PR #1, opened 2017-08-18
Creating CNAME file for domain

zechendorf:
This file should make the github page answer on https://www.btcpayserver.com/


zechendorf:
Sorry. My mistake. We need to do something different here.

PR #10, opened 2017-10-27
Checkout is now responsive, plus footer fixes

lepipele:
More info in commit messages


BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -1,638 +1,622 @@
...
+                                            <span i18n="">Scan</span>
+                                        </div>
+                                        <div class="payment-tabs__tab" id="copy-tab">
+                                            <span i18n="">Copy</span>
+                                        </div>
+                                        <div class="payment-tabs__slider"></div>
+                                    </div>
+                                </div>
+                                <div adjust-height="" class="payment-box">
+                                    <div class="bp-view payment scan" id="scan" style="opacity: 1;">
```
NicolasDorier: I think I needed to remove opacity 1 here, or the refund email view was broken.

lepipele: Yeah, I haven't got around to check how refund works... was busy with those other changes - but will review what you did and deal with this as time permits. Thanks for the fix though!

lepipele: I haven't got around to checking that refund view because of time constraints... but will review what you did and test. Thanks for the fix though @NicolasDorier

Issue #19, opened 2017-12-29
Non-exact BTC payments issue


Arkaroo:
Looks like it's not possible to request bitcoin payments in exact amounts.

Using container: nicolasdorier/btcpayserver:1.0.0.49

Repro steps:
1. Click "Create a new invoice" from the admin panel (e.g. http://localhost:49392/invoices)
2. Set currency to "BTC", set Amount to 1
3. Click "Create"
4. Click "checkout" on the resulting invoice
5. Observe that the amount charged is 1.00010000BTC, approx $1.50USD more than it should be

Expected outcome:
An invoice for exactly 1BTC

Actual outcome:
An invoice for slightly more than 1 BTC

Let me know if you need more info. Cool system by the way, I'm hoping to have it in production in a side-project I'm working on in a few months time :smile:

astupidmoose:
Do you have the "Network Fee" setting turned on in your store settings? 

This will add a .00010000BTC Charge for every expected payment, and I believe its turned on by default.

Arkaroo:
Ah, I did. In that case, I don't understand what the network fee setting is supposed to be doing. It's not setting a transaction fee, that still has to be set by the payee in their wallet, so what's it for?

But that's a separate, far less important issue. I can confirm that when the network fee checkbox under  http://localhost:49392/stores/{$store_id} is unchecked, the issue goes away.

astupidmoose:
> I don't understand what the network fee setting is supposed to be doing.

The network fee is being set to be able to cover some of the cost of sweeping the UTXO's from payments: https://support.bitpay.com/hc/en-us/articles/115002990803-Why-Am-I-Being-Charged-an-Additional-Network-Cost-on-My-BitPay-Invoice-

:)

NicolasDorier:
@Arkaroo imagine that the user send you a payment of 40 USD in one transaction. Because it creates an UTXO, as a merchant, when you will want to spend this UTXO (for example for sending the money to your exchange), you will have to pay around 120 bytes of fees. By default, those fees are passed to the customer, this represent the network fees.

If a user pay the invoice by sending 4 different bitcoin transactions, you will have 4 UTXO, so the fees network fees would be increased by 4 times.
Note that Signature aggregation (also called schnorr sig) would solve this problem, but this is not in bitcoin yet...

You can choose to disable this in your store settings if you don't want it.

Reopen if

NicolasDorier:
[...]

PR #23, opened 2018-01-03
Implementing indicator that shows total line row can be expanded

lepipele:
References:
https://github.com/btcpayserver/btcpayserver/issues/19#issuecomment-354586029
https://forkbitpay.slack.com/archives/C7M093Z55/p1514743072000005

### DIFF
--- a/BTCPayServer/Views/Invoice/Checkout.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout.cshtml
@@ -12,15 +12,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
     <title>BTCPay Invoice</title>
 
-    <environment include="Development">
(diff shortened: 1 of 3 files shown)


NicolasDorier:
it seems the arrow show only when I point the mouse on the header.

lepipele:
@NicolasDorier yeah I didn't want to change look too much because I loved how clean it is. Rather just give indication that row can be expanded. 

Give it a try for a day/two... if it doesn't stick all we need to do is remove `:hover` from CSS and arrow will be always visible.

NicolasDorier:
Another alternative if you like the clean look is to play on the opacity of the arrow (if possible) where by default it is super light, and when over more dark?

lepipele:
@NicolasDorier https://github.com/lepipele/btcpayserver/commit/ee3144f34a37d0adf8c9d99a27a8116a9914b9ef

PR #40, opened 2018-01-25
bootstrap.css deleted last line

Eskyee:
From bootstrap.css I removed the last line. and the problem is fixed
after a Failed to load resource: the server responded with a status of 404 (Not Found) debug


NicolasDorier:
I think you changed on your jungle and PRed on btcpayserver

Eskyee:
ok sorry my mistake ! I was in Visual Studio Code in jungle yes!

Issue #47, opened 2018-01-26
Automating bundling of JS/CSS files


lepipele:
We are currently not making distinction between different environments - development, staging, production. As described in #39 resources are linked statically... we link JS/CSS directly and that stays same between environments.

Ideally we want system where in development we would link full JS/CSS files which would help during development and debugging... but once project is deployed to production there should be automatic process of minimization and bundling of all referenced files so we ideally serve just one JS and one CSS file.

This way we will also resolve issue of versioning - if certain JS/CSS file is updated, server should automatically create new bundle with new version string (i.e. `http://btcpay-server-url.com/renderjs?version_guid`). That way you also prevent confusion with client browsers who'll always have latest versions of JS/CSS files (without requiring clients to CTRL+Refresh your page). See this url for example: https://stackoverflow.com/questions/16395595/mvc4-bundling-version-control

Eskyee:
yes I sure this is our best solution lepipele .. nice !

Eskyee:
https://docs.microsoft.com/en-us/aspnet/mvc/overview/performance/bundling-and-minification

NicolasDorier:
this link is not about asp.net core.

lepipele:
@NicolasDorier no worries - concept is pretty much the same and I've done it previously. Will be done with this by tomorrow evening and will update both you and @Eskyee

Eskyee:
I’m on my machine tonight!! Looking forward to see the Solution lepipele ;)

lepipele:
@Eskyee here is the first pass - comparison of HTTP requests and loading size when there is no bundling:

![non-bundled](https://user-images.githubusercontent.com/5191402/35494475-81af40c0-0480-11e8-97b6-724bdf6aa61f.png)

And then where bundling is activated:

![bundled](https://user-images.githubusercontent.com/5191402/35494481-88304f7a-0480-11e8-8972-01983023fd0f.png)

I want to make this automatic between different environments so I'll not create pull request yet, but rather when I am fully done with this. In the meantime feel free to take a break when it comes to that pull request you are submitting - I'll resolve that problem. There are some other errors with jquery.validation and I'll handle all that in this single issue.

Will keep both you and Nicolas updated.

Eskyee:
:)

lepipele:
For that Jquery Unobtrusive Validation... here is how it looked before:

[...]

PR #48, opened 2018-01-30
JS/CSS Bundling

lepipele:
Ref: https://github.com/btcpayserver/btcpayserver/issues/47

We now have new environment setting in `BTCPAY_BUNDLEJSCSS:true/false` that allows us to effortlessly turn bundling of JS/CSS files on or off.

Also, I've resolved jquery validation references... for now that's separate bundle but we can include it in main bundle if needed.

It would be good if we published this for more extensive testing on staging environment... if @Eskyee can help with ensuring there are no errors he mentioned - that would be great.


Eskyee:
👏 I will check for errors  later on this evening.  Good Job lepipele

BTCPayServer/Views/Shared/SyncModal.cshtml
```diff
@@ -77,7 +77,7 @@
 
 </div>
 
-<link href="~/vendor/animatecss/animate.css" rel="stylesheet" />
+@*<link href="~/vendor/animatecss/animate.css" rel="stylesheet" />*@
```
NicolasDorier: why not completely removed?

NicolasDorier:
@lepipele what about using webpack or stuff like which are the standard for web dev?

BTCPayServer/Views/Shared/SyncModal.cshtml
lepipele: Left it as reference where it's actually used... but feel free to remove it.

Related to this I am thinking about maybe modifying that BundleMinifier to support mixed bundles so it's easier to group files... like when you download bootstrap you have both CSS and JS files so it would be natural to add them together.

lepipele:
@NicolasDorier webpack and stuff this like that often require some extra installations on dev machines (like Node.js)... 

Visual Studio now does have better support for those scenarios (Node.js being bundled)... but you don't want to burden developers with ensuring their machine has all the tooling installed and configured properly.

BundleMinifier is NuGet dependency so it integrates effortlessly... plus it already has all we need.

Another thing I could do down the road is move our clientside (Javascript/CSS) dependencies to [Bower](https://bower.io/)... it is the only way to go as project grows and you need to figure out what requires what... much easier when you have Bower in that flow.

NicolasDorier:
Ok. Forget about Bower, I think it is deprecated now.

NicolasDorier:
BTCPAY_BUNDLEJSCSS should not exist, you should depend on the environment (Development or Production)

lepipele:
About Bower - lol... that clientside / javascript related stuff is now getting obsolete every 6 months. https://stackoverflow.blog/2018/01/11/brutal-lifecycle-javascript-frameworks/

About BTCPAY_BUNDLEJSCSS - it gives you finer degree of control so you can test bundling if needed and run configurations that are more complex like Production/Not bundled. Obviously bundling should be perfect... but my experience with production projects is that here and there some JS/CSS is done is such way that it'll break bundling and you want option to disable it regardless of environment.

NicolasDorier:
@lepipele ok. So given BTCpayServer would not work without bundling, you need to add the bundling registration (`services.AddBundles`) inside `AddBTCPayServer`.

Also, use an additional field in `BTCPayServerOptions` for configuring this, and add the command line documentation to `DefaultConfiguration`.

Because AddBTCPayServer do not have access to `BTCPayServerOptions` you need to to use a custom IConfigureOptions as specified [here](https://andrewlock.net/access-services-inside-options-and-startup-using-configureoptions/) for the `BundleOptions`.

By doing this it becomes possible to add BTCPay services to another application easily.

Eskyee:
Well done i check it out today !! lepipele 👍

NicolasDorier:
Missing:

1. Remove bundle from the commits
2. Add the bundles to gitignore

Eskyee:
@lepipele i be testing with Safari Web tools Develop & Safari Technology Preview just to see any problems, i also like to use https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fbtcpay-server-testnet.azurewebsites.net&tab=mobile.

Eskyee:
looking forward to the solution ;) you found.

NicolasDorier:
Merged

Eskyee:
Good Job @lepipele @NicolasDorier

PR #72, opened 2018-03-23
Add Brazilian Portuguese option as pt.js

rsandrade:


BTCPayServer/wwwroot/js/checkout/langs/pt.js
```diff
@@ -0,0 +1,56 @@
...
+    "Order ID": "Nº do Pedido",
+    "Return to StoreName": "Retornar à {{storeName}}",
+    // Invoice paid
+    "This invoice has been paid": "Essa fatura foi paga",
+    // Invoice archived
+    "This invoice has been archived": "Essa fatura foi arquivada",
+    "Archived_Body": "Por favor, contate o estabelecimento para informações e suporte"
+};
+
+    © 2018 GitHub, Inc.
```
NicolasDorier: what is that?

rsandrade: A Brazilian Portuguese translation. I did it in wrong place?

NicolasDorier: " © 2018 GitHub, Inc."

PR #76, opened 2018-03-23
Fix French translation

LinoxBE:


NicolasDorier:
That's why I hate french...

PR #85, opened 2018-03-28
Fixed cs translation

pajasevi:
I fixed a word that was not translated properly


lepipele:
Thanks! We have more two more strings incoming in new version so hopefully you'll be available to help with that.

pajasevi:
No problem, I'm always happy to help :)

PR #105, opened 2018-04-06
Migrating to Bootstrap4, CSS and JS cleanup

lepipele:


NicolasDorier:
I need to fix the colors still.
Will add theme.

NicolasDorier:
The POS is broken, will fix

PR #115, opened 2018-04-12
Fix Portuguese translation

iamvinny:


NicolasDorier:
It is brazilian portuguese, this might be the reason? @rsandrade ACK ?

rsandrade:
Most of suggestions are a better way to tell the same. I agree with the changes. :)

PR #122, opened 2018-04-14
Bootstrap theming 

lepipele:
Adding support for changing default theme for both Bootstrap and CreativeStart


NicolasDorier:
@rsandrade hey you LTC guys can create a bootstrap theme so we have something to play with? :D

lepipele:
@NicolasDorier I'll reach out to you in the morning on details - we still need to do some migration `style` and `class` migrations so that theming is fully supported. If @rsandrade wants to join us for talk that would also work - defining how we approach stuff that's not done with bootstrap but with other style is definitely something we need to go over.

lepipele:
OK @NicolasDorier - test and give it try - it should work well and both CSS files (bootstrap + creativestart) can now be replaced in policies.

NicolasDorier:
Man I am so happy to not have to deal with what you are doing... testing :p

NicolasDorier:
I suggest two things: 

* Rename the field "Creative Theme" to "Additional theme" because it might not be creative one right?
* Add a new tab in server (like "Theming"). Policies is not about design normally.

NicolasDorier:
So is it possible to keep the font for the title, but small font keep the old font?

lepipele:
@NicolasDorier ready to merge

Issue #133, opened 2018-04-23
Invoice window cutt-off (v1.0.1.88)


bitcoinshirt:
I don't think this happened in previous versions, but due to the length of the addresses, the invoice screen gets cut-off. 
It's just a minor inconvenience, a styling issue, and you can scroll the windows horizontally to view the entire address. 
I'm using default settings and styling, haven't done any customizations on my end.

![screenshot from 2018-04-23 09-57-26](https://user-images.githubusercontent.com/36959754/39113809-5cfdac7e-46cc-11e8-83b9-269d6c9220b1.png)
![screenshot from 2018-04-23 09-55-24](https://user-images.githubusercontent.com/36959754/39113810-5d1ba828-46cc-11e8-9522-1ad96aefd111.png)

NicolasDorier:
ping @lepipele  I checked it but could not find easy fix :(

NicolasDorier:
(also you took my no-wrap stuff!)

PR #134, opened 2018-04-23
Ellipsis when there is lots of info, preserving responsive tables

lepipele:
Ref: https://github.com/btcpayserver/btcpayserver/issues/133


NicolasDorier:
Is there a way to fix the first column length? 

image

lepipele:
We can set fixed width for first column so they align. You wanna do that?

NicolasDorier:
yes I think this would look better

lepipele:
Ready for review @NicolasDorier

NicolasDorier:
This is magic, it works!

PR #155, opened 2018-05-04
Payment Tolerance

Kukks:
This PR is not tested yet. 
This adds a new configurable feature to allow stores to accept invoices which are under paid to a specified percentage. The default is 0% so no store marks invoices as paid if not fully paid unless configured. 

@NicolasDorier  Currently I've decided to:

- send the paid event with the name `invoice_paidWithinTolerance`
- set the exceptionstatus to  `paidWithinTolerance`
- set the status to `paid`

I'm unsure if `invoice_paidWithinTolerance` and `paidWithinTolerance` breaks any bitpay compatibility, I'm still new to this api. :)

### DIFF
--- a/BTCPayServer/HostedServices/InvoiceWatcher.cs
+++ b/BTCPayServer/HostedServices/InvoiceWatcher.cs
@@ -96,11 +96,28 @@ namespace BTCPayServer.HostedServices
                     }
                 }
 
-                if (accounting.Paid < accounting.TotalDue && invoice.GetPayments().Count != 0 && invoice.ExceptionStatus != "paidPartial")
+                if (accounting.Paid < accounting.TotalDue && invoice.GetPayments().Count != 0)
                 {
-                    invoice.ExceptionStatus = "paidPartial";
-                    context.MarkDirty();
+                    if (invoice.PaymentTolerance > 0)
+                    {
+                        var tolerantAmount = (accounting.TotalDue.Satoshi * (invoice.PaymentTolerance / 100));
+                        var minimumTotalDue = accounting.TotalDue.Satoshi - tolerantAmount;
+                        if (accounting.Paid.Satoshi >= minimumTotalDue)
+                        {
+                            context.Events.Add(new InvoiceEvent(invoice, 1003, "invoide_paidWithinTolerance"));
+                            invoice.ExceptionStatus = "paidWithinTolerance";
+                            invoice.Status = "paid";
+                        }
+                    }
+
+                    if (invoice.ExceptionStatus != "paidPartial")
+                    {
+                        invoice.ExceptionStatus = "paidPartial";
+                        context.MarkDirty();
+                    }
                 }
+
+               
             }
 
             // Just make sure RBF did not cancelled a payment
--- a/BTCPayServer/HostedServices/InvoiceWatcher.cs
+++ b/BTCPayServer/HostedServices/InvoiceWatcher.cs
@@ -97,16 +97,11 @@ namespace BTCPayServer.HostedServices
                     }
                 }
 
-                if (!isPaid && invoice.GetPayments().Count != 0)
+                if (!isPaid && invoice.ExceptionStatus != "paidPartial" && invoice.GetPayments().Count != 0)
                 {
-                    if (invoice.ExceptionStatus != "paidPartial")
-                    {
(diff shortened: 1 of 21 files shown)


BTCPayServer/Models/InvoiceResponse.cs
```diff
@@ -178,7 +178,7 @@ public decimal Rate
         }
 
         //"exceptionStatus":false
-        //Can be `paidPartial`, `paidOver`, or false
+        //Can be `paidPartial`, `paidOver`, `paidWithinTolerance` or false
```
NicolasDorier: for compatibility purpose I prefer we don't add new stuff here

NicolasDorier:
I think it can be simplified.

I don't think we need event nor new exception status. (Especially the exceptionStatus which can break compatibility with bitpay)

The best is to just replace all `accounting.Paid < accounting.TotalDue` with `!accounting.Paid(tolerance)`.

Kukks:
Did the changes @NicolasDorier

NicolasDorier:
Actually you can simplify more: Use the `Money.Almost` methods ;)

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -523,6 +524,22 @@ public class PaymentMethodAccounting
         /// Total amount of network fee to pay to the invoice
         /// </summary>
         public Money NetworkFee { get; set; }
+
+        public bool IsPaid(double tolerance, out bool paidOver)
+        {
+            paidOver = false;
+            if (Paid < TotalDue)
```
NicolasDorier: This whole method can be summarized as following:

```
var totalDue = TotalDue - (TotalDue * (tolerance/100));
return Paid <= totalDue;
```

NicolasDorier: An alternative is to have a `PaymentMethodAccounting.MinimumTotalDue` and use it in place of of `TotalDue` everywhere in the Watcher! Note that this make everything stupid to review!

NicolasDorier:
I proposed some changes. But I can probably implement it myself after merge if it feel burdensome for you.

Though can you write a small test about this new feature?

Kukks:
I can do the changes and test tomorrow unless you can't wait, no worries :)

Kukks:
Added a very basic test and the MinimumTotaDue approach instead of IsPaid @NicolasDorier

BTCPayServer/HostedServices/InvoiceWatcher.cs
```diff
@@ -78,13 +78,13 @@ public void MarkDirty()
...
             {
-                if (accounting.Paid >= accounting.TotalDue)
+                if (accounting.Paid >= accounting.MinimumTotalDue)
                 {
                     if (invoice.Status == "new")
                     {
                         context.Events.Add(new InvoiceEvent(invoice, 1003, "invoice_paidInFull"));
                         invoice.Status = "paid";
-                        invoice.ExceptionStatus = accounting.Paid > accounting.TotalDue ? "paidOver" : null;
+                        invoice.ExceptionStatus = accounting.Paid > accounting.MinimumTotalDue ? "paidOver" : null;
```
NicolasDorier: this should be an exception where `TotalDue` is used. Else, a merchant will see the majority of people as `paidOver`

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -671,6 +676,10 @@ public PaymentMethodAccounting Calculate(Func<PaymentEntity, bool> paymentPredic
             accounting.Due = Money.Max(accounting.TotalDue - accounting.Paid, Money.Zero);
             accounting.DueUncapped = accounting.TotalDue - accounting.Paid;
             accounting.NetworkFee = accounting.TotalDue - totalDueNoNetworkCost;
+            accounting.MinimumTotalDue = new Money(Convert.ToInt32(Math.Ceiling(accounting.TotalDue.Satoshi -
```
NicolasDorier: should be a `ToInt64`

NicolasDorier:
awesome merging it, I can fix the two nits I pointed out.

NicolasDorier:
For reference 7872b3ec55607a103a378af6ba31c6cbb99f384c

PR #156, opened 2018-05-06
Added BTG Support

Vutov:
Support for BTG. We are going to create docker image for BTG Core and create PR for the docker repo as well.


BTCPayServer/BTCPayNetworkProvider.BitcoinGold.cs
```diff
@@ -0,0 +1,30 @@
...
+                CryptoCode = nbxplorerNetwork.CryptoCode,
+                BlockExplorerLink = NetworkType == NetworkType.Mainnet ? "https://explorer.bitcoingold.org/insight/tx/{0}/" : "https://test-explorer.bitcoingold.org/insight/tx/{0}",
+                NBitcoinNetwork = nbxplorerNetwork.NBitcoinNetwork,
+                NBXplorerNetwork = nbxplorerNetwork,
+                UriScheme = "bitcoingold",
+                DefaultRateRules = new[]
+                {
+                    "BTG_X = BTG_BTC * BTC_X",
+                    "BTG_BTC = bitfinex(BTG_BTC)",
+                    "BTG_USD = bitfinex(BTG_USD)",
```
NicolasDorier: Can you remove this? All new store have the rule `X_X = coinaverage(X_X)`. (coinaverage being overriden if the user selected another exchange in the dropdown)

Because coinaverage (https://apiv2.bitcoinaverage.com/indices/global/ticker/short) has BCHUSD already, this will work out of the box without this line.

NicolasDorier: This is unecessary I think:

Without this, `BTG_USD` would match rule `BTG_X` which would be `bitfinex(BTG_BTC) * BTC_X`
`BTC_X` would be expanded to BTC_USD which is well defined by the all exchanges that the user might have selected in his settings.

Vutov: Sounds reasonable, updated the PR

NicolasDorier:
cool, can you try with a real mainnet node?

BTCPayServer/BTCPayNetworkProvider.BitcoinGold.cs
```diff
@@ -0,0 +1,29 @@
...
+                UriScheme = "bitcoingold",
+                DefaultRateRules = new[]
+                {
+                    "BTG_X = BTG_BTC * BTC_X",
+                    "BTG_BTC = bitfinex(BTG_BTC)",
+                },
+                CryptoImagePath = "imlegacy/btg-symbol.svg",
+                LightningImagePath = "imlegacy/btg-symbol.svg",
+                DefaultSettings = BTCPayDefaultSettings.GetDefaultSettings(NetworkType),
+                CoinType = NetworkType == NetworkType.Mainnet ? new KeyPath("156'") : new KeyPath("1'")
```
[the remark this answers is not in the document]
Vutov: slip-0044 - https://github.com/satoshilabs/slips/blob/master/slip-0044.md . It should be the same for the ledger as well

Vutov:
It is working with mainnet node. Tried everything short of sending money to the address in the invoice

NicolasDorier:
ok then we are good to go! :)

PR #158, opened 2018-05-07
[WIP] Feature/ethereum

Kukks:
DO NOT MERGE

Just leaving this here as a progress report. 
@NicolasDorier I will open separate PRs to the segwit derivation path stuff as we talked about and will start extracting interfaces to reduce NBXplorer dependencies. 

This PR is to add in Ethereum support to BtcPayServer
The current approach looks at using the Web3 Client directly connecting to Infura hosted node services to avoid having to run an ethereum node which can be costly. 

But me and @NicolasDorier  talked it over and we may want to add in a full blown explorer to track Ethereum txs and accounts which also opens up easier support to ERC2xx token payments.

@anyone with Nethereum experience and Ethereum payment handling, please jump in with advice. :)


sharmaabhishek2207:
HI Kukks,
We want to use ethereum/ERC20 token in BTCPayerser for one of the use case. Could you please share the expected timeline when this feature will be available?

Regards.

Kukks:
no timeline unfortunately, would probably be q3 at the earliest

Kukks:
That would be nice, got a million things to do right now :P
ᐧ

Kukks:
https://btcpay.kukks.org/apps/3eFThAEycrZopgT5HeAMzZMFVWCa/crowdfund

image

pavlenex:
ahhahha finally

PR #168, opened 2018-05-11
[WIP] Feature/extended invoice

Kukks:
Depends on: https://github.com/MetacoSA/NBitpayClient/pull/6
I change the double types to be consistent with the new properties and fixed the unit tests accordingly.

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -12,6 +12,7 @@ using NBXplorer.Models;
 using NBXplorer;
 using NBXplorer.DerivationStrategy;
 using BTCPayServer.Payments;
+using NBitpayClient;
 
 namespace BTCPayServer.Services.Invoices
 {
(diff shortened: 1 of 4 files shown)


Kukks:
Havent updated the nuget reference in the csproj here as Im waiting for what would porbably be .20  of NBitPayClient

NicolasDorier:
Can you rebase? I incorporated the decimal changes.

Kukks:
Should be with latest now

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -396,10 +410,22 @@ public InvoiceResponse EntityToDTO(BTCPayNetworkProvider networkProvider)
                     dto.BTCDue = cryptoInfo.Due;
                     dto.PaymentUrls = cryptoInfo.PaymentUrls;
                 }
+
 #pragma warning restore CS0618
                 dto.CryptoInfo.Add(cryptoInfo);
+
+                dto.PaymentSubtotals.Add(cryptoCode, subtotalPrice.ToDecimal(MoneyUnit.Satoshi));
+                dto.PaymentTotals.Add(cryptoCode, accounting.TotalDue.ToDecimal(MoneyUnit.Satoshi));
```
NicolasDorier: The number of satoshi in decimal? are you sure?

Kukks: their docs stated that we need the amounts in the smallest factor:
```
paymentSubtotals
Equivalent of price for each of the supportedTransactionCurrencies. This value does not include the minerFees. The key is the currency and the value is an amount indicated in the smallest possible unit for each supported cryptocurrency (e.g satoshis for BTC and BCH)

paymentTotals
The total amount that the purchaser should pay. This is like paymentSubtotals but with the minerFees included. The key is the currency and the value is an amount indicated in the smallest possible unit for each supported cryptocurrency (e.g satoshis for BTC and BCH)
```

NicolasDorier: so this should be long, not decimal.

NicolasDorier: Yes, but as it is like this, it will be incompatible with Bitpay clients who expects an integer.

Kukks:
Just updated to reflect the new types in the invoice. 
I still have these placeholders which I'm unsure of how to fill out
```
            //TODO: Populate dto.AmountPaid
            //TODO: Populate dto.MinerFees
            //TODO: Populate dto.TransactionCurrency
```

Kukks:
Ok, I will just remove the todo and leave a note. Anything else to do?

NicolasDorier:
Yeah, can you add some line to an existing test to check if those fields are correctly filled?

Issue #148, opened 2018-04-30
Show fiat value on Checkout page


NicolasDorier:
https://twitter.com/riggsfinn/status/990922094853611521

rockstardev:
Here is first iteration - looking forward to crew's feedback:

![invoice with fiat 1 btc](https://user-images.githubusercontent.com/5191402/39851004-0fd52e0c-53da-11e8-8ab3-c92da95ad641.png)

![invoice with fiat 2 ltc](https://user-images.githubusercontent.com/5191402/39851005-0fe2f0fa-53da-11e8-893a-5e55031a87cf.png)

Like to commit: https://github.com/rockstardev/btcpayserver/commit/be1128a886aaf8ff102a7ca83e1151dbe56fa485

NicolasDorier:
let's commit this!

NicolasDorier:
For reference, before this change:

![image](https://user-images.githubusercontent.com/3020646/39909350-639c2bea-552d-11e8-9137-25af399f285a.png)

rockstardev:
And here is how it looks when its in item details: 

![fiat](https://user-images.githubusercontent.com/5191402/39909816-0b715332-54ba-11e8-9393-9cbb4d040db5.png)

I'm also curious what would be best term in item details... Fiat Total is basically - "Order Amount + Network Fee in Fiat"... but that one would be mouthful ;).

Looking forward to comments

hubiktomas:
I think that the original version where there was total amount and fiat total under it with all the details in the lower part was better.

hubiktomas:
Also can there be more than 1 transaction? Is it necessary to include "1 transaction x " in the Network Cost?

NicolasDorier:
yes there can be more transaction. If you pay an invoice with 2 transactoin you need to pay two time the network cost.

hubiktomas:
So if I pay lower than required amount while having this option to include network cost in the total enabled, the total will be increased after this first payment?

Eskyee:
good job, rockstardev I like it, 
total amount and fiat total under it with all the details in the lower part looks nice. I agree

NicolasDorier:
@hubiktomas yes

hubiktomas:
Ok, then I think that the original proposal with total amount and fiat total under it with all the details in the lower part was the best one. Thanks for clarification.

astupidmoose:
I prefer this:
[...]

PR #170, opened 2018-05-11
Display fiat value of invoice during checkout

rockstardev:
As requested by #148

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -239,9 +239,9 @@ namespace BTCPayServer.Controllers
                 CustomCSSLink = storeBlob.CustomCSS?.AbsoluteUri,
                 CustomLogoLink = storeBlob.CustomLogo?.AbsoluteUri,
                 BtcAddress = paymentMethodDetails.GetPaymentDestination(),
-                OrderAmount = (accounting.TotalDue - accounting.NetworkFee).ToString(),
                 BtcDue = accounting.Due.ToString(),
-                FiatDue = FiatDue(accounting.Due, paymentMethod),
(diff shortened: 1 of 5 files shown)


hubiktomas:
Still liked this version https://github.com/btcpayserver/btcpayserver/issues/148#issuecomment-387942990 more as it looked more consistent - all totals together and additional info under. But if the majority is for this one, then ok. It will not stop me from using BTCPay Server. :)

Kukks:
Maybe it can be configurable?  Not for this PR, but what if we allow stores to accept a template file with handlebars bindings or some simple templating engine?

rockstardev:
OK, after extended discussion with @NicolasDorier and taking all points here in consideration I've removed "Order Amount in Fiat" (notice my showoff use of `git rebase` that made that commit... poof... disappear ;)).

So now what we have is - if Invoice is defined using Fiat (like USD) then we will show that value below "Order Amount" this way, similar to rate:

showing fiat value

If however invoice is defined using Crypto Currency as a source... then we don't show that or exchange rate (although it would be funny to excuse DOGE for this and show 1Đ=1Đ):

not showing fiat for invoices that are in crypto

I hope this allows us to finally close this issue and that merchants who wanted fiat display will be satisfied ;).

@NicolasDorier feel free to merge at your convenience when you test locally on how new layout feels.

NicolasDorier:
let's merge this, we will adapt if there is complaints

hubiktomas:
The exchange rate should not be displayed if the selected currency != order currency. But if I have order in BTC and select LTC in the BTCPay, there should be FX rate and order amount in the original currency displayed.

See the case below - order for 1 BTC, LTC selected.

image

rockstardev:
@hubiktomas join on #179

PR #174, opened 2018-05-13
Support Monacoin.

cryptcoin-junkey:
It's roughly checked if payment transaction was stored into the blockchain.
But not enough tested. Please note that it is beta status.

In addition, I've already ported Elements' c-lightning to Monacoin.
But I've not checked yet whether it runs with BTCPayServer.


NicolasDorier:
Can you check POLIS?
You need to define a default way to calculate MONA rates, so that new users don't have to setup anything in their store's rate settings to make MONA work.

cryptcoin-junkey:
I've not check POLIS/DOGE yet. And I checked it's possible to get MONA/JPY by using rate by using information from Zaif. I see, I'll add some more patches later.

PR #177, opened 2018-05-15
[WIP] Docker compose visual studio project

Kukks:


NicolasDorier:
I don't understand why you need a docker-compose, there is already one for development purpose.

The Dockerfile in tests is meant to run the tests in this dev docker-compose. Can you use alpine image as in the root of this project?

NicolasDorier:
AH got it! You want to allow F5 debugging directly inside a container as well? That's cool.

NicolasDorier:
we probably what this as well https://www.reddit.com/r/dotnet/comments/8ibvup/manually_configuring_aspnet_core_docker_debugging/

Kukks:
Yup exactly, this allows you to debug while running inside a container. Also allows you to edit the cshtml files while running.  It can feel a bit sluggish to use sometimes but with this it will allow new users to just really just do dev work and f5. A solution clean deletes all the created images but volumes still persist so you would need to do a `docker volume prune` or deleting the docker volumes manually. 

I need to fix some issues as they arent working completely but it shouldn't be much work

Kukks:
killing this pr in favor of a new one with less chanegs

Kukks:
https://github.com/btcpayserver/btcpayserver/pull/196

PR #179, opened 2018-05-16
Showing exchange rate for cryptos

rockstardev:
Ref: https://github.com/btcpayserver/btcpayserver/pull/170#issuecomment-389462155

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -85,7 +85,7 @@ namespace BTCPayServer.Controllers
                 {
                     cryptoPayment.Address = onchainMethod.DepositAddress;
                 }
-                cryptoPayment.Rate = FormatCurrency(data);
+                cryptoPayment.Rate = ExchangeRate(data);
                 cryptoPayment.PaymentUrl = cryptoInfo.PaymentUrls.BIP21;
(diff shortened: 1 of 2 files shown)


rockstardev:
Here is how it looks with changes:

showing exchange rate

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -314,18 +323,13 @@ public static string FormatCurrency(decimal price, string currency, CurrencyName
...
-                return null;
-            }
 
-            return FormatCurrency(productInformation.Price, productInformation.Currency, _CurrencyNameTable);
+            if (_specialCryptoFormat.Contains(currency))
+                return price.ToString("C", provider);
+            else
+                return price.ToString("C", provider) + $" ({currency})";
         }
+        private static readonly string[] _specialCryptoFormat = new[] { "BTC", "LTC" };
```
NicolasDorier: wat? why only BTC and LTC?

NicolasDorier: Fixed https://github.com/btcpayserver/btcpayserver/commit/eeb522fe7d852e610e1c4b0ded7fc56ffa229ce1

rockstardev: Yeah, much better that way - was going to ask you if you were already maintaining that data somewhere.

PR #186, opened 2018-05-20
[WIP] LND integration

rockstardev:


NicolasDorier:
There is no end to end test (CanSendLightningPayment)

junderw:
@rockstardev Hey I was just wondering what the status of this PR is currently and if there's anywhere I can help?

rockstardev:
@junderw getting there - https://twitter.com/codeoverwhelmin/status/1007753403123097603

Will let you know as soon as we go production... I would ideally want this to work with any LND - and we need to cleanup some things related to that... will talk with @NicolasDorier on it.

NicolasDorier:
@rockstardev this is without the connectionString refactoring we talked about right?

The connectionString refactoring should be done in a separate PR I think. The "legacy" format should be used only for deserializing stored data in the DB. The legacy format should otherwise not work when set inside the textbox.

rockstardev:
Yeah, that's without refactoring. After going through code I saw that the main thing is how we save to database, meaning that `LightningSupportedPaymentMethod.cs` is the most important class. Now it ends up being:

  {
    "BTC": "tpubD93CJNkmGjLXnsBqE2zGDqfEh1Q8iJ8wueordy3SeWt1RngbbuxXCsqASuVWFywmfoCwUE1rSfNJbaH4cBNcbp8WcyZgPiiRSTazLGL8U9w",
    "BTC_LightningLike": {
      "CryptoCode": "BTC",
      "PaymentId": {
        "IsBTCOnChain": false,
        "CryptoCode": "BTC",
        "PaymentType": 1
      },
      "LightningChargeUrl": "http://127.0.0.1:53280/",
      "Username": "lnd",
      "Password": "lnd",
      "Macaroon": "xxx",
      "Tls": null
    }
  }

`LightningConnectionString` only play as to how you parse user supplied string and convert it into `LightningSupportedPaymentMethod`. I think it would make sense to refactor dependencies and push `ConnectionType` there. In that scenario you would switch to using `LightningSupportedPaymentMethod` for most of the stuff... and `LightningConnectionString` would keep being only for going from/to one-liners. We use one-liners quite often even for configuring docker, like:

      TEST_MERCHANTLIGHTNINGD: "/etc/merchant_lightningd_datadir/lightning-rpc"
      TEST_CUSTOMERLIGHTNINGD: "/etc/customer_lightningd_datadir/lightning-rpc"
      TEST_MERCHANTCHARGE: http://api-token:foiewnccewuify@lightning-charged:9112/

So I was debating whether to change that because for LND we would basically choose between:

      TEST_MERCHANTLND: http://lnd:lnd@127.0.0.1:53280?macaroon={macaroon}&tls={tls}

And something like:

      TEST_MERCHANTLND: TYPE=lnd;URI=http://127.0.0.1:53280;MACAROON={macaroon};TLS={tls}

While Clightning and Charge would be:

      TEST_MERCHANTLIGHTNINGD: TYPE=clightning;URI=unix://etc/merchant_lightningd_datadir/lightning-rpc
      TEST_CUSTOMERLIGHTNINGD: TYPE=clightning;URI=unix://etc/customer_lightningd_datadir/lightning-rpc
      TEST_MERCHANTCHARGE: TYPE=charge;URI=http://lightning-charged:9112;USERNAME=api-token;PASSWORD=foiewnccewuify

It definitely makes it more obvious, less black magic and better documents what we are doing... but, do we want more verbose format? Again, this part of discussion only influences one-liner format... main topic is format of `LightningSupportedPaymentMethod.cs` class and how it gets loaded/saved from database.

NicolasDorier:
We should not need to add any field it should be something like `TYPE=clightning;URI=unix://etc/merchant_lightningd_datadir/lightning-rpc` saved into  `LightningChargeUrl`. The name of the field is a legacy details, if I could rename the field it would be `LightningConnectionString`.

NicolasDorier:
@rockstardev I need to clean stuff, normally the macaroon should be mandatory, but the tests do not support it now.

rockstardev:
Sure thing @NicolasDorier - I expect refactoring, code needs to be integrated into whole... it's your brainchild man so do it the way you see makes most sense.

rockstardev:
During testing I wanted macaroon to be optional, but if want to make it mandatory just go:

https://github.com/btcpayserver/btcpayserver/blob/25dbf6445f88eeaf64050b418fdfe94a28a41740/BTCPayServer/Payments/Lightning/Lnd/LndSwaggerClientCustomHttp.cs#L35

And you can remove that default `byte[] grpcMacaroon = null` plus you can check for null and throw exception if class usage is attempted without macaroon.

Also if we'll use our docker container for first integration we need to take look into how be build image: https://github.com/btcpayserver/lnd/blob/dockerfile/docker-btcpayserver/start-lnd.sh#L53

NicolasDorier:
let's move to https://github.com/btcpayserver/btcpayserver/issues/210
I was thinking enforcing the macaroon just at the connectionstring level.

PR #191, opened 2018-05-26
Localizing datetimes based on browser's timezone

rockstardev:
Ready to merge


NicolasDorier:
`ToBrowserAgo` is not needed. A "time ago" does not depends on local time. Actually making it dependent on local time might create issue during daylight time change.

NicolasDorier:
Oh actually this is not about browser time.

NicolasDorier:
Will review and test more tonight, sorry for delay!

NicolasDorier:
I merged it but do you think we should add `(local)` to be more explicit?

PR #196, opened 2018-05-28
add docker compose projects and configure

Kukks:
Now you have a true f5 and done dev experience. :D This PR introduces visual studio's native docker support by creating a visual studio docker compose project.

Left the original compose files and dockerfiles in place to make sure I don't break any automated builds you have set up.  I'm also intentionally not using alpine images as visual studio seems to crash and burn with those .net specific images. 

There is room for improvement which we can do later:
* use alpine
* use a .env file for all the config in the compsoe file
* more launch profiles for a full local dev experience with a reverse proxy too if wanted
* test launch profile 
* Clean up existing docker compose files

### DIFF
--- a/BTCPayServer/Dockerfile
+++ b/BTCPayServer/Dockerfile
@@ -1,0 +1,31 @@
+FROM microsoft/dotnet:2.1-aspnetcore-runtime AS base
+
+
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
+
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
+
+WORKDIR /app
+
+RUN mkdir /datadir
+ENV BTCPAY_DATADIR=/datadir
+VOLUME /datadir
+
+EXPOSE 80
+
+FROM microsoft/dotnet:2.1-sdk AS build
+WORKDIR /src
+COPY BTCPayServer/BTCPayServer.csproj BTCPayServer/
+RUN dotnet restore BTCPayServer/BTCPayServer.csproj
+COPY . .
+WORKDIR /src/BTCPayServer
+RUN dotnet build BTCPayServer.csproj -c Release -o /app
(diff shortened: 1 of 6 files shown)


BTCPayServer/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY BTCPayServer/BTCPayServer.csproj BTCPayServer/
+RUN dotnet restore BTCPayServer/BTCPayServer.csproj
+COPY . .
+WORKDIR /src/BTCPayServer
+RUN dotnet build BTCPayServer.csproj -c Release -o /app
+
+FROM build AS publish
+RUN dotnet publish BTCPayServer.csproj -c Release -o /app
+
+FROM base AS final
```
NicolasDorier: Today I learnt something.

NicolasDorier:
Interesting. Does tests continue to work though? 
Does the old docker-compose is still necessary?

Kukks:
I left the old docker compose and dockerfile files there on purpose due to the issues with alpine images. You should still build using your existing setup and test with docker-compose up tests.

NicolasDorier:
Can't we make the tests work on this docker-compose so we don't have two ?

NicolasDorier:
I am just worried that you need to turn off one docker-compose and on the other one when you want to switch between F5 debug and debugging the tests.

Kukks:
Yup, I will add it in as well but just be aware we will lose alpine on the tests so the environments deployed and the dev  one may differ on the image base

NicolasDorier:
that's fine

Kukks:
Added in a test docker project but having issues with its dockerfile. Maybe you can take a look at what needs to be done to get it to work @NicolasDorier ?

Kukks:
Will close this PR. The stability of these docker projects aren't the best and unless you expand this repo to have more standalone services there's no need for more magic

PR #197, opened 2018-05-28
[WIP] Bitpay rates api

Kukks:
This PR introduces a bitpay compatible rates api.
```
/rates
/rates/{currencyCode}
/rates/{baseCurrencyCode}/{currencyCode}
```

I've added a new property to `ISupportedPaymentMethod` so that a payment method may specify which currency code it handles. This way the `/rates` & `/rates/{currencyCode}`  endpoint will map all available payment methods to available currency pairs to return.
E.g. a btcpayserver instance with btc and ltc configured would return pairs of "BTC_LTC" and "LTC_BTC" under `/rates` and "BTC_LTC" under `/rates/btc`.

`rates/x/y`  would map directly to the specified values.  

The only incompatibility left is that we still need to specify the store id but maybe that is fine and should be handled from the clients directly (see https://github.com/btcpayserver/btcpayserver/pull/197 )


NicolasDorier:
I updated the NBitpayClient package, can you add some tests?

Kukks:
Will do

Kukks:
Trying to write some UTs but the test execution is taking forever. Does it usually take >10mins to execute?
image

NicolasDorier:
mmh this is tricky. Unlike Bitpay, we have no way to enumerate all available rates from a cryptoCode.

NicolasDorier:
I am unsure if there is a nice way to get it like Bitpay. The root of the problem is that we have no generic way to enumerate the rates.

Kukks:
@NicolasDorier I'm going to go ahead and get the final touches on this done. Do you have any requirements to add?

NicolasDorier:
it seems good to me as long as it does not break stuff, ready to merge?

Kukks:
ok we can merge it in @NicolasDorier 

So to summarize, 
This adds in the bitpay compliant rates api with one variation to the original api: the request must have a storeid query paramater or the request must be authenticated.

I will add open up a PR to NBitPayClient to send those requests as authenticated.

Kukks:
Actually give me a min, need to verify some stuff

Kukks:
Added in some small changes and done from my end.

BTCPayServer/Payments/ISupportedPaymentMethod.cs
```diff
@@ -13,6 +13,7 @@ namespace BTCPayServer.Payments
     /// </summary>
     public interface ISupportedPaymentMethod
     {
+        string CryptoCode { get;}
```
NicolasDorier: Why? you can use `PaymentId.CryptoCode`.

Kukks: Ah, did not know exactly what PaymentId was until now, removed it now

PR #221, opened 2018-07-12
[Translation Dutch]: improve translation

romanornr:
some inconsistency between english an dutch translations.

English: Altcoins 
Dutch: Alternative cryptocurrencies 

We can remove it and make it altcoins too.
Dutch people use the word "altcoins" more often. It only leads to confusing. 

English: please enter
Dutch: Thank you enter 

That doesn't sound right. Actually the "please" in dutch can be skipped. 

some more inconsistency in translation
English: Network
Dutch: Blockchain

I changed it to Network, so it's the same as English.

I also changed some old dutch words to a bit more modern dutch. almost nobody uses those old words and it can lead to confusion. 

I did not edit this but
In dutch for the word "you" we have "je" but old fashioned we say "u" sometimes to people one respects like parents, teachers, customers etc.
This is old fashioned but people in Belgium who speak dutch too only use "u".

Might be worth to think about changing.

### DIFF
--- a/BTCPayServer/wwwroot/checkout/js/langs/nl.js
+++ b/BTCPayServer/wwwroot/checkout/js/langs/nl.js
@@ -8,7 +8,7 @@ const locales_nl = {
     "Contact_Body": "Bedankt om je email adres in te vullen voor een mogelijke opvolging. We contacteren je indien er een probleem optreedt.",
     "Your email": "Je email adres",
     "Continue": "Verdergaan",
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Ping @LinoxBE , ACK?

LinoxBE:
Most changes are mainly differences between Flemish and Dutch. But the changes look good. Thanks @romanornr !

ACK!

PR #224, opened 2018-07-13
Added Macaroons to 'v1/invoices/subscribe' call

Vutov:
This PR is more for reference what I've done to make it work. Probably it is not that good of an idea to expose `_Authentication`. Also there is `public async System.Threading.Tasks.Task<LnrpcInvoice> SubscribeInvoicesAsync(System.Threading.CancellationToken cancellationToken)` in `LndSwaggerClient` that might do the trick, haven't tried it.

The result of this 'fix' is successfully payed BTG LND invoice:
https://www.screencast.com/t/Z9EVhLQz
https://www.screencast.com/t/jEo6pIi2h


NicolasDorier:
Closing this, I took this but replaced public with internal.

PR #234, opened 2018-07-27
Added Dash support

BenjaminNitschke:
Please merge, let me know if anything else is needed, thanks :)


BTCPayServer.Tests/RateRulesTest.cs
```diff
@@ -18,6 +18,7 @@ public void CanParseRateRules()
             builder.AppendLine("// Some cool comments");
             builder.AppendLine("DOGE_X = DOGE_BTC * BTC_X * 1.1");
             builder.AppendLine("DOGE_BTC = Bittrex(DOGE_BTC)");
+            builder.AppendLine("DASH_BTC = Bittrex(DASH_BTC)");
```
NicolasDorier: remove this

BTCPayServer.Tests/RateRulesTest.cs
```diff
@@ -29,6 +30,7 @@ public void CanParseRateRules()
                 "// Some cool comments\n" +
                 "DOGE_X = DOGE_BTC * BTC_X * 1.1;\n" +
                 "DOGE_BTC = bittrex(DOGE_BTC);\n" +
+                "DASH_BTC = bittrex(DASH_BTC);\n" +
```
NicolasDorier: remove this

NicolasDorier:
Have you tested it?

BTCPayServer.Tests/RateRulesTest.cs
BenjaminNitschke: sure, it is removed again. it was tested and had no impact on the existing code, it was just to make sure the used string from InitDash would parse just fine.

BenjaminNitschke:
Yep, we tested it, will try add a store on https://MyDashWallet.org next week based on this code and BTCPayServer

NicolasDorier:
Nice, if you can make a docker image for `btcpayserver-docker` I can include it in one-click deploy.

BenjaminNitschke:
sure, will try, have a nice weekend.

Aranad:
Is dash working? I did
export BTCPAYGEN_CRYPTO2="ltc"
export BTCPAYGEN_CRYPTO3="dash"
and re-ran the install and LTC has activated fine, but no sign of dash...?

NicolasDorier:
@Aranad the dash community need to provide their docker image.

Aranad:
Thanks @NicolasDorier, could you recommend a link that would be a good starting point for how to get started creating such an image? I'd like to have a go at making one, also I've pointed here from a Dash forum thread about needing more payment processors.

BenjaminNitschke:
@Aranad @NicolasDorier any help on how to proceed generating the required docker image would be great, last time we tried creating a simple example website we got stuck pretty quickly as there was no easy way to test or setup things.

Aranad:
@BenjaminNitschke  I don't know the code at all, but I've gone through the process of setting up a server using docker from [this tutorial](https://medium.com/@BtcpayServer/hosting-btcpay-server-for-cheap-2b27761fdb9d) which was very straight forward and I was up and running in 20min.

I then added support for LTC and DASH by adding the following environment variables and re-running install,
```bash
export BTCPAYGEN_CRYPTO2="ltc"
export BTCPAYGEN_CRYPTO3="dash"
```
But only the LTC worked, when I looked at the output from the installation it seemed to indicate that everything required for dash is ready except that there is not docker image for dashd itself. I don't know what's required for that part, I'm guessing it's more complicated than just finding any existing docker image that runs a dashd...?

Aranad:
Thanks a lot @NicolasDorier for putting all this time in as well, it's really amazing work which is incredibly beneficial to the ecosystem, which I was I was inspired to make a donation to the project.

rockstardev:
@Aranad you are right on that basically docker image should basically just run `dashd`. @BenjaminNitschke I can help you out with preparing docker image - hit me up on Slack so we can agree on time and we can do remote desktop session where I can show you how to setup everything with Visual Studio.

Aranad:
There are a few existing dash docker images already such as [this official one](https://hub.docker.com/r/dashpay/dashd/), [thelazier](https://hub.docker.com/r/thelazier/docker-dashd/) and [zebralucky](https://hub.docker.com/r/zebralucky/dashd/), but all the existing altcoin docker images in btcpay seem to be based on the [amacneil/bitcoin](https://hub.docker.com/r/amacneil/bitcoin/) image.

rockstardev:
Well, for now we can go with the official image or whichever @BenjaminNitschke prefers.

Down the road I think we need to clean up hosting altcoin images - depend on official ones whenever we can... or if BtcPayServer depends on certain image we host it on https://hub.docker.com/r/btcpayserver/ like we do for LND.

But for now we can just point to image Benjamin decides on and add Dash fragment here referencing image: https://github.com/btcpayserver/btcpayserver-docker/tree/master/docker-compose-generator/docker-fragments

PR #235, opened 2018-07-30
Feature/payment method toggle

Kukks:
This PR will allow you to enable/disable payment methods on each store. Benefits include you not having to erase your derivation scheme each time you want to toggle a crypto payment method. Works for both LightningLike and BtcLike

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -505,7 +505,7 @@ namespace BTCPayServer.Controllers
                 return View(model);
             }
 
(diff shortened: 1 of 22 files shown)


NicolasDorier:
Cool, do you mind if I adapt your PR with the design I talked you about? (Using Filters saved at Store level)

Kukks:
go for it but I want to be able to add in a way to toggle these at an invoice level and then also at an admin level later on :)

NicolasDorier:
It is precisely because you want to do that later that I want to change your PR.

NicolasDorier:
I refactored into d480be925b4909a97b4bc7444e09f54ee1dcf31b and added tests in 7a32f692d1e256a7c8373cbe8f7632f4f1559b76

PR #252, opened 2018-08-23
Update Checkout.cshtml

Onurrr:
Center text & select for invoice page

result:
image


romanornr:
+1

NicolasDorier:
I would have swear that I tried it before but it did not work

Issue #258, opened 2018-08-30
Pay button doesn't work when user not authenticated


mluczak:
Pay buttons should work for users who don't have an account on the btcpayserver, instead the user is directed to the login/sign up form.

NicolasDorier:
Yes I saw, will fix this today or tomorrow.

NicolasDorier:
@rockstardev I also wanted a small secret ID in the button. Because the button is allowing anyone knowing the storeId to create infinite amount of invoice.

mluczak:
I guess it would be a lot of work, but it would be nice if the invoice was generated after a successful payment for pay buttons. This way we could easily use btcpayserver pay buttons in replacement of paypal pay buttons on simple commerce websites without filling the database with unpaid invoices. Even nicer, this behaviour could be configurable for each button... It would I believe require the addition of a draft invoice object (or just flag in the current object?) which could be removed by some garbage collecting routine periodically if expired.

NicolasDorier:
I could garbage collect, but this is not easy. Some users are paying after expiration because they keep the address and pay days after. If I delete the invoice, the merchant can't understand what happened.

ketominer:
Thank you for this fix! Small additional question, would it be possible to have a "cancel" button on the payment page, so the user could voluntarily cancel the invoice? As expected, we end up having tons of expired invoices of people who clicked on the button just to see what it does. Another (more complicated) solution would be to pass some session_id from the button to btcpay so if the same user clicks again on the same button on the same website, we could propose him to finish a previously initiated invoice instead of creating a new one (or show him that he already paid if he did).

ketominer:
As I understand it, it will fast create a lot of new addresses to watch (multiplied by the number of coins supported on a particular instance?) and impact scaling. I will think a little more about it and propose something in a new issue.

PR #260, opened 2018-08-30
Public controllers for PayButton and Apps / POS

rockstardev:
Handling #258


NicolasDorier:
@rockstardev I also wanted a small secret ID in the button. Because the button is allowing anyone knowing the storeId to create infinite amount of invoice.

This can be like "ApiKey" which can be rotated, which would be inside the Forms.

NicolasDorier:
OR, it can be done by adding a screen to ask the user to activate Pay Button before using it.

rockstardev:
Hey @NicolasDorier, I've added Enabling/Disabling of PayButton... review commits

BTCPayServer/Controllers/PublicController.cs
```diff
@@ -30,9 +30,15 @@ public async Task<IActionResult> PayButtonHandle(string storeId, [FromForm]PayBu
             var store = await _StoreRepository.FindStore(storeId);
             if (store == null)
                 ModelState.AddModelError("Store", "Invalid store");
+            else
```
NicolasDorier: `else` uneeded

NicolasDorier:
testing it right now

BTCPayServer/Controllers/PublicController.cs
rockstardev: If `store` is null you'll trigger `store.GetStoreBlob()` line and NullException will get thrown. That's why I added `else`.

NicolasDorier:
Merged with some of my changes on top

PR #267, opened 2018-09-05
Changelly Support

Kukks:
Quick fix for the current Shapeshift mess. This PR replaces the shapeshift option with Changelly.  The user currently still needs to input the correct amount but I'll post a  PR later that will fix this

**UPDATE:**

No longer a quick fix :D

This PR introduces support for Changelly.  It adds a new supported method in the store data to follow the same style as other payments and also introduces a new Payment Type `ThirdParty`

The Store settings UI has a new table at the bottom that holds third party payment methods. Chnagelly support can be toggled per store and needs to be configured( api key, api secret, enabled, merchant id).

The Checkout form has also gotten some UI work done( thanks @rockstardev  for helping me with some issues). 
There is a new Vue Component stored in a partial cshtml file `ChangellyComponent` which holds the Vuejs logic for it and then a partial cshtml to define the VUejs template used in CheckoutBody as `ChangellyTemplate`

Since Changelly does not have fixed rates, there is a new controller `ChangellyController` that takes a `storeid`  which  queries the list of currencies available from Changelly and also a GET for retrieving the current rate of a specific currency pair.

The UI works like this:
* Loads Changelly Currencies from btcpay.com/changelly/{storeid}/currencies
* Go to Conversion Page
* Select Currency
* Sends request to btcpay.com/changelly/{storeid}/getExchangeAmount and specifies currency pairs and amount of 1
* Calculate rate and add Changelly's 0.5% fee to result
* Send another request to btcpay.com/changelly/{storeid}/getExchangeAmount and verify that the computed amount is enough to cover the transacttion
* Generate url and show pay button


NicolasDorier:
I think we can't use changelly, @rockstardev already checked. With changelly you can never be sure about the rate you will eventually get ahead of time.

Kukks:
True but I can add in a way for store owners to configure an api key and then use https://changelly.com/developers#estimated-exchange

Kukks:
But at least for now, this adds in a possible way to pay for the invoice with unsupported currencies.

NicolasDorier:
it is does not, because if the user pays, then the amount going to the invoice does not match what the merchant expect.

Kukks:
Yes I plan to solve that. 

What do you think about making the changelly thing into an actual `PaymentType`? Was thinking of creating a `PaymentType.ThirdParty` and then having a `ChangellySupportedPaymentMethod`. What this does it you specify which enabled `BTCLikepaymentMethod` you would like the destination currency to be and then configure any other changelly specific settings in it
That way it's more consistent and I can do the computation from their api on how much the amount should be and whatnot

NicolasDorier:
This might work. Though I don't think using changelly should be the solution, they will eventually have the same issue.

Kukks:
Not the long term solution but I want to try this as a way to test
expansion of the payment methods :)

BTCPayServer/wwwroot/checkout/js/core.js
```diff
@@ -21,8 +21,11 @@ function resetTabsSlider() {
 }
 
 function onDataCallback(jsonData) {
-    // extender properties used 
+    // extender properties used
     jsonData.shapeshiftUrl = "https://shapeshift.io/shifty.html?destination=" + jsonData.btcAddress + "&output=" + jsonData.paymentMethodId + "&amount=" + jsonData.btcDue;
+    jsonData.changellyUrl = "https://changelly.com/widget/v1?auth=email&from=ETH&to=" + jsonData.paymentMethodId + "&address=" + jsonData.btcAddress;
```
NicolasDorier: `from=ETH` ? why? can't it be blank?

Kukks:
I started changing this pr Yesterday so it's not complete yet

Kukks:
@NicolasDorier  @rockstardev 

No longer a quick fix :D

This PR introduces support for Changelly.  It adds a new supported method in the store data to follow the same style as other payments and also introduces a new Payment Type `ThirdParty`

The Store settings UI has a new table at the bottom that holds third party payment methods. Chnagelly support can be toggled per store and needs to be configured( api key, api secret, enabled, merchant id).

The Checkout form has also gotten some UI work done( thanks @rockstardev  for helping me with some issues). 
There is a new Vue Component stored in a partial cshtml file `ChangellyComponent` which holds the Vuejs logic for it and then a partial cshtml to define the VUejs template used in CheckoutBody as `ChangellyTemplate`

Since Changelly does not have fixed rates, there is a new controller `ChangellyController` that takes a `storeid`  which  queries the list of currencies available from Changelly and also a GET for retrieving the current rate of a specific currency pair.

The UI works like this:
* Loads Changelly Currencies from btcpay.com/changelly/{storeid}/currencies
* Go to Conversion Page
* Select Currency
* Sends request to btcpay.com/changelly/{storeid}/getExchangeAmount and specifies currency pairs and amount of 1
* Calculate rate and add Changelly's 0.5% fee to result
* Send another request to btcpay.com/changelly/{storeid}/getExchangeAmount and verify that the computed amount is enough to cover the transacttion
* Generate url and show pay button

NicolasDorier:
need rebase

Kukks:
rebased

Kukks:
Ok ok, I'm done @NicolasDorier  & @rockstardev  
I've optimized it, reduced as much logic in js as I could and added quite a bit of test coverage to it. :)

BTCPayServer/Controllers/ChangellyController.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using BTCPayServer.Models;
+using BTCPayServer.Payments.Changelly;
+using Changelly.ResponseModel;
+using Microsoft.AspNetCore.Mvc;
+
+namespace BTCPayServer.Controllers
+{
+    [Route("[controller]/{storeId}")]
```
NicolasDorier: No permission required ?

Kukks:
So, controller does not need permissions. It should be accessible to the
checkout UI publicly.

I styled this feature in the payment method pattern because it needs the
data needs to live in the same manner as the actual payment methods and be
able to be toggled(excluded) just like them along with storing the data in
store blob and it seemed like the more organized way to do it.

That said, I think the payment id field `cryptocode` should be renamed
because a payment method could potentially handle multiple currencies
anyway. From your example, stripe would handle all fiat currency codes and
not just use for example.

What do you think?

NicolasDorier:
No, an invoice in stripe is always done in one specific currency.
We can't refactor payment id there is too much impact behind it. Even the fact that CryptoCode is not a crypto but say USD might be problematic.
One payment is done in a specific PaymentId, can't be several, if you need "several" it is just different payments.

 Anyway this is not a matter for this PR. 
Do not tweak PaymentId for changelly, because even if the user pay ETH via changelly, from merchant point of view the payment is style in BTC on chain.

BTCPayServer/Controllers/ChangellyController.cs
Kukks: nope, it needs to be public as there's ajax calls from the checkout ui

Kukks:
Refactored PR to not be in the paymentmethod pattern. Changelly settings are still stored in the blob and verified that it all still works. :)

Kukks:
It is more of the same but they do not fix the rates which is why they do not allow you to set the to amount upfront. They also enable over a hundred altcoins and even fiat currencies so once this is merged in we have USD/ EUR support ;)

NicolasDorier:
Question: Because the rate can fluctuate between when we calculate the rate and you pay to changelly, what happen if the user do not pay enough? Also does changelly send the money immediately?

Kukks:
From what I read a transaction can take around 5-30mins and I think they do some sort of limit buy order where if it times out it will give you a refund. There could be   a slight difference however so it might not match the amount required entirely but that is where the payment tolerance setting should come in handy

NicolasDorier:
@astupidmoose anything we should do for this? Maybe making the user over pay just in case?

astupidmoose:
Perhaps the merchant can add another "exchange fee" on top of the changelly fee. I don't know much about changelly so I can't really say too much. Underpaid invoices would be a real pain though if it was because their rates were not accurate enough. If the merchant could add a buffer of their choosing, this might help.

Think of it as a shitcoin tax :P

NicolasDorier:
@Kukks I think a shitcoin tax field is mandatory. It is very painful for merchants to deal with underpaid payments. Maybe defaulting to 2%, while explaining in the store settings why this is here. (avoiding underpayment because of how changelly works)

Kukks:
Shitcoin tax incoming...

But I will not make it an enforced thing just because I have no control over what the user decides to do once on the changelly ui( they can change the values, etc).

NicolasDorier:
@astupidmoose ready to be the guinea pig for this? :p

Kukks:
I pushed some code but i have not tested the shitcoin tax just fyi @NicolasDorier  :)

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -251,6 +251,11 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, string paymen
                 ? storeBlob.ChangellySettings
                 : null;
 
+            var changellyAmountDue = changelly!= null? 
+                (accounting.Due.ToDecimal(MoneyUnit.BTC) * 
+                 (changelly.AmountMarkupPercentage> 0? 
+                     (changelly.AmountMarkupPercentage/100): 1)) : (decimal?) null;
```
NicolasDorier: Not good. It is `(1 + (changelly.AmountMarkupPercentage/100))`. You can remove `changelly.AmountMarkupPercentage> 0`.

BTCPayServer/Models/StoreViewModels/UpdateChangellySettingsViewModel.cs
```diff
@@ -23,5 +23,11 @@ public class UpdateChangellySettingsViewModel
         public bool Enabled { get; set; } = true;
         
         public string StatusMessage { get; set; }
+        
+        [Display(Name = "Percentage to multiply amount requested at Changelly to avoid underpaid situations due to Changelly not guaranteeing rates. ")]
```
NicolasDorier: Add a `[Range(0, 100)]`

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -251,10 +251,10 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, string paymen
                 ? storeBlob.ChangellySettings
                 : null;
 
+            
             var changellyAmountDue = changelly!= null? 
                 (accounting.Due.ToDecimal(MoneyUnit.BTC) * 
-                 (changelly.AmountMarkupPercentage> 0? 
-                     (changelly.AmountMarkupPercentage/100): 1)) : (decimal?) null;
+                     (1+(changelly.AmountMarkupPercentage/100))) : (decimal?) null;
```
NicolasDorier: I thought you added 0.5% of changelly fee as well ?

Kukks: It is included in their result according to docs

NicolasDorier:
I think you need `(1m+(changelly.AmountMarkupPercentage/100m))` so stuff do not get converted to int

BTCPayServer/Controllers/ChangellyController.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                return Ok(result);
+            }
+
+            return BadRequest(result);
+        }
+
+        [HttpGet]
+        [Route("calculate")]
+        public IActionResult CalculateAmount(string storeId, string fromCurrency, string toCurrency,
```
NicolasDorier: the shitcoin tax is not applied?

Kukks: It is applied in the InvoiceController.UI's model which is then passed on to the vue component

BTCPayServer/Views/Invoice/Checkout-Body.cshtml
```diff
@@ -271,17 +271,32 @@
...
-                            <script type="text/javascript">function open_widget(a, e) { e.preventDefault(); var link = a.href; var changellyWindow = window.open(link, 'Changelly', 'width=600,height=470,toolbar=0,menubar=0,location=0,status=1,scrollbars=1,resizable=0,left=0,top=0'); changellyWindow.focus(); return false; }</script>
-                        <a onclick="open_widget(this, event);" href="https://changelly.com/widget/v1?auth=email&from=DASH&to=BTC&address=&amount=1&merchant_id=&ref_id=">
-                            <img src="https://changelly.com/pay_button_pay_with.png" alt="Changelly" />
-                        </a>*@
+                <center>             
+                    <changelly inline-template
+                               :merchant-id="srvModel.changellyMerchantId"
+                               :store-id="srvModel.storeId"
+                               :to-currency="srvModel.paymentMethodId"
+                               :to-currency-due="srvModel.changellyAmountDue"
```
Kukks: @NicolasDorier  right here

BTCPayServer/wwwroot/checkout/js/changellyComponent.js
```diff
@@ -0,0 +1,115 @@
+﻿var ChangellyComponent =
```
NicolasDorier: This script does not seems to be included anywhere?

BTCPayServer/Views/Invoice/Checkout-Body.cshtml
NicolasDorier: But  not in the return from calculate?

BTCPayServer/wwwroot/checkout/js/changellyComponent.js
Kukks: Through the bundle 
```
    {
        "outputFileName": "wwwroot/bundles/main-bundle.min.js",
        "inputFiles": [
            "wwwroot/vendor/jquery/jquery.js",
            "wwwroot/vendor/popper/popper.js",
            "wwwroot/vendor/bootstrap4/js/bootstrap.js",
            "wwwroot/vendor/jquery-easing/jquery.easing.js",
            "wwwroot/vendor/scrollreveal/scrollreveal.min.js",
            "wwwroot/vendor/magnific-popup/jquery.magnific-popup.js",
            "wwwroot/vendor/bootstrap4-creativestart/*.js",
            "wwwroot/main/**/*.js"
        ]
    },
```

BTCPayServer/Views/Invoice/Checkout-Body.cshtml
Kukks: The Calculate action in the controller is called from inside the vue component using `changellyAmountDue` which is including the shitcoin tax.

NicolasDorier:
@Kukks I needed to revert this PR.

Several issues:

* Some warnings
* The Changelly library you are using is not using async to make outside calls
* Test failed with NullReferenceException or database exceptions

PR #268, opened 2018-09-05
Feature/openiddict

Kukks:
This adds OpenIddict as an STS server to BtcpayServer!  We can now issue jwt tokens to users of the system and start creating a sexy new api.

Some notes:
* I have added a policy scheme that will act  as the default challenge towards the site. This policy scheme dynamically chooses between the bitpay, cookie and jwt schemes( thank you asp.net core 2.1)
* Sessions via JWT are also sticky in the case the server restarts via a persisted file in the data dir called `rsaparams`
* Works seamlessly with the btcPay Claims so we can decorate the new apis with the existing permissive stuff (tested with `[Authorize(Policy = Policies.CanModifyStoreSettings.Key)]`
* ~~I had to hack and create an instance of the `BtcPayServerOptions` during the ConfigureServices as I need it for setup.~~
* ~~I had to again do another hax to get the options in the BTCPayServerServcies hosted service in order to configure the `AddJwtBearer`.~~ Over here, there is a bit of an issue as I need to know at least one viable accessible url to the btcpayserver host.. Currently:
```
options.Authority = btcPayServerOptions.ExternalUrl?.ToString()?? "http://127.0.0.1:14142/";
options.Audience = btcPayServerOptions.ExternalUrl?.ToString()?? "http://127.0.0.1:14142/";
```


Kukks:
@NicolasDorier  @rockstardev Updated and looking for a review :)

rockstardev:
Sorry for not responding man... like every day I start typing and then something else comes up.

I'll ping you on Slack for some tricky questions I have on this.

NicolasDorier:
Can I see an example on how to use that?

NicolasDorier:
Can you try to divide that in smaller PRs? it is quite hard to review.

Kukks:
@NicolasDorier I am not sure I can break it up; this is pretty much the minimal code I could get to adding it in.

Any controller/controller action decorated with an `Authorize` attribute will automatically allow you to use any registered authentication scheme now. For testing purposes, I added a Test Controller at the route:  `http://btcpayhost.com/api/v1.0/test` that you can hit:
`GET: http://btcpayhost.com/api/v1.0/test ` - validates that the user is authenticated
`GET: http://btcpayhost.com/api/v1.0/test/{storeId}`  - this one also validates that the current user has permissions to modiy the specified storeid

If you are logged in the site via cookies, you should already be able to navigate to these endpoints. If you use the bitpay authentication method, you should also be able to hit these endpoints. The other option is through Bearer Authentication which requires you to add a Request Header `Authorization` with value `Bearer XXXXX` where XXXXX is the access token you acquired through the STS

To get an access token you should hit: 
`POST http://btcpayhost.com/connect/token`
with this body:
image

This is just one of the possible ways to get an access token. Here's a list of the other methods: http://docs.identityserver.io/en/release/topics/grant_types.html

I've only tested password flow but the toehrs should be quite straight forward in any case.

NicolasDorier:
Can you add unit test? Also I don't know what OpenIddict is from high level. What does it allow us to do?

NicolasDorier:
I still don't understand. Do you have an example of a client which use OpenID to query BTCPay API? (maybe in the tests)

Kukks:
I will create a few usecase examples under a different repo while also trying to add some more UTs.

Kukks:
@NicolasDorier  I set up a new branch https://github.com/Kukks/btcpayserver/tree/feature/openiddict-examples with some auth clients. All of them are a mix of different authenetication methods via OpenId. I haven't gotten around to hotwiring them all to hook up to btcpayserver (they are from the openiddict samples) but I've tried to adapt so some might already work.

Included are:
* An Aurelia based  SPA using Implicit Flow( trustless environment, user is redirected to btcpay login and gets an access token to use on the SPA. Great to build a new custom UI for btcpay if needed or for 3rd party apps)
* An Angular based SPA using Refresh Flow( same as implicit flow, but gives user a refresh token to be able to renew their access token without logging in again. This allows for short-lives access tokens which lets us to control upon refresh if a token can continue to operate, think breached users)
* .net console app using Password Flow( you provide a user and pass to an http endpoint and get an access token back)
* asp.net core app using Authentication Code Flow( This works in a similar way as Implicit/Refresh but allows the backend to get the access token from btcpayserver. This way a site can use the standard cookie auth mechanisms on its own site whil still having access to btcpayserver's users and apis)
* asp.net core app using Oauth Introspection( This allows an external api to btcpay to let authorized btcpayusers use it as if it was part of the original api)

I don't think I have time to hotwire each of them to btcpay honestly but I just wanted to give you an example of what kind of ecosystem we can build around this.

Beyond that, we can also create clients for each user to be able to say that Client A can only access API A while only using Implicit Flow.

PR #274, opened 2018-09-12
enable account lockout 

### LINKED ISSUE #273: Limit front-end login attempts
It's a good idea to have a way to limit brute-force login attempts on the front-end of BTCPay. 

I'm sure many users or even admins won't set 2FA or if we implement U2F (as I suggested earlier in #53).

My suggestions : 
- Limit the number of logins and lock account after few attempts for a certain time or require email verification
- Add google anti-spam "are you a robot" verification

Kukks:
closes #273

### DIFF
--- a/BTCPayServer/Controllers/AccountController.cs
+++ b/BTCPayServer/Controllers/AccountController.cs
@@ -90,7 +90,7 @@ namespace BTCPayServer.Controllers
                 }
                 // This doesn't count login failures towards account lockout
                 // To enable password failures to trigger account lockout, set lockoutOnFailure: true
-                var result = await _signInManager.PasswordSignInAsync(model.Email, model.Password, model.RememberMe, lockoutOnFailure: false);
+                var result = await _signInManager.PasswordSignInAsync(model.Email, model.Password, model.RememberMe, lockoutOnFailure: true);
                 if (result.Succeeded)
                 {
                     _logger.LogInformation("User logged in.");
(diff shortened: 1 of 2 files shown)


rockstardev:
ACK

If no NACKs I'll merge this in next release

rockstardev:
One note on this though - @NicolasDorier already has added Rate Limits... here:

https://github.com/btcpayserver/btcpayserver/blob/957d478865f26ac68ed70c2ee422a237126d0d56/BTCPayServer/Controllers/AccountController.cs#L74

So if you try too many logins at once you'll get `error 429`.

rockstardev:
Also @Kukks what is our process for unlocking account?

I can see that we also have some work to do on `Lockout.cshtml` because when you lock the account you don't see any text because content is not wrapped in `<section><div class="container"><div class="row">`. You want to take care of it or I should do it?

Kukks:
This is for account login attempts regardless of ip. :) If the account is being attempted over multiple ips at the same time, this will lock it for 5 minutes. :)

Kukks:
Go for it, bedtime for me :D

Kukks:
It unlocks automatically by the configured time from what I really don't
the asp net core identity docs. It is set to five minutes. If we want to
force unlock we can do it through psql cli command in the postgres docker
which is not optimal but again, it's only five mins right now.

We can enhance later by allowing admins to unlock from the ui or by maybe
sending an unlock code to the user's email or something like that.

rockstardev:
OK, I'll merge and then deal with UI - agree that for now unlock is no biggie when it happens on it's own in 5 minutes. :*

PR #280, opened 2018-09-20
Fixes to close off few open issues

rockstardev:


BTCPayServer/BTCPayServer.csproj
```diff
@@ -59,9 +59,9 @@
   </ItemGroup>
 
 	<ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.0" />
-		<PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version=" 2.1.0" PrivateAssets="All" />
-		<PackageReference Include="YamlDotNet" Version="4.3.1" />
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.1" />
+	<PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version=" 2.1.1" PrivateAssets="All" />
```
Kukks: Ef Core tools are bug infested. Be careful when changing these 

DO we need to update the dockerfile to reflect the .net core version bumped up here?

rockstardev: No, 2.1.1 is minor release that bugfixes problems in 2.1.0... so there should be no problems for dockerfile.

This change is definitely needed... just try running BtcPayServer in Debug without this commit, than try changing `.cshtml` - you'll see `The type 'RazorViewAttribute' exists in both 2.1.0 and 2.1.1` error.

PR #310, opened 2018-10-03
creating italian localization file

dennybiasiolli:


BTCPayServer/wwwroot/checkout/js/langs/it.js
```diff
@@ -0,0 +1,54 @@
+const locales_en = {
```
britttttk: const locales_it = {
(because const locales_en is being used for English)

dennybiasiolli:
@britttttk done 😃

PR #318, opened 2018-10-05
Displaying Node Info as QR code for Lightning payments

rockstardev:


bitcoinshirt:
I like this. I don't think there is a better way to do this, and I'm happy with how you solved the problem. Haven't tested it, but I like the concept.

rockstardev:
Preview:

https://twitter.com/i/status/1048356906971095041

NicolasDorier:
cool!

Issue #237, opened 2018-08-03
[REQUEST]  LOG FILE


forkyboy:
CAn you add an option to specify where the log file is written?

rockstardev:
Done

PR #322, opened 2018-10-08
Issue #237 add debug log file command line option --debuglog.

sipsorcery:
Adds a --debugfile command line option to log console messages to a rolling debug file (limited to 2MB in size and rolls when full). If warranted other options such as the size, number of files to retain etc. could also be added.

Example usage:
` dotnet run --project=BTCPayServer/BTCPayServer.csproj --debuglog c:\temp\debug.log`

### DIFF
--- a/BTCPayServer.Tests/BTCPayServer.Tests.csproj
+++ b/BTCPayServer.Tests/BTCPayServer.Tests.csproj
@@ -10,6 +10,9 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
+    <PackageReference Include="Serilog" Version="2.7.1" />
+    <PackageReference Include="Serilog.AspNetCore" Version="2.1.1" />
+    <PackageReference Include="Serilog.Sinks.File" Version="4.0.0" />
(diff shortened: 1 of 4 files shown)


sipsorcery:
Hmm Serilog doesn't work on Ubuntu... Back to the drawing board.

NicolasDorier:
I would want to test this on linux, as I prefer Serilog than NLog #323 when you say "doesn't work" you mean that it crash at startup or do not write file?

sipsorcery:
With Serilog on Ubuntu there was no crash. The problem was that no log file was produced. My 5 minutes of googling revealed other people had the same problem and no obvious solution.

If Serilog is the preferred option for btcpayserver I can look into it further. I'll dig into it.

nblumhardt:
We're using Serilog extensively on Ubuntu, logging to files and elsewhere. It's almost certainly a permissions issue of some kind - if you need a hand working through it, a bunch of us hang out on https://gitter.com/serilog/serilog 👍

NicolasDorier:
@nblumhardt the thing is that @sipsorcery achieved the same thing with NLog. There might be a problem. I will also take a look at it today.

PR #323, opened 2018-10-08
Issue #237 add debug log file command line option --debuglog.

sipsorcery:
Adds a --debugfile command line option to log console messages to a rolling debug file (limited to 2MB in size and rolls when full). If warranted other options such as the size, number of files to retain etc. could also be added.

Example usage:
` dotnet run -c Debug -p BTCPayServer/BTCPayServer.csproj --testnet --debuglog ~/btcpayserver.debug`

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -53,6 +53,9 @@
     <PackageReference Include="NicolasDorier.RateLimits" Version="1.0.0.3" />
     <PackageReference Include="NicolasDorier.StandardConfiguration" Version="1.0.0.18" />
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="2.1.0" />
+    <PackageReference Include="Serilog" Version="2.7.1" />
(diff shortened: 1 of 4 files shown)


NicolasDorier:
How can it works? I don't see it plugged to the logging infrastructure of ASP.NET log?

NicolasDorier:
I am not super fan of introducing an external dependency. I will make a quick check to see if there is an alternative inside .NET Core log system.

sipsorcery:
Serilog was my first choice as well #322. Unfortunately it didn't seem to work on Ubuntu and therefore probably other Linux OS's as well.

I can "enhance" the existing custom logging class already in use if that's more palatable? It obviously won't be as flexible as a proper logging library but should be sufficient for a basic debug file.

NicolasDorier:
wtf (O_o) ok I will test this

sipsorcery:
@NicolasDorier I've switched this PR back to Serilog. By moving the Serilog initialization code into the ConfigureLogging method I do now get a log file on Ubuntu.

It's the same amount of code for NLog vs Serilog. I'm not sure how you could do the config with one line unless the idea is to use a config file for Serilog?

PR #324, opened 2018-10-09
allow showing addresses from derivationscheme even if not updating it

### LINKED ISSUE #279: Confirm the addresses (BTC) in Derivation Path can be bypassed
Strange bug, in **Confirm the addresses (BTC)** when setting up the derivation scheme.

The checkup only happens when you clear the public key field, disable derivation scheme and save.

I think it's important this checkup is always shown as soon as the derivation scheme is changed. Strangely I was able to bypass this checkup most of the times. 

screen shot 2018-09-19 at 22 57 25

Kukks:
closes #279
This adds the option to check the derivation scheme's generated addresses at will.

### DIFF
--- a/BTCPayServer/Controllers/StoresController.BTCLike.cs
+++ b/BTCPayServer/Controllers/StoresController.BTCLike.cs
@@ -58,7 +58,7 @@ namespace BTCPayServer.Controllers
 
         [HttpPost]
         [Route("{storeId}/derivations/{cryptoCode}")]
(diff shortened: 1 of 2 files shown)


NicolasDorier:
I think a new button is confusing, it should have the same behavior as before.

NicolasDorier:
Fixed on https://github.com/btcpayserver/btcpayserver/commit/dcc859a86a0ef621c059912af32d8ef720898beb#diff-96fe038a36fee5ba45b97e848ec1d70eR106

PR #326, opened 2018-10-10
Add simplified chinese translation

BobleChinois:
Pretty self-explanatory :smiley: 

Thanks to @britttttk and her coworker that did the translation, I only readproofed it (with a little help from a chinese friend) and prepared the PR.


BobleChinois:
pou queue d'chie

PR #327, opened 2018-10-11
Solving the new version of btcpayserver caused btcpay-python not to create an order problem

GameXG:
https://github.com/btcpayserver/btcpay-python/blob/master/btcpay/client.py#L88


NicolasDorier:
Will release a new version tomorrow I think

PR #343, opened 2018-10-18
Changelly v2

Kukks:
Let's try this again!

- [x] Tests successful
- [x] UI For update  store
- [x] UI to update changelly settings
- [x] UI for Checkout 
- [x] Able to toggle feature per store
- [x] Able to toggle fiat currencies( BtcPayServer now supports VISA and Mastercard... LOL :D )
- [x] Able to set a markup percentage to avoid underpaid invoices
- [x] No Dependency on crappy unofficial Changelly lib( we now use `async` and `HttpClient` calls)
- [x] Pretty currency list with icons


NicolasDorier:
Is it ready to merge ?

BTCPayServer/Payments/Changelly/ChangellyClientProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace BTCPayServer.Payments.Changelly
+{
+    public class ChangellyClientProvider
+    {
+        private readonly StoreRepository _storeRepository;
+
+        private readonly ConcurrentDictionary<string, Changelly> _clientCache =
+            new ConcurrentDictionary<string, Changelly>();
+
+        public ChangellyClientProvider(StoreRepository storeRepository)
```
NicolasDorier: Can you use the `IHttpClientFactory`  to inject the HttpClient into the Changelly client?

BTCPayServer/Payments/Changelly/Changelly.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+
+        public virtual async Task<(decimal Amount, bool Success, string Error)> GetExchangeAmount(string fromCurrency,
+            string toCurrency,
+            decimal amount)
+        {
+            try
+            {
+                var message =
```
NicolasDorier: Use string interpolation `$`

BTCPayServer/Payments/Changelly/Changelly.cs
```diff
@@ -0,0 +1,135 @@
...
+                    "{\"id\": \"test\",\"jsonrpc\": \"2.0\",\"method\": \"getExchangeAmount\",\"params\":{\"from\": \"" +
+                    fromCurrency + "\",\"to\": \"" + toCurrency + "\",\"amount\": \"" + amount + "\"}}";
+
+                var result = await PostToApi<string>(message);
+                return !result.Success ? (0, false, result.Error) : (Convert.ToDecimal(result.Result.Result), true, "");
```
NicolasDorier: Instead of returning error Go-style I think it is better to throw a proper `ChangellyException`

Kukks:
did all the changes @NicolasDorier

NicolasDorier:
Let's see how it goes, if it makes more problem than it solves, we can always remove it.

astupidmoose:
Documented this on Slack, but thought I would also put here to keep history:

Upon starting BTCPay I receive this error: 
Controllers/StoresController.Changelly.cs(92,21): warning CS0162: Unreachable code detected [
/home/stufftech/btcpayserver/BTCPayServer/BTCPayServer.csproj]
Payments/Changelly/Changelly.cs(36,17): warning CA1305: The behavior of 'StringBuilder.Append
Format(string, object)' could vary based on the current user's locale settings. Replace this
call in 'Changelly.ToHexString(byte[])' with a call to 'StringBuilder.AppendFormat(IFormatPro
vider, string, object)'. [/home/stufftech/btcpayserver/BTCPayServer/BTCPayServer.csproj]
Payments/Changelly/Changelly.cs(105,20): warning CA1305: The behavior of 'Convert.ToDecimal(s
tring)' could vary based on the current user's locale settings. Replace this call in 'Changel
ly.GetExchangeAmount(string, string, decimal)' with a call to 'Convert.ToDecimal(string, IFor
matProvider)'. [/home/stufftech/btcpayserver/BTCPayServer/BTCPayServer.csproj]

Changelly still seems to work. 

I also have noticed some issues with Changelly, more as a service that should be documented to users maybe.

- Changelly has what I feel is an unreasonable "Minimum" purchase. Approx $30-40 USD. 
- Changelly does not issue new API keys without approval. This could prove troublesome for some vendors. It also means some vendors will use the default API keys potentially putting those keys at risk if the merchant is considered a bad actor and bans the default keys. 

Will add more if I discover through testing.

Issue #344, opened 2018-10-19
Feature request: MySQL/MariaDB/Galera support


ketominer:
It would be great to have support for multiple database engines. The top priority for me would be MySQL and its derivatives. Should probably be a drop-in replacement for the driver as I didn't find any Postgres specific queries in the code. MariaDB can be a challenge starting from version 10.2 because the default SQL_MODES are very strict but it can be changed in the database configuration if it breaks something (not ideal).
Galera requires a little more work because of possible deadlocks in the cluster leading to temporary errors (this requires the code to handle retries) but we can survive without this for small scale deployments (or by directing writes to only one node of the database cluster).

sipsorcery:
The list of supported data stores for Microsoft.EntityFrameworkCore (the database abstraction layer used by btcpayserver) is [here](https://docs.microsoft.com/en-us/ef/core/providers/).

Adding a new database option will add at least one additional dependency to btcpay's rather long list which currently stands at 26 different external components. That being said MySQL is currently more widely deployed than Postgresql so would seem like a good option to have. 

What do you think @NicolasDorier? I'm happy to wire up a MySQL Entity Framework provider if it's something you'd like in the project?

NicolasDorier:
I replied with my concerns of it in the PR. Not strictly opposed to it, but often, this is not as easy as adding one provider.

PR #345, opened 2018-10-20
MySQL Support

sipsorcery:
PR for #344 to add MySQL support using the open source [Pomelo.EntityFrameworkCore.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) driver.

The existing Postgresql configuration options have been duplicated.

Example command line option: 
`--mysql "User ID=root;Password=password;Host=localhost;Port=3306;Database=btcpay;"`

Example environment variable: 
`BTCPAY_MYSQL : User ID=root;Password=password;Host=127.0.0.1;Port=3306;Database=btcpayserver;`

mysql_btcpay

### DIFF
--- a/BTCPayServer.Tests/ServerTester.cs
+++ b/BTCPayServer.Tests/ServerTester.cs
@@ -60,8 +60,9 @@ namespace BTCPayServer.Tests
             {
                 NBXplorerUri = ExplorerClient.Address,
                 LTCNBXplorerUri = LTCExplorerClient.Address,
+                TestDatabase = Enum.Parse<TestDatabases>(GetEnvironment("TESTS_DB", TestDatabases.Postgres.ToString()), true),
                 Postgres = GetEnvironment("TESTS_POSTGRES", "User ID=postgres;Host=127.0.0.1;Port=39372;Database=btcpayserver"),
-                MySQL = GetEnvironment("TESTS_MYSQL", null),
+                MySQL = GetEnvironment("TESTS_MYSQL", "User ID=root;Host=127.0.0.1;Port=33036;Database=btcpayserver"),
                 IntegratedLightning = MerchantCharge.Client.Uri
             };
             PayTester.Port = int.Parse(GetEnvironment("TESTS_PORT", Utils.FreeTcpPort().ToString(CultureInfo.InvariantCulture)), CultureInfo.InvariantCulture);
--- a/BTCPayServer.Tests/ServerTester.cs
+++ b/BTCPayServer.Tests/ServerTester.cs
@@ -61,6 +61,7 @@ namespace BTCPayServer.Tests
                 NBXplorerUri = ExplorerClient.Address,
                 LTCNBXplorerUri = LTCExplorerClient.Address,
                 Postgres = GetEnvironment("TESTS_POSTGRES", "User ID=postgres;Host=127.0.0.1;Port=39372;Database=btcpayserver"),
+                MySQL = GetEnvironment("TESTS_MYSQL", null),
                 IntegratedLightning = MerchantCharge.Client.Uri
             };
             PayTester.Port = int.Parse(GetEnvironment("TESTS_PORT", Utils.FreeTcpPort().ToString(CultureInfo.InvariantCulture)), CultureInfo.InvariantCulture);
--- a/BTCPayServer/Hosting/BTCPayServerServices.cs
+++ b/BTCPayServer/Hosting/BTCPayServerServices.cs
@@ -75,12 +75,12 @@ namespace BTCPayServer.Hosting
             {
                 var opts = o.GetRequiredService<BTCPayServerOptions>();
                 ApplicationDbContextFactory dbContext = null;
-                if (String.IsNullOrEmpty(opts.PostgresConnectionString) == false)
+                if (!String.IsNullOrEmpty(opts.PostgresConnectionString))
                 {
                     Logs.Configuration.LogInformation($"Postgres DB used ({opts.PostgresConnectionString})");
                     dbContext = new ApplicationDbContextFactory(DatabaseType.Postgres, opts.PostgresConnectionString);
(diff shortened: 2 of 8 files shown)


NicolasDorier:
Hey, I am not against this, however I think we should add the possibility to run unit tests on it by adding it to docker-compose, and having an environment variable.

I ran into some many gotcha with sqlite, things did not just ran correctly.

NicolasDorier:
Same thing: some gotcha on postgres, which by default, create primary keys on string field which does not result in a point query, but in a full scan.

NicolasDorier:
I was also considering dropping Entity Framework in the future in favor of Dapper. Because of the number of table, the first request to entity framework (the call to `.Migrate`) take like 15 seconds. That said, I don't think I can get ride on entity framework now :(

sipsorcery:
I've updated the PR to be able to run the unit tests against MySQL (the docker-compose is getting pretty crowded...). By default Postgres is still used for unit tests. To use MySQL an environment variable needs to be set:

`TESTS_MYSQL:User ID=root;Host=127.0.0.1;Port=33306;Database=btcpayserver`

I get the same results when running the unit tests with Postgres or MySQL, 33 pass and 6 fail (the failures are to do with lightning).

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -75,17 +75,23 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
             {
                 var opts = o.GetRequiredService<BTCPayServerOptions>();
                 ApplicationDbContextFactory dbContext = null;
-                if (opts.PostgresConnectionString == null)
+                if (String.IsNullOrEmpty(opts.PostgresConnectionString) == false)
```
NicolasDorier: `!String.IsNullOrEmpty(opts.PostgresConnectionString)`

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -75,17 +75,23 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
...
                 {
-                    var connStr = "Data Source=" + Path.Combine(opts.DataDir, "sqllite.db");
-                    Logs.Configuration.LogInformation($"SQLite DB used ({connStr})");
-                    dbContext = new ApplicationDbContextFactory(DatabaseType.Sqlite, connStr);
+                    Logs.Configuration.LogInformation($"Postgres DB used ({opts.PostgresConnectionString})");
+                    dbContext = new ApplicationDbContextFactory(DatabaseType.Postgres, opts.PostgresConnectionString);
+                }
+                else if(String.IsNullOrEmpty(opts.MySQLConnectionString) == false)
```
NicolasDorier: same

BTCPayServer.Tests/ServerTester.cs
```diff
@@ -61,6 +61,7 @@ public ServerTester(string scope)
                 NBXplorerUri = ExplorerClient.Address,
                 LTCNBXplorerUri = LTCExplorerClient.Address,
                 Postgres = GetEnvironment("TESTS_POSTGRES", "User ID=postgres;Host=127.0.0.1;Port=39372;Database=btcpayserver"),
+                MySQL = GetEnvironment("TESTS_MYSQL", null),
```
NicolasDorier: Add `User ID=root;Host=127.0.0.1;Port=3306;Database=btcpayserver`  by default.

I think we should add a new environment variable:

`TESTS_DB` by default to `postgres`, but could be `mysql`

NicolasDorier:
super cool

sipsorcery:
I've made the suggested changes.

The only thing that's not really clear is how the TESTS_DB environment variable should be set when running the tests from VS? 

The test framework doesn't use use the launchSettings.json file by default it's not possible to set the environment variable from the VS Project page. There are solutions such as copying the launchSettings.json file to test directory but that's a more intrusive change.

NicolasDorier:
@sipsorcery that's fine, just have to use `$env:TEST_DB` before running `dotnet test`. That is enough for me, later we will have proper CI.

ketominer:
Looks postgresql specific as per https://www.postgresql.org/docs/9.3/static/indexes-opclass.html : 

The operator classes text_pattern_ops, varchar_pattern_ops, and bpchar_pattern_ops support B-tree indexes on the types text, varchar, and char respectively. The difference from the default operator classes is that the values are compared strictly character by character rather than according to the locale-specific collation rules. This makes these operator classes suitable for use by queries involving pattern matching expressions (LIKE or POSIX regular expressions) when the database does not use the standard "C" locale. As an example, you might index a varchar column like this:

CREATE INDEX test_index ON test_table (col varchar_pattern_ops);

Note that you should also create an index with the default operator class if you want queries involving ordinary <, <=, >, or >= comparisons to use an index. Such queries cannot use the xxx_pattern_ops operator classes. (Ordinary equality comparisons can use these operator classes, however.) It is possible to create multiple indexes on the same column with different operator classes. If you do use the C locale, you do not need the xxx_pattern_ops operator classes, because an index with the default operator class is usable for pattern-matching queries in the C locale.

ketominer:
But there is another issue with stock parameters against mariadb 10.1 in my first test:

MySql.Data.MySqlClient.MySqlException (0x80004005): Index column size too large. The maximum column size is 767 bytes. ---> MySql.Data.MySqlClient.MySqlException (0x80004005): Index column size too large. The maximum column size is 767 bytes.

Also the engine is not set in the create table statements (should be innodb). I'll add relevant information in the doc for database creation and settings.

ketominer:
One question, any particular use for longblobs and longtexts? Those are datatypes to store up to 4GB per column, not sure StoreName needs to be that long... :)

NicolasDorier:
@ketominer no particular reason.

PR #360, opened 2018-10-24
Added configuration options for BtcPayServer https binding

sipsorcery:
While the nginx/letsencrypt deployment approach should suit most people there are still going to be some that would like to have BtcPayServer listen directly on an HTTPS endpoint with an existing certificate file.

This PR adds two new options to the configuration file:

```
httpscertificatefilepath=btcpaydev.pfx
httpscertificatefilepassword=
```

If the `httpscertificatefilepath` is set and the file exists then BtcPayServer will attempt to bind to a HTTPS endpoint.

As an example to bind to a HTTPS endpoint that listens on all interfaces the configuration file options could be set as:

```
port=433
bind=0.0.0.0
httpscertificatefilepath=btcpaydev.pfx
httpscertificatefilepassword=
```

For ipv6 loopback:

`bind=::1`

Having this option is also very helpful for developers integrating with BtcPayServer as it makes it simpler to run and debug through Visual Studio or the command line without requiring IIS.

### DIFF
--- a/BTCPayServer/Hosting/Startup.cs
+++ b/BTCPayServer/Hosting/Startup.cs
@@ -118,15 +118,33 @@ namespace BTCPayServer.Hosting
                     b.AllowAnyMethod().AllowAnyHeader().AllowAnyOrigin();
                 });
             });
+            services.Configure<KestrelServerOptions>(kestrel =>
+            {
+                var networkType = DefaultConfiguration.GetNetworkType(Configuration);
+                var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
+                var btcPaySettings = Configuration.Get<BTCPayServerOptions>();
+                var bind = Configuration.GetOrDefault<IPAddress>("bind", IPAddress.Loopback);
+                int port = Configuration.GetOrDefault<int>("port", defaultSettings.DefaultPort);
 
-            // Needed to debug U2F for ledger support
-            //services.Configure<KestrelServerOptions>(kestrel =>
-            //{
-            //    kestrel.Listen(IPAddress.Loopback, 5012, l =>
-            //    {
-            //        l.UseHttps("devtest.pfx", "toto");
-            //    });
-            //});
+                if (!String.IsNullOrEmpty(btcPaySettings.HttpsCertificateFilePath))
+                {
+                    if (!File.Exists(btcPaySettings.HttpsCertificateFilePath))
+                    {
+                        // Note that by design this is a fatal error condition that will cause the process to exit.
+                        throw new ApplicationException($"The https certificate file could not be found at {btcPaySettings.HttpsCertificateFilePath}.");
+                    }
+
+                    Logs.Configuration.LogInformation($"Https certificate file path {btcPaySettings.HttpsCertificateFilePath}.");
+                    kestrel.Listen(bind, port, l =>
+                    {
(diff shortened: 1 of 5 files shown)


BTCPayServer/Hosting/Startup.cs
```diff
@@ -118,15 +118,27 @@ public void ConfigureServices(IServiceCollection services)
...
 
-            // Needed to debug U2F for ledger support
-            //services.Configure<KestrelServerOptions>(kestrel =>
-            //{
-            //    kestrel.Listen(IPAddress.Loopback, 5012, l =>
-            //    {
-            //        l.UseHttps("devtest.pfx", "toto");
-            //    });
-            //});
+                if (!String.IsNullOrEmpty(btcPaySettings.HttpsCertificateFilePath) && File.Exists(btcPaySettings.HttpsCertificateFilePath))
```
NicolasDorier: Can you remove the check on `File.Exists`? If the file do not exist, the user need some proper feedback.

sipsorcery:
@NicolasDorier certificate file existence check removed. Process will now exit if the certificate file does not exist.

NicolasDorier:
@sipsorcery the error message is understandable?

NicolasDorier:
Can you rebase as well?

sipsorcery:
> @sipsorcery the error message is understandable?

It was the standard file not found exception. I've added a custom application exception to make it obvious.

Rebase done.

NicolasDorier:
meh, will adjust

NicolasDorier:
Actually will keep it like that. I have the impression you did it the right way, and the way I create `BTCPayServerOptions` is wrong.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -118,15 +118,33 @@ public void ConfigureServices(IServiceCollection services)
                     b.AllowAnyMethod().AllowAnyHeader().AllowAnyOrigin();
                 });
             });
+            services.Configure<KestrelServerOptions>(kestrel =>
+            {
+                var networkType = DefaultConfiguration.GetNetworkType(Configuration);
+                var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
+                var btcPaySettings = Configuration.Get<BTCPayServerOptions>();
+                var bind = Configuration.GetOrDefault<IPAddress>("bind", IPAddress.Loopback);
```
NicolasDorier: actually I am not sure about this

NicolasDorier:
sorry I needed to revert.
Handling of `port` and `bind` is quite more difficult.

Look at https://github.com/NicolasDorier/NicolasDorier.Tools/blob/master/StandardConfiguration/DefaultConfiguration.cs#L89

Basically by doing what you do we break the fact the bind can have multiple bind address, and the fact `ASPNETCORE_URLS` is now ignored. (a property set by IDE when you set the URL in the project settings)

NicolasDorier:
I need to think about it a bit more.

BTCPayServer/Hosting/Startup.cs
sipsorcery: There seems to be at least 3 or 4 different ways to set the address/URL that Kestrel will use. But as best as I could tell the only two options for setting the HTTPS certificate were using code, as per this PR, or using the appSettings file. Using appSettings is an option but that would mean two config files for btcpayserver which would be very confusing.

PR #369, opened 2018-10-28
async i18n and json translation format

### LINKED ISSUE #364: Introduce Transifex support
Thanks to @britttttk 's insane amount of translation work, we need the locales system to scale a bit better. Someone mentioned Transifex( btc uses it) and we checked it out. It is free to OSS projects so that's good.  

Right now we load locale files from a .js and the only way that Transifex can load them is through Requirejs so I propose we change our locale loading to be through key-value json files. We can fetch them on demand in the ui when a suer changes language. :)

I've applied for an OSS license on Transifex and added in the first language file(english) to it at
https://www.transifex.com/btcpayserver-1/public/

Kukks:
Closes #364 
This Pr allows us to 

* use the  same locale files in Transifex for better scaling of languages. 
* load new language file on demand instead of all in the bundle to reduce bloat in initial load
* remove c# dependency when adding a new language
* default language set to en instead of en-us ( america ain't the boss of us :P )


Kukks:
@britttttk  I am guessing you have the "good" Transifex btcpayserver project name ( mine ended up being called `btcpayserver-1` because it existed). These new locale files should work unlike the current ones which gave you errors. 

@rockstardev  I'd like you review the FE stuff and see if I did anything bad, nothing major was changed though.

britttttk:
@Kukks Wow thank you. I moved it to here: [https://www.transifex.com/btcpayserver/btcpayserver/](https://www.transifex.com/btcpayserver/btcpayserver/)

NicolasDorier:
So we use https://www.transifex.com/btcpayserver/btcpayserver/ ?

NicolasDorier:
Let's use @britttttk https://www.transifex.com/btcpayserver/btcpayserver/

NicolasDorier:
tests do not pass

britttttk:
@NicolasDorier yes the en.json file that @Kukks created is uploaded to Transifex and then users can translate it by string. Transifex admin can download the new (ex: fr-FR.json) file that is created when someone translates some strings. If you want to see everything, message me your email for some reason can't find you or Kukks by username on Transifex to add you.

NicolasDorier:
@britttttk I normally did a request to be translator of french and english.

BTCPayServer/Services/LanguageService.cs
```diff
@@ -28,8 +29,13 @@ public class LanguageService
...
         public LanguageService(IHostingEnvironment environment)
         {
-            
-            var path = Path.Combine(environment.ContentRootPath, "wwwroot", "locales");
+            var path = (environment as HostingEnvironment)?.WebRootPath;
+            if (string.IsNullOrEmpty(path))
+            {
+                //test environment
+                path = Path.Combine(TryGetSolutionDirectoryInfo().FullName,"BTCPayServer", "wwwroot");
+            }
```
NicolasDorier: Why is it not set correctly on test environment?

BTCPayServer/Services/LanguageService.cs
```diff
@@ -44,6 +50,16 @@ public LanguageService(IHostingEnvironment environment)
             _languages = result.ToArray();
         }
 
+        private static DirectoryInfo TryGetSolutionDirectoryInfo(string currentPath = null)
```
NicolasDorier: this is ugly

BTCPayServer/Services/LanguageService.cs
Kukks: WebRootPath is not initialized on tests because of the  `IHostingEnvironment` pointing to the Tests project running instead of the main app.

BTCPayServer/Services/LanguageService.cs
Kukks: Yeah it is, I am not happy with test specific code in main app logic but it hard to find all usages of it in the tests without  too much refactoring

NicolasDorier: just compile the json file inside the assembly like I did for Currencies.txt

Kukks: I cannot do this as the json files will be generated dynamically later from transifex.  I also don't want translators to need to compile the source in order for new languages to be added until we add in automatic transifex generation

Kukks:
Apart from the small UT hack, everything in this PR is ok last I checked @NicolasDorier

NicolasDorier:
There is a conflict (probably introduced by https://github.com/btcpayserver/btcpayserver/pull/381/files) can you fix? then I merge.

PR #371, opened 2018-10-28
Add in derivation dropdown option

### LINKED ISSUE #361: Derivation Scheme - Dropdown format selection
@Kukks recently suggested that we have some sort of a drop-down menu, as we noticed that users sometimes just have no idea how to add an appropriate suffix (sometimes they forget the dash).

This is just to document his idea so that we don't forget it. 

1. Add drop down for formats that are added into the xpub field when you select them.
2. Have some sort of a tooltip or a hint what each address looks like 
3. Test button which leads users to the next page.

The problem is only ledger integration, I'm not sure if it can be easily adapted with this setup

new mockup 1

Kukks:
closes #361
What do you think so far @bitcoinshirt 

derivation dropdown


Kukks:
The toggle is just a wya to use the old UI just in caee. Its completely optional. I just wanted to show it to see how the options from the new UI and old UI stay updated in between themselves

Kukks:
The table is hidden because it shows options for the old UI way of editing.

bitcoinshirt:
Gotcha. My suggestion is hiding what's not necessary/may lead to confusion for the user.

Maybe "toggle legacy input method"? I somehow think users may click to that and be even more confused.  Or somehow make it less visible, it looks like a CTA instead of the help link. 

Any idea how we can keep the old input method (as a backup), but make it pop-up less?

Did you test it with Ledger by the way?

Kukks:
Maybe I can put the link lower in the page?

As for ledger, it shouldn't affect any of that but I'll give it a test tomorrow

bitcoinshirt:
I mean anything that can make it look less like a call to action. 
I'm now thinking if it would be helpful that I create a short video we can embed showing people how to add derivation schemes? Or use this one until then https://www.youtube.com/watch?v=xX6LyQej0NQ?

Derivation schemes are an important step, and I'd like if we can make sure to make it as smooth as butter.

Kukks:
I think a small table explaining each address type would be good too

NicolasDorier:
I am not even convinced that there is something to improve. I don't think it is a good idea to mention any P2*** at all. This is a difficult concept that the user should not have to care about.

The current UX with the step2 check and the user able to input the first address of his wallet work fine I think.

Kukks:
@NicolasDorier  We could rename them to more user friendly stuff such as 
"Legacy", "Backward Compatible Segwit", "Segwit", etc

bitcoinshirt:
@Kukks I don't think that would help. People don't know the difference anyway, they'll have to Google in both cases :) Most wallets have formats labeled as  P2***, instead of saying backward segwit, segwit etc. At least most of the ones I used.

NicolasDorier:
@Kukks they should not even care about the suffix. If they click `Next` then they can input the first address of their wallet so btcpay can figure out what they mean.

Kukks:
Hmm, guess I have been using the wrong workflow lol. We should better caption that address input field. I thought it was some troubleshooting tool

Kukks:
@bitcoinshirt  So what should I do here, close this or you have any better UX suggestions to get this aligned?  I think I would need to allow a user to either use the address type dropdown or to enter a test address

Kukks:
I will give it another go later tonight and see if it makes any sense

Kukks:
What if I put the dropdown next to the xpub field? And then move the test address input to the same page?

NicolasDorier:
@Kukks frankly, I would prefer not touching it. There has been no support issue on it at this moment.

The only thing that can be improved is the error message when somebody input a segwit derivation scheme on a non segwit crypto currency.

jmaurice:
While setting up BTC and LTC address derivation, I was confused. I would like to request this PR be re-opened, and I think some example language like the below could help. The pull-down for choosing address derivation scheme is also a good idea IMO.

```
P2PKH which begin with the number 1, eg: 1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2.
P2SH type starting with the number 3, eg: 3J98t1WpEZ73CNmQviecrnyiWrnqRhWNLy.
Bech32 type starting with bc1, eg: bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq.
```

britttttk:
> I think a small table explaining each address type would be good too

Maybe we could add a link to the docs using one of the (?) icons in the derivation area. We could expand on the address descriptions, and describe the table @Kukks mentioned, etc. If you don't want to over-complicate the page. 

@pavlenex I think we should also add information to the [Getting Started doc?](https://github.com/btcpayserver/btcpayserver-doc/blob/master/GettingStarted.md) suggesting that if you don't add a derivation you will not be able to create invoices or that the functionality of BTCPay will be limited.. something like that. Even if it's just a reminder for when people are trying to troubleshoot.

I originally agreed with this issue being too complicated UI. However I think it's inevitable that we should allow/help users understand and choose derivations. Even for me honestly, when it auto chooses I'm constantly confused and would prefer to know more about what is going on. :)

pavlenex:
Imo the whole process of adding store + derivation scheme should be simplified.

All of these suggestions are good, but won't prevent user from freaking out. 
For this process, I wouldn't educate them, I'd just guide them and handhold them step by step with a setup wizard of some sort.

1. Add Store Name and site
2. Paste your derivation scheme
3. Verify your derivation scheme
4. Integrate your store, create payment button, etc.

Ideal scenario would be: 
1. User plugs hardware wallet, we set up everything, he just needs store name and bam.

I'll open a separate issue for this.

Eskyee:
i like the idea for the address derivation UX, if this could be done correctly, I think it would help many people & newbies, to understand the address derivation, which for most is quite complex to understand, education is the way forward, looks like another great brain storm session for you guys to solve

NicolasDorier:
```
P2PKH which begin with the number 1, eg: 1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2.
P2SH type starting with the number 3, eg: 3J98t1WpEZ73CNmQviecrnyiWrnqRhWNLy.
Bech32 type starting with bc1, eg: bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq.
```
Is not user friendly.

I wished we don't touch this screen, people rarely screw up, let's not fix what is not broken.

britttttk:
If we don't want to touch this screen, anyone who would like to suggest adding info to the docs related to this step, feel free to add you comments here: [Improve Getting Started Guide](https://github.com/btcpayserver/btcpayserver-doc/issues/161)

NicolasDorier:
I added https://github.com/btcpayserver/btcpayserver/commit/575b6ca2226bcc8966440b8b47f66e0ac0f50fd8

PR #372, opened 2018-10-29
Adjusted mechanism for setting https binding configuration option

sipsorcery:
I've done some further investigation on the Ketstrel and BtcPayServer socket binding behaviour and documented it below (main [Kestrel reference](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-2.1#endpoint-configuration) page). 

Current Behaviour:
1. If `--bind` and/or `--port` are set on the command line they take precedence and Kestrel will attempt to bind on the corresponding end point (no way to set the HTTPS certificate options),

2. If `bind` and/or `port` are set in the <network>.settings file then they will be set as the `urls` environment variable and cause Kestrel to attempt to bind to that IP end point. If the command line option is also set the `port` on the command line gets used. If `bind` on the command line is also set then both are used resulting in multiple end points,

3. If neither of the previous 2 options are set then if present the `ASPNETCORE_URLS` environment variable will be used. This can set multiple end points (no way to set HTTPS certification options),

4. If none of the previous 3 options are set then if present the `applicationUrl` option from the `launchSettigns.json` file will be used. This can set multiple end points (no way to set HTTPS certification options),

5. If none of the above are set the default Kestrel endpoints of `https://localhost:5001` and `http://localhost:5000` will be attempted.

Options that don't work with BtcPayServer:

- The `--urls` command line argument,
- The `Kestrel` options from an appsettings.json (or custom config file), note this option does allow the setting of a HTTPS certificate file).

Given the above findings I've changed the mechanism in the original #360 PR so that the `services.Configure<KestrelServerOptions>` method is only called for the first two cases. That allows the default behaviour for cases 3 to 5 when the bind,port and https certificate options are not set on the command line or `<network>.settings` file.

An alternative could be to add `httpscertificatefilepath` and `httpscertificatefilepassword` as command line arguments but that seems clunky and less secure.

### DIFF
--- a/BTCPayServer/Hosting/Startup.cs
+++ b/BTCPayServer/Hosting/Startup.cs
@@ -119,29 +119,30 @@ namespace BTCPayServer.Hosting
                 });
             });
 
-            // Require that all the required options for a HTTPS certificate be set in the config file.
-            // If the HTTPS options can't be used don't use this mechanism and instead allow the defautl Kestrel bindings to occur.
+            // If the HTTPS certificate path is not set this logic will NOT be used and the default Kestrel binding logic will be.
             var networkType = DefaultConfiguration.GetNetworkType(Configuration);
             var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
             var btcPaySettings = Configuration.Get<BTCPayServerOptions>();
-            var bindAddress = Configuration.GetOrDefault<IPAddress>("bind", IPAddress.None);
-            int bindPort = Configuration.GetOrDefault<int>("port", 0);
+
             string httpsCertificateFilePath = btcPaySettings.HttpsCertificateFilePath;
 
-            if (bindAddress != IPAddress.None && bindPort != 0 && !String.IsNullOrEmpty(httpsCertificateFilePath))
+            if (!String.IsNullOrEmpty(httpsCertificateFilePath))
             {
+                var bindAddress = Configuration.GetOrDefault<IPAddress>("bind", IPAddress.Any);
+                int bindPort = Configuration.GetOrDefault<int>("port", 443);
+
                 services.Configure<KestrelServerOptions>(kestrel =>
                 {
-                    if (!File.Exists(btcPaySettings.HttpsCertificateFilePath))
+                    if (!File.Exists(httpsCertificateFilePath))
                     {
                         // Note that by design this is a fatal error condition that will cause the process to exit.
-                        throw new ConfigException($"The https certificate file could not be found at {btcPaySettings.HttpsCertificateFilePath}.");
+                        throw new ConfigException($"The https certificate file could not be found at {httpsCertificateFilePath}.");
                     }
 
-                    Logs.Configuration.LogInformation($"Https certificate file path {btcPaySettings.HttpsCertificateFilePath}.");
+                    Logs.Configuration.LogInformation($"Https certificate file path {httpsCertificateFilePath}.");
                     kestrel.Listen(bindAddress, bindPort, l =>
                     {
-                        l.UseHttps(btcPaySettings.HttpsCertificateFilePath, btcPaySettings.HttpsCertificateFilePassword);
+                        l.UseHttps(httpsCertificateFilePath, btcPaySettings.HttpsCertificateFilePassword);
                     });
                 });
             }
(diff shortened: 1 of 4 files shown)


NicolasDorier:
by `--urls` they mean `ASPNETCORE_URLS` environment variable which should work. (though if it does not work via the commandline `--urls` it seems a bug)

BTCPayServer/Hosting/Startup.cs
```diff
@@ -119,14 +119,32 @@ public void ConfigureServices(IServiceCollection services)
...
-            //    {
-            //        l.UseHttps("devtest.pfx", "toto");
-            //    });
-            //});
+            // Require that all the required options for a HTTPS certificate be set in the config file.
+            // If the HTTPS options can't be used don't use this mechanism and instead allow the defautl Kestrel bindings to occur.
+            var networkType = DefaultConfiguration.GetNetworkType(Configuration);
+            var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
+            var btcPaySettings = Configuration.Get<BTCPayServerOptions>();
+            var bindAddress = Configuration.GetOrDefault<IPAddress>("bind", IPAddress.None);
```
NicolasDorier: `IPAddress.None` is 0.0.0.0?

BTCPayServer/Hosting/Startup.cs
```diff
@@ -119,14 +119,32 @@ public void ConfigureServices(IServiceCollection services)
...
+            int bindPort = Configuration.GetOrDefault<int>("port", 0);
```
NicolasDorier: the default should be `443`

BTCPayServer/Hosting/Startup.cs
NicolasDorier: shouldn't it be IPAddress.Any?

BTCPayServer/Properties/launchSettings.json
```diff
@@ -15,8 +15,7 @@
             "ASPNETCORE_ENVIRONMENT": "Development",
             "BTCPAY_CHAINS": "btc,ltc",
             "BTCPAY_POSTGRES": "User ID=postgres;Host=127.0.0.1;Port=39372;Database=btcpayserver"
-        },
-      "applicationUrl": "http://127.0.0.1:14142/"
```
NicolasDorier: keep that, it is used during development

NicolasDorier:
small nits, but it seems good.

BTCPayServer/Hosting/Startup.cs
sipsorcery: The reason to use IPAddress.None is that if any of the four required settings are missing then the programmatic set up of the HTTPS binding should be ignored and the fallback option of using the other Kestrel binding mechanisms used.

BTCPayServer/Hosting/Startup.cs
sipsorcery: Reason for no default port is the same as the reason for IPAddress.None.

BTCPayServer/Hosting/Startup.cs
NicolasDorier: I would say if `httpsCertificateFilePath` is present, then it must be https, and thus you don't need to check IPAddress.None and port equals to 0.

sipsorcery: Ok, will adjust.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -119,14 +119,33 @@ public void ConfigureServices(IServiceCollection services)
...
-            //{
-            //    kestrel.Listen(IPAddress.Loopback, 5012, l =>
-            //    {
-            //        l.UseHttps("devtest.pfx", "toto");
-            //    });
-            //});
+            // If the HTTPS certificate path is not set this logic will NOT be used and the default Kestrel binding logic will be.
+            var networkType = DefaultConfiguration.GetNetworkType(Configuration);
+            var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
+            var btcPaySettings = Configuration.Get<BTCPayServerOptions>();
```
NicolasDorier: don't use this, just use `Configuration.GetOrDefault(HttpsCertificateFilePassword)` I think it is simpler.

PR #373, opened 2018-10-30
Logs UI in Server Admin

### LINKED ISSUE #370: Display logs on the front end
It would be helpful if users can access the logs from the front end easily. 
That can help them when troubleshooting or asking for help without doing too much back-end stuff.
Of course, if their server is down, they'll have to do it anyways.

Kukks:
Closes #370 

This PR allows users to view the server logs if `debuglog` is specified. 
It will show you a list of the debug log files (typically one per day) and then ou can can view each one individually in the UI.


Kukks:
Crap I have some other branch included in this one, one sec

PR #374, opened 2018-10-30
Logs UI in Server Admin

### LINKED ISSUE #370: Display logs on the front end

Kukks:
Closes #370

This PR allows users to view the server logs if debuglog is specified.
It will show you a list of the debug log files (typically one per day) and then ou can can view each one individually in the UI.

Update:
I've also added in an extra option: `debugloglevel`.  With this we can specify the minimum level of logging. I plan on exposing these two through the btcpayserver-docker


sipsorcery:
I've given this PR a spin and it works well and is very handy.

A concern I have is that the controller is using a fairly liberal regex to determine which files it will display on the web page. If a user inadvertently named their log file "mysupersecret.log" and there happened to be a sensitive file sitting in the system with the same starting name it would be exposed. I admit it's a bit of an edge case but when opening files best to be careful.

Two alternative approaches that I can think of:

1. Always use a pre-defined directory for  log files,
2. Don't allow the user to set the name of the log file and use a pre-defined prefix.

I think option 1 is the better one from a security point of view.

Kukks:
@sipsorcery garbage in garbage out?

Kukks:
Have to also keep in mind restricting it now would probably count as a breaking change?

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -625,5 +627,48 @@ public async Task<IActionResult> Emails(EmailsViewModel model, string command)
...
+            {
+                var di = Directory.GetParent(_Options.LogFile);
+                if (di == null)
+                {
+                    vm.StatusMessage = "Error: Could not load log files";
+                }
+
+                var x = Path.GetFileNameWithoutExtension(_Options.LogFile);
+                
+               var logFiles = di.GetFiles($"{x}*");
```
NicolasDorier: why not `di.GetFiles($"{x}*.{extension}");` ?

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -625,5 +627,48 @@ public async Task<IActionResult> Emails(EmailsViewModel model, string command)
...
         }
+        
+        [Route("server/logs/{file?}")]
+        public async Task<IActionResult> LogsView(string file = null)
+        {
+            var vm = new LogsViewModel();
+
+            if (string.IsNullOrEmpty(_Options.LogFile))
+            {
+                vm.StatusMessage = "Error: File Logging Option not specified.";
```
NicolasDorier: can you document the command line option / environment variable they must set?

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -625,5 +627,60 @@ public async Task<IActionResult> Emails(EmailsViewModel model, string command)
...
+            {
+                offset = 0;
+            }
+
+            var vm = new LogsViewModel();
+
+            if (string.IsNullOrEmpty(_Options.LogFile))
+            {
+                vm.StatusMessage = "Error: File Logging Option not specified. " +
+                                   "You need to set <i>debuglog</i> and optionally " +
```
NicolasDorier: you can't use markups here, it will be escaped.

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -625,5 +627,60 @@ public async Task<IActionResult> Emails(EmailsViewModel model, string command)
...
+                    .Skip(offset)
+                    .Take(5)
+                    .ToList();
+                vm.LogFileOffset = offset;
+
+                if (string.IsNullOrEmpty(file)) return View("Logs", vm);
+                vm.Log = "";
+                var path = Path.Combine(di.FullName, file);
+
+                using (var fileStream = new FileStream(
```
NicolasDorier: Replace with `vm.Log = File.ReadAllText(path);`

NicolasDorier:
also can you hijack one of the tests to use those routes so it is testable?

BTCPayServer/Controllers/ServerController.cs
Kukks: @NicolasDorier  I was doing that initially but the Serilog file writer locks the process and I need the filestream to only open a readable stream and make it work

Kukks:
I will add in some tests soon

Kukks:
Ok added in a basic UT for now

Kukks:
I think this PR is ready @NicolasDorier

NicolasDorier:
cool

PR #381, opened 2018-11-05
Modal invoice through btcpay.js

rockstardev:


NicolasDorier:
coooooool.

Though, how do I use it?

BTCPayServer/wwwroot/modal/btcpay.js
```diff
@@ -0,0 +1,113 @@
...
+    function showInvoice(invoiceId, params) {
+        window.document.body.appendChild(iframe);
+        var invoiceUrl = origin + '/invoice?id=' + invoiceId + '&v=2&view=modal';
+        if (params && params.animateEntrance === false) {
+            invoiceUrl += '&animateEntrance=false';
+        }
+        iframe.src = invoiceUrl;
+    }
+
+    function setApiUrlPrefix(urlPrefix) {
```
NicolasDorier: can you verify that the user does not set a `/` at the end by mistake? this is classical foot shooting.

rockstardev:
@NicolasDorier I have left `test.html` in same folder as example. Since we are _decentralized_ we won't have `btcpay.js` hosted at one place so setup needs to both reference `.js` file and then initialize site root for calls to show invoices, see: https://github.com/btcpayserver/btcpayserver/blob/364bffcda310b3ba214d263c6e138a9f6cae84db/BTCPayServer/wwwroot/modal/test.html#L5

I want to combine this with pay button so that users can create invoice when clicking button and then show that invoice... started conversation in Slack on stuff I need to make that happen.

NicolasDorier:
@rockstardev can we expose this test page inside BTCPay somewhere? Because how does the user can discover this feature?

NicolasDorier:
or better, just documenting somewhere

BTCPayServer/wwwroot/modal/test.html
```diff
@@ -0,0 +1,11 @@
+﻿<html>
+<head>
+    <script src="btcpay.js"></script>
+    <script type="text/javascript">
+        btcpay.setApiUrlPrefix("http://127.0.0.1:14142");
```
NicolasDorier: because NBXPlorer has the serverUrl, we don't have to use 127.0.0.1.

NicolasDorier:
Merging it, but @rockstardev can you delete the test.html, and use this popup in the Invoices list.

I am thinking this, in the invoice list, change the `Checkout` linke to  `Checkout [^]`. And if the user click on `^` we should the modal.

PR #388, opened 2018-11-09
Make language loading more efficient and solid

Kukks:


Kukks:
```
Now language loading has these priorities:
1. `lang` query string value
2. browser language (through `navigator.language` if available)
3. store default lang
4. fallback to `en`

And they all check to avoid uneccessary 404 loading
```

BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -109,32 +110,60 @@
...
-            .init({
-                backend: {
-                    loadPath: '/locales/{{lng}}.json',
-                },
-                lng: storeDefaultLang,
-                fallbackLng: 'en',
-                nsSeparator: false,
-                keySeparator: false
-            });
+<script type="text/javascript">
```
NicolasDorier: Please use `var srvModel = @Html.Raw(Json.Serialize(obj));` this code allow javascript injection.

Kukks: done

BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -109,32 +110,59 @@
...
     }
-
-    if (urlParams.lang) {
-        changeLanguage(urlParams.lang);
-    } 
-//    else if (storeDefaultLang) {
-//        changeLanguage(storeDefaultLang);
-//    }
-
+
```
NicolasDorier: replace with `return availableLanguages.indexOf(languageCode) >= 0`

BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -80,14 +82,13 @@
                             {{$t("nested.lang")}} >>
                         *@
                     
-                    <select asp-for="DefaultLang"  
-                            class="cmblang reverse invisible"
+                        <select asp-for="DefaultLang"  
+                                class="cmblang reverse invisible"
                             onchange="changeLanguage($(this).val())"
-                            asp-items="@langService.GetLanguages().Select((language) => new SelectListItem(language.DisplayName,language.Code, language.Code == Model.DefaultLang))"></select>
+                            asp-items="@langService.GetLanguages().Select((language) => new SelectListItem(language.DisplayName,language.Code, false))"></select>
```
NicolasDorier: Can you just add `PaymentModel.DefaultLanguage` and fill that in the in action ?

BTCPayServer/Views/Invoice/Checkout.cshtml
Kukks: done

BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -109,32 +110,56 @@
...
+            fallbackLng: fallbackLanguage,
+            nsSeparator: false,
+            keySeparator: false
+        });
+    
+    function computeStartingLanguage() {
+        if (urlParams.lang && isLanguageAvailable(urlParams.lang)) {
+            return urlParams.lang;
+        }
+//        else if (navigator &&
```
NicolasDorier: remove coments

BTCPayServer/Views/Invoice/Checkout.cshtml
Kukks: It was like that before but we are settings it in js before because we need to  validate and choose language based on priority and not just from that field

BTCPayServer/Views/Invoice/Checkout.cshtml
Kukks: done

PR #391, opened 2018-11-10
Improve POS UI

mariodian:
Hi, I was thinking the POS front-end could get a little facelift.

I put the items in the bootstrap's card and added option to add item image and description (both optional). Please have a look at the screenshots below.

I also fixed margins so that the title and bottom input isn't so close to the edge of the window.

default

description

image

image description

### DIFF
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -20,7 +20,9 @@ namespace BTCPayServer.Controllers
                 Template =
                     "tea:\n" +
                     "  price: 0.02\n" +
-                    "  title: Green Tea # title is optional, defaults to the keys\n\n" +
+                    "  title: Green Tea # title is optional, defaults to the keys\n" +
+                    "  description: Lovely, fresh and tender, Meng Ding Gan Lu is grown in the lush Meng Ding Mountains of the southwestern province of Sichuan # description is optional, defalts to none\n" +
+                    "  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg # image is optional, defaults to none\n\n" +
                     "coffee:\n" +
                     "  price: 1\n\n" +
                     "bamba:\n" +
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -152,7 +152,17 @@ namespace BTCPayServer.Controllers
                 .Where(kv => kv.Value != null)
                 .Select(c => new ViewPointOfSaleViewModel.Item()
                 {
+                    Description = c.Value.Children
+                             .Select(kv => new { Key = (kv.Key as YamlScalarNode)?.Value, Value = kv.Value as YamlScalarNode })
+                             .Where(kv => kv.Value != null)
+                             .Where(cc => cc.Key == "description")
+                             .FirstOrDefault()?.Value?.Value,
                     Id = c.Key,
+                    Image = c.Value.Children
+                             .Select(kv => new { Key = (kv.Key as YamlScalarNode)?.Value, Value = kv.Value as YamlScalarNode })
+                             .Where(kv => kv.Value != null)
+                             .Where(cc => cc.Key == "image")
+                             .FirstOrDefault()?.Value?.Value,
                     Title = c.Value.Children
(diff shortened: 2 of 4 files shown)


NicolasDorier:
so fucking coool

NicolasDorier:
love love it! Should we put some default images for the default template? :))

NicolasDorier:
I am thinking using the tea store that you just screenshot as default template. Can you share the content / images?

mariodian:
> I am thinking using the tea store that you just screenshot as default template. Can you share the content / images?

Yeah sure.

```
green tea:
  price: 1
  title: Green Tea
  description:  Lovely, fresh and tender, Meng Ding Gan Lu ("sweet dew") is grown in the lush Meng Ding Mountains of the southwestern province of Sichuan where it has been cultivated for over a thousand years.
  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg

black tea:
  price: 1
  title: Black Tea
  description: Tian Jian Tian Jian means 'heavenly tippy tea' in Chinese, and it describes the finest grade of dark tea. Our Tian Jian dark tea is from Hunan province which is famous for making some of the best dark teas available. 
  image: https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg

rooibos:
  price: 1.2
  title: Rooibos
  description: Rooibos is a dramatic red tea made from a South African herb that contains polyphenols and flavonoids. Often called “African redbush tea,” Rooibos herbal tea delights the senses and delivers potential health benefits with each caffeine-free sip.
  image: https://cdn.pixabay.com/photo/2017/01/08/08/14/water-1962388__480.jpg

pu erh:
  price: 2
  title: Pu Erh
  description: This loose pur-erh tea is produced in Yunnan Province, China. The process in a relatively high humidity environment has mellowed the elemental character of the tea when compared to young Pu-erh.
  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg

herbal tea:
  price: 1.8
  title: Herbal Tea
  description: Chamomile tea is made from the flower heads of the chamomile plant. The medicinal use of chamomile dates back to the ancient Egyptians, Romans and Greeks. 
  image: https://cdn.pixabay.com/photo/2015/07/02/20/57/chamomile-829538__480.jpg

fruit tea:
  price: 1.5
  title: Fruit Tea
  description: The Tibetan Himalayas, the land is majestic and beautiful—a spiritual place where, despite the perilous environment, many journey seeking enlightenment.
  image: https://cdn.pixabay.com/photo/2016/09/16/11/24/darts-1673812__480.jpg
```

I also wanted to make a small change to the padding. When you open the POS on a phone, the items are too close to left and ride edges. It irritates me a bit? Also the last two items should also be 100% wide. So I think I'll fix it in the next PR?

NicolasDorier:
cool yep! Change also the default template to this one. It rocks.

mariodian:
Alright, will do! But will have to figure out how to fix the last two items without changing the behavior on large screens.

mariodian:
@NicolasDorier there's a slight UI dilemma with this new picture layout.

Every `(i+4) % 4 = 1` item will be 100% wide. Which will look like this:

screen shot 2018-11-10 at 3 42 15 pm

Now the question is whether we want it that way or we want items to be uniform width. 

In that case items will have to be aligned to the left and it will look like so:

screen shot 2018-11-10 at 3 45 51 pm

Which one do you prefer?

ketominer:
This is so cool! Makes me think, what about the possibility to select multiple items then "checkout"?

mariodian:
Yeah was thinking about it but that’s for another thread. Also I don’t know C# so it will have to be done by someone else

NicolasDorier:
Multi item checkout is quite outside of scope. Might be good idea for future though.

@mariodian I like both. I would have prefered that it was uniform and centered though.

rockstardev:
Looks awesome @mariodian, great job

mariodian:
> I would have prefered that it was uniform and centered though.

It can't be uniform and centered with even number of columns though. It will look strange. 

Also, it won't be easy with bootstrap to center the last item while retaining responsiveness. The reason it worked with your old design is because items didn't have borders and you couldn't see that the last item was stretched from side to side.

I'd say let's go with the first design (the last item will be stretched) so it looks good even for those that only have <= 3 items in their POS. 

We could perhaps have switch in the admin area in the future that would let you set a different layout? We could even make it super-customizable by letting you add bootstrap grid classes directly and thus have any layout you want.

NicolasDorier:
good for me, anyway this look way better now than it was before. Both choice are good for me.

PR #393, opened 2018-11-11
Checkout Experience Language Setting

Kukks:
Fixes:
If you had en-US selected as the default store, the dropdown would select the first item in the list instead since we changed the locale to be en instead now.


Kukks:
@NicolasDorier

BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
```diff
@@ -58,7 +58,7 @@ public void SetCryptoCurrencies(ExplorerClientProvider explorerProvider, string
 
         public void SetLanguages(LanguageService langService, string defaultLang)
         {
-            defaultLang = defaultLang ?? "en";
+            defaultLang = langService.GetLanguages().Any(language => language.Code == default)? defaultLang : "en";
```
NicolasDorier: `defaultLang` not `default`.

PR #398, opened 2018-11-14
Tweaking UI for custom amounts

rockstardev:
Inspired by work @mariodian has done recently I've just done few small tweaks regarding Custom amount so that is now also rendered as card.

Plus I added ability for any card to be custom amount, price will be turned into suggested minimum. Here is comparison on how it now looks side by side

tweaking-custom


rockstardev:
@NicolasDorier on comments:

1. I thought about removing ShowCustom, but that renders in new row + legacy... so I decided to leave it.
2. I went with both price:10 and custom:true because that allows you to prepopulate field and have minimum... like I see this mostly used for donations where you can suggest minimum (0.01 or 5 or 10 or whatever).

NicolasDorier:
If you can get the popup working on this, it will be awesome

PR #400, opened 2018-11-15
Fix format on expired invoice DON'T MERGE

britttttk:
\n was showing up in expired invoice. This change hides it but I couldn't get a newline to work so maybe it could just be removed.

### DIFF
--- a/BTCPayServer/wwwroot/locales/en.json
+++ b/BTCPayServer/wwwroot/locales/en.json
@@ -31,7 +31,7 @@
   "Invoice expiring soon...": "Invoice expiring soon...",
   "Invoice expired": "Invoice expired",
   "What happened?": "What happened?",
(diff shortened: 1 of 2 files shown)


NicolasDorier:
you need to fix it in transiflex, I generate those file automatically now.

britttttk:
I uploaded the en.json file in this PR to Transifex for the new source file because Transifex recognizes /n and creates a symbol for it, for translator readability. It also brings your comment into the files.

britttttk:
I broke them only on Transifex right? A modified string in the source file will become an untranslated string in the translations. There is a history tab in the translation editor to bring them back easily.

britttttk:
If it's \n in Transifex in the source code, I don't understand how it becomes \\n from their translation. I think the user just copies the symbol. When I download the translated file for use it's still \n. Neither create a newline in the invoice so does the utility help this or only just converting it for fr and en.

NicolasDorier:
I mean right now if I run my scripts, all the translations for `InvoiceExpired_Body_1 ` are screwed, so can you revert what you did on transifex?

NicolasDorier:
I run the test `PullTransifexTranslations` to pull translation from transifex and update BTCPay project.

Issue #243, opened 2018-08-08
[i18n] Translate the PointOfSales app


junderw:
"Buy for `<<amount>>`" and "amount" and "Pay" are all English.

I would like to translate to Japanese:

"`<<amount>>`で購入", "金額", and "支払う" respectively

rockstardev:
When I did custom values in POS for "Buy" and "Donate" it was possible to localize whole page completely.

Now we have Cart mode that @mariodian has done, so we may want to consider refactoring all of this and providing either ability to customize all strings directly in POS or reference translation files from page.

I think doing robust translations like we have for Invoice is best course of action - we would just need to stabilize page enough and then plug in same process that @britttttk now does for checkout/invoice.

rockstardev:
I'm closing this issue since when initially opened we only needed to localize those two buttons - which was done.

If there is interest in localizing POS with cart checkout process - I would open new issue and track it independently. If we are to do this - I recommend start from clean slate - new BtcPayServer App that is specified and redesigned from ground up, so whoever is implementing it doesn't need to handle legacy dependencies and code for them

PR #403, opened 2018-11-16
Preserving title with custom amount

rockstardev:


rockstardev:
Button text was the only non-customizable part of the page so I added settings for that. Now your can customize your shop for whatever language as requested in #243.

Also, I added Custom CSS setting as suggested by @ketominer

Here is how everything looks now if you use [Darkly](https://bootswatch.com/darkly/) (https://bootswatch.com/4/darkly/bootstrap.min.css)

pos changes

Craddock7 (commented):
I approve this message

PR #408, opened 2018-11-19
Add POS Data to Invoice UI

Kukks:
This PR adds basic displaying of the pos data in the invoice details screen.


Kukks:
oops messy master head on my end

PR #409, opened 2018-11-19
Add POS Data to Invoice UI

Kukks:
This PR adds basic displaying of the pos data in the invoice details screen.


Kukks:
It passes now @NicolasDorier  :)

NicolasDorier:
Can you hijack one test somewhere and use this and check against the resulting view model?

Kukks:
added in a test

NicolasDorier:
Can you call `Invoices()` and check against the view model. It will cover more.

Kukks:
added some more @NicolasDorier

NicolasDorier:
Never use `ContinueWith` there is no reason to do so, `async/await`  is always better.

PR #411, opened 2018-11-20
Fixing MONA_BTC rate breaking tests and CircleCI because zaif is down

rockstardev:


BTCPayServer/BTCPayNetworkProvider.Monacoin.cs
```diff
@@ -24,7 +24,7 @@ public void InitMonacoin()
                 DefaultRateRules = new[] 
                 {
                                 "MONA_X = MONA_BTC * BTC_X",
-                                "MONA_BTC = zaif(MONA_BTC)"
```
NicolasDorier: Ping @cryptcoin-junkey what the hell with zaif?

PR #416, opened 2018-11-22
CircleCI multiarch Docker images

rockstardev:
Future is here ;)


Dockerfile.linuxarm32v7
```diff
@@ -0,0 +1,26 @@
+FROM microsoft/dotnet:2.1-sdk AS builder
```
NicolasDorier: I will change to specific version so we make sure we use same as x64

PR #428, opened 2018-11-30
Fix currency format for total amount

mariodian:
Fixed a bug where the total amount would be displayed in a weird way when using non-USD currencies (the bug had no impact on the functionality, just the UI) because they use `<space>`  instead of `,` for thousands, and `,` instead of `.` for decimals.

Since I don't have any useful information from the server about the currency format other than a currency sign ($, € etc.. and sometimes the srvModel returns no sign at all, for example for a currency like INR), and javascript can't natively work with currencies either, I have to hack it a bit.

It should display the correct total amount now for both type of currency formats.

### DIFF
--- a/BTCPayServer/wwwroot/cart/js/cart.js
+++ b/BTCPayServer/wwwroot/cart/js/cart.js
@@ -115,7 +115,7 @@ Cart.prototype.getTotal = function(plain) {
 }
 
 Cart.prototype.updateTotal = function() {
-    $('#js-cart-total').text(this.formatCurrency(this.getTotal()));
+    $('#js-cart-total').text(this.formatCurrency(this.getTotal(), srvModel.currencyCode, srvModel.currencySymbol));
 }
 
 Cart.prototype.updateAmount = function() {
@@ -135,7 +135,7 @@ Cart.prototype.escape = function(input) {
 Cart.prototype.listItems = function() {
     var $table = $('#js-cart-list').find('tbody'),
         self = this,
-        list = []
+        list = [],
         tableTemplate = '';
     
     if (this.content.length > 0) {
@@ -147,8 +147,9 @@ Cart.prototype.listItems = function() {
                 image = this.escape(item.image),
                 count = this.escape(item.count),
                 price = this.escape(item.price.formatted),
-                currencySymbol = this.escape(srvModel.currencySymbol),
+                total = this.escape(this.formatCurrency(this.getTotal(), srvModel.currencyCode, srvModel.currencySymbol)),
                 step = this.escape(srvModel.step),
+                tip = this.escape(this.tip || ''),
                 customTipText = this.escape(srvModel.customTipText);
 
             tableTemplate = '<tr data-id="' + id + '">' + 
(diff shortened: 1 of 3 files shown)


BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -163,17 +163,17 @@ Cart.prototype.listItems = function() {
...
 
-        tableTemplate = '<tr><td colspan="4"><div class="row"><div class="col-sm-8 py-2">' + customTipText + '</div><div class="col-sm-4">' +
+        tableTemplate = '<tr><td colspan="4"><div class="row"><div class="col-sm-7 py-2">' + customTipText + '</div><div class="col-sm-5">' +
     '<div class="input-group">' + 
         '<div class="input-group-prepend">' +
-            '<span class="input-group-text">' + currencySymbol + '</span>' +
+            '<span class="input-group-text">' + (currencySymbol != 'null' ? currencySymbol : '<i class="fa fa-money"></i>') + '</span>' +
         '</div>' +
-        '<input class="js-cart-tip form-control" type="number" min="0" step="' + step + '" name="tip" placeholder="Amount">' +
+        '<input class="js-cart-tip form-control" type="number" min="0" step="' + step + '" value="' + (this.tip || '') + '" name="tip" placeholder="Amount">' +
```
NicolasDorier: js injection vector

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -163,17 +163,17 @@ Cart.prototype.listItems = function() {
...
     '</div>' +
     '</div></div></td></tr>';
         list.push($(tableTemplate));
 
-        tableTemplate = '<tr class="bg-light h4"><td colspan="2">Total</td><td colspan="2" align="right"><span id="js-cart-total">' + this.formatCurrency(this.getTotal()) + '</span></td></tr>';
+        tableTemplate = '<tr class="bg-light h4"><td colspan="1">Total</td><td colspan="3" align="right"><span id="js-cart-total">' + this.formatCurrency(this.getTotal()) + '</span></td></tr>';
```
NicolasDorier: js injection vector.

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -239,16 +239,34 @@ Cart.prototype.emptyList = function() {
     $table.html('<tr><td colspan="4">The cart is empty.</td></tr>');
 }
 
-/* Get the currency symbol from an existing amount and use it with the new amount*/
+/* Get the currency symbol from an existing amount and use it with the new amount */
```
NicolasDorier: You should not have do such hack. You can have currency info by the server, and put this in `srvModel`, and just use it.

BTCPayServer/wwwroot/cart/js/cart.js
mariodian: It's escaped.

BTCPayServer/wwwroot/cart/js/cart.js
mariodian: True, I'll see what I can come up with. I'm afraid we'll have to load extra js dependency.

mariodian: Or, we'll only support a few major currencies (USD, GBP, JPY, EUR, THB) and won't show any currency symbol for the rest.

NicolasDorier: I disagree. Let me do a commit to give you the info you need.

mariodian: It's ok I've done it already. Just not pushed yet.

NicolasDorier: Ok, normally you should have all information you need to format from server side from any arbitrary currency. You found what you need?

mariodian: All I need is currency sign and currency (eur, usd, rub...). But currency sign sometimes returns `null` for currencies like IND (indonesian rupee) in the srvModel.

NicolasDorier: you need thousand separator and the number of decimal.

mariodian: I need more than that. I need to know whether the currency sign is on the left or right, whether it's separated by space etc.

Because sometimes the amounts are as followed:

$1,000.00
1 000,00 €
1 000,00 Kč
CHF 1'000.00

What I'm doing now is, I'm hardcoding it for major currencies and using a default format for exotic currencies. Should be good enough in most cases.

NicolasDorier: ok let me know when you want me to review and merge

mariodian: @NicolasDorier I realized separators are indeed missing in srvModel and I can't figure out how to add it there. I can only retrieve number of decimals, code, name and symbol. Seems like I will need that commit you mentioned. Thanks!

mariodian: @NicolasDorier sorry to change my mind all the time, but I find out the data in srvModel isn't very consistent. 

Sometimes it returns symbol, sometimes it doesn't (even when an item amount does have the symbol included), it always returns divisibility 2 even when in some cases (like indonesian rupee it should be 0) etc. 

If you haven't started working on it, I will just hard code major currencies and then default the rest as I wanted the first time. I'm testing the currencies against the server so the hardcoded currencies will match what comes out of server.

That will be the easiest approach I think. I'll push the commit today.

mariodian:
Please review.

As I explained earlier, the reason I'm hardcoding currencies is because the data coming form server isn't consistent.

I think we should merge this PR first because the bug makes UI look bad for non-usd currencies and perhaps later focus on fixing the data coming from the server (I didn't want to touch currencies.txt so I don't break displaying currencies elsewhere) and then modifying formatCurrency() accordingly.

NicolasDorier:
Will fix globally today.

PR #435, opened 2018-12-06
check for enabled flag on translations 

Kukks:
So that we can have in-progress translations


NicolasDorier:
what problem does it solve?

Kukks:
So that we can get rid of the arabic PR and move the translations in there until we add rtl support

Also good for when we have translations not fully complete.

Kukks:
Didn't know about that flag there. Good enough

NicolasDorier:
Not a flag, just in the script I filter stuff that are not ready

PR #440, opened 2018-12-10
invoice data public again 

Kukks:
Don't merge just yet.

this exposes the invoice ui data in a `invoices/{invoiceId}` route when `Content-type` is set to `application/json`

I also do  a simple 
```
result.NotificationEmail = null;
result.NotificationUrl = null;
```
to strip out potentially unwanted shared details
Was there more stuff to hide @nicolasDorier ?

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -31,31 +31,63 @@ namespace BTCPayServer.Controllers
     {
         [HttpGet]
         [Route("invoices/{invoiceId}")]
+        [MediaTypeConstraint("application/json")]
+        [AllowAnonymous]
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -29,33 +29,65 @@ namespace BTCPayServer.Controllers
...
+            if (result == null)
+            {
+                return NotFound();
+            }
+
+            //strip any potentially unwanted merchant info
+            result.NotificationEmail = null;
+            result.NotificationUrl = null;
+
+            return Json(result);
```
NicolasDorier: why twice

NicolasDorier:
Please, do not reuse the same Model. This model is done only for a view, make a separate model for JSON.
You can probaly include more data with better json shape.

PR #453, opened 2018-12-13
Update Anyone can enable invoices text so its much more clear

2pac1:
I think that: 

Please confirm you want to allow outside world to create invoices in your store (via API).");

is much better then

Please confirm you want to allow anyone to create invoices in your store")


NicolasDorier:
Ping @bitcoinshirt does it makes sense?

bitcoinshirt:
Yeah, agreed with @Kukks 
I don't think there's a way for newbs to understand this anyway, but we [recently updated docs](https://docs.btcpayserver.org/faq-and-common-issues/faq-stores#allow-anyone-to-create-invoice) to explain it.

2pac1:
I am satisfied with help that was updated

Issue #432, opened 2018-12-04
POS new design


mariodian:
Hi, I think POS should have a more POS-like design with an always visible "shopping cart" for easier manipulation on tablet devices. Which is where it shall mostly be used by physical vendors.

I'm using standard bootstrap elements (except for the sidebar "shopping cart" which is customized a bit) so it's easy to colorize with bootswatch themes. 

![](http://magicmockups.com/media/screen/guest/99/a63cb4fb965b40d985272d5ac124dd33_12_1600.jpg)

A first semi-working prototype is [here](http://btcpay-pos.freedomnode.com/).

- Let's brainstorm what the ideal POS should look like in terms of UI and UX. 
- Also the question is whether we want to have 2 separate apps (the current POS which could be renamed to something like "eshop" and the new POS), change the design of the current POS, or have a switch in the administration to switch between the eshop UI or the new POS UI.

Also can someone please check whether the sidebar works correctly on very old Android tablets? Thanks!

bitcoinshirt:
## PoS
I agree with the concept. Looks nice and solves important UX problem for physical merchants.

> A first semi-working prototype is here.

The link does not work for me.

I agree with the idea that we split the eshop and PoS apps. Even though at one point there will be a dedicated iOS or Android App, I think we should offer this, mostly because in theory some users may not have a compatible device for the actual mobile app or simply someone may want to have it on a desktop.

## eShop

In order to have an eShop app, imo we need: 
* To allow billing/shipping address entry that will be added to invoices in BTCPay.
* To have TOS and privacy policies at the checkout (GDPR)
* Make product adding easier and intuitive 

For an e-shop we don't need anything too complex, whoever needs complexity can use WordPress, but this sort of app would allow merchants to run their store on hardware device, which is great use case.

I have a very old Android tablet somewhere and will test it out if you can fix that link.

mariodian:
Sorry, wrong link. Try again please http://btcpay-pos.freedomnode.com/

bitcoinshirt:
@darkban that's more like quick invoice 
@mariodian 
Okay, tested on HP 7, 4.1.2
- On default browser does not work at all. screenshoot
![oldbrowser](https://user-images.githubusercontent.com/36959754/49465294-8bdb6d00-f7fd-11e8-8b04-2811732a8c72.jpg)
- On Chrome 70.0.3538.110 it works, but the Pay button is cut off as well as btcpay logo. screenshoot
[...]

Issue #450, opened 2018-12-13
srvModel.currencyInfo wrong data


mariodian:
I see that a new `currencyInfo` object was added to `srvModel` (for PoS). However for many currencies it shows wrong data.

For example for Indian Rupee it looks like this:

```
{
currencySymbol: "INR"
decimalSeparator: "."
divisibility: 2
prefixed: true
thousandSeparator: ","
}
```

However, `srvModel.items` will show amounts with the correct symbol: `Buy for ₹ 1,000.00`

For example for Euro, the object looks alright (the Euro symbol is provided):

```
{
currencySymbol: "€"
decimalSeparator: ","
divisibility: 2
prefixed: false
thousandSeparator: " "
}
```

But the Euro symbol in `srvModel.items` is separated by space like so: `100,00 €` and the `currencyInfo` object returns the symbol without the space. And I have no way of knowing which currency uses a space separator for symbols and which don't.

Is it possible to fix this? Otherwise i'll have to stick with my custom formatCurrency function which predefines the symbol for major currencies.

mariodian:
Found other issues.

For example one of the Arab currencies IRR says the `divisibility` is 2:

```
{
currencySymbol: "IRR"
decimalSeparator: "٫"
divisibility: 2
prefixed: true
thousandSeparator: "٬"
}
```

but the `srvModel.items` show amounts like this `Buy for ریال1٬000` (thousand) which completely omits the decimal separator.

There are inconsistencies for most currencies except for 10 or so major ones.

NicolasDorier:
Don't come back to what you were doing, it was not working for JPY nor for BTC. I will fix this.

NicolasDorier:
It does not show `Buy for ریال1٬000` for me but `1,000/00ريال`

PR #454, opened 2018-12-13
New PoS design

mariodian:
Changed the UI (based on #432) so it looks more like a real PoS compatible with phones and tablets.

What's new:

- cart in a sidebar
- custom amount
- discount
- tipping buttons
- cart summary
- new add-to-cart visualization
- product search
- custom +/- cart item buttons
- better responsiveness


BTCPayServer/Controllers/AppsController.PointOfSale.cs
```diff
@@ -65,6 +65,9 @@ public PointOfSaleSettings()
             public string CustomButtonText { get; set; } = CUSTOM_BUTTON_TEXT_DEF;
             public const string CUSTOM_TIP_TEXT_DEF = "Do you want to leave a tip?";
             public string CustomTipText { get; set; } = CUSTOM_TIP_TEXT_DEF;
+            public const string CUSTOM_TIP_PERCENTAGES = "15,18,20";
+            public string CustomTipPercentages { get; set; } = CUSTOM_TIP_PERCENTAGES;
```
NicolasDorier: should be `int[]`, not string

NicolasDorier:
`CustomTipPercentages` should be `int[]`, and you should parse it inside the controller, not inside the view.

BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
```diff
@@ -36,8 +36,11 @@ public class UpdatePointOfSaleViewModel
         public string CustomButtonText { get; set; }
         [Required]
         [MaxLength(30)]
-        [Display(Name = "Do you want to leave a tip?")]
+        [Display(Name = "Text to display in the tip input")]
         public string CustomTipText { get; set; }
+        [MaxLength(30)]
+        [Display(Name = "Tip percentage amounts (comma separated)")]
+        public string CustomTipPercentages { get; set; }
```
NicolasDorier: this one is OK as string

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -218,88 +370,329 @@ Cart.prototype.listItems = function() {
...
+* Storage
+*/
 Cart.prototype.saveLocalStorage = function() {
-    localStorage.setItem(this.getStorageKey(), JSON.stringify(this.content));
+    localStorage.setItem('cart', JSON.stringify(this.content));
 }
 
 Cart.prototype.loadLocalStorage = function() {
-    this.content = $.parseJSON(localStorage.getItem(this.getStorageKey())) || [];
+    this.content = $.parseJSON(localStorage.getItem('cart')) || [];
```
NicolasDorier: why did you removed my `this.getStorageKey()`? I did this because there you start having conflict where the user use several PoS.

NicolasDorier:
You remove lots of my changes and fixes. (fixes about currency formatting for example, and storage key)
Please rebase it on latest.

BTCPayServer/wwwroot/cart/js/cart.js
mariodian: I didn't realize anyone touched the cart.js

BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
mariodian: I have no idea how to do that. I can't code C#.

mariodian:
I didn't realize you changed something. But anyway, as I said in #450 the server returns wrong data.

BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
mariodian: @NicolasDorier figured it out. Will push it soon.

mariodian:
> Put back the this.getStorageKey() you removed.

@NicolasDorier it's back already (0a3b383) I just extended a bit so it works with tip, custom amount and discount too.

I tested the formatting and seems to work correctly now.

NicolasDorier:
@mariodian so it is ready to merge again?

mariodian:
Would be nice if more people tested it in case I missed something.

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -142,43 +313,23 @@ Cart.prototype.listItems = function() {
...
-                        '   <div class="input-group-append"><a class="js-cart-item-remove btn btn-danger btn-sm" href="#"><i class="fa fa-remove"></i></a></div>' + 
-                        '</div></td>' +
-                        '<td class="align-middle" align="right">' + price + '</td>' +
-                        '</tr>';
+                image = this.escape(item.image);
+
+            tableTemplate = this.template($('#template-cart-item'), {
+                'id': this.escape(item.id),
+                'image': image ? this.template($('#template-cart-item-image'), {
+                    'image' : image
```
NicolasDorier: no escape?

NicolasDorier:
I am just a bit worried about the home grown templating system. If there is only once JS injection vector, it might be used to gain control of the cookie of an unsupecting admin.

Wished you used Vue instead. But well, that will do for now.

BTCPayServer/wwwroot/cart/js/cart.js
mariodian: both image and id are escaped

mariodian:
I’m learning vue but it will take some time until I’m ready to rewrite it.

britttttk:
I'll test it for you @mariodian

britttttk:
Hey @mariodian,

1. Using trash icon to delete line item
    * **Issue**: If you press the Trash icon on the first product item in your store, it will empty the entire cart. No matter where this item is in the cart. It only happens when using the trash feature on the first item in your product list. In your default template for example, it's Green Tea. 

2. Exit the tip/confirmation page and return back to it
    * **Issue**: If you previously had a total and had chosen a tip, upon returning to the confirmation page, it displays the previous tip amount. If the cart total changes, between confirmation pages, the confirmation page should either update the tip value using the previous percent or reset to zero. Mainly because it's an extra step for the merchant to have to remember to clear it or ask the customer to re-tip. 
    * **Example**: If they request to remove an item and then return back to the confirmation page, they see a new total, their "previous tip" and they press confirm but they would actually be paying a % value higher than their original agreement but they would need to be pretty quick with math to see it. This could make a customer mad..

3. Line item as 'Qty' holds previous value
    * **Issue**: You can enter confirmation page state with line items as 'Qty' producing incorrect total.     
    * **Example**:
image

mariodian:
@britttttk 

> Using trash icon to delete line item

Got it

> Exit the tip/confirmation page and return back to it

I think the correct way is to reset it to 0 because if I type in the tip amount instead of click on the buttons (15% etc) I still may not want to change the tip when the value changes. 

So just remove it for both cases so I don't have to remember the method I used for tipping.

> Line item as 'Qty' holds previous value

Got it. Thanks! I'll have to figure out how to handle such situation. Whether to handle it as a 0 or not letting you click the confirm button.

mariodian:
1) fixed
2) I'm setting the tip to $0 every time the total amount changes
3) a cart item is considered $0 when the qty doesn't have a numeric value.

NicolasDorier:
@mariodian I was wondering, why are you saving the cart in storage instead of just a local variable scoped in the script?

NicolasDorier:
@britttttk time to test this once again? I will review again. I really want this merged.

britttttk:
@mariodian,

Thanks for fixing my comments 1-3. I tested it again and these are my only remaining comments:
<br>

Related to previous issues 2 & 3: 

> 2. I'm setting the tip to $0 every time the total amount changes
> 3. a cart item is considered $0 when the qty doesn't have a numeric value.

 *   **Issue A**: When the line item is equal to 'qty', it's value is equal to $0 in both the cart and confirmation page (your solution to previous issue 3 works perfectly). However, this situation is the only case where previous issue 2 is not solved. As you can see in the photo below, the previous tip has persisted across a change in cart value when line items are set to 'qty' (Tip shown is 33%).

image

<br>

 *   **Issue B**: It's possible to create negative cart values when deleting line items using the trash feature. This happens when you add some items to the cart, like this:

Product A = Quantity 1
Product B = Quantity 5

If you set Product B equal to 'qty' (by highlighting it and deleting the value of the line item, it defaults to 'qty') and you do this prior to deleting it as a line item, using the trash feature, you can produce a negative cart value of -4 items. You don't allow negative cart values to persist into the confirmation page, as it's locked when the cart is equal to a negative value. So it's really only a UI issue.

image

NicolasDorier:
@mariodian before fixing the points of @britttttk can you rebase?
I merged https://github.com/btcpayserver/btcpayserver/pull/470 which may or may not have impact.

mariodian:
@NicolasDorier rebased

@britttttk fixed. Thanks!

NicolasDorier:
@mariodian added a commit to fix the tests, if you make change, please pull your branch before.

PR #459, opened 2018-12-17
Update es-ES.json

andres-pinilla:
With the current translation the text is very long and looks bad: https://imgur.com/a/sqBYahO


britttttk:
Lo siento :) but you need to make the changes in Transifex only.

andres-pinilla:
🥇

Issue #455, opened 2018-12-13
"Disable user-registration" to "Allow/Enable" and uncheck by default


bitcoinshirt:
The ability for other users to register and use someone else's BTPay Server should be disabled by default and imo the user needs to give an explicit permission for others to register.

Admins in most cases do not want to allow others to use their server. In some jurisdictions it may be even tricky and the user (even though it's stupid) can be considered an illegal payment processor.

So my proposal is to disable the registration for others in policy and change it to "Enable user registration" With a description: Allow other users to register and use your server. 

The admin should give a permission explicitly.

kubko32459:
I agree.

sipsorcery:
Like this?

![serversettings](https://user-images.githubusercontent.com/197660/50058710-6e52bf80-017c-11e9-82e9-de35ebe4f040.png)

sipsorcery:
Subsequent to last post I noticed that the setting seems to already exist. Does the "Disable Registration" setting on the Policies page already satisfy this issue?

![policies](https://user-images.githubusercontent.com/197660/50058962-ad830f80-0180-11e9-9ca6-27c129941157.png)

bitcoinshirt:
@sipsorcery Yes, I was referring to "Disable Registration" in policies. 
I think that one is unchecked by default and that's bad for the reasons I provided in the first post.

Perhaps it's better to ask users if they want to Enable it and leave it unchecked, that's much better logic than with having it enabled out of the box, most of the people don't want other people on their server.

What do you think?

sipsorcery:
@bitcoinshirt personally I'd vote to have it disabled by default. It's an easy setting to miss, I did. I would think someone who want other users to create stores will find and apply the setting but not vice-versa.

I'll submit a PR to disable it by default and let the power(s) that be decide.

bitcoinshirt:
Nice, thanks marking this as "prepare to close".

PR #463, opened 2018-12-18
Set disable registration as default true

sipsorcery:
For safety user registration is defaulted to true as per #455.


NicolasDorier:
If I do this, how does the first user can create the admin account?

Kukks:
@sipsorcery  you need to enable the registration when there are no users registered at the very least then :)

sipsorcery:
Adjusted so that registrations will always be allowed if there are no user records.

BTCPayServer/Views/Shared/_Layout.cshtml
```diff
@@ -74,7 +74,7 @@
                         </li>}
                     else
                     {
-                        @if (themeManager.ShowRegister)
+                        @if (themeManager.ShowRegister || UserManager.Users.Count() == 0)
```
NicolasDorier: This mean we do one additional database call for every page if `ShowRegister` is false. Massive perf hit for a condition never satisfied.

NicolasDorier:
@sipsorcery I would prefer you keep True by default, and when the admin is registered, you change the setting to lock subscriptions.

sipsorcery:
@NicolasDorier PR is fixed to work as you recommended. 

For the tests are the instructions [here](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Tests/README.md) missing a step or two? `docker-compose up dev` reports that it can't find a valid configuration file. Has it been moved to the .circleci directory or is there some intermediate step needed to generate a docker-compose file?

NicolasDorier:
@sipsorcery `docker-compose up dev` if you want to develop and run tests from visual studio.
`docker-compose run tests` run the tests from inside a container.

NicolasDorier:
The tests still failing. You probably want to disable this feature for the tests.

NicolasDorier:
@sipsorcery anytime to fix the tests ? I think this should definitively be merged once fixed.

sipsorcery:
@NicolasDorier I was planning on fixing the tests once I’m back from Xmas travels (no computer at the moment).

sipsorcery:
@NicolasDorier added `debug` environment check so this policy setting won't be applied when running unit tests. Seemed a bit better than an additional command line option.

NicolasDorier:
@sipsorcery can you instead just add an option with the command line? like `--disable-registration`/`DisableRegistration` default to `true` and just set it in the `PayTester` for unit tests to `false`?

NicolasDorier:
Awesome, I really wanted that.

PR #467, opened 2018-12-20
CoinSwitch Integration

Kukks:


NicolasDorier:
Anybody willing to test that? @astupidmoose ?

BTCPayServer/wwwroot/checkout/coinswitch.html
```diff
@@ -0,0 +1,88 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>CoinSwitch</title>
+    <script>
+        var script = document.createElement('script'),
+            head = document.head || document.getElementsByTagName('head')[0];
+        script.src = window.location.protocol + '//files.coinswitch.co/public/js/cs_switch.js';
```
NicolasDorier: why not using always https?

Kukks: I figured it might cause some sort of mismatching protocol errors? Not 100% if that is a thing with actual script files.

NicolasDorier: I don't think so, I think you should just use https

astupidmoose:
Is there a test server this is implemented on? 

I'm #reckless, but not that much.. lol.

PR #492, opened 2019-01-06
Crowdfund Part 1: JS Dependencies

Kukks:
* AnimeJs : cool client side animations lib
* babel-polyfill: for older browser support needed for some js libs
* Bootstrap-Vue: Complete replacement for the jquery powered official bootstrap  js  and provides 1st class support in vue for bootstrap 4
*  flatpickr: date time picker for admin 
* jquery.richtext: HTML WYSIWYG editor for admin
* signlar: Multi transport method abstraction layer with .net core
* vue-toasted: toast notifications ofr displaying errors or payments to the app
* moment: time and date js lib


NicolasDorier:
no `min` for signalr?

NicolasDorier:
Can you try to use https://docs.microsoft.com/en-us/aspnet/core/client-side/libman/?view=aspnetcore-2.2 ? just to see if we can prevent adding tons of deps directly into the repo

Kukks:
I didn't download min files where possible on purpose: the bundler minifies them itself and when files are minified twice ive noticed some weird behavior

Kukks:
Ok I gave it a go but it creates a lot of junk files and not always the
correct things we need. I think we should stick with the current structure
and then find a better solution
ᐧ

NicolasDorier:
> the  bundler  minimizes it which is why i put in non min

Eskyee:
Looking good 🤛🥳 Kukks

PR #496, opened 2019-01-06
SMTP e-mail settings on a store level 

### LINKED ISSUE #468: SMTP e-mail settings on a store level
We should we allow non-admins to configure SMTP setting, possibly on store-level.

Use case: 

1. The user wants to have a payment button e-mail notification, but uses someone else’s server, sees the e-mail settings but can’t create SMPT since he’s not the admin.

2. Same goes for #313 I assume one would want to send e-mail confirmations or a thank you emails to whoever supports the campaign.

Kukks:
closes #468

### DIFF
--- a/BTCPayServer/Services/Mails/EmailSenderFactory.cs
+++ b/BTCPayServer/Services/Mails/EmailSenderFactory.cs
@@ -1,0 +1,31 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
+using BTCPayServer.Services.Stores;
+
+namespace BTCPayServer.Services.Mails
+{
+    public class EmailSenderFactory
+    {
+        private readonly IBackgroundJobClient _JobClient;
+        private readonly SettingsRepository _Repository;
+        private readonly StoreRepository _StoreRepository;
+
+        public EmailSenderFactory(IBackgroundJobClient jobClient,
+            SettingsRepository repository,
+            StoreRepository storeRepository)
+        {
+            _JobClient = jobClient;
+            _Repository = repository;
+            _StoreRepository = storeRepository;
+        }
+
+        public IEmailSender GetEmailSender(string storeId = null)
+        {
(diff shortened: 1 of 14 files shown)


NicolasDorier:
The email sender should not know about `StoreId`. Instead, the email sender instance should be created from the store instead.

Kukks:
@NicolasDorier  Should I make an `EmailSenderFactory` then?  I'm trying to recycle the instances so that we dont have to recreate them for each background task

NicolasDorier:
Yes I think having a EmailSenderFactory is good.

Kukks:
@NicolasDorier  done :)

NicolasDorier:
No `ContinueWith` please, just call it without waiting.
To remove the warning, I think you can use `Task.Run(blablaAsync())`

BTCPayServer/Services/Mails/EmailSenderFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+        public EmailSenderFactory(IBackgroundJobClient jobClient,
+            SettingsRepository repository,
+            StoreRepository storeRepository,
+            EventAggregator eventAggregator)
+        {
+            _JobClient = jobClient;
+            _Repository = repository;
+            _StoreRepository = storeRepository;
+            eventAggregator.Subscribe<EmailSettingsUpdated>(updated =>
+                _CachedClients.TryRemove(updated.StoreId ?? ServerSettingsKey, out var x));
```
NicolasDorier: you never unsubscribe, this is a leak.

Kukks:
```

> No `ContinueWith` please, just call it without waiting.
> To remove the warning, I think you can use `Task.Run(blablaAsync())`

Already fixed it

Kukks:
this is ready for review again @NicolasDorier

NicolasDorier:
I will rewrite this. `GetEmailSender` does not have to be asking. One sec.

PR #504, opened 2019-01-08
Fix Coinswitch Issues, Fix LN Node Info Clipboard, Fix Vue-Cloak Styles

### LINKED ISSUE #503: Coinswitch Issues
I've found a couple issues with Coinswitch (cc @Kukks):

1. When pay with coinswitch is clicked, the changelly button disappears. Screen needs refreshing before it comes back even if coinswitch is closed. 

2. (Browser Related) - If selected as a popup, many browsers will block this. Not really something we can control, but perhaps something that needs documentation

3. Coinswitch itself seems to have an issue calculating its own rate. Example: Coinswitch quotes "171.29 UBQ", then proceeds to ask for "169.309 UBQ". It quotes and then immediately asks for less. Its very strange, but not BTCPay related. 

4. Coinswitch BTCPay merchant setup page is broken, their setup process has errors which prevent merchants from properly signing up right now, again their issue not ours.

Kukks:
closes #503

Fixes the clipboard copy issue in the new LN Node Info Page for Firefox and Safari

Fixes the split second stutter on checkout page

### DIFF
--- a/BTCPayServer/Models/StoreViewModels/UpdateCoinSwitchSettingsViewModel.cs
+++ b/BTCPayServer/Models/StoreViewModels/UpdateCoinSwitchSettingsViewModel.cs
@@ -11,7 +11,7 @@ namespace BTCPayServer.Models.StoreViewModels
         public bool Enabled { get; set; }
 
         [Display(Name = "Integration Mode")]
-        public string Mode { get; set; } = "popup";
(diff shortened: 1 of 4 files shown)


NicolasDorier:
@britttttk @andres-pinilla the strings should be translateable @Kukks says

andres-pinilla:
Thanks @NicolasDorier

PR #521, opened 2019-01-13
WIP: Selenium Tests

britttttk:
I made some Chrome Selenium tests.

I'm not sure how to make it work with Docker and connect to the regtest instance to run the tests though. 

Right now I just check the tests on the mainnet.demo instance. They will fail on the regtest instance because the tests are currently using a 3rd party instance view (the header is different). 

Any advice is appreciated it's my first time using Selenium.

### DIFF
--- a/BTCPayServer.Tests/BTCPayServer.Tests.csproj
+++ b/BTCPayServer.Tests/BTCPayServer.Tests.csproj
@@ -11,6 +11,9 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="selenium-server-standalone" Version="3.9.1.1" />
+    <PackageReference Include="Selenium.WebDriver" Version="3.141.0" />
+    <PackageReference Include="Selenium.WebDriver.ChromeDriver" Version="2.45.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Awesome, so can you check the other tests? In each tests I start a new instance of btcpay with a very simple line of code.
I think you just need to do this, and use the URL of this new instance instead of the mainnet site.

Kukks:
Wrap you unit test method bodies in `using (var tester = ServerTester.Create())` 

With the `tester` instance, you can create users, stores, register derivations, etc and set up the local data you need for your e2e tests. To get the uri of the local instance it's ` tester.PayTester.ServerUri`.

Look at the other tests to get the bigger picture. :)

Kukks:
Did you also do `tester.Start();` after the `using` ?

NicolasDorier:
The address is not determist you need to use something like `tester.PayTester.HostUri` or something like this.

Kukks:
@britttttk  we got an update a few days ago and the test server should work now :) Can you give it another go or should I take over?

Kukks:
@britttttk  ping :)

britttttk:
@Kukks wow it does work now :D

NicolasDorier:
cool! Can you rebase so we merge this? not later than 1H ago, it could have catch a bug that is currently not caught into the current test suite. (That's why I bumped from version .97 to version .99 without ever publishing the .98 publicly :p)

Issue #523, opened 2019-01-14
AspNet.Core 2.1.6 or 2.1.4?


bolatovumar:
Documentation [here](https://github.com/btcpayserver/btcpayserver#how-to-build) specifies .NET Core v2.1.4 while the .csproj file specifies v2.1.6 ([see here](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/BTCPayServer.csproj#L72)). Are these things referring to the same thing but the versions are in conflict? Or am I missing something here?

Kukks:
So, we currently run .net core 2.1, I think specifically we use 2.1.500 SDK but the other newer ones (2.1.502 and 2.1.503) might work too. 

So it is v2.1.6 for now. We plan to update on the .net core 3 release.

bolatovumar:
Should the documentation be updated to say to use v2.1.6 instead of v2.1.4 then?

PR #528, opened 2019-01-15
Update .NET Core SDK version in documentation

bolatovumar:
Address #523


bolatovumar:
The patch version note was removed as it's already included in v2.1.6 and no separate security patch needs to be applied.

bolatovumar:
Looks like a certain unit test is failing but the test has nothing to do with this change.

PR #531, opened 2019-01-15
fix merge bug

Kukks:


Kukks:
@nicolasdorier  ping, this one is a bit important

Issue #461, opened 2018-12-18
Feature Request: Subscription KISS


nvk:
Create an end point for seller to generate/send subscription invoices/QR email. Or better let the seller create a rudimentary scheduler to send out these invoices. If the later, could be based customer email, choose SKU and interval.

Background: Many of the Patreon users choose it because of reoccurring payments. Bitcoin is push only, so a good work around is to give the seller an option send the 'patrons' a monthly invoice with QR.

PS: didn't spend a lot of time on this, one of your contributors suggested I place my [tweet/request here](https://twitter.com/nvk/status/1074762447020482564)

Keep up the great work.

bitcoinshirt:
Somewhat related and can be implemented in combination with #314 .

Sjors:
Subscriptions / recurring invoices would be great for SAAS in general. For the Patreon use case, it would allow some simplification of @JeffVandrewJr's LibePatron.

I wonder what's necessary for this. Is there a (daily) cron that can be leveraged to generate new invoices where needed?

At the data model level I think a `SubscriptionData` object could be like this:

```
public class SubscriptionData
    {
        public string StoreDataId
        {
            get; set;
        }
        public StoreData StoreData
        {
            get; set;
        }

        public string Id
        {
            get; set;
        }

        public DateTimeOffset Created
        {
            get; set;
        }

        public byte[] Blob
        {
            get; set;
        }
        public string ItemCode
        {
            get;
            set;
        }
        public string OrderId
        {
            get;
            set;
        }
        public string Status
        {
            get;
            set;
        }
        public string CustomerEmail
        {
            get;
            set;
        }
        public List<InvoiceData> Invoices { get; set; }
    }
```

For status and exceptions, the SAAS can always inspect invidual invoices, but it's nice to have a quick way of knowing if the subscription is active. 

```
public enum SubscriptionStatus
    {
        New, // First invoice sent, not complete
        Active, // Most recent invoice is complete
        PaymentPending, // Most recent invoice(s) incomplete (behavior is application specific),
        Cancelled
    }
```

Customer info gets copied into a `InvoiceData` object at each billing interval (the redundancy is fine, because if a customer changes their address that shouldn't impact previous invoices).

Kukks:
This is actually not that hard to implement at all. Just been busy :)

9mido:
I am not sure if this already exists in btcpayserver or not. It may need a separate issue so that it gets added in.

[...]

PR #532, opened 2019-01-15
Payment Requests

### LINKED ISSUE #314: Payment Requests
I regularly pay freelancers with Bitcoin. The problem is that once we agree on the amount, due to volatility, the amount changes, so sometimes I have to pay again, or they have to refund, etc. It's a pain. Especially if we agree on the amount a few days before the payment. Someone is always at a loss.

I've seen that Blokonomics[ launched a very nice tool.](https://www.blockonomics.co/invoicecreate/) From what I see it's an invoice which creates a countdown every time you open of the link with dynamic amounts refreshing.

screen shot 2018-10-05 at 09 50 28

I think we can do this easily to make BTCPay useful for freelancers. @Kukks and I discussed it a while ago, but we forgot about it, so I'm opening an issue for future reference and discussion.

### LINKED ISSUE #571: The App page is titled as Stores
This page title says 'Stores' when you are on the Apps page.

All the other pages have titles which match their header title. 

image

Kukks:
With this PR, I am introducing a new entity to BtcPay, Payment Requests as discussed in #314 

Closes #314 
Also Closes #571

A payment request is exactly what it sounds, a request to be paid. It consists of a title, description, amount and an optional expiry date. 

Until it expires, it can be paid at any time by creating a fresh invoice of the remaining balance on demand. Only one invoice can be active at a time per payment request.

It comes with the same real-time updates as the crowdfunding but with less colorful aspects to it. :) 

The UI will also display all invoices created through it along with all payment, tx ids and block explorer links. 

I will also add a print-friendly version that can be used to print for accounting purposes. 

Eventually, I will build #461  on top of Payment Requests since a payment request is perfect to send to an email on a schedule.

image


Kukks:
Added UTs, ready for you @NicolasDorier  💃

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -428,6 +431,62 @@ public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
...
+            }
+
+            await Task.WhenAll(finalTasks);
+
+            return result;
+        }
+
+         public Dictionary<string, decimal> GetCurrentContributionAmountStats(InvoiceEntity[] invoices, bool usePaymentData = true)
+        {
+            if(usePaymentData){
```
NicolasDorier: format

BTCPayServer/Migrations/20190116102421_AddPaymentRequests.Designer.cs
```diff
@@ -0,0 +1,602 @@
+﻿// <auto-generated />
+using System;
+using BTCPayServer.Data;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace BTCPayServer.Migrations
```
NicolasDorier: What is this file? I don't remember  any file like this needed to be generated by entity when changing the DB.

BTCPayServer/PaymentRequest/PaymentRequestHub.cs
```diff
@@ -0,0 +1,211 @@
...
+                AmountCollected = amountCollected,
+                CurrencyData = _AppsHelper.GetCurrencyData(blob.Currency, true),
+                LastUpdated = DateTime.Now,
+                Invoices = invoices.Select(entity => new ViewPaymentRequestViewModel.PaymentRequestInvoice()
+                {
+                    Id = entity.Id,
+                    Amount = entity.ProductInformation.Price,
+                    Currency = entity.ProductInformation.Currency,
+                    ExpiryDate = entity.ExpirationTime.DateTime,
+                    Status = Enum.GetName(typeof(InvoiceStatus), entity.Status),
```
NicolasDorier: Don't use this, there is a `GetState()` I think that now you can use which indicate both the status and the exceptionStatus properly.

BTCPayServer/Services/PaymentRequests/PaymentRequestRepository.cs
```diff
@@ -0,0 +1,196 @@
...
+using NBitcoin;
+using NBXplorer;
+
+namespace BTCPayServer.Services.PaymentRequests
+{
+    public class PaymentRequestRepository 
+    {
+        private readonly ApplicationDbContextFactory _ContextFactory;
+        private readonly InvoiceRepository _InvoiceRepository;
+        private readonly EventAggregator _EventAggregator;
```
NicolasDorier: No event aggregator in the Repository, this is not the goal. A repository just abstract the storage technology.

BTCPayServer/Migrations/20190116102421_AddPaymentRequests.Designer.cs
Kukks: It was generated by the .net cli ef core tool. I dare not touch these :)

Kukks:
PR changes done @NicolasDorier deet

Kukks:
don't forget to review or I will just keep adding stuff @NicolasDorier 

 I refactored the status to make the hosted service lighter

BTCPayServer/Migrations/20190121133309_AddPaymentRequests.Designer.cs
```diff
@@ -0,0 +1,604 @@
+﻿// <auto-generated />
```
NicolasDorier: What is this file? it just rebuild the whole Models of all table Oo

NicolasDorier: I think you can remove it

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -428,6 +428,62 @@ public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
...
+                    var tResult = await rateTask.Value;
+                    var rate = tResult.BidAsk?.Bid;
+                    if (rate == null) return;
+
+                    foreach (var stat in stats)
+                    {
+                        if (string.Equals(PaymentMethodId.Parse(stat.Key).CryptoCode, rateTask.Key.Right,
+                            StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            result += (1m / rate.Value) *  stat.Value;
```
NicolasDorier: bug, additions are not atomic.
You are over complicating, you don't need `finalTasks` at all.

BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -0,0 +1,315 @@
...
+            return View("Confirm", new ConfirmModel()
+            {
+                Title = $"Remove Payment Request",
+                Description = $"Are you sure to remove access to remove payment request '{blob.Title}' ?",
+                Action = "Delete"
+            });
+        }
+
+        [HttpPost]
+        [Route("remove/{id}")]
```
NicolasDorier: can you instead `{id}/remove` ?

BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -0,0 +1,315 @@
...
+            _RateFetcher = rateFetcher;
+            _BtcPayNetworkProvider = btcPayNetworkProvider;
+            _PaymentRequestRepository = paymentRequestRepository;
+            _PaymentRequestStreamer = paymentRequestStreamer;
+            _EventAggregator = eventAggregator;
+            _AppsHelper = appsHelper;
+        }
+
+        [HttpGet]
+        [Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]
```
NicolasDorier: Put the authorize at the `Controller` level.

BTCPayServer/Services/PaymentRequests/PaymentRequestRepository.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        public PaymentRequestStatus Status { get; set; }
+
+        public byte[] Blob { get; set; }
+
+        public PaymentRequestBlob GetBlob()
+        {
+            var result = Blob == null
+                ? new PaymentRequestBlob()
+                : new Serializer(Network.Main).ToObject<PaymentRequestBlob>(Encoding.UTF8.GetString(Blob));
```
NicolasDorier: Don't hard code `Network.Main`. Why do you need the serializer at all? It seems to me you can use JsonConvert directly.

Kukks:
ready for another round @NicolasDorier  :P

NicolasDorier:
btw, let's keep it like that, but in the future I would prefer if JS was not mandatory, and instead just used to improve the user experience if available.
The apps are different.

Kukks:
I can do it, it's not a big thing. Give me an hour

Kukks:
Added No js version @NicolasDorier

britttttk:
This is an awesome feature, I'm super exited for this one.
My comments:

1. **expiry** - I've never heard this word used before in the US :P usually expiration 
2. **Calendar**:
    - What is the purpose of allowing users to choose past dates, if there is an option for no expiration? I don't think it can be paid if it's created as expired?
    - Is it possible, when the user re-opens the calendar for it to open to the date they have chosen?
    - Is it possible for calendar info to be kept through a PayReq error? I mean if they choose Bitcoin instead of BTC, the error deletes their calendar info and they have to do it again.
3. **No Deriv** - Causes the 'Make Payment' button on the PayReq page to hang forever... would maybe be better if it does nothing when pressed (how it is for CF App).
4. **Currency Choice**:
2currencytypes
    - What is the correct option here? One choice? A list?
    - The error message is very strange requesting a string or an array type? BTC, LTC doesn't work.
    - The error message displays incorrectly: "The Field The Currency..."
5. **UI Dashes:** - Normally there are dashes here
dashes

Kukks:
resolved all above ^

britttttk:
@Kukks Thanks for fixing everything. 

Should the Currency Field* be starred to show it's required? The light BTC text is great but if the user assumes it's pre-typed for you, the page hangs when they try to save? It's giving me a null exception error in VS. I'm not sure how that appears to the user. Is it ok?

NicolasDorier:
there is still some `.designer`

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -441,6 +441,62 @@ public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
                 .ToArray();
         }
 
+        public async Task<decimal> GetCurrentContributionAmount(Dictionary<string, decimal> stats, string primaryCurrency, RateRules rateRules)
```
NicolasDorier: Why are you making it so complicated?

```csharp
        private async Task<decimal> GetCurrentContributionAmount(Dictionary<string, decimal> stats, string primaryCurrency, RateRules rateRules)
        {
            decimal result = 0;

            var ratesTask = _RateFetcher .FetchRates(
                stats.Keys
                    .Select((x) => new CurrencyPair( primaryCurrency, PaymentMethodId.Parse(x).CryptoCode))
                    .Distinct()
                    .ToHashSet(), 
                rateRules);
            foreach (var rateTask in ratesTask)
            {
                    var tResult = await rateTask.Value;
                    var rate = tResult.BidAsk?.Bid;
                    if (rate == null) continue;

                    foreach (var stat in stats)
                    {
                        if (string.Equals(PaymentMethodId.Parse(stat.Key).CryptoCode, rateTask.Key.Right,
                            StringComparison.InvariantCultureIgnoreCase))
                        {
                            result += (1m / rate.Value) *  stat.Value;
                        }
                    }
            }
            return result;
        }
```

Kukks: Simplified it. this thing is used in crowdfund app too

Kukks:
Ok @NicolasDorier  & @britttttk  

I added in server-side formatted amount like you wanted @NicolasDorier  and added * next to the required fields and fixed null error on invalid currency input.

Are you all done :P ?

britttttk:
@Kukks Thanks for fixing ^

The Amount field does not actually require user input (as you now denote with the asterisk).

As a result, the PayReq defaults to 0 which does not cause any problems in BTCPay. But what is the purpose of allowing PayReqAmt = 0 if it can't be paid? Only viewed, I can't imagine a use case for this. Some accounting reason perhaps, but it might be better to wait for someone to request this, imo. 

image

Specifically, what purpose does it serve for a user to create a PayReq which appears like this:

image

Kukks:
It's more work to "remove" the feature so let's not. While in creating
state, it is fine to have 0 in it anyway

britttttk:
tACK 

Thanks for your work @Kukks this is a really cool feature :D

BTCPayServer/Controllers/AppsPublicController.cs
NicolasDorier: yes, and it is probably bugged as it update result concurrently.

BTCPayServer/Views/PaymentRequest/MinimalPaymentRequest.cshtml
```diff
@@ -0,0 +1,161 @@
...
+@{
+    var status = Enum.Parse<PaymentRequestData.PaymentRequestStatus>(Model.Status, true);
+}
+<div class="container">
+    <div class="card">
+        <h1 class="card-header">
+            @Model.Title
+
+            <span class="text-muted float-right text-center">
+                @switch (status)
```
NicolasDorier: Can you put this code in the controller instead? Adding logic in the View really suck.

Kukks: (ノಠ ∩ಠ)ノ彡( \o°o)\

NicolasDorier: I will do it ok ok :D

NicolasDorier: But generally, logic in razor is bitch:
* Refactors does like rename does not change impact razor.cshtml
* It crashes at runtime

pavlenex:
This is a great PR with plenty of potential use-cases. It proved to be way better than I originally suggested.

My suggestions are more on the front-end side:

1. Rename Payment Request to Bills (BTCPay Bills sounds great), unless there is additional use-case beyond billing that you'd like to implement.

2. Remove text editor and use simple text field. (Text editor proved to be buggy sometimes. In billing, we don't need to add videos images etc.

3. Center the paymenr request back end (I did simple margin: auto;) and it looks much more focused.

margins

4. Move Store ID field bellow currency, it has more context there

5. Save settings and checkbox are not that good UX. Can we have "Save for Later" and ("Create Bill" or "Publish") without the checkbox. It does not make much sense and people will in most cases forget it, just like we do with the Crowfund app, so many people forget to check that box.

6. Add padding on top of the payment request so that it looks centered on the page and more focused.
capture44

7. Make a payment button, should be **Pay** or Pay Now.

Kukks:
I think the text editor should stay. Too many cool use cases can come out
of it.

britttttk:
> Remove text editor

I think the text editor could be used to send things like timesheets, contract information, videos can show progress of work, etc. with the payment request. I can think of more use cases for keeping it, than removing it. The text editor worked perfectly when I tested it :P

britttttk:
Kukks I like the improvments you have made. I see you fixed a bug in the pay button in your previous 2435187 commit. So I am going to mention that a bug still exists. I'm sorry if you are not ready for re-review. However, it exists like this:

----

**Example 1**:
- check box to allow payee to make the custom payment
- Payer chooses value, (in the Ex1 below it's 1 BTC) and they pay that value (1 BTC) in full.
- When they try to pay the remaining balance (in this case 9 BTC) it does not allow payment when the user clicks the pay button because it opens the Paid modal invoice from the previous payment. 

cantpayt8

----

**Example 2:** (same issue as Ex1)
- User pays 55 BTC of their custom 55 BTC invoice. Ok no problem.
- User pays 1 BTC of their custom 40 BTC invoice. Ok no problem.
- User tries to pay remaining balance of 44 BTC. Opens to a Completed Payment modal invoice. 

openstocompletedpayment

Kukks:
hmm ok, it's because the `/pay` mechanism grabs the last invoice. I will need to add a pay button next to the non-complete invoices and add some additional logic 

Thanks for the test cases

britttttk:
No problem. I like the custom pay button but it also seemed to work fine before when the user just underpaid the one request. If that was true, is it worth it to add extra pay buttons per request? It sounds to me like it could confuse the user. But I’m not sure how it was/will be.

Kukks:
Fixed, did not add any extra pay links. Instead, if there is an active invoice, you will only be able to pay that one.

britttttk:
@Kukks it still happens. It looks so much better though. Also the Tx Id it looks like this on Firefox :/ 

image

Kukks:
fiiiiiiiiixed

britttttk:
Works perfectly, thanks for fixing :D

NicolasDorier:
Nobody bitching about Kukks code anymore? @pavlenex ?
If so, that will be my turn after I reviewed it in my plane back to tokyo :D

Eskyee:
bonne voyage 🛩

pavlenex:
I haven't audited the code, but I tested the front end and UX of this PR. In my tests, I haven't found any bugs and all earlier bugs were fixed.

The only UX problem I've found is that the amount field always has 0, that needs to be deleted prior to entering a normal amount. I think that should be fixed.



I'm also not a fan of the 'actions' in the payment request view, especially on a mobile, it's hard to click. I'm afraid there's little to none to be done there until we improve the UI.

actions
actions2

My suggestion is that in future we make these requests even more customizable and probably allow drag and drop or color customization that will give more customization options for non-coders without having to use CSS.

So afaik, you can review it now @NicolasDorier

BTCPayServer/PaymentRequest/PaymentRequestHub.cs
```diff
@@ -131,14 +132,18 @@ protected override async Task ProcessEvent(object evt, CancellationToken cancell
...
+
                 await InfoUpdated(paymentRequestId);
             }
             else if (evt is PaymentRequestUpdated updated)
             {
-                if (updated.Data.Status != PaymentRequestData.PaymentRequestStatus.Creating)
+                if (updated.Published)
                 {
-                    await InfoUpdated(updated.PaymentRequestId);
+                    await _PaymentRequestService.UpdatePaymentRequestStateIfNeeded(updated.Data);
```
NicolasDorier: why? The `EditPaymentRequest` should have already updated the payment inside `CreateOrUpdatePaymentRequest`.

Kukks: When you save an item and publish( and was in creating state before), the state gets stuck on pending even if the payments were settled

NicolasDorier: How can payments be settled before publish?

Kukks: By setting the payment back to draft mode or a store owner pays it in create mode.

NicolasDorier: it should not be able to be back in draft mode

BTCPayServer/Models/PaymentRequestViewModels/ListPaymentRequestsViewModel.cs
```diff
@@ -80,6 +71,8 @@ public UpdatePaymentRequestViewModel(PaymentRequestData data)
         public string EmbeddedCSS { get; set; }        
         [Display(Name = "Allow payee to create invoices in their own denomination")]
         public bool AllowCustomPaymentAmounts { get; set; }
+
+        public bool Enabled { get; set; }
```
NicolasDorier: Can remove

BTCPayServer/Models/PaymentRequestViewModels/ListPaymentRequestsViewModel.cs
```diff
@@ -124,6 +107,8 @@ public ViewPaymentRequestViewModel(PaymentRequestData data)
             }
         }
 
+        public bool Enabled { get; set; }
```
NicolasDorier: Can remove

NicolasDorier:
Merged!

Issue #396, opened 2018-11-12
Feature Request: Admin users can view and modify other stores


ghtsto:
I would like to have the ability for my admin account to see other stores as well as have permissions to modify their settings. 

The use case would be, as an admin, inviting friends/clients to build their own stores, I'd want to offer assistance in the setup and management of the stores.

Additionally, I think it is important to have visibility into the number of stores and apps that have been created on a server that I run.

bitcoinshirt:
I think this can be dangerous to allow.
We already have a way for users to add admin manually as owners.
There are many third-party hosts , we can't push this issue, it would allow owner to easily replace xpub.

NicolasDorier:
Malicious third party host can already do this.

bitcoinshirt:
I understand, but should we make it easier to people to do it? 
I'm certain it brings more good than it does bad, but maybe we need to add a consent from already existing users for the first time it's added?

I just don't want for third-hosts to have out of the box way to do it, without user consent.

Kukks:
I'm with @bitcoinshirt on this. I don't see why we should make this easier. At least with the current way you need at least some technical skill to get in. There are plenty of people simply running btcpay instances for merchants for free with no technical knowledge and this would enable the asshole minority to wreak havoc. 

Maybe we can come to a compromise on this though and have a toggle in store settings to allow server admins to be able to manage it too.

Eskyee:
We have to secure more, I think and protect our users... not allow admin features to weaken our users privacy... I thought this was a BTCPay server strong point, Full Protection from any Malicious third party host. is our end goal.
This is Very Important... other wise what's the point! of Bitcoin ??

Eskyee:
ps I want everybody to run their own BTCPay Server node...  ;) 
Third party host's are second choice for the end users... but privacy is very important here too.

ghtsto:
> Full Protection from any Malicious third party host. is our end goal.

Wouldn't the solution to this be "run your own instance"? If I am providing the hardware, hosting, installation, maintenance of a btcpayserver instance, I'd argue that I should have full access to all aspects of what happens on my server.

I agree that privacy is very important and one of the pillars of this project was to allow people to get away from centralized providers but that shouldn't mean that an administrator doesn't have visibility into what's happening on _their own server_. Currently, if I give someone access to my server, they have the ability to create an unlimited number of stores and apps and I have absolutely no way of knowing if this is occurring, unless I'm directly monitoring the DB.

[...]

PR #544, opened 2019-01-20
Allow admins to view all stores

sipsorcery:
For #396 allows a user in the Admin role to view all the stores on their server. This PR does not allow an admin to modify the settings on a store they are not the owner of but now at least they'll be able to see them.

btcpay_ownerstores
btcpay_allstores


pavlenex:
@sipsorcery Wait, can the admin read invoices?

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -94,6 +148,28 @@ public async Task<StoreData[]> GetStoresByUserId(string userId)
             }
         }
 
+        //public async Task<StoreData[]> GetAllStores()
```
NicolasDorier: Remove this

BTCPayServer.Tests/StoreRepositoryTests.cs
```diff
@@ -0,0 +1,329 @@
...
+    /// <summary>
+    /// Integration test class for the Store database repsoitory.
+    /// </summary>
+    public class StoreRepositoryTests
+    {
+        /// <summary>
+        /// Attempts to emulate the parts of the BTCPayServer ASP.NET core intialisation needed to test the Store 
+        /// database repository.
+        /// </summary>
+        public class StoreRepositoryTestsStartup
```
NicolasDorier: Why so much hacks instead of using the BTCPay tester we are using already everywhere?

sipsorcery: Well I wouldn't say adding a new test class is a hack. The main reason for a new test class was to isolate the database/EF query logic. The BTCPayServerTester initialises a larger subset of the components used by BTCPayServer. This PR is only concerned with the `Stores` EF repository.

NicolasDorier:
Without the ability to read invoices, it is not that useful.

pavlenex:
I think that visibility of the store on your server run by others is reasonable request, admin should know who uses their server, as let's say the admin be held accountable if he acts as a custodial on behalf of the merchant who's doing something illegal in server host's jurisdiction.

> Should a server admin be able to view/manage settings/invoices on stores belonging to other users?

I think no, absolutely not. There's no real reason for it. If user wants to control multiple stores of his, he can already easily add himself as admin/store owner.

Eskyee:
Should a server admin be able to view/manage settings/invoices on stores belonging to other users?

I strongly disagree too, 

the next thing someone will suggest is.. 
Why not just add KYC to BTCPayServer...  lol 
I wanna know my customers lol "says the admin

britttttk:
* **Admin perspective:**
From the perspective of the server Admin, I think this PR is great. I absolutely think you should be able to see the number of stores created on your server. I also think that server admins should be able to delete all the random stores I have created on their instances for testing (not that many, sorry in advance). 

* **BTCPay Customer perspective:**
From the perspective of a BTCPay customer, I do have a concern. How can a user or customer who pays to some merchant using BTCPay, know that they are paying to a 3rd Party user? Now the admin can see that a user is paying to certain websites or stores. Is it not possible to use the information for doxxing? Is there a way for BTCPay users to know that they are using a 3rd party host vs. an Admin? Can anyone give an opinion on this?

* **BTCPay 3rd Party Merchant perspective:**
    * It's likely that some 3rd party BTCPay merchants would not want their business activity exposed to another party. However, at this point I feel as if the technical burden of setting up BTCPay has been reduced in terms of the necessary technical capability and cost. I believe that many of the 3rd party users are in a BTCPay exploratory phase and can easily improve security and privacy by self hosting when they are ready. 

    * I don't have an issue with letting the Admin see invoices because they are using their own full node anyways, so the transactions can already be watched at the node level. 

    * My biggest concern is the malicious 3rd Party Host. @sipsorcery does not allow modifying of the server settings so the xpub can't be replaced that way. However, what is stopping the malicious 3rd Party Host from creating stores identical to a user stores, adding the user as a new owner to the fake store and deleting their old store? Now you can replace their xpub by replacing their entire store. The user will have no idea that their old store was deleted and they simply see the new store in their account and it will appear the same as the old one. This would depend on the functionality of the store, but the admin can of course view and determine which stores this is possible with. 

    * Run your own server, problem solved. The cost of security and privacy is never free!! I believe that BTCPay should continue to uphold the idea that **running your own node** is the only way to ensure you can reap the full benefits of the Bitcoin Network and the applications build on top of it. 


* **Testing perspective:**
It's easier to test user permissions and store creation when they appear on the Admin account instead of branching off on accounts only the user knows about. This includes when 3rd party users share stores with each other not including the admin.

sipsorcery:
I've updated the PR so that the `Remove` link is only visible to store admins (with the current permissions serve admins can't delete stores any way) and the `Invoices` link is only shown if the user has been explicitly assigned to the store as an Owner or Guest.

btcpay_viewinvoices

@Eskyee  
> Allow admins to view all stores and invoices !! Still don't like this idea... just so you know!

Fair enough but I'm not sure how you could ever stop them.

```
select * from stores;
select * from invoices;
select * from settings;
```

Likewise with the other points about malicious 3rd party hosts and allowing admins to see/modify invoices and settings. It's a completely false sense of security to think there is any protection because a link or menu option isn't shown. It's good that this discussion is being had because the thinking should change.

From:

**Don't allow an admin to do such and such on the BTCPayServer user interface in case they harm or steal from a user** 

To:

 **Emphasis to users in the strongest terms possible that users should only use a 3rd party server IF they trust the server owner NOT to harm or steal from them**

Eskyee:
Running your own btcpay node is the first and best solution, for the end users. 
Using a BTCPay 3rd party server solution is the last choice, I fully agree.
 
Emphasis to users in the strongest terms possible that users should only use a 3rd party server IF they trust the server owner NOT to harm or steal from them

incase of any renegade admins.

@sipsorcery @pavlenex @britttttk @NicolasDorier

NicolasDorier:
Despite the fact that you added a link to the invoices, the admin still have no right to see the invoices. If he click on the link, I would expect the list to be empty.

NicolasDorier:
> Emphasis to users in the strongest terms possible that users should only use a 3rd party server IF they trust the server owner NOT to harm or steal from them

So if we go this way, there is a way simpler PR than this! We can just add an option "Log as user" for the server admin. This mean no complicated refactoring with corner cases like this PR.

NicolasDorier:
I am really torn about this PR.

I think, as @sipsorcery says, that it is giving a false sense of security to just hide something from the UI. An admin can do what he wants by just modifying the database directly. Admins can do this because they have access to the SSH keys to connect to the server.

I think this PR is half seat between not touching what we have today, and just allowing the admin to log as any user. I would largely prefer either not touching what we have today OR going full blown "log as any user".

This PR is invasive code wise for a marginal feature. Going full blown `log as any user` is way less invasive.

Or what about:

* Adding the `log as any user` feature for admin
* Disabled by default
* Can be enabled via a new option parameter?

sipsorcery:
I've wound back a number of the security changes in this PR so that no new permissions are created for a server admin (there weren't ever any new ones added but an additional policy setting was created to avoid using the `role` directly). Consequently the only security change is that an admin now has the option to **view** all the stores on their server as per the original request:

> Additionally, I think it is important to have visibility into the number of stores and apps that have been created on a server that I run.

A new `login as any user` feature would not satisfy the request for an admin to be able to see all the stores on their server.

I'd agree that it's getting to the stage where this feature is taking up more time than it's worth. If the consensus is that it's still too much to allow an admin to view all stores then it would still be worth salvaging the paging portion of this PR so that anyone with large number of stores doesn't get an endless screen full.

britttttk:
IMO the 'Show All Check Box' doesn't make sense for non-Admins, and would be confusing for 3rd Party Users. In this image below we have two 3P Users who share stores with eachother, one as owner, one as guest. It would make sense to me that perhaps guest store access would appear to 3P users when the box is checked but that is not the case. So that to me is unnecessary UI. 
image

Since the Server Admin can always delete any user (which deletes everything related to the user), it doesn't really matter if the Admin can delete the store on this page below, or on the (Server Settings > Users > Remove) page. This just allows the Admin to delete one item for the user (store) instead of their entire account. I'm basically indifferent on this as a feature. 
image

BTCPayServer.Tests/StoreRepositoryTests.cs
NicolasDorier: I understand the goal, I just feel this is a lot of boiler plate, which will need to be maintained.
The only reason to do this is so that tests are faster to execute?

BTCPayServer.Tests/StoreRepositoryTests.cs
```diff
@@ -0,0 +1,329 @@
...
+        public async void CanAddUser()
+        {
+            IWebHost host = new WebHostBuilder()
+                .UseStartup<StoreRepositoryTestsStartup>()
+                .Build();
+
+            using (IServiceScope scope = host.Services.CreateScope())
+            {
+                await scope.ServiceProvider.GetRequiredService<ApplicationDbContext>().Database.EnsureDeletedAsync();
+                await scope.ServiceProvider.GetRequiredService<ApplicationDbContext>().Database.EnsureCreatedAsync();
```
NicolasDorier: this is the kind of boilerplate that the btcpayserver tester take care for you,

BTCPayServer.Tests/StoreRepositoryTests.cs
sipsorcery: I started off using BTCPayServerTester. I added an Sqlite option and then started on dummy config files and settings via environment variables at which point decided it was a very convoluted way to initialise an EF integration test. IMHO 10 lines of code (the ASP.Net IdentityOptions could be set somewhere else) for an isolated approach to testing *Repository classes is a benefit.

Razasangi:
early evening hours, whereas 67.5% of the TV-14 programs were broadcast at 9
p.m. or later. TV-PG programs, however, were equally likely to be broadcast in the
early or later evening hours. Although there were relatively few TV-MA programs
in the sample, these were somewhat more likely to be broadcast in the later evening
hours (64.0%). Similarly, programs that were not rated were a little more likely to
be broadcast in the early evening (65.5%). Although all of the programs with con-
tent designations were slightly more likely to be broadcast in the later hours than
the early hours, only two designations, language and sex, were statistically signifi-
cant (see Table 7). Six out of 10 of the programs with the language and violence
designations—χ2(1, N = 838) = 14.45, p < .001, d = .26, and χ2(1, N = 838) = 7.12,
p < .01, d = .18, respectively—were broadcast at 9 p.m. or later, whereas more than
half of the programs without these designations were broadcast before 9 p.m. Al-
though 6 out of 10 of the programs with the designation for sex were also seen at 9
p.m. or later, because there were so few programs overall with this designation, the
differences were not statistically significant.
292 SIGNORIELLI
TABLE 6
Age-Based Ratings and Time of Broadcast, 1997 Fall–2003 Fall
Not Rated TV-G TV-PG TV-14 TV-M Total
Time R% C% R% C% R% C% R% C% R% C% N %
Early 8.6 65.5 15.9 73.3 61.3 49.6 12.0 32.5 2.2 36.0 416 49.6
Late 4.5 34.5 5.7 26.7 61.4 51.0 24.6 67.5 3.8 64.0 422 50.4
Total (N, %) 55 6.1 90 10.7 514 61.3 154 18.4 25 3.0 838
Note. R% = row percentages. C% = column percentages.
χ2(4, N = 838) = 45.74, p < .001.
TABLE 7
Content Designations and Time of Broadcast, 1997 Fall–2003 Fall
Violencea Sexb Languagec Dialogued Total
Time R% C% R% C% R% C% R% C% N %
Early 14.4 40.8 5.0 39.6 20.9 39.2 17.8 46.0 416 49.6
Late 20.6 59.2 7.6 60.4 32.0 60.8 20.6 54.0 422 50.4
Total (N, %) 147 17.5 53 6.3 222 26.5 161 19.2 838
Note. R% = row percentages. C% = column percentages. aχ2(1, N = 127) = 5.55, p < .05. bχ2(1, N = 53) = 2.27, ns. cχ2(1, N = 222) = 13.19, p < .001. dχ2(1, 161) = 1.08, ns

PR #550, opened 2019-01-22
Add delete Admin message

### LINKED ISSUE #549: Improve delete user message
It would be better if there was a unique message for when the Admin is deleting the Admin. 

Right now the 'delete user' message is the same for all users: "This user will be permanently deleted"

I think we should check:

- If the user being deleted is the admin, display a different message such as:
    * "Are you sure you want to delete this Admin and delete all accounts, users and data associated with the server account?"

I understand admins may want to start over and delete their server account for some reason. However when the Admin deletes himself, I think some users could be confused about what happens next and some hearts could be broken if they are not warned.

britttttk:
Add a warning for removing the admin.
(Server Settings > Users > Remove )

 Closes #549

### DIFF
--- a/BTCPayServer/Controllers/ServerController.cs
+++ b/BTCPayServer/Controllers/ServerController.cs
@@ -369,14 +369,24 @@ namespace BTCPayServer.Controllers
         public async Task<IActionResult> DeleteUser(string userId)
         {
             var user = userId == null ? null : await _UserManager.FindByIdAsync(userId);
+            var roles = await _UserManager.GetRolesAsync(user);
+            var isAdmin = IsAdmin(roles);
             if (user == null)
                 return NotFound();
-            return View("Confirm", new ConfirmModel()
-            {
-                Title = "Delete user " + user.Email,
-                Description = "This user will be permanently deleted",
-                Action = "Delete"
-            });
+            if (isAdmin == true)
+                return View("Confirm", new ConfirmModel()
+                {
+                    Title = "Delete user " + user.Email,
+                    Description = "Are you sure you want to delete this Admin and delete all accounts, users and data associated with the server account?",
+                    Action = "Delete"
+                });
+            else
+                return View("Confirm", new ConfirmModel()
+                {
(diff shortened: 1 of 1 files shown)


britttttk:
Actually it should only appear when there is one admin.

britttttk:
Also something to note: If the user is the only admin, and he un-checks the box to remove himself as admin (Server Settings > Users > Modify > Un-Check Admin Box > Save). 

When this user logs out & logs back in, they are no longer the admin and can't access the Server Settings, although they can still login. The next user to register on the server will become the new Admin and will be able to access SS.

If the User has disabled registration, then they will not be able to re-gain access to their BTCPay as an admin. They will be able to login, but never gain Admin capability back and would have to delete their BTCPay. 

I think there should be some protection on removing the only remaining admin using the checkbox feature as well, for the reasons above. So the user must add a second admin before removing the existing one, using the checkbox feature.

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -369,14 +369,30 @@ public new async Task<IActionResult> User(string userId, UserViewModel viewModel
         public async Task<IActionResult> DeleteUser(string userId)
         {
             var user = userId == null ? null : await _UserManager.FindByIdAsync(userId);
+            var roles = await _UserManager.GetRolesAsync(user);
+            var isAdmin = IsAdmin(roles);
+            var admincount = 0;
             if (user == null)
                 return NotFound();
-            return View("Confirm", new ConfirmModel()
+            foreach (var User in _UserManager.Users)
```
NicolasDorier: This request does not scale. Also it does not count admin counts but the number of user if the current user is admin.

britttttk: Ok thanks for explaining very clearly. I tried but I don't know how to do it. :)

NicolasDorier: unsure you can with the UserManager actually

PR #558, opened 2019-01-25
Add link to no stores error

### LINKED ISSUE #553: Rename: Error: You must have created at least one store
The message: 
> Error: You must have created at least one store 

is clear, but does not provide instructions on where are stores located. 

My suggestion is to replace with:

> Error: You must create a store (with a **link** to /stores/create) 

Suggested message not only gives a warring but also gives clear action on what to do next and eliminates UX confusion.

Kukks:
Closes #553


NicolasDorier:
Argh no, by doing this, I need to review all code calling StatusMessage to know if it is weak to injection.
Better a Temp data property, and use it on the view.

BTCPayServer/Controllers/AppsController.cs
```diff
@@ -73,7 +73,12 @@ public async Task<IActionResult> CreateApp()
             var stores = await _AppsHelper.GetOwnedStores(GetUserId());
             if (stores.Length == 0)
             {
-                StatusMessage = "Error: You must have created at least one store";
+                StatusMessage = new StatusMessageModel()
+                {
+                    Html =
+                        $"Error: You must have created at least one store. <a href='{(Url.Action("CreateStore", "UserStores"))}'>Create store</a>",
+                    Severity = "danger"
```
NicolasDorier: while you are at it, you can use an enum? :D

Kukks: bah!

Kukks:
ready @NicolasDorier  but I know you wanted to rape the code more :P

NicolasDorier:
For now, I would prefer that you make a simple test to know if the first character is a `{` to know if it is json before the try/catch. The reason is that raising exceptions are expensive.

Kukks:
good to go @NicolasDorier

britttttk:
tACK :)

PR #560, opened 2019-01-27
Convert Ledger account list to dropdown and add more accounts to list

Kukks:
image

(List goes to 20 accounts now though)


NicolasDorier:
Cool! This will be useful when we support Trezor.

Kukks:
Yeah this was because a user reported not being able to select their #7 account so I contained it a bit more and extended the loop to 20

PR #566, opened 2019-01-28
Part 1: OpenIddict - Minor Changes & Config prep

Kukks:
* Adds the new openiddict dependencies
* New configuration keys and options along with accompanying cleaner extension methods to load items from config 
* Preps authentication scheme setup to allow more schemes with part 2
* Make `BTCPayClaimsFilter` leaner -- nothing changes in terms of implementation flow
* Expand Error View with additional optional properties


BTCPayServer/Configuration/ConfigurationExtensions.cs
```diff
@@ -57,5 +58,41 @@ public static T GetOrDefault<T>(this IConfiguration configuration, string key, T
                 throw new NotSupportedException("Configuration value does not support time " + typeof(T).Name);
             }
         }
+
```
NicolasDorier: Please move this code in `OpenIdOptions`, and add a `OpenIdOptions.Load(IConfiguration configuration)` which read the config to set the properties.

BTCPayServer/Extensions/OpenIddictServerBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace BTCPayServer
+{
+    public static class OpenIddictServerBuilderExtensions
+    {
+        public static void ConfigureClientRequirements(this OpenIddictServerBuilder builder,
+            IConfiguration configuration)
+        {
+            if(!configuration.GetOpenIdEnforceClients())
```
NicolasDorier: Then use `OpenIdOptions` instead of `IConfiguration` here.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -165,8 +165,8 @@ private static void ConfigureCore(IApplicationBuilder app, IHostingEnvironment e
             {
                 app.UseDeveloperExceptionPage();
             }
-
-            app.UseCors();
+            
+            app.UseCors(builder => builder.AllowAnyOrigin().AllowAnyHeader().AllowAnyMethod());
```
NicolasDorier: I don't think you need this.

BTCPayServer/Configuration/BTCPayServerOptions.cs
```diff
@@ -27,6 +21,51 @@ public class NBXplorerConnectionSetting
...
+            EnforceClients = GetOpenIdEnforceClients(configuration);
+            EnforceEndpoints = GetOpenIdEnforceEndpoints(configuration);
+            EnforceScopes = GetOpenIdEnforceScopes(configuration);
+            EnforceGrantTypes = GetOpenIdEnforceGrantTypes(configuration);
+            return this;
+        }
+
+        public static bool GetOpenIdEnforceClients(IConfiguration configuration)
+        {
+            return configuration.GetValue("openid_enforce_clientId", false);
```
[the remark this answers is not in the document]
Kukks: If clientId is not enforced, you do not need to configure any Clients to use Password Flow. You can use the API through a user/password authorization.
Although not recommended, it's the simplest way to start using all the new features without any of the manual UI configurations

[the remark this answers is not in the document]
NicolasDorier: So it means then you cna't revoke your authorization?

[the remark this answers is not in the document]
Kukks: Revoking a token in Open Id is a little different. You do not revoke the actual tokens but you revoke the refresh tokens( that allow you to extend your session longer without requiring a user to be re-authenticated) 
So in this case, without using a client, you would not be able to request a refresh token which means there is no way to extend your session without broadcasting credentials again or an active authentication eg: browser redirection to btcpay login page)

Kukks:
changes done

BTCPayServer/Security/BTCPayClaimsFilter.cs
```diff
@@ -35,28 +30,34 @@ void IConfigureOptions<MvcOptions>.Configure(MvcOptions options)
...
+
+            var identity = ((ClaimsIdentity)principal.Identity);
+            if (principal.IsInRole(Roles.ServerAdmin))
+            {
+                identity.AddClaim(new Claim(Policies.CanModifyServerSettings.Key, "true"));
+            }
+
+            if (context.RouteData.Values.TryGetValue("storeId", out var storeId))
+            {
+                var userid = _userManager.GetUserId(principal);
```
NicolasDorier: There is no coupling between stores and users, so this will always be null.

Kukks: Actually, if you look at old code, this is identical except that it uses UserManager to locate userId instead of manually looking at Claims on identity principal ( error prone  because you hardcode the claim type for id instead of letting identity handle it from its Startup configuration) 

I'm referring to this in old code:
```
var claim = identity.Claims.FirstOrDefault(c => c.Type == ClaimTypes.NameIdentifier);
//claim.Value is user id
```

BTCPayServer/Configuration/ConfigurationExtensions.cs
```diff
@@ -57,5 +58,22 @@ public static T GetOrDefault<T>(this IConfiguration configuration, string key, T
...
+            return GetDataDir(configuration, networkType);
+        }
+
+        public static string GetDataDir(this IConfiguration configuration, NetworkType networkType)
+        {
+            var defaultSettings = BTCPayDefaultSettings.GetDefaultSettings(networkType);
+            return configuration.GetOrDefault("datadir", defaultSettings.DefaultDataDirectory);
+        }
+
+        public static Uri GetExternalUri(this IConfiguration configuration)
```
NicolasDorier: this does not exists anymore

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -225,16 +226,23 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
...
+                options.AddPolicy(CorsPolicies.All, p => p.AllowAnyHeader().AllowAnyMethod().AllowAnyOrigin());
             });
 
             var rateLimits = new RateLimitService();
             rateLimits.SetZone($"zone={ZoneLimits.Login} rate=5r/min burst=3 nodelay");
             services.AddSingleton(rateLimits);
             return services;
         }
+        
+        private static void AddBtcPayServerAuthenticationSchemes(this IServiceCollection services, IConfiguration configuration)
```
NicolasDorier: `IConfiguration configuration` not needed

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -53,7 +53,7 @@ namespace BTCPayServer.Hosting
 {
     public static class BTCPayServerServices
     {
-        public static IServiceCollection AddBTCPayServer(this IServiceCollection services)
+        public static IServiceCollection AddBTCPayServer(this IServiceCollection services, IConfiguration configuration)
```
NicolasDorier: `IConfiguration configuration` uneeded

BTCPayServer/Security/BTCPayClaimsFilter.cs
NicolasDorier: Or, it should be `GetUserIdAsync` though

BTCPayServer/Configuration/BTCPayServerOptions.cs
```diff
@@ -26,6 +20,31 @@ public class NBXplorerConnectionSetting
         public string CookieFile { get; internal set; }
     }
 
+    public class OpenIdOptions
```
NicolasDorier: Why not simply using `OpenIddictServerOptions`

BTCPayServer/Hosting/BTCPayServerServices.cs
NicolasDorier: You still have not removed configuration here

BTCPayServer/Hosting/Startup.cs
```diff
@@ -64,7 +64,7 @@ public void ConfigureServices(IServiceCollection services)
                 .AddEntityFrameworkStores<ApplicationDbContext>()
                 .AddDefaultTokenProviders();
             services.AddSignalR();
-            services.AddBTCPayServer();
+            services.AddBTCPayServer(Configuration);
```
NicolasDorier: not needed

BTCPayServer/Security/BTCPayClaimsFilter.cs
NicolasDorier: Use the async version

BTCPayServer/Security/BTCPayClaimsFilter.cs
```diff
@@ -35,28 +30,34 @@ void IConfigureOptions<MvcOptions>.Configure(MvcOptions options)
...
+
+                if (!string.IsNullOrEmpty(userid))
                 {
-                    var claim = identity.Claims.FirstOrDefault(c => c.Type == ClaimTypes.NameIdentifier);
-                    if (claim != null)
+                    var store = await _StoreRepository.FindStore((string)storeId, userid);
+                    if (store == null)
+                    {
+                        context.Result = new ChallengeResult();
```
NicolasDorier: This is a breaking change. Now trying to browse to any page containing a storeId will not redirect you to the login page.

Kukks: @NicolasDorier  ping

BTCPayServer/Security/BTCPayClaimsFilter.cs
Kukks: @NicolasDorier  ping

BTCPayServer/Hosting/Startup.cs
Kukks: need this for later

BTCPayServer/Hosting/BTCPayServerServices.cs
Kukks: need it for next parts

BTCPayServer/Security/BTCPayClaimsFilter.cs
NicolasDorier: Sorry I missed it, because the challenge result is interpreted by the authentication scheme. If is is cookie, then this is interpreted as a redirect to login page. You can make a test for it.

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -311,6 +311,37 @@ public void CanCalculateCryptoDue()
...
         }
 
+        [Fact]
+        [Trait("Integration", "Integration")]
+        public async Task GetRedirectedToLoginPathOnChallenge()
+        {
+            using (var tester = ServerTester.Create())
+            {
+                tester.Start();
+                var client = HttpClientFactory.Create();
```
NicolasDorier: Use `tester.HttpClient`

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -313,6 +313,37 @@ public void CanCalculateCryptoDue()
...
+            using (var tester = ServerTester.Create())
+            {
+                tester.Start();
+                var client = tester.PayTester.HttpClient;
+                //Wallets endpoint is protected
+                var response = await client.GetAsync("wallets");
+                var urlPath = response.RequestMessage.RequestUri.ToString()
+                    .Replace(tester.PayTester.ServerUri.ToString(), "");
+                //Cookie Challenge redirects you to login page
+                Assert.StartsWith("Account/Login", urlPath, StringComparison.InvariantCultureIgnoreCase);
```
NicolasDorier: You also need to check if the original path is in the redirect. If you do this, then I will merge.

Kukks: Done! Mergmergemerrge

PR #567, opened 2019-01-28
Part2: Openiddict: Init OpenIddict & Database Migration & Auth Policies

Kukks:


BTCPayServer/Extensions/OpenIddictExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public static class OpenIddictExtensions
+    {
+        public static OpenIddictServerBuilder ConfigureSigningKey(this OpenIddictServerBuilder builder,
+            IConfiguration configuration)
+        {
+            
+            var file = Path.Combine(configuration.GetDataDir(), "rsaparams");
+                    
+            RSACryptoServiceProvider RSA = new RSACryptoServiceProvider(2048);
```
NicolasDorier: Does this generate a key? If yes, we should not do this RSA key generation is super computationally expensive.

BTCPayServer/BTCPayServer.csproj
```diff
@@ -58,6 +58,7 @@
     <PackageReference Include="NicolasDorier.StandardConfiguration" Version="1.0.0.18" />
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="2.1.2" />
     <PackageReference Include="OpenIddict" Version="2.0.0" />
+    <PackageReference Include="OpenIddict" Version="2.0.0" />
```
NicolasDorier: dups

BTCPayServer/Security/Policies.cs
```diff
@@ -8,7 +8,9 @@ namespace BTCPayServer.Security
 {
     public static class Policies
     {
+        
         public const string BitpayAuthentication = "Bitpay.Auth";
+        public const string Dynamic = "dynamic";
```
NicolasDorier: Why dynamic? `OpenIddictAuthentication` would make better sense?

NicolasDorier: nevermind, I understand...

NicolasDorier:
Can you rebase? I added some tests testing some authentication stuff.

BTCPayServer/Views/Account/Login.cshtml
```diff
@@ -59,7 +60,7 @@
                     <h4>Use another service to log in.</h4>
                     <hr />
                     @{
-                        var loginProviders = (await SignInManager.GetExternalAuthenticationSchemesAsync()).ToList();
```
NicolasDorier: maybe we can cleanup all of this, we don't use it.

BTCPayServer/Extensions/OpenIddictExtensions.cs
Kukks: It is a one-time thing and only creates it if the file does not exist.

Kukks:
rebased

BTCPayServer/Views/Account/Login.cshtml
```diff
@@ -1,5 +1,6 @@
 ﻿@using System.Collections.Generic
 @using System.Linq
+@using BTCPayServer.Security
```
NicolasDorier: can probably remove this.

Kukks:
updated @NicolasDorier

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -245,9 +241,40 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
         
         private static void AddBtcPayServerAuthenticationSchemes(this IServiceCollection services, IConfiguration configuration)
         {
-            services.AddAuthentication()
+            JwtSecurityTokenHandler.DefaultInboundClaimTypeMap.Clear();
+            JwtSecurityTokenHandler.DefaultOutboundClaimTypeMap.Clear();
+
+            services.AddAuthentication(options =>
+                {
+                    options.DefaultAuthenticateScheme =  Policies.Dynamic;
```
NicolasDorier: I am not comfortable with this.
What if OpenIddict is not well configured, the hacker inject a `Authorization` header to force `JwtBearerDefaults.AuthenticationScheme`  authentication and gain access to everything?

Why can't we just precise explicitly at controller level, the policy we want to use?

Kukks: I've removed the Dynamic policy

PR #568, opened 2019-01-28
Part3: OpenIddict: Add Flows Event Handlers

Kukks:


Kukks:
@NicolasDorier  rebased this now

BTCPayServer.Tests/AuthenticationTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                user.GrantAccess();
+
+                await GetAccessTokenByPasswordGrant(tester, user);
+            }
+        }
+
+        private static async Task<OpenIdConnectResponse> GetAccessTokenByPasswordGrant(ServerTester tester,
+            TestAccount user)
+        {
+            using (var httpClient = new HttpClient())
```
NicolasDorier: use `tester.CreateClient()` or something like this.

NicolasDorier:
Need rebase

Kukks:
Rebased, I will be adding tests for each handler, at least as much as I can. Are there any specific tests you'd like to see for these @NicolasDorier ?

NicolasDorier:
@Kukks yes, basically think about how we will document this stuff for people.

As far as I understand OpenId does not allow anything more, except if you specifically specify it at the controller level?

It would be nice for this PR to have at least 1 new API using this feature.
I propose the simplest one: `GetStatus` returning the `NBXplorerDashboard` infos.

Kukks:
Yup, just keep in mind that even when this PR is merged in, there is still no way for users to create clients in order to make use of open id authorizations( that's in part 4)

NicolasDorier:
@Kukks that's fine, if that's not implemented in UI yet, it is not needed for the tests.

NicolasDorier:
@Kukks ha, I did not checked by can you make the Password flow disabled by default (then activable in the user's settings later?)

NicolasDorier:
I will take a look, I don't like changing the default. One fuckup and people's server settings are open on the wide.

Kukks:
I have a few more flows to test against and then I will improve on that dynamic policy default and make it less intrusive

Kukks:
It now has tests for all flows (Password, Client Credentials, Implicit, Code, Refresh). When creating an API client, you need to specify which flows are allowed( so the password flow needs to be specifically chosen). I've had to re-add the dynamic policy because I think I found a .net core bug where the authorization filter priority is not consistent depending on where the attribute is defined in controllers but I have made it more restrictive( JWT bearer is only allowed when URL  starts with `/api` and has the `Authorization Bearer xxx` header)

NicolasDorier:
I think you are mixing the concept of Policies with the concept of AuthenticationScheme. Will try to play around.

NicolasDorier:
Ok what about this: You change the default authenticationScheme as one scheme that always fail.
That basically force people to specify clearly which scheme is needed everywhere, my last commit does that.

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -230,9 +230,7 @@ public void CanCalculateCryptoDue()
 
             entity.Payments.Add(new PaymentEntity()
             {
-                Output = new TxOut(Money.Coins(0.5m), new Key()),
```
NicolasDorier: Can you stop changing that back and forth? Before it was on one line, then you formatted to be on several line, and now you put back on one line.

BTCPayServer.Tests/AuthenticationTests.cs
```diff
@@ -0,0 +1,351 @@
...
+
+        [Fact]
+        [Trait("Integration", "Integration")]
+        public void CanGetOpenIdConfiguration()
+        {
+            using (var tester = ServerTester.Create())
+            {
+                tester.Start();
+                var url = new Uri(tester.PayTester.ServerUri, "/.well-known/openid-configuration");
+                using (var wc = new WebClient())
```
NicolasDorier: Use `tester.PayTester.HttpClient`

NicolasDorier:
everything fail @kukks

NicolasDorier:
Rebase,it was maybe my fault as I broke the build at one point.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -142,6 +163,10 @@ private void ConfigureOpenIddict(IServiceCollection services)
                 })
                 .AddServer(options =>
                 {
+                    if (_Env.IsDevelopment())
+                    {
+                        options.DisableHttpsRequirement();
```
NicolasDorier: mmmh what is requiring this? Because it will not work on Tor.

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -253,20 +251,15 @@ public void CanCalculateCryptoDue()
...
-                Accounted = true,
-                NetworkFee = 0.1m
+                Output = new TxOut(Money.Coins(0.6m), new Key()), Accounted = true, NetworkFee = 0.1m
             });
 
             accounting = paymentMethod.Calculate();
             Assert.Equal(Money.Zero, accounting.Due);
             Assert.Equal(Money.Coins(1.3m), accounting.TotalDue);
 
-            entity.Payments.Add(new PaymentEntity()
```
NicolasDorier: please remove those changes.

BTCPayServer/BTCPayServer.csproj
```diff
@@ -1,6 +1,6 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
-  <Import Project="../Build/Version.csproj" Condition="Exists('../Build/Version.csproj')" />
-  <Import Project="../Build/Common.csproj" />
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <Import Project="../Version.csproj" Condition="Exists('../Version.csproj')" />
```
NicolasDorier: revert this

NicolasDorier:
tests still break, just rebase pelase

Kukks:
Should be ok now

NicolasDorier:
OK looking at it... again :D

NicolasDorier:
Seems almost good!

* `AuthorizationCode_RefreshTokenGrantTypeEventHandler` should be splitted into two classes (or alternatively, put all the handlers into 1 class)
* Need to test `Logout`
* `AuthorizationEventHandler`, ` OpenIddictConstants.Errors.ServerError` is for bugs, should be InvalidGrant or something else.

NicolasDorier:
tests not passing

Kukks:
tests pass now :)

PR #584, opened 2019-02-04
WIP: "Auto trade" feature

joemphilips:
Hi.

My company wants to start using BTCPayServer for receiving BTC from our customers.
But for some mysterious accounting reason, they want to treat incoming payments are payed as fiat.

So I decided to implement a feature to sell the pre-existing balance in the exchange automatically when our server received a payment. 
In this way, we still have to buy BTC beforehand but it can be treated as a necessary expense. so it can (hopefully) avoid the regulation.

I'm planning to add a section to set the "Auto Exchange" settings under third party payment methods.
btcpay_server

I've just started this PR, and it is far from finished. But I want to have a feedback from community beforehand if possible.

So far, the only exchange I'm planning to support is https://bitbank.cc/ because our company resides in Japan. But I want to make it extendable to support any exchange.
NOTE: My current plan is to support only one exchange per store. Since it might be too complicated otherwise.

TODO:
* [ ] finish writing controller and tests. 
* [ ] finish view.
* [ ] check if API token is not for withdrawing (only trading)
* [ ] send BTC automatically to the exchange when the balance in the exchange has become too low.
* [ ] add test case for the reorg.

Rationale ... Users should be able to pay in any currency they want. At the same time, merchant should be able to receive in any currency they want( including fiat)

### DIFF
--- a/BTCPayServer.Tests/AutoTradeTests.cs
+++ b/BTCPayServer.Tests/AutoTradeTests.cs
@@ -1,0 +1,59 @@
+using BTCPayServer.Controllers;
+using BTCPayServer.Models.StoreViewModels;
+using BTCPayServer.Payments.AutoTrade;
+using BTCPayServer.Tests.Logging;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace BTCPayServer.Tests
+{
+    public class AutoTradeTests
+    {
+        public AutoTradeTests(ITestOutputHelper helper)
+        {
+            Logs.Tester = new XUnitLog(helper) {Name = "Tests"};
+            Logs.LogProvider = new XUnitLogProvider(helper);
+        }
+
+        [Fact]
+        [Trait("Integration", "Integration")]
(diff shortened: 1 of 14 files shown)


joemphilips:
@Kukks cool! May I have a pointer to the document or the code? @rockstardev

Kukks:
It's still under closed development till we polish it out. Reach out to me on our chat, I can send you some demo videos

JayOkocha:
The same here^^. I have some stores who want exactly the same and only keep the tips in Bitcoin, or whatever is left after they get the fiat filled. But mostly to escape price fluctuation, make taxation easier and and still accumalate a little BTC :). But where are the exchanges with LN support?

rockstardev:
I just had call with Joe and explained how we will be attacking this problem through Service like Jeff did for LibrePatron: https://twitter.com/BtcpayServer/status/1087934018778288128

Anyone looking to contribute to the effort, DM me on Mattermost (http://chat.btcpayserver.org) so that we don't uncle fork our effort-chain and bring upon us wrath of philosopher child god Vitalik.

Issue #589, opened 2019-02-06
Fix Crowdfund missing user input fields


britttttk:
Improve the user input fields for Crowdfund:

1. Add asterisks for required user input.
2. Fix message for primary currency (remove extra "The")
![cf_createfields](https://user-images.githubusercontent.com/39231115/52370245-a30e2500-2a0f-11e9-9ec1-d17241042e21.PNG)

rockstardev:
Thanks Brit, I got it tonight

PR #593, opened 2019-02-08
Custom validation message for Crowdfund form primary currency

BenSanex:
Second half of #589

### DIFF
--- a/BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
@@ -39,9 +39,9 @@ namespace BTCPayServer.Models.AppViewModels
         public DateTime? StartDate { get; set; }
         public DateTime? EndDate { get; set; }
 
-        [Required(ErrorMessage="The primary currency is required")]
+        [Required]
         [MaxLength(5)]
-        [Display(Name = "The primary currency used for targets and stats. (e.g. BTC, LTC, USD, etc.)")]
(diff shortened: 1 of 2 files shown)


BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
```diff
@@ -39,7 +39,7 @@ public class UpdateCrowdfundViewModel
         public DateTime? StartDate { get; set; }
         public DateTime? EndDate { get; set; }
 
-        [Required]
+        [Required(ErrorMessage="The primary currency is required")]
         [MaxLength(5)]
         [Display(Name = "The primary currency used for targets and stats. (e.g. BTC, LTC, USD, etc.)")]
```
britttttk: It follows the style of the other missing field messages on this page, if you just remove `The` from the Display Name. Also need asterisk here :)

BenSanex: The asterisk should get taken care of when rockstars PR gets merged, didn't want to accidentally cause a merge conflict.
IMO repeating `used for targets and stats. (e.g. BTC, LTC, USD, etc.)` is redundant but can change it if that's preferred.

britttttk: There are three required fields: Title, Description, Primary Currency. The title field asterisk already exists. Rockstardev has added one for the description field, so there is one missing for the third field. (:

BenSanex: Alright I got it this time :D

rockstardev:
Whoho

PR #594, opened 2019-02-10
Feature/britk admin fixes

rockstardev:
This resolves issues #549 #550

### DIFF
--- a/BTCPayServer/Controllers/ServerController.cs
+++ b/BTCPayServer/Controllers/ServerController.cs
@@ -383,12 +383,29 @@ namespace BTCPayServer.Controllers
             var user = userId == null ? null : await _UserManager.FindByIdAsync(userId);
             if (user == null)
                 return NotFound();
-            return View("Confirm", new ConfirmModel()
+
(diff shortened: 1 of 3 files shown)


britttttk:
Thanks for the PR @rockstardev it works great. Knowledge drop as Mr. @Eskyee says :D

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -352,22 +352,27 @@ public new async Task<IActionResult> User(string userId, UserViewModel viewModel
...
-            var isAdmin = IsAdmin(roles);
-            bool updated = false;
 
-            if (isAdmin != viewModel.IsAdmin)
+            viewModel.StatusMessage = "";
+
+            var admins = await _UserManager.GetUsersInRoleAsync(Roles.ServerAdmin);
+            if (!viewModel.IsAdmin && admins.Count == 1)
+            {
+                viewModel.StatusMessage = "This is the last Admin, so IsAdmin role can't be removed";
```
britttttk: "This is the only Admin, so their role can't be removed until another Admin is added."

I think it's better to remove IsAdmin from the message and suggest a solution to the user.

PR #607, opened 2019-02-22
fix invoice fetcher in crowdfund

Kukks:
Fixes bug where invoices were not being counted at all.
Also fixes an issue where you could no longer view invoices of a specific app.  Had to introduce a new invoice filter `appid`. 

Not the most efficient code but couldn't think of a better way for now.


NicolasDorier:
we can't query on `appId` this is not a column and mean we must bring in memory all the invoices, which is not acceptable.

Kukks:
Yeah it is quite bad, I'm guessing there is no other way around it right?

NicolasDorier:
no way to work around.

NicolasDorier:
Normally fixed by https://github.com/btcpayserver/btcpayserver/commit/4016ded58466148119a07120543ae460abeac074

PR #629, opened 2019-02-28
Update text in Services to reflect LN in general not just LND

### LINKED ISSUE #573: Modify" You can get access here to LND (gRPC, Rest)..." in Services
As mentioned in #545 the text under Services > Crypto Settings should be modified to reflect current state of things.

Right now it says:

>  You can get access here to LND (gRPC, Rest) services exposed by your server 

screen shot 2019-01-29 at 08 29 14

It looks a bit strange when you have c-lightning for example.

Any ideas how we can make this a bit more up to date?

pavlenex:
Making it more generic to suit c-lightning not just LND.

Perhaps the headline "Crypto Services" should be changed to Lightning Services since we break down different plugins into categories anyway.

Close #573

### DIFF
--- a/BTCPayServer/Views/Server/Services.cshtml
+++ b/BTCPayServer/Views/Server/Services.cshtml
@@ -17,7 +17,7 @@
     <div class="col-md-8">
         <h4>Crypto services</h4>
         <div class="form-group">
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Closing this one, the text has been changed already and there is not only lightning network service in the future

Issue #628, opened 2019-02-28
Pay Request "No store" error consistent with warring in apps


pavlenex:
We should make the error message consistent with the one in apps as it gives the user a link and exact action on what to do when no store is created.

Add a link to store /stores.

![2019-02-28_22-06-53](https://user-images.githubusercontent.com/36959754/53598640-8e9ff100-3ba5-11e9-9587-336517fea749.png)
![2019-02-28_22-06-54](https://user-images.githubusercontent.com/36959754/53598641-8f388780-3ba5-11e9-96f3-c89a342e9a5b.PNG)

pavlenex:
Fixed grammar for tor the Apps [here](https://github.com/btcpayserver/btcpayserver/pull/630/commits/40efdf01099ab21dddfa3b7e6506df59fbea3e9f).

Kukks:
plz close @pavlenex ? :D

Kukks:
or not, my bad

PR #630, opened 2019-02-28
Fix typo in AppsController.cs

pavlenex:
Fixing grammar a bit in the AppsController.cs

capture


britttttk:
I'm in favor of using, "You need to create at least one store" for apps. As described in #628.

PR #646, opened 2019-03-05
fix no store error message for payment request 

### LINKED ISSUE #628: Pay Request "No store" error consistent with warring in apps

Kukks:
closes #628

### DIFF
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -101,7 +101,7 @@ namespace BTCPayServer.Controllers
                         StatusMessage = new StatusMessageModel()
                         {
                             Html =
-                                $"Error: You must have created at least one store. <a href='{Url.Action("CreateStore", "UserStores")}'>Create store</a>",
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -98,7 +98,12 @@ public async Task<IActionResult> EditPaymentRequest(string id, string statusMess
                 return RedirectToAction("GetPaymentRequests",
                     new
                     {
-                        StatusMessage = "Error: You need to create at least one store before creating a payment request"
+                        StatusMessage = new StatusMessageModel()
```
pavlenex: Can you please make the error message itself consistent with the one [here](https://github.com/btcpayserver/btcpayserver/pull/630/commits/40efdf01099ab21dddfa3b7e6506df59fbea3e9f)? 

Consistency in error message would help us with troubleshooting and documenting stuff.

Kukks: done

PR #647, opened 2019-03-05
Datetime picker  and small edit UI changes

### LINKED ISSUE #625: Payment request - No expiration
screenshot 2019-02-28 at 09 12 47

Wouldnt it be better, when nothing is selected, to put pre-selected text there that says: Invoice currently does not have an expiration date or something.
Makes for a better UX.

X on the right makes no sense, as nothing is selected. Makes sense when you have something selected.

Kukks:
* remove currency placeholder in payment requests
* do not allow negative amounts for crowdfund and payment requests
* Clear button only appears when a value is set. If no value is set, display a placeholder indicating it. Closes #625


pavlenex:
Great way to display a message in the UI and it's not intrusive. 
Tested in a dev environment, and everything appears to work.

PR #648, opened 2019-03-05
Clone Payment Requests

### LINKED ISSUE #615: Payment Request - Clone template
Users should be able to easily clone earlier made template. People using pay requests to invoice client won't change the design often, so allowing cloning will speed up the  creation process.

Kukks:
closes #615

### DIFF
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -309,5 +309,21 @@ namespace BTCPayServer.Controllers
         {
             return _UserManager.GetUserId(User);
         }
+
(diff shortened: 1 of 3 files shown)


pavlenex:
Cloning works well. I like how it's handled from the UI perspective.
However, currently, when you clone an invoice, it immediately goes into a saved state (pending).

I know that in that state, the invoice can be modified easily, but when user clones a request he will most likely have to do a tiny modification. Rarely will there be 100% identical pay request? 

My suggestion is that the payment request does not become saved as soon as it's cloned. The user should do modifications, then save.

Kukks:
Changes done

PR #653, opened 2019-03-06
Allow sounds and animation colors in crowdfund to be  configured

### LINKED ISSUE #652: Crowdfund: Configurable Sounds & animation colors
- Allow sounds to be configurable
- Allow animated background colors to be configurable

Kukks:
closes #652

### DIFF
--- a/BTCPayServer/Controllers/AppsController.Crowdfund.cs
+++ b/BTCPayServer/Controllers/AppsController.Crowdfund.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Text;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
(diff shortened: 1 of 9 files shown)


pavlenex:
Great change. Custom sounds can be a great use-case in the music industry. I already have an idea.

Is there a way we can make adding of the format a bit clearer?

We can add a tooltip that directs to docs or maybe just standardize the format and somehow explain what kind of file/links required.

sound


Lots of DJ's use SoundCloud, etc but if that does not work, we can find a workaround and explain it in a tutorial. With the voting feature enabled, their fans can even vote for songs. Though we'd have to then have sounds in perks.

Kukks:
it has to be direct music files that are supported in web browsers. With docs links , we can add it with next release once we write it up :)

pavlenex:
I see. I agree, we'd have to explain that through documentation somehow with examples.

I've tested this and everything works as expected when you add the appropriate direct download file link.

Kukks:
(protip:  you can use `red` `green`, etc in the colors)

PR #654, opened 2019-03-06
Add policy to discourage search engines + build robots.txt dynamcally

### LINKED ISSUE #390: Feature request- Robots Meta Tag and/or robots.txt
For security purposes, it would seem like a good idea to give the user the option to not have the web interface indexed by search engines, either through meta tagging or a robots.txt. Since the recommended install is one-click via Docker, it's not easy to have the user add a robots.txt to his btcpay subdomain himself manually.

Kukks:
closes #390


pavlenex:
Okay, I'd like to post a little feedback about this.

The way it currently works is that it creates yourserrver./robots.txt and it disallows robots entirely. This, however, means that search engines can still index your page, which I think is way more important.

To understand what I am saying better:

* Noindex: tells search engines not to include your page(s) in search results.
* Disallow: tells them not to crawl your page(s).
* Nofollow: tells them not to follow the links on your page

So we're telling Google, don't crawl my page. But it may (and most likely will index it and show it in a very ugly way in search results)

> Noindex (page) + Disallow: Disallow can’t be combined with noindex on the page, because the page is blocked and therefore search engines won’t crawl it to know that they’re not supposed to leave the page out of the index.

Google Webmasters Guidelines [recommend ](https://support.google.com/webmasters/answer/93710?hl=en)using noindex for not showing page in search engine.

> Important! For the noindex directive to be effective, the page must not be blocked by a robots.txt file. If the page is blocked by a robots.txt file, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.

That said, what's the goal here @JeffVandrewJr? Block bots, or prevent search engines from indexing our pages?

If we want to keep our private pages out of the search engines, I'd recommend, that we use:

> To prevent most search engine web crawlers from indexing a page on your site, place the following meta tag into the <head> section of your page: `meta name="robots" content="noindex"`

In past I was able to replicate this behavior of search engines, when they showed my pages blocked by bots and they looked even uglier.

JeffVandrewJr:
My thought process was to keep it out of search engines altogether, since it's an internal-use page that you wouldn't necessarily want customers to come across when googling your business.

pavlenex:
@JeffVandrewJr  Exactly what I though, so I then suggest we use meta name="robots" content="noindex" sitewide instead of blocking bots.

NicolasDorier:
I am not super fan to add a dependency on external lib for something so simple.
We can just add a route in the Home controller instead which return the file.

Kukks:
@NicolasDorier  Agreed

Kukks:
@pavlenex  & @NicolasDorier  updated

No dependencies, robots.txt generated dynamically anyway + meta tag in header dynamically as requested

Kukks:
discussed with @pavlenex  a bit more and removed robots.txt generation entirely and left noindex meta only

BTCPayServer/Views/Shared/_Layout.cshtml
```diff
@@ -15,6 +15,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="description" content="">
     <meta name="author" content="">
+    <cache expires-after="@TimeSpan.FromMinutes(5)">
```
NicolasDorier: I see coming the issue opening with people saying "I changed the settings but nothing changed on my page" because they tested before the 5min expiration.

Can you do the same hack we did for the ThemeManager to prevent querying the DB each time?

Kukks: yup, had no idea of that mechanism. Done

BTCPayServer/Models/RobotsMetaViewModel.cs
```diff
@@ -0,0 +1,7 @@
+namespace BTCPayServer.Models
+{
+    public class RobotsMetaViewModel
```
NicolasDorier: don't need it

Kukks: poof

Kukks: poof

rockstardev:
Why are tests failing? ;(

NicolasDorier:
was my fault, fixed it since

PR #657, opened 2019-03-07
Document LND supported wallets in services

NicolasDorier:
This add in Services / LND Rest or gRPC some links to supported wallets:

image

image

I am certainly missing some wallets, so I keep this PR open until I get more complete.


pavlenex:
I like the idea. I have two thoughts:

1. Should we maybe point the links instead of to websites to our documentation for how to configure each wallet? I'm a bit worried user may get lost once they get sent to the home page of a wallet. Or do we leave this on wallets to figure out?

2. Perhaps, `compatible wallets` should be below the QR code. Right now, it looks okay, but what will happen when we have 6, or 10 wallets, the QR code which is the most important piece of this page will o below the fold.

NicolasDorier:
With Zeus

image

NicolasDorier:
If the worry is about what happen when there is 10 wallets, let's just change it when there is 10 wallets.
Right now it looks good right?

I can change links, but I think it is not really our business to document other people project, they should have a good landing page.

pavlenex:
Sure, let's do it when there are more wallets. No need to change now and about docs I agree, anyways we will see if people have questions or not, if they do get trapped we'll fix it later.

PR #667, opened 2019-03-09
Add Issue Templates

pavlenex:
This PR adds a new folder ./github that has two files:
- feature_request.md
- bug_report.md

Those two files will provide an issue template to people when they're opening an issue. Templates are great way to get as much information as possible without spending too much time going back and fort just to get some basic information. 

Also templates eliminate easy questions that can easily be Googled since whoever is filing an issue will go down the path which takes less effort. 

I got this idea from[ RTL project](https://github.com/ShahanaFarooqui/RTL/issues/new/choose).

This is how user will be greeted when he clicks new issue

Screen Shot 2019-03-09 at 23 25 07

These templates are just a suggestion, feedback and ideas are welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,34 @@
+---
+name: Report a problem
+about: File a technical problem or report a bug
+---
+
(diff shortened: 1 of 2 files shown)


rockstardev:
I like it

PR #668, opened 2019-03-10
Improve scanability of payment QR codes

### LINKED ISSUE #609: Improve the scannability of QR codes in BOLT11 invoices
Several users reported that lately BOLT11 invoices take a while to be scanned.
It was proposed by @Kukks that we reduce the size of the icons in the middle or remove them entirely:

> They have too much data on them . We should remove the circle on them so it has around an extra 10% success rate

skddc:
Doubles the resolution of the payment QR codes, so they are consistently readable when shown on high-res displays.

We have to add `!important` to the CSS rules for now, because [node-qrcode doesn't have an option to change the rendered size](https://github.com/soldair/node-qrcode/issues/140) separately from the canvas/image size itself.

fixes #609

### DIFF
--- a/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
@@ -207,7 +207,7 @@
         <div class="payment__scan">
             <img v-bind:src="srvModel.cryptoImage" class="qr_currency_icon"
                  v-if="scanDisplayQr" />
-            <qrcode v-bind:value="scanDisplayQr" :options="{ width: 256, margin: 0, color: {dark:'#000', light:'#f5f5f7'} }"
+            <qrcode v-bind:value="scanDisplayQr" :options="{ width: 512, margin: 0, color: {dark:'#000', light:'#f5f5f7'} }"
                     v-if="scanDisplayQr">
             </qrcode>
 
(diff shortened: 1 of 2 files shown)


Kukks:
I think @rockstardev  did some work on this and replaced the qr code generator with a more efficient one. That said, I think your code improves more on it.

skddc:
The build timed out, but I don't think my changes break the tests. Maybe someone can restart, and merge if green?

Kukks:
can you rebase on latest master?

rockstardev:
Can anybody else verify this issue? I would prefer clean new issue and someone else reproducing the problem because we got this in middle of QR code replacing.

@skddc can you share the URL / picture of invoice so that others can try scanning code with their own Android devices?

skddc:
I don't see any conflicts with the master branch, so I don't see how my code changed something that was also changed in master. Could you please link to the commit or PR that changed the QR code rendering in the meantime?

rockstardev:
My worry was that you tested with old version... then pulled new version and changed it without testing. What you describe is kind of problem that happened before this change. That's why I'm asking if someone else can reproduce the problem -> because everyone that tried it out so far said it works better.

For canvas / svg - I agree... hate screen density problems.

rockstardev:
Also, can you do me a favor @skddc - can you try margin:1 in your environment:

`:options="{ width: 256, margin: 1, color: {dark:'#000', light:'#f5f5f7'} }"`

Seems like there is subtle bug with library - it cuts of few pixels on the right side of QR code when margin is 0.

rockstardev:
We'll be switching to SVG: #687

skddc:
:+1:

Issue #578, opened 2019-01-29
Server error when creating invoice with lnd and mysql [ Data too long for column 'Address' at row 2 ]


gradientskier:
How to replicate:

* Btcpay commit a815fad3f1d322d2fac6c3364d656b6db1685cda (HEAD, tag: v1.0.3.50) 
* `lnd version 0.5.1-beta commit=v0.5.1-beta`

Btcpay configuration (I'm using mysql):

```
testnet=1
port=8082 
mysql=Uid=root;Pwd=example;Server=localhost;Database=btcpay;
externalurl=https://dev.something.it/btcpay
rootpath=/btcpay
```

Lnd connection string:

`type=lnd-rest;server=https://localhost:8085/;macaroonfilepath=/home/gradientskier/.lnd/data/chain/bitcoin/testnet/admin.macaroon;allowinsecure=true`

Steps to reproduce:

0. Start with a new btcpay manual installation and no btcpay database
1. Register a new user
2. Create a new store
3. Configure derivation scheme for the store
4. Configure lightning node for the store
5. Create invoice

Error:

```
crit: PayServer:      Unhandled exception in BTCPayMiddleware
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> MySql.Data.MySqlClient.MySqlException: Data too long for column 'Address' at row 2 ---> MySql.Data.MySqlClient.MySqlException: Data too long for column 'Address' at row 2
   at MySqlConnector.Core.ServerSession.TryAsyncContinuation(Task`1 task) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ServerSession.cs:line 1245
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs:line 42
   --- End of inner exception stack trace ---
[...]

PR #703, opened 2019-03-21
Increase Address max length to 512

sipsorcery:
Should solve #578 when deploying new BtcPay server instances with MySQL.

### DIFF
--- a/BTCPayServer/Migrations/20171006013443_AddressMapping.cs
+++ b/BTCPayServer/Migrations/20171006013443_AddressMapping.cs
@@ -12,7 +12,7 @@ namespace BTCPayServer.Migrations
                 name: "AddressInvoices",
                 columns: table => new
                 {
-                    Address = table.Column<string>(nullable: false),
+                    Address = table.Column<string>(nullable: false, maxLength: 512),
                     InvoiceDataId = table.Column<string>(nullable: true)
                 },
                 constraints: table =>


Kukks:
You need to generate a new migration for this instead of editing an existing one. You need to annotate the c# model's property with this maxlength and generate the migration from there. 

With this, existing deployments won't notice the change and the new deployments wont get it because this only affects users who are on a version before this migration was created.

NicolasDorier:
mmh

NicolasDorier:
Actually I think this is fine.

NicolasDorier:
So there is two way we can fix this:

1. Make sure we never put too much info in the Addresses column.
2. Changing the migration script (it won't break existing deployment, it will not fix current deployments, but we don't really care as they were not working anyway)

NicolasDorier:
Also I think the problem is in the `HistoricalAddressInvoiceData`. `AddressInvoiceData` is now storing only the preimage of the payment.

Kukks:
New deployments would use the snapshot migration as far as I remember and not the incremental ones. Can't we use data annotations like `[MaxLength(500)] ` and genertae a migration then

NicolasDorier:
I like this one here because things working before will still work tomorrow. Your other solution might crash sqlite deployments. 

Also maybe it is better idea to either trunk the BOLT11 OR just use the payment hash

Kukks:
closed my other pr

Kukks:
@sipsorcery  Is it ok if I take over your PR? :)

sipsorcery:
@Kukks sure of course. I'm a bit out of my depth with dotnet core ef any way.

Kukks:
I think we should merge this in actually, @NicolasDorier  was right - this should work just fine. @ketominer currently has to do a bash script on nodl to fix this issue. I think he does a 1k char limit but 512 should be enough anyway

NicolasDorier:
Will be part of https://github.com/btcpayserver/btcpayserver/issues/1224

Fixing that is actually became way harder.

Issue #669, opened 2019-03-10
Blank page when I try product buttons 


danbadjar:
**Describe the problem/bug**
Payment buttons doesn't work in my App (Point of sale). A blank page is generated instead.

**Your environment**
v1.0.3.81

Hello I'm trying a bran new BTPayserver instance on my Linux Machine. Payment buttons generate a blank page.
In BTCPayserver logs I see critical errors concerning MySQL.
Can anybody help me? Thanks in advice

```
crit: PayServer:      Unhandled exception in BTCPayMiddleware
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> MySql.Data.MySqlClient.MySqlException: Data too long for column 'Address' at row 1 ---> MySql.Data.MySqlClient.MySqlException: Data too long for column 'Address' at row 1
   at MySqlConnector.Core.ServerSession.TryAsyncContinuation(Task`1 task) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ServerSession.cs:line 1245
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs:line 42
   --- End of inner exception stack trace ---
   at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet(ResultSet resultSet) in C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs:line 74
   at MySql.Data.MySqlClient.MySqlDataReader.ReadFirstResultSetAsync(IOBehavior ioBehavior) in C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs:line 299
[...]

PR #705, opened 2019-03-22
Add migration for maxLength to address columns 

Kukks:
For #669 #703 and #578

I tried to add both data annotations and to the fluent migration builder but migrations are not generated there so I manually made the migration script. 

The reasoning behind the bug is that any indexed key is given a a specific max length instead of MAX


NicolasDorier:
problem is, column migration might not be supported by Sqlite, so such migration would crash some people.

Kukks:
Should I check for the provider to not be sqlite?

NicolasDorier:
I frankly don't think it is worth the trouble to add a migration.

PR #707, opened 2019-03-22
set username on email change

### LINKED ISSUE #673: username stays the same after changing email
After I update email, username stays the same. And this is means to login I should use old email. 

**Your environment**
* `53cecc8c8a41033a88594b14521d3699f1a42d32`

**To Reproduce**
1. Go to `/Manage/Index`
2. Change email and confirm
3. Follow link in the email
4. New email used for password reset and notifications, but to actually login I still need to use the old one, as it actually not a email, but username on the login page

**Expected behavior**
username should be changed to the new email too

**Actual behavior**
username stays the same, and this is means I still need to use old details to login

Kukks:
closes #673

### DIFF
--- a/BTCPayServer/Controllers/ManageController.cs
+++ b/BTCPayServer/Controllers/ManageController.cs
@@ -110,10 +110,10 @@ namespace BTCPayServer.Controllers
             {
                 var setEmailResult = await _userManager.SetEmailAsync(user, model.Email);
                 if (!setEmailResult.Succeeded)
-                {
-                    await _userManager.SetUserNameAsync(user, model.Username);
+                {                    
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/ManageController.cs
```diff
@@ -111,6 +111,7 @@ public async Task<IActionResult> Index(IndexViewModel model)
                 var setEmailResult = await _userManager.SetEmailAsync(user, model.Email);
                 if (!setEmailResult.Succeeded)
                 {
+                    await _userManager.SetUserNameAsync(user, model.Username);
```
NicolasDorier: I don't think you want to put this line here. Probably after, if it succeeded

PR #708, opened 2019-03-22
Abstracted cloud storage - Amazon/Google/Azure/Local

### LINKED ISSUE #239: Add unified file storage solution
Add in a way to store files within the system

@NicolasDorier  suggested to store them in the db directly but I'm a bit wary of that due to issues that could come with it (more details here: https://softwareengineering.stackexchange.com/questions/150669/is-it-a-bad-practice-to-store-large-files-10-mb-in-a-database)

I propose we use something like this: https://github.com/geeklearningio/gl-dotnet-storage

Kukks:
Currently restricted only to Admins but this can be extended with extra settings to allow non-admin stores or apps to use it to upload css or image files( with a file size restriction)

I also plan to use this to handle the FastSync direct integration into Btcpay

Another use-case is https://github.com/btcpayserver/btcpayserver-docker/issues/116

closes #239

image
image
image
image
image
image


NicolasDorier:
I am worried that people get confused about it and think that everything get saved into the storage, when it is not the case.

How about, instead of having a menu specifically for it, just setting it up as a "Service"?

Also, I am worried that people use file storage on the docker deploy, because if they do, they will lose the files when the container restarts. So if file storage, the root of the storage should be in the datadir of the chain.

Kukks:
> Also, I am worried that people use file storage on the docker deploy, because if they do, they will lose the files when the container restarts. So if file storage, the root of the storage should be in the datadir of the chain.

Hmm, we need them to be in the wwwroot to be publicly available. Maybe another volume? Or I think there is a FileProvider trick I can do to map the folder to the web if I remember correctly

NicolasDorier:
not the wwwroot,  the wwwroot also lose the data

Kukks:
@NicolasDorier  using data dir now

Kukks:
Moved to services section and added a `/storage/{fileId}` endpoint

NicolasDorier:
tests crash

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -0,0 +1,166 @@
...
+            var savedSettings = await _SettingsRepository.GetSettingAsync<StorageSettings>();
+            if (forceChoice || savedSettings == null)
+            {
+                return View(new StorageSettings()
+                {
+                    Provider = savedSettings?.Provider ?? BTCPayServer.Storage.Models.StorageProvider.FileSystem
+                });
+            }
+
+            return RedirectToAction("StorageProvider", new
```
NicolasDorier: can you use `nameof(StorageProvider)`

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -0,0 +1,166 @@
...
+            return RedirectToAction("StorageProvider", "Server", new
+            {
+                provider = viewModel.Provider.ToString()
+            });
+        }
+
+        [HttpGet("server/storage/{provider}")]
+        public async Task<IActionResult> StorageProvider(string provider)
+        {
+            var storageProvider = Enum.Parse(typeof(StorageProvider), provider);
```
NicolasDorier: uncaught exception  if provider can't be parsed or null

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -0,0 +1,166 @@
...
+            var data = (await _SettingsRepository.GetSettingAsync<StorageSettings>()) ?? new StorageSettings();
+
+            var storageProviderService =
+                _StorageProviderServices.Single(service => service.StorageProvider().Equals(storageProvider));
+
+            var viewName = $"Edit{storageProvider}StorageProvider";
```
NicolasDorier: don't do magic strings, use `nameof()` when you call `View` instead.

BTCPayServer/BTCPayServer.csproj
```diff
@@ -69,7 +69,12 @@
   </ItemGroup>
 
 	<ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.9" />
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.6" />
```
NicolasDorier: don't touch this version

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -0,0 +1,186 @@
...
+                        Severity = StatusMessageModel.StatusSeverity.Error,
+                        Message = $"{provider} provider is not supported"
+                    }.ToString()
+                });
+            }
+
+            var data = (await _SettingsRepository.GetSettingAsync<StorageSettings>()) ?? new StorageSettings();
+
+            var storageProviderService =
+                _StorageProviderServices.Single(service => service.StorageProvider().Equals(storageProvider));
```
NicolasDorier: if not found, then unhandled exception. Should also be case insensitive.

BTCPayServer/BTCPayServer.csproj
Kukks: must have been a rebase fuckup

NicolasDorier:
Need tests as well.

You should use dev secrets as I did with the transifex thing, so a dev can run manually the tests if he has some amazon creds.

BTCPayServer.Tests/BTCPayServer.Tests.csproj
```diff
@@ -33,4 +33,8 @@
   <ItemGroup>
     <ProjectReference Include="..\BTCPayServer\BTCPayServer.csproj" />
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="uploadtestfile.txt" />
```
NicolasDorier: just create dummy file dynamically

Kukks:
Added tests

NicolasDorier:
Need rebase, then I think we can merge (will need to setup CI later for those dependencies though)

BTCPayServer.Tests/StorageTests.cs
```diff
@@ -0,0 +1,247 @@
...
+                var controller = tester.PayTester.GetController<ServerController>(user.UserId, user.StoreId);
+
+//                //For some reason, the tests cache something on circleci and this is set by default
+//                //Initially, there is no configuration, make sure we display the choices available to configure
+//                Assert.IsType<StorageSettings>(Assert.IsType<ViewResult>(await controller.Storage()).Model);
+//
+//                //the file list should tell us it's not configured:
+//                var viewFilesViewModelInitial =
+//                    Assert.IsType<ViewFilesViewModel>(Assert.IsType<ViewResult>(await controller.Files()).Model);
+//                Assert.False(viewFilesViewModelInitial.StorageConfigured);
```
NicolasDorier: `For some reason, the tests cache something on circleci and this is set by default` not a problem of circleCI.

The problem is that the tests all run on a shared instance that may have data from previous test.

NicolasDorier: Also where does it works? The new account you create should not be admin, so you should not be able to access ServerController stuff.

Kukks: resolved: we use controller directly in tests

Kukks:
I've disabled the google/amazon code until someone has time to test it further

Issue #433, opened 2018-12-04
UI of partially paid invoice


hubiktomas:
I have a feedback from our customers - when they do accidentally a partial payment (for instance by not calculating the fee), the whole UI just updates the remaining amount without any other notification of accepted payment and notification of the need to pay the remaining payment. So if they did not notice the changed number, it looks like just nothing happened. We should probably make more visible in the UI that some funds are received and some still have to be sent.

britttttk:
It's very hard to see partial payments when viewing the invoice on QR code screen. If it's on the copy address screen the amount due is updated in large text. But I do agree it's hard to see the change. I also think the dropdown arrows are very faint to reveal the details. Even though this is standard for invoices I think it's not very user friendly. 

Some solution ideas:

partial payment changes the color of the "Awaiting Payment..." bar
change the "Awaiting Payment..." message to like "Partial Payment Received"

NicolasDorier:
ping @rockstardev

astupidmoose:
I think a small popup wouldn't be too extreme either since its a time sensitive issue and one that could result in cancelled orders. I like the "Partial Payment Received", but i think it would also make sense to deliver a "UH OH" popup as a call to attention

bitcoinshirt:
@astupidmoose I think that can be a good idea, but it needs to be handled so that the pop-up is not affected by adblockers and that it works well on mobile devices (most mobile browsers will block pop up from what I know?),  but that's a good approach, we clearly need to give them info right away.

astupidmoose:
Maybe a popup is a bad term here. But a change in the UI that involves the customer acknowledgement before proceeding and something that's so different it makes the customer look cause it's not just ticking down with a remaining amount.

pavlenex:
@rockstardev any idea how we can tackle partially paid the invoice and make it clearer that the invoice is partially paid and that customer has to pay the remaining amount.

How about something that @mariodian suggested in #434 that got overlooked IMO, at least the clear red warrning.

![49869081-4399fb80-fe4a-11e8-96d1-01162bd1b9ce](https://user-images.githubusercontent.com/36959754/54391858-6d132f00-46a6-11e9-8448-80c69843d294.png)
.

pavlenex:
@rockstardev any chance you can snug a warring when the invoice is partially paid to make it clearer for people that they need to pay the rest, right now it's so unclear unless they select the drop-down arrow.

I think @mariodian suggestion for at least the warning in different color should be a good place to start.
[...]

PR #709, opened 2019-03-22
Improve design to be more sober

clintnetwork:
Hi make a little change about CSS, I think it's better now, have an hard background-shadow is not very beautiful.

Best regards

Before:
old

After:
new

### DIFF
--- a/BTCPayServer/wwwroot/checkout/css/normalizer.css
+++ b/BTCPayServer/wwwroot/checkout/css/normalizer.css
@@ -331,7 +331,7 @@ body {
     line-height: 1.5;
     background-color: #fff;
     margin: 0;
-    background: #E4E4E4;
+    background: #f7f7f7;
 }
 
 [tabindex="-1"]:focus {
@@ -1195,7 +1195,7 @@ input[type="button"].action-button {
     background-color: #fff;
     background-clip: padding-box;
     border: 1px solid rgba(0, 0, 0, 0.2);
-    border-radius: 0.3rem;
+    border-radius: 0;
     outline: 0;
 }
 
@@ -6831,7 +6831,7 @@ button.close {
     background-color: #fff;
     background-clip: padding-box;
     border: 1px solid rgba(0, 0, 0, 0.2);
-    border-radius: 0.3rem;
+    border-radius: 0;
     outline: 0;
 }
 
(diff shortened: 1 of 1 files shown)


britttttk:
I like previous dark shadow better. Perhaps there is something in between? New version is a bit light for me. :)

clintnetwork:
Hmm, something like this ?
image

clintnetwork:
I would like to add something: the trends is light / sober, watch by example https://now.sh :)

britttttk:
Sorry I've never heard of sober, but I like the original shadow best. It makes the invoice stand out more for me, in my opinion. :)

clintnetwork:
So we need poll :p

pavlenex:
Thanks for your PR, unfortunately I think you're trying to fix something that is not broken. Instead I suggest you take a look at this issue #433 and try to tackle it.

The current design makes invoice stand out which can be helpful for QR code scanning, especially on older phones.

clintnetwork:
First I don't have tried to fix a problem but improve an existing design.

Secondly if you love

britttttk:
@clintnetwork yes can you help with 443? :) it's been open a very long time!

clintnetwork:
You see bitpay has understanded the concept of design 💃 

image

pavlenex:
Hey @clintnetwork, luckily we're not trying to be BitPay.

We base our design based on feedback from community and merchants and not based on what we think looks pretty, so functional over pretty for now :)

As I mentioned earlier, there are many other issues concerning design which are a priority at this moment, if you want to help out, you can join us on our [Mattermost chat](https://chat.btcpayserver.org/) to discuss with developers, or just let's discuss what we can improve with functional design.

To be honest with you in my personal opinion, BitPay's recent changes in design are to decrease customers support for them, not make the shopping experience better for the buyers.

mariodian:
I agree with @clintnetwork here. In the era of flat designs the current shadow is way too dark and I would say for no apparent reason.

On the other hand the whole invoice page needs a major redesign (including the mentioned #433). For example, I don't see a reason why the invoice element would be so narrow as if it was viewed on a mobile when viewed on a PC. Also I don't think we need to display BTCPay logo to save some space for mobiles to display more important info etc. I'll likely draft something for the distant future.

PR #711, opened 2019-03-24
Javascript floating point math fix, closes #701

### LINKED ISSUE #701: Sats rounding error
@rockstardev 😅 
Screen Shot 2019-03-21 at 19 17 32

rockstardev:


NicolasDorier:
seems to be called only when showing satoshi

Issue #694, opened 2019-03-19
Noscript invoice


pavlenex:
As discussed in #320 we should have noscript invoices.

> I noticed in the code the other day that there is a noscript version of the checkout if you append /invoice-noscript to the url. There shoudl also be a failsafe in noscript mode to tell the user to follow that link

pavlenex:
[Added.](https://github.com/btcpayserver/btcpayserver/pull/732/commits/e8b71f36b2954cd45034287188305548823ed2fe)

Issue #320, opened 2018-10-07
Support Tor Browser and non-javascript invoice


ghost:
For all the evil that BitPay has done, they have, to their credit at least have the invoice working without JavaScript. I miss this feature in BtcPay.

Currently, to pay a BtcPay invoice:
- I have to whitelist the scripts in NoScript
- I have to change Tor Browser security level to 'Lowest' (No idea why, but the invoice still refuses to work with security 'High' and scripts whitelisted) 

Please make the Tor Browser a supported target - and bonus points for a NoScript functionality!

Kukks:
I noticed in the code the other day that there is a noscript version of the checkout if you append `/invoice-noscript` to the url. There shoudl also be a failsafe in noscript mode to tell the user to follow that link

NicolasDorier:
I doubt the noscript version works. We probably need to do a second checkout page.

pavlenex:
@NicolasDorier close?

NicolasDorier:
I don't quite understand, Tor Browser works for me.
That said I wished to have a non JS page as well.

pavlenex:
@NicolasDorier This is an old issue requesting Tor + non js invoice page. Since we now support Tor I wanted to know if we should close it or rename it only for non-js page?

Kukks:
Let's split the non js invoice into a new issue and close this one :)
ᐧ

NicolasDorier:
ACK

PR #732, opened 2019-03-31
UI fixes

rockstardev:
Fixing that console error when customer email is updated, and then also addressing #320

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -438,7 +438,7 @@ namespace BTCPayServer.Controllers
                 return BadRequest(ModelState);
             }
             await _InvoiceRepository.UpdateInvoice(invoiceId, data).ConfigureAwait(false);
-            return Ok();
+            return Ok("{}");
         }
 
         [HttpGet]
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -188,7 +188,7 @@ namespace BTCPayServer.Controllers
             //Keep compatibility with Bitpay
             invoiceId = invoiceId ?? id;
             id = invoiceId;
-            ////
+            //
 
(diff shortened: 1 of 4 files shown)


pavlenex:
Also closes #694

rockstardev:
I'm rarely happier then when I close issue opened by @pavlenex <3

BTCPayServer/Views/Invoice/Checkout.cshtml
```diff
@@ -51,9 +51,9 @@
 
             <p>Alternatively, click below to continue to our HTML-only invoice.</p>
 
-            <form action="/invoice-noscript" method="GET">
-                <button style="text-decoration: underline; color: blue">Continue to javascript-disabled invoice &gt;</button>
-            </form>
+            <a href="/invoice-noscript?id=@Model.InvoiceId" style="text-decoration: underline; color: blue">
```
Kukks: better to use `asp-action` and `asp-route-id` :)

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -213,6 +213,23 @@ private string ToString(PaymentMethodId paymentMethodId)
...
+            //Keep compatibility with Bitpay
+            invoiceId = invoiceId ?? id;
+            id = invoiceId;
+            //
+
+            var model = await GetInvoiceModel(invoiceId, paymentMethodId == null ? null : PaymentMethodId.Parse(paymentMethodId));
+            if (model == null)
+                return NotFound();
+
+            return View(nameof(CheckoutNoScript), model);
```
Kukks: I don't think you need to define the view name, it automatically uses the action method's name if non is specified

rockstardev: I got to that point by experimenting with View names... thinking to extract public parts of Invoice controller to CheckoutController. Thanks for suggestion, fixed it in commit b4b69394983bf1ee0b5771ff91953a1d9056c15c

BTCPayServer/Views/Invoice/Checkout.cshtml
rockstardev: Tried it... ended up prefering single `href` element than `asp-action`, `asp-controller` and `asp-route-id`

PR #735, opened 2019-04-04
Add link to andreas video during IBD

NicolasDorier:
image

Ping for feedback @pavlenex

### DIFF
--- a/BTCPayServer/Views/Shared/SyncModal.cshtml
+++ b/BTCPayServer/Views/Shared/SyncModal.cshtml
@@ -10,8 +10,7 @@
         </div>
         <div class="modal-body">
             <p>
-                Some of your nodes are still synchronizing...<br />
-                BTCPay Server will not create invoices with those cryptocurrencies.
+                Your node is synching the entire blockchain and validating the consensus rules...
             </p>
             @foreach (var line in dashboard.GetAll())
             {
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Motivation: While cat video are cool, I believe that sending the user look a video of Andreas is likely to keep  him occupied for the 2 days required to sync.

pavlenex:
I like that you included fast sync.

I'd just make it more contextual:

"_Your node is syncing the entire blockchain and validating the consensus rules_. 
[Watch this video ](https://www.youtube.com/watch?v=OrYDehC-8TU)to understand the importance of blockchain synchronization. 

Familiar with command line? Check [FastSync](https://github.com/btcpayserver/btcpayserver-docker/tree/master/contrib/FastSync)."

I think we can even deduct the first sentence. The reason I'd like to see less words is to make translating easier if in future if we decide to make it multilingual.

NicolasDorier:
image

NicolasDorier:
without break

image

pavlenex:
I think we can even remove `"Your node is syncing the entire blockchain and validating the consensus rules.` because newbies won't understand that at all until they watch the video anyway :-)

NicolasDorier:
image

NicolasDorier:
Syncing or synching?

pavlenex:
Ah, the eternal [dilemma](https://english.stackexchange.com/questions/681/synced-or-synched). In my mind, it's `Syncing` 

I think we should just leave it the way it is now so I guess `synching`. I remember people complained it the past about this. in past.

I see we have :
Your nodes are synching but later syncing 

Let's leave it for a while to see if a native can resolve the dilemma.

pavlenex:
Great solution.

PR #740, opened 2019-04-05
Button to switch between time formats, width fix

rockstardev:
pavlenex


pavlenex:
A good solution that will make both sides happy. ACK.

PR #745, opened 2019-04-08
Add payment data to crypto info in invoice api model

Kukks:
Depends on https://github.com/MetacoSA/NBitpayClient/pull/22


Kukks:
@NicolasDorier  done

PR #747, opened 2019-04-08
Pos decimal problem on french OS

Dolu89:
I'm on a french windows 10. 
In POS app, if total amount had a dot (example: 1.20 $), decimal amount in POST controller was at 0 because in JS, my system wanted 1,20 and not 1.20


NicolasDorier:
Thanks, did this one instead which is 2 line change (https://github.com/btcpayserver/btcpayserver/commit/2ca11ed692a43b0e41318f244319f67bc252716b)

PR #754, opened 2019-04-11
fix pos settings savings for notifications

### LINKED ISSUE #751: [PoS app] Unable to have notificationUrl callback working
**Describe the problem/bug**
I am unable to have notificationUrl  callback in PoS app working. 
To test it I set "https://test.com" in the "Callback Notification Url" form field in PoS settings, then I go to the PoS view and I do a payment.  
Then when I go to https://btcpay.example.com/invoices/{invoiceId} the "Notification url" and "Order Id" are empty. In nginx logs I don't see any call either.

Also, when I enter a value in the "Callback Notification Url" form field, click save and come back in settings, I don't see the value I have just set.

The feature doesn't seem to work.

**Your environment**
* Version of BTCPay Server: v1.0.3.92
* Deployment method: docker

Kukks:
closes #751

### DIFF
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -178,7 +178,10 @@ namespace BTCPayServer.Controllers
                 CustomButtonText = vm.CustomButtonText,
                 CustomTipText = vm.CustomTipText,
                 CustomTipPercentages = ListSplit(vm.CustomTipPercentages),
-                CustomCSSLink = vm.CustomCSSLink
+                CustomCSSLink = vm.CustomCSSLink,
+                NotificationUrl = vm.NotificationUrl,
+                NotificationEmail = vm.NotificationEmail
+                
             });
             await UpdateAppSettings(app);
             StatusMessage = "App updated";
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/AppsController.PointOfSale.cs
```diff
@@ -178,7 +178,10 @@ public async Task<IActionResult> UpdatePointOfSale(string appId, UpdatePointOfSa
                 CustomButtonText = vm.CustomButtonText,
                 CustomTipText = vm.CustomTipText,
                 CustomTipPercentages = ListSplit(vm.CustomTipPercentages),
-                CustomCSSLink = vm.CustomCSSLink
+                CustomCSSLink = vm.CustomCSSLink,
+                NotificationUrl = vm.NotificationUrl,
+                NotificationEmail = vm.NotificationEmail
+
```
NicolasDorier: couldn't this be used by a spammer to spam random emails?

Kukks: this one is  only configured by pos settings so store owner would need to set up email.

NicolasDorier: it does not, this come from the POST data.

Kukks: You are talking about this: https://github.com/btcpayserver/btcpayserver/blob/7a21c03896282e07efc8adb92a2b226233d5171e/BTCPayServer/Controllers/AppsPublicController.cs#L215

NicolasDorier: oops

PR #755, opened 2019-04-11
allow facade lock in for access token creation 

### LINKED ISSUE #692: Pairing Link that Locks in Merchant Facade
In apps that pair to BTCPay, right now we can give the user the link `<host>/api-tokens` and it conveniently takes them right to pairing. However many users get confused by the `facade` drop-down. 

It's a minor thing, but it would be cool to have it so a different sub-URI,  (maybe `<host>/merchant-api-tokens`) would take them to a different version of the pairing page where they are probibited from selecting a facade and `merchant` is just assumed. One less thing for them to screw up.

Kukks:
closes #692

### DIFF
--- a/BTCPayServer/Controllers/StoresController.cs
+++ b/BTCPayServer/Controllers/StoresController.cs
@@ -29,6 +29,7 @@ using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.Extensions.Options;
 using NBitcoin;
 using NBitcoin.DataEncoders;
+using NBitpayClient;
 
 namespace BTCPayServer.Controllers
 {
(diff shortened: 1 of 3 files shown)


NicolasDorier:
the facade feature is not used at all. Better to remove all trace of it in btcpay.

Kukks:
Maybe we can merge this one in anyway @NicolasDorier until we sort out the other PR( which may take a while since it is a bit big)

NicolasDorier:
you want me to bulldoze the thing?

Kukks:
https://giphy.com/gifs/just-do-it-24xRxrDCLxhT2

NicolasDorier:
4067d4b00f0e2cffedffa06600f791d8efaed974

PR #756, opened 2019-04-11
Allow POS to redirect invoices automatically after paid

### LINKED ISSUE #730: PoS - redirect to main screen a few seconds after payment
In a public Point of Sale, a customer should always see the main screen. 
Currently, if a customer make a payment and doesn't return to the main screen manually, the next customer will not see the main screen but will be presented with the receipt of the previous customer (which is terrible UX).  

In a PoS app, it should redirect automatically to the main screen a few seconds after a successful payment

Kukks:
closes #730

### DIFF
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -79,7 +79,7 @@ namespace BTCPayServer.Controllers
             public string CustomCSSLink { get; set; }
             public string NotificationEmail { get; set; }
             public string NotificationUrl { get; set; }
-            public bool RedirectAutomatically { get; set; }
+            public bool? RedirectAutomatically { get; set; }
         }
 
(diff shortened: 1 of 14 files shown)


NicolasDorier:
You are on fire today. I think this should be a store property in checkout experience though.

Kukks:
I ragequit Shekiro as well 🗡

Kukks:
> You are on fire today. I think this should be a store property in checkout experience though.

I don't fully agree. I can imagine this only being wanted on a POS app for a physical location but not on their online store or their online POS which would cause people to create multiple stores to get around it

NicolasDorier:
image

With the settings at app level being "Yes/No/Same as store settings"

NicolasDorier:
added `bool? RedirectAutomatically` to invoice type of NBitpay

Kukks:
changes done and tested @NicolasDorier

PR #787, opened 2019-04-25
Add U2f Login Support

### LINKED ISSUE #53: Add U2F verification to further secure the server login
Currently, there is 2FA authentication which is a great way to secure online accounts, but I would like to suggest adding U2F authentication (Ledger/Trezor support it) to further enhance the security of the server login.

Kukks:
Closes #53

### DIFF
--- a/BTCPayServer/Controllers/AccountController.cs
+++ b/BTCPayServer/Controllers/AccountController.cs
@@ -109,30 +109,16 @@ namespace BTCPayServer.Controllers
                         {
                             // we need to do an actual sign in attempt so that 2fa can function in next step
                             await _signInManager.PasswordSignInAsync(model.Email, model.Password, model.RememberMe, lockoutOnFailure: true);
-                            twoFModel = new LoginWith2faViewModel { RememberMe = model.RememberMe };
-                        }
-
-                        if (Request.IsHttps)
-                        {
-
-                            var u2fChallenge = await _u2FService.GenerateDeviceChallenges(user.Id,
-                                Request.GetAbsoluteUriNoPathBase().ToString().TrimEnd('/'));
-
-                            u2fModel = new LoginWithU2FViewModel()
+                            twoFModel = new LoginWith2faViewModel
                             {
-                                Version = u2fChallenge[0].version,
-                                Challenge = u2fChallenge[0].challenge,
-                                Challenges = JsonConvert.SerializeObject(u2fChallenge),
-                                AppId = u2fChallenge[0].appId,
-                                UserId = user.Id
+                                RememberMe = model.RememberMe
                             };
-
                         }
 
                         return View("SecondaryLogin", new SecondaryLoginViewModel()
                         {
                             LoginWith2FaViewModel = twoFModel,
-                            LoginWithU2FViewModel = u2fModel
+                            LoginWithU2FViewModel = await BuildU2FViewModel(model.RememberMe, user)
                         });
                     }
                     else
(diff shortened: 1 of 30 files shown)


pavlenex:
Tests are failing, but I've tested this long-awaited PR on over 10 U2F devices.

tACK.
Devices below work out of the box on both Chrome and Firefox (all latest).
U2F BTCPay

One exception is Nano S that has already known Chrome problem, so you need to quickly confirm when prompted, but that one should be fixed soon by Ledger.

This was so much fun to test.

Kukks:
I actually did all my dev work and testing on a Nano S so it was not that bad with it :)

BTCPayServer/Views/Account/LoginWithU2F.cshtml
```diff
@@ -0,0 +1,55 @@
...
+        2: "Bad request error, try again",
+        3: "This key isn't supported, please try another one",
+        4: 'The device is already registered, please login',
+        5: 'Authentication timed out. Please reload to try again.'
+    };
+    setTimeout(function() {
+        window.u2f.sign(
+            "@Model.AppId",
+            "@Model.Challenge",
+            @Html.Raw(@Model.Challenges), function (data) {
```
NicolasDorier: JS injection possible

BTCPayServer/Views/Account/LoginWithU2F.cshtml
```diff
@@ -0,0 +1,55 @@
...
+                console.log("call back data: ", data);
```
NicolasDorier: no logs

NicolasDorier:
U2F is so simple that I think you can remove the dependency to `U2F.Core`. IMHO, this is not worth it.

NicolasDorier:
@pavlenex you can test it?

Kukks:
> U2F is so simple that I think you can remove the dependency to `U2F.Core`. IMHO, this is not worth it.

I do not have experience with the lower level crypto stuff for the spec so I prefer trusting a tested package

BTCPayServer/Views/Account/LoginWithU2F.cshtml
Kukks: What should I do for this then? I don't think injection here would be harmful?

pavlenex:
My suggestions are all UX related.

## Setup Page (Registring U2F Device)
Capture
1. Add [font-awesome loading icon ](https://fontawesome.com/how-to-use/on-the-web/styling/animating-icons)next to Registring U2F Device, to indicate something is happening, loading, waiting.
2. Replace description text, with `Insert your U2F device or a hardware wallet into your computer's USB port. If it has a button, tap on it`

## Login page (U2F only enabled)
Same as for point 1. and 2. in the setup page.

## Login page (Both U2F and 2FA enabled)
Personally, if both are enabled, I'd show U2F with a link at the bottom `Try alternative authentication method`, but it really is not a big deal.

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -91,8 +97,39 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
...
+
+                        return View("SecondaryLogin", new SecondaryLoginViewModel()
+                        {
+                            LoginWith2FaViewModel = twoFModel,
+                            LoginWithU2FViewModel = await BuildU2FViewModel(model.RememberMe, user)
+                        });
+                    }
+                    else
+                    {
+                      var incrementAccessFailedResult = await  _userManager.AccessFailedAsync(user);
```
NicolasDorier: If you plan doing this, then you need to document to user how to unlock his account by modifying the DB directly.

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -123,6 +162,63 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
             return View(model);
         }
 
+        private async Task<LoginWithU2FViewModel> BuildU2FViewModel(bool rememberMe, ApplicationUser user)
+        {
+            if (Request.IsHttps)
```
NicolasDorier: If onion is also fine. I think there is a method somewhere I created search for "IsSecure" in the code.

BTCPayServer/Controllers/AccountController.cs
Kukks: The lockout is interval based and already enabled on normal login flow

Kukks: I think after 10 mins the user is no longer locked using ASP Identity

NicolasDorier:
Need tests

BTCPayServer/Views/Manage/AddU2FDevice.cshtml
```diff
@@ -0,0 +1,59 @@
...
+            2: "Bad request error, try again",
+            3: "This key isn't supported, please try another one",
+            4: 'The device is already registered, please login',
+            5: 'Authentication timed out. Please reload to try again.'
+        };
+
+        setTimeout(function() {
+                var request = { "challenge": "@Model.Challenge", "version": "@Model.Version", "appId": "@Model.AppId" };
+                var registerRequests = [{version: request.version, challenge: request.challenge}];
+                console.log(request);
```
NicolasDorier: don't log

BTCPayServer/Views/Account/LoginWithU2F.cshtml
NicolasDorier: I don't remember how I do, search `srvModel` in the code, I show how to get a json object properly deserialized.

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -123,6 +162,63 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
...
+            }
+            try
+            {
+                if (await _u2FService.AuthenticateUser(viewModel.UserId, viewModel.DeviceResponse))
+                {
+                    await _signInManager.SignInAsync(user, viewModel.RememberMe, "U2F");
+                    _logger.LogInformation("User logged in.");
+                    return RedirectToLocal(returnUrl);
+                }
+                throw new Exception("Invalid login attempt.");
```
NicolasDorier: wtf? just don't throw anything here and do what is in the catch next.

PR #789, opened 2019-04-25
fix failing pos test

Kukks:


rockstardev:
@Kukks I think POS test failing was caused by something else - take a look at this commit from PR #790 :

https://github.com/btcpayserver/btcpayserver/pull/790/commits/da04244ddef34a54ff4bc2b8a54de8785a5f3b4b

Kukks:
fixed with #790

PR #790, opened 2019-04-25
Filtering invoice list by startdate and enddate, other fixes

rockstardev:
One notice: since datetime can be specified like `4/11/2019 3:34:00 PM` I needed to enforce splitting of parameters in search string with comma.

Also because our client side dates are displayed based on browser timezone settings, I am now passing that value back to server to ensure smooth search (user doesn't need to convert his local time to UTC... which is how invoices are saved in database). Note this works both for invoice display and invoice export.

Finally, tons of small fixes, would appreciate if @Kukks can verify that `CanUsePoSApp` test was broken and that change I made is good.

Here is that main feature in action:

ability-to-search-1

And then when start date for search is provided:

ability-to-search-2

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -742,6 +742,12 @@ namespace BTCPayServer.Tests
                 AssertSearchInvoice(acc, false, invoice.Id, $"exceptionstatus:paidOver");
                 AssertSearchInvoice(acc, true, invoice.Id, $"unusual:true");
                 AssertSearchInvoice(acc, false, invoice.Id, $"unusual:false");
+
+                var time = invoice.InvoiceTime;
+                AssertSearchInvoice(acc, true, invoice.Id, $"startdate:{time.ToString("yyyy-MM-dd HH:mm:ss")}");
+                AssertSearchInvoice(acc, true, invoice.Id, $"enddate:{time.ToStringLowerInvariant()}");
+                AssertSearchInvoice(acc, false, invoice.Id, $"startdate:{time.AddSeconds(1).ToString("yyyy-MM-dd HH:mm:ss")}");
+                AssertSearchInvoice(acc, false, invoice.Id, $"enddate:{time.AddSeconds(-1).ToString("yyyy-MM-dd HH:mm:ss")}");
             }
         }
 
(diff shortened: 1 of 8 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -1665,10 +1677,10 @@ public void CanUsePoSApp()
                 
 
                 // testing custom amount
-                var action = Assert.IsType<RedirectToActionResult>(publicApps.ViewPointOfSale(appId, 5, null, null, null, null, "donation").Result);
```
NicolasDorier: why was it not working ???

NicolasDorier:
awesome!

NicolasDorier:
Just one question: I don't understand why the test was broken

BTCPayServer.Tests/UnitTest1.cs
Kukks: Because I added another test case on top of it with the same amount

Kukks: my bad from https://github.com/btcpayserver/btcpayserver/pull/784

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -465,25 +465,22 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody]Upda
         [Route("invoices")]
         [Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]
         [BitpayAPIConstraint(false)]
-        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50)
+        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0)
         {
-            if (searchTerm == null)
```
Kukks: Why did you remove this?

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -504,35 +501,35 @@ public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip
...
                 UserId = GetUserId(),
-                Unusual = !filterString.Filters.ContainsKey("unusual") ? null
-                          : !bool.TryParse(filterString.Filters["unusual"].First(), out var r) ? (bool?)null
-                          : r,
-                Status = filterString.Filters.ContainsKey("status") ? filterString.Filters["status"].ToArray() : null,
-                ExceptionStatus = filterString.Filters.ContainsKey("exceptionstatus") ? filterString.Filters["exceptionstatus"].ToArray() : null,
-                StoreId = filterString.Filters.ContainsKey("storeid") ? filterString.Filters["storeid"].ToArray() : null,
-                ItemCode = filterString.Filters.ContainsKey("itemcode") ? filterString.Filters["itemcode"].ToArray() : null,
-                OrderId = filterString.Filters.ContainsKey("orderid") ? filterString.Filters["orderid"].ToArray() : null
+                Unusual = fs.GetFilterBool("unusual"),
```
Kukks: This looks so much better :D

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -627,27 +624,6 @@ public async Task<IActionResult> CreateInvoice(CreateInvoiceModel model, Cancell
             }
         }
 
-        [HttpPost]
-        [Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]
-        [BitpayAPIConstraint(false)]
-        public IActionResult SearchInvoice(InvoicesModel invoices)
-        {
-            if (invoices.SearchTerm == null)
```
Kukks: Why did you remove?

rockstardev:
@Kukks for session search perseverance - you now also have timezone offset, plus I've moved parameters to GET. Meaning - if people want to get back to certain search results - they just go to URL.

Issue #22, opened 2018-01-03
Address reuse on same invoice


astupidmoose:
Currently BTCpay generates a new address for each payment. This can cause confusion for the customer and shop owner if the invoice is paid with multiple payments. It also potentially generates an unnecessary amount of addresses. 

I think reusing the same BTC address for all payments to a single invoice should not be an issue and much cleaner. 

Thoughts?

NicolasDorier:
While we can do that, I am unsure it is really useful. As when multiple crypto will be supported, it will be possible to pay partly in say LTC and partly in BTC in which case you would have 2 addresses anyway.

I think you got confused because the woocommerce plugin was showing you only one of the two addresses. If this is the case, then the best way is for my to update the invoice details you receive on woocommerce to include ALL the addresses and payment already done.

I am however unsure about this approach either, maybe a better way is for you to just have a link to the invoice details on the btcpay website? The details page should summarize everything. The only thing I would need to do is to remove authentication of this page so you can just click on the invoice url in woocommerce and it open the page in BTCPay.

bitcoinshirt:
Now that this is our oldest opened issue, let's see what to do with it.
@astupidmoose looking at you.
Was there really a confusion from your customers?
I assume with invoice export this is easily visible on your end.

Kukks:
Considering all the questions and "issues " people ask about the gap limit because of the amount of unused addresses we generate, we should consider reusing addresses for same invoice. 

After that, we take it to the next level and generate addresses on demand base don invoice creation or by letting the user toggle the crypto they want to pay in.

astupidmoose:
It's not really any longer an issue on our side and haven't heard anymore issues from this.

Using the same address for the same invoice though might reduce load on gap limits and maybe system resources?

At this point, I don't think its mission critical but maybe still a good idea

astupidmoose:
I like #2, default to bitcoin but have a drop-down of other choices. Eliminates the need for the extra step. 

As for paying in multiple coins, this actually is a thing and many people do it on our service. I don't understand it, but it exists lol.

bitcoinshirt:
Well isn't that quite close to an already existing solution?
![screen shot 2018-12-10 at 22 46 04](https://user-images.githubusercontent.com/36959754/49763612-7cf92c80-fccd-11e8-86db-a041c58c1cc3.png)
BTC Default, but the addres is generated only when a new coin is selected? And we just need to make selection a little bit better somehow in terms of UI.

[...]

PR #792, opened 2019-04-26
Allow invoice creation to only allow specific payment methods in UI

Kukks:
This PR introduces a new parameter to the invoice creation api and on the Create Invoice UI, "Payment Methods". It allows you to specify which coins( and their lightning counterparts) are to be enabled for this invoice.  If this list is left null or is an empty string, it enabled all by default. If a specified payment method is not available, it is ignored. 

This is slightly related to #22. With this new option, merchants can build UIs that allow them to choose what coins they would like to pay with before creating the invoice

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.cs
+++ b/BTCPayServer/Controllers/InvoiceController.cs
@@ -160,10 +160,16 @@ namespace BTCPayServer.Controllers
 
             var rateRules = storeBlob.GetRateRules(_NetworkProvider);
             var fetchingByCurrencyPair = _RateProvider.FetchRates(currencyPairsToFetch, rateRules, cancellationToken);
-
             var fetchingAll = WhenAllFetched(logs, fetchingByCurrencyPair);
             var supportedPaymentMethods = store.GetSupportedPaymentMethods(_NetworkProvider)
                                                .Where(s => !excludeFilter.Match(s.PaymentId))
+                                               .Where(s => invoice.PaymentMethods == null || 
+                                                           !invoice.PaymentMethods.Any() || 
+                                                           invoice.PaymentMethods.Any(selectedPaymentMethod =>
+                                                               s.PaymentId.ToString()
+                                                                   .Equals(
+                                                                       selectedPaymentMethod, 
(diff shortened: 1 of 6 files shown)


BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -160,10 +160,16 @@ internal async Task<DataWrapper<InvoiceResponse>> CreateInvoiceCore(CreateInvoic
 
             var rateRules = storeBlob.GetRateRules(_NetworkProvider);
             var fetchingByCurrencyPair = _RateProvider.FetchRates(currencyPairsToFetch, rateRules, cancellationToken);
-
             var fetchingAll = WhenAllFetched(logs, fetchingByCurrencyPair);
             var supportedPaymentMethods = store.GetSupportedPaymentMethods(_NetworkProvider)
                                                .Where(s => !excludeFilter.Match(s.PaymentId))
+                                               .Where(s => invoice.PaymentMethods == null ||
```
NicolasDorier: can't we just use the exludeFilters that I created especially for this purpose?

Kukks: This is on an invoice level or do you mean specifically to avoid that `Where` clause?

NicolasDorier: Actually aren't you coding the same feature that is already available on https://github.com/btcpayserver/btcpayserver/pull/792/files#diff-cf427292b2f76704cc7a734a60602ddeR139 ?

Kukks: WHAAAAAT

Kukks: DOH this is exactly the same thing

Kukks: Ok let me just make this PR a way to do this on create invoice page instead lol

NicolasDorier:
Can't select multiple payment methods.

Kukks:
> Can't select multiple payment methods.

As in it doesn't allow you through UI or just you shouldnt be able to?

The select list in UI is a multiselect

PR #794, opened 2019-04-26
Show email warning on apps when settings are not complete

### LINKED ISSUE #693: E-mail invoice notifications on a store-level
Currently, if people aren't using CMS, it's impossible for them to get/send e-mail notifications for different invoice status. Of course, technically advanced users can use the API, but noobs have no way of setting that up.

We now have SMTP settings at a store level and should extend the emails to be sent for different invoice statuses. Ideally it should be highly-customizable, but for starters, a standardized messages for each invoice status is fine imo.

**Provide examples**
This solution is mostly for people using BTCPay Apps.

Use case 1: Casa Hodl store uses payment button for their orders. They need to check BTCPay constantly to know when to ship the order.
Use case 2: Charity uses payment button to collect donation and wants to send thank you message to whoever donates.
Use case 3: Crowdfunding user wants to get notification when a new donation is received.

**Additional context**
Related #172 #468 #683

Kukks:
closes #693

image

@pavlenex Is this good enough?

### DIFF
--- a/BTCPayServer/Controllers/AppsController.Crowdfund.cs
+++ b/BTCPayServer/Controllers/AppsController.Crowdfund.cs
@@ -5,6 +5,7 @@ using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using BTCPayServer.Models.AppViewModels;
 using BTCPayServer.Services.Apps;
+using BTCPayServer.Services.Mails;
 using Microsoft.AspNetCore.Mvc;
 
 namespace BTCPayServer.Controllers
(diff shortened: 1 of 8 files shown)


pavlenex:
@Kukks Yup, quite clear.

BTCPayServer/Controllers/AppsController.Crowdfund.cs
```diff
@@ -31,8 +32,13 @@ public async Task<IActionResult> UpdateCrowdfund(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<CrowdfundSettings>();
+
+            var emailWarning =
+                !((await (_emailSenderFactory.GetEmailSender(app.StoreDataId) as EmailSender)?.GetEmailSettings())
```
NicolasDorier: You can replace all this shit with

```csharp
var emailWarning = !((await _emailSenderFactory.GetEmailSender(app.StoreDataId) as EmailSender)?.GetEmailSettings())?.IsComplete() is true);
```

BTCPayServer/Controllers/AppsController.PointOfSale.cs
```diff
@@ -90,8 +91,12 @@ public async Task<IActionResult> UpdatePointOfSale(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
+            var emailWarning =
```
NicolasDorier: Refactor with above comment

PR #800, opened 2019-04-30
Showing parts of invoice details in grid for quicker administration

rockstardev:
As requested by @pavlenex, here is preview of this functionality

invoices expansion

I've refactored loading of invoice details - taken out `EntityToDto` call because it was unnecessary.

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -47,9 +47,9 @@ namespace BTCPayServer.Controllers
             if (invoice == null)
                 return NotFound();
 
-            var dto = invoice.EntityToDTO(_NetworkProvider);
+            var prodInfo = invoice.ProductInformation;
             var store = await _StoreRepository.FindStore(invoice.StoreId);
-            InvoiceDetailsModel model = new InvoiceDetailsModel()
+            var model = new InvoiceDetailsModel()
             {
                 StoreName = store.StoreName,
                 StoreLink = Url.Action(nameof(StoresController.UpdateStore), "Stores", new { storeId = store.Id }),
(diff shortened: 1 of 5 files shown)


rockstardev:
Hey @Kukks that `BTCPayServer.Tests.StorageTests.CanUseAzureBlobStorage` test is again failing... can you please tell me what should I do in order to make this pass for my PRs?

Kukks:
@rockstardev  hey is it failing locally? You need to get an azureblob storage service on azure and save it on your machine through dotnet use secrets. `Run "dotnet user-secrets set AzureBlobStorageConnectionString <value>"`

Kukks:
Ah you mean in circleci, @NicolasDorier  didn't you add your own.conmection string to circle ci?

pavlenex:
Would it look better if we replaced arrow up/down with +/-?

rockstardev:
@Kukks what confuses me is that this test fails only sometimes... if you check your PRs they are passing.

@pavlenex - you sure about it? I think these double arrows are more intuitive, but if you are sure about + / - I can do it that way.

pavlenex:
@rockstardev not sure at all, just asking if it looks better or not. I remember the feature for modal pop up for created invoice, it's an arrow, I discovered it only a few weeks ago.

If you tried both plus/minus and arrows, and arrow look better, let's go with arrows then.

rockstardev:
I've tried plus - it looks too much as "add"... let's stick with arrows.

For mobile users - there is way more we can do usability standpoint... at this moment of time I wouldn't expand the scope. I'll ping you for private conversation on the topic - that's actually why I'm debating on whether to make Details an icon.

rockstardev:
I've tried plus - it looks too much as "add"... let's stick with arrows.

plus-as-option

For mobile users - there is way more we can do usability standpoint... at this moment of time I wouldn't expand the scope. I'll ping you for private conversation on the topic - that's actually why I'm debating on whether to make Details an icon.

PR #803, opened 2019-05-02
Fix Azure Storage 

Kukks:


NicolasDorier:
The test is still broken.

Kukks:
no it works, I think the secrets are not provided when running pull requests

NicolasDorier:
Can you fix so it does not run during PRs? I thought I did it, but somehow it does not work?

NicolasDorier:
Trying something which might work.

PR #808, opened 2019-05-03
Fix date time issues on crowdfund.payment requests

Kukks:
* Bumped flatpickr minor versions
* fixed bugs with live update of crowdfund/payment requests (styling and time-related elements stayed in place)
* fixed display of date/time in edit ui(timezone crap)


NicolasDorier:
why does your PR try to run azure tests? Normally it should skip it as it is a PR....

PR #809, opened 2019-05-03
do not force azure test if not set

Kukks:


NicolasDorier:
We should not do this. Else we won't know if the test is passing because the test is disabled by mistake of if it pass legitimately.

PR #822, opened 2019-05-08
Fix U2F Auth on Linux

Kukks:


Kukks:
I tried this out locally and on https://fiat.kukks.org/ which has this PR deployed. @pavlenex  Let me know if you have issues

PR #824, opened 2019-05-08
Allow import of coldcard wallet 

Kukks:


BTCPayServer/wwwroot/main/css/site.css
```diff
@@ -19,3 +19,7 @@
 .only-for-js, .input-group-clear{
     display: none;
 }
+
+.display-when-ledger-connected{
```
NicolasDorier: not a fan of having that in the site wide style. It's place is really better in `style=`

NicolasDorier:
Actually it is not perfect. Many info is not saved, like the fingerprint and root keypath.

PR #826, opened 2019-05-09
Move ElectrumMapping to BtcPayNetwork

Kukks:


BTCPayServer/BTCPayNetworkProvider.Bitcoin.cs
```diff
@@ -26,7 +26,19 @@ public void InitBitcoin()
                 LightningImagePath = "imlegacy/bitcoin-lightning.svg",
                 DefaultSettings = BTCPayDefaultSettings.GetDefaultSettings(NetworkType),
                 CoinType = NetworkType == NetworkType.Mainnet ? new KeyPath("0'") : new KeyPath("1'"),
-                SupportRBF = true
+                SupportRBF = true,
```
NicolasDorier: You need to check the `NetworkType` and only add the electrum mapping specific for the  network

PR #827, opened 2019-05-09
Remove CF modal footer text

britttttk:
I'm not sure what PRS is. On the PoS modal footer this text does not appear. 

image


Kukks:
Oops

PR #831, opened 2019-05-11
Helpers for quick filtering of invoices

rockstardev:
Also includes:
-BOLT 11 display fix
-Moving update of confirmation count to Invoice Watcher

Here is how it looks in action

invoice filtering in action


rockstardev:
I finally understand why `BTCPayServer.Tests.StorageTests.CanUseAzureBlobStorage` test fails - it's because PRs are executed on my CircleCI account where connection string is missing. When user doesn't have CircleCI then PR is tested on BTCPayServer CircleCI account.

@Kukks how critical this test is? And can we disable it by default... like if it's executed on account with no connection string it skips it, otherwise it executes it?

NicolasDorier:
@rockstardev the idea is that I want to execute it is master... I am annoyed by those tests.

NicolasDorier:
Kukks tried to make the test pass if the setting is not set, but I don't want this because then we might end up on master with the tests passing because it is mis configured.

rockstardev:
@NicolasDorier the problem is that bash `if` check is being done in context of commit to forked repository. So check needs to be different... after change here is CircleCI result when I commit to forked repository (external integration skipped):

https://circleci.com/gh/rockstardev/btcpayserver/130

And then when I push that same commit to btcpayserver repo (external integration tests executed):

https://circleci.com/gh/btcpayserver/btcpayserver/1297

If we merge this - then this issue will finally go away... you can't test those external integrations if you don't have connection strings and we'll simply need to live with that - catch eventual errors when we merge to main repo. Last thing - we can alternatively also support checking if `user-secrets` is initialized on forked repository and then run external integrations tests if it is.

NicolasDorier:
@rockstardev I would merge anything to get this error away now.

PR #832, opened 2019-05-12
Selenium Tests

britttttk:
Add Chrome tests


NicolasDorier:
that is so cool, I will expand those tests to do way more browsing later!

PR #836, opened 2019-05-14
Import Trezor Wallet

Kukks:
This PR allows you to import your Trezor wallets into a BTCPay Store
It also moves the Ledger detection to a dialog for a more streamlined UI and to avoid device detection collision
image
image
image
image

Currently hardcoded to only use the WebUsb mode which means no firefox but as soon as https://github.com/trezor/trezord-go/issues/155 is implemented it should work all across.

image


NicolasDorier:
Ready to merge?

Kukks:
I added a warning so that users double-check first and moved the import dropdown to the side and converted into a text link so that users do not think it is the save button at first glance

image
image

I would merge now, support is a little flaky( webusb bugs out on some usb ports apparently) but its basics work.

NicolasDorier:
coooooooooool will review and test today :)

NicolasDorier:
Do we really need the non minified 56K LOC trezor file?

Kukks:
removed it and pushed the js code to https://github.com/btcpayserver/trezor-webpack-wrapper

NicolasDorier:
conflicts

Issue #838, opened 2019-05-14
Small Typo


mandelmonkey:
**Describe the problem/bug**
Small typo on the sync screen

"If you really don't want to synch and that you are familiar with command line, check FastSync."
should be sync?

**Your environment**
* Version of BTCPay Server: Latest
* Deployment method: Lunanode

Eskyee:
The spelling "synch" is a variant of "sync", and "sync" (synced) is the one mostly used. Both are the informal short form of synchronize (as in synchronization).

mandelmonkey:
ah thanks I didnt know that, I learn something new everyday

PR #842, opened 2019-05-15
make sync match FastSync

britttttk:
#838 

Is this better to remove/change synch? We do call it FastSync maybe we should match that style. Sync is more familiar to me, I've seen both.


britttttk:
Ok :) issue is closed, closing this too

PR #847, opened 2019-05-17
Make wallet able to send to multiple destinations

Kukks:


BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -215,17 +222,77 @@ public async Task<IActionResult> ListWallets()
...
+                var index = int.Parse(command.Substring(command.IndexOf(":",StringComparison.InvariantCultureIgnoreCase) + 1),  CultureInfo.InvariantCulture);
+                vm.Outputs.RemoveAt(index);
+                return View(vm);
+            }
+            
+
+            if (!vm.Outputs.Any())
+            {
+                ModelState.AddModelError(string.Empty,
+                    "Please add at least one transaction output");
```
NicolasDorier: you can early return here and remove the `else`.

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -215,17 +222,77 @@ public async Task<IActionResult> ListWallets()
...
+            }
+            else
+            {
+                var subtractFeesOutputsCount = new List<int>();
```
NicolasDorier: Should not use checkboxes but radio boxes! (Like I did in WalletSettings)

NicolasDorier:
On top of my comments. I think the Add destination should be next to the `Sign` button.
Also, if there is no multiple destination, the UI should be the same as before. 

That way, it does not scare noobs and you don't have to change the selenium tests.

BTCPayServer/Controllers/WalletsController.cs
NicolasDorier: Actually we can't use radio box I think. Radio box, at least one much be checked right?

Kukks: Yup, that's why I didn't use a radiobox

Kukks:
Done, the feature is very unintrusive :) 

image
image
image

NicolasDorier:
Can we even remove the group box if there is only one destination?

Kukks:
image
done

PR #848, opened 2019-05-18
Support temporary links for local file system provider

Kukks:


BTCPayServer/Storage/Services/Providers/FileSystemStorage/FileSystemFileProviderService.cs
```diff
@@ -60,7 +66,28 @@ public override async Task<string> GetFileUrl(StoredFile storedFile, StorageSett
...
+            var name = Guid.NewGuid().ToString();
+            var fullPath = Path.Combine(GetTempStorageDir(_Options), name);
+            if (!File.Exists(fullPath))
+            {
+                File.Create(fullPath).Dispose();
+            }
+            
+            await File.WriteAllTextAsync(Path.Combine(GetTempStorageDir(_Options), name), JsonConvert.SerializeObject(localFileDescriptor));
+            return
+                _HttpContextAccessor.HttpContext.Request.IsOnion()
```
NicolasDorier: Use `Request.GetAbsoluteUri` or something like this.

NicolasDorier:
Mmmh the FileProvider is broken

```
        public FileSystemFileProviderService(BTCPayServerEnvironment btcPayServerEnvironment,
            BTCPayServerOptions options, IHttpContextAccessor httpContextAccessor)
        {
            _BtcPayServerEnvironment = btcPayServerEnvironment;
            _Options = options;
            _HttpContextAccessor = httpContextAccessor;
        }
```

You can't do this on a singleton instance. Because a singleton outlive a Request, `IHttpContextAccessor ` will either point to the first request that the website created, or to null.

You need to refactor this. Either by not making those provider singletons, or by passing the url as parameter of methods of the providers.

Kukks:
Done,  passed down the baseurl

BTCPayServer/Storage/StorageExtensions.cs
```diff
@@ -28,17 +26,27 @@ public static void AddProviderStorage(this IServiceCollection serviceCollection)
...
+            } 
             else
             {
                 dirInfo = new DirectoryInfo(dir);
             }
 
+            DirectoryInfo tmpdirInfo;
+            if (!Directory.Exists(tmpdir))
+            {
+                tmpdirInfo = Directory.CreateDirectory(tmpdir);
```
NicolasDorier: can you try/catch that and log a warning if it fails? (and not register the `UseStaticFiles` if that's the case)

Some people have tight permission deployment, we don't want to crash them when they update.

Kukks: done

BTCPayServer/Storage/StorageExtensions.cs
```diff
@@ -25,55 +28,63 @@ public static void AddProviderStorage(this IServiceCollection serviceCollection)
...
                     }
-                }
-            });
-            
+                });
+            }
+            catch (Exception e)
+            {
+                Logs.Utils.LogError(e, $"Could not initialize the Local File Storage system( uploading and storing files locally)");
+                Console.WriteLine(e);
```
NicolasDorier: no Console.WriteLine

BTCPayServer/Storage/StorageExtensions.cs
```diff
@@ -25,55 +28,63 @@ public static void AddProviderStorage(this IServiceCollection serviceCollection)
...
+                        }
                     }
-                }
-            });
-            
+                });
+            }
+            catch (Exception e)
+            {
+                Logs.Utils.LogError(e, $"Could not initialize the Local File Storage system( uploading and storing files locally)");
```
NicolasDorier: Fix parentheses `(uploading and storing files locally)`

PR #850, opened 2019-05-21
BTCPay Abstractions: Move PaymentMethod specific logic to their handlers

Kukks:
Currently, there is a lot of logic across the entire codebase doing switch cases based on `PaymentTypes`  and applying specific logic. This pull request abstracts that logic with new method definitions to the `IPaymentMethodHandler` interface and removes those switches with the exception of some smaller legacy use-cases. 
This helps pave the way for making the central code leaner and more atomic which will allow us to move the payment methods to their own projects (`BtcPayServer.Payments.BtcLike`, `BtcPayServer.Payments.LightningLike`)


NicolasDorier:
I don't like the fact now you have to inject the payment handlers everywhere. This complicate the code more than anything.

Kukks:
Yeah, I need to figure out a better way for that for sure

NicolasDorier:
Can you break stuff down in smaller PRs? So at least I can easily review it and agree or not on specific things.

Kukks:
Will do, right now just experimenting and making sure all the tests pass with this approach

NicolasDorier:
I also don't like the Accounting part being integration test.

Kukks:
If I can figure out a better way to distribute the payment method handlers I think I'll be able to switch it back to Fast tests

Kukks:
A global static `Payment Method handler Registry` outside of the DI? Although it is a bit of an antipattern, it does clean up and simplify the code overall greatly

Kukks:
@NicolasDorier  I think this PR is an ok size now. 

* Tests are back to being in "Fast" mode
* There is a more minimal footprint of Payment Handler passing across
* All specific payment method logic is contained in their own payment handler now

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -98,44 +99,45 @@ public void CanHandleUriValidation()
...
+                    .SetCryptoPaymentData(new BitcoinLikePaymentData()
+                    {
+                        Output = new TxOut() {Value = Money.Coins(0.00151263m)}
+                    }));
             accounting = btc.Calculate();
-            invoiceEntity.Payments.Add(new PaymentEntity() { Accounted = true, CryptoCode = "BTC", NetworkFee = 0.00000100m }.SetCryptoPaymentData(new BitcoinLikePaymentData()
-            {
-                Output = new TxOut() { Value = accounting.Due }
-            }));
+            invoiceEntity.Payments.Add(
```
NicolasDorier: what's your problem with formatting, coding or drinking you need to chose.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -362,7 +367,7 @@ public bool IsExpired()
             return DateTimeOffset.UtcNow > ExpirationTime;
         }
 
-        public InvoiceResponse EntityToDTO(BTCPayNetworkProvider networkProvider)
```
NicolasDorier: ugh. I am not fan of the Task here. Thinking about what can be done....

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -33,6 +42,98 @@ class Prepare
...
+
+        public override string GetCryptoImage(PaymentMethodId paymentMethodId)
+        {
+            var network = _networkProvider.GetNetwork(paymentMethodId.CryptoCode);
+            return GetCryptoImage(network);
+        }
+        
+        private string GetCryptoImage(BTCPayNetwork network)
+        {
+            return _environment.Context.Request.GetRelativePathOrAbsolute(network.CryptoImagePath);
```
NicolasDorier: You can't do this. A handler is a singleton. `Context` is per request. If this code is not executed in a request, they you get an exception.... I think you can completely remove `GetCryptoImage`. The network is already providing the relative path, which is what we want.

Kukks: removed it

BTCPayServer.Tests/UnitTest1.cs
Kukks: lol what do you mean? I use the auto formatting in vs and rider so there should not be an issue

BTCPayServer/Services/Invoices/InvoiceEntity.cs
Kukks: I might not need it to be async, let me make it sync

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -101,12 +105,15 @@ private InvoiceDetailsModel InvoicePopulatePayments(InvoiceEntity invoice)
...
                 var cryptoPayment = new InvoiceDetailsModel.CryptoPayment();
-                cryptoPayment.PaymentMethod = ToString(paymentMethodId);
+                cryptoPayment.PaymentMethod = _paymentMethodHandlers
+                    .GetCorrectHandler(paymentMethodId)
+                    .ToPrettyString(data.GetId());
                 cryptoPayment.Due = _CurrencyNameTable.DisplayFormatCurrency(accounting.Due.ToDecimal(MoneyUnit.BTC), paymentMethodId.CryptoCode);
                 cryptoPayment.Paid = _CurrencyNameTable.DisplayFormatCurrency(accounting.CryptoPaid.ToDecimal(MoneyUnit.BTC), paymentMethodId.CryptoCode);
                 cryptoPayment.Overpaid = _CurrencyNameTable.DisplayFormatCurrency(accounting.OverpaidHelper.ToDecimal(MoneyUnit.BTC), paymentMethodId.CryptoCode);
-
+//TODO: abstract
```
NicolasDorier: wat?

Kukks: marker for my next pull request :P

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -14,6 +15,7 @@
 using BTCPayServer.Payments;
 using NBitpayClient;
 using BTCPayServer.Payments.Bitcoin;
+using NUglify.Helpers;
```
NicolasDorier: what is this?

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -607,7 +610,8 @@ public async Task<PaymentEntity> AddPayment(string invoiceId, DateTimeOffset dat
                 var invoice = context.Invoices.Find(invoiceId);
                 if (invoice == null)
                     return null;
-                InvoiceEntity invoiceEntity = ToObject(invoice.Blob);
+                InvoiceEntity invoiceEntity = ToObject<InvoiceEntity>(invoice.Blob, network.NBitcoinNetwork);
```
NicolasDorier: Keep `ToObject`, it has different meaning than `ToObject<InvoiceEntity>`

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -397,13 +399,14 @@ public async Task<InvoiceEntity> GetInvoice(string id, bool inludeAddressData =
         private InvoiceEntity ToEntity(Data.InvoiceData invoice)
         {
             var entity = ToObject(invoice.Blob);
+            entity.PaymentMethodHandlers = _paymentMethodHandlers;
```
NicolasDorier: Don't assign here, assign inside the `ToObject`

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -818,23 +817,26 @@ public PaymentMethodAccounting Calculate(Func<PaymentEntity, bool> paymentPredic
             bool paidEnough = paid >= Extensions.RoundUp(totalDue, precision);
             int txRequired = 0;
 
-            ParentEntity.GetPayments()
+            var _ = ParentEntity.GetPayments()
```
NicolasDorier: you can remove `var`, `_` is actually a keyword.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -818,23 +817,26 @@ public PaymentMethodAccounting Calculate(Func<PaymentEntity, bool> paymentPredic
...
                     }
+
                     paidEnough |= Extensions.RoundUp(paid, precision) >= Extensions.RoundUp(totalDue, precision);
                     if (GetId() == _.GetPaymentMethodId())
                     {
                         cryptoPaid += _.GetCryptoPaymentData().GetValue();
                         txRequired++;
                     }
+
+                    return _;
```
NicolasDorier: if you don't call `ToArray`, this never get evaluated.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
Kukks: removed it

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -334,35 +316,41 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
...
+                                                  kv.ParentEntity.PaymentMethodHandlers.GetCorrectHandler(
+                                                      availableCryptoPaymentMethodId);
+                                              return new PaymentModel.AvailableCrypto()
+                                              {
+                                                  PaymentMethodId = kv.GetId().ToString(),
+                                                  CryptoCode = kv.GetId().CryptoCode,
+                                                  PaymentMethodName = availableCryptoHandler.GetPaymentMethodName(availableCryptoPaymentMethodId),
+                                                  IsLightning =
+                                                      kv.GetId().PaymentType == PaymentTypes.LightningLike,
+                                                  CryptoImage = Request.GetAbsoluteUri(availableCryptoHandler.GetCryptoImage(availableCryptoPaymentMethodId)),
```
NicolasDorier: `GetRelativePathOrAbsolute`

NicolasDorier:
You can remove the `IEnumerable<PaymentHandler>` from the InvoiceController. By encapsulating the `new InvoiceEntity()` inside the `InvoiceRepository` with a new method `InvoiceRepository.CreateNewInvoice()`.

Kukks:
There is still one use of it  for populating the select in the create invoice ui. Otherwise it seems good!

BTCPayServer/Payments/IPaymentMethodHandler.cs
```diff
@@ -32,8 +37,25 @@ public interface IPaymentMethodHandler
...
+
+        void PreparePaymentModel(PaymentModel model, InvoiceResponse invoiceResponse);
+        string GetCryptoImage(PaymentMethodId paymentMethodId);
+        string GetPaymentMethodName(PaymentMethodId paymentMethodId);
+
+        Task<string> IsPaymentMethodAllowedBasedOnInvoiceAmount(StoreBlob storeBlob,
+            Dictionary<CurrencyPair, Task<RateResult>> rate,
+            Money amount, PaymentMethodId paymentMethodId);
+        
+        IEnumerable<PaymentMethodId> GetSupportedPaymentMethods();
```
NicolasDorier: This and `CanHandle` are basically the same thing. An implementer should not have to implement both.

NicolasDorier: I would say this:

Create a `PaymentMethodHandlerCollection` type that replace `IEnumerable<IPaymentMethodHandler>` of now.
In this new collection, add the `GetCorrectHandler` that is currently just a extension method and remove the extension method.

You can then remove `CanHandle` from here this interface, and create a map in `PaymentMethodHandlerCollection` which map `PaymentMethodId => IPaymentHandler`.

`GetCorrectHandler` can then just use this map to fetch the correct handler.

Kukks: Ok I removed the CanHandle situation and wrapped in a custom type with an index. Looks a lot nicer now

NicolasDorier:
Other question: Does it makes sense for the InvoiceEntity to get the singleton dictionary? Should it not be a subset of this dictionary which filter only the handlers that are actually matching the supported payments?

Kukks:
> Other question: Does it makes sense for the InvoiceEntity to get the singleton dictionary? Should it not be a subset of this dictionary which filter only the handlers that are actually matching the supported payments?

It would, but we only find out what payment methods are enabled on an invoice AFTER we get the entity from the repository( at least in terms of creation). With that in mind, I would not make the instantiation different depending on if the invoice is being created or fetched for consistency's sake.

NicolasDorier:
Well one idea is to not affect the dictionnary in the `NewInvoiceEntity` and only do it when we save it in DB. Meh does not matter too much...

Kukks:
Yeah, I would not focus on that. Are we good to merge, because I have some more stuff to add which should probably be in another PR

NicolasDorier:
need to kick in the box for the tests to pass

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -27,6 +27,7 @@
 using NBitpayClient;
 using NBXplorer;
 using Newtonsoft.Json.Linq;
+using NUglify.Helpers;
```
NicolasDorier: what is nuglify?

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -95,6 +95,7 @@ public override string GetPaymentMethodName(PaymentMethodId paymentMethodId)
         public override IEnumerable<PaymentMethodId> GetSupportedPaymentMethods()
         {
             return _networkProvider.GetAll()
+                .Where(network => network.NBXplorerNetwork != null)
```
NicolasDorier: `== null` can't happen.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -137,10 +137,12 @@ public async Task TestConnection(NodeInfo nodeInfo, CancellationToken cancellati
                 throw new PaymentMethodUnavailableException($"Error while connecting to the lightning node via {nodeInfo.Host}:{nodeInfo.Port} ({ex.Message})");
             }
         }
-        
+
         public override IEnumerable<PaymentMethodId> GetSupportedPaymentMethods()
         {
-            return _networkProvider.GetAll().Select(network => new PaymentMethodId(network.CryptoCode, PaymentTypes.LightningLike));
+            return _networkProvider.GetAll()
+                .Where(network => network.NBXplorerNetwork != null)
```
NicolasDorier: `== null` can't happen.

BTCPayServer/Controllers/InvoiceController.UI.cs
Kukks: I'm not sure but it has an extension method available  that I like to use on IEnumerables: `ForEach`

Kukks: Don't know which package added NUglify, probably the bundle minifier

Kukks:
@NicolasDorier  mergeable now? :D

BTCPayServer/Controllers/InvoiceController.API.cs
```diff
@@ -20,14 +22,17 @@ public class InvoiceControllerAPI : Controller
         private InvoiceController _InvoiceController;
         private InvoiceRepository _InvoiceRepository;
         private BTCPayNetworkProvider _NetworkProvider;
+        private readonly IEnumerable<IPaymentMethodHandler> _paymentMethodHandlers;
 
         public InvoiceControllerAPI(InvoiceController invoiceController,
                                     InvoiceRepository invoceRepository,
-                                    BTCPayNetworkProvider networkProvider)
+                                    BTCPayNetworkProvider networkProvider,
+                                    IEnumerable<IPaymentMethodHandler> paymentMethodHandlers)
```
NicolasDorier: Why not the dictionary?

NicolasDorier: actually you may be able to remove it

Kukks: removed it

PR #851, opened 2019-05-22
small cleanup

Kukks:
Just a small subset of #850

### DIFF
--- a/BTCPayServer/Hosting/BTCPayServerServices.cs
+++ b/BTCPayServer/Hosting/BTCPayServerServices.cs
@@ -38,6 +38,8 @@ using BTCPayServer.Logging;
 using BTCPayServer.HostedServices;
 using System.Security.Claims;
 using BTCPayServer.PaymentRequest;
+using BTCPayServer.Payments;
+using BTCPayServer.Payments.Bitcoin;
 using BTCPayServer.Payments.Changelly;
 using BTCPayServer.Payments.Lightning;
 using BTCPayServer.Security;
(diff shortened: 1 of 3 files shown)


BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -177,15 +179,18 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
...
+            services.AddSingleton<BitcoinLikePaymentHandler>();
+            services.AddSingleton<IPaymentMethodHandler>(provider => provider.GetService<BitcoinLikePaymentHandler>());
+            services.AddSingleton<IPaymentMethodHandler<DerivationSchemeSettings>>(provider => provider.GetService<BitcoinLikePaymentHandler>());
+            services.AddSingleton<IHostedService, NBXplorerListener>();
 
-            services.AddSingleton<Payments.IPaymentMethodHandler<Payments.Lightning.LightningSupportedPaymentMethod>, Payments.Lightning.LightningLikePaymentHandler>();
             services.AddSingleton<LightningLikePaymentHandler>();
-            services.AddSingleton<IHostedService, Payments.Lightning.LightningListener>();
-            
+            services.AddSingleton<IPaymentMethodHandler>(provider => provider.GetService<LightningLikePaymentHandler>());
```
NicolasDorier: why??

Kukks: Before, we were creating the duplicate instances of the same service( under `LightningLikePaymentHandler`  and `IPaymentMethodHandler<LightningSupportedPaymentMethod>`). 
This makes them all use the same one while also registering them against the main interface `IPaymentMethodHandler`

This is for my next pull requests (starting at #852 )

NicolasDorier: If there is several singleton using `IPaymentMethodHandler`, then controllers can resolve `IEnumerable<IPaymentMethodHandler>` ?

Kukks: Yes

PR #852, opened 2019-05-22
Provide Pretty descriptions for payment methods from their handlers

Kukks:
Just a small subset of #850 
Builds on top of #851

### DIFF
--- a/BTCPayServer/Extensions.cs
+++ b/BTCPayServer/Extensions.cs
@@ -39,6 +39,11 @@ namespace BTCPayServer
 {
     public static class Extensions
     {
+        public static IPaymentMethodHandler GetCorrectHandler(
+            this IEnumerable<IPaymentMethodHandler> paymentMethodHandlers, PaymentMethodId paymentMethodId)
+        {
+            return paymentMethodHandlers.First(handler => handler.CanHandle(paymentMethodId));
+        }
         public static string PrettyPrint(this TimeSpan expiration)
         {
             StringBuilder builder = new StringBuilder();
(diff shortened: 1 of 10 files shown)


BTCPayServer/Extensions.cs
```diff
@@ -39,6 +39,11 @@ namespace BTCPayServer
 {
     public static class Extensions
     {
+        public static IPaymentMethodHandler GetCorrectHandler(
+            this IEnumerable<IPaymentMethodHandler> paymentMethodHandlers, PaymentMethodId paymentMethodId)
+        {
+            return paymentMethodHandlers.First(handler => handler.CanHandle(paymentMethodId));
```
NicolasDorier: `FirstOrDefault` better?

BTCPayServer/Payments/PaymentMethodId.cs
```diff
@@ -62,22 +62,9 @@ public override int GetHashCode()
...
         {
-            return ToString(false);
+            return PaymentType == PaymentTypes.BTCLike ? CryptoCode : $"{CryptoCode}_{PaymentType}";
         }
 
-        public string ToString(bool pretty)
-        {
-            if (pretty)
-            {
-                return $"{CryptoCode} ({PrettyMethod(PaymentType)})";
```
NicolasDorier: If not use, can delete `PrettyMethod` as well below.

BTCPayServer/Extensions.cs
Kukks: I want it to intentionally throw since any situation where it is null is probably a bug in the code

BTCPayServer/Payments/PaymentMethodId.cs
Kukks: removed

PR #866, opened 2019-05-29
 List Cards Grid

nakamotones:
ListApps, ListStores, ListWallets as Cards/Grid layout


NicolasDorier:
You need to rebase on `master`, also the tests don't pass you may have break something.

nakamotones:
Got it. Will do.

NicolasDorier:
Need to fix test `CanManageWallet`

nakamotones:
Selenium Test fails because i's looking for string "Manage", however I changed it to "Manage Wallet" in ListWallets.chtml `<a class="" asp-action="WalletTransactions" asp-route-walletId="@wallet.Id">Manage Wallet</a>`
Question is, do we keep "Manage Wallet"? Or set it back to just "Wallet"? 
Screenshot_2019-05-29 BTCPay Server

britttttk:
@nakamotones As long as it's not "wallets" which selenium finds in the main header, just replace the link text with what you choose. Personally I think that Manage Wallet is fine. :)

PR #869, opened 2019-05-29
Update explorer to blockstream.info

grubles:
Title is pretty self-explanatory.

### DIFF
--- a/BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
+++ b/BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
@@ -16,7 +16,7 @@ namespace BTCPayServer
             {
                 CryptoCode = nbxplorerNetwork.CryptoCode,
                 DisplayName = "Bitcoin",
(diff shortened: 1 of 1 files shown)


Kukks:
What's the issue with the current explorer?

NicolasDorier:
damn why this test is still running when not in master? @rockstardev any idea? I thought it was fixed.

rockstardev:
@NicolasDorier we didn't implement that @Kukks PR fix... can do it now.

NicolasDorier:
Personally I prefer blockstream.info, that said maybe a better idea is to all customization at the user's setting level.

grubles:
@Kukks Blockstream releases the code for blockstream.info, unlike Smartbit (AFAICT, at least). https://github.com/Blockstream/esplora

Another notable difference is Blockstream.info supports browsing javascript-less via tor and through tor hidden services.

britttttk:
ack - translated blockchain explorer

NicolasDorier:
Make me think that later we should use the onion link if BTCPay is used on Tor Browser instead of the clearnet link.

PR #876, opened 2019-06-03
New login page

NicolasDorier:
image

PR #877, opened 2019-06-03
New register form

NicolasDorier:
Related PR about login: https://github.com/btcpayserver/btcpayserver/pull/876

After:
image

Before:
image

### DIFF
--- a/BTCPayServer/Views/Account/Register.cshtml
+++ b/BTCPayServer/Views/Account/Register.cshtml
@@ -4,47 +4,54 @@
 }
 
 <section>
-    <div class="container">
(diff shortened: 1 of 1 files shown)


pavlenex:
What exactly is `is administrator`?
The tickbox should be inline before `is administrator anyway`
is administrator

NicolasDorier:
Note: The "Is administrator" only appear in tests.

pavlenex:
Yeah was a bit worried about that. In that case tACK for this one, looks nice.

NicolasDorier:
I inlined

image

PR #854, opened 2019-05-23
added View App button to POS settings

nakamotones:

PR #879, opened 2019-06-04
added buttons to POS settings

nakamotones:
updated POS settings form group buttons to match Crowdfund settings from group buttons, addressing #854

### DIFF
--- a/BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
@@ -60,9 +60,6 @@ namespace BTCPayServer.Models.AppViewModels
         [Display(Name = "Redirect invoice to redirect url automatically after paid")]
         public string RedirectAutomatically { get; set; } = string.Empty;
 
-        public string AppId { get; set; }
-        public string SearchTerm { get; set; }
-
         public SelectList RedirectAutomaticallySelectList =>
             new SelectList(new List< SelectListItem>()
             {
--- a/BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
@@ -60,6 +60,9 @@ namespace BTCPayServer.Models.AppViewModels
         [Display(Name = "Redirect invoice to redirect url automatically after paid")]
         public string RedirectAutomatically { get; set; } = string.Empty;
 
+        public string AppId { get; set; }
+        public string SearchTerm { get; set; }
+
         public SelectList RedirectAutomaticallySelectList =>
             new SelectList(new List< SelectListItem>()
             {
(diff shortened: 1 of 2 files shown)


BTCPayServer/Models/AppViewModels/UpdatePointOfSaleViewModel.cs
```diff
@@ -60,6 +60,9 @@ public class UpdatePointOfSaleViewModel
         [Display(Name = "Redirect invoice to redirect url automatically after paid")]
         public string RedirectAutomatically { get; set; } = string.Empty;
 
+        public string AppId { get; set; }
+        public string SearchTerm { get; set; }
```
NicolasDorier: How can this work? You are not affecting them any value.

nakamotones: Build failed with out it because the form group used  `asp-route-searchterm="@Model.SearchTerm"` and `asp-route-appId="@Model.Id" id="ViewApp"` in UpdatePointOfSale.cshtml.
AppId and SearchTerm are not needed for buttons to work: removed from both files.

NicolasDorier: I doubt it works, both are needed, but should be initialized in the code.

nakamotones: > I doubt it works, both are needed, but should be initialized in the code.

It's working.  What should be initialized and why?

NicolasDorier:
```
<a class="btn btn-secondary" target="_blank" asp-action="ListInvoices" asp-controller="Invoice" >Invoices</a>
                        <a class="btn btn-secondary" target="_blank" asp-action="ViewPointOfSale" asp-controller="AppsPublic" id="ViewApp">View App</a>
```

1. You are basically asking to open the Invoices page without filter. While what is needed is opening the page with a filter to find invoices related to this app.
2. In ViewApp, you are not passing the AppId, so how is it possible that the link open the right app?

nakamotones: Got it! Will test a few days before PR.
We define   `public string AppId { get; set; }
        public string SearchTerm { get; set; }`  within UpdatePointOfSaleViewmodel.cs and added GetPointOfSaleOrderId to AppService.cs  so AppsController.PointOfSale.cs  knows what we're talking about when we use the variables AppId and SearchTerm and pass them through `                         <a class="btn btn-secondary" target="_blank" asp-action="ListInvoices" asp-controller="Invoice" asp-route-searchterm="@Model.SearchTerm">Invoices</a>
                        <a class="btn btn-secondary" target="_blank" asp-action="ViewPointOfSale" asp-controller="AppsPublic" asp-route-appId="@Model.Id" id="ViewApp">View App</a>
` 
Now the PoS settings buttons behave like the CF settings buttons, as per @britttttk request!
Thanks for the lesson @NicolasDorier!

NicolasDorier: Yes, and `SearchTerm` and `Id` in the case of PointOfSale are filled by the controller.

PR #891, opened 2019-06-18
Allow language parameter in pay button endpoint

Kukks:


BTCPayServer/Controllers/PublicController.cs
```diff
@@ -58,7 +59,15 @@ public async Task<IActionResult> PayButtonHandle([FromForm]PayButtonViewModel mo
                 RedirectURL = model.BrowserRedirect,
                 FullNotifications = true
             }, store, HttpContext.Request.GetAbsoluteRoot(), cancellationToken: cancellationToken);
-            return Redirect(invoice.Data.Url);
+            if (string.IsNullOrEmpty(model.Language))
+            {
+                return Redirect(invoice.Data.Url);
+            }
+            var uriBuilder = new UriBuilder(invoice.Data.Url);
+            var paramValues = HttpUtility.ParseQueryString(uriBuilder.Query);
```
NicolasDorier: don't use this, just add a `string lang = null` parameter in `PayButtonHandle`

BTCPayServer/Controllers/PublicController.cs
```diff
@@ -58,7 +59,15 @@ public async Task<IActionResult> PayButtonHandle([FromForm]PayButtonViewModel mo
...
+            paramValues.Add("lang", model.Language);
+            uriBuilder.Query = paramValues.ToString();
```
NicolasDorier: I doubt it serialize the dictionarry as you expect.

NicolasDorier: it does

NicolasDorier:
What about allowing arbitrary paramters? Instead of `Language` you would have `QueryString`.
And you would just merge this query string to the other one from the invoice?

Kukks:
I can replace Language with Dictionary<string, string> CheckoutAdditionalParameters? and then  you can do <input type='hidden' name='checkoutAdditionalParameters[lang]' value='da-DK'>
this should work as FormData allows these kinds of structures

NicolasDorier:
@Kukks I think `checkoutQueryString` is enough.

PR #898, opened 2019-06-25
Multiple domains for apps in BTCPay 

### LINKED ISSUE #887: Multiple domains for apps in BTCPay
In #828 we added the ability for app to be on a website root. 

Next, we should allow multiple domains for apps so one btcpay instance can have apps on app 1 on domain1.com different app on app 2 on domain2.com etc.

Kukks:
closes #887

### DIFF
--- a/BTCPayServer/Controllers/ServerController.cs
+++ b/BTCPayServer/Controllers/ServerController.cs
@@ -461,7 +461,7 @@ namespace BTCPayServer.Controllers
         public async Task<IActionResult> Policies()
         {
             var data = (await _SettingsRepository.GetSettingAsync<PoliciesSettings>()) ?? new PoliciesSettings();
-            await GetAppSelectList();
+            ViewBag.AppsList = await GetAppSelectList();
             return View(data);
         }
 
@@ -469,7 +469,7 @@ namespace BTCPayServer.Controllers
         [HttpPost]
         public async Task<IActionResult> Policies(PoliciesSettings settings, string command = "")
         {
-            await GetAppSelectList();
+            ViewBag.AppsList = await GetAppSelectList();
             if (command == "add-domain")
             {
                 ModelState.Clear();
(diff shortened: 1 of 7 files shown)


BTCPayServer/Controllers/ServerController.cs
```diff
@@ -459,43 +461,61 @@ public string StatusMessage
...
-                var selectList = ctx.Users.Where(user => user.Id == userId)
-                                .SelectMany(s => s.UserStores)
-                                .Select(s => s.StoreData)
-                                .SelectMany(s => s.Apps)
-                                .Select(a => new SelectListItem($"{a.AppType} - {a.Name}", a.Id)).ToList();
-                selectList.Insert(0, new SelectListItem("(None)", null));
-                ViewBag.AppsList = new SelectList(selectList, "Value", "Text", data.RootAppId);
-            }
-
+            await GetAppSelectList();
```
NicolasDorier: Do `ViewBag.AppsList = GetAppSelectList();` this is horrible like this.

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -459,43 +461,61 @@ public string StatusMessage
...
-                        settings.RootAppType = null;
-                }
+                ModelState.Clear();
+                settings.DomainToAppMapping.Add(new PoliciesSettings.DomainToAppMappingItem());
+                return View(settings);
+            }
+            if (command.StartsWith("remove-domain", StringComparison.InvariantCultureIgnoreCase))
+            {
+                ModelState.Clear();
+                var index = int.Parse(command.Substring(command.IndexOf(":",StringComparison.InvariantCultureIgnoreCase) + 1),  CultureInfo.InvariantCulture);
```
NicolasDorier: don't do hack like this, just add a new parameter to `Policies` like `selectedAppId` which is somehow passed via `input hidden` to the submitted form.

Kukks: It is not that easy you need to have a form for each item which is not possible since there is a main form for all the fields( and you cannot have a form in form.)

We use this same logic in the wallet send UI too fro multiple destinations

PR #905, opened 2019-07-02
Paging on List Users page for big hosting providers like Esky

rockstardev:
It should help @Eskyee with identifying exactly how much users it currently hosts.

Here is paging in action:
Screenshot_2019-07-02 BTCPay Server


Eskyee:
Cheers 🍻 my man 🤗🙌🌟

PR #911, opened 2019-07-09
Allow apps server admin does not have access to to be mapped in domain-to-app policy settings

Kukks:
This is a bugfix imo:
If you have 2 server admins, where only Admin1 has access to a specific app and Admin2 goes to edit the policies, the mapping would be overwritten to "None" .
This PR exposes all apps on all stores to this mapping dropdown irrespective of user. I think it makes sense to leak this minor data here to server admins.

### DIFF
--- a/BTCPayServer/Controllers/ServerController.cs
+++ b/BTCPayServer/Controllers/ServerController.cs
@@ -579,18 +579,10 @@ namespace BTCPayServer.Controllers
 
         private async Task<List<SelectListItem>> GetAppSelectList()
         {
-            // load display app dropdown
-            using (var ctx = _ContextFactory.CreateContext())
-            {
-                var userId = _UserManager.GetUserId(base.User);
-                var selectList = await ctx.Users.Where(user => user.Id == userId)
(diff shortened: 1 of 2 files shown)


NicolasDorier:
@Kukks then it should not be a dropdown list, because if there is 99999999999999999 apps then it will be problematic.

Maybe should remove the dropdown in favor of showing the appid.

Kukks:
That would be bad UX imo, It is only 1 call for all the app dropdowns in that one specific page that should only be loaded once in a blue moon

I don't think it should be problematic even if there are 10000 apps on one instance

NicolasDorier:
@Kukks well now you have the problem is somebody is calling the app with the same name as yours.

Kukks:
I can add the name of the user alongside the app name?

NicolasDorier:
You can't an app is not attached to a user.

Kukks:
I've also added the store name to the display now. I dont think this is something we should worry about much tbh

NicolasDorier:
there is a conflict

Kukks:
updated

PR #912, opened 2019-07-10
Refactor Network Provider to add networks through interfaces 

Kukks:
Still a bit rough; this is a first step to eventually allow the possibility to use DI to add networks( and make the shitcoins code be in a separate dll/project). There's a static factory method so far that replaces the tricky parts of the code where you create new instances of the network provider whichI hope to clear out later on

### DIFF
--- a/BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
+++ b/BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
@@ -7,39 +7,55 @@ using NBXplorer;
 
 namespace BTCPayServer
 {
-    public partial class BitcoinBTCPayNetworkProvider
+    public class BitcoinBTCPayNetworkProvider : IBTCPayNetworkProvider
     {
-        public void InitBitcoin()
+        public IEnumerable<BTCPayNetworkBase> GetNetworks(NetworkType networkType)
         {
-            var nbxplorerNetwork = NBXplorerNetworkProvider.GetFromCryptoCode("BTC");
-            Add(new BTCPayNetwork()
+            var nbxplorerNetwork = GetNBXplorerNetworkProvider(networkType).GetFromCryptoCode("BTC");
+            return new[]
             {
-                CryptoCode = nbxplorerNetwork.CryptoCode,
-                DisplayName = "Bitcoin",
-                BlockExplorerLink = NetworkType == NetworkType.Mainnet ? "https://blockstream.info/tx/{0}" : "https://blockstream.info/testnet/tx/{0}",
-                NBitcoinNetwork = nbxplorerNetwork.NBitcoinNetwork,
-                NBXplorerNetwork = nbxplorerNetwork,
-                UriScheme = "bitcoin",
-                CryptoImagePath = "imlegacy/bitcoin.svg",
-                LightningImagePath = "imlegacy/bitcoin-lightning.svg",
-                DefaultSettings = BTCPayDefaultSettings.GetDefaultSettings(NetworkType),
-                CoinType = NetworkType == NetworkType.Mainnet ? new KeyPath("0'") : new KeyPath("1'"),
-                SupportRBF = true,
-                //https://github.com/spesmilo/electrum/blob/11733d6bc271646a00b69ff07657119598874da4/electrum/constants.py
-                ElectrumMapping = NetworkType == NetworkType.Mainnet
-                    ? new Dictionary<uint, DerivationType>()
-                    {
-                        {0x0488b21eU, DerivationType.Legacy }, // xpub
-                        {0x049d7cb2U, DerivationType.SegwitP2SH }, // ypub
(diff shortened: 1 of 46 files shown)


NicolasDorier:
@Kukks you need to keep the "unfiltered".

I got lot's of bug provoked for people who deactivated network afterward. That's why I need a way to get all the possible networks.

Kukks:
I got rid of the concept of storing a filtered instance of the provider. Instead, I changed the `Filter` method to return the list of networks that match the criteria. Need to test it out better first though :)

NicolasDorier:
Don't rename to `BitcoinBTCPayNetworkProvider`

NicolasDorier:
Frankly I prefer you don't touch this if not needed. I removed many bugs after the last refactoring.

BTCPayServer.Common/BTCPayNetworkProvider.cs
```diff
@@ -8,45 +8,106 @@
 
 namespace BTCPayServer
 {
-    public partial class BTCPayNetworkProvider
+    public static class BTCPayNetworkProviderFactory
     {
-        Dictionary<string, BTCPayNetworkBase> _Networks = new Dictionary<string, BTCPayNetworkBase>();
+        public static BTCPayNetworkProvider Instance { get; set; }
```
NicolasDorier: If it is a singleton, it's initialization must not depend on anything, so you need to init it here. (Via `Lazy<>` or normal singleton pattern).

It should not be instantiated outside.

BTCPayServer.Common/BTCPayNetworkProvider.cs
```diff
@@ -8,45 +8,106 @@
...
 
+        public static BTCPayNetworkProvider GetProvider(NetworkType type)
+        {
+            return Instance?.NetworkType == type ? Instance : new BTCPayNetworkProvider(GetDefaultNetworkProviders(), type);
```
NicolasDorier: Instance should never be able to be null

BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
```diff
@@ -7,39 +7,55 @@
 
 namespace BTCPayServer
 {
-    public partial class BTCPayNetworkProvider
+    public class BitcoinBTCPayNetworkProvider : IBTCPayNetworkProvider
```
NicolasDorier: Maybe `BitcoinLikeNetworkProvider` ?

BTCPayServer/Data/StoreData.cs
```diff
@@ -400,13 +399,13 @@ public string LightningDescriptionTemplate
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
         public double PaymentTolerance { get; set; }
 
-        public BTCPayServer.Rating.RateRules GetRateRules(BTCPayNetworkProvider networkProvider)
+        public BTCPayServer.Rating.RateRules GetRateRules(IEnumerable<BTCPayNetworkBase> networks)
```
NicolasDorier: I am not fan of having a `IEnumerable<>` instead of a class like `BTCPayNetworkProvider`.

With a class like `BTCPayNetworkProvider`, we can add helper methods like `BTCPayNetworkProvider.GetFromCryptoCode(cryptoCode)` or more. This prevent to scan all the list everytimes.

BTCPayServer/Services/Rates/CurrencyNameTable.cs
```diff
@@ -94,7 +94,7 @@ public IFormatProvider GetCurrencyProvider(string currency)
                         catch { }
                     }
 
-                    foreach (var network in new BTCPayNetworkProvider(NetworkType.Mainnet).GetAll())
+                    foreach (var network in BTCPayNetworkProviderFactory.Instance[NetworkType.Mainnet].GetAll())
```
Kukks: I tried to refactor the CurrencyNameTable to use DI instead of the current static way but it was too intertwined with everything

NicolasDorier:
It is not clear to me that places where you removed references to `UnfilteredNetworks` are now passed all the networks, and not only the one registered to DI.

NicolasDorier:
I am really not fan, we are now passing 2 types of NetworkProvider around (AvailableNetworkProvider and the unfiltered NetworkProvider), it makes it easy to should yourself in the foot by passing one instead of the other.

I am also unsure about the goal on registering shitcoins via DI, because all shitcoins need to be registered whatever the configuration. Else we will end up in a situation when things start crashing if a user disable a shitcoin after initially supporting it. (Your previous refactor on NetworkProvider broke lot's of things that way)

NicolasDorier:
I agree with the part of making singleton out of the unfiltered networks.

Kukks:
What should i do to this PR @NicolasDorier  Not sure what the next step is?

Kukks:
Im going to kill this PR since it's so dated now which is a shame.  I really hoped we could move out altcoin stuff to a separate project.

PR #920, opened 2019-07-15
"Manual" Payment Method

Kukks:
This PR introduces various small refactorings around payment methods along with introducing a new payment type called "manual". This payment type can be enabled and configured from the store settings and should enable a number of new use-cases.

Some of these use-cases:

- Accepting cash, cheques, gift vouchers in a physical store
- Accepting wire transfer, local banking quick transfer apps, cash, cheques, gift vouchers sent by mail in an online store
- Accepting a split payment of cash and bitcoin 

Configuration of this new payment type is flexible yet simple enough:

image

Checkout UI: 
image

image

When the setting "Mark payment as confirmed instantly (instead of Paid)" is disabled, the invoice is marked as paid when a manual payment is added, until a user with access to the store confirms it. 
image

Additional screenshots:
image


NicolasDorier:
Need test

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -174,11 +174,15 @@ public async Task<InvoiceEntity> CreateInvoiceAsync(string storeId, InvoiceEntit
 
                 foreach (var paymentMethod in invoice.GetPaymentMethods())
                 {
-                    if (paymentMethod.Network == null)
+                    if (paymentMethod.Network == null )
```
NicolasDorier: why a space here

BTCPayServer/Services/Rates/CurrencyNameTable.cs
```diff
@@ -142,7 +142,7 @@ public string DisplayFormatCurrency(decimal value, string currency, bool threeLe
             else
                 return value.ToString("C", provider) + $" ({currency})";
         }
-
+
```
NicolasDorier: why

NicolasDorier:
mmmh this is super error prone to have an "empty value" cryptocode. I am wondering if the best is not to move `CryptoCode` out of BTCPayNetworkBase and into BTCPayNetwork. This would make more sense.

NicolasDorier:
Ready for another round of review?

Kukks:
I would say so @NicolasDorier  :)

Kukks:
Good to note that the empty cryptoCode refactoring should be done in a separate PR( and not in #912 but after as it still does not fit with this specific refactoring)

NicolasDorier:
So replace  `string.Empty` with a constant `const string ManualCryptoCode = String.Empty.`, so we can easily cleanup by removing the constant later.

BTCPayServer/Payments/PaymentMethodId.cs
```diff
@@ -76,15 +76,22 @@ public string ToPrettyString()
         public static bool TryParse(string str, out PaymentMethodId paymentMethodId)
         {
             paymentMethodId = null;
-            var parts = str.Split('_', StringSplitOptions.RemoveEmptyEntries);
```
NicolasDorier: why do you modify this code?

BTCPayServer/Views/Invoice/InvoicePaymentsPartial.cshtml
```diff
@@ -1,101 +1,49 @@
...
-        </div>
-    </div>
+@{
+    var grouped = Model.Payments.GroupBy(payment => payment.GetPaymentMethodId().PaymentType);
+    
 }
-@if (Model.OffChainPayments.Count > 0)
+@foreach (var paymentGroup in grouped)
 {
-    <div class="row">
```
NicolasDorier: I think this complicate things for nothing. The partial is not really needed as there is no reuse.

BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -247,11 +248,23 @@ private async Task<PaymentMethod> CreatePaymentMethodAsync(Dictionary<CurrencyPa
                 var logPrefix = $"{supportedPaymentMethod.PaymentId.ToPrettyString()}:";
                 var storeBlob = store.GetStoreBlob();
                 var preparePayment = handler.PreparePayment(supportedPaymentMethod, store, network);
-                var rate = await fetchingByCurrencyPair[new CurrencyPair(network.CryptoCode, entity.ProductInformation.Currency)];
-                if (rate.BidAsk == null)
+                RateResult rate;
+                if (network.CryptoCode == BTCPayNetworkProvider.ManualCryptoCode)
```
NicolasDorier: Should be unecessary, `network,CryptoCode == entity.ProductInfo.Currency` I should return 1 already

Kukks: Necessary  as `network.CryptoCode` = "" so `network,CryptoCode != entity.ProductInfo.Currency`

BTCPayServer/Views/Invoice/InvoicePaymentsPartial.cshtml
Kukks: I honestly think  it cleans up the code clutter in this section by offloading them to individual sections but I can move them

NicolasDorier: Maybe we can do later, but this PR is already huge and super painful to review. Make it as small as possible so we can move on.

NicolasDorier:
1. I think that the manual payment should NOT have its own "network". This is not a network, this is a payment type.
2. The CryptoCode to empty is super smelly, nor should it be empty. a `PaymentID` of `USD_MANUAL` or `BTC_MANUAL` must be possible. I prefer that we fix the assumption in places where such cryptoCode can always get a BTCPayNetworkProvider rather than hack by saying "empty crypto code", which would completely fuck up the export for example. (The Paid column of the export currency should be what is in the "CryptoCode" column in all case)

woutersamaey:
Can’t we call this “offline” payment or “other” instead of “manual” with a bag of coins icon? Looks confusing on the checkout page. I’d drop out of that checkout.

Kukks:
Superceded by #971

Issue #881, opened 2019-06-04
[UI] Unneeded string on login page with registering disabled.


Zaxounette:
**UI issue.**

**Describe the problem/bug**
The "Don't have an account ? Create one" string is unneeded when registering disabled.

**Your environment**
* Version of BTCPay Server:
1.03.113
* Deployment method: 
Lunanode

**To Reproduce**
1. Disable account registering in server admin dashboard Policies section
2. Logout
3. Click on "Log In"
4. Visualize the login page with the aforementioned string

**Actual behavior**
While the "disable registering" option is enabled in the server dashboard, the login page still shows the following string: "Don't have an account ? Create one" 
The "Create one" words being a link that redirects to the main page unnecessarily.

**Expected behavior**
Hiding the aforementioned string.

**Screenshots/Links**
[Login page](http://prntscr.com/nxgurv)
[Policies section](http://prntscr.com/nxgvyf)

// This is my first "issue" report, don't hesitate to bash me on the head if something wasn't done properly.

NicolasDorier:
Was fixed.

PR #935, opened 2019-07-21
Update Login.cshtml

radWorx:
[UI] Removed Unneeded string on login page with registering disabled. #881


NicolasDorier:
This PR remove the link even when registering is enabled.

radWorx:
will fix

Issue #949, opened 2019-07-30
Setting "Do not propose lightning payment if value of the invoice is above..." have no effect


HugoDoyon:
**Describe the problem/bug**
In manage store > Checkout experience, setting a value to the field _"Do not propose lightning payment if value of the invoice is above..."_ have no effect on the lightning network payment method type in the POS apps

**Your environment**
* Version of BTCPay Server: v1.0.3.124
* Deployment method: Prod on the cloud and regtest local debug
* Other relevant environment details: Debian 9

**To Reproduce**
1. Go to 'manage store > Checkout experience'
2. Set 100.00 USD in the "Do not propose lightning payment if value of the invoice is above..." field
3. Click on 'Save'
4. Create a store and a pos apps
5. Create a product in the store of value 120.00$
6. Save POS App settings
7. Click "View" POS App
8. Click on the POS product of value 120.00$ 
9. Lightning Network Payment method will still be visible.

**Expected behavior**
In the scenario describe above, Lightning Network Payment method should not be visible.

**Actual behavior**
Lightning Network Payment method still visible

**Screenshots/Links**
![Annotation 2019-07-30 190446](https://user-images.githubusercontent.com/17368121/62171469-001ecf80-b2fd-11e9-9b05-32c14c82da95.jpg)

**Additional context**
LightningMaxValue field of the store settings have not been use in InvoiceController.cs. The logic does not seem to have been implemented yet.

pavlenex:
Related #922

NicolasDorier:
Bug caused by 90852fe9 @Kukks

pavlenex:
@NicolasDorier I've just tested this upon the report of @HugoDoyon and we were both able to confirm the issue is still there even with [717cadc](https://github.com/btcpayserver/btcpayserver/commit/717cadc64b596a83761304686ca5e9a138255fc2) 

To replicate:
1. Do not propose lightning payment - Enter 100 USD
2. Create an invoice for 150 USD
3. Lightning is still proposed.

HugoDoyon:
Will take care of it thank you!

NicolasDorier:
@pavlenex  can you try on master please? I normally fixed it the commit after 717cadc

pavlenex:
I did try it on the master, and it didn't work for me. The issue is still there. @NicolasDorier you can't replicate?

pavlenex:
Could really be somehow related to #922

HugoDoyon:
[...]

Issue #922, opened 2019-07-15
Setting minimum on chain amount disabled lightning above that amount


Sjors:
**Describe the problem/bug**

If I configure a minimum amount for on chain payments:

<img width="596" alt="Schermafbeelding 2019-07-15 om 20 59 13" src="https://user-images.githubusercontent.com/10217/61241226-6b847100-a743-11e9-950e-418358798574.png">

Then the checkout page will no longer show Lightning for amounts above that.

**Your environment**
* Version of BTCPay Server: v1.0.3.124
* Deployment method: Ubuntu without Docker
* Other relevant environment details: 

**Logs (if applicable)**
<img width="774" alt="Schermafbeelding 2019-07-15 om 21 00 17" src="https://user-images.githubusercontent.com/10217/61241307-940c6b00-a743-11e9-8294-fac4753fb4af.png">
`"BTC (Off-Chain): The amount of the invoice is too high to be paid with lightning"`

**To Reproduce**

See above.

**Expected behavior**
I expect Lightning invoices for amount above the maximum on-chain amount. That's the job for the other option "Do not propose lightning payment if value of the invoice is above" .

**Additional context**
This looks incorrect: 

https://github.com/btcpayserver/btcpayserver/blob/01787e2662efb4eece7c0144bce9f9a0c19e3aa8/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs#L149-L169

pavlenex:
This is a bug, I was able to replicate it, thanks for reporting it.
Pinging @NicolasDorier

HugoDoyon:
See Pull Request #951

PR #951, opened 2019-08-01
Fix for bug #922 & bug #949 

HugoDoyon:
Fix for "Do not propose lightning payment if value of the invoice is above..." and minimum on chain amount. For your review ;)

### DIFF
--- a/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
@@ -74,38 +74,11 @@ namespace BTCPayServer.Payments.Bitcoin
             CurrencyValue limitValue = null;
             string errorMessage = null;
             
-            if (paymentMethodId.PaymentType == PaymentTypes.BTCLike)
+            if (storeBlob.OnChainMinValue != null)
             {
-                if (storeBlob.LightningMaxValue == null && 
-                    storeBlob.OnChainMinValue != null)
-                {
-                    compare = (value, limit) => value < limit;
-                    limitValue = storeBlob.OnChainMinValue;
-                    errorMessage = "The amount of the invoice is too low to be paid on chain";
-                }
-                else if (storeBlob.LightningMaxValue != null && 
-                         storeBlob.OnChainMinValue != null)
-                {
-                    if (storeBlob.LightningMaxValue.Currency == storeBlob.OnChainMinValue.Currency &&
-                        storeBlob.LightningMaxValue.Value <= storeBlob.OnChainMinValue.Value)
-                    {
-                        //Case where both fields are set but OnChainMinValue is greater
-                        // --> then use LightningMaxValue as limit
-                        compare = (value, limit) => value < limit;
-                        limitValue = storeBlob.LightningMaxValue;
-                        errorMessage = "The amount of the invoice is too low to be paid on chain";
-                    }
-                    else if (storeBlob.LightningMaxValue.Currency == storeBlob.OnChainMinValue.Currency &&
-                             storeBlob.LightningMaxValue.Value > storeBlob.OnChainMinValue.Value)
-                    {
-                        //Case where both fields are set but LightningMaxValue is greater
-                        // --> then use OnChainMinValue as limit
-                        //    (Otherwise a gap of price value with no payment method is possible) 
-                        compare = (value, limit) => value < limit;
-                        limitValue = storeBlob.OnChainMinValue;
-                        errorMessage = "The amount of the invoice is too low to be paid on chain";                        
-                    }
-                }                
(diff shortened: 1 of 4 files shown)


NicolasDorier:
I am not super fan of such whitebox testing and would prefer a test via selenium or through the real server. But that is not blocking, @pavlenex you can test it?

NicolasDorier:
This is not correct, the OnChainMinValue should remove onchain payments.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -73,37 +69,62 @@ public override string GetPaymentMethodName(PaymentMethodId paymentMethodId)
...
             Dictionary<CurrencyPair, Task<RateResult>> rate, Money amount, PaymentMethodId paymentMethodId)
         {
+            
             Func<Money, Money, bool> compare = null;
             CurrencyValue limitValue = null;
             string errorMessage = null;
-            if (paymentMethodId.PaymentType == PaymentTypes.LightningLike &&
-               storeBlob.LightningMaxValue != null)
+            
+            if (paymentMethodId.PaymentType == PaymentTypes.BTCLike)
```
NicolasDorier: remove this test, it is actually useless, we are always BTCLike if we reach here.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -73,37 +69,62 @@ public override string GetPaymentMethodName(PaymentMethodId paymentMethodId)
...
-                limitValue = storeBlob.LightningMaxValue;
-                errorMessage = "The amount of the invoice is too high to be paid with lightning";
-            }
-            else if (paymentMethodId.PaymentType == PaymentTypes.BTCLike &&
-               storeBlob.OnChainMinValue != null)
-            {
-                compare = (a, b) => a < b;
-                limitValue = storeBlob.OnChainMinValue;
-                errorMessage = "The amount of the invoice is too low to be paid on chain";
+                if (storeBlob.LightningMaxValue == null &&
```
NicolasDorier: `LightningMaxValue ` is orthogonal to onchain payment, no test should be done on that.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -73,37 +69,62 @@ public override string GetPaymentMethodName(PaymentMethodId paymentMethodId)
...
-                limitValue = storeBlob.OnChainMinValue;
-                errorMessage = "The amount of the invoice is too low to be paid on chain";
+                if (storeBlob.LightningMaxValue == null && 
+                    storeBlob.OnChainMinValue != null)
+                {
+                    compare = (value, limit) => value < limit;
+                    limitValue = storeBlob.OnChainMinValue;
+                    errorMessage = "The amount of the invoice is too low to be paid on chain";
+                }
+                else if (storeBlob.LightningMaxValue != null &&
```
NicolasDorier: same

NicolasDorier:
I made the code more complicated that it should have been. 
Basically the lightning handler should only use MaxLightningValue while the bitcoin handler should only use MinLightningValue.

No need for the strange hacks I did.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
HugoDoyon: No problem, that actually overkill indeed

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
HugoDoyon: That is true LightningMaxValue and OnChainMinValue are orthogonal. 

However, while testing I realize that a strange edge case could happen if LightningMaxValue is smaller than OnChainMinValue. In that situation there will be a gap where the no payment method will be available to the user. 

For exemple: IF LightningMaxValue = 50 USD$ AND OnChainMinValue = 52 USD$ AND InvoiceAmount =51 USD$ THEN No payment method available for user

HugoDoyon:
Glad to know that you guys prefer E2E test (Selenium/CodeUI/Protractor) over simple unit test as I do like them more too and I also very comfortable with those tech as well. Next time, I will propose those kind of test.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
NicolasDorier: That's fine. The use will get an error which explain clearly the issue.

HugoDoyon:
Change done ;)

PR #954, opened 2019-08-02
Prevent autofilling SMTP config user and password

woutersamaey:
Prevent autofilling SMTP config user and password.
Browser won't try to auto-fill the username and password anymore. This can mess up your form and your settings if you hit "save".

### DIFF
--- a/BTCPayServer/Services/Mails/EmailSettings.cs
+++ b/BTCPayServer/Services/Mails/EmailSettings.cs
@@ -24,7 +24,7 @@ namespace BTCPayServer.Services.Mails
         {
             get; set;
         }
-        [DataType(DataType.Password)]
+        
         public string Password
         {
             get; set;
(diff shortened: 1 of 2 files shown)


woutersamaey:
Damn, doesn't work on latest version of Chrome. Chrome ignores "autocomplete" it appears :(
Still might be a good change for other browsers.

woutersamaey:
I will experiment further by giving the fields different names etc. Maybe even change the password type to plain text.
Personally I’ve never seen the benefit in using a password field for config settings if the user is an admin. Admins should see everything, imho.

woutersamaey:
Would you approve such a change @btcpayserver?

NicolasDorier:
@woutersamaey I think it makes sense to use plaintext

woutersamaey:
I've changed the type of the password field to plain text and this fixes the auto-fill once and for all. Much better.

PR #955, opened 2019-08-03
Fix typo / grammar for dns error message

PatrickLemke:


PatrickLemke:
Hmm maybe something like
`If you are using the docker deployment process`
otherwise I'd say the wording is fine as it is.

Kukks:
how about `If you are using the docker deployment method` which is what we call them on the docs (https://docs.btcpayserver.org/deployment/deployment)

PatrickLemke:
Good idea, updated

PR #956, opened 2019-08-04
Prevent layout from breaking on hyperlinks with very long URLs as the visible text

woutersamaey:
Fixes broken layout like this.
The link text should be word wrapped so the table doesn't stretch wider.

image


woutersamaey:
I limited this CSS rule to the invoice detail page and added a comment in the CSS otherwise very hard to tell later on why that CSS rule is there.

PR #957, opened 2019-08-04
Openiddict - Authorize App view

Kukks:
Request consent from user before giving application access to the user's data & services.

We can extend this with permissions in another PR. 

image


BTCPayServer/Views/Authorization/Authorize.cshtml
```diff
@@ -0,0 +1,25 @@
...
+                <h2 class="section-heading">Authorization Request</h2>
+                <hr class="primary">
+                <p>@Model.ApplicationName is requesting access to your account</p>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-lg-12 text-center">
+                <form method="post">
+
+                    @Html.AntiForgeryToken()
```
NicolasDorier: don't need this, should be by default from the fact you use `form`.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -147,7 +147,7 @@ private void ConfigureOpenIddict(IServiceCollection services)
                 })
                 .AddServer(options =>
                 {
-                    
+                    options.EnableRequestCaching();
```
NicolasDorier: why?

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -0,0 +1,133 @@
...
+            // the original authorization request from the cache.
+            return View(new AuthorizeViewModel
+            {
+                ApplicationName = await _applicationManager.GetDisplayNameAsync(application),
+                RequestId = request.RequestId,
+                Scope = request.Scope
+            });
+        }
+
+        [Authorize(AuthenticationSchemes = Policies.CookieAuthentication), HttpPost("~/connect/authorize")]
```
NicolasDorier: Can you put on two lines and just `connect/authorize` ? I don't know what the `~` stand for.

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -0,0 +1,133 @@
...
+            IOptions<IdentityOptions> identityOptions)
+        {
+            _applicationManager = applicationManager;
+            _signInManager = signInManager;
+            _authorizationManager = authorizationManager;
+            _userManager = userManager;
+            _IdentityOptions = identityOptions;
+        }
+
+        [Authorize(AuthenticationSchemes = Policies.CookieAuthentication), HttpGet("~/connect/authorize")]
```
NicolasDorier: Same as below

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -0,0 +1,133 @@
...
+            string consent, bool createAuthorization = true)
+        {
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                return View("Error",
+                    new ErrorViewModel
+                    {
+                        Error = OpenIddictConstants.Errors.ServerError,
+                        ErrorDescription = "An internal error has occurred"
```
NicolasDorier: More descriptive error please, so it is easy to debug

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -0,0 +1,133 @@
...
+                    });
+            }
+
+            string type = null;
+            switch (consent.ToUpperInvariant())
```
NicolasDorier: what about consent `BLAHBLAH` ?

BTCPayServer/Hosting/Startup.cs
Kukks: This is needed if you use  the authorize screen. It caches the connection request momentarily so that you can pass around a request id until the user authorizes

BTCPayServer/Controllers/AuthorizationController.cs
Kukks: What do you mean?

NicolasDorier: that you don't handle the case where a consent is not the expected value.

BTCPayServer/Authentication/OpenId/OpenIdExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        public static async Task<AuthenticationTicket> CreateAuthenticationTicket(
+            OpenIddictApplicationManager<BTCPayOpenIdClient> applicationManager,
+            OpenIddictAuthorizationManager<BTCPayOpenIdAuthorization> authorizationManager,
+            IdentityOptions identityOptions,
+            SignInManager<ApplicationUser> signInManager,
+            OpenIdConnectRequest request,
+            ApplicationUser user,
+            AuthenticationProperties properties = null)
+        {
```
NicolasDorier: There is so much magic here, I don't know where to start. Are you using recommended oppeniddict practices? If so, where did you take it from?

Kukks: This magic is already merged in master, I just moved it to a new file for reuse.
It is based on code from https://github.com/openiddict/openiddict-samples and https://github.com/openiddict/openiddict-core/blob/dev/samples/Mvc.Server/Controllers/AuthorizationController.cs

BTCPayServer/Controllers/AuthorizationController.cs
Kukks: Blowing up is a correct response since no other Ui or api should be setting this.

Issue #649, opened 2019-03-05
Inventory limit for store item


britttttk:
Allow users to produce invoices for a store "item" up to a certain quantity. 

Use case example: ticket/event sales of limited quantity

Maybe it can be done with PoS?

pavlenex:
We should add some sort of quantity limit for items when refactoring the PoS.

royserpa:
> 
> 
> Allow users to produce invoices for a store "item" up to a certain quantity.
> 
> Use case example: ticket/event sales of limited quantity
> 
> Maybe it can be done with PoS?

You can keep track of that from your website's server. You could ask, on your site, how many tickets/items they would like to buy and if the number of tickets/items are less or equal than the ones available, send the users through a form and generate an invoice in your btcpayserver instance.

I think lots of these problems can be handled directly from the websites' server through databases and not really have to do that on the btcpayserver server.

pavlenex:
@royserpa  Not all people use BTCPay + website. PoS is one of many examples of why such structure is needed.

royserpa:
> 
> 
> @royserpa Not all people use BTCPay + website. PoS is one of many examples of why such structure is needed.

Wouldn't that affect the btcpayserver performance? Like, I think the server where btcpayserver is hosted on would be more efficient if it's only job is to handle btc payments, but just my humble opinion.

pavlenex:
It wouldn't, btcpay is way more than just a payment processor.

PR #961, opened 2019-08-10
New feature: Apps limited inventory

Kukks:
This PR introduces a new field to the Perks in Crowdfunding and items in PoS. 

It allows you to restrict the number of available items.

Some good use-cases: 
* Limited-edition merchandise
* Early-bird prices
* Event tickets

### DIFF
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -26,33 +26,39 @@ namespace BTCPayServer.Controllers
                     "  price: 1\n" +
                     "  title: Green Tea\n" +
                     "  description:  Lovely, fresh and tender, Meng Ding Gan Lu ('sweet dew') is grown in the lush Meng Ding Mountains of the southwestern province of Sichuan where it has been cultivated for over a thousand years.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg\n" +
+                    "  inventory: -1\n\n" +
                     "black tea:\n" +
                     "  price: 1\n" +
                     "  title: Black Tea\n" +
                     "  description: Tian Jian Tian Jian means 'heavenly tippy tea' in Chinese, and it describes the finest grade of dark tea. Our Tian Jian dark tea is from Hunan province which is famous for making some of the best dark teas available.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg\n" +
+                    "  inventory: -1\n\n" +
                     "rooibos:\n" +
                     "  price: 1.2\n" +
                     "  title: Rooibos\n" +
                     "  description: Rooibos is a dramatic red tea made from a South African herb that contains polyphenols and flavonoids. Often called 'African redbush tea', Rooibos herbal tea delights the senses and delivers potential health benefits with each caffeine-free sip.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2017/01/08/08/14/water-1962388__480.jpg\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2017/01/08/08/14/water-1962388__480.jpg\n" +
+                    "  inventory: -1\n\n" +
                     "pu erh:\n" +
                     "  price: 2\n" +
                     "  title: Pu Erh\n" +
                     "  description: This loose pur-erh tea is produced in Yunnan Province, China. The process in a relatively high humidity environment has mellowed the elemental character of the tea when compared to young Pu-erh.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n" +
+                    "  inventory: -1\n\n" +
                     "herbal tea:\n" +
                     "  price: 1.8\n" +
                     "  title: Herbal Tea\n" +
                     "  description: Chamomile tea is made from the flower heads of the chamomile plant. The medicinal use of chamomile dates back to the ancient Egyptians, Romans and Greeks. Pay us what you want!\n" +
                     "  image: https://cdn.pixabay.com/photo/2015/07/02/20/57/chamomile-829538__480.jpg\n" +
+                    "  inventory: -1\n" +
                     "  custom: true\n\n" +
                     "fruit tea:\n" +
                     "  price: 1.5\n" +
                     "  title: Fruit Tea\n" +
                     "  description: The Tibetan Himalayas, the land is majestic and beautiful—a spiritual place where, despite the perilous environment, many journey seeking enlightenment. Pay us what you want!\n" +
                     "  image: https://cdn.pixabay.com/photo/2016/09/16/11/24/darts-1673812__480.jpg\n" +
+                    "  inventory: 5\n" +
                     "  custom: true";
                 EnableShoppingCart = false;
                 ShowCustomAmount = true;
@@ -198,22 +204,11 @@ namespace BTCPayServer.Controllers
                 RedirectAutomatically = string.IsNullOrEmpty(vm.RedirectAutomatically)? (bool?) null: bool.Parse(vm.RedirectAutomatically)
                 
             });
-            await UpdateAppSettings(app);
+            await _AppService.UpdateAppSettings(app);
             StatusMessage = "App updated";
             return RedirectToAction(nameof(UpdatePointOfSale), new { appId });
         }
 
-        private async Task UpdateAppSettings(AppData app)
-        {
-            using (var ctx = _ContextFactory.CreateContext())
-            {
-                ctx.Apps.Add(app);
-                ctx.Entry<AppData>(app).State = EntityState.Modified;
-                ctx.Entry<AppData>(app).Property(a => a.Settings).IsModified = true;
-                ctx.Entry<AppData>(app).Property(a => a.TagAllInvoices).IsModified = true;
-                await ctx.SaveChangesAsync();
-            }
-        }
 
         private int[] ListSplit(string list, string separator = ",")
         {
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -26,32 +26,27 @@ namespace BTCPayServer.Controllers
                     "  price: 1\n" +
                     "  title: Green Tea\n" +
                     "  description:  Lovely, fresh and tender, Meng Ding Gan Lu ('sweet dew') is grown in the lush Meng Ding Mountains of the southwestern province of Sichuan where it has been cultivated for over a thousand years.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg\n" +
-                    "  inventory: -1\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2015/03/26/11/03/green-tea-692339__480.jpg\n\n" +
                     "black tea:\n" +
                     "  price: 1\n" +
                     "  title: Black Tea\n" +
                     "  description: Tian Jian Tian Jian means 'heavenly tippy tea' in Chinese, and it describes the finest grade of dark tea. Our Tian Jian dark tea is from Hunan province which is famous for making some of the best dark teas available.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg\n" +
-                    "  inventory: -1\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg\n\n" +
                     "rooibos:\n" +
                     "  price: 1.2\n" +
                     "  title: Rooibos\n" +
                     "  description: Rooibos is a dramatic red tea made from a South African herb that contains polyphenols and flavonoids. Often called 'African redbush tea', Rooibos herbal tea delights the senses and delivers potential health benefits with each caffeine-free sip.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2017/01/08/08/14/water-1962388__480.jpg\n" +
-                    "  inventory: -1\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2017/01/08/08/14/water-1962388__480.jpg\n\n" +
                     "pu erh:\n" +
                     "  price: 2\n" +
                     "  title: Pu Erh\n" +
                     "  description: This loose pur-erh tea is produced in Yunnan Province, China. The process in a relatively high humidity environment has mellowed the elemental character of the tea when compared to young Pu-erh.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n" +
-                    "  inventory: -1\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n\n" +
                     "herbal tea:\n" +
                     "  price: 1.8\n" +
                     "  title: Herbal Tea\n" +
                     "  description: Chamomile tea is made from the flower heads of the chamomile plant. The medicinal use of chamomile dates back to the ancient Egyptians, Romans and Greeks. Pay us what you want!\n" +
                     "  image: https://cdn.pixabay.com/photo/2015/07/02/20/57/chamomile-829538__480.jpg\n" +
-                    "  inventory: -1\n" +
                     "  custom: true\n\n" +
                     "fruit tea:\n" +
                     "  price: 1.5\n" +
--- a/BTCPayServer/Controllers/AppsController.PointOfSale.cs
+++ b/BTCPayServer/Controllers/AppsController.PointOfSale.cs
@@ -199,7 +199,7 @@ namespace BTCPayServer.Controllers
                 RedirectAutomatically = string.IsNullOrEmpty(vm.RedirectAutomatically)? (bool?) null: bool.Parse(vm.RedirectAutomatically)
                 
             });
-            await _AppService.UpdateAppSettings(app);
+            await _AppService.UpdateOrCreateApp(app);
             StatusMessage = "App updated";
             return RedirectToAction(nameof(UpdatePointOfSale), new { appId });
         }
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -304,7 +304,7 @@ namespace BTCPayServer.Controllers
                 
                 if (choice.Inventory == 0)
                 {
-                    return RedirectToAction(nameof(ViewPointOfSale), new { appId = appId });
+                    return NotFound("Option was out of stock");
                 }
                 if (choice.Inventory != -1)
                 {
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -1,33 +1,21 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Globalization;
-using System.IO;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Configuration;
-using BTCPayServer.Data;
 using BTCPayServer.Filters;
 using BTCPayServer.ModelBinders;
 using BTCPayServer.Models;
 using BTCPayServer.Models.AppViewModels;
-using BTCPayServer.Payments;
-using BTCPayServer.Rating;
 using BTCPayServer.Security;
 using BTCPayServer.Services.Apps;
-using BTCPayServer.Services.Invoices;
-using BTCPayServer.Services.Rates;
-using Ganss.XSS;
 using Microsoft.AspNetCore.Cors;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.EntityFrameworkCore;
-using NBitpayClient;
-using Newtonsoft.Json.Linq;
-using YamlDotNet.RepresentationModel;
 using static BTCPayServer.Controllers.AppsController;
 
 namespace BTCPayServer.Controllers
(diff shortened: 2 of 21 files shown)


BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
```diff
@@ -0,0 +1,88 @@
...
+                if (new[] {InvoiceEvent.Expired, InvoiceEvent.MarkedInvalid}.Contains(invoiceEvent.Name) &&
+                    !string.IsNullOrEmpty(invoiceEvent.Invoice.ProductInformation.ItemCode))
+                {
+                    var appIds = AppService.GetAppInternalTags(invoiceEvent.Invoice);
+
+                    if (!appIds.Any())
+                    {
+                        return;
+                    }
+
```
NicolasDorier: You should be able to check the appType BEFORE making a request to the database.

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
```diff
@@ -0,0 +1,88 @@
...
+                        {
+                            case AppType.PointOfSale:
+                                var possettings = data.GetSettings<AppsController.PointOfSaleSettings>();
+                                return (Data: data, Settings: (object)possettings,
+                                    Items: _AppService.Parse(possettings.Template, possettings.Currency));
+                            case AppType.Crowdfund:
+                                var cfsettings = data.GetSettings<CrowdfundSettings>();
+                                return (Data: data, Settings: (object)cfsettings,
+                                    Items: _AppService.Parse(cfsettings.PerksTemplate, cfsettings.TargetCurrency));
+                            default:
```
NicolasDorier: A timing bomb ticking. When  a new AppType will be created, this will explode.

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -130,13 +131,68 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
...
+                        if (updateNeeded)
+                        {
+                            settings.Template = _AppService.SerializeTemplate(choices);
+                            app.SetSettings(settings);
+                            await _AppService.UpdateAppSettings(app);
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine(e);
```
NicolasDorier: no

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -268,7 +270,40 @@ public Task<StoreData> GetStore(AppData app)
...
+            {
+                price =  item.Price.Value,
+                title= item.Title,
+                description = item.Description,
+                image = item.Image,
+                custom = item.Custom,
+                inventory = item.Inventory
+            }));
+            
+            // the frontend editor is not a proper yml parser so we need to add some very specific formatting to help it( new line before each item in list)
```
NicolasDorier: why not yaml or json

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -268,7 +270,40 @@ public Task<StoreData> GetStore(AppData app)
...
+
+            var split = result.Split(Environment.NewLine);
```
NicolasDorier: might not work on linux or windows as this is platform dependant, not user dependent,

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -293,7 +328,8 @@ public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
                                  Value = decimal.Parse(cc.Value.Value, CultureInfo.InvariantCulture),
                                  Formatted = Currencies.FormatCurrency(cc.Value.Value, currency)
                              }).Single(),
-                    Custom = c.GetDetailString("custom") == "true"
+                    Custom = c.GetDetailString("custom") == "true",
+                    Inventory = int.Parse( c.GetDetailString("inventory")?? "-1", CultureInfo.InvariantCulture)
```
NicolasDorier: formatting

britttttk:
Solves #649 :)

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
Kukks: I don't think there is a way( the tag is `APP#{appId}`) 

This is the first DB call anyway `await _AppService.GetApps(appIds)`

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
Kukks: fixed

BTCPayServer/Services/Apps/AppService.cs
Kukks: I was referring to: https://github.com/btcpayserver/btcpayserver/blob/63df6ac5eb13adc0f4e540bdada37c385bb8557a/BTCPayServer/wwwroot/products/js/products.js#L12

BTCPayServer/Controllers/AppsPublicController.cs
Kukks: refactored

BTCPayServer/Services/Apps/AppService.cs
Kukks: In https://github.com/btcpayserver/btcpayserver/blob/63df6ac5eb13adc0f4e540bdada37c385bb8557a/BTCPayServer/Controllers/AppsController.PointOfSale.cs#L25
we add the default template with `\n`

I've changed it to be the same along with doing a `Replace(Environment.NewLine, "\n") to normalize`

Kukks:
> Not sure if you're ready for ui/ux review but I have a few comments :)
> 
> PoS + Cart Enabled
> 
> 1. Can you please fix the cart count, if the user adds an item which has no remaining inventory, it decrements the cart count. (i.e. 1 Left is added twice, the second time the cart goes to 0)
>    Cart_Counter
> 2. The inventory values don't persist very well for the Store owner. If the Store Owner changes anything on the settings page it sets the current inventory back to the default inventory values.
> 3. Is it possible for at least the Sold out to persist after a payment? I understand the Pay button may have to be shown after the inventory is out but it's a bit unclear to the user.
>    For example: (before any payments)
>    ShowInventory
>    and after one payment to the 1 Left item:
>    PaidOnce

Thanks for the review! I cannot do anything with issue 2. The only solution is to not edit while the POS is in use.

I'll fix the rest today

britttttk:
It's much more clear now and thanks for fixing the cart counter.  :) 

My only other comment:

A. In the PoS + Cart the items with inventory less than -1 I think could be better handled the way that zero is handled. First so that the store owner only has to understand that -1 means without inventory. I don't think it's a big deal to just document that but it's also the case that if the store owner chooses any number less than -1, customers can't pay for those items when the cart is enabled and it currently doesn't show as Sold Out like 0 does. Maybe just make a change to the cart? I'm not sure.

CF looks great :)

image
image

Kukks:
Added some helper text + made any negative value setting disable inventory

BTCPayServer/Services/Apps/AppService.cs
NicolasDorier: Why not parsing the yaml server side and just give a json object to the front?

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
```diff
@@ -0,0 +1,89 @@
...
+                                return (null, null, null);
+                        }
+                    }).Where(tuple => tuple.Data != null &&  tuple.Items.Any(item =>
+                        item.Inventory >= 0 && item.Id == invoiceEvent.Invoice.ProductInformation.ItemCode));
+                    foreach (var valueTuple in apps)
+                    {
+                        foreach (var item1 in valueTuple.Items.Where(item =>
+                            item.Id == invoiceEvent.Invoice.ProductInformation.ItemCode))
+                        {
+                            item1.Inventory++;
```
NicolasDorier: why inventory increase? it should decrease

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                                ((AppsController.PointOfSaleSettings) valueTuple.Settings).Template =
+                                    _AppService.SerializeTemplate(valueTuple.Items);
+                                break;
+                            case AppType.Crowdfund:
+                                ((CrowdfundSettings) valueTuple.Settings).PerksTemplate =
+                                    _AppService.SerializeTemplate(valueTuple.Items);
+                                break;
+                            default:
+                                throw new InvalidOperationException();
```
NicolasDorier: timing bomb

BTCPayServer/HostedServices/AppInventoryUpdaterHostedService.cs
Kukks: This is for when an invoice does not get paid( either through expiring or being marked invalid). The inventory must be marked back to the original stock

BTCPayServer/Services/Apps/AppService.cs
Kukks: this is for the editor not the customer  facing UI

BTCPayServer/Controllers/AppsController.PointOfSale.cs
```diff
@@ -26,33 +26,39 @@ public PointOfSaleSettings()
...
                     "  description: This loose pur-erh tea is produced in Yunnan Province, China. The process in a relatively high humidity environment has mellowed the elemental character of the tea when compared to young Pu-erh.\n" +
-                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n\n" +
+                    "  image: https://cdn.pixabay.com/photo/2018/07/21/16/56/tea-cup-3552917__480.jpg\n" +
+                    "  inventory: -1\n\n" +
                     "herbal tea:\n" +
                     "  price: 1.8\n" +
                     "  title: Herbal Tea\n" +
                     "  description: Chamomile tea is made from the flower heads of the chamomile plant. The medicinal use of chamomile dates back to the ancient Egyptians, Romans and Greeks. Pay us what you want!\n" +
                     "  image: https://cdn.pixabay.com/photo/2015/07/02/20/57/chamomile-829538__480.jpg\n" +
+                    "  inventory: -1\n" +
```
NicolasDorier: why this is changed? what was working before should still work now.

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -130,13 +119,62 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
                 price = choice.Price.Value;
                 if (amount > price)
                     price = amount;
+
+                if (choice.Inventory == 0)
+                {
+                    return RedirectToAction(nameof(ViewPointOfSale), new { appId = appId });
+                }
+                if (choice.Inventory > -1)
```
NicolasDorier: Can't you use nullable instead of magic value -1?

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -241,6 +278,18 @@ public async Task<IActionResult> ContributeToCrowdfund(string appId, ContributeT
                 price = choice.Price.Value;
                 if (request.Amount > price)
                     price = request.Amount;
+                
+                if (choice.Inventory == 0)
+                {
+                    return NotFound("Option was out of stock");
+                }
+                if (choice.Inventory > -1)
```
NicolasDorier: should use nullable

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
```diff
@@ -20,6 +20,7 @@ public class ItemPrice
             public ItemPrice Price { get; set; }
             public string Title { get; set; }
             public bool Custom { get; set; }
+            public int Inventory { get; set; } = -1;
```
NicolasDorier: should be nullable

BTCPayServer/Services/Apps/AppService.cs
NicolasDorier: why not then just serializing the object via YamlJson?

BTCPayServer/Views/Apps/UpdateCrowdfund.cshtml
```diff
@@ -267,6 +267,12 @@
                     <textarea rows="3" cols="40" class="js-product-description form-control mb-2">{description}</textarea>
                 </div>
             </div>
+            <div class="form-row">
+                <div class="col">
+                    <label>Inventory (set to -1 to not use inventory feature)</label>
```
NicolasDorier: should be nullable

BTCPayServer/Views/AppsPublic/Crowdfund/ContributeForm.cshtml
```diff
@@ -43,11 +44,29 @@
...
             </div>
-            @if (Model.ViewCrowdfundViewModel.PerkCount.ContainsKey(item.Id))
+            @if (Model.ViewCrowdfundViewModel.PerkCount.ContainsKey(item.Id) || item.Inventory > -1)
             {
                 <div class="card-footer d-flex justify-content-between">
-                    <span></span>
-                    <span> @Model.ViewCrowdfundViewModel.PerkCount[item.Id] Contributors</span>
+                        @switch (item.Inventory)
+                        {
+                            case -1:
```
NicolasDorier: Should be nullable

BTCPayServer/Views/AppsPublic/Crowdfund/ContributeForm.cshtml
```diff
@@ -43,11 +44,29 @@
...
+                                <span></span>
+                                break;
+                            case 0:
```
NicolasDorier: should test for `item.Inventory <= 0`, because in your code it logically possible to have a race condition which end up with negative inventory.

BTCPayServer/wwwroot/products/js/products.js
```diff
@@ -43,6 +43,9 @@ Products.prototype.loadFromTemplate = function() {
             if (productProperty.indexOf('custom:') !== -1) {
                 custom = productProperty.replace('custom:', '').trim();
             }
+            if (productProperty.indexOf('inventory:') !== -1) {
```
NicolasDorier: still `-1`

BTCPayServer/wwwroot/crowdfund/app.js
```diff
@@ -38,7 +38,7 @@ addLoadEvent(function (ev) {
         },
         computed: {
             canExpand: function(){
-                return !this.expanded && this.active && (this.perk.price.value || this.perk.custom)
+                return !this.expanded && this.active && (this.perk.price.value || this.perk.custom) && this.perk.inventory !== 0
```
NicolasDorier: `inventory` can potentially be negative (if there is a race) or null, you don't seem to handle it here.

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -625,15 +635,37 @@ Cart.prototype.saveLocalStorage = function() {
...
+        var matchedItem = srvModel.items.find(function(item){
+            return item.id === self.content[i].id;
+        });
+        if(!matchedItem){
+            //remove if nio longer available
+            this.content.splice(i,1);
+            continue;
+        }else{
+            
+            if(matchedItem.inventory === 0){
```
NicolasDorier: inventory can be negative

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
```diff
@@ -370,29 +397,51 @@
 
                             </div>
                             <div class="card-footer bg-transparent border-0">
-                                @if (item.Custom)
+                                @if (!item.Inventory.HasValue || item.Inventory.Value != 0)
```
NicolasDorier: `if (item.Inventory is int inv && inv > 0)`

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
```diff
@@ -294,6 +304,23 @@
                                 <div class="card-footer pt-0 bg-transparent border-0">
 
                                     <span class="text-muted small">@String.Format(Model.ButtonText, @item.Price.Formatted)</span>
+                                     @if (item.Inventory.HasValue)
```
NicolasDorier: prefer `item.Inventory is int`

BTCPayServer/Views/AppsPublic/Crowdfund/VueCrowdfund.cshtml
```diff
@@ -301,9 +301,11 @@
                 </div>
             </div>
         </div>
-        <div class="card-footer d-flex justify-content-between" v-if="perk.sold">
-           <span ></span>
-           <span x >{{perk.sold}} Contributor{{perk.sold > 1? "s": ""}}</span>
+        <div class="card-footer d-flex justify-content-between" v-if="perk.sold || perk.inventory >= 0">
+           
+            <span v-if="perk.inventory > 0" class="text-center text-muted">{{perk.inventory}} left</span>
+            <span v-if="perk.inventory === 0" class="text-center text-muted">Sold out</span>
```
NicolasDorier: can be negative

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -303,6 +303,14 @@ public string SerializeTemplate(ViewPointOfSaleViewModel.Item[] items)
...
                     first = false;
                 }
+                else
+                {
+                    //YAMLDotNet's serializer with EmitDefaults() will print out nullable types with no value but we dont want to pollute the template with junk
+                    if (str.TrimEnd().Split(':', StringSplitOptions.RemoveEmptyEntries).Length == 1)
+                    {
+                        continue;
+                    }
+                }
```
NicolasDorier: What is this monster hack here, why are you splitting str, to join it again I don't understand why all of that is needed at all.

NicolasDorier: You can also just build the yaml by hand with YamlNodes, it will be less hack.

Kukks: Yeah,I do that actually, this was supposed to be simple through the serializer but instead went crappy.

I should do a pr later to allow json as an alternative to this annoying yml syntax

NicolasDorier:
Why are you concatenating strings that may be subject to bug if the user enter wrong char? We have a library to manipulate YAML via a DOM object model, just use this.

Kukks:
> Why are you concatenating strings that may be subject to bug if the user enter wrong char? We have a library to manipulate YAML via a DOM object model, just use this.

Because this is not from the user but from a .net class instance so the values are already valid and safe. And that YAML code looks way too messy without using their high level de/serializer.

NicolasDorier:
> Because this is not from the user but from a .net class instance so the values are already valid and safe. 

It is not for example your code does not support line-breaks and would act weird if the user is using Block Style Indicator/Block Chomping Indicator

NicolasDorier:
> And that YAML code looks way too messy without using their high level de/serializer.

We have 6 property, it means it should take 6 lines.

Kukks:
> > And that YAML code looks way too messy without using their high level de/serializer.
> 
> We have 6 property, it means it should take 6 lines.

I can't find any reference to this DOM API in their wiki? Do you have a code sample somewhere?.

NicolasDorier:
YamlDocument (https://github.com/btcpayserver/btcpayserver-docker/blob/master/docker-compose-generator/src/DockerComposeDefinition.cs#L190)

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -396,5 +426,39 @@ public async Task<AppData> GetAppDataIfOwner(string userId, string appId, AppTyp
                 return app;
             }
         }
+
+        public async Task UpdateAppSettings(AppData app)
+        {
+            using (var ctx = _ContextFactory.CreateContext())
+            {
+                ctx.Apps.Update(app);
```
NicolasDorier: this is not equivalent to the previous one, this potentially fuck up `Created` depending on how it is used.

NicolasDorier:
@britttttk can you give a new try ?

britttttk:
@Kukks It's really good. :)

The only issue I notice is that if the store owner views the settings page, after a payment has been made to an item with inventory, it always, only shows the last item created (on the settings page in the items-photos area). The template still shows all items. It's not an issue for the customer. This makes some kind of weird things for a newbie store owner who doesn't understand the template vs pictures method of viewing/adding items. 

For example: 5 items exist in the template but only 1 item shows as a photo-item. Then when a new item is added via the add product (or add perk) button (not using the template) it deletes all items in the template, even though they may have remaining inventory or items not using inventory. It will remove them because the template will reflect the photo-items when they press save changes. 

If it's not clear I can send some pictures. It's not really that bad just a bit strange when you see it.

Also I noticed in crowdfund if the inventoried perk gets removed (on purpose or by accident from the issue above), then the contributors count at the top of the CF page doesn't match the total number of contributors on the perks. Personally i'm ok with that, but maybe it interests you.

Kukks:
> @Kukks It's really good. :)
> 
> The only issue I notice is that if the store owner views the settings page, after a payment has been made to an item with inventory, it always, only shows the last item created (on the settings page in the items-photos area). The template still shows all items. It's not an issue for the customer. This makes some kind of weird things for a newbie store owner who doesn't understand the template vs pictures method of viewing/adding items.
> 
> For example: 5 items exist in the template but only 1 item shows as a photo-item. Then when a new item is added via the add product (or add perk) button (not using the template) it deletes all items in the template, even though they may have remaining inventory or items not using inventory. It will remove them because the template will reflect the photo-items when they press save changes.
> 
> If it's not clear I can send some pictures. It's not really that bad just a bit strange when you see it.
I'm not sure what that means

> Also I noticed in crowdfund if the inventoried perk gets removed (on purpose or by accident from the issue above), then the contributors count at the top of the CF page doesn't match the total number of contributors on the perks. Personally i'm ok with that, but maybe it interests you.

This is unrelated, it has something to do with the way  the editor UI saves the yaml id of items afaik

BTCPayServer/Services/Apps/AppService.cs
Kukks: This is not used for create https://github.com/Kukks/btcpayserver/blob/641a0674b35cbe659643c7b238ddc7a9516b22ff/BTCPayServer/Controllers/AppsController.cs#L131 bit it probably should as we do the db call straight in the controller

I can adapt it to handle create if you want

NicolasDorier:
tests stuck, can yuo run again?

britttttk:
@Kukks If you want to reproduce what I'm talking about.

1. Make a PoS app with the default items as you have it.
2. Pay for one item that has inventory (Fruit Tea)
3. Go to store settings and see the template contains all items... but the Products under the Add Product button... only shows the last item you have created (Fruit Tea)

Issue: 
4. If the store owner adds a New Item using the "Add Product" Button
5. Press save settings, 
6. The app removes all items except your New Item and Fruit Tea.

It means that the store owner can't add new items when inventory is being used. 
Because the old ones automatically get deleted. 

I hope it makes sense. :) 

image

BTCPayServer/Services/Apps/AppService.cs
NicolasDorier: @Kukks by created I mean the column.
You are updating fields that should not be updated.

NicolasDorier:
@Kukks you may have broken tests

Kukks:
Not anymore :D

Kukks:
@britttttk  I fixed that issue too now

PR #970, opened 2019-08-17
Add description and embedded css to POS

Kukks:


BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
```diff
@@ -42,11 +42,16 @@
...
-                max-width: 320px;
-                margin: auto;
-            }
+        .card-deck .card:only-of-type {
+            max-width: 320px;
+            margin: auto;
+        }
+        @if (!string.IsNullOrEmpty(Model.EmbeddedCSS))
+        {
+                @Safe.Raw(Model.EmbeddedCSS);
```
NicolasDorier: will not work, `Raw` can only be used in the context of an HTML doc, but you are inside the style tag. Look what I did when I fixed the vulnerability.

Kukks: hmm but it did work for me?

NicolasDorier: `@Safe.Raw` is meant to insert HTML. Not using it in this way mean there is probably a corner case that is not handled correctly, I am not comfortable with that. Please, just add the `<style></style>` tags to the `EmbeddedCSS` directly.

Kukks:
done
ᐧ

PR #973, opened 2019-08-19
Set Buyer email to refund email if not set

### LINKED ISSUE #941: Add refund email to invoice export
I use pure crowdfunding app without woocommerce integration. The problem is that in invoice view, I cannot see and especially export the refund email address field, which is the only contact information for customer that I have.

I'd like to have an option to easily export the list of paid invoices with the selected perk and e-mail address that they gave during payment.

Running:

```
docker exec -it c59817c0f6fe psql -U postgres btcpayservermainnet

select "CustomerEmail" from "Invoices" where "OrderId" like 'crowdfund-app%' and "Status"='complete'  order by "Created";
```

Kukks:
#fixes #941 (i think)

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceRepository.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceRepository.cs
@@ -122,12 +122,25 @@ retry:
             using (var ctx = _ContextFactory.CreateContext())
             {
                 var invoiceData = await ctx.Invoices.FindAsync(invoiceId).ConfigureAwait(false);
-                if (invoiceData == null)
+                if (invoiceData == null && data.Email != null)
                     return;
-                if (invoiceData.CustomerEmail == null && data.Email != null)
+                if (invoiceData.CustomerEmail == null)
                 {
                     invoiceData.CustomerEmail = data.Email;
                 }
+                
+                var invoice = ToObject(invoiceData.Blob);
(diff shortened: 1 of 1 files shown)


NicolasDorier:
I don't like changing the blob, the blob is meant to be read only.

NicolasDorier:
It should be fixed at the export level, not UpdateInvoice.

Kukks:
The problem is that it is a bit inconsistent: When you create an invoice
with a pre-existing buyer email, it sets both refund and blob email but
when you set the refund, it does not set the buyer. That would mean on
export, we would need to add 2 columns for possible emails, which is a bit
of a mess
ᐧ

NicolasDorier:
@Kukks I think just showing the RedundMail is enough.

Kukks:
For me, I need the buyer email to be set as the refund email is not available through the current API. One example of this use-case is the ticketing system we used for BTCPayDay  where I needed the invoice email to send the ticket. I did not care if it was a refund email or a buyer email, as long as it was the recipient.

One solution is to make the get invoice  method populate the buyer email from the refund email on load instead of mutating the data upon refund email being set. This solves it for the Bitpay API and for exports in one go

NicolasDorier:
`One solution is to make the get invoice method populate the buyer email from the refund email on load` I think this is good yes.

PR #976, opened 2019-08-20
simplify settings changed listening

Kukks:
Might as well use the Event aggregator to listen in on changes instead of the current complicated magic in the settings repo. This also reduces db calls( it sends the new settings through the event instead of fetching them again)

### DIFF
--- a/BTCPayServer/Events/SettingsChanged.cs
+++ b/BTCPayServer/Events/SettingsChanged.cs
@@ -1,0 +1,7 @@
+namespace BTCPayServer.Events
+{
+    public class SettingsChanged<T>
+    {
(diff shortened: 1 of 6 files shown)


NicolasDorier:
Can you just add an extension (or not extension) method for `eventAggregator.WaitSettingsChanged<TSettings>(settings => {})` ? I am not fan of the new syntax that is hard to read.

NicolasDorier:
I was not suggesting to go back to old style, just having `eventAggregator.WaitSettingsChanged` instead of `eventAggregator.WaitEvent<SettingsUpdated<MySettings>>(obj => obj.Settings)`

PR #987, opened 2019-08-23
Ensure raised amount is showing with correct precision in crowdfound in app tooltip

### LINKED ISSUE #926: Decimal denomination in the crowdfuding header
**Describe the problem/bug**
Denomination in the header has 7 decimals, in the wallet, we have 6, 6 is also standard for BTC denomination.

**Your environment**
* Version of BTCPay Server:  v1.0.3.124
* Deployment method: Docker

**To Reproduce**
1. Go to Crowdfunding app
2. Hover over the amount in BTC, see 7 decimals.
3. Go to the wallet, check the amount it's 6 decimals.

**Expected behavior**
The value in the wallet and in crowdfunding should be the same.

**Actual behavior**
The denomination is different in wallet and crowdfunding app.

**Screenshots/Links**
Screen Shot 2019-07-16 at 23 22 59

bolatovumar:
fix #926

### DIFF
--- a/BTCPayServer/Views/AppsPublic/Crowdfund/VueCrowdfund.cshtml
+++ b/BTCPayServer/Views/AppsPublic/Crowdfund/VueCrowdfund.cshtml
@@ -128,8 +128,7 @@
                 <b-tooltip target="crowdfund-body-raised-amount" v-if="paymentStats && paymentStats.length > 0">
                     <ul class="p-0 text-uppercase">
                         <li v-for="stat of paymentStats" class="list-unstyled">
-                            
-                            {{stat.label}} <span v-if="stat.lightning" class="fa fa-bolt"></span> {{stat.value}}
+                            {{stat.label}} <span v-if="stat.lightning" class="fa fa-bolt"></span> {{stat.value.toFixed(srvModel.currencyData.divisibility)}}
                         </li>
                     </ul>
                 </b-tooltip>      


BTCPayServer/Views/AppsPublic/Crowdfund/VueCrowdfund.cshtml
```diff
@@ -128,8 +128,7 @@
                 <b-tooltip target="crowdfund-body-raised-amount" v-if="paymentStats && paymentStats.length > 0">
                     <ul class="p-0 text-uppercase">
                         <li v-for="stat of paymentStats" class="list-unstyled">
-                            
-                            {{stat.label}} <span v-if="stat.lightning" class="fa fa-bolt"></span> {{stat.value}}
+                            {{stat.label}} <span v-if="stat.lightning" class="fa fa-bolt"></span> {{stat.value.toFixed(srvModel.currencyData.divisibility)}}
```
Kukks: You mentioned `A related issue here is that you get a tooltip even if you don't have any LN contributions so your tooltip doesn't add anything. See here:`

Can you add a `v-if` check against stat.value and see if it is null or 0 so that we avoid a js exception otherwise and also solve that issue

bolatovumar: @Kukks Actually this might be more complex than what I initially wrote. There is a problem of duplicate content only if the content in the tooltip is exactly the same as what's below the tooltip. For example, it's weird if you you are displaying raised funds in BTC and the tooltip also shows the same amount in BTC. If you displaying USD value instead then showing BTC value in tooltip is fine.

So the check needs to be that the content in the tooltip is not exactly the same as what's below it that triggers it. If not then we show the tooltip. Makes sense?

NicolasDorier:
@pavlenex can you take a look if it fix https://github.com/btcpayserver/btcpayserver/issues/926

NicolasDorier:
Another idea is that instead of rounding at the JS level we round at the C# level, passing the value as a string.

Kukks:
Yes, we already do this with some of the other values, will look into it

NicolasDorier:
@Kukks you could look into it? Can we merge?

pavlenex:
> 
> 
> @pavlenex can you take a look if it fix #926
tACK

It fixes it, however @bolatovumar mentioned he wants to tackle tooltip showing [only if there are more than 2 items shown.](https://github.com/btcpayserver/btcpayserver/issues/926#issuecomment-524235018) If you guys want it in this PR, then let's wait for that before we merge it.

bolatovumar:
@pavlenex I guess the issues I mentioned with only showing the tooltip if there is additional stuff to show there is a separate issue. This PR can be merged as is and I can open a separate issue for the other thing.

PR #990, opened 2019-08-24
Add display attributes.  Fix #98

bitcoinbrisbane:


BTCPayServer/Models/InvoicingModels/CreateInvoiceModel.cs
```diff
@@ -70,12 +75,15 @@ public SelectList Stores
...
-        
+
+        [DisplayName("Supported Transaction Currencies")]
         public List<string> SupportedTransactionCurrencies
         {
             get;
             set;
         }
+
+        [DisplayName("Available Payment Methods")]
```
Kukks: This doesn't show in the UI as far as I know. Would make sense to be the text for `SupportedTransactionCurrencies` actually

NicolasDorier: It should show in the UI.

PR #971, opened 2019-08-17
Manual payment no network

Kukks:
This PR is based on top of #920 but removes the dependency of the Network for a payment metho

PR #992, opened 2019-08-24
Allow overriding UI of checkout by payment handler

Kukks:
Taken off #920 / #971

### DIFF
--- a/BTCPayServer/Views/Invoice/CheckoutNoScript.cshtml
+++ b/BTCPayServer/Views/Invoice/CheckoutNoScript.cshtml
@@ -16,16 +16,22 @@
     <h1>Pay with @Model.StoreName</h1>
     @if (Model.Status == "new")
     {
-        <div>
-            <p>To complete payment, please send <b>@Model.BtcDue @Model.CryptoCode</b> to <b style="word-break: break-word;">@Model.BtcAddress</b></p>
-            <p>Time remaining: @Model.TimeLeft</p>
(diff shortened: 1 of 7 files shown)


BTCPayServer/Views/Invoice/CheckoutNoScript.cshtml
```diff
@@ -16,16 +16,22 @@
...
-            <p><a href="@Model.InvoiceBitcoinUrl" style="word-break: break-word;">@Model.InvoiceBitcoinUrl</a></p>
-            @if (Model.IsLightning)
-            {
-                <p>Peer Info: <b>@Model.PeerInfo</b></p>
-            }
-        </div>
-
+        if (!string.IsNullOrEmpty(Model.UISettings?.NoScriptPartialName))
+        {
+            <partial model="@Model" name="@Model.UISettings.NoScriptPartialName"/>
```
NicolasDorier: while you are at it, maybe you can move the else part in a partial?

Kukks: Already on it in a separate PR( I am migrating code to vue from the legacy jquery stuff)

NicolasDorier: should you do it in this PR so that's done?

Kukks: Some of the vue migration has to happen for the move to the partial so I think separate PR is better

PR #993, opened 2019-08-25
Revert "Add bisq as supporting P2P service"

wiz:
This reverts commit 0c5c6233c7d8f1577ce20fc933a071b775142032.

I think we discussed this the other day, but Bisq only supports Tor v2 so it cannot work with BTCPay's Tor v3 onion hostnames, and random plebs are coming into the Bisq support channel asking why their Bisq doesn't work with it:

Screen Shot 2019-08-25 at 16 38 49
Screen Shot 2019-08-25 at 16 38 38


NicolasDorier:
fuck I thought it would be easy to fix :(

PR #1000, opened 2019-08-29
[Refactor] Move SQL classes in their own assembly

NicolasDorier:
This PR attempt to move SQL classes in a separate assembly.

Entity Framework, for each migration, generate a snapshot of the database and generate more and more classes which never actually need to be recompiled.

If those classes are in BTCPayServer, then any change to a file to this project will trigger all the entity framework classes to be recompiled as well. This is a major waste of time which impact the coding/debug feedback loop.

I still have to check the impact of having changing the namespace of some types on migrations.


Kukks:
This is going to kill my pull requests with conflicts lol

NicolasDorier:
I don't think so, if you don't change the DB schema

NicolasDorier:
if it is too much issue, I might just change the migrations, not the data classes.

NicolasDorier:
Ok `13,277` less lines. @Kukks from now Entity Framework classes should be in this new assembly.

Issue #606, opened 2019-02-22
Point of Sale app - Cart Button and pop-up window size, invoice generation indicator


pavlenex:
After [watching this video](https://twitter.com/turm_im/status/1098735185481383937) I see we can make a few UX improvements in the PoS app.

- [x] Cart button size should be increased
- [x] Increase the size of pop-up confirmation windows, it's too small and fields are hard to navigate and click
- [x] When pay button is clicked, it should give some sort of indicator that the invoice is being generated

bolatovumar:
@pavlenex I also have some suggestions and nits to consider:

- [ ] Increase the default step amount for tip percentage to 1 instead of 0.01 (see screenshot below). You are unlikely to want to increase the tip in 0.01% amounts
- [ ] Ensure tip amount always shows up in cent amount (see screenshot below where it shows up as 0.7560000000000000001)
- [ ] Ensure selected tip amount is prominently displayed as selected (see screenshot below). Right now it has a background color on hover but when selected it only has a border. It should have a background color when selected and not just on hover
![tip](https://user-images.githubusercontent.com/1934678/63585486-e6914080-c5a7-11e9-8209-4269c44af8f2.png)
- [ ] Don't show cart in opened state on desktop (same as on mobile)
- [ ] Don't show anything in the cart when it's empty. Show a message like "Your shopping cart is empty." instead
- [ ] Center align amount input vertically. Right now it's shifted to the top
![input alignment](https://user-images.githubusercontent.com/1934678/63585462-d8432480-c5a7-11e9-9b7b-171c7b8d7af0.png)
- [ ] Left align the search bar with products below it. Right now it's floating out of alignment
- [ ] Apply consistent border radius to the search bar. Right now it's rounded on the left but not on the right for some reason
![search bar](https://user-images.githubusercontent.com/1934678/63585457-d4170700-c5a7-11e9-864d-dfc2738e9a6c.png)
- [ ] Consider increasing the search bar size to extend all the way to the cart button. Right now it's not obvious why it's taking up a little bit more than half the space available and it's just sort of floating there

bolatovumar:
@pavlenex I've address the three items you mentioned in the above three commits. I can continue with implementing the suggestions I listed above if deemed appropriate.

pavlenex:
@bolatovumar Thanks, feel free to tackle the remaining ones, nicely spotted.

NicolasDorier:
should update the checkboxes

[...]

PR #1004, opened 2019-08-31
Tackle additional issues in #606

bolatovumar:
This PR tackles remaining issues in #606

### DIFF
--- a/BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
+++ b/BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
@@ -49,7 +49,7 @@
     </style>
     @if (!string.IsNullOrEmpty(Model.EmbeddedCSS))
     {
-        @Safe.Raw($"<style>{Model.EmbeddedCSS}</style>");   
+        @Safe.Raw($"<style>{Model.EmbeddedCSS}</style>");
     }
 
 </head>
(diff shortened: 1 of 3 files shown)


NicolasDorier:
@pavlenex can you give it a try?

pavlenex:
@NicolasDorier tACK

pavlenex:
@bolatovumar 
Any idea why the closing "X" always appears in the search even though I didn't enter search query? It should be there only if I enter the query.

https://imgur.com/a/9o8r5M0

Besides that, everything else looks good.

bolatovumar:
@pavlenex it doesn't show up for me before i enter a query but i will take another look

bolatovumar:
@pavlenex ok, I took a look and looks like the issue you are describing is also in the master branch and has been there for a while. If you enter a query then the X icon shows up. If you click on it, it clears the input field and disappears. However, if you manually clear out the input field the icon stays there. I can add this issue to the task list and fix it.

Also, your video has another issue. The icon is meant to appear within the input field and not below it. Looks like you are missing some CSS. I'm pretty sure I pushed updated CSS but I will double-check.

bolatovumar:
@pavlenex alright, should be fixed now

NicolasDorier:
@pavlenex can you test once again?

pavlenex:
@bolatovumar it all looks good now, could be that my caching messed up with particular parts of css styling but the "x" no longer appears and is now inline.

My concern now (after clearing the cache) is that the sidebar is hidden until you click the cart icon. I'm not sure that's the best UX. 

imo, for desktop in particular it should be right away with PAY cta button, the icon click is not that intuitive and adds another step which is not really needed, what do you think? Should we:
- Show sidebar everywhere except on mobile
- Hide sidebar everywhere until icon is clicked

I'm nitpicking at this point, but it's not the UX I prefer.

NicolasDorier:
I think the cart must be open by default as well

NicolasDorier:
@pavlenex why you said on #606 that you don't want cart open by default on desktop and now say the reverse? :p

pavlenex:
> 
> 
> @pavlenex why you said on #606 that you don't want cart open by default on desktop and now say the reverse? :p

Well I didn't say specifically I want it opened, I thought it would be good if it's `bigger` in size.

I misunderstood @bolatovumar's checkbox that said "Don't show cart in opened state on desktop (same as on mobile)"

NicolasDorier:
Ok then waiting for this one to be fixed, then I will make a new version, BTCPay had tons of commits since the last one.

NicolasDorier:
@bolatovumar can you just keep the cart open by default? I will to release new version as soon as I merge this one.

NicolasDorier:
Merging that for now!

PR #1007, opened 2019-09-03
Remove dependency of network on payments (Excerpts from #971 (manual PR))

Kukks:
Remove dependency on network for payments / invoices + other small refacotrings

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.cs
+++ b/BTCPayServer/Controllers/InvoiceController.cs
@@ -10,6 +10,7 @@ using BTCPayServer.Events;
 using BTCPayServer.Logging;
 using BTCPayServer.Models;
 using BTCPayServer.Payments;
+using BTCPayServer.Payments.Bitcoin;
 using BTCPayServer.Rating;
 using BTCPayServer.Security;
 using BTCPayServer.Services.Apps;
@@ -20,6 +21,7 @@ using BTCPayServer.Services.Wallets;
 using BTCPayServer.Validation;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore.Internal;
 using NBitcoin;
 using NBitpayClient;
 using Newtonsoft.Json;
@@ -134,18 +136,9 @@ namespace BTCPayServer.Controllers
                                                  PaymentFilter.Where(p => !supportedTransactionCurrencies.Contains(p)));
             }
 
-            foreach (var network in store.GetSupportedPaymentMethods(_NetworkProvider)
-                                                .Where(s => !excludeFilter.Match(s.PaymentId))
-                                                .Select(c => _NetworkProvider.GetNetwork<BTCPayNetworkBase>(c.PaymentId.CryptoCode))
-                                                .Where(c => c != null))
-            {
-                currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, invoice.Currency));
-                //TODO: abstract
-                if (storeBlob.LightningMaxValue != null)
-                    currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, storeBlob.LightningMaxValue.Currency));
-                if (storeBlob.OnChainMinValue != null)
-                    currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, storeBlob.OnChainMinValue.Currency));
-            }
+            currencyPairsToFetch.AddRange(store.GetSupportedPaymentMethods(_NetworkProvider)
+                .Where(s => !excludeFilter.Match(s.PaymentId))
+                .SelectMany(method => method.PaymentId.PaymentType.GetCurrencyPairs( method, invoice.Currency, storeBlob)));
 
             var rateRules = storeBlob.GetRateRules(_NetworkProvider);
             var fetchingByCurrencyPair = _RateProvider.FetchRates(currencyPairsToFetch, rateRules, cancellationToken);
@@ -156,11 +149,10 @@ namespace BTCPayServer.Controllers
                                                 (Handler: _paymentMethodHandlerDictionary[c.PaymentId],
                                                 SupportedPaymentMethod: c,
                                                 Network: _NetworkProvider.GetNetwork<BTCPayNetworkBase>(c.PaymentId.CryptoCode)))
-                                                .Where(c => c.Network != null)
-                                                .Select(o =>
+                                               .Select(o =>
                                                     (SupportedPaymentMethod: o.SupportedPaymentMethod,
                                                     PaymentMethod: CreatePaymentMethodAsync(fetchingByCurrencyPair, o.Handler, o.SupportedPaymentMethod, o.Network, entity, store, logs)))
-                                                .ToList();
+                                               .ToList();
             List<ISupportedPaymentMethod> supported = new List<ISupportedPaymentMethod>();
             var paymentMethods = new PaymentMethodDictionary();
             foreach (var o in supportedPaymentMethods)
@@ -239,7 +231,16 @@ namespace BTCPayServer.Controllers
                 var logPrefix = $"{supportedPaymentMethod.PaymentId.ToPrettyString()}:";
                 var storeBlob = store.GetStoreBlob();
                 var preparePayment = handler.PreparePayment(supportedPaymentMethod, store, network);
-                var rate = await fetchingByCurrencyPair[new CurrencyPair(network.CryptoCode, entity.ProductInformation.Currency)];
+                var currencyPairs = supportedPaymentMethod.PaymentId.PaymentType.GetCurrencyPairs(supportedPaymentMethod,
+                    entity.ProductInformation.Currency, storeBlob);
+                var matchingCurrencyPairs = fetchingByCurrencyPair
+                    .Where(pair => currencyPairs.Contains(pair.Key));
+                if (!matchingCurrencyPairs.Any())
+                {
+                    return null;
+                }
+
+                var rate = await matchingCurrencyPairs.First().Value;
                 if (rate.BidAsk == null)
                 {
                     return null;
@@ -253,7 +254,7 @@ namespace BTCPayServer.Controllers
 
                 using (logs.Measure($"{logPrefix} Payment method details creation"))
                 {
-                    var paymentDetails = await handler.CreatePaymentMethodDetails(supportedPaymentMethod, paymentMethod, store, network, preparePayment);
+                    var paymentDetails = await handler.CreatePaymentMethodDetails(supportedPaymentMethod, paymentMethod, entity.ProductInformation.Currency, store, network, preparePayment);
                     paymentMethod.SetPaymentMethodDetails(paymentDetails);
                 }
 
--- a/BTCPayServer/Models/InvoicingModels/PaymentModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/PaymentModel.cs
@@ -1,11 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
-using BTCPayServer.Payments;
-using BTCPayServer.Payments.Bitcoin;
-using BTCPayServer.Services.Invoices;
-using Newtonsoft.Json;
+﻿using System.Collections.Generic;
 
 namespace BTCPayServer.Models.InvoicingModels
 {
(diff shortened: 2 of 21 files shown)


BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -639,26 +641,26 @@ public async Task AddRefundsAsync(string invoiceId, TxOut[] outputs, BTCPayNetwo
         /// <param name="cryptoCode"></param>
         /// <param name="accounted"></param>
         /// <returns>The PaymentEntity or null if already added</returns>
-        public async Task<PaymentEntity> AddPayment(string invoiceId, DateTimeOffset date, CryptoPaymentData paymentData, BTCPayNetworkBase network, bool accounted = false)
+        public async Task<PaymentEntity> AddPayment(string invoiceId, DateTimeOffset date, CryptoPaymentData paymentData, string cryptoCode, BTCPayNetworkBase network, bool accounted = false)
```
NicolasDorier: can you make another overload which keep the old signature so you don't have to update all the places that just use `network.CryptoCode` ? This is easier to review.

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
```diff
@@ -21,31 +24,33 @@ private BitcoinPaymentType()
         public override string ToPrettyString() => "On-Chain";
         public override string GetId() => "BTCLike";
 
-        public override CryptoPaymentData DeserializePaymentData(string str)
+        public override CryptoPaymentData DeserializePaymentData(string str, params object[] additionalData)
         {
-            return JsonConvert.DeserializeObject<BitcoinLikePaymentData>(str);
+            var result = JsonConvert.DeserializeObject<BitcoinLikePaymentData>(str);
+            result.Network = (BTCPayNetwork)additionalData[0];
```
NicolasDorier: wtf is this hack

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
```diff
@@ -21,31 +24,33 @@ private BitcoinPaymentType()
...
+            return result;
         }
 
         public override IPaymentMethodDetails DeserializePaymentMethodDetails(string str)
         {
             return JsonConvert.DeserializeObject<Payments.Bitcoin.BitcoinLikeOnChainPaymentMethod>(str);
         }
 
-        public override ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkBase network, JToken value)
+        public override ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkProvider networkProvider, PaymentMethodId paymentMethodId, JToken value)
```
NicolasDorier: this does not make any sense, you have now 3 parameters instead of 2 for doing exactly the same thing.

Kukks: It makes sense when you look at the full Manual PR, where there is no network available to use its serializer or to get the cryptoCode

NicolasDorier: So what about you keep `DeserializeSupportedPaymentMethod(BTCPayNetworkBase network, JToken value)` but also add `DeserializeSupportedPaymentMethod(PaymentMethodId paymentMethodId, JToken value)`

Kukks: ok, done

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
```diff
@@ -39,20 +39,26 @@ public override IPaymentMethodDetails DeserializePaymentMethodDetails(string str
         public override ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkProvider networkProvider, PaymentMethodId paymentMethodId, JToken value)
         {
             var network = networkProvider.GetNetwork<BTCPayNetwork>(paymentMethodId.CryptoCode);
-            if (network == null)
+            return DeserializeSupportedPaymentMethod(network, value);
+        }
+        
+        public override ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkBase network, JToken value)
+        {
+            if (network == null || !(network is BTCPayNetwork btcPayNetwork))
```
NicolasDorier: The parameter should be of type `BTCPayNetwork`

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
```diff
@@ -39,20 +39,26 @@ public override IPaymentMethodDetails DeserializePaymentMethodDetails(string str
         public override ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkProvider networkProvider, PaymentMethodId paymentMethodId, JToken value)
```
NicolasDorier: `(BTCPayNetworkProvider networkProvider, PaymentMethodId paymentMethodId, JToken value)` should not exist, only `( PaymentMethodId paymentMethodId, JToken value)`

Kukks: I can't do this as it would break the abstraction layer

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
Kukks: need to let it be generic or else introduce a a generic argument to `PaymentType` with a constraint to `BTCPayNetworkBase`

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
NicolasDorier: but why? you said that the Network can be null, but the only way it can be null is for the Manual payment.

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
NicolasDorier: I don't understand. This method can't be called with a `BTCPayNetworkBase` which is not a `BTCPayNetwork`. This mean that it should break as fast as possible if somebody is doing it. And the best time to break is at compile time.

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
Kukks: Because `DeserializeSupportedPaymentMethod` is called from a location that does not care if it uses a network or not. 

```csharp
var supportedPaymentMethod = paymentMethodId.PaymentType.DeserializeSupportedPaymentMethod(networks, paymentMethodId,
                        strat.Value);
```

This call needs to be able to  work with all variations of PaymentType

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -172,6 +176,10 @@ public override void PreparePaymentModel(PaymentModel model, InvoiceResponse inv
             model.PaymentMethodName = GetPaymentMethodName(network);
             model.InvoiceBitcoinUrl = cryptoInfo.PaymentUrls.BOLT11;
             model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls.BOLT11.ToUpperInvariant();
+            model.LightningAmountInSatoshi = storeBlob.LightningAmountInSatoshi;
+            
+            model.PreparePaymentModelWithChangelly(storeBlob, accounting);
```
NicolasDorier: it makes no sense to have LightningLikePaymentHandler adding Changelly and Coinswitch.

BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
Kukks: This method comes from an abstract class as an abstract method..the only way to strong type it is to have a generic argument on `PaymentType` like `public abstract class PaymentType<TNetwork> where TNetwork:BTCPayNetworkBase`

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
Kukks: true

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -47,6 +53,9 @@ public override void PreparePaymentModel(PaymentModel model, InvoiceResponse inv
             model.PaymentMethodName = GetPaymentMethodName(network);
             model.InvoiceBitcoinUrl = cryptoInfo.PaymentUrls.BIP21;
             model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls.BIP21;
+
+            model.PreparePaymentModelWithChangelly(storeBlob, accounting);
+            model.PreparePaymentModelWithCoinswitch(storeBlob, accounting);
```
NicolasDorier: Make no sense to have it here neither. IMHO you should just keep it outside really.

Kukks: I think it does, these converters are only useful on the onchain payment method so no need for the code to run on the main method

NicolasDorier:
need rebase

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -647,21 +655,21 @@ public async Task AddRefundsAsync(string invoiceId, TxOut[] outputs, BTCPayNetwo
...
```
NicolasDorier: Does passing `BTCPayNetworkBase network` is still useful?

Kukks: Yeah, it is useful in case there is a network associated to this cryptoCode that has its own serializer. 
Referring to this:
```cs
 private T ToObject<T>(byte[] value, BTCPayNetworkBase network)
{
    if (network == null)
    {
      return NBitcoin.JsonConverters.Serializer.ToObject<T>(ZipUtils.Unzip(value), null);
    }
  return network.ToObject<T>(ZipUtils.Unzip(value));
}
```

BTCPayServer/Payments/Bitcoin/NBXplorerListener.cs
```diff
@@ -362,7 +362,7 @@ private async Task<InvoiceEntity> ReceivedPayment(BTCPayWallet wallet, InvoiceEn
             invoice = (await UpdatePaymentStates(wallet, invoice.Id));
             if (invoice == null)
                 return null;
-            var paymentMethod = invoice.GetPaymentMethod(wallet.Network, PaymentTypes.BTCLike);
```
NicolasDorier: can you keep everywhere you can like it was before? `invoice.GetPaymentMethod(wallet.Network` should still be possible if the caller has Network in the scope.

This would simplify the review.

BTCPayServer/Models/InvoicingModels/PaymentModel.cs
```diff
@@ -76,5 +80,6 @@ public class AvailableCrypto
         public string RootPath { get; set; }
         public decimal CoinSwitchAmountMarkupPercentage { get; set; }
         public bool RedirectAutomatically { get; set; }
+         public Dictionary<string, object> AdditionalSettings { get; set; } = new Dictionary<string, object>();
```
NicolasDorier: why is that?

BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -240,7 +232,16 @@ private async Task<PaymentMethod> CreatePaymentMethodAsync(Dictionary<CurrencyPa
...
-                var rate = await fetchingByCurrencyPair[new CurrencyPair(network.CryptoCode, entity.ProductInformation.Currency)];
+                var currencyPairs = supportedPaymentMethod.PaymentId.PaymentType.GetCurrencyPairs(supportedPaymentMethod,
+                    entity.ProductInformation.Currency, storeBlob);
+                var matchingCurrencyPairs = fetchingByCurrencyPair
+                    .Where(pair => currencyPairs.Contains(pair.Key));
+                if (!matchingCurrencyPairs.Any())
+                {
+                    return null;
+                }
+
```
NicolasDorier: This code smell bad. All of this because we don't have `network` instead?

BTCPayServer/Models/InvoicingModels/PaymentModel.cs
Kukks: I plan to move payment specific properties to this so we stop spamming the main object with conditional values. Its usage is in the main PR

BTCPayServer/Controllers/InvoiceController.cs
Kukks: Open to suggestions but I think it does the job: `GetCurrencyPairs` needs to return a list of currency pairs that need the rate fetched because in the btclike and lightninglike payment handlers, we support querying for minimum onchain and maximum off chain rates. If no rates are matched, then the payment handler is not available.

Kukks:
@NicolasDorier I need to get back to this and the #971  PR so that I can start work on Liquid asset support in BTCPay as I will need to decouple the payments from the network. 

As far as I can tell there shouldnt be much work on this PR left.

NicolasDorier:
Can you rebase this?

Kukks:
> 
> 
> Can you rebase this?

done

PR #1028, opened 2019-09-10
Allow removing transaction labels by clicking on an icon

### LINKED ISSUE #1025: Suggestion: allow removing transaction labels by clicking an icon next to label
This is super minor but as mentioned in a [comment](https://github.com/btcpayserver/btcpayserver/issues/1017#issuecomment-528671815) in #1017 it would be nice to be able to remove labels from transaction by clicking some sort of "x" icon next to it instead of having to open a list of labels by clicking an icon on the very right and removing them from there.

bolatovumar:
close #1025


NicolasDorier:
@Kukks @rockstardev if @bolatovumar continued like this we will get sacked by @pavlenex very soon.

PR #1034, opened 2019-09-13
Replace pay button PNG image with an SVG

### LINKED ISSUE #1030: Pay button image looks blurry on iPhone
See how the "Pay with BTCPAY" button text looks blurry in comparison to other elements on the page. To resolve this the underlying image needs to be re-exported in 2x the original size (in pixels).

IMG_0015

bolatovumar:
fix #1030


NicolasDorier:
Awesome. yeah just need to keep the png.

bolatovumar:
@Kukks would it make sense to export the SVG as a PNG in 2x resolution to fix the blurriness for those using the PNG as well? So this way I don't delete the PNG but replace it with one in 2x resolution so that it doesn't look blurry on high-res screens.

Kukks:
Sounds good to me

bolatovumar:
@Kukks alright, done

PR #1039, opened 2019-09-18
Do not link to external image for default PoS app item

Kukks:
@ketominer pointed out that the PoS app reveals your BTCPay instance to whoever hosts those tea pictures. This PR uses a base64 image hardcoded in the app instead. We cannot make use of an local image file as it would complicate things in terms of multiple-domains or root path usage


NicolasDorier:
It makes the default template unreadable.

Kukks:
Yeah you're right, it's a bit much

PR #1041, opened 2019-09-18
Add sats as a native currency

Kukks:
This will allow you to create an invoice where its primary currency is denominated in sats

### DIFF
--- a/BTCPayServer.Rating/RateRules.cs
+++ b/BTCPayServer.Rating/RateRules.cs
@@ -99,6 +99,8 @@ namespace BTCPayServer.Rating
         RuleList ruleList;
 
         decimal _Spread;
+        private const string ImplicitSatsRule = "SATS_X = SATS_BTC * BTC_X;\nSATS_BTC = 0.00000001;\n";
+
         public decimal Spread
         {
             get
@@ -126,6 +128,7 @@ namespace BTCPayServer.Rating
         }
         public static bool TryParse(string str, out RateRules rules, out List<RateRulesErrors> errors)
         {
+            str = ImplicitSatsRule + str;
             rules = null;
             errors = null;
             var expression = CSharpSyntaxTree.ParseText(str, new CSharpParseOptions(LanguageVersion.Default).WithKind(SourceCodeKind.Script));
@@ -195,6 +198,7 @@ namespace BTCPayServer.Rating
         {
             return root.NormalizeWhitespace("", "\n")
                 .ToFullString()
+                .Replace(ImplicitSatsRule, string.Empty, StringComparison.OrdinalIgnoreCase)
                 .Replace("{\n", string.Empty, StringComparison.OrdinalIgnoreCase)
                             .Replace("\n}", string.Empty, StringComparison.OrdinalIgnoreCase);
         }
--- a/BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.Bitcoin.cs
+++ b/BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.Bitcoin.cs
@@ -21,11 +21,6 @@ namespace BTCPayServer
                 UriScheme = "bitcoin",
                 CryptoImagePath = "imlegacy/bitcoin.svg",
                 LightningImagePath = "imlegacy/bitcoin-lightning.svg",
-                DefaultRateRules = new[]
-                {
-                    "SATS_X = SATS_BTC * BTC_X",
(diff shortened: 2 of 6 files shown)


Kukks:
@opennoms  This is what you wanted right?

Kukks:
oops @openoms *

NicolasDorier:
this is not good. If an exchange can solve USD_BTC, then it should bt able to solve USD_SATS. Basically it should just be an implicit rate rule.

Kukks:
cleaned iy up to only use the rates scripting engine

BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
```diff
@@ -22,6 +22,11 @@ public void InitBitcoin()
                 UriScheme = "bitcoin",
                 CryptoImagePath = "imlegacy/bitcoin.svg",
                 LightningImagePath = "imlegacy/bitcoin-lightning.svg",
+                DefaultRateRules = new[]
+                {
+                    "SATS_X = SATS_BTC * BTC_X",
```
NicolasDorier: you should not do this, look at the implicit rules like `1/X`, you should modify the  the RateRules class.

Kukks: RateRules class is so complex, where are these implicit rules

bolatovumar:
I kinda wanted this as well and thought of working on it but good to know someone is working on it.

dennisreimann:
This will be great for donations pages. Had to fiddle around to make mine handle BTC fractions.

BTCPayServer.Common/BTCPayNetworkProvider.Bitcoin.cs
Kukks: ping @NicolasDorier

NicolasDorier:
Can't be a default rules, as default rules can be overidden.

I can't think of a way to implement this which would not result in crazy corner case.

Kukks:
> 
> 
> Can't be a default rules, as default rules can be overidden.
> 
> I can't think of a way to implement this which would not result in crazy corner case.

that's fine though, if someone wants to override the rule then it's a deliberate action?

NicolasDorier:
Maybe by changing `FindBestCandidate` such that, if the pair is `SATS_X` or `X_STATS`, then the rule is replaced by `BTC_X` or `X_BTC`.
Then later we take the matched expression * 10^8

NicolasDorier:
> that's fine though, if someone wants to override the rule then it's a deliberate action?

No. This will create crazy corner case that will take ages to debug and support. And most likely the person will do it by mistake when he wants to create his own script.

Kukks:
How about this now?

BTCPayServer.Rating/RateRules.cs
```diff
@@ -100,6 +100,8 @@ public SyntaxNode GetSyntaxNode()
         RuleList ruleList;
 
         decimal _Spread;
+        private const string ImplicitSatsRule = "SATS_X = SATS_BTC * BTC_X;\nSATS_BTC = 0.00000001;\n";
```
NicolasDorier: Only use `SATS_BTC = 0.00000001`, inverses are implicit.

Also you need to test that properly

BTCPayServer.Rating/RateRules.cs
```diff
@@ -127,6 +129,7 @@ public static bool TryParse(string str, out RateRules rules)
         }
         public static bool TryParse(string str, out RateRules rules, out List<RateRulesErrors> errors)
         {
+            str = ImplicitSatsRule + str;
```
NicolasDorier: Mofidy the rule list instead.

Kukks: The Rule list is black magic to me, no clue what or how it does anything

BTCPayServer.Rating/RateRules.cs
Kukks: Are you sure? As soon as I remove it, it makes every other rate for it unavailable.

NicolasDorier:
Write tests about it, then I can try to tweak it the correct way

Kukks:
rebased and added a test

NicolasDorier:
still need test

pavlenex:
tACK, though I must admit I first tried  to create an invoice with `SAT` which made more sense, but then realize `SATS` is needed.

PR #1044, opened 2019-09-21
xmr

Kukks:


BTCPayServer/Monero/Payments/MoneroPaymentType.cs
```diff
@@ -0,0 +1,44 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace BTCPayServer.Payments.Monero
+{
+    public class MoneroPaymentType: PaymentType
+    {
+        public static MoneroPaymentType Instance { get; } = new MoneroPaymentType();
+        public override string ToPrettyString()
+        {
+            return "";
```
NicolasDorier: why?

NicolasDorier: Should be On-Chain like bitcoin.

BTCPayServer/BTCPayServer.csproj
```diff
@@ -41,6 +41,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="MoneroRPC.NET" Version="1.0.1" />
```
NicolasDorier: no external dependency for shitcoins, you can probably just do a simple HttpClient wrapper for only the stuff you do.

Kukks:
Yeah, I built this nuget instead of putting it in btcpay directly to reduce
code in this repo. Should I add it back in?

BTCPayServer/BTCPayServer.csproj
Kukks: Yeah, I built this nuget instead of putting it in btcpay directly to reduce
code in this repo. Should I add it back in?

NicolasDorier: Just put a one file `MoneroClient` with only what you need, you don't need the full lib.

BTCPayServer/Hosting/Startup.cs
```diff
@@ -45,6 +46,7 @@ public void ConfigureServices(IServiceCollection services)
         {
             Logs.Configure(LoggerFactory);
             services.ConfigureBTCPayServer(Configuration);
+            services.AddMoneroLike();
```
NicolasDorier: Call that only if monero is in the options.

Kukks: I can't do it at this level as i need to wait for BTCPayServerOptions to be created in the DI. Any service running in the background checks if there is xmr configured else it wont do anything at all ( or modify the UI)

NicolasDorier: you can just check directly inside the `Configuration` object without the option. This is shitcoin code, if they break I don't care, but I care bitcoin break because of them.

BTCPayServer/Monero/Services/MoneroRPCProvider.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        public ConcurrentDictionary<string, MoneroLikeSummary> Summaries => _summaries;
+
+        public MoneroRPCProvider(MoneroLikeConfiguration moneroLikeConfiguration, EventAggregator eventAggregator)
+        {
+            _moneroLikeConfiguration = moneroLikeConfiguration;
+            _eventAggregator = eventAggregator;
+            DaemonRpcClients =
+                _moneroLikeConfiguration.MoneroLikeConfigurationItems.ToImmutableDictionary(pair => pair.Key,
+                    pair => new JsonRpcClient(pair.Value.DaemonRpcUri, "", ""));
```
NicolasDorier: Can you inject a `HttpClient` there via the IHttpClientFactory so there is no weird problem of dns change?

Kukks: done

BTCPayServer/Payments/PaymentMethodId.cs
```diff
@@ -70,7 +70,7 @@ public override string ToString()
 
         public string ToPrettyString()
         {
-            return $"{CryptoCode} ({PaymentType.ToPrettyString()})";
+            return string.IsNullOrEmpty(PaymentType.ToPrettyString()) ? CryptoCode : $"{CryptoCode} ({PaymentType.ToPrettyString()})";
```
NicolasDorier: Remove this, it would show as `Monero (On-Chain)` as it should.

PR #1045, opened 2019-09-21
add LND Hub 

yannick10800:
Hello
Could you add the feature of LNDhub ? (such as https://github.com/dangeross/guides/blob/master/raspibolt/README.md)

To allow the Node to be a LND hub?
This allows user to be hub for different people, which reduces the cost of opening channels for end users (but is more centralized)
Thanks


Kukks:
Feature requests should be issues not pull requests. This also is suited more on the btcpayserver-docker not this one

PR #1048, opened 2019-09-23
Authorize granular permissions

Kukks:
This PR makes the new API authorization require integrations to specify what they would like to be allowed to do once the user authorizes it. Controller action protection is already in place and tested :)
image

### DIFF
--- a/BTCPayServer/Authentication/BTCPayScopes.cs
+++ b/BTCPayServer/Authentication/BTCPayScopes.cs
@@ -1,0 +1,97 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using OpenIddict.Abstractions;
+
+namespace BTCPayServer.Authentication
+{
+    public static class RestAPIPolicies
+    {
+        public static class BTCPayScopes
+        {
+            public const string ViewStores = "view_stores";
+
+            //Create and manage stores
+            public const string StoreManagement = "store_management";
+
+            //create and manage invoices
+            public const string ViewInvoices = "view_invoices";
+
+            //create and manage invoices
+            public const string CreateInvoices = "create_invoice";
+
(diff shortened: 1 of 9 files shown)


BTCPayServer/Authentication/BTCPayScopes.cs
```diff
@@ -0,0 +1,97 @@
...
+                builder => builder.AddRequirements(new MultipleScopeGroupsRequirement(scopeGroups)));
+        }
+    }
+
+    public class MultipleScopeGroupsRequirement :
+        AuthorizationHandler<MultipleScopeGroupsRequirement>, IAuthorizationRequirement
+    {
+        private readonly IEnumerable<IEnumerable<string>> _ScopeGroups;
+
+        public MultipleScopeGroupsRequirement(IEnumerable<IEnumerable<string>> scopeGroups)
```
NicolasDorier: why is it enumerable of enumerable? It is not used.

Kukks: It is for ORs between scopes.
For example: 
to list and view apps, if you have a scope for `View Apps` OR `Manage apps`, the requirement is satisifed

NicolasDorier: this is awful to read. Can you use instead a lambda so instead of:
```csharp
    AddScopePolicy(options, CanViewInvoices,
                new[] {new[] {BTCPayScopes.ViewInvoices}, new[] {BTCPayScopes.InvoiceManagement}});
```
We get like

```csharp
    AddScopePolicy(options, CanViewInvoices,
`sG => HasScope(sG, BTCPayScopes.ViewInvoices) || HasScope(sG, BTCPayScopes.InvoiceManagement)`)
```

Kukks: done

NicolasDorier:
testse fail?

PR #1049, opened 2019-09-23
Introduce Swagger/ Open API document generator

Kukks:
image


Kukks:
We might be doing this manually so I guess I'll close this for now

Issue #677, opened 2019-03-14
Virtual Lightning Wallet (VLW) for Stores


lippoliv:
## Feature-Proposal

For Stores accepting BTC On-Chain, we do have wallets and thatfor, we see a kind of a "balance". Now, while LN gets more popular, On-Chain transactions will not be used for each invoice. For Off-Chain BTCPayServer does not have both:

1. A "wallet"/balance for the Store (partial available #538)
1. A payout-function (there's an API request partially matching this #618)

Having different stores means you do have a kind of different companys. You definetly need to keep track of what's the balance of those companys.

## My workaround

What I do for now is having an Excel-Sheet. I Export all invoices from BTCPayServer. Also I keep track of each Payment I do from BTCPayServer (using RTL) and assign it to an Store (manually, not cool).

Thanks to Pivot tables, I come up with this

| Store   | Invoices   | Payments   | Balance    |
|---------|------------|------------|------------|
| Store 1 | 0.00129564 | 0.00060004 | 0.00069560 |
| Store 2 | 0.00001442 | 0.00000000 | 0.00001442 |
| Store 3 | 0.00166419 | 0.00090010 | 0.00076409 |

This could be handled by BTCPayServer easily too.

## What needs to change

1. Add a "virtual lightning wallet" for stores
1. For each Paid invoice, add an "income" to that VLW, linking to the invoice
1. Add an "withdraw" function to that wallet (like it is for on-chain BTC already)
1. Enable the user to paste any BOLT11 Invoice, of Course it must be lower than the current balance of the store
1. On each payout, add an "payment" to that VLW including the current DATE and the BOLT11 description

## Benefits

Let's see the benefits with the knowledge of current mankind, trusting in Banks and being not able to manage their funds independently:

1. It enables us to use the same LN node for multiple Store owners (of course the ADMIN needs to decide whether this should be able or not)
1. Having this, it get's more easy to demonstrate BTCPayServer / Lightning Network to new store owners, ending up in more instances of BTCPayServer (as of they will not trust me "manging" their funds forever)
1. Small stores which just wants to deal with BTC payments ore test them are not forced to Setup and maintain an own BTCPayServer, they do not have any Knowledge for Linux/Docker/SSH/Servers/IPs/Domains, … While just moving $100 per month, you'll not learn all this

## Risks

BTCPayServer container must be able to Access the lightning node. This is already done.
Instead of just generating Invoices, BTCPayServer must be able to advice lightning node to pay one invoice.

[...]

PR #1053, opened 2019-09-25
Lightning Open Channel UI

Kukks:
New page to allow websites to request a btcpay store to open a channel towards it without having to input data. Useful for liquidity providers that reciprocate channel opening.

Route: `btcpay.com/lightning-node/{cryptocode}/open-channel`
Parameters available:
* storeId : pre-select the store to open a lightning channel from
* nodeInfo : which node to connect and open to (format is `node@ip:port`)
* sats : amount of satoshis to open channel with
* callback: redirect to this url after attempting channel opening. Replaces `{NodeId}` with the store's node id and `{Result}` with the channel opening result ( Ok, CannotAffordFunding,
PeerNotConnected, NeedMoreConf, AlreadyExists)
* description : a description to show on the UI
* servicename: the name of your service to display in the title

For example:
image
with the url
`https://btcpay.com/lightning-node/btc/open-channel?servicename=bitrefill&nodeinfo=test:1:0&sats=10000&description=bitrefill%20is%20requesting%20you%20to%20open%20a%20channel%20to%20their%20nodes.%20Upon%20completion,%20they%20will%20reciprocate%20with%20an%20inbound%20channel%20to%20you.`


NicolasDorier:
I don't think this is needed at this stage, but let's keep it open.

What we want is the reverse, bitrefill opening channel to provide liquidity, not BTCPay opening the channel, I talked with @ziggamon about simplest to do this. (Lightning channel product in bitrefill) I can explain at next meeting, this is even simpler than this.

Kukks:
I talked with @zigamon too. We can just add another param there for `lnurl` or an input form which would parse it an populate everything else

NicolasDorier:
like we talked about, liquidity provider create the channel, not the receiver. The workflow liquidity provider want is different from this one. Let's keep this PR aside for now, let's see when we find a company wanting to push this to their users.

Kukks:
I think we should merge it anyway. I can build a connect endpoint similar to this but accept the lnurl scheme

BTCPayServer/Controllers/LightningNodeController.cs
```diff
@@ -0,0 +1,234 @@
...
+            _BtcPayNetworkProvider = btcPayNetworkProvider;
+            _LightningClientFactory = lightningClientFactory;
+            _UserManager = userManager;
+            _LightningLikePaymentHandler = lightningLikePaymentHandler;
+            _FeeRateProvider = feeRateProvider;
+            _StoreRepository = storeRepository;
+        }
+
+        [HttpGet("{cryptoCode}/open-channel")]
+        [Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]
```
NicolasDorier: Should be able to modify the store, and `storeId` should be in the user (say `stores/{storeId}/open-channel`)

NicolasDorier:
There is permission issues, anybody can open a channel as long as they know the `storeId`.
Also, you should expose this open-channel page somewhere. Maybe in AddLightningNode?

BTCPayServer/Controllers/LightningNodeController.cs
Kukks: storeId is optional here so it could not be in the route. If a store id is nor provided, a dropdown with your stores is shown. Store did is checked in the POST to verify you have the permissions

Kukks:
Ok, so I should adjust this PR so that:
* it can handle connection to node only
* lnurl transaltion to our UI

NicolasDorier:
@Kukks I don't think lnurl  is relevant here. This PR is for opening a channel, lnurl is for having a liquidity provider opening a channel to you.

pavlenex:
There are some liquidity providers that open channel back to you once you open a channel to them,  I think that's what Kukk's is trying to do. Still this is complicated imo.

I think this PR should be left for the future, once we decide to have an internal wallet #677, as I already said, I think external wallets now do way better UI then what we can do at this point.

I would leave this on liquidity providers like Bitrefill to do.

In Berlin we explained to @ziggamon that a pop-up is needed where user can select one of many liquidity providers once they launch their own node, and get liquidity that way.

Kukks:
Yeah but it still comes down to a point where you have to show a UI for:
* entering the lnurl
* selecting which store to use `lnurl` against
* showing the information from lnurl and confirming against the user to connect to their node

You basically end up with the same UI, albeit simpler because it is adapted only for connecting to their node instead of opening channels.

PR #1057, opened 2019-09-28
Authorize granular permissions 

Kukks:
This PR makes the new API authorization require integrations to specify what they would like to be allowed to do once the user authorizes it. Controller action protection is already in place and tested :)
image

### DIFF
--- a/BTCPayServer/Controllers/RestApi/TestController.cs
+++ b/BTCPayServer/Controllers/RestApi/TestController.cs
@@ -66,38 +66,49 @@ namespace BTCPayServer.Controllers.RestApi
 
         [Authorize(Policy = RestAPIPolicies.CanViewStores,
             AuthenticationSchemes = OpenIddictValidationDefaults.AuthenticationScheme)]
+        [HttpGet(nameof(ScopeCanViewStores))]
         public bool ScopeCanViewStores() { return true; }
 
         [Authorize(Policy = RestAPIPolicies.CanManageStores,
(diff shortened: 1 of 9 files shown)


BTCPayServer/Controllers/RestApi/TestController.cs
```diff
@@ -52,10 +54,63 @@ public async Task<StoreData[]> GetCurrentUserStores()
...
+        [Authorize(Policy = RestAPIPolicies.CanManageWallet,
+            AuthenticationSchemes = OpenIddictValidationDefaults.AuthenticationScheme)]
+        [HttpGet(nameof(ScopeCanManageWallet))]
+        public bool ScopeCanManageWallet() { return true; }
+
+        [Authorize(Policy = RestAPIPolicies.CanViewProfile,
+            AuthenticationSchemes = OpenIddictValidationDefaults.AuthenticationScheme)]
+        
+        [HttpGet(nameof(ScopeCanViewProfile))]
+        public bool ScopeCanViewProfile() { return true; }
```
NicolasDorier: Can you test them? or maybe keep only one with a proper HTTP path and test it. (negative test and positive test)

Then we are good to merge.

Kukks: https://github.com/btcpayserver/btcpayserver/pull/1057/files#diff-edabefc2a7d0b3be0cb9a9feaeb3dc9bR141
There are tests against these routes here

PR #1063, opened 2019-09-30
validate xmr endpoint hit before proceeding

Kukks:


NicolasDorier:
so we close this? As you said it should not be exploitable.
Just in case, I just disable those endpoint for those not using monero.

PR #972, opened 2019-08-17
Basic localization support

woutersamaey:
@NicolasDorier please review the code in this PR as a basis for localization.

It is fully working with:
- Language switcher put in an easily visible place
- Locale stored in cookie as per .NET's best-practise, remembers for 1 year
- Fully compatible with .NET's IStringLocalizer (so no parallel solution - we can use native stuff)
- Data is stored in JSON files in a "Resources" folder, compatible with the Transiflex format
- Currently I have translated the main menu and left sidebar menus for the Dutch language so you can test and see content change
- English content is not affected as we discussed.
- All translation is done through 2 TagHelpers: 1 is for HTML content, and 1 is for translating HTML attributes like "title" and "alt".

Todo's are:
- Automatically detect the user's language based on the `Accept` request header as a last resort (if he doesn't have the cookie & no language stored in his user profile)
- Automatically append missing translations to a `en.json` file during development, so the file can be easily uploaded to Transiflex without needing to manually copy-paste keys. This is a pain.
- Automatically detect available languages (which JSON files are in the Resources dir?) - should be relatively easy
- More translations need to be added - also easy
- No translations with parameters encountered yet - not sure what to expect here
- No number formatting encountered yet - not sure what to expect here
- Text right-to-left support (like for Arabic) - appears not to be supported in Bootstrap HTML. Unofficial workarounds exist, but not sure which is best
- Sending localized emails in the user's language
- Translating form validation messages

Some things I'd like some help with are:
- Store the user's "preferred locale" in his profile. Would be essential for sending emails to the user in his language. Default to the language the user was using when creating his account or English for already existing users.
- ~~I'm not 100% sure the translations are kept in memory and the JSON file isn't re-read. Maybe you could help me find out? There is a flag to "reload on change" and that is set to "true", so maybe it's decent. I dunno.~~

I'm planning to address all of these TODOs, but would like to get your thought so far.

I don't expect you to approve the PR in this current state, though you probably can!! After all it is functional, just not 100% translated yet.

PR #1065, opened 2019-09-30
new feature: Wallet Receive Page

### LINKED ISSUE #965: Allow receiving funds on BTCPay Wallet
BTCPay is a very feature-rich wallet. However, there is no way outside of generating invoices(through apps or manuall) to be able to get an address to send bitcoin to directly.

Add a new section to the wallet UI: "Receiving"
Here, you can click a button "Generate new address" which would reserve a new address to send bitcoin to and display it on screen with a QR code. Optionally, you should also be able to pre-assign labels to this receiving address so that once funds are received, the wallets shows the label.

Create an invoice manually or through an app. This has the drawback that the invoice shows a specific amount and an unnecessary UI for something you want quick. 

**Provide examples**
Transferring some bitcoin from an exchange to your cold-storage wallet linked to your btcpay store. Just generate a receiving address and attach a label "funds bought from binance exchange after receiving August salary "

Kukks:
Very simple for now but we can improve on UX and UI afterwards
image

closes #965

### DIFF
--- a/BTCPayServer/Controllers/WalletsController.cs
+++ b/BTCPayServer/Controllers/WalletsController.cs
@@ -295,6 +295,26 @@ namespace BTCPayServer.Controllers
         }
 
         [HttpGet]
+        [Route("{walletId}/receive")]
+        public async Task<IActionResult> WalletReceive([ModelBinder(typeof(WalletIdModelBinder))]
+            WalletId walletId)
+        {
+            if (walletId?.StoreId == null)
+                return NotFound();
+            var store = await Repository.FindStore(walletId.StoreId, GetUserId());
+            DerivationSchemeSettings paymentMethod = GetDerivationSchemeSettings(walletId, store);
+            if (paymentMethod == null)
+                return NotFound();
+            var network = this.NetworkProvider.GetNetwork<BTCPayNetwork>(walletId?.CryptoCode);
+            if (network == null)
+                return NotFound();
+
+            var wallet = _walletProvider.GetWallet(network);
+            var address = await wallet.ReserveAddressAsync(paymentMethod.AccountDerivation);
+            return View(new WalletReceiveViewModel(){CryptoCode = walletId.CryptoCode, Address = address.ToString()});
+        }
+
+    [HttpGet]
         [Route("{walletId}/send")]
         public async Task<IActionResult> WalletSend(
             [ModelBinder(typeof(WalletIdModelBinder))]
@@ -957,6 +977,12 @@ namespace BTCPayServer.Controllers
                 return NotFound();
             }
         }
+    }
+
+    public class WalletReceiveViewModel
+    {
+        public string CryptoCode { get; set; }
+        public string Address { get; set; }
     }
 
 
--- a/BTCPayServer/Controllers/WalletsController.cs
+++ b/BTCPayServer/Controllers/WalletsController.cs
@@ -314,13 +314,22 @@ namespace BTCPayServer.Controllers
             DerivationSchemeSettings paymentMethod = GetDerivationSchemeSettings(walletId);
             if (paymentMethod == null)
                 return NotFound();
-            var network = this.NetworkProvider.GetNetwork<BTCPayNetwork>(walletId?.CryptoCode);
+            var network = NetworkProvider.GetNetwork<BTCPayNetwork>(walletId?.CryptoCode);
             if (network == null)
                 return NotFound();
 
             var address = _WalletReceiveStateService.Get(walletId)?.Address;
-            
-            return View(new WalletReceiveViewModel(){StatusMessage = statusMessage,CryptoCode = walletId.CryptoCode, Address = address?.ToString(), CryptoImage = GetImage(paymentMethod.PaymentId, network)});
+            if (!string.IsNullOrEmpty(statusMessage))
+            {
+                TempData[WellKnownTempData.SuccessMessage] = statusMessage;
+            }
+
+            return View(new WalletReceiveViewModel()
+            {
+                CryptoCode = walletId.CryptoCode,
+                Address = address?.ToString(),
+                CryptoImage = GetImage(paymentMethod.PaymentId, network)
+            });
         }
 
         [HttpPost]
@@ -1042,7 +1051,6 @@ namespace BTCPayServer.Controllers
 
     public class WalletReceiveViewModel
     {
-        public string StatusMessage { get; set; }
         public string CryptoImage { get; set; }
         public string CryptoCode { get; set; }
         public string Address { get; set; }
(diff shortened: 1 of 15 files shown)


NicolasDorier:
YES!
Can you make it the default page instead of /transactions though? Because for big wallet /transactions is slow to open.

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -271,6 +271,26 @@ private static string GetLabelTarget(WalletId walletId, uint256 txId)
...
+            var store = await Repository.FindStore(walletId.StoreId, GetUserId());
+            DerivationSchemeSettings paymentMethod = GetDerivationSchemeSettings(walletId, store);
+            if (paymentMethod == null)
+                return NotFound();
+            var network = this.NetworkProvider.GetNetwork<BTCPayNetwork>(walletId?.CryptoCode);
+            if (network == null)
+                return NotFound();
+
+            var wallet = _walletProvider.GetWallet(network);
+            var address = await wallet.ReserveAddressAsync(paymentMethod.AccountDerivation);
```
NicolasDorier: Don't reserve everytimes.

NicolasDorier:
So here are the difficult parts.

1. You should not generate a new address at each refresh. This is a GET and should not have side effect.
2. You can't just show the next unused address, because it is possible that an invoice comes in with this address, and the payment then get attributed to the invoice by mistake.

It means that either: 
1. You must show a button showing "Get a new address".
2. You must reserve the address the first time and show it until it get actually used. (I can maybe add something in NBXplorer to know this)

Kukks:
Yes I ran into this dilemma and opted to still make the most basic version of this PR for the sake of discussion.
Coin logo is no issue, will add it in a bit.

I think it should be that
* browse to /receive
* Check if we have reserved address for wallet receive ( memory cache?)
* if we do  have a reserved address, check if it was used.
  * If used, clear cache and generate new address( and save to cache)
  * If not used, display address
* If we do not, generate new address( and save to cache)
*show address on screen + a button to "Generate a new address"

We will also need an even subscription to sore derivation changes so that we clear cache so that we dont show an incorrect address from cache

The main issue here is that you still generate a new address per btcpay run but it's not too bad

pavlenex:
Oh boy, I wanted this so bad.

About the new unused address, I think, generate a new address button should be enough?

Bisq has an interesting UI for that, but perhaps it's a bit of an overkill, but the bottom part of the image can be handy for us?

bisq

Kukks:
> 
> 
> Yes I ran into this dilemma and opted to still make the most basic version of this PR for the sake of discussion.
> Coin logo is no issue, will add it in a bit.
> 
> I think it should be that
> 
>     * browse to /receive
> 
>     * Check if we have reserved address for wallet receive ( memory cache?)
> 
>     * if we do  have a reserved address, check if it was used.
>       
>       * If used, clear cache and generate new address( and save to cache)
>       * If not used, display address
> 
>     * If we do not, generate new address( and save to cache)
>       *show address on screen + a button to "Generate a new address"
> 
> 
> We will also need an even subscription to sore derivation changes so that we clear cache so that we dont show an incorrect address from cache
> 
> The main issue here is that you still generate a new address per btcpay run but it's not too bad

Done!

NicolasDorier:
> We will also need an even subscription to sore derivation changes so that we clear cache so that we dont show an incorrect address from cache

Or I maybe can just add an API at NBXplorer level to know about this.

Kukks:
rebased

NicolasDorier:
tests don't pass

pavlenex:
Looks great. I've tested it, but is it just me or the wording `Unreserve BTC Address` is a bit odd? Perhaps `Skip`? I completely understand what it does, but not sure if there's a better wording that may even be easier to translate in the future.

britttttk:
tACK - The `Unreserve this Address` text is clear to me.

Kukks:
tests pass @NicolasDorier  
It can definitely be improved with some extra functionality from NBX but this preserves addresses as much as possible and should be a decent base

pavlenex:
> tACK - The `Unreserve this Address` text is clear to me.

Dunno, I can't even think how `unreserve` can be translated. `Next Address` is easier, but if you guys are for the term, I'm fine with it. We'll see if it bothers translators, we're still far away from #972

Kukks:
> 
> 
> > tACK - The `Unreserve this Address` text is clear to me.
> 
> Dunno, I can't even think how `unreserve` can be translated. `Next Address` is easier, but if you guys are for the term, I'm fine with it. We'll see if it bothers translators, we're still far away from #972

Next address is not the same as unreserve

Unreserve means that you can use this address in the next invoice if you click it
next address means that you leave the current reserved and reserve the next one too

pavlenex:
> > > tACK - The `Unreserve this Address` text is clear to me.
> > 
> > 
> > Dunno, I can't even think how `unreserve` can be translated. `Next Address` is easier, but if you guys are for the term, I'm fine with it. We'll see if it bothers translators, we're still far away from #972
> 
> Next address is not the same as unreserve
> 
> Unreserve means that you can use this address in the next invoice if you click it
> next address means that you leave the current reserved and reserve the next one too

Wait, but the address changes upon receiving balance? Why would I want to `unreserve` the address, users don't care, if no balance show this address, when an invoice is generated with that address, show different one, I don't see why users even need to do anything, as it may encourage address re-use, no?

I'd probably have to take another look at the PR, I forgot how it works, but if there's a concensus on the term, I'm fine with it, perhaps it's just me that finds it unconventional (I've never seen a software or a wallet using the term) and difficult to translate.

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -276,6 +286,68 @@ private static string GetLabelTarget(WalletId walletId, uint256 txId)
...
+            if (walletId?.StoreId == null)
+                return NotFound();
+            DerivationSchemeSettings paymentMethod = GetDerivationSchemeSettings(walletId);
+            if (paymentMethod == null)
+                return NotFound();
+            var network = this.NetworkProvider.GetNetwork<BTCPayNetwork>(walletId?.CryptoCode);
+            if (network == null)
+                return NotFound();
+
+            var address = _WalletReceiveStateService.Get(walletId)?.ScriptPubKey
```
NicolasDorier: won't work with liquid, maybe you can cache the address instead of the ScriptPubKey ?

NicolasDorier:
This need rebase, this is lot's of change for a small feature though.
What about removing the `WalletReceiveCacheUpdater.cs`, and instead just having `WalletReceiveStateService.Get` an async method, which reserve the next address to NBXplorer and cache it for the next request in memory.

BTCPayServer/Controllers/WalletsController.cs
Kukks: yeah this was before we added the address in nbx . Fixed

Kukks:
Rebased

Kukks:
> 
> 
> This need rebase, this is lot's of change for a small feature though.
> What about removing the `WalletReceiveCacheUpdater.cs`, and instead just having `WalletReceiveStateService.Get` an async method, which reserve the next address to NBXplorer and cache it for the next request in memory.
That's what I do though. I reserve and cache an address in memory until it is either:
* unreserved manually
* the wallet changes
* a payment comes in 

I don't think it is bad at all

PR #1070, opened 2019-10-03
[WIP] Switch to .netcoreapp3.0

NicolasDorier:
This build but crashes at runtime because the following dependencies are still stuck in .netcoreapp2.2:

* [x] [Pomelo.EntityFrameworkCore.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) (see [following issue](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/839))
* [ ] [openiddict-core ](https://github.com/openiddict/openiddict-core)
* [ ] [BundlerMinifier PR](https://github.com/madskristensen/BundlerMinifier/pull/444)

Both library have nightly build for 3.0 support, but we should probably better wait for official release.

### DIFF
--- a/BTCPayServer.Tests/BTCPayServer.Tests.csproj
+++ b/BTCPayServer.Tests/BTCPayServer.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
     <TargetFramework Condition="'$(TargetFrameworkOverride)' != ''">$(TargetFrameworkOverride)</TargetFramework>
     <IsPackable>false</IsPackable>
     <NoWarn>NU1701,CA1816,CA1308,CA1810,CA2208</NoWarn>
--- a/BTCPayServer.Tests/Dockerfile
+++ b/BTCPayServer.Tests/Dockerfile
@@ -1,8 +1,7 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:2.1.505-alpine3.7 AS builder
-ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
-RUN apk add --no-cache icu-libs
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.100 AS builder
+
+RUN apt-get update && apt-get install -y --no-install-recommends chromium-driver \
+    && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /source
 COPY nuget.config nuget.config
@@ -11,15 +10,12 @@ COPY BTCPayServer/BTCPayServer.csproj BTCPayServer/BTCPayServer.csproj
 COPY BTCPayServer.Common/BTCPayServer.Common.csproj BTCPayServer.Common/BTCPayServer.Common.csproj
 COPY BTCPayServer.Rating/BTCPayServer.Rating.csproj BTCPayServer.Rating/BTCPayServer.Rating.csproj
 COPY BTCPayServer.Data/BTCPayServer.Data.csproj BTCPayServer.Data/BTCPayServer.Data.csproj
-COPY BTCPayServer.Tests/BTCPayServer.Tests.csproj BTCPayServer.Tests/BTCPayServer.Tests.csproj
-RUN dotnet restore BTCPayServer.Tests/BTCPayServer.Tests.csproj
-
-ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT false
-
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
-
-RUN apk add --no-cache chromium chromium-chromedriver icu-libs
+RUN cd BTCPayServer && dotnet restore
+COPY BTCPayServer.Common/. BTCPayServer.Common/.
+COPY BTCPayServer.Rating/. BTCPayServer.Rating/.
+COPY BTCPayServer.Data/. BTCPayServer.Data/.
+COPY BTCPayServer/. BTCPayServer/.
+COPY Build/Version.csproj Build/Version.csproj
 
 ENV SCREEN_HEIGHT 600 \
     SCREEN_WIDTH 1200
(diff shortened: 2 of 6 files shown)


PinpointTownes:
FYI, the latest OpenIddict 3.0 alpha packages (3.0.0-alpha1.19504.68) are now compatible with ASP.NET Core 2.1, 2.2 and 3.0, which should muchly simplify your code base, since you'll no longer need to reference different OpenIddict versions to support both ASP.NET Core 2.x and 3.x: https://github.com/openiddict/openiddict-core/issues/736#issuecomment-538601538

BTCPayServer/Authentication/OpenId3/ClientCredentialsGrantTypeEventHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+                return;
+            }
+
+            var application = await _applicationManager.FindByClientIdAsync(request.ClientId);
+            if (application == null)
+            {
+                context.Reject(
+                    error: OpenIddictConstants.Errors.InvalidClient,
+                    description: "The client application was not found in the database.");
+                context.HandleRequest();
```
PinpointTownes: `context.HandleRequest()` is a way to tell OpenIddict that you want to take care of returning the response yourself (e.g by returning a custom HTTP response). It's likely not what you want to do here. In general, `context.Reject()`, `context.HandleRequest()` and `context.SkipRequest()` are all mutually exclusive.

NicolasDorier:
hey @PinpointTownes I am happy to know OpenIddict3.0 is compatible with 2.1.

Not because we intend to use OpenIddict3.0 for long in 2.1, but because it means we can gradually move the code to 3.0 instead of having one big PR.

The main pain points in the migration have been the following:

* The event handlers do not have access to the HttpContext anymore, so we used the `IHttpContextAccessor` (see BTCPayServer/Authentication/OpenId3/LogoutEventHandler.cs). Still unsure if it is good.
* In our code, somewhere we [were creating an AuthenticationTicket](https://github.com/btcpayserver/btcpayserver/pull/1070/files#diff-16b21d27631a1db3d3202d86e4f7568aL125). Since the scopes are not attached to the ticket anymore, I am unsure how to migrate this code.
* The `options.UseJsonWebTokens();` from openiddict 2.0 disappeared. I don't know how to keep the behavior.

PinpointTownes:
> The event handlers do not have access to the HttpContext anymore, so we used the IHttpContextAccessor (see BTCPayServer/Authentication/OpenId3/LogoutEventHandler.cs). Still unsure if it is good.

As you probably figured out, it's a deliberate change, to make `OpenIddict.Server` completely host-agnostic. What you have is fine, but you can access the ASP.NET Core context by using `context.Transaction.GetHttpRequest().HttpContext` (which doesn't use `IHttpContextAccessor` internally, so no need to register it).

> In our code, somewhere we were creating an AuthenticationTicket. Since the scopes are not attached to the ticket anymore, I am unsure how to migrate this code.

All the extensions have been moved to `ClaimsPrincipal`, as `AuthenticationTicket` is an ASP.NET Core type we couldn't depend on.

> The options.UseJsonWebTokens(); from openiddict 2.0 disappeared. I don't know how to keep the behavior.

JWT is now the default token format for all types of tokens (authorization codes and refresh tokens are always encrypted using JWE, so don't try to read them using jwt.io 😄).

NicolasDorier:
Thanks for your help, @PinpointTownes, another confusing part: Should we use `OpenIddictServerAspNetCoreDefaults.AuthenticationScheme` or `OpenIdConnectDefaults.AuthenticationScheme` as authentication scheme?

NicolasDorier:
@PinpointTownes I must say, I am a bit afraid about our use of OpenIddict. I don't understand enough the design, and I fear there is huge security hole in our code as a result.

Right now we don't use OpenIddict for anything important, but we will. Can I pay for your time later on so you can make a review of our OpenIddict code once I finished completely the 3.0 migration? If yes, contact me on nicolas.dorier@gmail.com

PinpointTownes:
> Should we use OpenIddictServerAspNetCoreDefaults.AuthenticationScheme or OpenIdConnectDefaults.AuthenticationScheme as authentication scheme?

`OpenIddictServerAspNetCoreDefaults.AuthenticationScheme` when you want to return an OIDC authorization or token response (with a token).

`OpenIdConnectDefaults.AuthenticationScheme` when you want to trigger an OIDC authentication dance from an OIDC ASP.NET Core client.

NicolasDorier:
We still need https://github.com/madskristensen/BundlerMinifier/pull/443 

I am unsure if I will migrate to 3.0 since 3.1 is LTS and should release next week. But at least I brought the gap to make .NET 3.0 work.

NicolasDorier:
I officially give up on .netcoreapp3.0

I keep this PR open just to have something to experiment to keep making the gap to cross smaller between netcore2.1.

Will reconsider on netcoreapp3.1

NicolasDorier:
There is still a bug in open iddict https://github.com/openiddict/openiddict-core/issues/834 that make most of the test fail. Bug I think there is also other bug not caught by tests related to MVC.

bolatovumar:
@NicolasDorier just wondering, have you abandoned this PR for now?

NicolasDorier:
@bolatovumar yeah, was waiting for .net 3.1, which is now out.
There was some bugs in OpenIddict which prevented me to continue, but those should now be fixed, and if not fixed,  I would just stop supporting it until it is fixed (we don't use it yet anyway)

PinpointTownes:
@NicolasDorier just to be clear, the "bug" mentioned earlier is not an OpenIddict regression, it's an EF Core breaking change (it no longer generates IDs for string primary keys by default). And it's certainly not blocking as you can fix it on your side by assigning the ID in the constructor (e.g https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore.Models/OpenIddictAuthorization.cs#L18-L22)

https://github.com/openiddict/openiddict-core/issues/834 tracks a possible future improvement consisting in calling `.ValueGeneratedOnAdd()` for you.

NicolasDorier:
@PinpointTownes I think the problematic classes are in OpenIddict assembly, so if you did it on your side, I should just be able to update assembly and it would work right?

PinpointTownes:
They are not. OpenIddict exposes the base generic types, but you own the final entities: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Data/Data/BTCPayOpenIdAuthorization.cs

It’s worth noting this "issue" is not specific to OpenIddict: you’d have the same issue with Identity if you inherited from their generic entities.

Kukks:
But you should be marking them, on the base types, inside OpenIddict?

PinpointTownes:
@Kukks marking? Not sure what you mean exactly.

PR #1077, opened 2019-10-08
Update to OpenIddict3.0

NicolasDorier:
As @PinpointTownes said he supported netcoreapp2.1 with OpenIddict3.0, we will start by migrating that before continuing the work to jump to netcorapp3.0 started in this PR https://github.com/btcpayserver/btcpayserver/pull/1070 .

I am still at loss migrating [this part](https://github.com/btcpayserver/btcpayserver/compare/master...NicolasDorier:migration/openiddict30?expand=1#diff-16b21d27631a1db3d3202d86e4f7568aL125).

Work in progress...


NicolasDorier:
With those change we get empty `OpenIddictResponse` even if password is correct.

BTCPayServer/Authentication/OpenId/ClientCredentialsGrantTypeEventHandler.cs
```diff
@@ -33,32 +35,29 @@ public class
...
             }
 
             var user = await _userManager.FindByIdAsync(application.ApplicationUserId);
-
-            notification.Context.Validate(await CreateTicketAsync(request, user));
-            // Don't allow other handlers to process the event.
-            return OpenIddictServerEventState.Handled;
+            await UpdateClaimsPrincipal(context.Principal, request, user);
+            context.HandleAuthentication();
+            context.HandleRequest();
```
PinpointTownes: You'll want to remove it since it's a way to tell OpenIddict not to apply its default logic to return the response (i.e when you want to completely take care of returning a custom response yourself).

BTCPayServer/Authentication/OpenId/OpenIdGrantHandlerCheckCanSignIn.cs
```diff
@@ -27,44 +34,41 @@ public abstract class
...
-                return OpenIddictServerEventState.Handled;
+                return;
             }
 
-            notification.Context.Validate(await CreateTicketAsync(request, user));
-            // Don't allow other handlers to process the event.
-            return OpenIddictServerEventState.Handled;
+            await this.UpdateClaimsPrincipal(notification.Principal, request, user);
+            notification.HandleAuthentication();
+            notification.HandleRequest();
```
PinpointTownes: You'll want to remove that too.

BTCPayServer/Authentication/OpenId/PasswordGrantTypeEventHandler.cs
```diff
@@ -8,56 +8,60 @@
...
-                // Don't allow other handlers to process the event.
-                return OpenIddictServerEventState.Handled;
+                return;
             }
 
-            notification.Context.Validate(await CreateTicketAsync(request, user));
-            // Don't allow other handlers to process the event.
-            return OpenIddictServerEventState.Handled;
+            await UpdateClaimsPrincipal(notification.Principal, request, user);
+            notification.HandleRequest();
```
PinpointTownes: That too.

BTCPayServer/Authentication/OpenId/PasswordGrantTypeEventHandler.cs
```diff
@@ -8,56 +8,60 @@
...
                     description: "The specified credentials are invalid.");
-                // Don't allow other handlers to process the event.
-                return OpenIddictServerEventState.Handled;
+                return;
             }
 
-            notification.Context.Validate(await CreateTicketAsync(request, user));
-            // Don't allow other handlers to process the event.
-            return OpenIddictServerEventState.Handled;
+            await UpdateClaimsPrincipal(notification.Principal, request, user);
```
PinpointTownes: When using a non-derivative grant type like code or client credentials (i.e everything that doesn't use an existing authorization code or refresh token), `context.Principal` is always null and must be set with the principal containing the user claims.

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -118,26 +112,27 @@ await OpenIdExtensions.IsUserAuthorized(_authorizationManager, openIdConnectRequ
...
-            if (createAuthorization)
-            {
-                var application = await _applicationManager.FindByClientIdAsync(openIdConnectRequest.ClientId);
-                var authorization = await _authorizationManager.CreateAsync(User, user.Id, application.Id,
-                    type, ticket.GetScopes().ToImmutableArray(),
-                    ticket.Properties.Items.ToImmutableDictionary());
-                ticket.SetInternalAuthorizationId(authorization.Id);
-            }
+            //// Create a new authentication ticket.
+            //var ticket =
```
PinpointTownes: Instead of creating an `AuthenticationTicket`, you must create a `ClaimsPrincipal` so you'll want to update `OpenIdExtensions.CreateAuthenticationTicket` to return that instead of a ticket.

BTCPayServer/Controllers/RestApi/TestController.cs
```diff
@@ -19,7 +17,7 @@ namespace BTCPayServer.Controllers.RestApi
     /// </summary>
     [Route("api/[controller]")]
     [ApiController]
-    [Authorize(AuthenticationSchemes = OpenIddictValidationDefaults.AuthenticationScheme)]
+    [Authorize(AuthenticationSchemes = OpenIddictServerAspNetCoreDefaults.AuthenticationScheme)]
```
PinpointTownes: Use `OpenIddictValidationAspNetCoreDefaults` instead of `OpenIddictServerAspNetCoreDefaults` everywhere in this file. If you try to run this controller as-is, you'll get an exception indicating that no identity can be retrieved, since the OpenIddict server handler is not responsible of validating API requests for endpoints it doesn't manage.

BTCPayServer/Extensions/OpenIddictExtensions.cs
```diff
@@ -29,7 +29,9 @@ public static SecurityKey GetSigningKey(IConfiguration configuration)
         public static OpenIddictServerBuilder ConfigureSigningKey(this OpenIddictServerBuilder builder,
             IConfiguration configuration)
         {
-            return builder.AddSigningKey(GetSigningKey(configuration));
+            return builder
+                    .AddSigningKey(GetSigningKey(configuration))
+                    .AddEncryptionKey(GetSigningKey(configuration));
```
PinpointTownes: While it's not illegal to use the same RSA key for both encryption and signing, consider using 2 different keys.

If you want your access tokens to be encrypted, you can add a symmetric signing key (`SymmetricSecurityKey`) instead of a RSA key. For signing, you can also use an EC key instead of RSA (after all, ECDSA is the algorithm used by Bitcoin 😄).

PinpointTownes:
Not sure where the validation handler is registered as I didn't see it in this PR. In case you missed it:

```csharp
services.AddOpenIddict()
    .AddValidation(options =>
    {
        // Import the configuration from the local OpenIddict server instance.
        options.UseLocalServer();

        // Register the ASP.NET Core host.
        options.UseAspNetCore();
    });
```

Alternatively, you can use Microsoft's JWT handler if you prefer.

BTCPayServer/Authentication/OpenId/OpenIdGrantHandlerCheckCanSignIn.cs
NicolasDorier: Should I remove HandleAuthentication, HandleRequest or both ?

NicolasDorier: Brute forcing via tests seems to say I need to keep `HandleAuthentication`

NicolasDorier:
Ok, I think we are soon good!

One more error, the token does not seem to be validated. Trying to find if I can find more detail about why.
```
10/9/2019 4:00:11 AM +00:00 :PayServer:   A new user just registered e08b1d9f-2750-4242-9df1-6e0f03332734@toto.com 
10/9/2019 4:00:11 AM +00:00 :Configuration:   Should have sent email, but email settings are not configured
10/9/2019 4:00:14 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The token request was successfully extracted: {
  "grant_type": "password",
  "username": "e08b1d9f-2750-4242-9df1-6e0f03332734@toto.com",
  "password": "[removed for security reasons]",
  "client_id": "cc174c7f-fdf6-4942-8228-a84b1eeb1ace",
  "client_secret": "[removed for security reasons]"
}.
10/9/2019 4:00:14 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The token request was successfully validated.
10/9/2019 4:00:15 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The response was successfully returned as a JSON document: {
  "access_token": "[removed for security reasons]",
  "token_type": "Bearer",
  "expires_in": 3600
}.
10/9/2019 4:00:15 AM +00:00 :PayServer:   A new user just registered 41b92cc0-8ba7-4d29-8d1f-667ffa406c84@toto.com 
10/9/2019 4:00:15 AM +00:00 :Configuration:   Should have sent email, but email settings are not configured
10/9/2019 4:00:15 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The response was successfully returned as a JSON document: {
  "error": "invalid_token",
  "error_description": "The access token is not valid."
}.
10/9/2019 4:00:15 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was forbidden.
10/9/2019 4:00:16 AM +00:00 :PayServer:   Disconnected from WebSocket of NBXplorer (BTC)
10/9/2019 4:00:16 AM +00:00 :PayServer:   Disconnected from WebSocket of NBXplorer (LTC)
10/9/2019 4:00:16 AM +00:00 :PayServer:   Lightning listened stopped
```

NicolasDorier:
Debug logs:

```
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The token request was successfully extracted: {
  "grant_type": "password",
  "username": "4d2f81ac-5784-4098-aab2-f8ebbcd2563f@toto.com",
  "password": "[removed for security reasons]",
  "client_id": "9e09f548-83e7-4bc7-8494-203a10ea9be5",
  "client_secret": "[removed for security reasons]"
}.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The token request was successfully validated.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   Authentication was handled in user code.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   'AspNet.Identity.SecurityStamp' was excluded from the access token claims.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The response was successfully returned as a JSON document: {
  "access_token": "[removed for security reasons]",
  "token_type": "Bearer",
  "expires_in": 3599
}.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The request was handled in user code.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The request was handled in user code.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Server.OpenIddictServerProvider:   The request was handled in user code.
10/9/2019 4:07:22 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/9/2019 4:07:23 AM +00:00 :PayServer:   A new user just registered d01f3ca2-5150-4901-9ba3-ef88ce10927a@toto.com 
10/9/2019 4:07:23 AM +00:00 :Configuration:   Should have sent email, but email settings are not configured
10/9/2019 4:07:23 AM +00:00 :BTCPayServer.Security.Bitpay.BitpayAuthenticationHandler:   AuthenticationScheme: Bitpay.Auth was not authenticated.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The response was successfully returned as a JSON document: {
  "error": "invalid_token",
  "error_description": "The access token is not valid."
}.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The request was handled in user code.
10/9/2019 4:07:23 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was forbidden.
10/9/2019 4:07:23 AM +00:00 :PayServer:   Disconnected from WebSocket of NBXplorer (BTC)
```

NicolasDorier:
Our handlers are registered in the `AddServer`, so we should not have to change that right?

BTCPayServer/Authentication/OpenId/OpenIdGrantHandlerCheckCanSignIn.cs
PinpointTownes: Yep, remove just `HandleRequest()` and `SkipRequest()` everywhere you're currently using them.

NicolasDorier:
Wow thanks to look into it, @PinpointTownes you need to run the `docker-compose up dev` in `BTCPayServer.Tests`. This setup all the deps.

NicolasDorier:
> Anyway, I reintroduced the log traces in the last version (3.0.0-alpha1.19509.59), so it should be easier to figure out what's going on. Make sure you change the log level to Trace.

Will report the logs.

NicolasDorier:
```
Microsoft.AspNetCore.Mvc.ModelBinding.UnsupportedContentTypeFilter, BTCPayServer.Filters.XFrameOptionsAttribute, BTCPayServer.Filters.XContentTypeOptionsAttribute, BTCPayServer.Filters.XXSSProtectionAttribute, BTCPayServer.Filters.ReferrerPolicyAttribute, BTCPayServer.HostedServices.ContentSecurityPolicyCssThemeManager (Order: 1001)
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Execution plan of exception filters (in the following order): None
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Execution plan of result filters (in the following order): Microsoft.AspNetCore.Mvc.ViewFeatures.Internal.SaveTempDataFilter
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: Before executing OnAuthorizationAsync on filter BTCPayServer.Security.BTCPayClaimsFilter.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: After executing OnAuthorizationAsync on filter BTCPayServer.Security.BTCPayClaimsFilter.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: Before executing OnAuthorizationAsync on filter Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The self-contained JWT token 'eyJhbGciOiJSUzI1NiIsImtpZCI6IjdYUl9OVEY5NlFMSDhBSkVMNElTS1hOTlNVUEotQVpRODZFTERCQ0MiLCJ0eXAiOiJKV1QifQ.eyJzdWIiOiI0Njk3MGM0OS1hNjRkLTQ3OGItYTI1Yy1lMzE5NzA4OTZhYjEiLCJuYW1lIjoiM2IxMWFlMjktOTY5OS00ZDUzLTg0NzItMGQ4Y2NkOWQ4YTVkQHRvdG8uY29tIiwianRpIjoiNmM3NjcyMzAtYzhkOC00NjM1LWJjOWMtYjI0MmRlOWUzNjI2IiwiaWF0IjoxNTcwNjY5NzUyLCJleHAiOjE1NzA2NzMzNTIsImF6cCI6IjkwODc1MGI3LWFlYjMtNDljZi1iYTBlLTk3MzE4MDRjNGFjMCIsIm9pX3Rrbl91c2UiOiJhY2Nlc3NfdG9rZW4iLCJpc3MiOiJodHRwOi8vMTI3LjAuMC4xOjYxMTQwLyJ9.mnLcRXxHHaKdQLyFFEzzfhA_NxPKLKromub9yGAm4BYgDLBMgZUPPQ7y6WD94UXZriwrDyOFzGHe6x6pLpALxQ_6j0LigYyV2iK_rZYRRae2xL2TCnBV_SqP8ZAOXTGrNtLYOwtW7TSUh8suO0m91vyv_5go5skjY99FvVxM5gSRA7AbVuFx_VzlSa3Zsj1KfacjGGGqGkSQ8swd29COvQvraCZPzNf3umdIU1SNLtkAbpKK8A0bFyZLf4nPFzzbWEDxyNhCu-mUQ9J_DOpvpVvW7GZmVcGLgLb5DFGQ070nPCspbDaFughgahHS0OuaueJBZ51YjE35c6l0ygtOJg' was successfully validated and the following claims could be extracted: sub: 46970c49-a64d-478b-a25c-e31970896ab1, name: 3b11ae29-9699-4d53-8472-0d8ccd9d8a5d@toto.com, jti: 6c767230-c8d8-4635-bc9c-b242de9e3626, iat: 1570669752, exp: 1570673352, azp: 908750b7-aeb3-49cf-ba0e-9731804c4ac0, oi_tkn_use: access_token, iss: http://127.0.0.1:61140/.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Authorization.DefaultAuthorizationService:   Authorization was successful.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: After executing OnAuthorizationAsync on filter Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: Before executing OnAuthorizationAsync on filter Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was successfully authenticated.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Authorization.DefaultAuthorizationService:   Authorization failed.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization Filter: After executing OnAuthorizationAsync on filter Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Authorization failed for the request at filter 'Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter'.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Before executing action result Microsoft.AspNetCore.Mvc.ForbidResult.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.ForbidResult:   Executing ForbidResult with authentication schemes (OpenIddict.Validation.AspNetCore).
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionManager:   Scanning connections.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionManager:   Scanned connections in 00:00:00.0000992.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The response was successfully returned as a JSON document: {
  "error": "invalid_token",
  "error_description": "The access token is not valid."
}.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.OpenIddictValidationProvider:   The request was handled in user code.
10/10/2019 1:09:13 AM +00:00 :OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler:   AuthenticationScheme: OpenIddict.Validation.AspNetCore was forbidden.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   After executing action result Microsoft.AspNetCore.Mvc.ForbidResult.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker:   Executed action BTCPayServer.Controllers.RestApi.TestController.CanEdit (BTCPayServer) in 65.7271ms
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Server.Kestrel:   Connection id "0HLQD3POF5OLN" completed keep alive response.
10/10/2019 1:09:13 AM +00:00 :Microsoft.AspNetCore.Hosting.Internal.WebHost:   Request finished in 77.1215ms 401 application/json;charse
```

NicolasDorier:
The authorize on the control should be good:

```
[Authorize(AuthenticationSchemes = OpenIddictValidationAspNetCoreDefaults.AuthenticationScheme)]
```

NicolasDorier:
ah one moment we have also at the route level

```
[Authorize(Policy = Policies.CanModifyStoreSettings.Key,
            AuthenticationSchemes = OpenIddictValidationAspNetCoreDefaults.AuthenticationScheme)]
```

Taking a look.

NicolasDorier:
I think that it is our problem with the `Policies`. Other routes seem to work fine...

NicolasDorier:
So it seems the scope are not included in the accessToken... almost there!

NicolasDorier:
@Kukks now I am confused. I don't see how the test could pass before,

1. `CanUseNonInteractiveFlows` is not requesting the ManageStore scope, but still the user could access the route
2. The CanEdit had the wrong policy attached to it.

PinpointTownes:
I took a brief look and it seems you are currently using the JWT bearer handler and not the OpenIddict validation handler: https://github.com/btcpayserver/btcpayserver/blob/5025e0dd4ddd6d5ab459cbe3c6aa947bb882b5f1/BTCPayServer/Hosting/BTCPayServerServices.cs#L277-L325

Could that explain the behavior difference?

In 2.x, both the JWT and OpenIddict handlers used the `Bearer` scheme, so if you used `OpenIddictValidationDefaults.AuthenticationScheme`, it could also point to the JWT handler.

PinpointTownes:
I opened https://github.com/openiddict/openiddict-core/issues/826 to track future improvements regarding the error description we return for forbidden responses (in this case, the access token is not exactly invalid, it's just unsuitable for the requested operation, as authorization doesn't pass)

NicolasDorier:
@PinpointTownes I finallly found out the cause:

While the token seems to be validated, `OnTokenValidated` never get called.
```csharp
 services.AddAuthentication()
                .AddJwtBearer(options =>
                {
                    //Disabled so that Tor works witt JWT auth
                    options.RequireHttpsMetadata = false;
                    options.TokenValidationParameters.ValidateAudience = false;
                    //we do not validate the issuer directly because btcpay can be accessed through multiple urls that we cannot predetermine
                    options.TokenValidationParameters.ValidateIssuer = false;
                    options.TokenValidationParameters.IssuerSigningKey =
                        OpenIddictExtensions.GetSigningKey(configuration, "signing.rsaparams");
                    options.TokenValidationParameters.TokenDecryptionKey = OpenIddictExtensions.GetSigningKey(configuration, "encrypting.rsaparams");
                    options.IncludeErrorDetails = true;
                    options.Events = new JwtBearerEvents()
                    {
                        OnTokenValidated = async context =>
                        {
                            var routeData = context.HttpContext.GetRouteData();
                            var identity = ((ClaimsIdentity)context.Principal.Identity);
                            if (context.Principal.IsInRole(Roles.ServerAdmin))
                            {
                                identity.AddClaim(new Claim(Policies.CanModifyServerSettings.Key, "true"));
                            }

                            if (context.HttpContext.GetStoreData() != null ||
                                !routeData.Values.TryGetValue("storeId", out var storeId))
                            {
                                return;
                            }
                            var userManager = context.HttpContext.RequestServices
                                .GetService<UserManager<ApplicationUser>>();
                            var storeRepository = context.HttpContext.RequestServices
                                .GetService<StoreRepository>();
                            var userid = userManager.GetUserId(context.Principal);

                            if (!string.IsNullOrEmpty(userid))
                            {
                                var store = await storeRepository.FindStore((string)storeId, userid);
                                if (store == null)
                                {
                                    context.Fail("Could not authorize you against store access");
                                }
                                else
                                {
                                    context.HttpContext.SetStoreData(store);
                                    identity.AddClaims(store.GetClaims());
                                }
                            }
                        }
                    };
                })
```

NicolasDorier:
@PinpointTownes ah you found the same things as me. Maybe we should just disable `AddJwtBearer` since openiddict take care of things?

I think our stuff was quite hackish.

PinpointTownes:
Yep, you could remove the JWT handler. To replace the `OnTokenValidated` event, you could either:
  - Add the claims to the access token when issuing the tokens (with the `Destinations.AccessToken` destination). It would make sense at least for the `CanModifyServerSettings.Key` claim.

  - Add the claims to the principal dynamically using the claims transformation mechanism: https://stackoverflow.com/questions/50296369/asp-net-coreclaims-transformation-and-authorization-policy

NicolasDorier:
Glad to get ride of a huge uncomprehensible blob :)

BTCPayServer/Authentication/OpenId/OpenIdGrantHandlerCheckCanSignIn.cs
```diff
@@ -1,22 +1,29 @@
...
 using OpenIddict.Server;
+using Microsoft.AspNetCore;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using OpenIddict.Server.AspNetCore;
 
 namespace BTCPayServer.Authentication.OpenId
 {
-    public abstract class
-        OpenIdGrantHandlerCheckCanSignIn : BaseOpenIdGrantHandler<OpenIddictServerEvents.HandleTokenRequest>
+    public class OpenIdGrantHandlerCheckCanSignIn :
```
Kukks: how come this isn't abstract anymore?

BTCPayServer/Authentication/OpenId/PasswordGrantTypeEventHandler.cs
```diff
@@ -8,56 +8,59 @@
...
-                return OpenIddictServerEventState.Unhandled;
+                return;
             }
 
-            // Validate the user credentials.
-            // Note: to mitigate brute force attacks, you SHOULD strongly consider
-            // applying a key derivation function like PBKDF2 to slow down
-            // the password validation process. You SHOULD also consider
-            // using a time-constant comparer to prevent timing attacks.
+            var httpContext = notification.Transaction.GetHttpRequest().HttpContext;
```
Kukks: cool with theb throttle in the event handler!

BTCPayServer/Security/ClaimTransformer.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Authentication;
+using BTCPayServer.Authentication;
+
+namespace BTCPayServer.Security
+{
+    public class ClaimTransformer : IClaimsTransformation
```
Kukks: This is awesome

BTCPayServer/Hosting/Startup.cs
```diff
@@ -198,36 +181,29 @@ private void ConfigureOpenIddict(IServiceCollection services)
...
-                    options.AddEventHandler(RefreshTokenGrantTypeEventHandler.Descriptor);
+                    options.AddEventHandler(OpenIdGrantHandlerCheckCanSignIn.Descriptor);
                     options.AddEventHandler(ClientCredentialsGrantTypeEventHandler.Descriptor);
                     options.AddEventHandler(LogoutEventHandler.Descriptor);
-#endif
                     options.ConfigureSigningKey(Configuration);
+                })
+                .AddValidation(options =>
+                {
+                    options.UseLocalServer();
```
Kukks: We need to verify that this jwt handler will work when you have multiple domains pointing to the server. 
The following settings were ion the jwt handler:
```
                    options.RequireHttpsMetadata = false;
                    options.TokenValidationParameters.ValidateAudience = false;
                    //we do not validate the issuer directly because btcpay can be accessed through multiple urls that we cannot predetermine
                    options.TokenValidationParameters.ValidateIssuer = false;
```

BTCPayServer/Authentication/OpenId/OpenIdGrantHandlerCheckCanSignIn.cs
NicolasDorier: Because we could replace 2 class by just 1 line in `OpenIdGrantHandlerCheckCanSignIn`.

NicolasDorier:
Ok almost everything pass now! Only need to fix selenium tests.

BTCPayServer/Hosting/Startup.cs
NicolasDorier: In openiddict 3.0 the whole JWT from identiy stuff is skipped.
Previously, I guess we were doing some double validation.

BTCPayServer/Authentication/OpenId/LogoutEventHandler.cs
```diff
@@ -21,19 +29,19 @@ public class LogoutEventHandler : BaseOpenIdGrantHandler<OpenIddictServerEvents.
...
             // when the user agent is redirected from the external identity provider
             // after a successful authentication flow (e.g Google or Facebook).
             await _signInManager.SignOutAsync();
 
             // Returning a SignOutResult will ask OpenIddict to redirect the user agent
             // to the post_logout_redirect_uri specified by the client application.
-            await notification.Context.HttpContext.SignOutAsync(OpenIddictServerDefaults.AuthenticationScheme);
-            notification.Context.HandleResponse();
-            return OpenIddictServerEventState.Handled;
+            await httpContext.SignOutAsync(OpenIddictServerAspNetCoreDefaults.AuthenticationScheme);
```
NicolasDorier: ping @PinpointTownes here if I understand it should be `OpenIddictValidationAspNetCoreDefaults` right?

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -118,26 +112,27 @@ await OpenIdExtensions.IsUserAuthorized(_authorizationManager, openIdConnectRequ
                 default:
                     // Notify OpenIddict that the authorization grant has been denied by the resource owner
                     // to redirect the user agent to the client application using the appropriate response_mode.
-                    return Forbid(OpenIdConnectDefaults.AuthenticationScheme);
+                    return Forbid(OpenIddictServerAspNetCoreDefaults.AuthenticationScheme);
```
NicolasDorier: @PinpointTownes here it should be `OpenIddictServerAspNetCoreDefaults` because this is the not about validating the token, but validating the grant to the token right?

BTCPayServer/Authentication/OpenId/LogoutEventHandler.cs
PinpointTownes: Yep. Or you can simplify things and let OpenIddict do the logout redirection stuff by removing L. 43 and L. 44.

Note: if you're concerned about unsolicited logouts (which can be considered as a form of DOS), you can instead handle that in MVC and render a logout confirmation page.

BTCPayServer/Controllers/AuthorizationController.cs
PinpointTownes: Yep, `OpenIddictServerAspNetCoreDefaults.AuthenticationScheme` is what you need here.

BTCPayServer/Controllers/AuthorizationController.cs
```diff
@@ -74,26 +70,26 @@ public async Task<IActionResult> Authorize(OpenIddictRequest openIdConnectReques
 
             var userId = _userManager.GetUserId(User);
             if (!string.IsNullOrEmpty(
-                await OpenIdExtensions.IsUserAuthorized(_authorizationManager, openIdConnectRequest, userId, application.Id)))
+                await OpenIdExtensions.IsUserAuthorized(_authorizationManager, request, userId, application.Id)))
             {
-                return await Authorize(openIdConnectRequest, "YES", false);
+                return await Authorize("YES", false);
```
Kukks: Is this right?

NicolasDorier: it seems so, why not? The request is taken form the httpcontext.

NicolasDorier:
Making the tests pass. I think I broke something somewhere.

NicolasDorier:
@PinpointTownes 
The `Set/GetScopes` seems to set/get `OpenIddictConstants.Claims.Private.Scopes` in the claim the openiddict server creates. However, in my authorization handler the claim on the validation, the claim I have is `scopes` instead, so the OpenIddictExtensions.HasScope fails. What is going on?

NicolasDorier:
So I looked around and you seem to strip private claims on https://github.com/openiddict/openiddict-core/blob/204b87dbeaf693bb5cf1994fa3d443f72acd1594/src/OpenIddict.Server.DataProtection/OpenIddictServerDataProtectionFormatter.cs#L224

So how can I check in a clean way that the principal has the expected scope?

NicolasDorier:
YEAH BABY ALL TESTS PASS!

NicolasDorier:
Waiting for the official nuget release and I pull the plug! :)))

PinpointTownes:
Actually, it has nothing to do with the Data Protection formatter (that you don't use since you only use the default JWT token format).

OpenIddict stores granted scopes as multiple private claims `oi_scp` (that are automatically serialized as JSON arrays in JWT tokens). These claims - and the extensions on top of them - are not meant to be used by your app.

The "standard" way to represent the granted scopes in access tokens is to have a single claim, with space-separated values: https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-02#section-2.2.2

This special claim is added by this handler: https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.Server/OpenIddictServerHandlers.cs#L1480-L1484

I agree this is not really "pretty", but that's what the specification requires. Splitting the string is the correct thing to do here. If you're concerned about performance (more precisely allocations), you can use `StringTokenizer` instead of `string.Split`.

Now, I'm aware the fact we expose these extensions makes the UX poor. I'm not sure what do yet, but I'll explore multiple options (removing these extensions, making them pubternal, checking for both claim types...)

PinpointTownes:
> Waiting for the official nuget release and I pull the plug! :)))

I currently don't have any firm date for the alpha1 release. If you want to migrate to ASP.NET Core 3.0 AND keep using a released version, you can use OpenIddict 2.0.1, that was released yesterday and includes workarounds to work with EF Core 3.0.

NicolasDorier:
@PinpointTownes for now I just want to migrate to openiddict3.0 without .net core 3.0. But I wish I could remove the nuget.config changes.

NicolasDorier:
I will merge it for now. I think this is not a big deal.

PR #1085, opened 2019-10-12
Update launchSettings.json

Eskyee:
I`m  the Removing Docker-Regtest-https-monero as this loads as a default, which I didn't ask for, it  default in my builds config, launchSettings json, also I think since this json was changed, it gave my Mac VS a indent bug somehow, ?? But i may be wrong there, 
can Nicolas please review ?? this is the correct way I would remove Docker-Regtest-https-monero from my folk ?? thanks


esky33kubernetes:
❤️

Eskyee:
😋

PR #1086, opened 2019-10-12
Refactor authorizations

NicolasDorier:
After talking with @PinpointTownes, it became clear that the way btcpay was handling authorization was really hacky.

The attempt to use `AuthorizationHandlers` and `Policy` instead of claims.


BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -185,8 +187,14 @@ public partial class WalletsController : Controller
             return RedirectToAction(nameof(WalletTransactions), new { walletId = walletId.ToString() });
         }
 
+        [HttpGet]
+        [AllowAnonymous]
```
Kukks: Why mark it as public and then check for a user as the first thing?

BTCPayServer/Security/BitpayAuthorizationHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+                                StoreRepository storeRepository)
+        {
+            _HttpContext = httpContextAccessor.HttpContext;
+            _storeRepository = storeRepository;
+        }
+        protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context, PolicyRequirement requirement)
+        {
+            if (context.User.Identity.AuthenticationType != Policies.BitpayAuthentication)
+                return;
+            var storeId = context.User.Claims.Where(c => c.Type == Claims.StoreId).Select(c => c.Value).FirstOrDefault();
```
Kukks: How is this claim set on the user?

BTCPayServer/Security/OpenIdAuthorizationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+                                UserManager<ApplicationUser> userManager,
+                                StoreRepository storeRepository)
+        {
+            _HttpContext = httpContextAccessor.HttpContext;
+            _userManager = userManager;
+            _storeRepository = storeRepository;
+        }
+        protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context, PolicyRequirement requirement)
+        {
+            if (context.User.Identity.AuthenticationType != "AuthenticationTypes.Federation")
```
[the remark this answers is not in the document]
NicolasDorier: I was wondering, got response on https://github.com/openiddict/openiddict-core/issues/829

BTCPayServer/Services/Claims.cs
```diff
@@ -10,6 +10,7 @@ public class Claims
     {
         public const string SIN = "BITID_SIN";
 
-        public const string OwnStore = "BTCPAY_OWN_STORE";
+        public const string StoreId = "BTCPAY_STOREID";
+        public const string ApiKey = "BTCPAY_APIKEY";
```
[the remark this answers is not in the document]
NicolasDorier: The Bitpay AuthenticationHandler is checking if the bitpay authentication is using:

1. API Key or BitId SIN
2. Is attached or not to a store

To pass both info to the authorization handler, I use claims in the Principal.
This might be hacky, so I will check if instead I can stop relying on those claims and directly do the checks in the authorization handlers.

BTCPayServer.Tests/CheckoutUITests.cs
```diff
@@ -63,7 +41,12 @@ public async Task CanHandleRefundEmailForm()
                 s.Driver.AssertElementNotFound(By.Id("emailAddressFormInput"));
                 s.GoToHome();
                 var invoiceId = s.CreateInvoice(store.storeName);
-                s.GoToInvoiceCheckout(invoiceId);
+                s.Driver.FindElement(By.ClassName("invoice-details-link")).Click();
```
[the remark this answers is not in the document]
NicolasDorier: This is the detail link, not the checkout link. (I am just making sure I can browse the invoice details page)

BTCPayServer/Controllers/WalletsController.cs
NicolasDorier: Because at the controller level, I normally use `[Authorize(Policy = Policies.CanModifyStoreSettings.Key, AuthenticationSchemes = Policies.CookieAuthentication)]`.

But `CanModifyStoreSettings` can only work if you are in a store specific page (like `wallets/{walletId}/send`)
`ListWallet` however is not tied to a specific store, which mean it can't satisfy `Policies.CanModifyStoreSettings.Key`.

I thought just using
`[Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]`, but sadly this does not override the controller's permission. So  instead, I just remove authentication and check that we have a `UserId`.

The is the same thing I did for `ListStores` in the store controller.
I think a better solution might be to put `ListStores` and `ListWallets` in a common controller with `[Authorize(AuthenticationSchemes = Policies.CookieAuthentication)]`

BTCPayServer/Security/BitpayAuthorizationHandler.cs
NicolasDorier: At `BitpayAuthenticationHandler`. As said above, it smells bad so I will check if I can't solve it.

NicolasDorier:
@Kukks it is not WIP anymore. If you can give it another quick review.
I will sleep on it and review myself again tomorrow, if good I will merge, rebase openiddict30 and wait for @PinpointTownes to officially release openiddict3.0.

PR #1092, opened 2019-10-17
Detection of seed walletunlock.json LND autoinit file

rockstardev:
New deployments of LND will be auto initialized and have `walletunlock.json` file that will preserve seed and wallet unlock password. See this PR: https://github.com/btcpayserver/lnd/pull/4

BTCPayServer's detection of whether LND has seed backup / auto-unlock file will depend on new setting `BTCPAY_LND_SEEDPATH`. If setting is present we'll try to check if file exists and read it to parse & ensure format is right. If everything is correct we won't show warning about seedless LND deployment.

Otherwise user will be presented with big warning dialog explaining the migration path from legacy deployment with `noseedbackup=1` to new deployment that auto-inits the seed and auto-unlocks the wallet.

image

### DIFF
--- a/BTCPayServer/Services/Lightning/LndMigrationHelper.cs
+++ b/BTCPayServer/Services/Lightning/LndMigrationHelper.cs
@@ -20,35 +20,59 @@ namespace BTCPayServer.Services.Lightning
 
         public LndSeedFile UnlockFile { get; set; }
 
-        public bool IsSeedlessLnd { get; set; }
-        public async Task PerformDetection(string seedFilePath)
+        public bool IsSeedlessLndNotice { get; set; }
+        public bool PerformDetection(BTCPayServerOptions _BtcpayServerOptions, string connStr, string cryptoCode)
         {
-            // if seed file path is empty, show no warning
-            if (String.IsNullOrEmpty(seedFilePath))
-                IsSeedlessLnd = false;
+            // by default we don't show seedless LND notice
+            IsSeedlessLndNotice = false;
 
-            // by default all LND instalations were started as seedless
-            IsSeedlessLnd = true;
+            var isLndConnString = !String.IsNullOrEmpty(connStr) &&
+                (connStr.Contains("lnd-rest", StringComparison.OrdinalIgnoreCase) ||
+                connStr.Contains("lnd-grpc", StringComparison.OrdinalIgnoreCase));
+            if (!isLndConnString)
+            {
+                // if lightning connection string is not LND, return
+                return false;
+            }
+
+            if (!_BtcpayServerOptions.LndSeedPath.ContainsKey(cryptoCode))
+            {
+                // if LNDSEEDPATH is not set for crypto, don't run check
+                return false;
+            }
+                       
+            var lndSeedFilePath = _BtcpayServerOptions.LndSeedPath[cryptoCode];
+            if (String.IsNullOrEmpty(lndSeedFilePath))
+            {
+                // if LNDSEEDPATH is empty, don't run check
+                return false;
+            }
+
             
+            // if we got this far, it means we have all settings and it's LND connection string
+            // legacy settings initalized LND as seedless, so to get IsSeedlessLnd = false, we need to do check
+            IsSeedlessLndNotice = true;
             try
             {
                 // with new setting we'll try to check if unlock file existits and is in right format
                 // if that's true and parsing passes, we know this is not seedless LND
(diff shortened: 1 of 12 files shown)


rockstardev:
Example `walletunlock.json`:

```
{"wallet_password":"hellorockstar", "cipher_seed_mnemonic":["about","over","million","coast","nature","gold","fabric","aware","adult","salute","sock","envelope","oblige","shed","basic","need","analyst","spot","worth","garbage","planet","genre","leave","cereal"]}
```

BTCPayServer/Services/Lightning/LndMigrationHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+        public async Task PerformDetection()
+        {
+            // by default all LND instalations were started as seedless
+            IsSeedlessLnd = true;
+
+            try
+            {
+                // with new setting we'll try to check if unlock file existits and is in right format
+                // if that's true and parsing passes, we know this is not seedless LND
+                var seedFilePath = Environment.GetEnvironmentVariable("BTCPAY_LND_SEEDPATH");
```
NicolasDorier: need to use BTCPayOptions stuff

NicolasDorier:
1. I think in can be simplified. The `LndMigrationHelper` does not need to exist, can just be a private method in the `ServerController`.
2. Should use the `BTCPayServerOptions` so people cna use command line as well like all other options.
3. If BTCPAY_SEED_FILE is empty, it should not show any warning.

rockstardev:
Updating PR as soon as I'm done with travel

rockstardev:
Switching to BTCPayServerOptions per feedback, support per cryptoCode.

I've left LndMigrationHelper because it does help with preserving cached value so we don't need to recalculate it in POST or later at other spots where we may depend on the cached value.

If we'll expose Seed in server settings, we may do this check on start.

BTCPayServer/Views/Stores/AddLightningNode.cshtml
```diff
@@ -69,6 +69,16 @@
...
+
+                @if (Model.IsSeedlessLnd)
+                {
+                <div class="alert alert-danger alert-dismissible" role="alert">
+                    <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+                    <p>You have old version of LND deployment that was auto-initialized using `noseedbackup=1`.</p>
+                    <p>Please migrate to new version of LND deployment that is auto-initialized through script which creates seed backup file as part of the startup process.</p>
+                    <p><a href="https://github.com/btcpayserver/lnd/pull/4" target="_blank">Click here for more information</a></p>
+                </div>
+                }
```
NicolasDorier: I think this should be inside `server settings / lnd services` because only the admin can do something about it.

NicolasDorier:
and need test to pass :p

rockstardev:
Need to see why those other tests for me are failing... stopped looking at test results because of that.

NicolasDorier:
@rockstardev you can rebase, I tried to stabilize the tests a bit. I think there is two error on those tests which are not really related to this PR.

BTCPayServer/Views/Stores/AddLightningNode.cshtml
rockstardev: I think the best option would be in both places... store owner can also see notice and notify server admin. But for now I would not complicate PR further but rather proceed like this... go toward release... it's just notice UI.

NicolasDorier:
This message must NOT show up in AddLightningNode but in `Server Settings / LND`.

The user of `AddLightningNode` is not necessarily the owner of the internal node and would just be confused by the message.

NicolasDorier:
So here is what I thought:

1. No warning should appear in `AddLightningNode`, because the viewer of this page is not necessarily the owner of the internal node and would be very confused about what the warning is about.
2. No warning should appear anywhere.

Instead, the simplest I can see is adding a new service in `Server Settings / Services / Crypto Service` called LND Backup.

If the user go there with a old wallet, we can say this feature is disabled for him and redirect to a doc page which tell him the steps to do to clear his LND node.

If the user go there with a wallet with the seed, we should just have:
1. One button that can reveal the seed and password.
2. One button to delete the seed
3. An explanation to the LND doc about static channel backups and seed.

If the user go there with a wallet with the seed which has been deleted into the past, only the explanation `3.` should appear.

rockstardev:
OK, I'm breaking down this pull request into multiple PRs for easier integration.

Let's first add new LND with seed to solution (#1103) and I'll keep developing UI per instructions above.

PR #1104, opened 2019-10-29
Improve pay button display

dennisreimann:
These changes should provide a more solid CSS approach to display the pay button in different scenarios. Also it fixes minor functionality issues like preselecting the merchant-chosen currency in the custom amount and slider dropdown.


NicolasDorier:
hey @dennisreimann I am completely out of the loop concerning the javascript parts. Maybe the best is to ping @pavlenex  to test it or @Kukks  for review the js part.

PR #1107, opened 2019-10-31
Redesign confirm forms

NicolasDorier:
From now, all confirm form would have this design:

image

ACK?

### DIFF
--- a/BTCPayServer/Views/Shared/Confirm.cshtml
+++ b/BTCPayServer/Views/Shared/Confirm.cshtml
@@ -1,26 +1,56 @@
 ﻿@model ConfirmModel
+
+@inject BTCPayServer.HostedServices.NBXplorerDashboard dashboard
+@inject BTCPayServer.HostedServices.CssThemeManager themeManager
+@addTagHelper *, BundlerMinifier.TagHelpers
 @{
-    Layout = "_Layout.cshtml";
+    Layout = null;
 }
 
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Ping @pavlenex

pavlenex:
Concept ACK, will test this in a few hours.

pavlenex:
Design wise I like it. One thing that may be useful specifically when deleting a new store is some sort of back button. This way user has no choice where to click except `Delete` which looks like CTA. Of course they can go back in the browser, but still not a perfect UX.

britttttk:
ACK

BTCPayServer/Views/Shared/Confirm.cshtml
```diff
@@ -1,26 +1,55 @@
...
             </div>
-        </div>
-        @if (!String.IsNullOrEmpty(Model.Action))
-        {
-        <div class="row">
-            <div class="col-lg-12 text-center">
-                <form method="post">
-                    <button id="continue" type="submit" class="btn btn-secondary @Model.ButtonClass" title="Continue">@Model.Action</button>
-                </form>
+            <div class="modal-body">
```
britttttk: `text-center` here looks better to me

Example Without:
image

NicolasDorier:
Added a `Go back` button and using text-center.

PR #1108, opened 2019-10-31
More pay button improvements

dennisreimann:
Some more improvements and cleanups around the previous PR (see #1104) as well as some display fixes for the slider version.

The CSS is more maintainable this way, but it is [compressed for the output](https://github.com/btcpayserver/btcpayserver/compare/master...dennisreimann:paybutton?expand=1#diff-fd3646cf1a43c3bdf4bb5c980b64fe80R41), so that the styles do not take up too much space in the code the user sees.


pavlenex:
tACK

PR #1115, opened 2019-11-03
feat(stores): add ledger specific xpub help

escapedcat:
**problem**: maximum of 20 ledger accounts are being listed in the derivation-scheme select
**solution**: help users with i.e. multiple ledger accounts by providing detailed info on where to find the xpub key manually

proposed result:

collapsed:
image

uncollapsed:
image


NicolasDorier:
I like the idea. However, @pavlenex and @britttttk  are already  documenting how to find the xpub in the docs, so a link to the doc should be enough.

escapedcat:
Right, I saw the issue but wanted to propose something in place. But I agree the content could be moved to the docs and a link should be fine. Happy to adjust this if wanted or just close it.

pavlenex:
Concept ACK.

I don't mind altering docs as long as the changes improve experience for the end-users. I actually prefer solutions like this in the UI.

Not sure how will changes that @NicolasDorier is working on influence this though.

Just prior to merging public key stuff, always make sure to open an issue in docs so we can document it as it's sensitive part.

NicolasDorier:
@pavlenex why do you say "altering the docs?" I thought this was already in the docs, I definitely remember somewhere where it was documented how to get xpub on different wallets.

pavlenex:
@NicolasDorier I meant, don't you worry about the docs as long as you improve things for users. I'll alter the docs when needed, just ping me :)
In this case, @escapedcat already did the [PR in docs](https://github.com/btcpayserver/btcpayserver-doc/pull/358).

@escapedcat So Nicolas's suggestion is to simply leave the UI as it is, but provide a link to the docs so people can read it there. I've just merged the PR from docs, so you can link to it and I assume modify the existing PR.

escapedcat:
@NicolasDorier @pavlenex  

I updated the PR to point at the docs as suggested. Looks like this now, pointing to https://docs.btcpayserver.org/getting-started/connectwallet/ledgerwallet#manual-setup:

image

escapedcat:
Will do @pavlenex

NicolasDorier:
So I can merge that or there is still something to do?

escapedcat:
@NicolasDorier this is ready. I'll extend the docs regarding @pavlenex comment in the docs repo. Thanks.

Issue #1111, opened 2019-11-02
Pay Button Preview too low on page


britttttk:
**Is your feature request related to a problem? Please describe.**
No, but I think payment button higher on the page so the user does not have to scroll to the bottom of the page to preview their button selection(s) would be better for UX. Sometimes checking boxes like `Use a simple input style` I'm not even sure what changed because you don't see it change. 

**Describe the solution you'd like**
Move button preview above optional Payment Notifications / Advanced fields.

**Additional context**
Keep code at the bottom of page

bolatovumar:
This makes sense. I'm thinking something like this would be nice?

![Screen Shot 2019-11-02 at 8 53 08 PM](https://user-images.githubusercontent.com/1934678/68080172-d7eab500-fdb2-11e9-8ec8-4435667e1cdf.png)

pavlenex:
Done in #1116 🚀

PR #1116, opened 2019-11-03
Improve pay button display options section

bolatovumar:
address #1111


bolatovumar:
I moved the preview to the top where the button display options are. I also made some additional minor improvements by adding a display options section heading and making the pay button options checkbox styling to align with the rest of the app.

PR #1118, opened 2019-11-04
Liquid & Liquid Assets Support

Kukks:
~Requires https://github.com/dgarage/NBXplorer/pull/198 first~

This PR introduces support for the Liquid Network and its assets. This currently supports LBTC(the main asset, 1:1 with BTC) and Liquid Tether (USDT). 

Since Liquid seems to break the PSBT support we have, the wallet section for them is disabled.


BTCPayServer.Tests/BTCPayServerTester.cs
```diff
@@ -109,13 +110,15 @@ public async Task StartAsync()
                 config.AppendLine($"bind=0.0.0.0");
             }
             config.AppendLine($"port={Port}");
-            config.AppendLine($"chains=btc,ltc");
+            config.AppendLine($"chains=btc,ltc,lbtc,usdt");
```
NicolasDorier: add lbtc and usdt only for tests actually testing liquid.

NicolasDorier:
Need rebase. Please rebase and force push instead of merging.

Kukks:
> 
> 
> Need rebase. Please rebase and force push instead of merging.

rebased

BTCPayServer.Tests/BTCPayServerTester.cs
Kukks: done

BTCPayServer.Tests/BTCPayServerTester.cs
```diff
@@ -93,7 +94,7 @@ public TestDatabases TestDatabase
 
         public bool MockRates { get; set; } = true;
 
-        public async Task StartAsync()
+        public async Task StartAsync(bool addLiquid = false)
```
NicolasDorier: no parameter, just use a property that your test set before running Start. (will do the same for litecoin later)

Kukks: fixed

PR #1120, opened 2019-11-04
Fixing CircleCi to run external test only in repo context, small refactor of LndServiceViewModel

rockstardev:


rockstardev:
@Kukks the problem is how CircleCI runs tests... look at this PR as example - if I have Circle CI account it runs in the context of my account: https://github.com/rockstardev/btcpayserver/commits/cifixrefactor (if you expand `All checks have passed` you'll see tests are linked to my account)

Then there is also the branch on which you are executing... `external_tests` are only running on master:
https://github.com/btcpayserver/btcpayserver/blob/bf59114a00000284d14af2ee10cb35e376c37261/.circleci/config.yml#L98

This commit just ensures that if I say submit PR that's directly on `master` branch `external_tests` don't run in the context of my repository. Then when PR is merged to `master` on `btcpayserver` `external_tests` will finally run (because they have access to credentials) and throw error in case of problems.

PR #1121, opened 2019-11-05
Add sponsor and new design to the login and registration page

NicolasDorier:
This make a better register and login page with the sponsors advertised.

The new design would be:

image

The ideal design of ttyl is

image

I know we are far from that, but we need to move step by step

Because I am using bootstrap, this login/register page adapt nicely on mobile.


pavlenex:
It's great start in my opinion. Also what you said earlier, that `automatically redirecting to login if the user is not logged or to the register page if there is no user` is good approach towards nailing this down.

I think @dennisreimann @rockstardev and @radWorx are working on something as well, so I'd wait for them to test this approach.

### Things I think can be improved (now):
- Mobile view, which pushes register form below the fold, which is bad UX, if you can please put sponsors at the bottom on mobile
mobile

- Introductory text - I really don't like the current one. We need something better. I'd rather use what we already have (I'll work on making this better in the future)
```
Free, self-hosted and open-source payment processor for sovereign businesses and individuals.
```

### Things I think can be improved (later)
- Better intro text
- More consistent fonts

dennisreimann:
Yes, definitely better and the homepage needs some extra love. As @pavlenex said it would be good to tackle this after my css themeing branch is reviewed and merged. Will prepare it together with @rockstardev, so we can get it in quickly and iterate based on that.

pavlenex:
A follow up on my end,[ I've found a resource I've read months ago, ](https://opensource.com/open-organization/18/12/state-purpose-single-sentence)and will use it to state a purpose of BTCPay in a single sentence that can then be used not only on the front page, but in the dos, website, etc.

However wants to team up on this, hit me up on Mattermost.

moneyball:
ACK

NicolasDorier:
> if you can please put sponsors at the bottom on mobile

No I can't. Or at least I have no idea how to do. that in CSS, I am using bootstrap which does 95% of the things right without be having to think about anything. But small detail like this can take me easily a whole day of development. :(

@dennisreimann you can just dump what I did when your branch has been reviewed. We need to show the sponsors for next release, so I made the simplest PR which did the job.

rockstardev:
When this is merged, I can set aside an hour and do touchups like that display of supporters at the bottom... no worries.

NicolasDorier:
@pavlenex I noticed the mobile view has shit fonts oO

NicolasDorier:
I added automatic redirect to login if not logged (except if there is rooted apps) and redirect to register if that's the first use.

dennisreimann:
Ok, will rebase my branch. I could also take care of moving that section down on mobile.

Issue #1117, opened 2019-11-03
Add refresh page text to server update (response message)


britttttk:
**Is your feature request related to a problem? Please describe.**
No, but can we add "refresh browser" text to the update server response message (in server maintenance UI). 

**Describe the solution you'd like**
The server will restart soon if an update is available... refresh the page and log back in. 

**Describe alternatives you've considered**
Add tooltip or doc FAQ

**Provide examples**
Something similar to the wallet rescan message which prompts for page refresh "(Scanning in progress, refresh the page to see the progress...")

**Additional context**
If they refresh the page before the server is back up they will get an error, not sure if users will know to keep refreshing the page.

pavlenex:
Good idea. 
I'd suggest we add similar to the wallet re-scan page, it's very poor UX asking user to manually refresh the page.

PR #1125, opened 2019-11-06
Improve message in Maintainance page

NicolasDorier:
Trying to fix https://github.com/btcpayserver/btcpayserver/issues/1117


NicolasDorier:
ping @britttttk

NicolasDorier:
I don't wanted to write "You need to refresh this page" because then the noob would think that without refresh it will not update.

britttttk:
ack

PR #1126, opened 2019-11-06
WIP password for payment request

Kukks:
This adds a password feature to payment requests. Currently only lets you input password through query string but will work on a UI soon

### DIFF
--- a/BTCPayServer.Data/Data/PaymentRequestData.cs
+++ b/BTCPayServer.Data/Data/PaymentRequestData.cs
@@ -33,6 +33,8 @@ namespace BTCPayServer.Data
             public string EmbeddedCSS { get; set; }
             public string CustomCSSLink { get; set; }
             public bool AllowCustomPaymentAmounts { get; set; }
+
+            public string PasswordHash { get; set; }
         }
 
         public enum PaymentRequestStatus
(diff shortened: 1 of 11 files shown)


NicolasDorier:
Why a password?

Kukks:
If you do official company  invoices, there isually a lot of sensitive data on the payment request that becomes readily available to anyone as soon as the url is discovered. An optional password helps my paranoia just enough.

This feature was requested a few times when we introduced payment requests.

britttttk:
does not show payment requests for me in testing :/ always 404

Kukks:
Yeah I forgot to mark as WIP :)

britttttk:
> Why a password?

I am also confused because the credentials are already provided.

Kukks:
> > Why a password?
> 
> I am also confused because the credentials are already provided.

What do you mean?

A password isn't provided when you share the payment request url, you ll have to enter it before you can see it

NicolasDorier:
Payment requests are not discoverable. Password is redundant.

Kukks:
I still feel safer having a password to view it. If anyone gets access to the url, they can view too much confidential data. This way I can share the password via different channels. It also helps for recurring payment requests where they can have one password for all future payment requests(many companies I worked for used an identical system for payroll slips)

NicolasDorier:
It is not tested, never saw anybody complaining about it and it add a layer of code just because "it does not feel safe". I am not fan of this PR.

pavlenex:
While it's a bit paranoid, I don't see why people wouldn't password protect their payment requests. After all, a simple Google search for a particular address of instance may lead you to payment request with company details tied to it, provided that search engine bots are smart enough to index that page, no?

NicolasDorier:
> a simple Google search for a particular address of instance may lead you to payment request with company details tied to it, provided that search engine bots are smart enough to index that page, no?

no

Issue #1110, opened 2019-11-02
Make recommended fee confirmation target configurable


bolatovumar:
**Is your feature request related to a problem? Please describe.**
Now that we have an option to show recommended fee on checkout invoice, it would be nice to allow merchants to set the confirmation target they want. It is currently hardcoded at 20 blocks.

**Describe the solution you'd like**
Add an option to set confirmation target in blocks on the checkout experience page.

**Describe alternatives you've considered**
Hardcode some other value instead of 20. Probably not a good idea as there was a rationale for setting it to 20 blocks.

**Additional context**
This issue was brought to my attention by @astupidmoose. He is using this feature on CoinCards and would like the transaction to clear as quickly as possible.

astupidmoose:
Yea, 20 blocks do not work for customers who are looking for fast transactions. 

It would be best to either make this configurable or use a 1 block time. My understanding is 20 blocks is hardcoded for wallet users as most people can wait for money going to an exchange, but for a customer-facing section of btcpay it should be conveying the urgency

pavlenex:
Added in #1127

PR #1127, opened 2019-11-07
Make recommended fee confirmation target configurable

bolatovumar:
Address #1110

### DIFF
--- a/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
+++ b/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
@@ -40,6 +40,10 @@ namespace BTCPayServer.Models.StoreViewModels
         [Display(Name = "Show recommended fee")]
         public bool ShowRecommendedFee { get; set; }
 
+        [Display(Name = "Recommended fee confirmation target blocks")]
+        [Range(1, double.PositiveInfinity)]
+        public int RecommendedFeeBlockTarget { get; set; }
+
         [Display(Name = "Do not propose on chain payment if the value of the invoice is below...")]
         [MaxLength(20)]
         public string OnChainMinValue { get; set; }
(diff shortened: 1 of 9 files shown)


bolatovumar:
Thanks to @rockstardev for going over the codebase with me and helping with this.

BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
```diff
@@ -40,6 +40,10 @@ public class Format
         [Display(Name = "Show recommended fee")]
         public bool ShowRecommendedFee { get; set; }
 
+        [Display(Name = "Recommended fee confirmation target blocks")]
```
NicolasDorier: add `(in satoshi per byte)`

NicolasDorier:
Cool. The only bad point is that this also impact the network fee which will become way more expensive.

bolatovumar:
@rockstardev this is what i was trying to get at in our session but couldn't quite express as I wasn't sure about how the whole network fee thing works. I wanted to have the network fee still target 20 blocks (leave it hardcoded) but have the invoice fee be configurable. Maybe something we take a look at next time?

rockstardev:
This won't impact network fee because when those other controllers call GetFeeRateAsync(), they don't specify parameter and for them default is 20.

https://github.com/btcpayserver/btcpayserver/pull/1127/commits/0a9290a980f87096baaac2957c19611f72eccb52#diff-8a2c0e8027c709ab0fb4ed0996572b0cR39

Analyzing this further, I think we made mistake with introducing new setting - target fee rate for invoices should be whatever number of confirmations merchants has already set in their store.

I wish I have time to refactor this... will add it to the list of stuff to work on when I have time.

bolatovumar:
@rockstardev not sure what you mean by "whatever number of confirmations merchants has already set in their store"? There is currently no setting for how many blocks a merchant wants to have the recommended fee target. There is a setting for how many blocks a merchants wants to wait for before the invoice is considered paid but they may wanna set that value to be something different. These are not the same things the way I see it.

@NicolasDorier does rockstar's reasoning make sense to you?

NicolasDorier:
fixed with https://github.com/btcpayserver/btcpayserver/commit/163ef031b97e43a449c4b78ba97c211aecfcb91b

NicolasDorier:
and https://github.com/btcpayserver/btcpayserver/commit/9ff9377bc7314f2b7395e37a49b8c4b34f2803ae

bolatovumar:
Thanks @NicolasDorier, looks like we are good now.

PR #1136, opened 2019-11-08
LND Seed Backup service

rockstardev:
This PR (continuation of #1092) makes use of auto-init and unlock functionality and displays information fetched from LND docker container in UI. When tested and merged it'll allow us to migrate from `noseedbackup=1` setting.

LND Seed Backup service is now displayed in Server Settings

01-services

Legacy deployments will show warning and provide link that explains the migration process (sweeping funds from legacy deployment to on-chain address, starting new LND instance, sending funds back)

02-legacy-deployments

New deployments will show wallet password, provide option to reveal the seed and delete it

03-with-seed

Seed can be deleted from the server, since all we need is wallet unlock password for subsequent restarts

04-seed-deleted

Deleting the seed can be problem in situations where BTCPayServer doesn't have write permission on the path... in that case error message will be shown and server admin can delete the password

### DIFF
--- a/BTCPayServer/Configuration/ExternalConnectionString.cs
+++ b/BTCPayServer/Configuration/ExternalConnectionString.cs
@@ -144,7 +144,7 @@ namespace BTCPayServer.Configuration
         }
         public bool? IsOnion()
         {
-            if (!this.Server.IsAbsoluteUri)
+            if (this.Server == null || !this.Server.IsAbsoluteUri)
                 return null;
             return this.Server.DnsSafeHost.EndsWith(".onion", StringComparison.OrdinalIgnoreCase);
         }
--- a/BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
+++ b/BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
@@ -1,0 +1,13 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace BTCPayServer.Models.ServerViewModels
+{
+    public class LndSeedBackupViewModel
+    {
+        public bool IsWalletUnlockPresent { get; set; }
+
+        public string WalletPassword { get; set; }
+
+        public string[] Seed { get; set; }
+    }
+}
--- a/BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
+++ b/BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
@@ -1,4 +1,8 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.IO;
+using Newtonsoft.Json;
 
 namespace BTCPayServer.Models.ServerViewModels
 {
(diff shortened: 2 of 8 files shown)


BTCPayServer/Configuration/ExternalConnectionString.cs
```diff
@@ -144,7 +144,7 @@ public ExternalConnectionString Clone()
         }
         public bool? IsOnion()
         {
-            if (!this.Server.IsAbsoluteUri)
+            if (this.Server == null || !this.Server.IsAbsoluteUri)
```
NicolasDorier: you can keep it like this but another cool way is `this.Server?.IsAbsoluteUri is true`

NicolasDorier:
Cool, ACK for me.

I will work on a way to test more the `Server Settings` later.

NicolasDorier:
merge it if you want, or wait for other review as you want.

BTCPayServer/Configuration/ExternalConnectionString.cs
rockstardev: I kinda prefer this... to me it's cleaner than alternative (wish I could do it without `is true`)

BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+            catch
+            {
+            }
+
+            return new LndSeedBackupViewModel();
+        }
+
+        private class LndSeedFile
+        {
+            public string wallet_password { get; set; }
```
Kukks: PascalCase + `JsonProperty` attribute would have been nicer

BTCPayServer/Properties/launchSettings.json
```diff
@@ -11,7 +11,7 @@
                 "BTCPAY_BTCLIGHTNING": "type=charge;server=http://127.0.0.1:54938/;api-token=foiewnccewuify",
                 "BTCPAY_BTCEXTERNALLNDGRPC": "type=lnd-grpc;server=https://lnd:lnd@127.0.0.1:53280/;allowinsecure=true",
                 "BTCPAY_BTCEXTERNALLNDREST": "type=lnd-rest;server=https://lnd:lnd@127.0.0.1:53280/lnd-rest/btc/;allowinsecure=true;macaroonfilepath=D:\\admin.macaroon",
-                "BTCPAY_BTCLNDSEEDPATH": "/etc/merchant_lnd/data/chain/bitcoin/regtest/walletunlock.json",
+                "BTCPAY_BTCEXTERNALLNDSEEDBACKUP": "D:\\walletunlock.json",
```
Kukks: maybe make this read into a local volume dir map with the docker setup?

BTCPayServer/Views/Server/LndSeedBackup.cshtml
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+    </div>
+
+</div>
+
+@section Scripts {
+    <script type="text/javascript">
+        function confirmSeedDelete() {
+            var conf = confirm('This operation is not undoable. Are you sure you want to proceed with deleting seed from LND container?');
+            if (conf) {
```
Kukks: `return conf;`

BTCPayServer/Properties/launchSettings.json
NicolasDorier: no, follow what I did on https://github.com/btcpayserver/btcpayserver/commit/28396206de5a67bc60562861fb22dcda954b08a2 we don't want people to do special hack on their machine to debug or test it.

BTCPayServer/Models/ServerViewModels/LndSeedBackupViewModel.cs
rockstardev: I prefer it like this considering I'm keeping this class private and mapping properties to parent ViewModel.

BTCPayServer/Properties/launchSettings.json
rockstardev: done

BTCPayServer/Views/Server/LndSeedBackup.cshtml
rockstardev: I did some debugging and had that leftover, thanks for catching this.

rockstardev:
OK, considering review passed I am going to close this PR and push separate branch for production testing. Once production testing passes I'll merge to master and we push it out to in new version.

rockstardev:
Continuation here: https://github.com/btcpayserver/btcpayserver/commits/feature/lndseedbackup

NicolasDorier:
Why do you close the PR? The PR automatically update when you update your branch.

PR #1137, opened 2019-11-08
UI: Support themes via CSS variables

dennisreimann:
This PR adds support for custom themes based on CSS custom properties (variables). The main changes from a high level perspective:

- Added "Classic" and "Casa" theme files for containing the individual custom properties for colors and fonts
- Modified bootstrap, creative-start and custom styles to use the custom properties provided by the theme files
- Added theme picker to switch between themes in the Server Settings (thanks to @radWorx) and integrate the chosen theme file in the head part of the layout
- Restructured bootstrap and theme related asset files (CSS and JS) to draw a distinction about what has been customized and what is original vendor code.
- Added bootstrap kitchensink page to see all elements on one page and easily theme them. (accessible via URL `/_bootstrap_kitchensink.html?theme=casa` – defaults to "classic" theme)

In addition to that the PR also contains some bootstrap usage fixes (i.e. for the grid on some pages), but I tried to keep the markup changes as minimal as possible.

The branch is rebased against the current master. 

## Things that need to be discussed/decided/done still:

- [ ] Thoroughly test pages and scenarios as there might be constellations I have overlooked or not seen yet
- [ ] Test backwards compatibility with current themeing approach

Optional things that can be done based on this:

- Provide more BTCPay themes (i.e. the "Modern" theme, which is not done yet)
- Set Casa theme as default for BTCPayServer instances running on a Casa node
- Allow the user to upload and choose own themes


Kukks:
whooooa, that is big, lets try and split it up :)

pavlenex:
I've already began testing it so will provide my feedback here regardless of you wanting to split it at later or not.

First, I'd switch theme selector to top. I assume most people would use pre-made themes, then customize. It kinda makes more sense, right?

push to top


## Classic Theme

1. The front page has very weird design. I don't understand, grey color overlay makes no sense. I understand you wanted to emphasize login CTA but it really does not look good at least to me. Could we find a better way to emphasize CTA?


2. Font used in the boxes seems too small imo, too light and even on my desktop barely visible. (maybe I am blind?)
font
It just blends in way too much and it's too small to what I'm used to.
blend in
Increasing the font slightly makes things better imo (my example)
font increase
Even adding higher font weight looks better, I mean invoices in particular are important I want to scan amounts quickly and stuff and not have to use a telescope 👓 
(my example)
heavier font

Def use same font weight/sizes as on the Casa theme, it looks better on Casa.

## Casa Theme
Looks quite powerful.
casa

Def need someone from Casa to take a look at it regarding the coloring, but I personally like it.

If I'd really had to nitpick I'd say amount field in wallet could use some love to be a bit more visible
amount

That's it for now, great work so far Dennis.

dennisreimann:
@pavlenex Thanks for your feedback – I addressed your points with some added commits 👍

NicolasDorier:
coooool

@dennisreimann imho you should not try to do more with this PR, what is in `Things that need to be discussed/decided/done still:` can be done later.

NicolasDorier:
Try to make the default design as near as possible from the current one. Font, weight, colors.

britttttk:
tACK

dennisreimann:
Heads-up: I just rebased the branch again with the current master.

Also did some testing for the cases where an additional bootswatch theme is provided. It does not break anything as far as I could see, nevertheless this and the new themeing approach are orthogonal and mixing them might lead to inconsistencies. Here's what it looks like with the sketchy theme:

sketchy
 
Tomorrow @rockstardev and I will go through this and hopefullt get this ready to merge :)

NicolasDorier:
@dennisreimann I am not worry about breaking other people theme, more about the points that @pavlenex said above. The default theme should just look the same as it is today.

rockstardev:
@NicolasDorier I'm merging this to include in next version... after review with @dennisreimann I found it's great foundation for going forward

Issue #1138, opened 2019-11-08
Redirect when adding user manually


pavlenex:
**Describe the problem/bug**
With our new register/login page re-design, user gets the impression that the new login page is a home page. Ok, when user goes to Server Settings > Users > Add new user he's redirected to a newly designed home page which disrupts the UX way too much. 

I suggest that we handle addition of a new user similarly to how we do it in Store Settings, internally without re-directing people.

1. Go to Server Settings > Users
2. Click on `Add new user`
3. See the disruption.

This is up for discussion, not sure how you feel about this, don't want to hurt any feelings with my language, so let's kindly discuss it.

pavlenex:
Ok, this may not be a bug as @britttttk clarified, but really terrible UX.

britttttk:
@pavlenex can you describe why you think it is terrible UX, so we can improve it? 
It was redirected to that link before as well.

pavlenex:
As I said on Mattermost, it disrupts the UX way too much and looks weird, user gets a feeling that  he's logged out.

bolatovumar:
Agree with @pavlenex here. This probably won't be a big deal for experienced users but would probably be pretty jarring for a new user.

NicolasDorier:
I think it is quite easy to solve actually:

1. Set `ViewData["LayoutPage"]` to `_Layout.cshmtl`.
2. In Register, if  `ViewData["LayoutPage"]` is not empty, then use that instead of the `WelcomeLayout`.

bolatovumar:
@NicolasDorier wouldn't you have to check if there is a return url to make sure you don't accidentally use the regular layout when you mean to use the welcome layout (see my PR).

@pavlenex what do you think about this solution?

pavlenex:
Closing this as it's been fixed in #1139

PR #1139, opened 2019-11-09
Vary page layout for registering new accounts

bolatovumar:
address #1138

### DIFF
--- a/BTCPayServer/Controllers/AccountController.cs
+++ b/BTCPayServer/Controllers/AccountController.cs
@@ -364,7 +364,7 @@ namespace BTCPayServer.Controllers
 
         [HttpGet]
         [AllowAnonymous]
-        public async Task<IActionResult> Register(string returnUrl = null, bool logon = true)
+        public async Task<IActionResult> Register(string returnUrl = null, bool logon = true, bool useBasicLayout = false)
         {
             var policies = await _SettingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
             if (policies.LockSubscription && !User.IsInRole(Roles.ServerAdmin))
@@ -372,6 +372,7 @@ namespace BTCPayServer.Controllers
             ViewData["ReturnUrl"] = returnUrl;
             ViewData["Logon"] = logon.ToString(CultureInfo.InvariantCulture).ToLowerInvariant();
             ViewData["AllowIsAdmin"] = _Options.AllowAdminRegistration;
+            ViewData["UseBasicLayout"] = useBasicLayout;
             return View();
         }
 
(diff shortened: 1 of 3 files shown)


BTCPayServer/Controllers/AccountController.cs
```diff
@@ -369,6 +369,7 @@ public async Task<IActionResult> Register(string returnUrl = null, bool logon =
             var policies = await _SettingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
             if (policies.LockSubscription && !User.IsInRole(Roles.ServerAdmin))
                 return RedirectToAction(nameof(HomeController.Index), "Home");
+            ViewData["WithReturnUrl"] = returnUrl != null ? true : false;
```
NicolasDorier: I don't think you should use `returnUrl`. There can be a returnUrl if you try to go to a page that is forbidden, you get redirected to login page and then if you click on "Create an account" in the login page, the returnUrl get carried out.

BTCPayServer/Views/Account/Register.cshtml
```diff
@@ -1,10 +1,10 @@
 ﻿@model RegisterViewModel
 @{
     ViewData["Title"] = "Register";
-    Layout = "_WelcomeLayout.cshtml";
+    Layout = (bool)ViewData["WithReturnUrl"] ? "../Shared/_Layout.cshtml" : "_WelcomeLayout.cshtml";
```
NicolasDorier: You should use Another ViewData that is only set in `Add new user` for the reason I specified above.

BTCPayServer/Views/Account/Register.cshtml
```diff
@@ -1,10 +1,10 @@
...
 }
 
-<div class="modal-dialog modal-login">
+<div class="modal-dialog @(Layout != "_WelcomeLayout.cshtml" ? "modal-dialog-centered" : "")">
```
NicolasDorier: this is ugly to read. I advise you do to on the top block.

```
var isDisruptive = ViewData["DisruptiveLayout"] is bool;
Layout = isDisruptive ? "_WelcomeLayout.cshtml" : "../Shared/_Layout.cshtml";
```

then use` isDisruptive` in your condition here.

BTCPayServer/Views/Account/Register.cshtml
bolatovumar: Makes sense.

BTCPayServer/Views/Account/Register.cshtml
bolatovumar: Makes sense.

BTCPayServer/Views/Account/Register.cshtml
NicolasDorier: no, I am saying that `ViewData["BoringLayout"] = true;` should be set in the ListUser's add new user.

bolatovumar: Oh, ok.

NicolasDorier:
way simpler! @pavlenex you take  a look?

pavlenex:
tACK.

PR #1142, opened 2019-11-10
Fix U2F bug when using multiple devices

### LINKED ISSUE #1109: Registering multiple U2F devices breaks all but most recent
**Describe the problem/bug**
If you register multiple U2F devices for 2FA and then try to log in, only the last device that was registered will actually authenticate.

**Your environment**
* Version of BTCPay Server:  v1.0.3.135
* Deployment method: docker
* Other relevant environment details: Ubuntu 18.04.1 LTS

**To Reproduce**
1. Go to /Manage/U2FAuthentication
2. Click on 'Add New Device'
3. Register a U2F device
4. Click on 'Add New Device'
5. Register a second U2F device
6. Log Out
7. Log In
8. Try to authenticate with the first U2F device you registered

**Expected behavior**
Log in should succeed

**Actual behavior**
Login fails with a timeout error

Kukks:
closes #1109 

@pavlenex  Can you test( with a variety of setups: 1 configured device, 2 configured devices, trying to sign in with non registered ones when 1 or 2 devices registered, etc)

### DIFF
--- a/BTCPayServer/U2F/U2FService.cs
+++ b/BTCPayServer/U2F/U2FService.cs
@@ -157,14 +157,22 @@ namespace BTCPayServer.U2F
                 var authenticationRequest =
                     UserAuthenticationRequests[userId].First(f =>
                         f.KeyHandle.Equals(authenticateResponse.KeyHandle, StringComparison.InvariantCulture));
+                
                 var registration = new DeviceRegistration(device.KeyHandle, device.PublicKey,
                     device.AttestationCert, Convert.ToUInt32(device.Counter));
 
                 var authentication = new StartedAuthentication(authenticationRequest.Challenge,
                     authenticationRequest.AppId, authenticationRequest.KeyHandle);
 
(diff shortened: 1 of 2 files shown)


pavlenex:
So far seems to work, still testing.
Noticed two things though, that are unrelated  to this PR probably but would be worth addressing:
- When you try to log in with a U2F device that's not registered, a very weird message appears. `This device is already registered, please login` makes no sense, instead when user tries to register with an unregistered device, give error like this : `This U2F device is not registered on this website, please try a different one`.
Capture
- User can register same U2F device multiple times, does it make sense to allow that? `This U2F device is already registered`

pavlenex:
tACK, I can confirm that the bug with multiple devices is no longer present. I've tested on 4 different U2F devices in different combinations.

@Kukks not sure if you want to address the above two issues in this PR or should we open a new issue, but should be a quick fix hopefully.

Kukks:
fixed the text but the duplicate u2f devices is not easy to do so I would leave it for now

pavlenex:
Thanks for fixing that nit, ACK.

PR #1143, opened 2019-11-10
Display proper rate for ln sats feature

### LINKED ISSUE #1128: Wrong USD rate for sats on invoice
**Describe the problem/bug**
If you check the "Display lightning payment amounts in Satoshis" checkbox in "Checkout experience" tab and then switch to lightning payment on an invoice you see exchange rate which says something like "1 Sats = $9,307.54" which is obviously incorrect (at least for now).

**Your environment**
* Version of BTCPay Server: v1.0.3.136

**Expected behavior**
Either show the actual value of a sat (something like $ 0.0093 USD) or show "1 BTC = $9,307.54".

**Actual behavior**
Shows price of 1 BTC for 1 Sat

**Screenshots/Links**
Screen Shot 2019-11-06 at 7 10 06 PM

Kukks:
fixes #1128

### DIFF
--- a/BTCPayServer/Models/InvoicingModels/PaymentModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/PaymentModel.cs
@@ -78,5 +78,6 @@ namespace BTCPayServer.Models.InvoicingModels
         public string RootPath { get; set; }
         public decimal CoinSwitchAmountMarkupPercentage { get; set; }
         public bool RedirectAutomatically { get; set; }
+        public string RateBaseAmount { get; set; } = "1";
     }
 }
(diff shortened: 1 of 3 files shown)


pavlenex:
@bolatovumar Can you test this one when you grab a spare moment please?

bolatovumar:
@pavlenex just looking at the code it looks like this will now show something like `100,000,000 Sats = $9000` or something like that if a user decides to show lightning amount in sats.

pavlenex:
@bolatovumar  Is that what you had in mind when opening the issue?
I'll review this one tomorrow.

bolatovumar:
@pavlenex I didn't really have an exact solution in mind but this solution shows correct data at least. I will test it and see how well it works.

bolatovumar:
This is what it looks like with the fix:

Screen Shot 2019-11-10 at 2 42 01 PM

Looks correct but I'm wondering if the number should be formatted first before rendering it? Reading `100000000 Sats` is hard. Something like `100,000,000 Sats` would be easier to read. This is not critical to me, however.

NicolasDorier:
I think `1 sat = xxx USD` would be better though.

Kukks:
For sure, but we dont have access to the rate in the payment handler at
that point so it would need more work and params passed around

pavlenex:
> 100000000 Sats is hard. Something like 100,000,000 Sats 

Agreed. Please use spacing. It's the best `100 000 000` as it eliminates confusion of `,` and `.` in certain countries. Spacing works everywhere and it's much cleaner to the eye.

Kukks:
Let's do this in a separate PR as this is mainly to address that rate bug.

bolatovumar:
@Kukks sounds good to me.

@pavlenex do you want to create a separate issue for this?

pavlenex:
For the spacing I opened #1146 , but should the [rate display that Nicolas mentioned ](https://github.com/btcpayserver/btcpayserver/pull/1143#issuecomment-552414223)addressed in this PR?

Kukks:
I would really like to get this merged just to get the fix out of the way and then deal with the aesthetic stuff with discussion later

NicolasDorier:
@Kukks we have the rates at the view level I don't understand why it should be passed in the handler.

PR #1150, opened 2019-11-13
Add SyncModal partial to welcome layout

### LINKED ISSUE #1148: Sync pop up window not shown on the login page
**Describe the problem/bug**
In the past, users had a pop-up window showing the progress of the blockchain sync displayed in the bottom right corner.

With new login page as a front page, it seems we lost that critical information for people. Right now user has to be logged in to see that information.

While most people will create an account right away, it may be confusing for newbies since plenty of tutorials and docs tell them the blockchain has to sync and that indicator is in the corner.

Maybe it's good to also address the design of that pop-up to fit more the new login page design.

**Your environment**
* Version of BTCPay Server: 1.0.3.137
* Deployment method: Docker

**To Reproduce**
1. Deploy btcpay server
2. Go to the domain
3. No sync progress is shown until you log in

**Expected behavior**
See the sync progress on the login page.

**Actual behavior**
No sync progress unless you log in.

bolatovumar:
fix #1148

### DIFF
--- a/BTCPayServer/Views/Account/_WelcomeLayout.cshtml
+++ b/BTCPayServer/Views/Account/_WelcomeLayout.cshtml
@@ -2,6 +2,8 @@
     Layout = null;
 }
 @inject BTCPayServer.Services.BTCPayServerEnvironment env
+@inject BTCPayServer.HostedServices.NBXplorerDashboard dashboard
+
 <!DOCTYPE html>
 <html lang="en">
 <head>
@@ -106,5 +108,10 @@
         </div>
     </section>
     @await Html.PartialAsync("_ValidationScriptsPartial")
+
+    @if (!dashboard.IsFullySynched())
+    {
+        <partial name="SyncModal" />
+    }
 </body>
 </html>


bolatovumar:
@pavlenex this should fix it

NicolasDorier:
I am not against it, but should we? In a sense, this protect privacy.

bolatovumar:
@NicolasDorier hard to tell without user feedback. It may not matter at all or it may be confusing to new users. Not sure.

NicolasDorier:
@pavlenex strong opinion on that?

pavlenex:
Can you elaborate a bit more on the privacy @NicolasDorier what's the attack vector here that invades privacy?

Honestly, when user sees this, because of how our page is designed I assume that they would register instantly, right? In that case it's not of a big deal.
However some of our tutorials may say once you launch you will see a pop-up. 

Could be unnecessary and if it protects privacy, would make sense to leave this one opened until we get more feedback from people who deployed for the first time, what do you guys think?

NicolasDorier:
Well because you see the height of the sync, you can now know which fullnode this guy run. (Full nodes advertize the height)

Ok let's keep this open.

Eskyee:
yes excellent topic, I thought too like Pav to have it the same as before, loading!! 
but I prefer protect privacy!! as a first choice, that's my idea here!! guys

Well because you see the height of the sync, you can now know which fullnode this guy run!!
lol yes its true, I too can see who runs what other shit coins lol and stuck in mid flight lol

bolatovumar:
@NicolasDorier your point here is that any anonymous user could technically go to the welcome page and see the height and this way figure out what node that server is running? If that's the case then it's probably best not to show the sync state until you authenticate as admin.

pavlenex:
Yes, fair point then. I guess I'll modify a few nits in docs to tell users to register right away. Theoretically if they don't register, someone can and see their block height anyway.

britttttk:
Concept Ack - sync popup is very good in my opinion, but on my computer I need to close the sync pop up to see the register area. 

Would a user know/think to do that? 
image

pavlenex:
@britttttk for privacy reasons and the fact that other people don't care about the sync process, it makes more sense to display the sync pop up after you register. From the screenshoot it seems you're running an outdated version, the pop up should not be displayed unless you're logged in.

britttttk:
>outdated version

@pavlenex It is the version of this pull request. :) I agree with you, sync after register is best.

pavlenex:
Yeah, sorry, anyway we agreed to give some more time to this, if people don't complain, we'll close this PR as agreed by @bolatovumar  earlier.

rockstardev:
I'm also for not doing this on register page... @bolatovumar you ok with closing this PR without merging?

We can revisit this later by maybe showing different indicator (blockchain syncing, please login to see progress).

bolatovumar:
@rockstardev closed

PR #1152, opened 2019-11-13
Add hardware support via BTCPayServer Vault

NicolasDorier:


NicolasDorier:
Ping @rockstardev @Kukks @bolatovumar I was wondering if I should not make use of async/await feature in the javascript.

I don't want to introduce a dependency like nodejs for developers in this project, but the waterfall of callback is really not super cool to read.

async/await seems supported by modern browsers.

NicolasDorier:
I am referring to the cascade https://github.com/btcpayserver/btcpayserver/pull/1152/files#diff-057004722b99872931b30b06d3f6a969R138

Kukks:
I would introduce async await only if you have a transpiler hooked up (Babel, typescript).  Truth is even if modern browsers support it, many users take forever to update( or in some scenarios arent able to)

Kukks:
We' ve had issues for example with small usages of `const` and `let` in the js breaking a few users

NicolasDorier:
`transpiler hooked up (Babel, typescript).` I don't want to add any new dev time dependency, so this is not an option.

bolatovumar:
@NicolasDorier I guess it's pretty safe to use async/await if you don't care about IE 11 (you probably don't). See support stats here: https://caniuse.com/#search=async

Overall it has 91.5% support globally.

NicolasDorier:
Well if `We' ve had issues for example with small usages of const and let in the js breaking a few users` I guess I should not even try with async/await :(

bolatovumar:
@NicolasDorier I'm wondering if in this case lack of support in IE 11 and some very old browsers is less of an issue because it's an internal page that's only available after you login. IE 11 is the only browser without async/await support that has any significant usage.

NicolasDorier:
Well @Kukks said some users complained because of const, which is itself more compatible than async/await.

NicolasDorier:
@bolatovumar @Kukks so with the new themeing feature coming, we are using named variable which is supported at 93% versus 91% for async await.

I think to show a popup warning page in case the user have outdated browser and just use async.

Kukks:
Yeah code the latest spec and then we can add a warning or a polyfill

bolatovumar:
@Kukks you can't really polyfill async/await easily without introducing a transpilation step.

NicolasDorier:
Still lot's of things to improve, but that will be for another day.

PR #1158, opened 2019-11-14
Do not allow login or register over an insecure connection

### LINKED ISSUE #1149: New Login page lets me log in via port 80
I just noticed that the new login page lets me login via an unsecured connection. 

I can access and login via port 80 and only after the login, it displays the "Your access to BTCPay Server is over an unsecured network..." message. Even though the message displays, I'm logged in and have access to all functions.

Kukks:
fixes #1149

### DIFF
--- a/BTCPayServer/Controllers/AccountController.cs
+++ b/BTCPayServer/Controllers/AccountController.cs
@@ -74,14 +74,18 @@ namespace BTCPayServer.Controllers
         [AllowAnonymous]
         public async Task<IActionResult> Login(string returnUrl = null)
         {
+            
             if (User.Identity.IsAuthenticated && string.IsNullOrEmpty(returnUrl))
                 return RedirectToLocal();
             // Clear the existing external cookie to ensure a clean login process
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Can you rebase and run the tests again?

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -580,6 +619,21 @@ private IActionResult RedirectToLocal(string returnUrl = null)
                 return RedirectToAction(nameof(HomeController.Index), "Home");
             }
         }
+        
+        
+        private bool CanLoginOrRegister()
```
NicolasDorier: Can you refactor this? This is not obvious that this method has side effects. I was wondering why you were calling it without checking the result.

Kukks: What do you want me to do? Rename it or split into two

NicolasDorier: Split in two

Issue #1141, opened 2019-11-10
Errors not working in Apps 


britttttk:
**Describe the problem/bug**
Some Apps errors are not working properly

**Your environment**
* Version of BTCPay Server: BTCPayServer v1.0.3.137
* Deployment method: LunaNode

**To Reproduce**

**Crowdfund** (Update App without required fields)
1. Click View App without required fields provided.
2. Page always shows "A Target Currency must be set for this app in order to be loadable."
3. Even if missing required field is something like not providing Description or Title

**POS** (Pay invoice without wallet connected to store)
1. Create POS without wallet 
2. Try to pay POS invoice
3. 500 error page

**Expected behavior**
Crowdfund - Missing required fields should not allow them to continue, right?
POS - Normally there is a warning, you need to connect wallet (like crowdfund does).

**Additional context**
I think this was not happening before.

bolatovumar:
> POS - Normally there is a warning, you need to connect wallet (like crowdfund does).
Shouldn't there be a check when you assign an app to a store to make sure that store actually has a wallet connected to it? If there isn't a wallet, it shouldn't let you connect and give you a message as to why. This way you will never get to a stage where you created an app, tried to pay for something and got a 500 (or a warning in the UI). Same goes for the crowdfund app.

britttttk:
> > POS - Normally there is a warning, you need to connect wallet (like crowdfund does).
> 
> Shouldn't there be a check when you assign an app to a store to make sure that store actually has a wallet connected to it? If there isn't a wallet, it shouldn't let you connect and give you a message as to why. This way you will never get to a stage where you created an app, tried to pay for something and got a 500 (or a warning in the UI). Same goes for the crowdfund app.

Normally the user can build the app before connecting a wallet. For why, I think an example is like if you are collaborating with a company, you can design the app and show it to them, before obtaining their xpub. This is the error for CF without wallet. 

I thought before, it was a text warning displayed on the page that opened, but I don't remember. This is how it is for CF. 

![image](https://user-images.githubusercontent.com/39231115/68538247-92883380-032e-11ea-8c23-c83596a77960.png)

bolatovumar:
>Normally the user can build the app before connecting a wallet. For why, I think an example is like if you are collaborating with a company, you can design the app and show it to them, before obtaining their xpub.
[...]

PR #1161, opened 2019-11-15
Hide View button in crowdfund

britttttk:
#1141 Hide View button if user does not provide required crowdfund info in settings.

### DIFF
--- a/BTCPayServer/Views/Apps/UpdateCrowdfund.cshtml
+++ b/BTCPayServer/Views/Apps/UpdateCrowdfund.cshtml
@@ -217,7 +217,10 @@
                     <div class="form-group">
                         <button name="command" type="submit" class="btn btn-primary" value="save" id="SaveSettings">Save settings</button>
                         <a class="btn btn-secondary" target="_blank" asp-action="ListInvoices" asp-controller="Invoice" asp-route-searchterm="@Model.SearchTerm">Invoices</a>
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/Apps/UpdateCrowdfund.cshtml
```diff
@@ -217,7 +217,10 @@
                     <div class="form-group">
                         <button name="command" type="submit" class="btn btn-primary" value="save" id="SaveSettings">Save settings</button>
                         <a class="btn btn-secondary" target="_blank" asp-action="ListInvoices" asp-controller="Invoice" asp-route-searchterm="@Model.SearchTerm">Invoices</a>
+                        @if ((Model.Description != null) && (Model.Title != null) && (Model.TargetCurrency != null))
```
Kukks: You should check if the ID is setinstead of these fields as that indicates the app has been created

britttttk: Sorry I don't know what ID you mean, the AppId is created before they update the app.

Kukks: ah crap you're right

Kukks: I think if you just check one of the required fields, it would be enough. `TargetCurrency ` is probably a good one.

PR #1162, opened 2019-11-15
Theme fixes

dennisreimann:
Some more things I found when going throught the pages.

### DIFF
--- a/BTCPayServer/Views/Manage/Disable2fa.cshtml
+++ b/BTCPayServer/Views/Manage/Disable2fa.cshtml
@@ -11,7 +11,7 @@
     </p>
     <p>
         Disabling 2FA does not change the keys used in authenticator apps. If you wish to change the key
-        used in an authenticator app you should <a asp-action="ResetAuthenticatorWarning">
+        used in an authenticator app you should <a asp-action="ResetAuthenticatorWarning" class="alert-link">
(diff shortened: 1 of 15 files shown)


BTCPayServer/Views/Manage/Disable2fa.cshtml
```diff
@@ -11,7 +11,7 @@
     </p>
     <p>
         Disabling 2FA does not change the keys used in authenticator apps. If you wish to change the key
-        used in an authenticator app you should <a asp-action="ResetAuthenticatorWarning">
+        used in an authenticator app you should <a asp-action="ResetAuthenticatorWarning" class="alert-link">
```
Kukks: do we need a class for this? Wouldn't `.alert  a` work?

dennisreimann: I wondered about this too until I saw that they have [this special class](https://getbootstrap.com/docs/4.3/components/alerts/#link-color) for this case. You don't notice the difference in the classic theme (because font colors fortunately fit), but we need to apply it so that it also works with the Casa theme.

pavlenex:
UX wise, there are logical improvements. I've gone step by step verifying what each commit did and looks good to me. I'll let you review the code and classes changed.

PR #1175, opened 2019-11-19
Add CSS variable for preformatted text color

### LINKED ISSUE #1169: Can't See UI Logs in Casa Theme (font color)
**Describe the problem/bug**
Log text color in UI Server Settings logs is too dark in Casa Theme and can't read them easily.

**Your environment**
* Version of BTCPay Server: BTCPayServer v1.0.3.138
* Deployment method: LunaNode

**Logs (if applicable)**
image

**To Reproduce**
1. Go to Server Settings
2. Click on Maintenance
3. Click on Logs
4. Logs shown are hard to read.

**Additional context**
Actual logs are fine.

bolatovumar:
fix #1169

### DIFF
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -122,3 +122,7 @@ a.nav-link:hover {
     background: white;
     display: inline-block;
 }
+
+pre {
+    color: var(--btcpay-preformatted-text-color); 
+}
--- a/BTCPayServer/wwwroot/main/themes/casa.css
+++ b/BTCPayServer/wwwroot/main/themes/casa.css
@@ -62,6 +62,8 @@
   --btcpay-header-bg: var(--btcpay-brand-darker);
   --btcpay-footer-bg: var(--btcpay-brand-darkest);
 
+  --btcpay-preformatted-text-color: var(--btcpay-color-white);
+
   --btcpay-font-size-base: 16px;
   --btcpay-font-family-head: 'Open Sans', 'Helvetica Neue', Arial, sans-serif;
   --btcpay-font-family-base: -apple-system, 'Open Sans', BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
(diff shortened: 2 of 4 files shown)


bolatovumar:
@britttttk should look like this now:

Screen Shot 2019-11-18 at 8 22 54 PM

NicolasDorier:
ping @dennisreimann is it normal to change directly boostrap css ?

dennisreimann:
I think it'd be best to have modifications like this in the [site.css](/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/main/site.css) file to separate custom stuff, like css custom properties we define after the `/* Color definitions for specific sections - try to reuse colors defined above */` comment.

The general approach should be to not modify the bootstrap.css, so that we can keep it updatable. 

-----
Further info/Caveat: The bootstrap.css file itself is generated based on what the vendor bootstrap.css provides. Right now bootstrap does not use custom properties, but I think in the future it is likely that they might switch to this approach as well. Until then I created a build script in [this repo](https://github.com/dennisreimann/btcpayserver-ui-prototype) which generates the bootstrap.css file we are using here.

dennisreimann:
One more addition: In general I think it is a good idea to introduce specific variables for cases like this. It allows us to address individual portions of the styles without affecting other parts which might be tight to a general variable – so thumbs up for that!

At best these specific variables should reuse other predefined variables, so that we keep the appearance consistent. (like @bolatovumar already did here)

I think I should write this up somewhere to provide some general guidance on how to use and grow this themeing approach :)

bolatovumar:
@NicolasDorier bootstrap.css has already been modified with a bunch of btcpay-specific styling variables so I figured it's OK to do.

bolatovumar:
@dennisreimann @NicolasDorier so are we thinking this new variable should be moved to `site.css` or kept in our modified `bootstrap.css` alongside the other variables?

dennisreimann:
It is supposed to be in `site.css`, which contains our modifications of `bootstrap.css`.

Yes, `bootstrap.css` contains variables too, but these are the general modifications we had to make in order to allow for this themeing approach (as bootstrap does not offer this itself). Because bootstrap has colors spread all over the place we'd have to override mostly everything, that's why these modifications are in the main bootstrap.css file.

Customizations like the one you did are useful and necessary, but in order to keep the `bootstrap.css` file updatable (and at best automatically generatable like I mentioned above) we should have them in `site.css`.

As I said I'll document this on friday.

bolatovumar:
Alright, moved the modification from `bootstrap.css` into `site.css` as @dennisreimann suggested.

PR #1184, opened 2019-11-22
[Wallet] Specify validation range for wallet send amount

### LINKED ISSUE #1182: Blank screen when sending 0 from the wallet
**Describe the problem/bug**
When user tries to send 0 from the wallet, he's prompted to `subtract the fees` and then the wallet proceeds to the next page, which is a blank screen. Instead user should get an error in wallet that he can't send and must enter the amount. ` The Amount field is required. `.

**Your environment**
* Version of BTCPay Server: v1.0.3.143
* Deployment method: Docker
* Other relevant environment details: Replicated on mainnet demo server

**To Reproduce**
1. Go to Wallet > Send
2. Enter the address and balance `0`
3. Scroll down, and `subtract the fees`
4. Click `Sign with` Ledger, Vault or anything else
5. See error the blank screen

**Expected behavior**
Sometimes in a hurry people can forget to enter the amount, and we should `provide a warring in a wallet itself` and not allow them to see a blank screen which is confusing.

Possible warnings:
- `The amount field is required`
- `Enter the amount`

**Actual behavior**
User sees blank screen when sending 0 balance from the wallet.

**Screenshots/Links**

bolatovumar:
fix #1182

### DIFF
--- a/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
+++ b/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
@@ -17,8 +17,10 @@ namespace BTCPayServer.Models.WalletViewModels
             [Required]
             public string DestinationAddress { get; set; }
 
-            [Display(Name = "Amount")] [Required]  [Range(0.0, double.MaxValue)]public decimal? Amount { get; set; }
-
+            [Display(Name = "Amount")]
+            [Required]
+            [Range(1E-08, 21E6)]
+            public decimal? Amount { get; set; }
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
actually it does not fix it. This can happen if there is one UTXO whose value can't even pay for fee.

NicolasDorier:
There is another bug in the form though: If you put wrong input (invalid address) no error messgae

bolatovumar:
> There is another bug in the form though: If you put wrong input (invalid address) no error messgae

@NicolasDorier isn't this a separate issue which should be addressed in a separate PR?

PR #1190, opened 2019-11-23
Integrate Configurator External Service

Kukks:


NicolasDorier:
Can you run test again? it seems to fail. Do you want I merge this for this version?

Kukks:
tests pass

BTCPayServer/Configuration/ExternalConnectionString.cs
```diff
@@ -77,6 +79,20 @@ public async Task<ExternalConnectionString> Expand(Uri absoluteUrlBase, External
...
+                {
+                    if (string.IsNullOrEmpty(connectionString.Password))
+                    {
+                        connectionString.Password = Guid.NewGuid().ToString();
+                    }
+
+                    await File.WriteAllTextAsync(connectionString.PasswordFile, connectionString.Password);
+                }
+            }
+
```
NicolasDorier: This should be called "CookieFile" and if "CookieFile", the value inside should not be cached, the file should be generated by the configurator, not by btcpayserver,

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -665,6 +668,15 @@ public async Task<IActionResult> Service(string serviceName, string cryptoCode,
                     case ExternalServiceTypes.LNDGRPC:
                     case ExternalServiceTypes.LNDRest:
                         return LndServices(service, connectionString, nonce);
+                    case ExternalServiceTypes.Configurator:
+                        
+                            return View("ConfiguratorService", new LightningWalletServices()
+                            {
+                                ShowQR = showQR,
+                                WalletName = service.ServiceName,
+                                ServiceLink = $"{connectionString.Server}?password={connectionString.Password}"
```
NicolasDorier: The password will end up in the browser history.
The only way is to read the cookiefile, and the configurator to regenerate a new cookiefile once the value has been used. (RTL does that)

Kukks:
updated @NicolasDorier

PR #1193, opened 2019-11-24
UI: Left align admin section headings

dennisreimann:
Removes the custom property and consistently align the headings left, which I agree is the better choice.

Context: https://chat.btcpayserver.org/btcpayserver/pl/upb3nzch7j8nmpbt7kfmoe6beo

### DIFF
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -42,11 +42,6 @@ a {
     transition-property: background, color;
 }
 
-/* Admin Sections */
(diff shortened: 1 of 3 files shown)


NicolasDorier:
I will revert this, it does not looks good everywhere:

image

dennisreimann:
We coulso try to fix this on these specific page(s), becaus it‘d be the same for the Casa theme. I think that would work better.

NicolasDorier:
make sense.

NicolasDorier:
doing it.

PR #1202, opened 2019-11-30
Change create with connect in wallet page

### LINKED ISSUE #1201: Remove create a wallet
People are confused by wording `Create and manage wallet` which leads them into thinking they can actually create a wallet in btcpay.
It would be better to use a word `Connect`.

pavlenex:
Closes #1201


britttttk:
Ack

PR #1212, opened 2019-12-04
Generate wallet 

Kukks:
(typos on images have been fixed)

image

image

### DIFF
--- a/BTCPayServer/Controllers/StoresController.BTCLike.cs
+++ b/BTCPayServer/Controllers/StoresController.BTCLike.cs
@@ -389,7 +389,7 @@ namespace BTCPayServer.Controllers
             }
             vm.Confirmation = true;
             ModelState.Remove(nameof(vm.Config)); // Remove the cached value
-            return View(vm);
+            return View(nameof(AddDerivationScheme),vm);
         }
         
         private bool CanUseGenerateWallet()
--- a/BTCPayServer/Controllers/StoresController.BTCLike.cs
+++ b/BTCPayServer/Controllers/StoresController.BTCLike.cs
@@ -132,6 +132,7 @@ namespace BTCPayServer.Controllers
                 vm.Config = derivation.ToJson();
             }
             vm.Enabled = !store.GetStoreBlob().IsExcluded(new PaymentMethodId(vm.CryptoCode, PaymentTypes.BTCLike));
+            vm.CanUseGenerateWallet = CanUseGenerateWallet();
         }
 
         private DerivationSchemeSettings GetExistingDerivationStrategy(string cryptoCode, StoreData store)
@@ -179,7 +180,7 @@ namespace BTCPayServer.Controllers
                         Message = "Config file was not in the correct format"
                     });
                     vm.Confirmation = false;
-                    return View(vm);
+                    return View("AddDerivationScheme",vm);
                 }
             }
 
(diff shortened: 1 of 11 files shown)


BTCPayServer/Controllers/StoresController.BTCLike.cs
```diff
@@ -319,6 +320,38 @@ private DerivationSchemeSettings GetExistingDerivationStrategy(string cryptoCode
...
+          var client =   _ExplorerProvider.GetExplorerClient(cryptoCode);
+          var response = await client.GenerateWalletAsync(request);
+          
+          var store = HttpContext.GetStoreData();
+          var result = await AddDerivationScheme(storeId, new DerivationSchemeViewModel()
+          {
+              Confirmation = false,
+              Network = network,
+              RootFingerprint = response.AccountKeyPath.MasterFingerprint.ToString(),
+              RootKeyPath = response.AccountKeyPath.KeyPath,
```
NicolasDorier: not right, should be `network.GetRootKeyPath()`

BTCPayServer/Controllers/StoresController.BTCLike.cs
```diff
@@ -319,6 +320,38 @@ private DerivationSchemeSettings GetExistingDerivationStrategy(string cryptoCode
             return ShowAddresses(vm, strategy);
         }
 
+        [HttpPost]
+        [Route("{storeId}/derivations/{cryptoCode}/generatenbxwallet")]
+        public async Task<IActionResult> GenerateNBXWallet(string storeId, string cryptoCode, GenerateWalletRequest request)
+        {
+            var network = _NetworkProvider.GetNetwork<BTCPayNetwork>(cryptoCode);
```
NicolasDorier: too much space

britttttk:
Creating a wallet this way.. How can a user check addresses as instructed in the UI? `Please check that your BTC wallet is generating the same addresses as below.`

tACK - I'm able to send/receive fine with mnemonic seed for now, I think that more functionality on sign comes when #1218 is merged. However I don't quite understand the RPC checkbox, is that something that is to be added in another PR as well?

Kukks:
RPC checkbox works already:
it imports any address generated to the Bitcoin Core wallet node so that
you can view your balance through the bitcoin-cli or rpc.
If you enable save private keys, you'll also have private keys imported to
the wallet and be able to spend from it.

NicolasDorier:
`Save private keys` should be named `Is hotwallet`.  With proper warning about what it means (danger about funds being stolen, but more convenient)

Eskyee:
Welcome to BTCPay Server ... Secure. The payment server does not need to know your private keys, so your money can't be stolen.  
that was the good old days ... lol

sorry but  have I missed something here ??? 
we are going save private keys in btcpay or NBX ??  I'm very dumbstruck here !!  
this sounds like a very very bad idea,

Kukks:
> 
> 
> Welcome to BTCPay Server ... Secure. The payment server does not need to know your private keys, so your money can't be stolen.
> that was the good old days ... lol
> 
> sorry but have I missed something here ???
> we are going save private keys in btcpay or NBX ?? I'm very dumbstruck here !!
> this sounds like a very very bad idea,

Opt-in feature for specific use-cases. Just ignore it if you don't need it.

Eskyee:
Opt-in feature for specific use-cases !! yes ok, that's much better for use-cases, I understand a lot more now,  Is a hotwallet, and not the store main xpub wallet, is that correct ??  could you also explain how secure this new method would be, as  a hot wallet. thanks guys
sorry for my rants!! ;)

pavlenex:
A hot wallet is needed mostly for developers. Nothing will change and the method of adding xpub key remains the same, as Kukks said, this just extend the use-cases and is completely optional and advanced feature. Lightning is a hot wallet as well by the way. It won't influence you if you're not using it if that's your question.

Eskyee:
yes now I`m very happy, just got myself confused.. as a extra opt-in feature it's a kool idea, @pavlenex

Kukks:
Updated to this
image

Kukks:
can we merge now @NicolasDorier

PR #1214, opened 2019-12-04
Fixing sats exchange rate display

### LINKED ISSUE #1147: 1 Sat = USD in invoice rates when using satoshi
[As described here by @NicolasDorier ](https://github.com/btcpayserver/btcpayserver/pull/1143#issuecomment-552414223)using 1 sat = USD would be better looking and future-proof.
sats

Display the rate for 1 satoshi if people are displaying sats on the invoice.

MrPaz:
Fixed #1147

image

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -294,6 +294,10 @@ namespace BTCPayServer.Controllers
             };
 
             paymentMethodHandler.PreparePaymentModel(model, dto, storeBlob);
+            if (model.IsLightning && storeBlob.LightningAmountInSatoshi && model.CryptoCode == "Sats")
+            {
+                model.Rate = _CurrencyNameTable.DisplayFormatCurrency(paymentMethod.Rate / 100_000_000, paymentMethod.ParentEntity.ProductInformation.Currency);
+            }
             model.UISettings = paymentMethodHandler.GetCheckoutUISettings();
             model.PaymentMethodId = paymentMethodId.ToString();
             var expiration = TimeSpan.FromSeconds(model.ExpirationSeconds);
--- a/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
@@ -100,7 +100,8 @@
                 <span>{{ srvModel.btcDue }} {{ srvModel.cryptoCode }}</span>
             </div>
             <div class="single-item-order__right__ex-rate" v-if="srvModel.orderAmountFiat && srvModel.cryptoCode">
-                {{srvModel.rateBaseAmount}} {{ srvModel.cryptoCodeSrv }} = {{ srvModel.rate }}
+                <span v-if="srvModel.cryptoCodeSrv === 'Sats'">1 Sat = {{ srvModel.rate }}</span>
+                <span v-else>1 {{ srvModel.cryptoCodeSrv }} = {{ srvModel.rate }}</span>
             </div>
         </div>
         <span class="fa fa-angle-double-down"></span>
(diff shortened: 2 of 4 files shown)


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -294,6 +294,10 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
             };
 
             paymentMethodHandler.PreparePaymentModel(model, dto, storeBlob);
+            if (model.IsLightning && storeBlob.LightningAmountInSatoshi && model.CryptoCode == "Sats")
```
Kukks: Please try and do this in `LightningLikePaymentHandler.PreparePaymentModel`. I tried hard to move payment specific logic to their own handlers.

BTCPayServer/Views/Invoice/Checkout-Body.cshtml
```diff
@@ -100,7 +100,8 @@
                 <span>{{ srvModel.btcDue }} {{ srvModel.cryptoCode }}</span>
             </div>
             <div class="single-item-order__right__ex-rate" v-if="srvModel.orderAmountFiat && srvModel.cryptoCode">
-                {{srvModel.rateBaseAmount}} {{ srvModel.cryptoCodeSrv }} = {{ srvModel.rate }}
+                <span v-if="srvModel.cryptoCodeSrv === 'Sats'">1 Sat = {{ srvModel.rate }}</span>
```
Kukks: I don't think the changes are needed at all

BTCPayServer/Controllers/InvoiceController.UI.cs
rockstardev: He can't do it because currency exchange rate formatting is tied to controller.

I would merge this as quickfix for display and then if we want to do bigger refactoring of code - we revisit assumptions made... someone from core team should really dig deep into PreparePaymentModel.

BTCPayServer/Views/Invoice/Checkout-Body.cshtml
rockstardev: There needs to be change here because `rateBaseAmount` is not needed property... you always have 1 SOMETHING = ...

Extra line of code here is because he didn't want `1 Sats =`... but `1 Sat =`... ocd plural (which, as fellow ocd, I approve of)

bolatovumar: I approve of the OCD change as well even though now the code is less simple/elegant.

knocte:
While this is better than before, why not show the USD amount of the purchase too? Otherwise you force the user to make calculations in his mind...

Kukks:
> 
> 
> While this is better than before, why not show the USD amount of the purchase too? Otherwise you force the user to make calculations in his mind...
It will show the currency the invoice was created in + the current payment method's currency. Not everyone cares about USD( I certainly don't, I would argue for EUR since Im in EU or even DKK  since im in DK)

knocte:
> Not everyone cares about USD(

Sure, when I said USD I meant **user's preferred fiat currency**, still my point applies.

Kukks:
I still think the best option is there already: the unit of account of the invoice, the payment method's currency and the exchange rate between those two.

knocte:
> and the exchange rate between those two.

If you're not a hardcore bitcoin maximalist you want to know the fiat price, to check that it's the same as the one you found on the merchant website before it redirected you to BtcPayServer.

> While this is better than before

Actually, I take that back. This is worse than before, because what we had was the price of 1 BTC in fiat money, which can be compared very easily to any price-tracking website (e.g. coinmarketcap). Any sane person would compare that, to double check if the merchant is screwing you with the exchange rate or not. Now one has to open the calculator to get the BTCUSD (or BTCEUR, ...) price from the BtcSat price.

Kukks:
Yeah, personally I wouldn't use this feature but I can see it useful for
fun ln mini projects

Let's create a new issue to discuss this there to not hijack this one. I do
have some ideas

knocte:
> Let's create a new issue

I opened https://github.com/btcpayserver/btcpayserver/issues/1220

PR #1218, opened 2019-12-06
Sign with NBX Seed

Kukks:
Builds on #1212 

This only shows if the metadata is available.
image

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -550,9 +550,6 @@ namespace BTCPayServer.Tests
                 }
                 
                 SignWith(mnemonic);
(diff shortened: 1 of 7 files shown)


pavlenex:
I approve the use-case, but I'm a bit scared with what we have in the UI for the `Sign with...` will get out of control at some point and is a bit too much for normies to handle. It's somehow getting unorganized and confusing.

This is not a comment on this PR in particular, but does anyone have an idea how we can better present this in the UI? Normie goes to wallet, wants to send and then sees bunch of options they don't understand and gets overwhelmed.

Kukks:
@pavlenex  it does not show if the seed isn't available

pavlenex:
Ah, I missed that one, though my comment is not directly related to the PR, but to how we handle the UX here in general, I'll open an issue if I have a better idea of how to display these, thanks for clarifying.

NicolasDorier:
can you fix the conflicts?

PR #1228, opened 2019-12-18
Update HomeController.cs

Eskyee:
live share collaboration rockstardev channel fix

### DIFF
--- a/BTCPayServer/Controllers/HomeController.cs
+++ b/BTCPayServer/Controllers/HomeController.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


NicolasDorier:
what does it fix

Eskyee:
shows the user invoice entries in their local timezone.

bolatovumar:
@Eskyee this patch doesn't seem to be actually doing that though

Eskyee:
@rockstardev if u can remind me what this change  was On our live share session!! Since we made a few fixes !! On that session!! Cheers

rockstardev:
@Eskyee timezone was in different file... when I have time I'll revisit and co-author you on commit if I fix.

PR #1244, opened 2019-12-29
Remove OpenIddict

Kukks:


NicolasDorier:
Closing that for now, does not require that much work.
It has been painful to switch to 3.0 because of it, but the work has now been done anyway.

Kukks:
Rebased this PR and updated the migrations

NicolasDorier:
Tests do not pass?

NicolasDorier:
We remove openiddict for now.

The main issue concern UX. OpenIddict is meant for centralized services where clients connect to it.(like facebook, google, some intranet)

In such settings, the UX is the following:

* The app developer register to the centralized server and get his credentials, hard code/config them on the config of their app's server
* The user try the app asks for permissions
* The user get redirected to centralized server, and accept permissions

However, in btcpayserver, the user is also "the centralized server", which mean that the process is the following:

* The app developer asks the user to register the app to his server
* The user then, after filling out a very technical form for the app registration, need to give the credential back to the app.
* Then the app need to save specific clientid/secret/server for each user.
* The user try the app asks for permissions
* The user get redirected to centralized server, and accept permissions

This is a nightmare from UX perspective and will require lot's of support.
On top of it, the breaking changes of OpenIddict and the breaking changes of Entity Framework consumed lot's of maintainance resources for something we were not even using yet.

It has been decided to remove it from the time being. I hope we put it back when we have a solution which does not require the user to register the app and pass clientid/secret back and forth.

PR #1258, opened 2020-01-09
[WIP] Switch to .netcoreapp3.1

NicolasDorier:
Supersede https://github.com/btcpayserver/btcpayserver/pull/1070

### DIFF
--- a/BTCPayServer.Data/BTCPayServer.Data.csproj
+++ b/BTCPayServer.Data/BTCPayServer.Data.csproj
@@ -6,14 +6,14 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.1.14" />
     <PackageReference Include="Pomelo.EntityFrameworkCore.MySql" Version="2.1.4" />
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="2.1.2" />
-    <PackageReference Include="OpenIddict.EntityFrameworkCore" Version="3.0.0-alpha1.19515.63" />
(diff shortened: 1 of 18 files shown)


NicolasDorier:
Ping @PinpointTownes you make breaking change on OpenIddict3.0? How can I fix this?

image

PinpointTownes:
> Ping @PinpointTownes you make breaking change on OpenIddict3.0?

Yes. And it's pretty much the whole point of alpha packages (technically, SemVer also allows you to adopt breaking changes in all types of pre-releases, even RCs).

The `HandleAuthentication` API was removed recently - based on your feedback - and you can remove the calls safely, as your handlers are specific to a single `grant_type` and won't interfere with each other.

Regarding `EnableRequestCaching`, it was split into `EnableAuthorizationEndpointCaching` and `EnableLogoutEndpointCaching` (you very likely only need the first one in your case).

You'll also need a migration as the OpenIddict EF schema was updated to support the device code flow. I'm pretty sure we chatted about that on Skype last month.

That said, I see you have a pending PR to remove OpenIddict, so you should probably merge it before moving to 3.1, to save some time (there's no much point migrating to a new version of a lib' you're not going to use).

NicolasDorier:
@PinpointTownes I don't plan on merging the removal of OpenIddict.

PinpointTownes:
Roger. Feel free to ping me on Skype if you need help or hit any issue.

NicolasDorier:
Something broke in openiddict on https://github.com/btcpayserver/btcpayserver/pull/1259 then this PR should be ready.

PR #1259, opened 2020-01-10
Update libs

NicolasDorier:
Updating OpenIddict 3.0 because of some bugs fixed on .NET Core 3.1.
Updating other 3.1 libraries to prepare for the switch.

### DIFF
--- a/BTCPayServer/Hosting/Startup.cs
+++ b/BTCPayServer/Hosting/Startup.cs
@@ -161,8 +161,7 @@ namespace BTCPayServer.Hosting
                         .EnableStatusCodePagesIntegration()
                         .EnableAuthorizationEndpointPassthrough()
                         .EnableLogoutEndpointPassthrough()
-                        .EnableTokenEndpointPassthrough()
-                        .EnableRequestCaching()
+                        .EnableAuthorizationEndpointCaching()
                         .DisableTransportSecurityRequirement();
 
                     // Enable the token endpoint (required to use the password flow).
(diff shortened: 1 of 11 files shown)


NicolasDorier:
@PinpointTownes the tests do not pass with latest version. :(

PinpointTownes:
I cloned BTCPayServer locally and wasn't able to repro this specific error.

In my case, the `CanUseNonInteractiveFlows` test fails because you enabled the pass-through mode for the token endpoint, for which you have no corresponding MVC action (that's why I get a 404 error). This is a BTCPayServer bug, that was previously hidden by the fact you called `context.HandleAuthentication()` (that basically bypassed the pass-through mode flag). Remove `EnableTokenEndpointPassthrough()` from Startup and it should work.

Note: you'll also need to revamp your tests to stop using `JObject.Parse(content).ToObject<OpenIddictResponse>()`, as OpenIddict no longer providers a `JsonConverter` for JSON.NET (it now uses `System.Text.Json` instead). Here's how you could replace it:

```csharp
var result = System.Text.Json.JsonSerializer.Deserialize<OpenIddictResponse>(content);
```

BTCPayServer/Hosting/Startup.cs
```diff
@@ -162,7 +162,7 @@ private void ConfigureOpenIddict(IServiceCollection services)
                         .EnableAuthorizationEndpointPassthrough()
                         .EnableLogoutEndpointPassthrough()
                         .EnableTokenEndpointPassthrough()
-                        .EnableRequestCaching()
+                        .EnableLogoutEndpointCaching()
```
PinpointTownes: I suspect that's not what you wanted. You likely want `EnableAuthorizationEndpointCaching()` (chances are high the consent view won't work since it uses `RequestId`, that doesn't exist when you don't enable caching.

PR #1262, opened 2020-01-10
Api keys with openiddict

Kukks:
variant of #1245 but with openid still intact

This PR introduces a new authentication and authorization system, completely independent of the OpenIddict/OpenID work that has been done so far( and in fact, I propose to remove all related code as this should be sufficient for most use-cases while relying on less third-party code and a lot less code( migrations make up the biggest difference here)).

# Authentication
It follows the common API concept of having users on the system generating an API Key, to which they then use to authenticate and hit the API available. The usage approach is to add a header `X-APIKEY` with the API Key as the value. 

# Authorization
I've also added granular permissions, similar to what we had done for OpenId with the scopes. You can assign specific permissions per API key such as server management or store management. In addition, I've extended the store management permissions to allow specific store permissions too in case you would like to authorize an app for only a subset of stores under your account's controls.
The current permissions in the system are:
* CanModifyServerSettings - modify the server settings and manage users
* CanModifyStoreSettings - create or manage all stores under your account
* CanListStoreSettings - list all stores available to your API Key

# Application Integration
While it can be trivial to simply tell a user to generate and copy an API key to your application for BTCPay integration, I've added support for external applications to request an api-key. 
This is very similar to the OpenId Authorize view we added not too long ago.
The path is `api-keys/authorize` and the available options(query parameters) are:
* applicationName - string, optional, the name to show on the view
* redirect - url, optional, the URl to redirect to after the user consents, with the query paramters appended to it: `permissions`, `user-id`, `api-key`. If not specified, user is redirect to their API Key list.
* permissions - array of string, optional. format: `permissions=CanModifyServerSettings&permissions=CanModifyStoreSettings`. Those are currently the only 2 permissions you can request currently. If no permissions are specified, the API-key can still be useful as a means to link a BTCPay user to your own system.
* strict - boolean, optional, default: true. If permissions are specified, and strict is set to false, it will allow the user to reject some of permissions the application is requesting.
* selectiveStores - boolean, optional , default: false. If the application is requesting the `CanModifyStoreSettings` permission and selectiveStores is set to true, this allows the user to only grant permissions to selected stores under the user's control.
* applicationIdentifier - string, optional. If specified, BTCPay will check if there is an existing API key stored associated with the user that also has this application identifer, redirect host AND the permissions required match(takes `selectiveStores` and `strict` into account). `applicationIdentifier` is ignored if `redirect` is not specified.

# Use-case
Even without an API currently available(prototyping branch for a full API can be found at https://github.com/Kukks/btcpayserver/tree/api-preview), this Auth system is still useful. 
We can create a "Log in with BTCPay" feature with the authorize endpoint:
* Create your application and use the domain `bitpay2.com`
* Define in your application a BTCPay url, or ask the user for one
* Generate an authorize url: `btcpayserver.com/api-keys/authorize?applicationName=Bitpay2&redirect=bitpay2.com/btcpay-response&applicationIdentifier=bitpay2` 
* Show to user
* User navigates to url & accepts
* Browser redirects to `bitpay2.com/btcpay-response?user-id=1&api-key=xyz`
* Use user id and api key for whatever you want to do. The next time the user wants to log in, the user will be redirected directly without requesting permission.
* You now have a federated BTCPay user system!


BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -0,0 +1,344 @@
...
+            if (viewModel.StoreMode == AddApiKeyViewModel.ApiKeyStoreMode.Specific)
+            {
+                permissions.AddRange(viewModel.SpecificStores.Select(APIKeyConstants.Permissions.GetStorePermission));
+            }
+            else if (viewModel.StoreManagementPermission)
+            {
+                permissions.Add(APIKeyConstants.Permissions.StoreManagement);
+            }
+
+            if (viewModel.IsServerAdmin && viewModel.ServerManagementPermission)
```
NicolasDorier: `viewModel.IsServerAdmin` is in the control of the attacker, you should not use this as permission check. Instead you should use the IAuthorizationService

BTCPayServer/Security/APIKeys/APIKeyAuthorizationHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (context.User.Identity.AuthenticationType != APIKeyConstants.AuthenticationType)
+                return;
+
+            bool success = false;
+            switch (requirement.Policy)
+            {
+                case Policies.CanListStoreSettings.Key:
+                    var selectiveStorePermissions =
+                        APIKeyConstants.Permissions.ExtractStorePermissionsIds(context.GetPermissions());
+                    if (context.HasPermissions(APIKeyConstants.Permissions.StoreManagement) || selectiveStorePermissions.Any())
```
NicolasDorier:
```
if (blah)
 toto = true;
else
 toto = false
```
can be simplified to

```
toto = blah
```

NicolasDorier:
The tests do not pass.
The code seems ready aside of some nits. After the test pass will do deeper review and manual testing.

BTCPayServer/Controllers/ManageController.APIKeys.cs
Kukks: fixed

BTCPayServer/Security/APIKeys/APIKeyAuthorizationHandler.cs
Kukks: fixed

BTCPayServer/BTCPayServer.csproj
```diff
@@ -220,4 +220,8 @@
 	    <Pack>$(IncludeRazorContentInPack)</Pack>
 	  </Content>
 	</ItemGroup>
+
+	<ItemGroup>
+	  <_ContentIncludedByDefault Remove="Views\Authorization\Authorize.cshtml" />
```
NicolasDorier: ??

BTCPayServer.Data/Data/APIKeyData.cs
```diff
@@ -11,15 +13,31 @@ public class APIKeyData
...
+        public string Permissions { get; set; }
+        
+        public StoreData StoreData { get; set; }
+        public ApplicationUser User { get; set; }
+        public string[] GetPermissions() { return Permissions?.Split(';') ?? new string[0]; }
+
+        public void SetPermissions(IEnumerable<string> permissions)
         {
-            get; set;
+            Permissions = string.Join(';', permissions ?? new string[0]);
```
NicolasDorier: You should check that no `permissions` contains a `;` else, funny stuff can be done.

BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -75,41 +99,56 @@ protected override void BuildModel(ModelBuilder modelBuilder)
             modelBuilder.Entity("BTCPayServer.Data.ApplicationUser", b =>
                 {
                     b.Property<string>("Id")
-                        .ValueGeneratedOnAdd();
```
NicolasDorier: you need to keep `ValueGeneratedOnAdd`

BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -14,56 +14,80 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
                     b.HasIndex("StoreId");
 
+                    b.HasIndex("UserId");
+
                     b.ToTable("ApiKeys");
                 });
 
+            modelBuilder.Entity("BTCPayServer.Data.AddressInvoiceData", b =>
+                {
+                    b.Property<string>("Address")
```
NicolasDorier: need `ValueGeneratedOnAdd`

BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -437,23 +548,31 @@ protected override void BuildModel(ModelBuilder modelBuilder)
             modelBuilder.Entity("BTCPayServer.Data.StoreData", b =>
                 {
                     b.Property<string>("Id")
-                        .ValueGeneratedOnAdd();
```
NicolasDorier: Check all the `ValueGeneratedOnAdd` I am unsure if that matter in reality... the snapshot is only to create a new migration I think.

NicolasDorier:
Can you also remove the Designer file? I never figure out what are they useful for, but I always deleted them.

Kukks:
Updated the migrations and db stuff now

NicolasDorier:
tests fail

BTCPayServer/Views/Manage/AddApiKey.cshtml
```diff
@@ -0,0 +1,110 @@
...
+                    <span asp-validation-for="ServerManagementPermission" class="text-danger"></span>
+                    <p>@APIKeyConstants.Permissions.PermissionDescriptions[APIKeyConstants.Permissions.ServerManagement].Description.</p>
+                </div>
+            }
+            @if (Model.StoreMode == ManageController.AddApiKeyViewModel.ApiKeyStoreMode.AllStores)
+            {
+                <div class="list-group-item form-group">
+                    @Html.CheckBoxFor(model => model.StoreManagementPermission, new Dictionary<string,string>(){ {"class", "form-check-inline"}} )
+                 
+                    <label asp-for="StoreManagementPermission" class="h5">@APIKeyConstants.Permissions.PermissionDescriptions[APIKeyConstants.Permissions.StoreManagement].Title</label>
```
rockstardev: We really need some alias usage here, like `GetDescription(passConstant)`

BTCPayServer/Security/APIKeys/APIKeyRepository.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                return await queryable.ToListAsync();
+            }
+        }
+
+        public async Task CreateKey(APIKeyData key)
+        {
+            if (key.Type == APIKeyType.Legacy || !string.IsNullOrEmpty(key.StoreId) || string.IsNullOrEmpty(key.UserId))
+            {
+                throw new InvalidOperationException("cannot save a bitpay legacy api key with this repository");
```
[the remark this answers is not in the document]
Kukks: ` Or this is just to prevent legacy key creation through new method calls?`
This

Kukks:
> 
> 
> @Kukks can you start pushing these bigger PRs as branches on origin repository? That way I can join in with these small cleanups instead of asking you to do all the changes.

You should be able to edit the branch as well since I have the ` Allow edits from maintainers. ` checked

BTCPayServer/Security/AuthenticationSchemes.cs
```diff
@@ -12,5 +12,6 @@ public class AuthenticationSchemes
         public const string Cookie = "Identity.Application";
         public const string Bitpay = "Bitpay";
         public const string OpenId = OpenIddictValidationAspNetCoreDefaults.AuthenticationScheme;
+        public const string ApiKey = "ApiKey";
```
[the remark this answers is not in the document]
Kukks: done

BTCPayServer/Views/Manage/AddApiKey.cshtml
Kukks: done

BTCPayServer/Views/Manage/APIKeys.cshtml
```diff
@@ -0,0 +1,48 @@
...
+    ViewData.SetActivePageAndTitle(ManageNavPages.APIKeys, "Manage your API Keys");
+}
+
+<partial name="_StatusMessage"/>
+<h4>API Keys</h4>
+<table class="table table-lg">
+    <thead>
+    <tr>
+        <th >Key</th>
+        <th >Permissions</th>
```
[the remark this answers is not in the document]
Kukks: done

rockstardev:
> You should be able to edit the branch as well since I have the `Allow edits from maintainers.` checked

Let me give that a try tonight... so I can check off "coding on the same feature branch with Kukkie" off the bucket list.

NicolasDorier:
Do not use X-APIKEY for two reason:
1. `X-` [is deprecated](https://tools.ietf.org/html/rfc6648).
2. This is controlling authorization.

Instead, use like github. `Authorization: token $GITHUB_TOKEN`

Kukks:
> 
> 
> Do not use X-APIKEY for two reason:
> 
>     1. `X-` [is deprecated](https://tools.ietf.org/html/rfc6648).
> 
>     2. This is controlling authorization.
> 
> 
> Instead, use like github. `Authorization: token $GITHUB_TOKEN`

@NicolasDorier Done

BTCPayServer.Data/Migrations/20200119130108_ExtendApiKeys.cs
```diff
@@ -0,0 +1,83 @@
...
+                name: "UserId",
+                table: "ApiKeys",
+                maxLength: 50,
+                nullable: true);
+
+            migrationBuilder.CreateIndex(
+                name: "IX_ApiKeys_UserId",
+                table: "ApiKeys",
+                column: "UserId");
+            if (this.SupportDropForeignKey(migrationBuilder.ActiveProvider))
```
NicolasDorier: Why is this conditional?

Kukks: Adding foreign keys breaks sqlite.. (tests failed before this was conditional)

NicolasDorier: Mmh maybe you should add a `SupportAddForeignKey` instead.

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -0,0 +1,346 @@
...
+                    UserId = new[] {_userManager.GetUserId(User)}
+                })
+            });
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> RemoveAPIKey(string id)
+        {
+            await _apiKeyRepository.Remove(id, _userManager.GetUserId(User));
+            return RedirectToAction("APIKeys", new {StatusMessage = "API Key removed"});
```
NicolasDorier: Please use the `TempData` instead like we do everywhere

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -0,0 +1,346 @@
...
+                });
+                return RedirectToAction("APIKeys");
+            }
+
+            return View("AddApiKey", await SetViewModelValues(new AddApiKeyViewModel()));
+        }
+
+        [HttpGet("~/api-keys/authorize")]
+        public async Task<IActionResult> AuthorizeAPIKey( string[] permissions, string applicationName = null, Uri redirect = null,
+            bool strict = true, bool selectiveStores = false, string applicationIdentifier = null)
```
NicolasDorier: This is a huge security flaw, please remove this feature.

Now if a `applicationIdentifier` becomes public, any attacker can fetch the API keys and use the keys to impersonate the applicationIdentifier.

Kukks: Why is it a security flaw? The `applicationIdentifier` can be public as it is linked to the redirect domain and can only be used when redirecting to it?

Kukks: In fact, I have a test for exactly this: 
https://github.com/btcpayserver/btcpayserver/pull/1262/files/1551ccb08f54f74f8cadbe30715f923a634e31ea#diff-e3ffadb8e98d10c6155d685f59cbb51dR216

NicolasDorier: Alice wants to use Carol's app. (applicationIdentifier = abc)
Malicia wants to get Alice's key.

Malicia ask Alice to to browse `https://alice.btcpayserer.org/api-keys/authorize?applicationIdentifier=abc&redirect=https://malicia.org`.

Malicia now has Alice's apikey.

NicolasDorier: I prefer you remove all reference to Redirect Url. I know that in theory, it saves a copy/paste, but this open up a can of worms.

Kukks: In that scenario, Alicia would be prompted to create a new key and also promoted that the browser would be redirected to malicia.org

NicolasDorier: It is not what is being done if a key is already created.

Kukks:
@NicolasDorier  I have removed all the redirect url + application Identifier functionality to get this moving.

BTCPayServer/Controllers/ManageController.APIKeys.cs
MrNaif2018: Hello! Writing it there not to be lost.
This issue could be handled like so:
When creating api key, also store redirect url(in parsed form, host:port) in database.
When `/api-keys/authorize` page is hit, fetch an api key with that app id and parsed redirect url.
If it exists-redirect instantly.
If it doesn't exist, fetch other api keys by that app id, if such a key exists, present that to a user with a warning like:
`The application has already been authorized before at {{ old domain from database}}, but is now requiring a different redirect url({{ parsed redirect url from query params}} ). Proceed with Caution`
And present 4 options to the user:
- Authorize app(one time) - basically just redirect to that url, adding api token and other data as query parameters, but not changing the database. So next time same URL is hit, same warning page is presented.
- Change current key's redirect URL - edit redirect url in database to the newly parsed one, and perform the redirect. Next time visiting same URL instant redirect will be done, but when visiting old URL warning will be displayed now
- Generate a new key - generate another key with that new redirect url, allowing two redirect URL's coexist for one app id. Both pages will instantly redirect now
- Cancel - well, that's just cancel.
Just my idea how to keep it secure and still functional.

NicolasDorier: I think such UX is still confusing for normal users, and does not seem really better than just adding a API key with a label that can be revoked and just copy/pasting that.

Kukks: https://github.com/btcpayserver/btcpayserver/pull/1363

Kukks: It is more useful than you think from a third party app perspective. For example, a native app would have horrible UX on mobile to set this up without a custom url

PR #1264, opened 2020-01-10
CoinGecko Rate Provider

Kukks:
The CoinGecko rate provider is similar to the bitcoin average one, in that you can ask it for a rate from its aggregated sourcing or you can get rates from specific exchanges. I've added support for both.
I haven't integrated it or replaced coinaverage just to see if we should use it as the default and switch everyone to it or what other action to take.


Kukks:
IMo, we should move anyone using coinaverage to coingecko automatically. We can leave the settings for coinaverage and they can set it back again if they want to pay for it

BTCPayServer.Rating/Providers/CoinGeckoRateProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+using BTCPayServer.Rating;
+using Microsoft.AspNetCore.Http;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace BTCPayServer.Services.Rates
+{
+    public class CoinGeckoRateProvider: IRateProvider, IHasExchangeName
+    {
+        static HttpClient Client = new HttpClient()
```
NicolasDorier: inject the `HttpClient` in the constructor.

Kukks:
I've updated PR, it now replaces CoinAverage as the default + also loads supported exchanges from it and the generic exchange rate providers from it.  
We now have waaaaay more exchange ticket support.

NicolasDorier:
tests don't pass

BTCPayServer.Rating/Services/RateProviderFactory.cs
```diff
@@ -98,25 +97,26 @@ public void InvalidateCache()
...
-        private void InitExchanges()
+        public async Task InitExchanges()
         {
             // We need to be careful to only add exchanges which OnGetTickers implementation make only 1 request
-            Providers.Add("binance", new ExchangeSharpRateProvider("binance", new ExchangeBinanceAPI(), true));
-            Providers.Add("bittrex", new ExchangeSharpRateProvider("bittrex", new ExchangeBittrexAPI(), true));
-            Providers.Add("poloniex", new ExchangeSharpRateProvider("poloniex", new ExchangePoloniexAPI(), true));
-            Providers.Add("hitbtc", new ExchangeSharpRateProvider("hitbtc", new ExchangeHitBTCAPI(), true));
-            Providers.Add("ndax", new ExchangeSharpRateProvider("ndax", new ExchangeNDAXAPI(), true));
+            Providers.TryAdd("binance", new ExchangeSharpRateProvider("binance", new ExchangeBinanceAPI(), true));
```
NicolasDorier: It should be `Add` not `TryAdd`.

BTCPayServer/Views/Stores/Rates.cshtml
```diff
@@ -21,9 +21,9 @@
                     <span>Rate script allows you to express precisely how you want to calculate rates for currency pairs.</span>
                     <p class="text-muted">
                         <b>Supported exchanges are</b>:
-                        @for (int i = 0; i < Model.AvailableExchanges.Length; i++)
+                        @for (int i = 0; i < Model.AvailableExchanges.Count; i++)
                         {
-                            <a href="@Model.AvailableExchanges[i].Url">@Model.AvailableExchanges[i].Name</a><span>@(i == Model.AvailableExchanges.Length - 1 ? "" : ",")</span>
+                            <span>@Model.AvailableExchanges.ElementAt(i).Value</span><span>@(i == Model.AvailableExchanges.Count - 1 ? "" : ",")</span>
```
NicolasDorier: Should keep the link so people can see which is the source we take.

NicolasDorier:
tests don't pass

cryptofuture:
I tried it and it seems it still tries coinaverage, even removed api keys from the settings. Settings tab still would be with coinaverage right?

Kukks:
Have you updated to .147? Store->Rates->Coin Gecko

cryptofuture:
Store rate Coin Gecko is here, but in main settings its only coinaverage. How I reset for coingecko for every store and user?

Kukks:
There aren't any main settings for Coin Gecko. If you create a new store, it will be with Coin Gecko. If you have an existing store, it stays configured.

cryptofuture:
> There aren't any main settings for Coin Gecko. If you create a new store, it will be with Coin Gecko. If you have an existing store, it stays configured.
Could you suggest an sql in psql to reset all stores in the table to coingecko? As Its not possible to set those to coin gecko manually.

Kukks:
No, messing around the database directly should only be done for extreme scenarios such as losing your admin credentials. 
To complicate further, the selected rate provider is saved as a stringified json converted to a Blob column type. 
@NicolasDorier  I think we should consider switching every store from CoinAverage to CoinGecko for anyone without bitcoin average credentials saved in their server on startup

cryptofuture:
> @NicolasDorier I think we should consider switching every store from CoinAverage to CoinGecko for anyone without bitcoin average credentials saved in their server on startup
Good idea!

britttttk:
Concept Ack - Prefer not to default to payment required features if possible.
>I think we should consider switching every store from CoinAverage to CoinGecko for anyone without bitcoin average credentials saved in their server on startup

Thanks for CoinGecko solution! @Kukks @NicolasDorier

PR #1265, opened 2020-01-11
Update ListApps.cshtml

radWorx:
From ListApps, View App opens new tab.

### DIFF
--- a/BTCPayServer/Views/Apps/ListApps.cshtml
+++ b/BTCPayServer/Views/Apps/ListApps.cshtml
@@ -58,7 +58,7 @@
                                     {
                                         <a asp-action="@app.UpdateAction" asp-controller="Apps" asp-route-appId="@app.Id">Settings</a><span> - </span>
                                     }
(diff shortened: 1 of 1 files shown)


bolatovumar:
People might already be used to the existing behaviour so changing it now might confuse people.

NicolasDorier:
@pavlenex this open app on a new tab when one click "View App" button instead of changing the current page.

I don't really know what's the best.

pavlenex:
@NicolasDorier An unwritten rule is that the new tab is used to open external website links, for anything internal, it opens up in a current page, so I'm not that fond on changing what we currently   have.

radWorx:
It's a UX thing in my opinion. Keep the ListApps page open and App in new tab as opposed  haveing to click back when on CF or PoS page.

rockstardev:
I propose proceeding by creating linked "New tab" icon next to the "View" link... that opens new tab and re-uses that tab for App if user clicks link again.

We can replicate this with split button in Settings (where we have ViewApp opening link in new tab). For example:

split-button

pavlenex:
Ok, that sounds like a good trade-off imo.

rockstardev:
Continuing in #1303

PR #1283, opened 2020-01-16
Remove the next address to pay to from Invoice details page (Fix #1056)

### LINKED ISSUE #1056: Wrong address and due amount on "paid summary" of invoice page.
I had and invoice in my Invoices page, which one was marked as paidPartial. To check details i opened invoice:

image

Well, first of all i have two BTC addresses in invoice page. As Pavlenex mentioned, this is because partialpaid status (on partial paid new addr is generated). So that's not the issue. That is fine.
But as you can see on status page, the addr 34idTd8JmCq8WnntpiGAn5isf1yMLgNZmQ is mentioned as used address in "paid summary", while in fact, the only address which was used is 3EuCWPZkX2YWQs55H9ySpSQi64LAMzmboY
Also, as you can see from tx details, the amount paid is 0.01000000 BTC, while invoice created for 0.01000099 BTC (99 more sats than paid). But in the summary view it shows due 199 sats. So the wrong address is entioned, as well as amount is not precise.

https://chain.so/address/BTC/3EuCWPZkX2YWQs55H9ySpSQi64LAMzmboY
https://chain.so/address/BTC/34idTd8JmCq8WnntpiGAn5isf1yMLgNZmQ

* Version of BTCPay Server:  v1.0.3.132
* Deployment method: BTCPayJungle
* Other relevant environment details: conversion enabled using changelly

**To Reproduce**
1. Create invoice
2. Pay it in partial
3. Wait until invoice is expired
4. Open invoice detailed view in INVOCES page

**Expected behavior**
It should show that 3EuCWPZkX2YWQs55H9ySpSQi64LAMzmboY was used to pay invoice and the due amount is 99 sats (100 sats less then it shows to me)

**Actual behavior**
It seems that btcpay uses first invoice bitcoin address, not actual one. I have no idea, why the amount does not match as well.

NicolasDorier:
This address was confusing. It represent the next address to pay to, but could easily confuse users as #1056 shows.

Before
image

Now
image

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -104,7 +104,6 @@ namespace BTCPayServer.Controllers
                 cryptoPayment.Paid = _CurrencyNameTable.DisplayFormatCurrency(accounting.CryptoPaid.ToDecimal(MoneyUnit.BTC), paymentMethodId.CryptoCode);
                 cryptoPayment.Overpaid = _CurrencyNameTable.DisplayFormatCurrency(accounting.OverpaidHelper.ToDecimal(MoneyUnit.BTC), paymentMethodId.CryptoCode);
                 var paymentMethodDetails = data.GetPaymentMethodDetails();
-                cryptoPayment.Address = paymentMethodDetails.GetPaymentDestination();
                 cryptoPayment.Rate = ExchangeRate(data);
                 model.CryptoPayments.Add(cryptoPayment);
             }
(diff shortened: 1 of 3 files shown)


Kukks:
how about we rename it instead "Current unused address"

NicolasDorier:
@Kukks I thought about it, but I think it is still strange to find it here.

Kukks:
Is there a place to see the current address at least?

NicolasDorier:
@Kukks yes, it is expandable in the Invoices page. (which is more useful than going to invoice details)

Kukks:
Good enough for me

NicolasDorier:
Apparently edgars was the one bitching about the page, so waiting for his feedback

bolatovumar:
This is indeed confusing as the address looks like the address to which the initial payment was made, not the next address to pay to.

pavlenex:
Since Edgars is MIA for quite a while (Tried reaching out to him privately a few times), I think we can merge this one, it's obviously confusing and bad UX.

PR #1286, opened 2020-01-17
Refactor rate handling to prevent error of exchange name and re-reroute coinaverage to coingecko

NicolasDorier:


NicolasDorier:
This PR remove the requirement for exchange providers to know their own name, removing some foot shooting risks.

This also allow to easily reroute `coinaverage` rate rules to coingecko.

PR #1289, opened 2020-01-18
Move the Receive to Transactions page

NicolasDorier:
This is meant to be an improvement over the current UX.
In the current UX, we show the Receive as a separate tab in the wallet menu.

This tries to put it in the /transactions page

image

image

To be honest I don't think this look nice, but posting it here in case somebody has an idea to improve it.


NicolasDorier:
This don't pass the tests, will fix if there is a consensus to change the current UX. Maybe it is not really needed after all.

pavlenex:
Honestly I don't see how this improves the UX, it just hides an option all users are used to seeing on normal wallets, send/receive.

NicolasDorier:
Ok closing that then

PR #1294, opened 2020-01-20
Add currency precision based on network

Kukks:


BTCPayServer/Payments/PaymentTypes.cs
```diff
@@ -66,6 +66,11 @@ public override string ToString()
         public abstract ISupportedPaymentMethod DeserializeSupportedPaymentMethod(BTCPayNetworkBase network, JToken value);
 
         public abstract string GetTransactionLink(BTCPayNetworkBase network, string txId);
+
+        public virtual int GetDivisibility(BTCPayNetworkBase network)
```
NicolasDorier: This is used only at one place, remove this one.

BTCPayServer.Common/Altcoins/Monero/MoneroLikeSpecificBtcPayNetwork.cs
```diff
@@ -3,5 +3,6 @@ namespace BTCPayServer
     public class MoneroLikeSpecificBtcPayNetwork : BTCPayNetworkBase
     {
         public int MaxTrackedConfirmation = 10;
+        public override int Divisibility { get; set; } = 12;
```
NicolasDorier: there is no need to override anything

BTCPayServer.Common/BTCPayNetwork.cs
```diff
@@ -124,7 +124,7 @@ public abstract class BTCPayNetworkBase
         public string CryptoCode { get; internal set; }
         public string BlockExplorerLink { get; internal set; }
         public string DisplayName { get; set; }
-
+        public virtual int Divisibility { get; set; } = 8;
```
NicolasDorier: no need for it to be virtual

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -420,14 +420,14 @@ public InvoiceResponse EntityToDTO()
...
                 cryptoInfo.Rate = info.Rate;
-                cryptoInfo.Price = subtotalPrice.ToString();
+                cryptoInfo.Price = subtotalPrice.ToString(false, true);
 
-                cryptoInfo.Due = accounting.Due.ToString();
-                cryptoInfo.Paid = accounting.Paid.ToString();
-                cryptoInfo.TotalDue = accounting.TotalDue.ToString();
+                cryptoInfo.Due = accounting.Due.ToString(false, true);
+                cryptoInfo.Paid = accounting.Paid.ToString(false, true);
+                cryptoInfo.TotalDue = accounting.TotalDue.ToString(false, true);
```
NicolasDorier: why changing this?

Kukks: based on discussion https://github.com/ElementsProject/elements/issues/805

Kukks: so that we dont generate bip21 with too many trailing zeros

BTCPayServer/Payments/PaymentTypes.cs
NicolasDorier: what I mean is that you can just inline `network?.Divisibility ?? 8` where you use it.

Kukks: What if there is a manual payment type that has no network defined and uses only 2decimal points?

BTCPayServer/Services/Invoices/InvoiceEntity.cs
NicolasDorier: This part is unrelated with bip21

NicolasDorier: Rather change `GenerateBIP21` to accept `Money` and format it correctly there

BTCPayServer/Payments/PaymentTypes.cs
NicolasDorier: Then you will need to pass divisibility as an arg to Calculate instead because PaymentTypes are singleton so there should be only one PaymentTypes.Manual

BTCPayServer/Services/Invoices/InvoiceEntity.cs
Kukks: done

PR #1304, opened 2020-01-24
Update display text on the view model.

bitcoinbrisbane:
Simple patch to fix the below:

Screen Shot 2020-01-24 at 5 30 52 pm


rockstardev:
Looking good, thanks for contribution @bitcoinbrisbane

bitcoinbrisbane:
@rockstardev happy to.  Wish I had more time to commit more!

PR #1307, opened 2020-01-26
Add nullable enable directive to HttpClientRequestMaker.MakeRequestAsync

bolatovumar:
My VSCode complains that `The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. [BTCPayServer.Rating]`. To fix this I added the `#nullable enable` directive. See here for details: https://stackoverflow.com/questions/55492214/the-annotation-for-nullable-reference-types-should-only-be-used-in-code-within-a

### DIFF
--- a/BTCPayServer.Rating/Providers/HttpClientRequestMaker.cs
+++ b/BTCPayServer.Rating/Providers/HttpClientRequestMaker.cs
@@ -196,11 +196,9 @@ namespace BTCPayServer.Services.Rates
                     throw new APIException(text);
                 }
                 api.ProcessResponse(new InternalHttpWebResponse(webHttpResponse));
-                #nullable enable
-                Action<IAPIRequestMaker, RequestMakerState, object>? requestStateChanged = RequestStateChanged;
-                if (requestStateChanged != null)
+                if (RequestStateChanged != null)
                 {
-                    requestStateChanged!(this, RequestMakerState.Finished, text);
+                    RequestStateChanged(this, RequestMakerState.Finished, text);
                     return text;
                 }
                 return text;
--- a/BTCPayServer.Rating/Providers/HttpClientRequestMaker.cs
+++ b/BTCPayServer.Rating/Providers/HttpClientRequestMaker.cs
@@ -196,6 +196,7 @@ namespace BTCPayServer.Services.Rates
                     throw new APIException(text);
                 }
                 api.ProcessResponse(new InternalHttpWebResponse(webHttpResponse));
+                #nullable enable
                 Action<IAPIRequestMaker, RequestMakerState, object>? requestStateChanged = RequestStateChanged;
                 if (requestStateChanged != null)
                 {


NicolasDorier:
No need of nullable here.
Just remove remove the `!` in `requestStateChanged!(this, RequestMakerState.Finished, text);`

bolatovumar:
@NicolasDorier it's actually the `?` that it complains about. I assumed both `?` and '!' are there for a reason so didn't want to remove them. Can I remove them then?

NicolasDorier:
Indeed you need to remove both.
FYI nullable are quite cool, but we don't use it yet in this project.
If for example you have `Blah a`, then the compiler make sure that `a` is never null or it fails compiling. It catch lot's of bug, as NullReferenceException are happening very often.

`Blah? a` means that `a` can be null.

Now if you do `a.DoSomething` the compiler will complained if you did not check `a` for null, because it can throw at runtime. If you are absolutely sure it is not null, you can do `a!.DoSomething`.

The compiler though in this context should be able to know that `requestStateChanged` is not null, so I wonder why it would complain if nullable were activated without the `!`.

Anyway, we don't use nullable for now.

rockstardev:
@bolatovumar we closing this?

rockstardev:
@bolatovumar reopen if needed.

bolatovumar:
@rockstardev I will make a follow-up PR which removes both `?` and `!` as @NicolasDorier suggested.

rockstardev:
You can also reopen this then force push update to branch to preserve history @bolatovumar

rockstardev:
I would say also remove that assigning:

```
Action<IAPIRequestMaker, RequestMakerState, object>? requestStateChanged = RequestStateChanged;
if (requestStateChanged != null)
{
    requestStateChanged!(this, RequestMakerState.Finished, text);
    return text;
}
```

instead just have:

```
if (RequestStateChanged != null)
{
    RequestStateChanged(this, RequestMakerState.Finished, text);
    return text;
}
```

Nullable (i.e. `Blah? a`) was introduced only for value types (current state of BtcPayServer codebase); here is seems as it was used for reference type... casting to nullable Action class. Maybe it's copy paste from codebase that has #nullable directive enabled. So, remove ?, ! and that assigning (as it's unnecessary).

Unless @NicolasDorier disagrees ofc.

bolatovumar:
@rockstardev alright, done.

NicolasDorier:
```
Action<IAPIRequestMaker, RequestMakerState, object>? requestStateChanged = RequestStateChanged;
if (requestStateChanged != null)
{
    requestStateChanged!(this, RequestMakerState.Finished, text);
    return text;
}
```

Can be rewritten correctly as

```csharp
RequestStateChanged?(this, RequestMakerState.Finished, text);
return text;
```
This remove the risk of having `RequestStateChanged` turning to null between the time of check and time of invocation.

rockstardev:
@NicolasDorier you mean `RequestStateChanged?.Invoke(this, RequestMakerState.Finished, text);`, right?

Also, here is the link to extended discussion we had on the commit, for sake of future generations: https://github.com/btcpayserver/btcpayserver/commit/00747906849f093712c3907c99404c55b3defa66#r37022103

Issue #921, opened 2019-07-15
Some URLs are CamelCased while most are lowercased


woutersamaey:
**Describe the problem/bug**
Some URLs are have Capitals while most URLs are all lowercase.
The app is case insensitive, so didn’t want to report this as a real issue. Would be prettier though.

After looking at the code, it seams the “route” is not set for the Capitalized controllers.

**Your environment**
* Version of BTCPay Server: latest
* Deployment method: docker

**Expected behavior**
All URLs should be lowercase and have "-" between words.

**Actual behavior**
Some URLs aren't pretty:
- /Account/Login
- /Account/Logout
- /Account/ForgotPassword
- /Account/Register
- /Manage/Index
- /Manage/ChangePassword
- /Manage/TwoFactorAuthentication
- /Manage/U2FAuthentication

woutersamaey:
@NicolasDorier here it is.

woutersamaey:
A solution seams already available, but I'm not experienced enough to implement.
Details are here: https://stackoverflow.com/questions/40334515/automatically-generate-lowercase-dashed-routes-in-asp-net-core/45777372#45777372

This will automatically slugify the URLs, meaning lowercase and only [a-z0-9\-]

NicolasDorier:
@woutersamaey you only need to add say `[Route("login")]` on top of each action.

bolatovumar:
@NicolasDorier you could do that but it would get rather tedious to do it everywhere. Seems like the method posted by @woutersamaey would be easier and solve the issue everywhere at once. Unfortunately it requires .NET Core v2.2.0 or later. Any change BTCPay could be upgraded to use .NET Core v2.2.0 or later (latest is v2.2.5)?

Link to MS documentation: https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-2.2#use-a-parameter-transformer-to-customize-token-replacement
Link to Github issue where this was implemented: https://github.com/aspnet/Routing/issues/186#issuecomment-421493328

NicolasDorier:
@bolatovumar I don't think this is tedious. Those are the only links impacted, this is easy to review and our tests cover all.

We won't move to .net core 2.2 because .net core 3.0 will release in a few week.

woutersamaey:
I'll try the "clean" solution once we are on 3.0.
[...]

PR #1308, opened 2020-01-26
Slugify URLs

bolatovumar:
address #921

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -85,7 +85,7 @@ namespace BTCPayServer.Tests
                 var email = s.RegisterNewUser();
                 s.Driver.FindElement(By.Id("Logout")).Click();
                 s.Driver.AssertNoError();
-                Assert.Contains("Account/Login", s.Driver.Url);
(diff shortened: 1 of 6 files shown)


NicolasDorier:
The tests do not pass, it must break something

Kukks:
Do all the links still work without the slug version?

bolatovumar:
@NicolasDorier yeah, it's because some tests try to hit the old non-slugified urls.

@Kukks no, old non-slugified URLs do not work which might be a problem.

NicolasDorier:
I think it is also easier to directly edit the Route attributes.

bolatovumar:
Cool, I will just add the attributes then.

bolatovumar:
Added attribute routes for  `Account` and `Manage` controllers. Can do other controllers if this change looks good to you @NicolasDorier @Kukks.

NicolasDorier:
Sorry for looking at it only now, tests fail.

bolatovumar:
@NicolasDorier sorry, was busy lately. Will fix this up soon.

bolatovumar:
I really don't understand why the remaining tests are failing so I give up now.

PR #1309, opened 2020-01-26
Do not show assets in sync modal

Kukks:


NicolasDorier:
Can you remove them as well from the logs when the config is loaded?

Issue #275, opened 2018-09-18
Improve look of 404 (and other) error pages


Kukks:
When a user clickc on a pay button and the feature is disabled on the store, they are greeted with 
![image](https://user-images.githubusercontent.com/1818366/45693191-d7ff0580-bb5c-11e8-855a-cff28d26e6e6.png)

perhaps we should have a more friendly page version

rockstardev:
This is wider issue... how our error pages (404, 500, etc) are displayed... I guess I can take this on when I have time.

PR #1314, opened 2020-02-01
Adding error pages to handle HTTP errors

rockstardev:
Resolves: #275

As you can tell from the screenshot the _look_ of pages is vastly improved, as requested in issue.

errorpage02


BTCPayServer/Views/Error/404.cshtml
```diff
@@ -0,0 +1,15 @@
+﻿@{
+    ViewData["ErrorTitle"] = "404 - Page not found";
+}
+
+<p class="lead-login">
+    This is like searching for person more beautiful than <a href="https://twitter.com/NicolasDorier" target="_blank">Nicolas Dorier</a>.
```
Kukks: `for a person` *

BTCPayServer/Hosting/Startup.cs
```diff
@@ -241,6 +241,10 @@ private static void ConfigureCore(IApplicationBuilder app, IWebHostEnvironment e
             forwardingOptions.KnownProxies.Clear();
             forwardingOptions.ForwardedHeaders = ForwardedHeaders.All;
             app.UseForwardedHeaders(forwardingOptions);
+
+
+            app.UseStatusCodePagesWithReExecute("/Error/Handle", "?statusCode={0}");
```
NicolasDorier: Can you check that the API calls are not impacted by this change? Say if you call a get invoice on the bitpay API, it should not rewrite.

rockstardev: You are handling `BitpayHttpException` in middleware so they continue to produce same responses as before.

https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Hosting/BTCpayMiddleware.cs#L68

The problem is that NBitpayClient doesn't expose errors returned so we can't test for them.

rockstardev: Figured it out... it was bug in the pipeline; see: 82393eb8bbcaa10865b21e890c01f41eeb389660

PR #1321, opened 2020-02-05
BIP79

Kukks:
This PR adds support for [BIP79 ](https://github.com/bitcoin/bips/blob/master/bip-0079.mediawiki) on both the invoicing side and the spending wallet side. I've also extended the BIP so that there is support for PSBT to the endpoint( you can still send a tx in hex instead as per spec). 

## How does it work?
Invoices generated by stores that enable BIP79 and have the correct wallet configuration, will have an addition to the BIP21 payment url specifying an endpoint. This url is used to generate the QR code and the payment button. 

Wallets that support BIP79 will create a transaction and collaborate with the store wallet automatically to create a modified transaction that enhances privacy by breaking certain blockchain analysis heuristics, also known as [Payjoin](https://en.bitcoin.it/wiki/PayJoin) or [P2EP](https://blockstream.com/2018/08/08/en-improving-privacy-using-pay-to-endpoint/). 

## Why should I enable BIP79 as a merchant?
* It consolidates your incoming funds into bigger utxos without costing you transaction fees.
* It breaks certain heuristics for blockchain analysis companies providing greater privacy for your customers and yourself. With greater use of Payjoin long-term, blockchain analysis companies lose one of their strongest tooling to track, harass and censor Bitcoin users.

[Take a look at the following article for a greater explanation of why payjoin is amazing.
](https://joinmarket.me/blog/blog/payjoin/)
## How can I set up a store to use BIP79
* Your store MUST use a segwit derivation scheme
* Your store MUST generate the wallet using the seed and store the private keys on the server
* You MUST use BTC
* Store Settings -> Checkout Settings -> `BIP79 Enabled` must be checked

## How can I see which invoices were paid using BIP79

Check your completed invoices and their payments list. If there was a transaction that used BIP79, it will note down the amount contributed from your store's wallet.

## How can I pay a BIP79 enabled invoice?
Use a BIP79 compatible client. In the case that you do not have one, BIP79 is backwards compatible with any BIP21 wallet. 

The supported wallet list for BIP79 is currently non-existent but I've added support in the BTCPay spending wallet if you use any of the following signing methods:
* BTCPay Vault
* Ledger 
* Seed
* NBX Seed

## Adding support for BIP79 as a wallet
I've extended BIP79 with PSBT support. If you are implementing support in your wallet for BIP79, BTCPay Server will accept any of the following transaction formats:
* PSBT hex format
* PSBT base64 format
* raw transaction hex format

As per the BIP79 spec, the transaction must 
* be signed and ready to be broadcasted when sent over to the endpoint
* have all inputs be segwit
* inputs used cannot have been used to attempt another BIP79 invoice payment that is still pending
* fee should be slightly higher than usual as the server adds an extra input.

and that:
* after receiving the proposed transaction, it should be signed and broadcasted within 5 mins else the original transaction is broadcasted as an anti-spam mechanism.

If you visit BTCPay through its Tor hidden service(.onion), the endpoint will also be available using .onion so you should have support for sending http requests to onion endpoints.

## Do I need to keep all my incoming funds in a hot wallet?
Payjoin requires the server to have a hot wallet to be able to contribute some inputs and partially sign transactions. 
My recommendation is to pair your hot wallet with BTC Transmuter to schedule fund forwarding to a cold wallet occasionally.


Kukks:
blocked by #1322 
blocked by #1323

rockstardev:
Taking myself out as reviewer because I don't have in-depth knowledge. Commits look good though, keep it up Kukkie.

NicolasDorier:
So there is two thing that bother me with the general approach:

Using a hotwallet is not very safe, also, allowing payjoin also mean that a customer can enumerate the utxos of the merchant. So payjoin has a form of disutility to it.

As such, I think it should be a different payment method on which the merchant can charge an extra network cost for this disutility.

Kukks:
> So there is two thing that bother me with the general approach:
> 
> Using a hotwallet is not very safe, also, allowing payjoin also mean that a customer can enumerate the utxos of the merchant. So payjoin has a form of disutility to it.
> 
> As such, I think it should be a different payment method on which the merchant can charge an extra network cost for this disutility. But with the current UX, this is an impossible thing to do.
> 
> One way would be to modify BIP79 to allow the merchant to set up extra fee.

There is anti spam protection to not allow users to enumerate multiple utxos from the merchant. 

When you request to collaborate on a payjoin, you send in a broadcastable TX, signed and finalized that pays the invoice to completion.if the user does not use the payjoin transaction proposed by the merchant in a timely fashion, BTCPay broadcasts the original transaction.

Also, if a user submits transactions using the same inputs to multiple invoices as a way to circumvent the spam, the payjoin is rejected and no utxos are exposed. 

If the original transaction is broadcasted because the user does not publish the payjoin transaction, any exposed utxos are placed in a prioritized list for the next payjoin. 

I think separating the payjoin from the normal payment method and allowing the merchant to charge extra for it is counter productive and makes the feature much less approachable, from both parties. The benefits to the merchant should be enough compensation for the risk presented( enhanced privacy, free utxo consolidation)  and the costs to the users is enough already - it has higher fees due to more inputs). Further to that, the feature should be used as a seamless extension to bitcoin payment urls instead of reinventing the wheel with new protocols. If both the server and the wallet support payjoin, this can be communicated sufficiently with bip21.

NicolasDorier:
Make sense. I think one feature we will need then is a way to automatically transfer money to another xpub when it pass a threshold like the transmutter does but integrated to btcpay.

This would at least limit the money at risk by the merchant. Can do that later though.

RHavar:
It's not a real wallet, but for testing if you use bitcoin-core you can send using the utility:
https://github.com/rhavar/bustapay

with `bustapay send $BITCOIN_ADDRESS $BUSTAPAY_URL $AMOUNT_IN_BITCOIN`

BTCPayServer.Tests/TestAccount.cs
```diff
@@ -125,6 +125,20 @@ await store.AddDerivationScheme(StoreId, new DerivationSchemeViewModel()
             return new WalletId(StoreId, cryptoCode);
         }
 
+        public async Task EnablePayJoin()
+        {
+            var storeController = parent.PayTester.GetController<StoresController>(UserId, StoreId);
+            var checkoutExperienceVM =
+                Assert.IsType<CheckoutExperienceViewModel>(Assert
+                    .IsType<ViewResult>(storeController.CheckoutExperience()).Model);
```
NicolasDorier: this should not be checkout experience. Checkout experience is about cosmetic, not about actual logic.

BTCPayServer/Services/SocketFactory.cs
```diff
@@ -1,25 +1,66 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Configuration;
+using MihaZupan;
 using NBitcoin.Protocol.Connectors;
 using NBitcoin.Protocol;
 
 namespace BTCPayServer.Services
 {
+    public class Socks5HttpClientFactory : IHttpClientFactory
```
NicolasDorier: Can you keep the whole HttpClient and tor stuff in a future PR? BIP79 does not require Tor, don't do it for now, we need a small PR so we can easily review and ship so people can actually try the user experience and give feedback before we go farther.

Kukks: Done

BTCPayServer.Tests/TestAccount.cs
Kukks: done

BTCPayServer/Payments/Bitcoin/NBXplorerListener.cs
```diff
@@ -324,6 +350,31 @@ private async Task<int> FindPaymentViaPolling(BTCPayWallet wallet, BTCPayNetwork
             return totalPayment;
         }
 
+        private decimal GetPayJoinContributedAmount(BitcoinLikeOnChainPaymentMethod paymentMethod, decimal amount, uint256 transactionHash)
+        {
+            if (paymentMethod.PayJoin.Enabled &&
```
NicolasDorier: must be `paymentMethod.PayJoin?.Enabled is true` is we don't get NRE on old payment methods.

BTCPayServer.Tests/docker-compose.yml
```diff
@@ -318,6 +319,19 @@ services:
       - "bitcoin_datadir:/deps/.bitcoin"
     links:
       - bitcoind
+  tor:
```
NicolasDorier: remove that for now

BTCPayServer/Controllers/WalletsController.PSBT.cs
```diff
@@ -153,6 +156,70 @@ private async Task<PSBT> UpdatePSBT(DerivationSchemeSettings derivationSchemeSet
...
+                            if (!input.TryFinalizeInput(out _))
+                            {
+                                valid = false;
+                                break;
+                            }
+                            // a new signed input was provided
+                            valid = true;
+                        }
+
+                        if (!valid || existingInputs.Any())
```
NicolasDorier: the logic is incorrect, if the last input is valid but not the first, valid will be true when it should be false.

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -675,9 +683,17 @@ private IActionResult ViewVault(WalletId walletId, PSBT psbt)
...
-        public IActionResult SubmitVault([ModelBinder(typeof(WalletIdModelBinder))]
+        public async Task<IActionResult> SubmitVault([ModelBinder(typeof(WalletIdModelBinder))]
             WalletId walletId, WalletSendVaultModel model)
         {
+            var network = NetworkProvider.GetNetwork<BTCPayNetwork>(walletId.CryptoCode);
+            var newPSBT = await TryGetBPProposedTX(PSBT.Parse(model.PSBT, network.NBitcoinNetwork), GetDerivationSchemeSettings(walletId), network);
+            if (newPSBT != null)
+            {
+                model.PSBT = newPSBT.ToBase64();
+                return View("WalletSendVault", model);
```
NicolasDorier: use `nameof`

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -747,9 +763,16 @@ private ViewResult ViewWalletSendLedger(WalletId walletId, PSBT psbt, BitcoinAdd
       
         [HttpPost]
         [Route("{walletId}/ledger")]
-        public IActionResult SubmitLedger([ModelBinder(typeof(WalletIdModelBinder))]
+        public async Task<IActionResult> SubmitLedger([ModelBinder(typeof(WalletIdModelBinder))]
             WalletId walletId, WalletSendLedgerModel model)
         {
+            var network = NetworkProvider.GetNetwork<BTCPayNetwork>(walletId.CryptoCode);
+            var newPSBT = await TryGetBPProposedTX(PSBT.Parse(model.PSBT,network.NBitcoinNetwork ), GetDerivationSchemeSettings(walletId), network);
```
NicolasDorier: Please don't update this. The ledger is depreciated.

NicolasDorier:
Please remove the TempData for bpu. There is weird side effect to it that can happen, which are not a bg deal when you show a success message, but in this particular case, the user may end up paying the wrong merchant. Instead, add payjoin to each viewmodel.

Also, you should check the PSBT after having signed, and if invalid, send the Tx1:  (you can with NBX) imagine malicious merchant
1. The merchant get Tx1, then returns an invalid PSBT,
2. You sign,
3. Broadcast fails.
4. Now merchant claims you did not paid.
5. You pay again
6. He broadcast tx1


but if it is not (say the merchant fucked up something) then the first PSBT should be sent instead

else, if there is a malicious merchant, he might claim that you did not paid the good. You may believe him because BTCPayServer show you that it failed

then you pay again, but then he sends the first tx

BTCPayServer/Controllers/WalletsController.PSBT.cs
```diff
@@ -153,6 +156,70 @@ private async Task<PSBT> UpdatePSBT(DerivationSchemeSettings derivationSchemeSet
...
+                
+                var bpuresponse = await httpClient.PostAsync(bpu.ToString(), new StringContent(cloned.ToHex(), Encoding.UTF8, "text/plain"));
+                if (bpuresponse.IsSuccessStatusCode)
+                {
+                    var hex = await bpuresponse.Content.ReadAsStringAsync();
+                    if (PSBT.TryParse(hex, btcPayNetwork.NBitcoinNetwork, out var newPSBT))
+                    {
+                        //check that all the inputs we provided are still there and that there is at least one new(signed) input.
+                        bool valid = false;
+                        var existingInputs = psbt.Inputs.Select(input => input.PrevOut).ToList();
```
NicolasDorier: `.ToHashSet` is better here

BTCPayServer/Controllers/WalletsController.PSBT.cs
Kukks: the loops `breaks` when an invalid input is found.

Kukks:
> 
> 
> Please remove the TempData for bpu. There is weird side effect to it that can happen, which are not a big deal when you show a success message, but in this particular case, the user may end up paying the wrong merchant. Instead, add payjoin to each viewmodel.
> 
> Also, you should check the PSBT after having signed, and if invalid, send the Tx1: (you can with NBX) imagine malicious merchant
> 
>     1. The merchant get Tx1, then returns an invalid PSBT,
> 
>     2. You sign,
> 
>     3. Broadcast fails.
> 
>     4. Now merchant claims you did not paid.
> 
>     5. You pay again
> 
>     6. He broadcast tx1
> 
> 
> Note that between 5 and 6 he can wait a month or so so you forget about it.

Done,
I replaced the bpu temp data with a property in each view model.
I've also kept state of the original transaction, so that if the payjoin tx is not broadcastable, the original tx is broadcasted instead.

BTCPayServer/Controllers/WalletsController.PSBT.cs
```diff
@@ -153,18 +157,83 @@ private async Task<PSBT> UpdatePSBT(DerivationSchemeSettings derivationSchemeSet
...
+        
         [HttpGet]
         [Route("{walletId}/psbt/ready")]
         public async Task<IActionResult> WalletPSBTReady(
             [ModelBinder(typeof(WalletIdModelBinder))]
             WalletId walletId, string psbt = null, 
             string signingKey = null,
-            string signingKeyPath = null)
+            string signingKeyPath = null,
+            string originalPsbt = null)
```
NicolasDorier: you can't pass PSBT in query string, if the transaction is too big there is many obscure error at proxy level happening that give lot's of support request. I did that before, eradicated all of them.

NicolasDorier:
1. You can't pass PSBT by query strings, there is side effect kicking in if the transaction get too big
2. The new option in the PSBT ready will be confusing for users.

The easiest way to fix this is to check that the transaction is valid in the ViewVault etc...

BTCPayServer/Controllers/WalletsController.PSBT.cs
Kukks: This is the same as the psbt parameter just above it. Its not actually used for the query string and is just called from another method

Kukks:
> 
> 
>     1. You can't pass PSBT by query strings, there is side effect kicking in if the transaction get too big
> 
>     2. The new options in the PSBT ready will be confusing for users. Don't add more buttons.
> 
> 
> The easiest way to fix this is to check that the transaction is valid in the ViewVault etc...
What if I rename the new option?

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+            PSBT psbt = null;
+            if (!Transaction.TryParse(rawBody, network.NBitcoinNetwork, out var transaction) &&
+                !PSBT.TryParse(rawBody, network.NBitcoinNetwork, out psbt))
+            {
+                return UnprocessableEntity("invalid raw transaction or psbt");
+            }
+
+            if (psbt != null)
+            {
+                transaction = psbt.ExtractTransaction();
```
NicolasDorier: This can try, you should use TryExtract and sent an error if does not work.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+            }
+
+            if (transaction.Check() != TransactionCheckResult.Success)
+            {
+                return UnprocessableEntity($"invalid tx: {transaction.Check()}");
+            }
+
+            if (transaction.Inputs.Any(txin => txin.ScriptSig == null || txin.WitScript == null))
```
NicolasDorier: useless check as NBX will tell us if valid or not. Also I don't think this is even possible to extract a tx from psbt with null scriptsig.

Kukks: the transaction can come from the body and not from psbt

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+            //UNLESS the request specified an invoice Id, which is mandatory :) 
+            var matchingInvoice = await _invoiceRepository.GetInvoice(invoice);
+            if (matchingInvoice == null)
+            {
+                return UnprocessableEntity($"invalid invoice");
+            }
+            
+            var invoicePaymentMethod = matchingInvoice.GetPaymentMethod(paymentMethodId);
+            //get outs to our current invoice address
+            var currentPaymentMethodDetails =
```
NicolasDorier: should be a normal cast

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+                invoicePaymentMethod.GetPaymentMethodDetails() as BitcoinLikeOnChainPaymentMethod;
+
+            if (!currentPaymentMethodDetails.PayJoin.Enabled)
```
NicolasDorier: PayJoin can be null `PayJoin?.Enabled is true

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+            _btcPayWalletProvider = btcPayWalletProvider;
+            _payJoinStateProvider = payJoinStateProvider;
+        }
+
+        [HttpPost("{invoice}")]
+        [IgnoreAntiforgeryToken]
+        [EnableCors(CorsPolicies.All)]
+        [MediaTypeConstraint("text/plain")]
+        [RateLimitsFilter(ZoneLimits.PayJoin, Scope = RateLimitsScope.RemoteAddress)]
+        public async Task<IActionResult> Submit(string cryptoCode, string invoice)
```
NicolasDorier: the invoice should not be needed in the URL, as we can know which invoice from the outputs.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -0,0 +1,418 @@
...
+            if (!mempool.Success)
+            {
+                return UnprocessableEntity($"provided transaction isn't mempool eligible {mempool.RPCCodeMessage}");
+            }
+
+            var paymentMethodId = new PaymentMethodId(network.CryptoCode, PaymentTypes.BTCLike);
+
+            //multiple outs could mean a payment being done to multiple invoices to multiple stores in one payjoin tx which makes life unbearable
+            //UNLESS the request specified an invoice Id, which is mandatory :) 
+            var matchingInvoice = await _invoiceRepository.GetInvoice(invoice);
```
NicolasDorier: matching invoice should be retrieved from the transaction's output itself.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
Kukks: There is no `TryExtract`

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
Kukks: Basically if the wallet paying the payjoin tx tries to pay 2 invoices on the same btcpay store with the same tx, you would need to do a bunch of logic to determine which takes priroity, etc

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
Kukks: https://github.com/btcpayserver/btcpayserver/pull/1321#discussion_r399166164

Kukks:
Note from talking with @NicolasDorier 
Consider switching to only PSBT to handle fee validations on the receiver side better

Kukks:
Note: @NicolasDorier 
Detect signature grinding and apply accordingly

Kukks:
Client Todo:
* move all payjoin logic to psbtready viewmodel
* pass the signing method to the psbtready model
* Reverse the option from "skip payjoin" to "Pay with payjoin"

PR #1322, opened 2020-02-05
BIP21 Support for Wallet spending

Kukks:


BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -462,6 +462,50 @@ private async Task<bool> CanUseHotWallet()
             if (network == null || network.ReadonlyWallet)
                 return NotFound();
             vm.SupportRBF = network.SupportRBF;
+            
+            if (!string.IsNullOrEmpty(bip21))
```
rockstardev: I would recommend this extracted into separate `SendBip21` method. https://martinfowler.com/bliki/FlagArgument.html

Kukks: done

NicolasDorier:
seems good. Can you add a test in CanManageWallet?

Kukks:
@NicolasDorier  Added test

PR #1323, opened 2020-02-05
Fix PSBT Redirect No-access issues

Kukks:
There were issues with  redirection after the 3.1 bump. This fixes and simplifies it


NicolasDorier:
I did this hack because there was no other way. You probably broke something else, let me remember.

BTCPayServer/Controllers/WalletsController.PSBT.cs
```diff
@@ -325,7 +325,7 @@ private async Task FetchTransactionDetails(DerivationSchemeSettings derivationSc
             }
             else if (command == "analyze-psbt")
             {
-                return RedirectToWalletPSBT(walletId, psbt);
+                return await WalletPSBT(walletId, new WalletPSBTViewModel() {PSBT = psbt.ToBase64()});
```
NicolasDorier: Not right, if you do this then you will show `WalletPSBT` on `{walletId}/psbt/ready` instead of `{walletId}/psbt`

BTCPayServer/Controllers/WalletsController.PSBT.cs
```diff
@@ -359,7 +359,7 @@ private IActionResult FilePSBT(PSBT psbt, string fileName)
             }
             sourcePSBT = sourcePSBT.Combine(psbt);
             TempData[WellKnownTempData.SuccessMessage] = "PSBT Successfully combined!";
-            return RedirectToWalletPSBT(walletId, sourcePSBT);
+            return await WalletPSBT(walletId, new WalletPSBTViewModel() {PSBT = sourcePSBT.ToBase64()});
```
NicolasDorier: Same problem as above

NicolasDorier:
Do not rewrite this, it took me a while to get it right.
When we export the PSBT, we should be redirected to `{walletId}/psbt`, not just showing the view on the current URL.

What is the exact issue?

BTCPayServer/Controllers/WalletsController.cs
```diff
@@ -603,22 +603,6 @@ private IActionResult ViewVault(WalletId walletId, PSBT psbt)
             });
         }
 
-        private IActionResult RedirectToWalletPSBT(WalletId walletId, PSBT psbt, string fileName = null)
```
NicolasDorier: Maybe what can work is instead use `RedirectToAction` instead of the PostRedirect page.

Though I am surprised I did not think about it and did this hack instead.

Kukks: updated the code. I used `RedirectToAction` but also `TempData` to transmit it between the redirect.

BTCPayServer/Controllers/WalletsController.PSBT.cs
Kukks: fixed

BTCPayServer/Controllers/WalletsController.PSBT.cs
Kukks: fixed

Kukks:
> 
> 
> Do not rewrite this, it took me a while to get it right.
> When we export the PSBT, we should be redirected to `{walletId}/psbt`, not just showing the view on the current URL.
> 
> What is the exact issue?

The issue is that we get another "Not authroized" redirection issue with the PostRedirect action because the url loses some route data

NicolasDorier:
Fixed by https://github.com/btcpayserver/btcpayserver/commit/ae33b1d0a876c46c2e2de6cce114a6311d519546

PR #1328, opened 2020-02-09
Minor readme cleanup + license clarification

pavlenex:
This PR provides a very tiny cleanup of the readme.md and adds a single paragraph that hopefully clarifies license (especially logos and graphics).


Kukks:
the branch you based this off is super old

pavlenex:
I don't think that should matter since the readme file is intact since then?

Kukks:
> 
> 
> I don't think that should matter since the readme file is intact since then?
Just makes the tests fail :)

pavlenex:
Ah okay, lemme do a new PR, saw a thing or two that I'd like to fix anyways.

Kukks:
nah its fine

PR #1331, opened 2020-02-12
Wallet prep work for BPU

Kukks:
This PR prepares the wallet for #1321. It makes transfers from the vault and ledger to go to their own post actions for processing (not particularly useful in this PR but is needed in BPU to propose a new tx)  It also makes the Sign with seed consistent with redirect to /psbt/ready after signing which it did not do (it stayed on the seed route)

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -564,6 +564,7 @@ namespace BTCPayServer.Tests
                 Assert.Contains(jack.ToString(), s.Driver.PageSource);
                 Assert.Contains("0.01000000", s.Driver.PageSource);
                 s.Driver.FindElement(By.CssSelector("button[value=analyze-psbt]")).ForceClick();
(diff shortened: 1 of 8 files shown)


BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -566,7 +566,7 @@ void SetTransactionOutput(int index, BitcoinAddress dest, decimal amount, bool s
                 s.Driver.FindElement(By.CssSelector("button[value=analyze-psbt]")).ForceClick();
                 s.Driver.FindElement(By.CssSelector("#OtherActions")).ForceClick();
                 s.Driver.FindElement(By.CssSelector("button[value=broadcast]")).ForceClick();
-                Assert.EndsWith("psbt", s.Driver.Url);
```
NicolasDorier: can you add this assert after the click on `analyze-psbt`? This is what I wanted to do actually.

Kukks: done

Kukks:
Done @NicolasDorier

Issue #1302, opened 2020-01-23
Expired invoice remains as new


mondoir:
Sometimes when an invoice expires, its status wont get updated as Expired.
This happens when there are more than 20 invoices generated despite their status.
According to the attached image, it shows the timer reaches zero but the address still showing.
This particular invoice expired 40 minutes prior to the time of this report.
There were no IPN call after the invoice was created according to the picture bellow.

<img width="363" alt="Screenshot 2020-01-23 at 21 40 35" src="https://user-images.githubusercontent.com/56967158/73026594-9d307000-3e29-11ea-8a4c-7bf8699dc507.png">
<img width="877" alt="Screenshot 2020-01-23 at 21 45 45" src="https://user-images.githubusercontent.com/56967158/73026683-c5b86a00-3e29-11ea-8502-4dbe0b147bb9.png">
<img width="384" alt="Screenshot 2020-01-23 at 21 45 56" src="https://user-images.githubusercontent.com/56967158/73026689-c8b35a80-3e29-11ea-9506-a3f37fd634d7.png">

pavlenex:
@mondoir Can you please follow the template next time when you're opening an issue, it helps us troubleshoot faster.

So to replicate this: 
Create 20 invoices in BTCPay one after another?
Which version of BTCPay are you running?
What's your deployment method?

mondoir:
@pavlenex sorry mate. My bad.

version: BTCPayServer v1.0.3.153
Deployment: LunaNode

This is for today.
I am not certain that generating 20 invoices one after another would certainly recreate this, but when there are many open invoices, at least one is not getting expired properly as others.

<img width="1113" alt="Screenshot 2020-01-25 at 10 37 32" src="https://user-images.githubusercontent.com/56967158/73119789-d3bbd700-3f5e-11ea-98bb-95910a0ad637.png">

pavlenex:
No worries, thanks for providing the info. Someone already reported this a while ago but we were never able to replicate it. Knowing that it probably happens when there are a lot of invoices may be a progress.

pavlenex:
I've tried this on v1.0.3.155 and can confirm it's a bug.

Link to expired invoice https://pay.pavlenex.com/i/UrRjBb9JcV5jJMMfrFfmig

```

Date | Message
-- | --
1/27/2020 12:56:05 PM | Creation of invoice starting
1/27/2020 12:56:05 PM | BTC_USD: The rating rule is coinaverage(BTC_USD)
[...]

PR #1345, opened 2020-02-20
Adding logging to try and catch situations where invoice is not expired

rockstardev:
Debugging #1302 and adding log entry to check if this is the reason why some invoices are not expired occasionally.

### DIFF
--- a/BTCPayServer/HostedServices/InvoiceWatcher.cs
+++ b/BTCPayServer/HostedServices/InvoiceWatcher.cs
@@ -180,7 +180,11 @@ namespace BTCPayServer.HostedServices
         {
             if (invoiceId == null)
                 throw new ArgumentNullException(nameof(invoiceId));
-            _WatchRequests.Writer.TryWrite(invoiceId);
+            
+            if (!_WatchRequests.Writer.TryWrite(invoiceId))
+            {
+                Logs.PayServer.LogWarning($"Failed to write invoice {invoiceId} into WatchRequests channel");
+            }
         }
 
         private async Task Wait(string invoiceId)
--- a/BTCPayServer/HostedServices/InvoiceWatcher.cs
+++ b/BTCPayServer/HostedServices/InvoiceWatcher.cs
@@ -54,7 +54,7 @@ namespace BTCPayServer.HostedServices
         private async Task UpdateInvoice(UpdateInvoiceContext context)
         {
             var invoice = context.Invoice;
-            if (invoice.Status == InvoiceStatus.New && invoice.ExpirationTime < DateTimeOffset.UtcNow)
+            if (invoice.Status == InvoiceStatus.New && invoice.ExpirationTime <= DateTimeOffset.UtcNow)
             {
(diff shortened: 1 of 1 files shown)


NicolasDorier:
You want that to be a warning rather than information though right?

rockstardev:
@NicolasDorier sure, if you think warning is better.

rockstardev:
OK, we got it... @pavlenex tested on his server and problem is resolved. This seems to have been tricky DateTime compare problem. 

https://github.com/btcpayserver/btcpayserver/pull/1345/commits/2a8464c57c6d04429ad75dd871c511c358fe9eab#diff-c60857d4c415ad8e4b397a08462027c7R196

This would also depend on database provider (does it round DateTime to seconds)... but occasionally that `var delay = invoice.ExpirationTime - DateTimeOffset.UtcNow;` would produce delay that would trigger invoice monitoring milliseconds before invoice expires. Thus logic in `InvoiceWatcher.Watch` would keep invoice as new and only update it to expired by the time next `Watch` check was triggered (MonitoringExpiration).

NicolasDorier:
I will probably do a new release only for this one.

pavlenex:
@NicolasDorier two minor PR's can probably be snugged in in the new release as well:
#1349 #1346 but if you prefer not to that's fine, I just think they are easy enough to review and merge if everything is fine so we can continue to decrease the PR backlog.

NicolasDorier:
@rockstardev even reading your comment and your PR I can't figure out why this is happening.
The call to `Watch` should make an update to the invoice, even if this one is not monitored anymore.

rockstardev:
@NicolasDorier it won't make an update to the invoice because that `Watch` call will trigger milliseconds before invoice expires.

Eventually invoice will absolutely be expired (see this comment by @pavlenex https://github.com/btcpayserver/btcpayserver/issues/1302#issuecomment-578941169). But it'll stay new for approx 4 hours by default.

Here is that piece of code with more comments:

```
// set delay to default 15 minutes so that if there are no other events 
// we'll hit `Watch(invoiceId)` exactly when invoice expires
var delay = invoice.ExpirationTime - DateTimeOffset.UtcNow;
if (delay > TimeSpan.Zero)
{
    // OK, we got in here... we are chilling for 15 minutes
    await Task.Delay(delay, _Cts.Token);
}

// but alas Task.Delay finished 100 milliseconds before invoice expired
Watch(invoiceId);
// we are out of Watch and invoice status hasn't changed because 
// invoice we had 100 milliseconds left before date compare check declared
// it expired

// Oh no, now we need to wait for 4 hours for another status update
delay = invoice.MonitoringExpiration - DateTimeOffset.UtcNow;
if (delay > TimeSpan.Zero)
{
    // chill for 4 hours
    await Task.Delay(delay, _Cts.Token);
}
// here we will finally expire invoice and stop monitoring
Watch(invoiceId);

// see https://github.com/btcpayserver/btcpayserver/issues/1302#issuecomment-578941169
```

PR #1349, opened 2020-02-24
add GET endpoint for pay button

### LINKED ISSUE #889: Extend Payment Button to support GET link
I want to send link in email in a form like: http://my.btcpay.server/api/v1/invoices?storeId=xxx&price=10&currency=USD&mail=customer@example.com

I would expect the same behavior like after click to payment button on a website, but I just want to be able to use it also in generated email where I just fill values with price and customer e-mail (so I will see who paid it).

I know I can create Payment Request but I will have to create it for every customer and I see no API how to do it easilly in docs.

Kukks:
closes #889

### DIFF
--- a/BTCPayServer/Controllers/PublicController.cs
+++ b/BTCPayServer/Controllers/PublicController.cs
@@ -26,6 +26,15 @@ namespace BTCPayServer.Controllers
         private InvoiceController _InvoiceController;
         private StoreRepository _StoreRepository;
 
+        [HttpGet]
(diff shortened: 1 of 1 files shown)


NicolasDorier:
not super fan of it, as this is not documented, nobody will use it other than the person in the issue...

iBobik:
> not super fan of it, as this is not documented, nobody will use it other than the person in the issue...

True, I never used it because have not found it in docs after a while.

PR #1341, opened 2020-02-18
 Add Swagger and Redoc

Kukks:
Blocked by #1262

PR #1351, opened 2020-02-24
 GreenField API #1: Get current API Key info 

Kukks:
~~blocked by #1341~~

Adds an endpoint to check info on the currently being used API Key 
 `GET` `api/v1/api-keys/current`  or  `api/v1/users/me/api-keys/current`


Kukks:
I am curious if this should be at `api/v1/api-keys/current` or at `api/v1/users/me/api-keys/current`

NicolasDorier:
@Kukks I think for next endpoint, you should create a client like the `ExplorerClient` of NBXplorer.
All the logic you put in the tests to make a request can be instead reused in a properly typed client.

PR #1356, opened 2020-02-28
CreateInvoiceRequest: remove redundant fields

toxeus:
Remove fields that are already part of the `Buyer` object.

I'm not sure what the goal regarding the `Buyer` object of commit 63c88be5336ecde4e22dac561ec9276e4377616a was. Keeping it maintains compatibility with the BitPay API and is more structured than the inlined fields that have been removed.

### DIFF
--- a/BTCPayServer/Models/CreateInvoiceRequest.cs
+++ b/BTCPayServer/Models/CreateInvoiceRequest.cs
@@ -14,22 +14,6 @@ namespace BTCPayServer.Models
     {
         [JsonProperty(PropertyName = "buyer")]
         public Buyer Buyer { get; set; }
-        [JsonProperty(PropertyName = "buyerEmail", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string BuyerEmail { get; set; }
-        [JsonProperty(PropertyName = "buyerCountry", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string BuyerCountry { get; set; }
-        [JsonProperty(PropertyName = "buyerZip", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string BuyerZip { get; set; }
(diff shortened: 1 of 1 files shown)


toxeus:
I will fix the rest of the code once I get an ACK that the intention of this PR is sound.

NicolasDorier:
this is a breaking change from the API of Bitpay, so I won't do it.

toxeus:
@NicolasDorier no, this is not a breaking change from the [API of Bitpay](https://bitpay.com/api/#rest-api-resources-invoices). These fields don't exist in Bitpay's API documentation of the invoice request.

rockstardev:
Does anybody look at tests passing at all now when submitting PR?

rockstardev:
PR doesn't build, reopen when PR builds and tests pass.

toxeus:
@rockstardev please read my [second comment](https://github.com/btcpayserver/btcpayserver/pull/1356#issuecomment-592333156) before being patronizing about passing tests. I will not invest time in fixing tests before I have an ACK about the intention of this PR. Currently, @NicolasDorier and me are trying to find out what the Bitpay API really is. I don't know if it changed recently but BTCPay deviates from it.

NicolasDorier:
Sorry @toxeus just saw your reply I think bitpay broke their own API, as before it was clearly not this shape.

I think I will still not changing to not break current clients of BTCPayServer.
Note that with the new GreenField API we develop, we will have a clean API free from those kind of smell later.

toxeus:
@NicolasDorier thanks for clarifying. It might be helpful to update the docs and remove lines such as
> Full compatibility with BitPay API (easy migration)

At least it would have been less confusing for me :wink:

NicolasDorier:
@toxeus yeah good idea.
That said, the new API have proper documentation and even swagger compatibility so it will be simpler.
For example on the demo server:

https://mainnet.demo.btcpayserver.org/docs/index.html?url=/swagger/v1/swagger.json

Still a work in progress though.

PR #1358, opened 2020-02-28
0 Amount invoices handling

Kukks:
Let 0 amount invoices pass through(they currently get stuck on new status) and allow email to be set when required but invoice has been paid( makes sense for 0 invoices, but also for when someone displays the btc address or invoice elsewhere from the api)

Specifically requested by nodl guys but it is super useful when you combine it with the POS and want to offer free tickets/ items.
@ketominer @Askuwheteau @IAskuwheteau

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -888,7 +888,11 @@ namespace BTCPayServer.Services.Invoices
             accounting.Due = Money.Max(accounting.TotalDue - accounting.Paid, Money.Zero);
             accounting.DueUncapped = accounting.TotalDue - accounting.Paid;
             accounting.NetworkFee = accounting.TotalDue - totalDueNoNetworkCost;
-            var minimumTotalDueSatoshi = Math.Max(1.0m, accounting.TotalDue.Satoshi * (1.0m - ((decimal)ParentEntity.PaymentTolerance / 100.0m)));
+            // If the total due is 0, there is no payment tolerance to calculate
+            var minimumTotalDueSatoshi = accounting.TotalDue.Satoshi == 0
+                ? 0
+                : Math.Max(1.0m,
+                    accounting.TotalDue.Satoshi * (1.0m - ((decimal)ParentEntity.PaymentTolerance / 100.0m)));
             accounting.MinimumTotalDue = Money.Satoshis(minimumTotalDueSatoshi);
             return accounting;
         }
--- a/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
@@ -155,7 +155,7 @@
 </div>
 
 <div class="payment-box" v-bind:class="{ 'payment-box--with-recommended-fee': showRecommendedFee && !showEmailForm }">
-    <div class="bp-view payment manual-flow enter-contact-email" id="emailAddressView" v-bind:class="{ 'active': showEmailForm}">
+    <div class="bp-view enter-contact-email" id="emailAddressView" v-bind:class="{ 'active': showEmailForm}">
         <form class="manual__step-one refund-address-form contact-email-form" id="emailAddressForm" name="emailAddressForm" novalidate="" v-on:submit.prevent="onEmailSubmit">
             <div class="manual__step-one__header">
                 <span>{{$t("Contact and Refund Email")}}</span>
@@ -190,7 +190,7 @@
         </component>
     </div>
 
-    <div class="bp-view" id="paid" v-bind:class="{ 'active': invoicePaid}">
+    <div class="bp-view" id="paid" v-bind:class="{ 'active': invoicePaid && !showEmailForm}">
         <div class="status-block">
             <div class="success-block">
                 <div class="status-icon">
--- a/BTCPayServer/Views/Invoice/Checkout.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout.cshtml
@@ -184,7 +184,7 @@
                 return !this.showEmailForm && !this.invoiceUnpayable && !this.invoicePaid;
             },
             showEmailForm: function(){
-                return this.srvModel.requiresRefundEmail && (!this.srvModel.customerEmail || !this.validateEmail(this.srvModel.customerEmail)) && !this.invoiceUnpayable && !this.invoicePaid;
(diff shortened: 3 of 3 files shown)


ketominer:
:+1:

pavlenex:
A reasonable use-case, concept ACK.

pavlenex:
Ok, so I've tested this PR and it works as expected. It makes sense for the use-case.

However, what's a bit scary  is that anybody can spam merchant with 0 invoice.

For example, people can spam the crowdfund app to get the sound and it would work out of the box. I'm thinking if this is something that would require some sort of permission prior to being allowed, but I'm unsure where we can put such option.

Kukks:
> 
> 
> Ok, so I've tested this PR and it works as expected. It makes sense for the use-case. Created invoice goes to paid state. Merchant can enable e-mail request get email and be able to deliver a free item such as ticket.
> 
> However, what's a bit scary is that anybody can spam merchant with 0 invoice. Imagine that you sell items via POS or accept donations, someone with a lot of time can make your invoice management way harder by showing a bunch of paid invoices in your dashboard and if you depend on BTCPay invoices to fulfill orders it can be tough.
> 
> For example, people can spam the crowdfund app to get the sound and it would work out of the box. I'm thinking if this is something that would require some sort of permission prior to being allowed, but I'm unsure where we can put such option.
> 
> Not sure how you guys feel about it? It's quite reasonable use-case but I feel like it can break plenty of stuff it not handled properly.

I dont think it would spam anything since no payments happen so no notifications occur.

pavlenex:
@Kukks What notifications are you referring to?

My example is a charity fulfilling orders  and having a pay button, then seeing 100 completed invoices that someone decided to have fun with?

Kukks:
> 
> 
> @Kukks What notifications are you referring to?
> 
> My example is a charity fulfilling orders and having a pay button, then seeing 100 completed invoices that someone decided to have fun with?

You would still get spammed by 100 pending( and stuck) invoices.

pavlenex:
True, but when a merchant wants to fulfill orders they'd filter by paid/completed. 

What do you mean stuck? They would expire otherwise?

Should we have a different label for the `Free` invoices?

I don't know I just see an annoying attack vector that can be just that `annoying`.

Kukks:
I call upon you demons (@ketominer @Askuwheteau ) to argue for this PR to be merged within this realm

NicolasDorier:
I wait for their feedback before merging this?

Kukks:
> 
> 
> I wait for their feedback before merging this?

I would merge but  I'm biased :P

pavlenex:
If I remember @rockstardev gave a NACK on this one.

I also think it can mess things and needs to be properly reviewed and figured out if really needed.

rockstardev:
Yeah, I gave NACK... but I guess no point in fighting it since as @Kukks says we already can create bunch of 0 invoices... they'll just not be properly processed.

This hole will only be properly closed once we have pre & post hooks for invoices... that could redirect user to custom flow of merchant. But as we can't have that before Bitcoin 2020, I guess merge and live with the pain.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -888,7 +888,11 @@ public PaymentMethodAccounting Calculate(Func<PaymentEntity, bool> paymentPredic
             accounting.Due = Money.Max(accounting.TotalDue - accounting.Paid, Money.Zero);
             accounting.DueUncapped = accounting.TotalDue - accounting.Paid;
             accounting.NetworkFee = accounting.TotalDue - totalDueNoNetworkCost;
-            var minimumTotalDueSatoshi = Math.Max(1.0m, accounting.TotalDue.Satoshi * (1.0m - ((decimal)ParentEntity.PaymentTolerance / 100.0m)));
+            // If the total due is 0, there is no payment tolerance to calculate
+            var minimumTotalDueSatoshi = accounting.TotalDue.Satoshi == 0
+                ? 0
+                : Math.Max(1.0m,
+                    accounting.TotalDue.Satoshi * (1.0m - ((decimal)ParentEntity.PaymentTolerance / 100.0m)));
```
NicolasDorier: ahaha

britttttk:
I assume the following cases would be the user accidentally creating a 0 invoice:

1. Payment Requests: Improvement (ux) if 0 invoice is created, since it will go to paid and return to the payment request (instead of user waiting for expiry or exiting the invoice)
2. Crowdfund: Weird to create 0 invoice and make it paid instantly and add to contributor count. If Expired status was used, it would not increase the contributor count.

I assume the following case could be purposeful user 0 invoice creation:
1. POS: 
- Custom Input Items: 0 Invoice is handled the same as before.
- Items that Cost 0 Amount: I believe this is why the PR exists. It is not ideal IMO as @pavlenex mentioned to have Paid invoices for 0 payment. It would mainly be annoying. Although I'm not a big fan of introducing a new status either... what about using Invalid or Expired? Since this person creating an Email Only Event can look at invoices by Store (and they could create a store for this without a real xpub..) so that people looking at only Paid Invoices don't have to encounter this issue.  
- Note: I don't think this can work if Redirect to Url on paid invoice is enabled... since it will redirect before they can provide an email. 

I would say NACK but only for PAID status reason above, else ACK.

Kukks:
> 
> 
> I assume the following cases would be the user accidentally creating a 0 invoice:
> 
>     1. Payment Requests: Improvement (ux) if 0 invoice is created, since it will go to paid and return to the payment request (instead of user waiting for expiry or exiting the invoice)
> 
>     2. Crowdfund: Weird to create 0 invoice and make it paid instantly and add to contributor count. If Expired status was used, it would not increase the contributor count.
> 
> 
> I assume the following case could be purposeful user 0 invoice creation:
> 
>     1. POS:
> 
> 
>     * Custom Input Items: 0 Invoice is handled the same as before.
> 
>     * Items that Cost 0 Amount: I believe this is why the PR exists. It is not ideal IMO as @pavlenex mentioned to have Paid invoices for 0 payment. It would mainly be annoying. Although I'm not a big fan of introducing a new status either... what about using Invalid or Expired? Since this person creating an Email Only Event can look at invoices by Store (and they could create a store for this without a real xpub..) so that people looking at only Paid Invoices don't have to encounter this issue.
> 
>     * Note: I don't think this can work if Redirect to Url on paid invoice is enabled... since it will redirect before they can provide an email.
> 
> 
> I would say NACK but only for PAID status reason above, else ACK. Unless @Kukks your comment
> 
> > but also for when someone displays the btc address or invoice elsewhere from the api
> 
> can't use a non-paid status.

Would https://github.com/btcpayserver/btcpayserver/pull/1397 resolve your concerns?

britttttk:
>Would #1397 resolve your concerns?
yes @Kukks

PR #1359, opened 2020-02-28
Use proper divisibility for payments in crowdfund and do not show too…

### LINKED ISSUE #1003: [Crowdfund app] Show raised amount tooltip only if content in the tooltip is different from what's below it
**Describe the problem/bug**
As mentioned [here](https://github.com/btcpayserver/btcpayserver/issues/926#issuecomment-524235018) we should show the raised amount tooltip only if the content in the tooltip is different from what you see right below it.

63580381-5cdc7580-c59d-11e9-861d-4a5a03679e09

So, what you see above is 3 BTC raised and 3 BTC again in the tooltip. This shouldn't happen. The tooltip should only show up if the tooltip content has something additional to show.

**Your environment**
* Version of BTCPay Server: v1.0.3.127

**To Reproduce**
1. Create a crowdfunding app
2. Raise some on-chain BTC 
3. Hover over the raised amount
4. See how the tooltip is exactly the same content as what's below it

**Expected behavior**
Don't show the tooltip if it doesn't provide any additional information

**Actual behavior**
Shows the tooltip even if the info in it is exactly the same as what's below it

### LINKED ISSUE #1037: [Crowdfund app] Raised amount tooltip shows wrong amount if app is raising in fiat
**Describe the problem/bug**
If your crowdfund app specifies amount to raise in fiat then your raised amounts tooltip shows amount rounded to two decimal points instead of eight.

**Your environment**
* Version of BTCPay Server: v1.0.3.131 

**To Reproduce**
1. Create a crowdfund app
2. Specify an amount to raise in USD or any other fiat option
3. Contribute some amount in crypto with more than two decimal points in amount (something like 0.1192345)
4. See that the amount raised will be shown as 0.12 in the tooltip

The issue is even worse if you are contributing a smaller amount like 0.001245. You will get 0.00 in the tooltip.

**Expected behavior**
Tooltip always shows contributed amounts in correct divisibility amounts.

**Screenshots/Links**
See screenshot below:
Screen Shot 2019-09-15 at 12 23 15 PM

**Additional context**
This issue was introduced by me earlier in a [PR](https://github.com/btcpayserver/btcpayserver/pull/987) that fixed issue #926.

The issue was that I assumed `srvModel.currencyData.divisibility` always equals divisibility number for crypto amounts (so 8) but it's equal to 2 if you are raising in fiat. I guess the raised amount number might have to be formatted server-side instead.

Kukks:
…ltip if identical data

fixes #1037 and fixes #1003

### DIFF
--- a/BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
@@ -75,6 +75,7 @@ namespace BTCPayServer.Models.AppViewModels
         public bool DisplayPerksRanking { get; set; }
         public bool Enabled { get; set; }
         public string ResetEvery { get; set; }
+        public Dictionary<string, CurrencyData> CurrencyDataPayments { get; set; }
     }
 
     public class ContributeToCrowdfund
--- a/BTCPayServer/Services/Apps/AppService.cs
+++ b/BTCPayServer/Services/Apps/AppService.cs
@@ -153,6 +153,10 @@ namespace BTCPayServer.Services.Apps
                 Sounds = settings.Sounds,
                 AnimationColors = settings.AnimationColors,
                 CurrencyData = _Currencies.GetCurrencyData(settings.TargetCurrency, true),
+                CurrencyDataPayments = currentPayments.Select(pair => pair.Key)
+                    .Concat(pendingPayments.Select(pair => pair.Key)).Distinct()
+                    .Select(id => _Currencies.GetCurrencyData(id.CryptoCode, true))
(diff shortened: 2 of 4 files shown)


pavlenex:
Ok, so I've tested this PR and it works as expected. It makes sense for the use-case.

However, what's a bit scary  is that anybody can spam merchant with 0 invoice.

For example, people can spam the crowdfund app to get the sound and it would work out of the box. I'm thinking if this is something that would require some sort of permission prior to being allowed, but I'm unsure where we can put such option.

bolatovumar:
I haven't tested this but code looks good to me.

PR #1360, opened 2020-02-28
Modal option for Pay Button

### LINKED ISSUE #796: Modal pop-up invoice for payment button
We should give an option to display the payment button invoice in pop-up out of the box, where the user just ticks the checkbox. It makes sense for it to even be checked by default.

- Add modal invoice support to pay button 
- Add checkbox in the UI

Current page redirect is not good UX most of the times.

** Screenshots
I kinda liked the position here, but any other would be okay. Maybe after button size makes even more sense.
New Mockup 1 copy

Kukks:
closes #796

### DIFF
--- a/BTCPayServer/Controllers/PublicController.cs
+++ b/BTCPayServer/Controllers/PublicController.cs
@@ -37,7 +37,6 @@ namespace BTCPayServer.Controllers
         
         [HttpPost]
         [Route("api/v1/invoices")]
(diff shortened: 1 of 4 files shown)


pavlenex:
tACK.

bolatovumar:
Just when I finally have some time to look into implementing this I find out that @Kukks has already done it.

I haven't tested this PR myself but code is looking good.

Issue #1316, opened 2020-02-03
[BUG] 1.0.3.156: We are not showing the next available address in the invoices list anymore


NicolasDorier:
![image](https://user-images.githubusercontent.com/3020646/73663778-90413580-46e1-11ea-83ea-666cd5248f58.png)

NicolasDorier:
Caused by 68484829 I did not anticipated this view was shared. Should consider removing partial view

PR #1361, opened 2020-02-28
Showing the next available address in the invoices list

rockstardev:
Fixes: #1316


BTCPayServer/Views/Invoice/Invoice.cshtml
```diff
@@ -183,7 +183,48 @@
             </div>
         }
 
-        <partial name="InvoicePaymentsPartial" model="Model" />
+        <div class="row">
```
Kukks: Instead of duplicating all this view code for one extra field, cant we just keep the partial and pass a parameter to it?

rockstardev: Not in a clean way... the best I could do was magic ViewBag parameter. If you can figure out a way for it to be like:

`<partial name="InvoicePaymentsPartial" model="Model" magic-param-show-address="true/false" />`

I would do it... otherwise I would keep it as is, since you are conceptually having different things.

Kukks: How about making the model for the partial a tuple `( InvoiceDetailsModel model, ( bool expand, bool otherParam))`

rockstardev: f852020

https://www.youtube.com/watch?v=MZ552T1vzn8

PR #1362, opened 2020-02-29
add enviroment configuration to issue template

pavlenex:
This PR adds `cat /etc/profile.d/btcpay-env.sh` in the bug report issue template that will help us better with deployment issues. I've noticed it's important info we often ask people to provide.

Unsure if wording was the best here so if there are better ideas, let me know.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,10 +14,8 @@ A clear and concise description of what the bug is.
 **Logs (if applicable)**
 Basic logs can be found in Server Settings > Logs.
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
mmmh I don't know if that's the best, because there is some private info here (say a xpub from eps)

pavlenex:
Should I add "Obscure any private information"?

NicolasDorier:
The best is that they run `. btcpay-setup.sh -i` then copy the logs, we show all variable there, but people need to make sure to remove sensitive info like xpub

pavlenex:
@NicolasDorier should be good now I think.

PR #1368, opened 2020-03-02
BTCPayServer.Client library + Revoke API Key

Kukks:


BTCPayServer.Client/BTCPayServer.Client.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
```
NicolasDorier: use netstandard2.1 instead.

Kukks: done

Kukks:
@NicolasDorier changed to netstandard 2.1

PR #1370, opened 2020-03-04
Prettify wallet receive tab screen

bolatovumar:
I saw that we now have a new "Receive" tab for the wallet and decided to prettify it a bit.

Before:
Screen Shot 2020-03-03 at 9 13 55 PM

After:
Screen Shot 2020-03-03 at 8 58 27 PM

Before:
Screen Shot 2020-03-03 at 9 14 21 PM

After:
Screen Shot 2020-03-03 at 8 58 03 PM

Also, one thing that's not clear to me is as someone who has never seen this screen before is this:
1. Is there a reason you need to click on "Generate BTC address" instead of just showing the QR code with the address right away?
2. What does "unreserve this address" button do? Once you click on it you get a success message like: `Address bcrt1qkjr980k3mnswq02mpnn627njuswycul0q9sgj2 was unreserved.`. It's not clear what "unreserve" means in this context and there is no explanation provided. I'm guessing this means that this address will never be used again but it's not entirely clear to me.

I'm anticipating that other users who are not familiar with this screen will probably have the same questions in their head as me when they first use this feature.

### DIFF
--- a/BTCPayServer/Views/Wallets/WalletReceive.cshtml
+++ b/BTCPayServer/Views/Wallets/WalletReceive.cshtml
@@ -14,21 +14,39 @@
     </div>
 }
 <div class="row no-gutters">
-    <div class="col-lg-6  mx-auto my-auto ">
+    <div class="col-lg-7  mx-auto my-auto ">
         <form method="post" asp-action="WalletReceive" class="card text-center">
-            @if (string.IsNullOrEmpty(Model.Address))
-            {
+            <div class="card-body">
+                @if (string.IsNullOrEmpty(Model.Address))
+                {
 
-                <h2 class="card-title">Receive @Model.CryptoCode</h2>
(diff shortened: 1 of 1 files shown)


Kukks:
Is there a reason you need to click on "Generate BTC address" instead of just showing the QR code with the address right away?

Yes, this helps to at least reduce the amount of reserved addresses to prevent gap limit issues.

UnReserving an address means that the next address used by btcpay on that store would be that one. Reserving implies that it won't be shown as an unused address in the next invoice

bolatovumar:
@Kukks @pavlenex ok, it would be great if we could explain what "unreserving" means in text below that link or in a tooltip or something. Otherwise it's really not clear what it does.

I still feel like it would be way nice if a user didn't have to click the "Generate BTC address" button. I understand why it's done this way now and can't really suggest a better solution.

Issue #1330, opened 2020-02-11
Include Replace By Fee (RBF) in transaction resources


haarts:
Bitcoin transactions can be replaced an other transaction spending the same (or subset) UTXOs. This feature is called [Replace By Fee](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki) and is a useful addition to the semantics of a transaction. 
Transactions can explicitly signal they support this feature via the `sequence` field setting this to less than `0xffffffff - 1`. Or do so implicitly when one of their parents explicitly signals RBF. 

The risk is that malicious users can pay with an RBF transaction, get the goods/services from the merchant and the broadcast another transaction spending the same UTXOs to a different address. 
This risk only applies when the merchant doesn't wait for the first confirmation. This happens often (all the time?) when doing face to face transactions.

It would be great if the responses from the API involving transactions would include a field telling the API consumer about RBF signalling. Perhaps a boolean field `isRBFSignalling` would suffice. This way the API consumer can decide what is best in their scenario.

Obviously Lightning is the better payment solution in such a case as we're dealing with (probably) small values and a face to face transaction.

mjlamb:
This should also be include in the IPN notification response, so action can be taken if required by  signalling.

Kukks:
> 
> 
> This should also be include in the IPN notification response, so action can be taken if required by signalling.

The IPN response should never be trusted. You should only ever grab the invoice ID from the IPN and call `get invoice` through the API to be sure no one faked an IPN call to you.
Even then, when using BTCPay, integrating against the `paid` status vs `confirmed` or `completed` is wrong. 

IMO, the only risk I see here is for physical shops, who just display the checkout and wait for payment. Maybe we can add an opt-in option to display a yellow warning on the checkout UI on RBFed txs

shivaenigma:
Another simple way is to lookup the txid with blockexplorer API will return the RBF flag
- [Blockstream](https://blockstream.info/tx/7f48851af8e2dcb35503da7839b6271ea88f557e196ecc06d882dca982056810) - Works if tx itself is RBF opt-in (Not sure about inherited RBF)
- [Blockonomics](https://www.blockonomics.co/views/api.html#txdetail) - Will return data for both explicit and [inherited signalling](https://twitter.com/blockonomics_co/status/1209409221499867141)

haarts:
> Another simple way is to lookup the txid with blockexplorer API that will return the RBF flag

[...]

PR #1375, opened 2020-03-09
Check for Double Spends

### LINKED ISSUE #1371: Call GetMempoolEntry on RBF unconfirmed txs on next block event/ periodically

Kukks:
~~Blocked by https://github.com/dgarage/NBXplorer/pull/247~~
fixes #1371
potentially helps #1330
needed for #1321

Currently, BTCPay can only handle double spends on incoming txs when at least 1 output is still pointing to the store derivation. Of course, the invoice would never get marked as confirmed, but it does stay stuck in the `paid` state. This PR introduces a checker that runs every minute that verifies that any unconfirmed, RBF enabled transaction has not been double spent by calling `getmempoolentry`.

### DIFF
--- a/BTCPayServer.Tests/ServerTester.cs
+++ b/BTCPayServer.Tests/ServerTester.cs
@@ -138,6 +138,19 @@ namespace BTCPayServer.Tests
             await CustomerLightningD.Pay(bolt11);
         }
 
+        public async Task<T> WaitForEvent<T>(Func<Task> action)
+        {
+            var tcs = new TaskCompletionSource<T>();
+            var sub = PayTester.GetService<EventAggregator>().Subscribe<T>(evt =>
+            {
+                tcs.SetResult(evt);
+            });
+            await action.Invoke();
+            var result = await tcs.Task;
+            sub.Dispose();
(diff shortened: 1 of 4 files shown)


NicolasDorier:
> Of course, the invoice would never get marked as confirmed, but it does stay stuck in the paid state

No, it would go into invalid state eventually.

Kukks:
> 
> 
> > Of course, the invoice would never get marked as confirmed, but it does stay stuck in the paid state
> 
> No, it would go into invalid state eventually.

In any case, this is for that period of time before expiration. Physical stores and payjoin require that quicker status update so that a physical knows immediately if it was "reverted" and for payjoin so that it can put exposed utxos into the prioritized queue and broadcast the original tx as spy protection

NicolasDorier:
I am thinking of another approach that does not require the rpc proxy and would also simplify the code:

In the update payment, I would try to broadcast the unconfirmed transaction. Then by analyzing the potential error message, I can flag it as double spend. Not only it simplify the logic a lot, but also prevent to handle a lot of corner cases.

PR #1382, opened 2020-03-11
Simplify RBF handling, and handle case of double spend happening outs…

NicolasDorier:
…ide of wallet (Fix #1375)

Alternative to

1375#issuecomment-597552085


Kukks:
This still only runs on new block or new tx. Can we have a background loop checking the unconfirmed txs every minute?

NicolasDorier:
@Kukks is it really important? I prefer not doing this because there would be concurrencies issue as several thread would want to update payments.

Kukks:
> @Kukks is it really important? I prefer not doing this because there would be concurrency issues as several thread would want to update payments.

I need it for the spy protection in payjoin and to improve the checkout UI for physical stores #1330

PR #1389, opened 2020-03-12
Greenfield API: God Mode

Kukks:
When the `ServerManagement` permission is granted, you should be able to do everything in the system.
Maybe I should rename it to GodMode as a permission to not have any confusion with managing server settings (currently `ServerManagement`)?

### DIFF
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -224,6 +224,12 @@ namespace BTCPayServer.Tests
                         tester.PayTester.HttpClient);
                 });
             }
+            else
+            {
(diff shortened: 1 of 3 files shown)


BTCPayServer.Tests/ApiKeysTests.cs
```diff
@@ -224,11 +225,14 @@ public async Task CanCreateApiKeys()
                 });
             }
 
-            await Assert.ThrowsAnyAsync<HttpRequestException>(async () =>
+            if (!permissions.Contains(Permissions.ServerManagement))
```
NicolasDorier: `else`, test that that Api Access work

Kukks: done

PR #1390, opened 2020-03-13
Greenfield API: Create User

Kukks:
Slightly big PR because I started refactoring to reduce code duplication between the UI based business logic and the api one.

### DIFF
--- a/BTCPayServer/Controllers/RestApi/Users/UsersController.cs
+++ b/BTCPayServer/Controllers/RestApi/Users/UsersController.cs
@@ -1,12 +1,18 @@
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using System.Threading.Tasks;
 using BTCPayServer.Client.Models;
+using BTCPayServer.Configuration;
 using BTCPayServer.Data;
+using BTCPayServer.Events;
 using BTCPayServer.Hosting.OpenApi;
 using BTCPayServer.Security;
+using BTCPayServer.Services;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 using NSwag.Annotations;
 
 namespace BTCPayServer.Controllers.RestApi.Users
@@ -18,14 +24,24 @@ namespace BTCPayServer.Controllers.RestApi.Users
     public class UsersController : ControllerBase
     {
         private readonly UserManager<ApplicationUser> _userManager;
+        private readonly BTCPayServerOptions _btcPayServerOptions;
+        private readonly RoleManager<IdentityRole> _roleManager;
+        private readonly SettingsRepository _settingsRepository;
+        private readonly EventAggregator _eventAggregator;
 
-        public UsersController(UserManager<ApplicationUser> userManager)
+        public UsersController(UserManager<ApplicationUser> userManager, BTCPayServerOptions btcPayServerOptions,
+            RoleManager<IdentityRole> roleManager, SettingsRepository settingsRepository,
+            EventAggregator eventAggregator)
         {
             _userManager = userManager;
+            _btcPayServerOptions = btcPayServerOptions;
+            _roleManager = roleManager;
+            _settingsRepository = settingsRepository;
+            _eventAggregator = eventAggregator;
         }
-    
+
         [OpenApiOperation("Get current user information", "View information about the current user")]
-        [SwaggerResponse(StatusCodes.Status200OK, typeof(ApiKeyData),
+        [SwaggerResponse(StatusCodes.Status200OK, typeof(ApplicationUserData),
             Description = "Information about the current user")]
         [Authorize(Policy = Policies.CanModifyProfile.Key, AuthenticationSchemes = AuthenticationSchemes.ApiKey)]
         [HttpGet("~/api/v1/users/me")]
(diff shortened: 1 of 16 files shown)


BTCPayServer/Controllers/RestApi/Users/UsersController.cs
```diff
@@ -34,14 +50,82 @@ public async Task<ActionResult<ApplicationUserData>> GetCurrentUser()
...
+            Description = "A list of validation errors that occurred")]
+        [SwaggerResponse(StatusCodes.Status400BadRequest, typeof(ValidationProblemDetails),
+            Description = "A list of errors that occurred when creating the user")]
+        [Authorize(Policy = Policies.CanModifyServerSettings.Key, AuthenticationSchemes = AuthenticationSchemes.ApiKey)]
+        [HttpPost("~/api/v1/users")]
+        public async Task<ActionResult<ApplicationUserData>> CreateUser(CreateApplicationUserRequest request)
+        {
+            var policies = await _settingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
+            var anyAdmin = (await _userManager.GetUsersInRoleAsync(Roles.ServerAdmin)).Any();
+            var admin = request.IsAdministrator.GetValueOrDefault(request.IsAdministrator == null && !anyAdmin);
```
NicolasDorier: var admin = (request.IsAdministrator is true && isAdmin) || !anyAdmin;

NicolasDorier: actually if `request.IsAdministrator is true && !isAdmin` should return be HTTP forbidden.

BTCPayServer/Controllers/RestApi/Users/UsersController.cs
```diff
@@ -34,14 +50,82 @@ public async Task<ActionResult<ApplicationUserData>> GetCurrentUser()
...
-        
+
+        [OpenApiOperation("Create user", "Create a new user")]
+        [SwaggerResponse(StatusCodes.Status201Created, typeof(ApplicationUserData),
+            Description = "Information about the new user")]
+        [SwaggerResponse(StatusCodes.Status422UnprocessableEntity, typeof(ValidationProblemDetails),
+            Description = "A list of validation errors that occurred")]
+        [SwaggerResponse(StatusCodes.Status400BadRequest, typeof(ValidationProblemDetails),
+            Description = "A list of errors that occurred when creating the user")]
+        [Authorize(Policy = Policies.CanModifyServerSettings.Key, AuthenticationSchemes = AuthenticationSchemes.ApiKey)]
```
NicolasDorier: Should allow anonymous

BTCPayServer/Hosting/Startup.cs
```diff
@@ -72,6 +72,16 @@ public void ConfigureServices(IServiceCollection services)
                 //    ScriptSrc = "'self' 'unsafe-inline'"
                 //});
             })
+            .ConfigureApiBehaviorOptions(options =>
+            {
+                var builtInFactory = options.InvalidModelStateResponseFactory;
+
+                options.InvalidModelStateResponseFactory = context =>
```
NicolasDorier: It would be great if we don't have to depend to ModelState. You can't have both, complex type serialization (say a BitcoinAddress) and model state validation. It is one or the other.

BTCPayServer/Controllers/RestApi/Users/UsersController.cs
Kukks: No, there is a special exception at `APIKeyAuthorizationHandler.cs` for this. I don't want to bypass the whole auth and permission system just for that one case

NicolasDorier: I think your approach is more risky. 
Because you are applying a rule globally, at the `APIKeyAuthorizationHandler` level, while adding `AllowAnonymous` here will make sure this apply ONLY to this action.

BTCPayServer/Hosting/Startup.cs
Kukks: We don't fully depend on it. If there is soemthing that is too complicated to handle through the simple validations, we can do it in our action like we currently do in the rest of btcpay uis

BTCPayServer/Controllers/RestApi/Users/UsersController.cs
NicolasDorier: Also, creating users should be allow if the "Disable registration" is not checked. And that throttling should be checked if the user is not admin.

Kukks: If we bypass the auth system, we can't control permissions through policies and api key permissions and will make it a headache to maintain though. 
For example, I wanted to add later on a way to generate api keys that would only allow you to create users. If we bypass the whole auth system and hotwire it in the action method,  we lose the ability to do that easily.

Kukks: With the way it is now:
* throttling is not required as only 1 call will work to this endpoint if you are unauthorized
* Disable registration should not effect this endpoint as only a server admin that can call this endpoint( Server admins can always create users even when "Disable Registration" is enabled, through the Server Settings->Users admin ui)
* The rule is explicit to the policy `Policies.CanCreateUser`

Kukks: IMHO I think the special rule is better handled in the same place where we handle all other authorizations since it is all grouped together and although slightly hacky ( that is the nature of this weird special rule anyway), it is still very readable

NicolasDorier: You can control permissions through policies and API Keys by using the AuthorizatinService (I cna take a look).

NicolasDorier:
Can you fix the conflicts? After I will try to implement with the AuthorizationManager or something like that.

BTCPayServer/Controllers/RestApi/Users/UsersController.cs
```diff
@@ -34,14 +50,82 @@ public async Task<ActionResult<ApplicationUserData>> GetCurrentUser()
...
+        }
+    }
+
+    [ModelMetadataType(typeof(CreateApplicationUserRequestMetadata))]
+    public class CreateApplicationUserRequest : BTCPayServer.Client.Models.CreateApplicationUserRequest
+    {
+        
+    }
+
+    public class CreateApplicationUserRequestMetadata
```
NicolasDorier: What is the reason with doing such hack instead of doing it simply in the controller? Is it so this is documented in the generated doc?

Kukks: Yeah It is for docs

BTCPayServer/Controllers/RestApi/Users/UsersController.cs
Kukks: Go for it

PR #1394, opened 2020-03-14
WIP: Add Tor Http Client Factory

Kukks:
This PR adds a Tor Http Client Factory so that we can do HTTP requests over socks5( which we usually use tor's socks5).
Using it for #1321 for sending when possible. Would be great to also use for rates and other external calls if possible.

 Currently have an issue where a 503 is always returned.  Must be something wrong with my tor config or the socks proxy lib Im using.

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -35,6 +35,7 @@
     <PackageReference Include="BundlerMinifier.Core" Version="3.2.435" />
     <PackageReference Include="BundlerMinifier.TagHelpers" Version="3.2.435" />
     <PackageReference Include="HtmlSanitizer" Version="4.0.217" />
+    <PackageReference Include="HttpToSocks5Proxy" Version="1.4.0" />
     <PackageReference Include="LedgerWallet" Version="2.0.0.5" />
     <PackageReference Include="Microsoft.Extensions.Logging.Filter" Version="1.1.2" />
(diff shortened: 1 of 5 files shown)


NicolasDorier:
The default IHttpClientFactory of dotnet core is a very complicated piece of code that we can't reimplement by ourselves. Instead we should manage to use their extensibility point to support Tor in the default factory.

BTCPayServer/BTCPayServer.csproj
```diff
@@ -35,6 +35,7 @@
     <PackageReference Include="BundlerMinifier.Core" Version="3.2.435" />
     <PackageReference Include="BundlerMinifier.TagHelpers" Version="3.2.435" />
     <PackageReference Include="HtmlSanitizer" Version="4.0.217" />
+    <PackageReference Include="HttpToSocks5Proxy" Version="1.4.0" />
```
NicolasDorier: No external deps, we have several piece of code to use

https://github.com/NTumbleBit/NTumbleBit/blob/master/NTumbleBit/Tor/SocksMessageHandler.cs

Or better, just use NBitcoin https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/Socks/SocksHelper.cs

NicolasDorier:
I searched a bit, and I don't think we have access to underlying socket via the default IHttpClientFactory, so we will probably need to do our own like you do now, but we should limit it only to Payjoin, I don't want to destabilize the rest.

I remember seeing somewhere that in .net5, we will have better access to internal.

Kukks:
Let's revisit this later

PR #1395, opened 2020-03-15
Allow Pay Button to work on Apps

### LINKED ISSUE #1392: Pay with BTCPay button can have payment amount changed using inspect element
**Describe the problem/bug**
Users can modify the html of the button to change how much they pay

**Your environment**
Just the embedded BTCPay button in HTML

example: https://www.youtube.com/watch?v=VrQBkPkfIk8

A work around is to explicitly set the value of the price input to the actual amount?

Kukks:
This PR allows you to use the pay button generator to create buttons that target apps. This means that you can generate an invoice that is linked to an item on the POS/Crowdfund (targeting the item is optional). The POS/Crowdfund item amount -> invoice creation amount validation works too so that the user cannot modify the amount of a perk using just html ( fixes #1392 )

### DIFF
--- a/BTCPayServer.Data/Data/AppData.cs
+++ b/BTCPayServer.Data/Data/AppData.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
--- a/BTCPayServer/Controllers/StoresController.cs
+++ b/BTCPayServer/Controllers/StoresController.cs
@@ -17,6 +17,7 @@ using BTCPayServer.Rating;
 using BTCPayServer.Security;
 using BTCPayServer.Security.Bitpay;
 using BTCPayServer.Services;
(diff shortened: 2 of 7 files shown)


NicolasDorier:
@MiLeung can you test it?

MiLeung:
I'm not sure how to. I'm using a third-party hosted BTCPay server. Also I think I will go the route of creating an invoice via the api and using the orderId. That would be the most secure way to ensure the user pays the correct amount, right?

NicolasDorier:
@MiLeung can you try? https://www.youtube.com/watch?v=GWR_CcMsEV0 

It works on mac, but it should work for any windows and linux.

MiLeung:
I only have a mac - you have to find someone else who has windows/linux

NicolasDorier:
hey @pavlenex if you test this one we can go back under 80 issues!

Kukks:
If it helps @pavlenex, this is basically a noob friendly way to this section in the apps:
image

PR #1398, opened 2020-03-18
Adding loggin in CanManageWallet test, fixing UserControllerTests

rockstardev:
For CanManageWallet: If all logs are present it's possibly issue with TempData... merge this to master so we can debug in future builds.

Also verify if my refactoring of `CreateUser` method aligns with what you wanted to do with recent changes.


NicolasDorier:
@rockstardev can you just run the tests for this PR several time? I fear we forget to remove that later.

PR #1399, opened 2020-03-19
New feature: Coin Selection

Kukks:
This opt-in feature allows you to select which utxos you want to use for a specific transaction.

https://i.imgur.com/7Xl9lyt.gifv

Has a very minimal no javascript version too


NicolasDorier:
Excellent. There is no test, will write one.

PR #1402, opened 2020-03-19
Fix liquid asset BIP21 decimal precision

Kukks:
The way liquid assets decimal precision works is just an ui layer. Each unit of an asset is actually 1sat.
Fixes https://github.com/Blockstream/green_android/issues/86

### DIFF
--- a/BTCPayServer.Common/Altcoins/Liquid/ElementsLikeBtcPayNetwork.cs
+++ b/BTCPayServer.Common/Altcoins/Liquid/ElementsLikeBtcPayNetwork.cs
@@ -25,7 +25,7 @@ namespace BTCPayServer
 
         public override string GenerateBIP21(string cryptoInfoAddress, Money cryptoInfoDue)
         {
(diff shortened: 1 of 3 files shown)


alessandro-saglimbeni:
> Each unit of an asset is actually 1sat.

Actually units and decimals of an asset are supposed to be determined based on the precision specified by the issuer on the [registry](assets.blockstream.info/).

The decimal precision is applied to satoshi units.

Few examples from the registry:
- if a merchant asks 10.00000000 USDt they're asking `amount=10.00000000` because `precision=8` for `assetID=ce091c998b83c78bb71a632313ba3760f1763d9cfcffae02258ffa9865a37bd2`

- if a merchant asks 10 SCAM they're asking `amount=0.00000010` because `precision=0` for `assetID=123465c803ae336c62180e52d94ee80d80828db54df9bedbb9860060f49de2eb`

- if a merchant asks 10.00 IPA they're asking `amount=0.00001000` because `precision=2` for `assetID=beebee1a548fbb20280e539b697de076d87859a25c2983ebc55f2d8bec40abc3`

PR #1403, opened 2020-03-20
Greenfield API: Basic Auth

Kukks:


BTCPayServer/Controllers/RestApi/TestApiKeyController.cs
```diff
@@ -15,7 +14,7 @@ namespace BTCPayServer.Controllers.RestApi
     /// </summary>
     [Route("api/test/apikey")]
     [ApiController]
-    [Authorize(AuthenticationSchemes = AuthenticationSchemes.ApiKey)]
+    [Authorize(AuthenticationSchemes = AuthenticationSchemes.ApiKeyOrBasic)]
```
NicolasDorier: Can you just call it "Greenfield" instead of "ApiKeyOrBasic" ? That's the name we use in the doc as well.

NicolasDorier:
I would have thought of a different, easier approach: Renaming the `ApiKey*Handler` to `GreenField*Handler` and just check for basic in the `GreenFieldAuthenticationHandler`

BTCPayServer/Security/AuthenticationSchemes.cs
```diff
@@ -5,5 +5,7 @@ public class AuthenticationSchemes
         public const string Cookie = "Identity.Application";
         public const string Bitpay = "Bitpay";
         public const string ApiKey = "GreenfieldApiKey";
+        public const string Basic = "Basic";
+        public const string ApiKeyOrBasic = ApiKey + "," + Basic;
```
NicolasDorier: I think you should remove the 3 of them and jsut have GreenField.

NicolasDorier:
Such way of doing does not seems to be standardizable via swagger, but I don't think this is a problem.

BTCPayServer/Security/AuthenticationSchemes.cs
Kukks: If I do this I cannot register individual auth handlers(unique key required for each handler)

NicolasDorier: Do we need individual auth handelrs?

Kukks:
@NicolasDorier  I consolidated the auth into one

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -102,6 +102,13 @@ public async Task CheckSwaggerIsConformToSchema()
             var schema = JSchema.Parse(await resp.Content.ReadAsStringAsync());
             IList<ValidationError> errors;
             bool valid = swagger.IsValid(schema, out errors);
+            //the schema is not fully compliant to the spec. We ARE allowed to have multiple security schemas.
```
NicolasDorier: I think if we are not compliant, we should just fallback to comment into the "Greenfield authentication" in plain language instead.

Kukks: we are compliant, it is just the schema json that is wrong. 
```
As a reminder, the JSON Schema is not the source of truth for the Specification. In cases of conflicts between the Specification itself and the JSON Schema, the Specification wins. Also, some Specification constraints cannot be represented with the JSON Schema so it's highly recommended to employ other methods to ensure compliance.
```

Kukks: from https://github.com/OAI/OpenAPI-Specification/tree/master/schemas/v3.0

PR #1411, opened 2020-03-24
GreenField: Create/Get/Delete Stores

Kukks:


NicolasDorier:
I think a really cool feature in the API would be a "Clone" so that someone can automatically configure a template store and duplicate it for several customer. Maybe another PR though

Kukks:
yeah another PR, I need to add update store to this too anyway

BTCPayServer/Controllers/GreenField/StoresController.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+            await _storeRepository.RemoveStore(storeId, _userManager.GetUserId(User));
+            return Ok();
+        }
+        
+        [HttpPost("~/api/v1/stores")]
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        public async Task<ActionResult<Client.Models.StoreData>> CreateStore(CreateStoreRequest request)
+        {
+            if (request is null)
```
NicolasDorier, line 70: `request?.Name is null`

BTCPayServer/Controllers/GreenField/StoresController.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        private readonly StoreRepository _storeRepository;
+        private readonly UserManager<ApplicationUser> _userManager;
+
+        public StoresController(StoreRepository storeRepository, UserManager<ApplicationUser> userManager)
+        {
+            _storeRepository = storeRepository;
+            _userManager = userManager;
+        }
+        [Authorize(Policy = Policies.CanViewStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
```
NicolasDorier, line 27: `CanViewStoreSettings` is a scoped permission. Should use a separate permission.

PR #1415, opened 2020-03-26
Scan address/bip21 with camera

Kukks:
More compact icons in wallet
firefox_2kae3TbmVm

camera interaction
https://i.imgur.com/i8pjow0.gifv


BTCPayServer/wwwroot/main/site.css
```diff
@@ -11,9 +11,6 @@ html {
 .input-group-clear {
     display: none;
 }
-.only-for-js {
```
Kukks: this is actually an unrelated bug fix

pavlenex:
Lovely PR. Much needed function and now the UI seems a bit cleaner than before with BIP21. I think this is definitely a good direction. I've tested this PR and my suggestions are only (as usual) UI-related.

## UI Improvements

Current UI is as following:

Screenshot 2020-03-26 at 18 29 17


### Re-position buttons
I think that current `+` Icon, especially because of the size of the `Sign with...` button may look as a `Send` CTA most mobile wallet users are used to. 
The two main CTA in wallet sections are `Sign with` and `add another destination`. Since people are used to them already, I suggest we leave the `add another destination` and increase size of the `Sign with...` while we are at it I suggest we make that button finally more clearer and add `Sign a transaction with`. The other two icons fit in great and are mostly clear what they are.

Screenshot 2020-03-26 at 18 21 08

### Permission denied warring 
Screenshot 2020-03-26 at 18 26 12

Currently when you deny the permission the warring looks too informal. I think it's good that we formalize it and explain why this permission is needed. I am not sure about the exact text, but something like: `A permission to the camera is needed to scan the QR code. Fill in data manually if you don't want to grant a permission`
 
### Scanned info gets cut off
Self-explanatory, the address & amount gets cut off and looks weird when scanned
Screenshot 2020-03-26 at 18 34 02

BTCPayServer/wwwroot/main/site.css
pavlenex: When you reach that level of goodness to review your own PR 🤣

Kukks:
@pavlenex  I don't think the text should be "Sign a transaction", becuase you are creating the transaction as you click that button and not selecting a new one. 

Updated everything else.

pavlenex:
@Kukks Awesome. One thing that doesn't work is `Retry` button when you deny the access.
Screenshot 2020-03-27 at 10 34 07

pavlenex:
Also when using multiple destinations, the camera option should be disabled or give a proper error message instead of:

Screenshot 2020-03-27 at 10 36 58

Kukks:
> 
> 
> Also when using multiple destinations, the camera option should be disabled or give a proper error message instead of:
> 
> Screenshot 2020-03-27 at 10 36 58

Yes, this is because the (specific) browser blocks future attempts( usually temporraily for the session) until you unblock it.

Kukks:
> 
> 
> @Kukks Awesome. One thing that doesn't work is `Retry` button when you deny the access.
> Screenshot 2020-03-27 at 10 34 07

Yes, this is because the (specific) browser blocks future attempts( usually temporraily for the session) until you unblock it.

pavlenex:
Yeah it's a bit clunky but I understand.

tACK.

NicolasDorier:
@Kukks can you fix the conflict?

PR #1430, opened 2020-04-01
GreenField: Payment Requests CRUD

Kukks:


NicolasDorier:
conflicts

Kukks:
fixed conflicts

NicolasDorier:
I think it should have its own permission where CanModifyStore include the permission.

Check method `ContainsPolicy` to do this.

NicolasDorier:
You did not changed `Permission.ContainsPolicy` to make modify implies view.

Kukks:
Doing a separate PR for that

Kukks:
Ah I just saw the method and its already there. Will do it now

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
```diff
@@ -0,0 +1,370 @@
...
+                    },
+                    "description": {
+                        "type": "string",
+                        "description": "The description of the payment request",
+                        "nullable": true,
+                        "format": "html"
+                    },
+                    "expiryDate": {
+                        "type": "string",
+                        "description": "The expiry date of the payment request",
```
NicolasDorier, line 339: which format?

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
```diff
@@ -0,0 +1,370 @@
...
+                    }
+                }
+            },
+            "PaymentRequestBaseData":{
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "amount": {
+                        "type": "number",
+                        "format": "decimal",
```
NicolasDorier, line 312: I don't think this is a good idea to specify amounts in decimal. The JS decimal lose a lot of precision. What are best practice in json for that?

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
Kukks: the .net default : `"2019-08-01T00:00:00-07:00"`

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
Kukks: json in itself is agnostic of the shit JS does with regards to numbers. Let me check a bit more on this though

Kukks: Would .NET be able to parse a json string to decimal type correctly?

NicolasDorier: I don't think json is agnostic to JS shit... however we use decimal already for create invoice. So I will check.

Also I remember seeing that satoshi decided to make 8 decimal because of some shit on floats.

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
Kukks: swagger format for `date-time` is also specific to one type which is also the `Newtonsoft.Json` default date format 
```csharp
    public enum DateFormatHandling
    {
        /// <summary>
        /// Dates are written in the ISO 8601 format, e.g. <c>"2012-03-21T05:40Z"</c>.
        /// </summary>
        IsoDateFormat,
```

BTCPayServer/wwwroot/swagger/v1/swagger.template.payment-requests.json
Kukks: I have added a converter for `decimal` to always be converted to string. It can read a decimal or a string type though

BTCPayServer.Client/JsonConverters/DecimalStringJsonConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+                default:
+                    throw new JsonSerializationException("Unexpected token type: " +
+                                                         token.Type);
+            }
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            if (value != null)
+                writer.WriteValue(value.ToString());
```
NicolasDorier, line 37: should case to decimal and use CultureInfo.Invariant

Kukks: done

BTCPayServer.Client/Models/PaymentRequestBaseData.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using BTCPayServer.Client.JsonConverters;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Client.Models
+{
+    public class PaymentRequestBaseData
+    {
+        [JsonProperty(ItemConverterType = typeof(PermissionJsonConverter))]
```
NicolasDorier, line 9: Should be `DecimalStringJsonConverter`

BTCPayServer.Client/JsonConverters/DecimalStringJsonConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue,
+            JsonSerializer serializer)
+        {
+            JToken token = JToken.Load(reader);
+            switch (token.Type)
+            {
+                case JTokenType.Float:
+                case JTokenType.Integer:
+                    return token.ToObject<decimal>();
```
NicolasDorier, line 23: There is some gotchas around this with newtonsoft. Like if it is an integer, it can't convert to decimal. Or sometimes it can to double, but not decimal or shit like that... got the case one :(

NicolasDorier: You should do a unit test specifically on this jsonconverter I think if you want to support that.

Kukks:
@NicolasDorier  added test

PR #1431, opened 2020-04-01
Swagger Generator: Merge multiple documents

Kukks:
It's becoming very hard to edit the swagger file as it grows (especially with multiple PRs altering it). This PR allows the swagger file to be generated from multiple jsons instead which are merged in the controller.


NicolasDorier:
Can you remove the CheckSwaggerIsConformToSchema  from Fast tests and instead use an integration test with the generated swagger file?

Kukks:
> 
> 
> Can you remove the CheckSwaggerIsConformToSchema from Fast tests and instead use an integration test with the generated swagger file?

@NicolasDorier  done

PR #1435, opened 2020-04-04
Dark checkout theme

rockstardev:
I've left the most colors from default scheme... tried to match what's currently on: https://btcpayserver.flat18.co.uk/

Here is picture of new dark checkout:

Dark checkout

And here is how it looks in action with all themes we have now:

dark invoice design


NicolasDorier:
holy shit my heart can't take it anymore!

NicolasDorier:
Ah, I think the QR code without some white padding will not work correctly.

NicolasDorier:
Yep, tested. Scanning the one in the dark theme is quite difficult. We should have some padding.
I think the QR lib we are using allow us to set it up though.

rockstardev:
@NicolasDorier try now.

Here are images if it's easier

image

Lightning

image

rockstardev:
> holy shit my heart can't take it anymore!

https://i.imgur.com/t9bScLe.gif

Kukks:
Ack, a thousand times ack

pavlenex:
Great work, I have a few comments which are mostly questions, since I'm curious about some decisions you've made.

1. The `Legacy | Dark` - Is there a reason why you didn't use drop-down here? I noticed that people often skip those links (for example lots of people find that "Click here" link on LN confusing. Also it would kinda look consistent with all drop-downs there.
drop-down I see how it works with custom link, so maybe it makes sense, this is more of a question than remark.

2. I'm not a fan of the black color used here, imo the color-scheme could match a bit better.
Also the background and the header are of same color so the header blends in.
Capture

Here's an example from the demo site:
Capture2

Suggestion: use darker color instead of black and change the background color, basically use the same scheme as on demo site, but if there are practical reasons why you've gone with these, let me know, perhaps I'm missing something.

3. I really like the green color used as a link color in `Powered by BTCPay Server` should we do the same for the Light Theme which has blue?

rockstardev:
@pavlenex:

1. We can't have dropdown there since that field is custom value that on legacy installations... pretty inline with your observation.

2. Black fix: 1a8f222e46fab2a2d35d863284e5a80dd46ad46c. 

For page background, problem is the smoothness of gradient. This is well known issue with CSS (https://stackoverflow.com/questions/13151331/smooth-css-gradients). If you want to see how it looks with gradient, checkout 1a8f222e46fab2a2d35d863284e5a80dd46ad46c.

If you like gradient, I can just put it back... and we can look into how to additionally smooth it in a way that will consistently work across devices. Maybe @dennisreimann can help.

3. f085a5618b16ca6c2fda137d2773446e4d18719d

NicolasDorier:
fuck this is so good.

pavlenex:
@rockstardev Just tested, looks great now!

rockstardev:
@pavlenex I'll hit you up on chat.btcpayserver.org to talk about gradients and other elements tomorrow; we'll keep improving after release.

PR #1434, opened 2020-04-04
New default theme

dennisreimann:
Based on the discussion in the [accompanying issue](/btcpayserver/btcpayserver/issues/1429). Closes #1429.

Helpful for reviewing: Use the [kitchensink page](http://127.0.0.1:14142/_bootstrap_kitchensink.html?theme=default) to see all design permutations – especially regarding the light/dark theme.

PR #1440, opened 2020-04-06
Fix transaction dropdowns

dennisreimann:
As brought up by @kukks with [this comment](https://github.com/btcpayserver/btcpayserver/pull/1434#issuecomment-609720019). The problem exists across themes, so the fix is separated from the originating PR.


dennisreimann:
The test failure `BTCPayServer.Tests.PayJoinTests.CanUseBIP79Client` does not seem to be related to the changes in this PR as there are also pipeline runs that fail with the same error on master.

PR #1443, opened 2020-04-07
Wallet view improvements

dennisreimann:
As brought up by @kukks with [this comment](https://github.com/btcpayserver/btcpayserver/pull/1434#issuecomment-609724703). These improvements can be applied across themes, so these fixes are separated from the originating PR.
Pure visual finetunings for the Coin Selection form and amount/fee rate links.

wallet

### DIFF
--- a/BTCPayServer/Views/Wallets/CoinSelection.cshtml
+++ b/BTCPayServer/Views/Wallets/CoinSelection.cshtml
@@ -23,22 +23,22 @@
         <li class="list-group-item d-flex justify-content-between align-items-center cursor-pointer"
             v-for="item of filteredItems"
             :key="item.outpoint"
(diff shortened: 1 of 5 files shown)


Kukks:
Since the tx id takes up more space, it's a bit harder to click to select now..  
It looks better but is harder to use. Any suggestions? I was thinking a left/right aligned button or checkbox?

dennisreimann:
You mean an explicit select button? Let's use left aligned checkboxes then, ok?

Kukks:
yeah that should work

dennisreimann:
Added the checkbox for selecting the UTXO – the whole line is clickable and toggles the selection state.

coin-selection-with-checkbox

Also prefixed the amount in the non-js version, because it wasn't visible before as the outpoint took up the whole space.
 
non-js-coin-selection

PR #1445, opened 2020-04-08
Add Tor Http Client Factory

Kukks:
This PR adds a Tor Http Client Factory so that we can do HTTP requests over socks5( which we usually use tor's socks5).


NicolasDorier:
Can you rebase this one? the tests should pass

PR #1447, opened 2020-04-08
New default theme

### LINKED ISSUE #1429: Default theme coloring re-design
After redesign of the invoice re-design and login page (#1428) it would be good if we can make colors consistent in our default theme, without doing anything fancy but color changes, to avoid breaking changes for users using default theme.

We can either leave classic theme as a legacy and keep new one as default, that's something that we should discuss.

Unfortunately, we do not have mockups available for it yet, but it should be done like @dennisreimann did for the #1137 Casa theme. 

Design guidelines are available [here](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY?node-id=0%3A1), and should be consistent with what we have [here](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY?node-id=25%3A0). 

If time allows of course a dark theme on top would be great, if not we can do that in the future.

dennisreimann:
Yikes, I @pavlenex'ed my forked repo 🙈 and cannot reset the original PR #1434. I'll close it and leave this for further discussions and the final merge. Closes #1429.

All of @Kukks comments in the other PR have been addressed and we can continue with new feedback here in case someone finds additional things that require fixing.

Sorry about the hassle!


NicolasDorier:
Ping @pavlenex @Kukks you let me know when I merge?

NicolasDorier:
Ping @Kukks the tests on greenfield seems to not pass for this PR, weird nothing here should impact it.

Kukks:
@dennisreimann do you think the `alert-link` should just be an implicit style in a stylesheet instead?
like `.alert a, .alert code`

dennisreimann:
@Kukks I had the same thoughts regarding merging and hence being able to spot potential issues earlier than by more review here. In case we find something I'll fix it with a new finetuning PR.

The alert-links: We'd either have to modify the bootstrap file for that or duplicate those styles into our custom extensions. I'd prefer to just append the `alert-link` class in HTML where necessary as this is how Bootstrap suggests to do it.

Kukks:
fixed the tests hopefully. Nothing big

PR #1449, opened 2020-04-09
Add payment link in checkout with payjoin indicator

Kukks:
This shows the BIP21 in the Copy tab in the invoice checkout. If the link is payjoin enabled, it shows the `user-secret` icon. Else, it shows the coin image.
image

### DIFF
--- a/BTCPayServer/Views/Shared/Bitcoin_Lightning_LikeMethodCheckout.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin_Lightning_LikeMethodCheckout.cshtml
@@ -1,3 +1,4 @@
+@using BTCPayServer.Services
 @model BTCPayServer.Models.InvoicingModels.PaymentModel
 
 <script type="text/x-template" id="bitcoin-lightning-method-checkout-template">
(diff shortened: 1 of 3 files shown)


NicolasDorier:
seems good for me, I am unsure about the CSS change, if you need to change the css, it means this will not render correctly on other themes? @dennisreimann

NicolasDorier:
Also I would say payment link should be below address.

dennisreimann:
@NicolasDorier The checkout CSS is decoupled from the backend/admin CSS, that's also why @rockstardev and I could work in parallel on the theme updates.

Rest looks good to me and I like the proposal of moving the link below the address.

Kukks:
moved it to below address

NicolasDorier:
Ping @rockstardev do you see any problem with the change in CSS? I fear that it break on existing checkout themes.

Kukks:
@NicolasDorier  just tried it

dennisreimann:
@NicolasDorier I don't think it will be a problem, as the classes in the HTML are already used elsewhere and hence present in the CSS.

PR #1457, opened 2020-04-12
Add mnemonic phrase color CSS variable

### LINKED ISSUE #1456: Mnemonic seed not visible when generating a hot wallet
**Describe the problem/bug**

**EDIT: This happens on every theme not just dark theme as I originally reported.**

When someone uses dark theme and tries to generate a hot wallet, the mnemonic is not that visible and it's hard for user to even notice there's a seed there, which can be quite dangerous.

**To Reproduce**
1. Go to Store > Derivation Scheme > Modify
2. Click on `Import from` ...`new/existing seed`
3. Enable hot wallet 
4. Generate a wallet.
5. See error

**Expected behavior**
User should clearly see, possibly emphasized in danger color his seed and should be aware that they need to back it up.

**Actual behavior**\
The seed is barely visible

**Screenshots/Links**
seed

mnemonic

Happens on light theme as well

light

bolatovumar:
fix #1456

### DIFF
--- a/BTCPayServer/Controllers/StoresController.BTCLike.cs
+++ b/BTCPayServer/Controllers/StoresController.BTCLike.cs
@@ -380,7 +380,7 @@ namespace BTCPayServer.Controllers
                 TempData.SetStatusMessageModel(new StatusMessageModel()
                 {
                     Severity = StatusMessageModel.StatusSeverity.Success,
-                    Html = $"Your wallet has been generated. Please store your seed securely! <br/><code>{response.Mnemonic}</code>"
+                    Html = $"Your wallet has been generated. Please store your seed securely! <br/><code class="alert-link">{response.Mnemonic}</code>"
                 });
             }
             else


dennisreimann:
I‘d rather apply the `alert-link` class to the `code` tag (like we did with the API Key) than introduce a new variable for this single use case. The `alert-link` class gives it the correct text color. What do you think @kukks?

Kukks:
Let's stay consistent with the rest of the codebase

bolatovumar:
@dennisreimann @Kukks ok, done.

dennisreimann:
@bolatovumar thanks! 👌🏻

@Kukks do we have the same problem in the tests as with the API key?

Kukks:
Most likely, will fix

Kukks:
fixed conflict and test i think

NicolasDorier:
@Kukks that's still failing.

NicolasDorier:
will fix tests

Issue #1458, opened 2020-04-12
Mobile menu icon is not shown on light theme


bolatovumar:
**Describe the problem/bug**
Mobile menu icon is not shown on light theme. Also, notice how the onion icon is shifted weirdly to the right.

**Your environment**
* Version of BTCPay Server: v1.0.3.164 

**Screenshots/Links**
![Screen Shot 2020-04-12 at 10 45 24 AM](https://user-images.githubusercontent.com/1934678/79076768-8cb5c600-7cb1-11ea-9578-971e06046147.png)

Kukks:
fixed by #1466

PR #1460, opened 2020-04-12
Make sure navbar looks good on smaller screen sizes

bolatovumar:
Addresses #1458


Kukks:
I'll leave this one to @dennisreimann  to review, I'm not sure if the hard-coded light/dark check is good since we can have other darker themes.

Kukks:
I am thinking this needs to be fixed in their respective theme files where:
 `*-dark` styles are light on dark themes
 `*-light` styles are dark on dark themes
vice versa for light themes

dennisreimann:
Yes, as @Kukks said I think we should refrain from using the light/dark-classes offered by Bootstrap. Instead we should use something more intention revealing.

I'll take a closer look later today.

Kukks:
superseded by #1466

PR #1465, opened 2020-04-13
add additional server policy for hot wallet RPC import

Kukks:
So that if you enable hot wallets for all, you can still not allow them to import to your RPC


NicolasDorier:
I don't really like the hack with the ViewData

Kukks:
Yeah me neither, I will fix in a new version

PR #1466, opened 2020-04-13
Navbar fixes

dennisreimann:
Alternative implementation for what @bolatovumar proposed with #1460.

As said in Umar's PR I think it is best to refrain from using the ligfht/dark classes, as they make it hard to deal with inverted themes. We are still using some of these classes, but I'd like to migrate away from these in favor of something more intention revealing.

An alternative approach would be to invert the colors of the dark themes, but there are cases like this one with the nav icon, where Bootstrap uses CSS background images and we'd have to override them in the individual themes. That would make things harder to maintain.

### DIFF
--- a/BTCPayServer/Views/Shared/_Layout.cshtml
+++ b/BTCPayServer/Views/Shared/_Layout.cshtml
@@ -51,11 +51,9 @@
             </a>
             @if (env.OnionUrl != null)
             {
-                <span>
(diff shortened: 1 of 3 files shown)


NicolasDorier:
is this alternative to #1460 ?

bolatovumar:
@NicolasDorier yeah. @dennisreimann has a more thorough fix. My fix was a quick and dirty one.

PR #1471, opened 2020-04-16
GreenField: Add health check endpoint

dennisreimann:
As discussed with @Kukks, this is a simple health check endpoint for the new API.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.Health.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Health.cs
@@ -1,0 +1,15 @@
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BTCPayServer.Client
+{
+    public partial class BTCPayServerClient
+    {
+        public virtual async Task<bool> IsHealthy(CancellationToken token = default)
+        {
+            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health"), token);
+            return response.IsSuccessStatusCode;
+        }
+    }
+}
--- a/BTCPayServer.Client/BTCPayServerClient.Health.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Health.cs
@@ -1,15 +1,16 @@
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using BTCPayServer.Client.Models;
 
 namespace BTCPayServer.Client
 {
     public partial class BTCPayServerClient
     {
-        public virtual async Task<bool> IsHealthy(CancellationToken token = default)
+        public virtual async Task<ApiHealthData> GetHealth(CancellationToken token = default)
         {
-            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health"), token);
-            return response.IsSuccessStatusCode;
+            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health", bodyPayload: new {}), token);
+            return await HandleResponse<ApiHealthData>(response);
         }
     }
 }
--- a/BTCPayServer.Client/BTCPayServerClient.Health.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Health.cs
@@ -9,7 +9,7 @@ namespace BTCPayServer.Client
     {
         public virtual async Task<ApiHealthData> GetHealth(CancellationToken token = default)
         {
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Should it send back OK 200 if it is still not fully synched?

Kukks:
> Should it send back OK 200 if it is still not fully synched?

For this specific one, yes, we can expand it after for when it is authenticated

dennisreimann:
That's actually a good point and imho it'd make sense to distinguish synching and synched. Especially since checkout won't work when not fullt synched.

NicolasDorier:
@Kukks yeah, I was thinking about it, but that's only if we were showing too much information.
I don't think that's a big deal to allow just checking if it is fully synched as anonymous.

dennisreimann:
What about `200` for fully synched, `202` for synch in progress and `503` for unavailable?

The [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) is imho the closest status description there is and even though checkout does not work I think something in the 2xx range is appropriate.

Kukks:
I would stick with empty 200 for anon call
And then add Sync status for authenticated call with no permissions

Kukks:
You cannot use a status code for sync as you could have multiple nodes on
the same instance

NicolasDorier:
@dennisreimann that's strange. I would say, add an parameter instead.
@Kukks this is not for multiple node, but just the `IsFullySynced` of the NBXplorerDashboard (saying if everything is completely sync, or not, no details).

dennisreimann:
So `200` in both cases but with `{ synched: bool }`?

NicolasDorier:
@dennisreimann yeah

BTCPayServer.Client/BTCPayServerClient.Health.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Threading.Tasks;
+using BTCPayServer.Client.Models;
+
+namespace BTCPayServer.Client
+{
+    public partial class BTCPayServerClient
+    {
+        public virtual async Task<ApiHealthData> GetHealth(CancellationToken token = default)
+        {
+            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health", bodyPayload: new {}), token);
```
Kukks, line 12: Remove the body payload

BTCPayServer/wwwroot/swagger/v1/swagger.template.health.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "components": {
+        "schemas": {
+            "ApplicationHealthData": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "synchronized": {
+                        "type": "boolean",
+                        "description": "True if the instance is fully synchronized"
```
Kukks, line 35: Mention that it is NBXplorer that is checked for sync

BTCPayServer/Controllers/GreenField/HealthController.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace BTCPayServer.Controllers.GreenField
+{
+    [ApiController]
+    public class HealthController : ControllerBase
+    {
+        [AllowAnonymous]
+        [HttpGet("~/api/v1/health")]
+        public ActionResult GetHealth(NBXplorerDashboard dashBoard)
+        {
+            ApiHealthData model = new ApiHealthData() {Synchronized = dashBoard.IsFullySynched()};
```
Kukks, line 15: nit: formatting

BTCPayServer.Client/BTCPayServerClient.Health.cs
```diff
@@ -1,15 +1,16 @@
...
 namespace BTCPayServer.Client
 {
     public partial class BTCPayServerClient
     {
-        public virtual async Task<bool> IsHealthy(CancellationToken token = default)
+        public virtual async Task<ApiHealthData> GetHealth(CancellationToken token = default)
         {
-            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health"), token);
-            return response.IsSuccessStatusCode;
+            var response = await _httpClient.SendAsync(CreateHttpRequest("api/v1/health", bodyPayload: new {}), token);
```
dennisreimann, line 12: It's weird that I have to pass an empty body here, otherwise I get an `415 Unsupported Media Type` error. @Kukks doesn't have this problem, do you know what might be happening here, @NicolasDorier?

PR #1472, opened 2020-04-17
Fix Payjoin p2sh

Kukks:
Since the test was not correct and that p2sh was never being enabled before, we missed these issues

Also having this on the payjoin receiver when adding p2sh inputs. 
```
System.InvalidOperationException: You need to provide P2WSH or P2SH redeem script with Coin.ToScriptCoin()
   at NBitcoin.Coin.GetScriptCode()
   at NBitcoin.TransactionBuilder.EstimateScriptSigSize(ICoin coin, Int32& witSize, Int32& baseSize)
   at NBitcoin.TransactionBuilder.EstimateSizes(Transaction tx, Int32& witSize, Int32& baseSize)
   at NBitcoin.TransactionBuilder.EstimateSize(Transaction tx, Boolean virtualSize)
   at NBitcoin.TransactionBuilder.EstimateFees(Transaction tx, FeeRate feeRate)
   at BTCPayServer.Payments.PayJoin.PayJoinEndpointController.Submit(String cryptoCode) in C:\Git\btcpayserver\BTCPayServer\Payments\PayJoin\PayJoinEndpointController.cs:line 391

```

### DIFF
--- a/BTCPayServer/Services/PayjoinClient.cs
+++ b/BTCPayServer/Services/PayjoinClient.cs
@@ -34,7 +34,7 @@ namespace BTCPayServer.Services
             if (i.WitnessUtxo.ScriptPubKey.IsScriptType(ScriptType.P2WPKH))
                 return ScriptPubKeyType.Segwit;
             if (i.WitnessUtxo.ScriptPubKey.IsScriptType(ScriptType.P2SH) &&
-                i.FinalScriptWitness.ToScript().IsScriptType(ScriptType.P2WPKH))
+                i.FinalScriptWitness.GetSigner().ScriptPubKey.IsScriptType(ScriptType.P2WPKH))
                 return ScriptPubKeyType.SegwitP2SH;
             return null as ScriptPubKeyType?;
         }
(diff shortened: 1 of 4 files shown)


BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -34,7 +34,7 @@ public static class PSBTExtensions
             if (i.WitnessUtxo.ScriptPubKey.IsScriptType(ScriptType.P2WPKH))
                 return ScriptPubKeyType.Segwit;
             if (i.WitnessUtxo.ScriptPubKey.IsScriptType(ScriptType.P2SH) &&
-                i.FinalScriptWitness.ToScript().IsScriptType(ScriptType.P2WPKH))
+                i.FinalScriptWitness.GetSigner().ScriptPubKey.IsScriptType(ScriptType.P2WPKH))
```
NicolasDorier, line 37: That's not correct. `GetSigner().ScriptPubKey` will give you P2SH.

Kukks: No, it gives `P2WPKH` i just tried it

Kukks:
Bumped and ran tests and removed the hacks, seems to work now

PR #1473, opened 2020-04-18
Payjoin: Better UIH1 & UIH2 based selection

### LINKED ISSUE #1470: Improve Payjoin utxo selector
Currently it only takes into consideration all outputs + the utxo you are considering.
Since we now have access to the values of each input of the original tx, we can expand this properly. 
Further, if we cannot find a single utxo to match the criteria, we can attempt to add multiple utxos.

Kukks:
closes #1470 

~I need to add some tests.~

### DIFF
--- a/BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
+++ b/BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
@@ -276,8 +276,8 @@ namespace BTCPayServer.Payments.PayJoin
                 utxos = utxos.Where(u => !prevOuts.Contains(u.Outpoint)).ToArray();
                 Array.Sort(utxos, UTXODeterministicComparer.Instance);
                 
-                foreach (var utxo in await SelectUTXO(network, utxos, psbt.Inputs.Select(input => input.WitnessUtxo.Value.ToDecimal(MoneyUnit.BTC)),  output.Value.ToDecimal(MoneyUnit.BTC),
-                    psbt.Outputs.Where(psbtOutput => !psbtOutput.Equals(output)).Select(psbtOutput => psbtOutput.Value.ToDecimal(MoneyUnit.BTC))))
+                foreach (var utxo in (await SelectUTXO(network, utxos, psbt.Inputs.Select(input => input.WitnessUtxo.Value.ToDecimal(MoneyUnit.BTC)),  output.Value.ToDecimal(MoneyUnit.BTC),
+                    psbt.Outputs.Where(psbtOutput => !psbtOutput.Equals(output)).Select(psbtOutput => psbtOutput.Value.ToDecimal(MoneyUnit.BTC)))).selectedUTXO)
                 {
                     selectedUTXOs.Add(utxo.Outpoint, utxo);
                 }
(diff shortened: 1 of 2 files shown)


BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -275,8 +275,9 @@ async Task UnlockUTXOs()
                 var prevOuts = originalTx.Inputs.Select(o => o.PrevOut).ToHashSet();
                 utxos = utxos.Where(u => !prevOuts.Contains(u.Outpoint)).ToArray();
                 Array.Sort(utxos, UTXODeterministicComparer.Instance);
-                foreach (var utxo in await SelectUTXO(network, utxos, output.Value,
-                    psbt.Outputs.Where(o => o.Index != output.Index).Select(o => o.Value).ToArray()))
+                
+                foreach (var utxo in (await SelectUTXO(network, utxos, psbt.Inputs.Select(input => input.WitnessUtxo.Value.ToDecimal(MoneyUnit.BTC)),  output.Value.ToDecimal(MoneyUnit.BTC),
+                    psbt.Outputs.Where(psbtOutput => !psbtOutput.Equals(output)).Select(psbtOutput => psbtOutput.Value.ToDecimal(MoneyUnit.BTC)))).selectedUTXO)
```
NicolasDorier, line 280: Why use `Equals` this is not well defined.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -504,42 +505,51 @@ private JObject CreatePayjoinError(int httpCode, string errorCode, string friend
...
                 }
 
+                if (invalid)
+                {
+                    continue;
+                }
                 if (await _payJoinRepository.TryLock(availableUtxo.Outpoint))
                 {
-                    return new UTXO[] { availableUtxo };
+                    return (new[] {availableUtxo}, false);
```
NicolasDorier, line 550: this is actually randomized, the result depends on how `availableUtxos` are sorted, and they are sorted randomly.
You should rename this if you use it in the tests. (Like an enum SelectedType with value like `UIH1,UIH2,Ordered`)

Kukks: yeah good point, this is specifically for tests

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -275,8 +275,9 @@ async Task UnlockUTXOs()
                 var prevOuts = originalTx.Inputs.Select(o => o.PrevOut).ToHashSet();
                 utxos = utxos.Where(u => !prevOuts.Contains(u.Outpoint)).ToArray();
                 Array.Sort(utxos, UTXODeterministicComparer.Instance);
-                foreach (var utxo in await SelectUTXO(network, utxos, output.Value,
-                    psbt.Outputs.Where(o => o.Index != output.Index).Select(o => o.Value).ToArray()))
+                
+                foreach (var utxo in (await SelectUTXO(network, utxos, psbt.Inputs.Select(input => input.WitnessUtxo.Value.ToDecimal(MoneyUnit.BTC)),  output.Value.ToDecimal(MoneyUnit.BTC),
+                    psbt.Outputs.Where(psbtOutput => psbtOutput != output).Select(psbtOutput => psbtOutput.Value.ToDecimal(MoneyUnit.BTC)))).selectedUTXO)
```
NicolasDorier, line 280: why do you even change this?

Kukks: the params separate the payment output from the others to compute what the modified payment output value will be

NicolasDorier: I know but before I was comparing the index, why did you modified to compare the object?

Kukks: ah, no clue why. Does it make a difference?

Kukks: reverted back t yours

NicolasDorier: I don't think so, but I had a doubt that I override the `.Equals` (which was a bad idea), I prefer not touching what does not break.

PR #1477, opened 2020-04-18
Improve static asset caching

dennisreimann:
Cache static assets for one year and set the correct cache control header. Adds the cache busting version based on file content to asset references to invalidate the cache on change. ([further details on the approach](https://andrewlock.net/adding-cache-control-headers-to-static-files-in-asp-net-core/) and [why one year](https://ashton.codes/set-cache-control-max-age-1-year/))

Most of the changes are the additions of the `asp-append-version="true"` attribute, the main configuration change is in `Startup.cs`.

### DIFF
--- a/BTCPayServer/Hosting/Startup.cs
+++ b/BTCPayServer/Hosting/Startup.cs
@@ -18,6 +18,7 @@ using System.IO;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using BTCPayServer.Security;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Net.Http.Headers;
 using System.Net;
 using BTCPayServer.PaymentRequest;
 using BTCPayServer.Services.Apps;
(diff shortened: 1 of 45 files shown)


NicolasDorier:
I think it should be the contrary. We should be explicit about which asset are cached for long.
If ever we fuck up with a new asset, then some people will be obliged to clear the cache of their browser.

Instead we should apply the caching strategy on specific asset which did not changed since the beginning (like the image in the background in the home page)

dennisreimann:
> If ever we fuck up with a new asset, then some people will be obliged to clear the cache of their browser.

They won't have to clear the cache, because the cache busting kicks in when we ship a fixed version and the appended version hashed changes (because it's based on the file content).

Right now we can have the opposing effect where people might have to clear their cache manually, because we have no cache busting. Haven't witnessed it myself, but @kukks [came across such a case](https://chat.btcpayserver.org/btcpayserver/pl/c8y8pxzak7bjxcuowro6t9o4ur) just recently.

dennisreimann:
Unit test fail with this error (output taken locally with #1478 applied):

```
4/18/2020 5:28:16 PM +00:00 :Tests:   FAILED: https://apps.apple.com/us/app/fully-noded/id1436425586 (/Users/dennisreimann/Sources/btcpay/server/BTCPayServer/Views/Server/RPCService.cshtml) The SSL connection could not be established, see inner exception.
4/18/2020 5:28:16 PM +00:00 :Tests:   FAILED: https://apps.apple.com/us/app/fully-noded/id1436425586 (/Users/dennisreimann/Sources/btcpay/server/BTCPayServer/Views/Server/RPCService.cshtml) The SSL connection could not be established, see inner exception.
4/18/2020 5:28:16 PM +00:00 :Tests:   FAILED: https://itunes.apple.com/us/app/google-authenticator/id388497605?mt=8 (/Users/dennisreimann/Sources/btcpay/server/BTCPayServer/Views/Manage/EnableAuthenticator.cshtml) The SSL connection could not be established, see inner exception.
```

NicolasDorier:
Ah so that's awesome. @Kukks @rockstardev  any downside you see?

dennisreimann:
@NicolasDorier Thanks for merging, I‘ll need to take a look at the integration tests tomorrow.

dennisreimann:
Ok, the integration tests worked for the merge commit.

PR #1484, opened 2020-04-19
Fix MySQL support (fixed width columns instead of blobs)

ketominer:
This is an attempt to fix MySQL support (works for me, probably requires further testing). Some columns didn't have fixed width but keys so creation was failing. It's also taking into account the max key size (767 Bytes) so the total of the columns from one key shouldn't exceed that.


ketominer:
ok scrap that, the Address fields are now too short for lightning.

PR #1485, opened 2020-04-19
Fix: Payment request redirects to non-existing (404) URL after paymen…

### LINKED ISSUE #1479: Payment request redirects to non-existing (404) URL after payment completed
**Describe the problem/bug**
I create a payment request, pay it and then the payment request is redirected to a non-existing URL within btcpayserver.

Tried with Bitcoin Lightning and Monero, seems to be independent of the payment method.

Happens everytime, with different URL that's 404, so it is not a rare bug, I can reproduce it everytime.

**Your environment**
* Version of BTCPay Server: 1.0.4.0
* Deployment method: docker

**Logs (if applicable)**
Basic logs can be found in Server Settings > Logs.

**Screenshots/Links**
Screencast:
https://www.screencast.com/t/SMh8tclC

Kukks:
fixes #1479

### DIFF
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -25,6 +25,7 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using NBitpayClient;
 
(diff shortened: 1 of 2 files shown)


britttttk:
@Kukks it uses the correct path/ID now when you try to view the payment request from invoices. But in local dev it still fails on redirect. I can't remember if that's expected or not in local dev.

Kukks:
Can you try with a fresh payment request? I've tried creating invoices both from the payment request UI and from the `/pay` endpoint and it worked

britttttk:
Actually @Kukks i just updated Chrome and it fixed it

britttttk:
Sorry I didn't get a good screenshot before update but it was saying something like OP's error about X-Frame-Options = Deny
image

britttttk:
@Kukks I thought it was just a localhost issue because it looks like this. If I refresh the page it will show the payment request fine. 
image

NicolasDorier:
Merging, I guess it is a localhost issue because it could not even connect.

Kukks:
Yeah, I think it is because you are using http and that can cause some mess sometimes with websockets

PR #1489, opened 2020-04-20
Re-add versioning for btcpay.js

dennisreimann:
As discussed with @Kukks – let me know if this works or I have overlooked something here.


Kukks:
Does it work ok if the js is referenced without the version parameter

dennisreimann:
As far as I checked it works for both cases, but please have a look as I don‘t know about everything that the script is supposed to do/scenarios to work in.

NicolasDorier:
@Kukks, merge it when you are satisfied.

Kukks:
I forgot but we actually have a selenium test that tests the modal so this is fine

PR #1495, opened 2020-04-23
Greenfield: Add update store API

Kukks:


BTCPayServer/Controllers/GreenField/StoresController.cs
```diff
@@ -73,6 +73,23 @@ public async Task<ActionResult<Client.Models.StoreData>> CreateStore(CreateStore
...
+        [HttpPut("~/api/v1/stores/{storeId}")]
+        public async Task<ActionResult> UpdateStore(string storeId, UpdateStoreRequest request)
+        {
+            var store = HttpContext.GetStoreData();
+            if (store == null)
+            {
+                return NotFound();
+            }
+
+            if (request?.Name is null)
```
NicolasDorier, line 87: if the name is null, this is not a bug. This may simply mean that the user don't want to update the name.

When we will have other updatable properties this may be the case.

Kukks:
No, on PUT you replace the whole entity. On PATCH you do partial updating

NicolasDorier:
that mean that we will have two different method one PUT and one PATCH with exactly the same code, that we will have both maintain except for this small difference? There is no advantage to that IMHO, why would people wants PUT when they have PATCH to achieve same thing?

Kukks:
> that mean that we will have two different method one PUT and one PATCH with exactly the same code, that we will have both maintain except for this small difference? There is no advantage to that IMHO, why would people wants PUT when they have PATCH to achieve same thing?

It is easier to work with PUT, since you have to track unit of work for PATCH and can be hard depending on app. Also sometimes values are null able and make your proposal hard to manage when doing a PATCH update.

We can make PATCH update the model and provide it to PUT and not have duplicate code

NicolasDorier:
In our case, the unit of work is very easy. We get the blob, we update it, we store it again.
It is way easier to just have a PUT and ignore the fields that are not set.

Or the reverse: You do the PATCH, then the PUT just fill out what is missing `req.Name ??= "Default Name"` before calling the PATCH's method.

Kukks:
> In our case, the unit of work is very easy. We get the blob, we update it, we store it again.
> It is way easier to just have a PUT and ignore the fields that are not set.
> 
> Or the reverse: You do the PATCH, then the PUT just fill out what is missing `req.Name ??= "Default Name"` before calling the PATCH's method.
Ah I was referring to it being hard depending on the client implementation. Some rest based orms in js can be a bitch to deal with partial diff updates for example.

Kukks:
> 
> 
> In our case, the unit of work is very easy. We get the blob, we update it, we store it again.
> It is way easier to just have a PUT and ignore the fields that are not set.
> 
> Or the reverse: You do the PATCH, then the PUT just fill out what is missing `req.Name ??= "Default Name"` before calling the PATCH's method.

Also, it is more complex when doing partial updates when you need to take into account other possible scenarios (to stay consistent across the full api). For example, if we had a property `decimal? X` on the model, how do you decide if:
* you want to set X to null
* you want to not update X


Anyway, there is a spec for partial updates, to help make this consistent across APIs and reduce pain for everyone, https://docs.microsoft.com/en-us/aspnet/core/web-api/jsonpatch?view=aspnetcore-3.1

Kukks:
@NicolasDorier  more info at http://jsonpatch.com/

NicolasDorier:
There is no need of doing something generic when we can do specific. JsonPatch is unbearable to read and write. Can't strongly type it.

This stuff is not simpler than just having a plain route checking if the field is null or present and skipping it.

> you want to set X to null 
you want to not update X

Good point, though I would be surprised there is no way to do

NicolasDorier:
I was thinking the user can just GetStore then do an update. We can just share same datatype between both.

BTCPayServer.Client/Models/UpdateStoreRequest.cs
```diff
@@ -0,0 +1,7 @@
+namespace BTCPayServer.Client.Models
+{
+    public class UpdateStoreRequest
```
NicolasDorier, line 3: as per https://github.com/btcpayserver/btcpayserver/pull/1495#issuecomment-619821716 we should use the same type for `GetStore` call, maybe should call `GreenFieldStoreData`?

Kukks:
> 
> 
> I was thinking the user can just GetStore then do an update. If we share datatypes, that's good enough.

Agreed, this is why I would like to start  with `PUT` and keep it simple, For the client, it is easy enough:
* x = GET
* x.y= z
* PUT x


We can add PATCH after by:
* x = GET
* x.y= z
* PATCH x.y (call GET then apply x.y to x then PUT on server side )

Kukks:
> 
> 
> There is no need of doing something generic when we can do specific. JsonPatch is unbearable to read and write. Can't strongly type it. Again a "meta protocol" which make only sense if you are updating JSON docs, which is not the case.
> 
> This stuff is not simpler than just having a plain route checking if the field is null or present and skipping it.
> 
> > you want to set X to null
> > you want to not update X
> 
> Good point, though I would be surprised there is no way to do

In the past , I have had to build similar systems for partial updating, and it always ends up very similar to jsonpatch, where you pass a list of changes with `property name`,`new value` and `change type`

BTCPayServer.Client/Models/UpdateStoreRequest.cs
Kukks: Yeah makes sense

NicolasDorier:
test fail

PR #1499, opened 2020-04-24
Auto label utxos based on invoice and payjoin

Kukks:
This PR introduces automatic labelling to transactions.
* If it is an incoming tx to an invoice, it will tag it as such.
* If it was a payjoin tx , it will tag it as such.
* If a transaction's inputs were exposed to a payjoin sender, we tag it as such.

image

### DIFF
--- a/BTCPayServer.Data/Data/WalletData.cs
+++ b/BTCPayServer.Data/Data/WalletData.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Data
 {
@@ -24,12 +25,35 @@ namespace BTCPayServer.Data
                 throw new ArgumentNullException(nameof(value));
             if (color == null)
                 throw new ArgumentNullException(nameof(color));
+            if (value.StartsWith("{"))
+            {
+                var jObj = JObject.Parse(value);
+                if (jObj.ContainsKey("value"))
+                {
+                    switch (jObj["value"].Value<string>())
+                    {
+                        case "invoice":
+                            Text = "invoice";
+                            Tooltip = "Received through an invoice";
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                        case "pj-exposed":
+                            Text = "payjoin-exposed";
+                            Tooltip = "This utxo was exposed through a payjoin proposal";
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                    }
+                }
+            }
             Value = value;
             Color = color;
         }
 
         public string Value { get; }
         public string Color { get; }
+        public string Link { get; }
+        public string Tooltip { get; }
+        public string Text { get; }
 
         public override bool Equals(object obj)
         {
--- a/BTCPayServer.Data/Data/WalletData.cs
+++ b/BTCPayServer.Data/Data/WalletData.cs
@@ -31,22 +31,31 @@ namespace BTCPayServer.Data
                     {
                         case "invoice":
                             Value = "invoice";
-                            Tooltip = "Received through an invoice";
+                            Tooltip = $"Received through an invoice ({jObj["id"].Value<string>()})";
                             Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
                             break;
                         case "pj-exposed":
                             Value = "payjoin-exposed";
-                            Tooltip = "This utxo was exposed through a payjoin proposal";
+                            Tooltip = $"This utxo was exposed through a payjoin proposal for an invoice ({jObj["id"].Value<string>()})";
                             Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                        default:
+                            Value = value;
                             break;
                     }
                 }
             }
-            Value = value;
+            else
+            {
+                Value = value;
+            }
+            RawValue = value;
+            
             Color = color;
         }
 
         public string Value { get; }
+        public string RawValue { get; }
         public string Color { get; }
         public string Link { get; }
         public string Tooltip { get; }
--- a/BTCPayServer.Data/Data/WalletData.cs
+++ b/BTCPayServer.Data/Data/WalletData.cs
@@ -1,8 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Data
@@ -33,12 +30,12 @@ namespace BTCPayServer.Data
                     switch (jObj["value"].Value<string>())
                     {
                         case "invoice":
-                            Text = "invoice";
+                            Value = "invoice";
                             Tooltip = "Received through an invoice";
                             Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
                             break;
                         case "pj-exposed":
-                            Text = "payjoin-exposed";
+                            Value = "payjoin-exposed";
                             Tooltip = "This utxo was exposed through a payjoin proposal";
                             Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
                             break;
@@ -53,7 +50,6 @@ namespace BTCPayServer.Data
         public string Color { get; }
         public string Link { get; }
         public string Tooltip { get; }
-        public string Text { get; }
 
         public override bool Equals(object obj)
         {
--- a/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
+++ b/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
@@ -1,0 +1,70 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using BTCPayServer.Events;
+using BTCPayServer.Payments;
+using BTCPayServer.Payments.Bitcoin;
+using BTCPayServer.Services;
+using NBitcoin;
+
+namespace BTCPayServer.HostedServices
+{
+    public class TransactionLabelMarkerHostedService : EventHostedServiceBase
+    {
+        private readonly WalletRepository _walletRepository;
+
+        public TransactionLabelMarkerHostedService(EventAggregator eventAggregator, WalletRepository walletRepository) :
+            base(eventAggregator)
+        {
+            _walletRepository = walletRepository;
+        }
+
+        protected override void SubscribeToEvents()
+        {
+            Subscribe<InvoiceEvent>();
+        }
+
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            await base.StartAsync(cancellationToken);
+            _ = StartLoop();
+
+        }
+
+        private async Task StartLoop()
+        {
+            while (!_Cts.IsCancellationRequested)
+            {
+                await _walletRepository.ProcessLabels();
+                await Task.Delay(1000, _Cts.Token);
+            }
+        }
+
+        protected override Task ProcessEvent(object evt, CancellationToken cancellationToken)
+        {
+            if (evt is InvoiceEvent invoiceEvent && invoiceEvent.Name == InvoiceEvent.ReceivedPayment && 
+                invoiceEvent.Payment.GetPaymentMethodId().PaymentType == BitcoinPaymentType.Instance && 
+                invoiceEvent.Payment.GetCryptoPaymentData() is BitcoinLikePaymentData bitcoinLikePaymentData)
+            {
+                var payjoinLabelColor = "#51b13e";
+                var invoiceLabelColor = "#0f3b21";
+                var walletId = new WalletId(invoiceEvent.Invoice.StoreId, invoiceEvent.Payment.GetCryptoCode());
+                var transactionId = bitcoinLikePaymentData.Outpoint.Hash;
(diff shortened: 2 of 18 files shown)


NicolasDorier:
@Kukks I would suggest something else:

1. For payjoin exposed, the link to the invoice is quite useless
2. For transactions tagged by the invoice, I think what we should do is to have the label "invoice", but add a link emoticon next to it so the user understand that clicking on it will bring them to the invoice details.

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -97,5 +99,55 @@ public async Task SetWalletTransactionInfo(WalletId walletId, string transaction
                 }
             }
         }
+
+
+        private ConcurrentQueue<Func<Task>> labelProcessor = new ConcurrentQueue<Func<Task>>();
```
NicolasDorier, line 104: You should use a Channel in the `TransactionLabelMarkerHostedService` instead.

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -97,5 +99,55 @@ public async Task SetWalletTransactionInfo(WalletId walletId, string transaction
...
+
+        public async Task ProcessLabels()
```
NicolasDorier, line 106: should be in the `TransactionLabelMarkerHostedService `

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -97,5 +99,55 @@ public async Task SetWalletTransactionInfo(WalletId walletId, string transaction
...
+        {
+            while (labelProcessor.TryDequeue(out var func))
+            {
+                await func.Invoke();
+            }
+        }
+
+        public void AddLabels(WalletId walletId,
+            Dictionary<uint256, List<(string color, string label)>> transactionLabels)
+        {
```
NicolasDorier, line 116: should returns Task and not queuing anything.

NicolasDorier:
Not too fan of the hard coded coloring, but anyway this is a fix for later. Will later allow people to chose their own color based on the label text.

Kukks:
> 
> 
> @Kukks I would suggest something else:
> 
>     1. For payjoin exposed, the link to the invoice is quite useless
> 
>     2. For transactions tagged by the invoice, I think what we should do is to have the label "invoice", but add a link emoticon next to it so the user understand that clicking on it will bring them to the invoice details.

I was going to do a simple `payjoin-exposed` label, but we would valuable information in my opinion. Knowing which and how many invoices provided the utxo as a payjoin proposition helps point out if there were spy attacks and if so, how big and long of an attack. If not for attack analysis reasons, to know how many times on average the utxo is leaked to failed payjoin attempts.

Kukks:
> 
> 
> Not too fan of the hard coded coloring, but anyway this is a fix for later. Will later allow people to chose their own color based on the label text.

I decided on hard coded colors simply because I wanted them to appear grouped together in UI. I think in a future PR, we should do label coloring deterministically based on label content

Kukks:
> 
> 
> @Kukks I would suggest something else:
> 
>     1. For payjoin exposed, the link to the invoice is quite useless
> 
>     2. For transactions tagged by the invoice, I think what we should do is to have the label "invoice", but add a link emoticon next to it so the user understand that clicking on it will bring them to the invoice details.
> 
> 
> For 1 and 2, the goal is that later on, when we allow the user to customize label's color given the label's text, then all "invoice" label will have the same color. It also fix the horrible UI big label makes.

For point 2, how should we do it? Should I hotwire the label to a json value `{label:"invoice", type:"invoice', ref:"invoiceId"}` for example?

NicolasDorier:
@Kukks `{label:"invoice", type:"invoice', link:"relativeLink"`

Kukks:
> 
> 
> @Kukks `{label:"invoice", link:"relativeLink"` no need of "type"

Can't use a relative link  in case we change url routes for action at some point in future. Better to have a "type" that can be used to map

NicolasDorier:
@Kukks if route were to change, we can redirect.

Kukks:
> @Kukks if route were to change, we can redirect. And really, even if we did not redirect, it is not like this should be a big issue for anybody. This is a just a UI concern.

If we expose this through api eventually, the redirect url would not be useful unless you parse it through hacks, still think a type is better overall

Kukks:
Updated it to support all we talked about, i think it is bitchin' now ( did I use that correctly?)!
Will upload gif soon

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        protected override void SubscribeToEvents()
+        {
+            Subscribe<InvoiceEvent>();
+        }
+
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            await base.StartAsync(cancellationToken);
+            _ = StartLoop();
```
NicolasDorier, line 38: We have a proper base class for starting loops like this already, which correctly wait the loop to be over before stopping.

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -0,0 +1,151 @@
...
+                            }
+
+                            if (update)
+                            {
+                                await _walletRepository.SetWalletTransactionInfo(updateTransactionLabel.WalletId,
+                                    pair.Key.ToString(), walletTransactionInfo);
+                            }
+                        }));
+                    }
+                    catch when (_Cts.Token.IsCancellationRequested)
```
NicolasDorier, line 81: if you use the proper base class, you can remove those catches

BTCPayServer.Data/Data/WalletData.cs
```diff
@@ -24,12 +25,35 @@ public Label(string value, string color)
...
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                        case "pj-exposed":
+                            Text = "payjoin-exposed";
+                            Tooltip = "This utxo was exposed through a payjoin proposal";
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                    }
+                }
+            }
```
NicolasDorier, line 47: what is the difference between text and value? old label only have value set, is this normal?

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            await base.StartAsync(cancellationToken);
+            _ = StartLoop();
+        }
+
+        private async Task StartLoop()
+        {
+            while (await _labels.Reader.WaitToReadAsync(_Cts.Token))
```
NicolasDorier, line 43: what is this loop? you are in 2019 or what?

```csharp
await foreach (var updateTransactionLabel in _labels.Reader.ReadAllAsync())
```

Kukks: I copy pasta the loop from your code from the base class :)

BTCPayServer.Data/Data/WalletData.cs
Kukks: Yeah, but maybe I can reduce this to one property actually

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
NicolasDorier: yeah, that was 2019 code! :p

pavlenex:
Great feature! This will be very useful for people. I have a few remarks and notes:
1. For me, when an invoice is created manually and paid, it's broken.
2. It is unclear what the `X` icon should do, remove the label I suppose? It doesn't for me.
x
3. Is it **really** useful to include `invoiceid` in the label? I don't see how it benefits. It's kinda hard to assume merchant would group the UTXO by the invoice ID. It's better to be by stores honestly. Because if I was a business, I wouldn't care, all those UTXO's come from my customers. `storeid` would be better, that way I can group donations, my woocommerce, etc
4. The `X` color on the dark green is barely visible

Kukks:
> 
> 
> Great feature! This will be very useful for people. I have a few remarks and notes:
> 
>     1. For me, when an invoice is created manually and paid, it's broken.
> 
>     2. It is unclear what the `X` icon should do, remove the label I suppose? It doesn't for me.
>        x
> 
>     3. Is it **really** useful to include `invoiceid` in the label? I don't see how it benefits. It's kinda hard to assume merchant would group the UTXO by the invoice ID. It's better to be by stores honestly. Because if I was a business, I wouldn't care, all those UTXO's come from my customers. `storeid` would be better, that way I can group donations, my woocommerce, etc
> 
>     4. The `X` color on the dark green is barely visible

Will check the issues  soon.
* Makes no sense to group by store, as all utxos in the wallet belong to the one store. It makes sense to have the invoice id to see which customer sent you that utxo. 
* Suggest a new color for invoice

pavlenex:
Ah right... completely overlooked that, perhaps in the future when we extract wallet from stores.

I still think invoice ID  brings very little if any usefulness. `Invoice` is enough. I find it extremely unlikely that anybody would group by invoice ID. And if there's somebody that's crazy to do that, they'd have the `i`. Normally a sane person would group by `invoice` (this is my business) `donation`, etc.
Labels should be simple thing to help people get around, not cover every edge-case, anyway just how I feel about it. With wallet we should encourage people to use it and avoid overwhelming them with edge-case stuff.

> Suggest a new color for invoice

Don't have any preference here honestly. Perhaps `  --btcpay-brand-secondary` `CEDC21`. It'll make that X more visible at least. But may look bad with white text color.

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -0,0 +1,122 @@
...
+            _walletRepository = walletRepository;
+        }
+
+        protected override void SubscribeToEvents()
+        {
+            Subscribe<InvoiceEvent>();
+            Subscribe<UpdateTransactionLabel>();
+        }
+
+        public override async Task StartAsync(CancellationToken cancellationToken)
```
NicolasDorier, line 33: useless

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -0,0 +1,122 @@
...
+            return ("#cedc21", JObject.FromObject(new {value = "invoice", id = invoice}).ToString());
+        }
+
+        public static (string color, string label) PayjoinExposedLabelTemplate(string invoice)
+        {
+            return ("#51b13e", JObject.FromObject(new {value = "pj-exposed", id = invoice}).ToString());
+        }
+
+        public WalletId WalletId { get; set; }
+        public Dictionary<uint256, List<(string color, string label)>> TransactionLabels { get; set; }
```
NicolasDorier, line 114: Add proper `ToString` events are shown in the logs.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -474,7 +478,18 @@ async Task UnlockUTXOs()
             }
             await _btcPayWalletProvider.GetWallet(network).SaveOffchainTransactionAsync(originalTx);
             _eventAggregator.Publish(new InvoiceEvent(invoice, 1002, InvoiceEvent.ReceivedPayment) {Payment = payment});
-
+            _eventAggregator.Publish(new UpdateTransactionLabel()
+            {
+                WalletId = new WalletId(invoice.StoreId, network.CryptoCode),
+                TransactionLabels = selectedUTXOs.Select(utxo =>
```
NicolasDorier, line 484: this will crash if two `selectedUTXOs` have same transaction Id

BTCPayServer.Data/Data/WalletData.cs
```diff
@@ -24,12 +22,43 @@ public Label(string value, string color)
...
+                    switch (jObj["value"].Value<string>())
+                    {
+                        case "invoice":
+                            Value = "invoice";
+                            Tooltip = $"Received through an invoice ({jObj["id"].Value<string>()})";
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
+                            break;
+                        case "pj-exposed":
+                            Value = "payjoin-exposed";
+                            Tooltip = $"This utxo was exposed through a payjoin proposal for an invoice ({jObj["id"].Value<string>()})";
```
NicolasDorier, line 39: Crash if somebody call a label manually "pj-exposed"

BTCPayServer.Data/Data/WalletData.cs
```diff
@@ -24,12 +22,43 @@ public Label(string value, string color)
...
+                            Link = jObj.ContainsKey("id") ? $"/invoices/{jObj["id"].Value<string>()}" : "";
```
NicolasDorier, line 40: Should use the action url builder of MVC.

NicolasDorier:
Need to make sure the user can't manually create a tag starting by `[` or `{`

BTCPayServer.Data/Data/WalletData.cs
Kukks: No, needs to be a json obj (starting with `{`)

BTCPayServer.Data/Data/WalletData.cs
Kukks: Yeah, was keeping it for another PR just to get this POC out of the way

BTCPayServer.Data/Data/WalletData.cs
NicolasDorier: This mean that you need to validate manual labels to enforce they don't start by `{`. Crash if somebody call a label manually "{pj-exposed".

Kukks: made it trim {

NicolasDorier:
Can you refactor with the `LabelFactory`?

Kukks:
doing it now

PR #1507, opened 2020-04-26
Add tooltip for store deriv + show "Not set" + "show end piece of string"

### LINKED ISSUE #1506: Store Settings UI: Show begining and end of Derivation scheme
Currently, the Derivation scheme is formatted as "xpubxxxxxx..." whereas I would like it to be "xpubxxxxxx...xxxx-[p2sh]" with the ... in the middle. Since the Derivation scheme format places most of the meaning at the beginning and end of the scheme.

**Provide examples**
If applicable provide examples, wireframes, sketches or images to better explain your idea.
範囲を選択_001

Kukks:
fixes #1506

image

image

### DIFF
--- a/BTCPayServer/Views/Stores/UpdateStore.cshtml
+++ b/BTCPayServer/Views/Stores/UpdateStore.cshtml
@@ -107,7 +107,21 @@
                     {
                         <tr class="@(@scheme.Collapsed? "collapsed": "")">
                             <td>@scheme.Crypto</td>
-                            <td class="smMaxWidth text-truncate">@scheme.Value</td>
+                            <td title="@scheme.Value" data-toggle="tooltip">
+                                @if (string.IsNullOrEmpty(scheme.Value))
+                                {
+                                    <span class="smMaxWidth">Not set</span>
+                                }
+                                else
+                                {
+                                    <div class="d-inline-flex justify-content-between smMaxWidth">
+                                        <span class="text-truncate ">
+                                            @scheme.Value.Substring(0, scheme.Value.Length - 6)
+                                        </span>
+                                        @scheme.Value?.Substring(scheme.Value.Length - 6, 6)
+                                    </div>
+                                }
+                            </td>
                             <td style="text-align:center;">
                                 @if(scheme.Enabled)
                                 {
--- a/BTCPayServer/Views/Stores/UpdateStore.cshtml
+++ b/BTCPayServer/Views/Stores/UpdateStore.cshtml
@@ -1,4 +1,5 @@
-﻿@model StoreViewModel
+﻿@using System.Text.RegularExpressions
+@model StoreViewModel
 @{
     Layout = "../Shared/_NavLayout.cshtml";
     ViewData.SetActivePageAndTitle(StoreNavPages.Index, "Profile");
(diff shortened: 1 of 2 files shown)


junderw:
instead of 6, you should use regexp to get these two groups:

`^(\d+-of-\S{4}|\S{6}).*?((?:-\[(?:[^\]])+\])+|\S{6})$`

Here you can look at each match and the two groups each match has: (it takes a little while to load the page for the explanation and matches/groups to show up.
https://regex101.com/r/amNOxd/3

Just add ... between that.

The longest one is: `12-of-xpub...-[legacy]-[keeporder]` which is 31 characters without the ...

junderw:
In English, the Regex above says:

"Give me m-of-4letters OR the first 6 letters at the beginning,
then for the end, give me all portions connected that look like -[letters] OR the last 6 letters if there is none"

junderw:
Or you can take the last part:

`((?:-\[(?:[^\]])+\])+|\S{6})$`

And use this to just grab the second group of the link I sent you. Then use that instead of the last 6 you put on the end.

junderw:
currently your PR will show a legacy keeporder multisig as `...order]`

Kukks:
That would be too long, the column width space is very limited, I chose 6 just for basic scenarios. if you had legacy + multisig+ others, it would not fit either way

junderw:
If you just keep the 2nd group, thats `-[legacy]-[keeporder]` which is 21 characters.

BTCPayServer/Views/Stores/UpdateStore.cshtml
```diff
@@ -107,7 +107,21 @@
...
+                            <td title="@scheme.Value" data-toggle="tooltip">
+                                @if (string.IsNullOrEmpty(scheme.Value))
+                                {
+                                    <span class="smMaxWidth">Not set</span>
+                                }
+                                else
+                                {
+                                    <div class="d-inline-flex justify-content-between smMaxWidth">
+                                        <span class="text-truncate ">
+                                            @scheme.Value.Substring(0, scheme.Value.Length - 6)
```
NicolasDorier, line 119: what happen if the length has less than 6 chars?

junderw: TIL it shows the label. haha.

NicolasDorier: Yeah, this is useful as the internal representation of those descriptors is using the nbxplorer format. But some people (importing from electrum) get very confused when they don't see zpub... So the solution was, if you import zpub we show zpub as label, but convert to our format internally.

This label can be changed in the wallet's settings.

junderw: - Name of column is Derivation Scheme
- Content of column is Derivation Scheme if label is empty else label
- Electrum zpub ypub etc. are automatically detected and automatically entered as label of wallet

Am I correct here?...

Why not just:

- Change column label to "Label"
- Default label = Derivation Scheme / Electrum converted Derivation Scheme
- Show label instead of switching between two different sets of data.

Kukks: Sounds correct BUT the original issue still remains when you use p2sh/legacy/etc and not import an electrrum-ish format

Maybe this PR can be merged now (I'm using that 2nd regex you provided)?

junderw:
I have tried with a few labels, even a few labels that mimic derivation Schemes.

Looks good to me.

junderw:
ahhhh... maybe we should change + to {1,2}... otherwise if I label my wallet to something like

`wallet-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]` then all of those -[a] will be shown lol

```csharp
var match =Regex.Match(scheme.Value, @"((?:-\[(?:[^\]])+\]){1,2}|\S{6})$");
```

Kukks:
> 
> 
> ahhhh... maybe we should change + to {1,2}... otherwise if I label my wallet to something like
> 
> `wallet-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]-[a]` then all of those -[a] will be shown lol
> 
> ```cs
> var match =Regex.Match(scheme.Value, @"((?:-\[(?:[^\]])+\]){1,2}|\S{6})$");
> ```

Quite an edge case though...it's like somebody noticed you can get bruised skin by hitting yourself with a stick...and then you do it for 20 times :)

junderw:
yeah you're right... lol...

Someday if there are 3 modifiers we'd need to change this lol

PR #1490, opened 2020-04-21
Make both sides of payment request table look same

bolatovumar:
fix #1488

# Screenshots

Light theme:
Screen Shot 2020-04-20 at 8 51 18 PM

Dark theme:
Screen Shot 2020-04-20 at 8 51 36 PM

Classic theme:
Screen Shot 2020-04-20 at 8 51 54 PM

Casa theme:
Screen Shot 2020-04-20 at 8 52 11 PM

Issue #1488, opened 2020-04-19
Dark Theme Contrast Issue - Payment Requests


britttttk:
**Describe the problem/bug**
The default theme used in payment requests is hard to read when using overall Dark Theme mode.

**Your environment**
* Version of BTCPay Server: BTCPayServer v1.0.4.1
* Deployment method: Debug Regtest

**To Reproduce**
1. Create Payment Request (looks ok)
2. Click Pay now (can't see the invoice and transaction info text very well)
3. Payments completed are also hard to read (see photo below)

**Expected behavior**
There is more contrast on the payment text in payment requests when using Dark Theme.

**Actual behavior**
Text is very hard to read. 

**Screenshots/Links**
![image](https://user-images.githubusercontent.com/39231115/79699451-f2272b00-824c-11ea-8f6e-a78ae797a503.png)

**Additional context**
No payment request functionality issues, just viewing payment/transaction text.

NicolasDorier:
ping @dennisreimann

PR #1514, opened 2020-04-27
Finetune colors

dennisreimann:
This builds on the discussions we had in #1488 and #1490.

Getting rid of the light/dark classes would be a bigger change as we use thesee quite a bit in our current templates. Hence I decided to flip their meaning for dark themes and inverted the color variables there: Light in dark mode means just a little lighter than dark instead of bright light. IMHO this works very well …

dark

casa

light

The other changes are related to finetuning the neutral color progression: As we haven't specified these completely in Figma, this was a mix of new and old grayscale colors. I've made the progression more harmonic and adjusted some of the styles that made use of the changed variables to fit the look we had before these changes (e.g. lightening the input borders, because the neutral-400 is now darker than before).

/cc @bolatovumar

### DIFF
--- a/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
+++ b/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
@@ -1186,7 +1186,7 @@ pre {
   color: var(--btcpay-color-neutral-700);
   background-color: var(--btcpay-color-white);
   background-clip: padding-box;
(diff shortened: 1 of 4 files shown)


dennisreimann:
To make it easier to spot the difference: Here's the old vs. the new progression – the 500 didn't fit in there.

OLD

old

NEW

new

PR #1519, opened 2020-04-28
Updates from design system

dennisreimann:
Sorry, I know this is annoying, please bear with me! 😬

The first commit streamlines the CSS variable names and makes them consistent with what we use in the design system. 
The second one reintroduces some of the light/dark classes in Bootstrap I once removed and improves the badge colors.

Basically this PR makes it easier to use design system as a single source of truth, as I'd like to maintain the bootstrap.css there (hopefully we don't have to change anything anymore until Bootstrap 5 comes out and likely support CSS variables). The design system now also features the [bootstrap kitchensink page](https://design.btcpayserver.org/views/bootstrap/), which shows all the elements and can be used to check the themes.


Eskyee:
high graded work @dennisreimann 💚

PR #1529, opened 2020-04-30
Fix decimal points shown in Checkout UI based on currency ( always showed btc decimal precision before)

Kukks:


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -224,6 +224,27 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
...
             var paymentMethodHandler = _paymentMethodHandlerDictionary[paymentMethodId];
+            
+            var divisibility = _CurrencyNameTable.GetNumberFormatInfo(paymentMethod.GetId().CryptoCode, false)?.CurrencyDecimalDigits;
+
+            string ShowMoney(Money money)
+            {
+                if (!divisibility.HasValue) return money.ToString();
+                var res = money.ToString(false, true);
+                var split = res.Split('.');
+                var decimals = res.Split('.')[1];
```
NicolasDorier, line 235: will crash if there is no decimal

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -224,6 +224,27 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
                 : (decimal?)null;
 
             var paymentMethodHandler = _paymentMethodHandlerDictionary[paymentMethodId];
+            
+            var divisibility = _CurrencyNameTable.GetNumberFormatInfo(paymentMethod.GetId().CryptoCode, false)?.CurrencyDecimalDigits;
+
+            string ShowMoney(Money money)
```
NicolasDorier, line 230: I really don't like those string hacks, just use proper .NET Feature on string conversion of decimal.

Kukks:
@NicolasDorier  updated

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -230,19 +230,8 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
...
-                    return res;
-                }
-                if (decimals == "00")
-                {
-                    decimals = string.Empty;
-                }
-                res = $"{split[0]}{(divisibility.Value > 0? ".": string.Empty)}{decimals.PadRight(divisibility.Value, '0')}";
-                return res;
+                var format = $"0{(divisibility.Value > 0 ? "." : string.Empty)}".PadRight(divisibility.Value, '0');
+                return money.ToDecimal(MoneyUnit.BTC).ToString(format, CultureInfo.InvariantCulture);
```
NicolasDorier, line 234: drake_meme.gif

PR #1533, opened 2020-05-02
Ensure dropdown option doesn't overflow container

### LINKED ISSUE #1526: Brazilian Portuguese Localization issue - Checkout experience
**Describe the problem/bug**
One of the translated strings is breaking the design. I believe this specific string should have the char limit set to 44 chars with spaces to avoid bug in all languages.
Fixed the string to fit in Transifex - please update

**To Reproduce**
1. Create invoice (in pt_BR version)
2. Click on 'Conversion`
3. See error

**Expected behavior**
Clean design

**Actual behavior**
Broken design

**Screenshots/Links**
btc_design-break

**Additional context**
Added fix to translation and comment here:
https://www.transifex.com/btcpayserver/btcpayserver/translate/#pt_BR/enjson/156624077

I've also reviewed all the translations and made other minor changes/improvements to the checkout experience strings.

bolatovumar:
fix #1526

Should look like this now:
Screen Shot 2020-05-01 at 8 39 07 PM

### DIFF
--- a/BTCPayServer/wwwroot/vendor/jquery-prettydropdowns/prettydropdowns.css
+++ b/BTCPayServer/wwwroot/vendor/jquery-prettydropdowns/prettydropdowns.css
@@ -1,5 +1,6 @@
 .prettydropdown {
     position: relative;
+    max-width: 100%; /* Make sure it's never larger than parent container */
     min-width: 72px; /* 70px + borders */
     display: inline-block;
 }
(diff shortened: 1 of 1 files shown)


bitcoinheiro:
Still pt_BR translation needs to be updated - but this fixes the char limit issue. Cheers!

NicolasDorier:
I do before release

PR #1535, opened 2020-05-02
Bump elements and fix test

Kukks:


NicolasDorier:
why the test was passing before?

Kukks:
because the CI does not run altcoin tests

PR #1539, opened 2020-05-03
Support specifying payment method through apps

### LINKED ISSUE #1525: Support specifying payment method through apps
We should expose the invoice creation param `SupportedTransactionCurrencies` to apps. This way you can offer items at a discounted/steeper rate if they pay with a specific payment method or have items available exclusively to just onchain or off chain payments.

Kukks:
closes #1525
This is great if you want to offer items with an incentive to use a specific payment method without messing with the rates.
For example, you can have `Item A` which costs 25$ and your store is configured for USDT and BTC. You can create two items, with different prices, where Item A costs 20$ if you pay with bitcoin or 25$ if you paid in dirty fiat pegs

### DIFF
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -112,6 +112,7 @@ namespace BTCPayServer.Controllers
             }
             string title = null;
             var price = 0.0m;
+            Dictionary<string, InvoiceSupportedTransactionCurrency> paymentMethods = null;
             ViewPointOfSaleViewModel.Item choice = null;
             if (!string.IsNullOrEmpty(choiceKey))
             {
@@ -130,6 +131,12 @@ namespace BTCPayServer.Controllers
                     {
                         return RedirectToAction(nameof(ViewPointOfSale), new { appId = appId });
                     }
+                }
+
+                if (choice?.PaymentMethods?.Any() is true)
+                {
+                    paymentMethods = choice?.PaymentMethods.ToDictionary(s => s,
+                        s => new InvoiceSupportedTransactionCurrency() {Enabled = true});
                 }
             }
             else
@@ -183,10 +190,7 @@ namespace BTCPayServer.Controllers
                     ExtendedNotifications = true,
                     PosData = string.IsNullOrEmpty(posData) ? null : posData,
                     RedirectAutomatically = settings.RedirectAutomatically,
-                    SupportedTransactionCurrencies = !(choice?.PaymentMethods?.Any() ?? false)
-                        ? null
-                        : choice?.PaymentMethods?.ToDictionary(s => s,
-                            s => new InvoiceSupportedTransactionCurrency() {Enabled = true})
+                    SupportedTransactionCurrencies = paymentMethods,
                 }, store, HttpContext.Request.GetAbsoluteRoot(),
                     new List<string>() { AppService.GetAppInternalTag(appId) },
                     cancellationToken);
@@ -275,6 +279,7 @@ namespace BTCPayServer.Controllers
             var store = await _AppService.GetStore(app);
             var title = settings.Title;
             var price = request.Amount;
+            Dictionary<string, InvoiceSupportedTransactionCurrency> paymentMethods = null;
             ViewPointOfSaleViewModel.Item choice = null;
             if (!string.IsNullOrEmpty(request.ChoiceKey))
             {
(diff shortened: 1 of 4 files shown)


BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -182,6 +183,10 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
                     ExtendedNotifications = true,
                     PosData = string.IsNullOrEmpty(posData) ? null : posData,
                     RedirectAutomatically = settings.RedirectAutomatically,
+                    SupportedTransactionCurrencies = !(choice?.PaymentMethods?.Any() ?? false)
+                        ? null
+                        : choice?.PaymentMethods?.ToDictionary(s => s,
```
ivopashov, line 188: i guess if we got this far, we don't need the Elvis operator. Still refreshing my rusty C# so sorry if i am wrong.

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -311,6 +316,10 @@ public async Task<IActionResult> ContributeToCrowdfund(string appId, ContributeT
                         NotificationURL = settings.NotificationUrl,
                         FullNotifications = true,
                         ExtendedNotifications = true,
+                        SupportedTransactionCurrencies = !(choice?.PaymentMethods?.Any() ?? false)
+                            ? null
+                            : choice?.PaymentMethods?.ToDictionary(s => s,
```
ivopashov, line 321: Ditto.

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
```diff
@@ -21,6 +21,7 @@ public class ItemPrice
             public string Title { get; set; }
             public bool Custom { get; set; }
             public int? Inventory { get; set; } = null;
+            public string[] PaymentMethods { get; set; } = null;
```
ivopashov, line 24: Can we go with an empty array as default and not null so we avoid all the null coalesce operators in the controller?

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -333,7 +333,9 @@ public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
                                  Formatted = Currencies.FormatCurrency(cc.Value.Value, currency)
                              }).Single(),
                     Custom = c.GetDetailString("custom") == "true",
-                    Inventory = string.IsNullOrEmpty(c.GetDetailString("inventory")) ?(int?) null:  int.Parse(c.GetDetailString("inventory"), CultureInfo.InvariantCulture)
+                    Inventory = string.IsNullOrEmpty(c.GetDetailString("inventory")) ?(int?) null:  int.Parse(c.GetDetailString("inventory"), CultureInfo.InvariantCulture),
+                    PaymentMethods = c.GetDetailStringList("payment_methods")
```
ivopashov, line 337: it would help to include some test data. I guess that goes in the `PointOfSaleSettings`.

BTCPayServer/Controllers/AppsPublicController.cs
NicolasDorier: this is super hard to read. I don't understand this code. It would be better if it was broken in 2 tertiary expression.

Kukks:
@ivopashov  Thanks for the review! I've cleaned up the logic a bit and added test coverage

BTCPayServer/Services/Apps/AppService.cs
Kukks: I'd rather not expose this by default. If there is a server without BTC configured, it would make the POS item unable to create invoices.

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
Kukks: Fixed the logic that uses `PaymentMethods` instead.

BTCPayServer/Controllers/AppsPublicController.cs
Kukks: Fixed

Kukks:
> 
> 
> I am kind of new to the project, just cloned it today and that's the first PR i am trying to understand.
> 
> Just want to make sure i understand correctly with the following example:
> 
> If i sell Black coffee and i want to sell it for 1$ in BTC and 2$ in LTC i should duplicate the item with 2 prices (1 and 2) and `payment_methods` of `BTC` and `LTC` respectively. Is that how that is supposed to work?

Yup, that's one usecase for it. :)

NicolasDorier:
test fails

Kukks:
It is only Tor failing now @NicolasDorier

Issue #938, opened 2019-07-22
Make ASCII non-alphanumerical characters valid in a POS item description


Zaxounette:
**_Describe the problem/bug_**
In the PoS app, when editing the text description of a shopping item, it seems the character inputs can only be alphanumerical.

**_Your environment_**
* _Version of BTCPay Server:_ V1.0.3.120
* _Deployment method:_ Lunanode

**_To Reproduce_**
_Steps to reproduce the behavior:_
1. Create a POS app
2. Edit a shopping item
3. Try to input any ASCII non-alphanumerical character (#&@%µ£$§...)
4. Submit changes and get thrown back to general POS app settings (error) instead of POS app list (success).

**_Additional context_**
I have not tested every ASCII non-alphanumerical character, but as an example, I wanted to put a hashtag (#) for a shopping item in the description.
Without the hashtag the submission goes thought, with it, it fails.

bitcoinbrisbane:
Asp.net defends against some characters to stop cross script attacks. It’s most likely that the error is because of that.  Do you have an error message?

Zaxounette:
> Asp.net defends against some characters to stop cross script attacks. It’s most likely that the error is because of that. Do you have an error message?

I might have to rewrite this PR.

1°) The error "Invalid template" appears in fact when I have a backspace in the description (to separate sentences, for instance).

2°) On the other hand, when inputting for example a hashtag + a word ( #example ), update goes through, BUT hashtag does not appear in the UI viewer side.
On top of that, if put in the middle of the description, the hashtag makes the following characters appear blank in the UI viewer side.

No error ever appears on this second part.

bolatovumar:
@Zaxounette the issue seems to be specifically with # and & characters. If you input # character then nothing after this character shows up (so in #&@%µ£$§🤣 nothing will show up). If you input & character then you get &amp; (so &@%µ£$§🤣 --> &amp;@%µ£$§🤣). If you don't use any of these characters then it seems to be fine.

bolatovumar:
@Zaxounette see this screenshot for a workaround:
![Screen Shot 2019-08-21 at 11 30 59 AM](https://user-images.githubusercontent.com/1934678/63415963-397fc200-c407-11e9-9676-bc2e5b887760.png)

If you surround the your product description with single or double quotes down in the "Template" section then it seems to display these characters correctly.

bolatovumar:
[...]

Issue #843, opened 2019-05-15
POS: It's impossible to learn why template is invalid


Kixunil:
**Describe the problem/bug**
When an invalid template is entered into POS settings and saved, it says "Invalid template" without explaining what's invalid about it, so it's impossible to find out why it's invalid.

**Your environment**
* Version of BTCPay Server: commit ad0edb5f4cfa872d959fabcf55d22a3ed29507dc
* Deployment method: manual

**To Reproduce**
1. Go to POS app settings
2. Enter an invalid template
3. Click save
4. See unhelpful error message

**Expected behavior**
When an invalid template is entered, the error message clearly explains why it's invalid.

**Actual behavior**
The message contains no information about the reason why the template is invalid.

pavlenex:
This is a well-known and quite annoying bug that I experienced numerous times. Unfortunately, I was not able to figure out when it happens. Usually, it happens when you add lots of new items or when you enter an inappropriate currency/price format.

I believe nobody was able to figure out exactly when it happens, but the weird thing is if you copy/paste that template entirely into a newly made PoS app, it usually works.

I think it's best to label this as a bug, perhaps @mariodian knows why it happens, but overall it will probably disappear once we refactor the PoS #749

Eskyee:
I can recreate this bug, if any text is in front of the Title* this will create a Invalid template,
I have noticed this lots of times, but just figured it out today. Pav

Eskyee:
<img width="1142" alt="Screenshot 2019-05-15 at 15 25 34" src="https://user-images.githubusercontent.com/30862925/57779068-b7a43c80-7725-11e9-9e13-d33c8b43fab0.png">

dennisreimann:
I talked to @Kukks about this as I was looking for an issue to help with.

IMHO we should question whether it is a good idea to expose the underlying YAML representation.
As the js implementation for handling the data does not use a YAML parser but naive string parsing this is very fragile.
Unless I'm missing something essential I'd say it would be best to hide the textarea via JS (this way users without JS can still edit the perks). This way the YAML representation could be used as an implementation detai. Users with JS can still edit the perks via the UI and do not need the YAML.

Kixunil:
Please do **not** remove the ability to import the data *somehow*, [we just started to rely on it](https://github.com/ParalelnaPolis/pos_articles_conversions). Funnily enough, I got it to work by finding out it's Yaml and using Yaml serializer.

[...]

PR #1544, opened 2020-05-04
Redo App template editor with Vue based app

Kukks:


pavlenex:
closes #1544 
closes #938 
closes #843

Kukks:
@pavlenex  I think i got it

Issue #1537, opened 2020-05-03
Qr codes in Dark mode do not scan well


Kukks:
**Describe the problem/bug**
When using dark mode, the QR codes do not work (LN node info page, Services page, Receive wallet page, Checkout, enable 2fa  aare all the pages where I think QR codes are displayed)

**Your environment**
* Version of BTCPay Server:
Latest
* Deployment method: 
Dev
* Other relevant environment details:

**Expected behavior**
Add a white border like we do for Checkout

PR #1547, opened 2020-05-05
add qrcode margins

chewsta:
addresses #1537 by setting the qrcode properties in ShowLightningNodeInfo and WalletReceive templates the same as Checkout


NicolasDorier:
@chewsta we released your fix in 1.0.4.4

PR #1555, opened 2020-05-07
Try to tell browser that autocomplete is not welcome in generate wallet

### LINKED ISSUE #1553: Credentials autofill in generate wallet
**Describe the problem/bug**
When I generate a wallet from new/existing seed, the browser autofills my login credentials in the seed and passphrase fields. This only happens if you have saved credentials in the browser. It happens for me in chrome and firefox.

**Your environment**
* Version of BTCPay Server: v1.0.4.4
* Deployment method: LunaNode

**To Reproduce**
1. Go to: Stores > Settings 
2. Click on: Modify (derivation scheme)
3. Drop down to: New/Existing seed

**Expected behavior**
Should have empty fields

**Actual behavior**
My login is auto filled in seed and passphrase fields.

**Screenshots/Links**
image

**Additional context**
I guess it's just annoying. Not really a big deal if there isn't a fix.

Kukks:
fixes #1553

image

### DIFF
--- a/BTCPayServer/Views/Stores/AddDerivationSchemes_NBXWalletGenerate.cshtml
+++ b/BTCPayServer/Views/Stores/AddDerivationSchemes_NBXWalletGenerate.cshtml
@@ -24,7 +24,7 @@
 
                 <div class="form-group">
                     <label asp-for="ExistingMnemonic">Existing Seed</label>
-                    <input type="text" asp-for="ExistingMnemonic" class="form-control"/>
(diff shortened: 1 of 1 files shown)


britttttk:
"off" doesn't work but "new-password" does

Kukks:
That's not html compliant though. I fear some password managers will just use that as a form of key and some browsers would just straight out ignore the noncompliant value

britttttk:
type text instead of password?

NicolasDorier:
merging this one as at least it is a bit better than now...

PR #1561, opened 2020-05-09
Refactoring of payjoin to support new spec

NicolasDorier:


BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -115,25 +116,37 @@ public int Compare([AllowNull] UTXO x, [AllowNull] UTXO y)
...
+            bool noadjustfee = false,
+            int feebumpindex = -1,
+            int v = 1)
         {
+            if (v != 1)
+            {
+                return BadRequest(new JObject
+                {
+                    new JProperty("errorCode", "version-unsupported"),
+                    new JProperty("supported", new JArray(1)),
```
Kukks, line 129: You should use the standard method `CreatePayjoinError` for this and just return the versions in the header?

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -146,69 +159,69 @@ public async Task<IActionResult> Submit(string cryptoCode)
...
-                originalTx = psbt.ExtractTransaction();
-            }
 
             async Task BroadcastNow()
             {
                 await _explorerClientProvider.GetExplorerClient(network).BroadcastAsync(originalTx);
             }
 
             var sendersInputType = psbt.GetInputsScriptPubKeyType();
-            if (sendersInputType is null)
```
Kukks, line 176: You removed all checks for different inputs in the original PSBT + p2pkh inputs this way

NicolasDorier: It is on purpose, there is two cases we wanted to consider.

Case 1: The original transaction has mixed inputs
If it has mixed input, then it does not matter whether we use P2WPKH or P2SH, a blockchain analyst can't spot which one is our input.

Case 2: The user is using P2PKH.
We do not support it in BTCPay, it should give a out-of-utxos error but if other implementation decide to support p2pkh that's fine.

PR #1567, opened 2020-05-12
Require loggedin user for docs?

Kukks:
We had talked before that docs should be for authorized users only. We had it in when we had Nswag but must have lost it after we switch to manual swagger files

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -101,9 +101,13 @@ namespace BTCPayServer.Tests
             using (var tester = ServerTester.Create())
             {
                 await tester.StartAsync();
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Test fail. I dropped it when I removed NSwag, because I was relying on the static file provider directly.

I think it does not hurt to be public, because I fixed the redoc page with the SHA512 in the script markup. So a malicious peer can't change the JS.

However, the downside to expose it is that a malicious actor can infer the version of BTCPay Server...
So yeah, better blocking it.

NicolasDorier:
@Kukks that said, it would be nice if we had a public link to share the API.

Maybe later we can do a github page with the latest swagger and just integrate that to the Circle CI.

PR #1570, opened 2020-05-13
See if decoupling wallet is possible

Kukks:
Here is my first very rough attempt at decoupling the wallets from stores.

It seems doable but damn you have to touch a lot of files..

The approach I am taking is around moving the DerivationStrategies column from StoreData to the WalletData ( and merging the LabelColors blob into DerivationSchemeSettings)


NicolasDorier:
@Kukks did not reviewed yet.
You should not move anything. DerivationSchemes of store should stay stored in the store.
What should be abstracted is how to get the DerivationScheme based on the WalletId, some wallet ID being a store's wallet id, other id could be "independant id".

If the user create a wallet separately and then link to the store, the StoreData should have a `LinkedDerivationSchemeSettings`.

So basically there would 2 types of Id: Wallet Store Id (like now) and Independant Wallet Id (new)

For the wallet store Id, there is two cases:
* The wallet is stored in StoreData (like now)
* The wallet is a `LinkedDerivationSchemeSettings` which reference an external wallet.

How to get the DerivationScheme from a wallet id should be abstracted away.

It also mean that a linked derivation scheme can have several wallet id: The Store's wallet id format, and the independant derivation scheme ID. We can't prevent this because our permission's framework is using store's WalletId to represent the meaning "Wallet of the store".

PR #1583, opened 2020-05-19
Refactor server info data structure

NicolasDorier:
Refactoring a bit @dennisreimann PR on the simplification of data structures.


NicolasDorier:
Just to explain: I make a difference in sync between what the internal full node state, and what NBXplorer state.

I also removed a useless level in the JSON, and applied nullable where needed.

PR #1591, opened 2020-05-22
Remove dependency on BTCPayServer.Common for BTCPayServer.Rating

Kukks:


BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -119,6 +121,8 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
             services.TryAddSingleton<BTCPayNetworkProvider>(o => 
             {
                 var opts = o.GetRequiredService<BTCPayServerOptions>();
+                CurrencyPair.AdditionalCurrencies.AddRange(opts.NetworkProvider.UnfilteredNetworks.GetAll()
```
NicolasDorier, line 124: there is no guarantee that a class who need `CurrencyPair` is not created before the first `BTCpayNetworkProvider`.
If we go this route, I would prefer we have a `CurrencyPairFactory`.

Kukks: Yeah, this is tricky. I'm not sure what you mean though. It is quite hard to refactor this code since `CurrencyPair` is used in static contexts and json converters.

NicolasDorier: Ok so why not just hard coding the currencies inside CurrencyPair?

Kukks: We have to do this for both `CurrencyPair` and `CurrencyValue` though (in the other PR) though.. Not the cleanest solution but might just be best. Maybe we can just extend the currencies.txt file

NicolasDorier: @Kukks I don't think this is that bad. This is not something that often change.

Moreover, if you are worried, you can write a Fast test which just check that all the possible networks are in CurrencyValue and CurrencyPair.

PR #1596, opened 2020-05-24
Randomize RBF if the user does not care

NicolasDorier:
NBXplorer will start randomizing the fingerprint of transactions.
To accomodate this change, I removed the checkbox in advanced settings of the wallet and replaced by a tri state.

image

### DIFF
--- a/BTCPayServer/Controllers/WalletsController.PSBT.cs
+++ b/BTCPayServer/Controllers/WalletsController.PSBT.cs
@@ -40,7 +40,10 @@ namespace BTCPayServer.Controllers
            
             if (network.SupportRBF)
             {
-                psbtRequest.RBF = !sendModel.DisableRBF;
+                if (sendModel.AllowFeeBump is WalletSendModel.ThreeStateBool.Yes)
+                    psbtRequest.RBF = true;
+                if (sendModel.AllowFeeBump is WalletSendModel.ThreeStateBool.No)
+                    psbtRequest.RBF = false;
             }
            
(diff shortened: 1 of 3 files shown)


pavlenex:
Strange wording imo, it's unclear what `I don't care` will do. Perhaps `Random` or whatever it actually does?

Kukks:
I would also add (RBF) to the label with a link to some docs somewhere

NicolasDorier:
I search for some doc about RBF but nothing is useful... :(

NicolasDorier:
@pavlenex I am not fan of "Random". I would prefer something like "Whichever". I wanted to use a three state checkbox but that is pain in the ass to code.

NicolasDorier:
What about, "Undecided, privacy first"

NicolasDorier:
what about `Randomize for higher privacy`

NicolasDorier:
Added link to https://bitcoinops.org/en/rbf-in-the-wild/

pavlenex:
> 
> 
> what about `Randomize for higher privacy`

Of all suggested this one sounds the best. However it's still unclear what's happening behind the scenes. To me it _sounds_ like if RBF is `random` then sometimes it'll be enabled and sometimes disabled? That link form optech is good, but it somehow does not even explain RBF it tackles it more from a UI and usability perspective.

NicolasDorier:
Let's wait I add RBF support for the wallet for the doc. Most is done, only UI left.

PR #1597, opened 2020-05-24
Refactor parameter passing in wallet functions

NicolasDorier:
In addition we use the suggestion of NBXplorer about whether or not enforcing low R signature.


NicolasDorier:
ping @Kukks I think you will like the refactoring of the SigningContext.

PR #1602, opened 2020-05-25
Rename Coldcard to Electrum/Airgapped hardware wallets

Kukks:
firefox_lIKlxJDjwp
firefox_I6AvCmmNkl

Also adds support to read the xpub from the file directly ( Cobo Vault support)


NicolasDorier:
should we rename `... a hardware wallet` to `a USB hardware wallet` ?

Eskyee:
Since you can have Bluetooth as well as USB hardware wallet!! I think this maybe clearer 

external hardware wallet 

This sounds more future proof. and generic I think 🤔

NicolasDorier:
Merging that for now. I think it deserve better naming though.

PR #1604, opened 2020-05-25
Clarify/clean up warning shown when adding a lightning node

janoside:
Simple text cleanup.


janoside:
This is a text-only change and I don't think it's responsible for the test failures (would be glad to have someone confirm). The details of the failure:

    Not found: server/services/ssh
    In value:  <html lang="en"><head>
    ...

PR #1605, opened 2020-05-26
Point of Sale Refactoring

### LINKED ISSUE #749: Pos App refactoring
So right now, the POS app splits into two:
the static item mode, where you pick an item and click buy and it creates the invoice
cart mode, which lets you pick multiple items, charge tips, allows donations, etc
The code for both of these are in one html file and it is a bit messy in there. The way we determine which is used is through a config setting in the POS app. I propose we remove that configuration, split these into separate views and allow any pos to use either view at any time through. 
So for example: 
* btcpay.com/apps/xyz/pos/static -> static item mode
* btcpay.com/apps/xyz/pos/cart -> cart mode
* btcpay.com/apps/xyz/pos/keypad-> keypad mode( no items, just an on-screen keyboard and an input field; people seem to really love this for some reason)

We can try and make the code between all the apps as reusable as possible between each view. I would say the static view doesn't need any js at all since it's, well, static.

Additionally, we can replace the "EnableShoppingCart" config item with a "DefaultView" which can be either of the views we create. This way `btcpay.com/apps/xyz/pos/` redirects to it.

dennisreimann:
As described in #749 and talked about with @kukks.

This is mostly UI related as it separates the simple static and cart views.
Replaces the `EnableShoppingCart` setting with a `DefaultView`, which can be set to `Static` or `Cart`.

Also gives the ability to define the view type via an optional URL parameter `[Route("/apps/{appId}/pos/{viewType?}")]`, falling back to the default view defined in the settings.

These are the things I'd appreciate input on:

- How to ensure we are backwards compatible for apps that had `EnableShoppingCart` set?
- Is there a better way to [solve this](https://github.com/btcpayserver/btcpayserver/pull/1605/files#diff-c8b2defbc97458f2f445a7ca48187735R41)? `Html.GetEnumSelectList<PosViewType>().Where(type => Convert.ToInt32(type.Value) >= 0)`

Closes #749.


dennisreimann:
The test failure seems unrelated, can someone retrigger the selenium tests? 

`OpenQA.Selenium.WebDriverException : Cannot start the driver service on http://localhost:45594/`

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -106,7 +108,7 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
-            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && !settings.EnableShoppingCart)
+            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && settings.DefaultView != PosViewType.Cart)
```
Kukks, line 111: This should be loaded from the route parameter and then falls back to the setting


```csharp
settings.DefaultView = settings.EnableShoppingCart? PosViewType.Cart : settings.DefaultView ;
viewType ?= settings.DefaultView;
```

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -141,14 +143,14 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
             }
             else
             {
-                if (!settings.ShowCustomAmount && !settings.EnableShoppingCart)
+                if (!settings.ShowCustomAmount && settings.DefaultView != PosViewType.Cart)
```
Kukks, line 146: same as above

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -141,14 +143,14 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
...
                     return NotFound();
                 price = amount;
                 title = settings.Title;
                 
                 //if cart IS enabled and we detect posdata that matches the cart system's, check inventory for the items
                 if (!string.IsNullOrEmpty(posData) && 
-                    settings.EnableShoppingCart &&
```
Kukks, line 151: same as above

BTCPayServer/Views/Apps/UpdatePointOfSale.cshtml
```diff
@@ -35,10 +36,10 @@
                         <span asp-validation-for="Currency" class="text-danger"></span>
                     </div>
                     <div class="form-group">
-                        <div class="form-check">
-                            <input asp-for="EnableShoppingCart" type="checkbox" class="form-check-input" />
-                            <label asp-for="EnableShoppingCart" class="form-check-label"></label>
-                            <span asp-validation-for="EnableShoppingCart" class="text-danger"></span>
+                        <div class="form-group">
+                            <label asp-for="DefaultView" class="control-label"></label>*
+                            <select asp-for="DefaultView" asp-items="@Html.GetEnumSelectList<PosViewType>().Where(type => Convert.ToInt32(type.Value) >= 0)" class="form-control"></select>
```
Kukks, line 41: No need to limit, I would allow user to select unspecified anyway.

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
```diff
@@ -36,7 +33,9 @@ public class CurrencyInfoData
 
         public CurrencyInfoData CurrencyInfo { get; set; }
 
-        public bool EnableShoppingCart { get; set; }
+        public PosViewType ViewType { get; set; }
+
+        public bool IsCartView { get { return ViewType == PosViewType.Cart; } }
```
Kukks, line 38: no need for a property, just do `ViewType == PosViewType.Cart` in the view

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
```diff
@@ -81,429 +79,12 @@
...
-                                }
-                            </div>
-                        </div>
-                    }
-                </div>
-            </div>
-        </div>
-    }
+    
+    <partial name="./PointOfSale/@Model.ViewType" />
```
Kukks, line 88: Instead of loading them as a partial, why not tell the action to load the specific view directly. We can avoid having this file altogether

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -106,7 +108,7 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
-            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && !settings.EnableShoppingCart)
+            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && settings.DefaultView != PosViewType.Cart)
             {
                 return RedirectToAction(nameof(ViewPointOfSale), new { appId = appId });
```
Kukks, line 113: Provide `viewType` as parameter

BTCPayServer/Controllers/HomeController.cs
```diff
@@ -69,7 +69,7 @@ private async Task<ViewResult> GoToApp(string appId, AppType? appType)
                         var controller = (AppsPublicController)serviceProvider.GetService(typeof(AppsPublicController));
                         controller.Url = Url;
                         controller.ControllerContext = ControllerContext;
-                        var res = await controller.ViewPointOfSale(appId) as ViewResult;
+                        var res = await controller.ViewPointOfSale(appId, PosViewType.Unspecified) as ViewResult;
```
Kukks, line 72: This doesn't look right?

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
dennisreimann: Then we'd have to duplicate some things from the layout file, which we can share this way. But maybe it's just me not knowing al available options here :)

BTCPayServer/Views/Apps/UpdatePointOfSale.cshtml
dennisreimann: imho this wouldn't make sense: The reason I introduced the `Unspecified` type is that I needed a way to handle the `/apps/{appId}/pos/` (without viewType) route. Another way I see would be to have it as a separate route which redirects to the route with the `DefaultView` – then we could get rid of `Unspecified` alltogether.

dennisreimann: Updated it using the described redirect approach.

NicolasDorier:
Build fail

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
Kukks: You can create a `_PosLayout` file and tll the razor pages to use it `Layout="PosLayout"` at the top

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -38,8 +38,8 @@ public class AppsPublicController : Controller
         private readonly AppService _AppService;
         private readonly BTCPayServerOptions _BtcPayServerOptions;
         private readonly InvoiceController _InvoiceController;
-        private readonly UserManager<ApplicationUser> _UserManager;
-
+        private readonly UserManager<ApplicationUser> _UserManager;
```
Kukks, line 41: nit: spacing :P

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -49,15 +49,28 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
-
+            PosViewType viewType = settings.DefaultView;
```
Kukks, line 52:
```suggestion
            PosViewType viewType = settings.EnableShoppingCart? PosViewType.Cart : settings.DefaultView;
```

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -106,9 +120,9 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
             if (app == null)
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
```
Kukks, line 122:
```suggestion
            var settings = app.GetSettings<PointOfSaleSettings>();
            settings.DefaultView = settings.EnableShoppingCart? PosViewType.Cart : settings.DefaultView;
```

BTCPayServer/Views/AppsPublic/ViewPointOfSale.cshtml
dennisreimann: Ok, thanks!

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -49,11 +49,11 @@ public async Task<IActionResult> ViewPointOfSale(string appId)
...
-            PosViewType viewType = settings.DefaultView;
-            
+            PosViewType viewType = settings.EnableShoppingCart? PosViewType.Cart : settings.DefaultView;
+
             return RedirectToAction(nameof(ViewPointOfSale), new { appId, viewType });
         }
-        
+
         [HttpGet]
         [Route("/apps/{appId}/pos/{viewType}")]
```
Kukks, line 58: viewType should be optional and if null: `viewType ?= settings.EnableShoppingCart? PosViewType.Cart : settings.DefaultView`

BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -38,8 +38,8 @@ public class AppsPublicController : Controller
         private readonly AppService _AppService;
         private readonly BTCPayServerOptions _BtcPayServerOptions;
         private readonly InvoiceController _InvoiceController;
-        private readonly UserManager<ApplicationUser> _UserManager; 
-        
+        private readonly UserManager<ApplicationUser> _UserManager;
+
         [HttpGet]
         [Route("/apps/{appId}/pos")]
```
Kukks: This action is not needed if https://github.com/btcpayserver/btcpayserver/pull/1605/files/58f56eac9034c8be31a1aa294bb5e2d819a55fa8..559015c70dc25d3ef4155e44083d64a091805ff0#r431649223 is done

BTCPayServer/Controllers/AppsPublicController.cs
Kukks, line 58:
```suggestion
        [Route("/apps/{appId}/pos/{viewType?}")]
```

Kukks: `PosViewType viewType` should be nullable: `PosViewType? viewType`

NicolasDorier:
@pavlenex  you want to manually test?

pavlenex:
tACK. I tested a bit and haven't found any problems.

PR #1612, opened 2020-05-29
Api/swaggerln

Kukks:


BTCPayServer/wwwroot/swagger/v1/swagger.template.lightning.common.json
```diff
@@ -4,41 +4,35 @@
...
+                                "nullable": true
+                            },
+                            "nodeHost": {
+                                "type": "string",
+                                "nullable": true
+                            },
+                            "nodePort": {
+                                "type": "integer",
+                                "format": "int32"
+                            }
```
NicolasDorier, line 31: don't use 3 parameters a node id is normally consiered having the format "pubkey@endpoint:port", no need to break it up in 3 parameters.

PR #1615, opened 2020-05-30
Always use second based TimeSpan in Greenfield

NicolasDorier:


NicolasDorier:
mmh actually this PR also change the behavior.
Like if the fields are not specified, I don't update them. We already talked about it.

@Kukks do you really have strong opinion on that? Personally if I call an update without specifying some fields, I don't expect them to be changed.

NicolasDorier:
Example of slack API that follow the rule "if unset, ignore it"

https://api.slack.com/methods/users.profile.set

NicolasDorier:
@Kukks outside the partial update that I will revert from this PR, does it seem good?

Kukks:
> @Kukks outside the partial update that I will revert from this PR, does it seem good? 


I think so, don't have a dtrong opinion on which time format to use unless it is the old shitty asp.net Json one 


I'm wondering if it's the correct format to use

NicolasDorier:
@Kukks we need to see how to handle null value.

```csharp
JsonConvert.DeserializeObject<CreateStoreRequest>("{ \"invoiceExpiration\": null}")
```

Give InvoiceExpiration equal to 0.
Tried to use DefaultValueAttribute, but sadly it does not work nice with json converters.

NicolasDorier:
@Kukks this handle default value as expected in any cases.

```csharp
        [JsonConverter(typeof(TimeSpanJsonConverter))]

        public TimeSpan InvoiceExpiration { get; set; } = TimeSpan.FromMinutes(15);
```

PR #1616, opened 2020-05-30
Make create store only about store creation

NicolasDorier:


Kukks:
In terms of REST, this is wrong. The resource endpoint should try to be consistent with the same model.

AFAIK, you can probably send a json body with just the name and it would still work with the current api

NicolasDorier:
The three reasons I have is:
* We should try to mirror the UI as much as possible.
* Later the create store could have new parameter only available at creation time (for example `copyFromStore`)
* If somebody wants to update their store in a certain way during creation, they can already do with a second call.

NicolasDorier:
There is also another point:

If you really wanted to be pure REST, the POST for creation, PUT for update. Both should be at the same endpoint (the id should not be in the URL).

Kukks:
> 
> 
> There is also another point:
> 
> If you really wanted to be pure REST, the POST for creation, PUT for update. Both should be at the same endpoint (the id should not be in the URL).
> We could do that too...

That sounds very wrong. If anything the `id` should not be in the body but only in the url.

A lot of my rest ideas were inspired from this many years ago:
https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#restful

PR #1619, opened 2020-05-31
[Greenfield] Document store API

NicolasDorier:
This PR is fixing a number of issue over default in the store API.

It is also documenting all the fields.

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoresController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoresController.cs
@@ -128,7 +128,7 @@ namespace BTCPayServer.Controllers.GreenField
                 CustomLogo = storeBlob.CustomLogo,
                 CustomCSS = storeBlob.CustomCSS,
                 HtmlTitle = storeBlob.HtmlTitle,
-                AnyoneCanInvoice = storeBlob.AnyoneCanInvoice,
+                AnyoneCanCreateInvoice = storeBlob.AnyoneCanInvoice,
                 LightningDescriptionTemplate = storeBlob.LightningDescriptionTemplate,
                 PaymentTolerance = storeBlob.PaymentTolerance,
                 RedirectAutomatically = storeBlob.RedirectAutomatically,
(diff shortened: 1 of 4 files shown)


BTCPayServer/Controllers/GreenField/StoresController.cs
```diff
@@ -128,7 +128,7 @@ private static Client.Models.StoreData FromModel(Data.StoreData data)
                 CustomLogo = storeBlob.CustomLogo,
                 CustomCSS = storeBlob.CustomCSS,
                 HtmlTitle = storeBlob.HtmlTitle,
-                AnyoneCanInvoice = storeBlob.AnyoneCanInvoice,
+                AnyoneCanCreateInvoice = storeBlob.AnyoneCanInvoice,
```
Kukks, line 131: If we're renaming this, maybe `PublicInvoiceEndpointEnabled`?

BTCPayServer/wwwroot/swagger/v1/swagger.template.stores.json
```diff
@@ -306,58 +304,79 @@
...
+                        "description": "If true, the checkout page will ask to enter an email address before accessing payment information."
                     },
                     "networkFeeMode": {
                         "$ref": "#/components/schemas/NetworkFeeMode"
                     },
                     "payJoinEnabled": {
-                        "type": "boolean"
+                        "type": "boolean",
+                        "default": false,
+                        "description": "If true, payjoin will be proposed in the checkout page if possible. ([More information](https://docs.btcpayserver.org/Payjoin/#btcpay-server-payjoin-guide))"
```
Kukks, line 374: does that link actually work in redoc?

NicolasDorier: it does! redoc is awesome

BTCPayServer/Controllers/GreenField/StoresController.cs
NicolasDorier: I find it more confusing.

PR #1620, opened 2020-05-31
[GreenField] Add additional data to allow roundtrip of upper version fields

NicolasDorier:
Ping @rockstardev @Kukks in the future, think about adding this field to JSON data that may have roundtrips.

Imagine a store in btcpayserver V1 have only a field "name" and "defaultLang" and btcpayserver client lib is also V1.

The client wants to change the name
First the client would do a GET
```json
{
  "name": "Hello",
  "defaultLang": "fr"
}
```

Then client would do a POST
```json
{
  "name": "NEW NAME",
  "defaultLang": "fr"
}
```

Then imagine the server update to V2, and a new field, enablePayjoin is added, but the client library is still V1. Without the AdditionalData field, here what would happen.

GET
```json
{
  "name": "Hello",
  "defaultLang": "fr",
  "payjoinEnabled": true
}
```

During deserialization, without AdditionalData, `payjoinEnabled` is simply thrown away. So when the client is running the same code than before on V2, his update on the name will actually disable payjoinEnabled because this default to `false`!

POST
```json
{
  "name": "NEW NAME",
  "defaultLang": "fr"
}
```

Adding AdditionalData make sure that the payjoinEnabled value is correctly roundtripping even for V1 clients.

### DIFF
--- a/BTCPayServer.Client/Models/PaymentRequestBaseData.cs
+++ b/BTCPayServer.Client/Models/PaymentRequestBaseData.cs
@@ -1,6 +1,8 @@
 using System;
+using System.Collections.Generic;
 using BTCPayServer.JsonConverters;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Client.Models
 {
@@ -17,5 +19,8 @@ namespace BTCPayServer.Client.Models
         public string EmbeddedCSS { get; set; }
         public string CustomCSSLink { get; set; }
         public bool AllowCustomPaymentAmounts { get; set; }
+
+        [JsonExtensionData]
+        public IDictionary<string, JToken> AdditionalData { get; set; }
     }
 }
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using BTCPayServer.Client.JsonConverters;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Client.Models
 {
@@ -54,6 +56,9 @@ namespace BTCPayServer.Client.Models
         public bool PayJoinEnabled { get; set; }
         public bool LightningPrivateRouteHints { get; set; }
 
+
+        [JsonExtensionData]
+        public IDictionary<string, JToken> AdditionalData { get; set; }
     }
     
     public enum NetworkFeeMode


rockstardev:
Wouldn't server update the path for v2? Then if you are running v1 and v2 in parallel on the server - you would have specific handling on how POSTs to v1 are maintaining compatibility without requiring clients to implement anything (definitely not this special behavior of AdditionalFields serialization)

NicolasDorier:
@rockstardev changing route version is useful in case of complete breaking change of the API.

Adding a field should not be a breaking change. If every field added is a breaking change, this will be unmaintainable as we would need 1 separate class every times we are adding a new field + one additional route populating the additional field with a default value from the old entity.

NicolasDorier:
Note that this change has only an impact on the client library, this has no impact server side.

rockstardev:
That's the core problem for me - with all my experience with REST APIs, I've never seen one where you push this type of functionality onto client libraries. If there is some standard/docs that inspired you to go in this direction, link it.

From your follow up messages - I understand now that there won't be v1 or v2... or even v1.0 and v1.1... rather you prefer not to version functionality. You want to add new field to existing version and then all clients should keep working. 

Considering all that, you would want to keep the scope of change on the server without depending on the clients. You already have ability to detect unset fields, so for these type of changes you would want server to maintain full control and fill in the gaps not depending on the client to "send the right data".

My advice is that you go in other direction - keep the scope of change on the server and think more about versioning the clients. I'm not a fan of client version headers... but with your preference of backwards compatibility and approach you are describing they would be useful. They would allow server to fill in the gaps differently depending on the version. Think about that approach rather than adding `AdditionalData` to endpoints.

NicolasDorier:
> so for these type of changes you would want server to maintain full control and fill in the gaps not depending on the client to "send the right data".

That's actually a reason why I was advocating for partial update over full update. Partial update would just ignore fields that are not set, but full update, as we do now, consider all fields unset, set to default value.

> I'm not a fan of client version headers... but with your preference of backwards compatibility and approach you are describing they would be useful.

I think we have the version in path already.

>  They would allow server to fill in the gaps differently depending on the version.

So there is some downside to it. If all field addition need a new version for the route.

1. If for some reason you can't update the library, but want the new feature, you can't simply add the field not in the library in the AdditionalData dictionary. The library need to update the version number of the route as well.
2. Everybody updating the library will have to think about bumping the route version on top of adding the new field. Since most dev don't test such simple change as adding a field, I guess there will be surprises there.

I never saw bumping a API endpoint just for parameter addition. (I know that since the day I passed my WCF SOAP services MS certification :D)

https://restfulapi.net/versioning/

rockstardev:
It seems we're completely disconnected here.

> That's actually a reason why I was advocating for partial update over full update. Partial update would just ignore fields that are not set, but full update, as we do now, consider all fields unset, set to default value. I lost the battle though.

Not sure I understand what kind of decision was made... but this is why you would have different methods for creating (POST) and updating partial (PATCH) and updating full (PUT).

> So there is some downside to it. If all field addition need a new version for the route.

Field addition DOES NOT require new version in route if you code it in backwards compatible matter. But obviously you need to provide behavior for legacy clients that didn't update. How you do that - it's up to implementation... again this DOESN'T require that you immediately go toward two separate routes/methods if you have facility to detect that field was not set by client (which we have).

I see that PR is merged regardless - so I'll stop adding feedback. If you or @Kukks want my input on anything API related, feel free to DM on Mattermost.

NicolasDorier:
@rockstardev I think you misunderstand what this PR is doing.

`Field addition DOES NOT require new version in route if you code it in backwards compatible matter.`

This is exactly our case.

PR #1592, opened 2020-05-22
Remove dep currencynametable

Kukks:
closes #1571 
Builds on top of #1591 so review that one first

PR #1622, opened 2020-05-31
Remove dependency on Common from Rating

### LINKED ISSUE #1571: Refactor CurrencyNameTable
There is a `CurrencyValue` type in `StoreBlob`, which needs a `CurrencyValueJsonConverter` which needs `CurrencyNameTable` which needs `BtcPayNetworkProvider`. Safe to say this is a clusterfuck. 
I was trying to finish the Store GreenField API but this piece requires moving and refactoring a lot of related logic.

Kukks:
replaces #1591 and #1592 
closes #1571 

Replaces the Currencies.txt to a json format which includes the networks too ( there is also a test to check now) Also makes CurrencyPair not depend on network and instead use CurrencyNameTable


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -2363,7 +2377,7 @@ public async Task CanUsePoSApp()
                         ExpectedThousandSeparator: ",", ExpectedPrefixed: true, ExpectedSymbolSpace: true),
                     (Code: "JPY", ExpectedSymbol: "¥", ExpectedDecimalSeparator: ".", ExpectedDivisibility: 0,
                         ExpectedThousandSeparator: ",", ExpectedPrefixed: true, ExpectedSymbolSpace: false),
-                    (Code: "BTC", ExpectedSymbol: "BTC", ExpectedDecimalSeparator: ".", ExpectedDivisibility: 8,
+                    (Code: "BTC", ExpectedSymbol: "Ƀ", ExpectedDecimalSeparator: ".", ExpectedDivisibility: 8,
```
NicolasDorier, line 2380: Ƀ is not bitcoin symbol. It should be ₿

NicolasDorier:
Seems good, but the symbol for BTC must be fixed.

PR #1627, opened 2020-06-03
Create GreenField Api Development Docs

Kukks:


docs/greenfield-development.md
```diff
@@ -0,0 +1,57 @@
+# GreenField API Development Documentation
+## Adding new API endpoints
+
+* Always document all endpoints and model schemas in swagger. OpenAPI 3.0 is used as a specification, in JSON formatting, and is written manually. The specification is split to a file per controller and then merged by the server through a controller action at  `/swagger/v1/swagger.json`.
+* All `JsonConverter` usage should be registered through attributes within the model itself.
```
dennisreimann, line 5: This refers to the client, correct? Maybe we can add a *General* section that explains what needs to be considered in general (controller, client, docs, tests).

Kukks: It applies to both. Since we use the client models directly in the api, we only need to apply once.

Yes, I originally, wanted to do  a checklist but ended up with these general pointers for now

docs/greenfield-development.md
```diff
@@ -0,0 +1,57 @@
...
+### Scenario 3: Adding a property on the model
+Adding a property on a model can potentially be a breaking change. It is a breaking change if:
+* the property is required.
+* the property has no default value.
+
+#### Solutions
+*  Check if the payload has the property present. If not, either set to the default value (in the case of a`POST`) or set to the model's current value. See [Detecting missing properties in a JSON model](#missing-properties-detect) for how to achieve this.
+
+#### Alternatives considered
+* Bump the version of the endpoint.
```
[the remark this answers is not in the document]
Kukks: I think it is on a case-by-case basis. With how current discussions have gone, we seem to be skipping minor version bumps and ensuring compatibility on the same url.

docs/greenfield-development.md
```diff
@@ -0,0 +1,57 @@
...
+
+#### Alternatives considered
+* Create a `JsonConverter` that allows conversion between the original type and the new type. However, if this option is used, you will need to ensure that the response model returns the same format. In the case of the `GET` endpoint, you will break clients expecting the original type.
+
+### Scenario 2: Removing a property on the model
+Removing a property on a model is a breaking change. 
+
+#### Solutions
+* Bump the version of the endpoint.
```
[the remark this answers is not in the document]
Kukks: Yeah, needs another bigger section elaborating on version bump procedure. Will keep for a separate PR as I mostly want to get the discussions around the solutions for non-version bumps settled

docs/greenfield-development.md
```diff
@@ -0,0 +1,57 @@
...
+* `decimal` and `long` and other similar types should be serialized to a string and able to deserialize from the original type and a string.
+* Ensure that the correct security permissions are set on the endpoint. Create a new permission if none of the existing ones are suitable.
+* Use HTTP methods according to REST principles when possible. This means:
+  * `POST` - Create or custom action
+  * `PUT` - Update full model
+  * `PATCH` - Update partially
+  * `DELETE` - Delete or Archive
+* When returning an error response, we should differentiate from 2 possible scenarios:
+  * Model validation - an error on the request was found - [Status Code 422](https://httpstatuses.com/422) with the model [ValidationProblemDetails](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.validationproblemdetails?view=aspnetcore-3.1).
+  * Generic request error - an error resulting from the business logic unable to handle the specified request - [Status Code 400](https://httpstatuses.com/400) with the model [ProblemDetails](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails?view=aspnetcore-3.1).
```
NicolasDorier, line 15: An example: If you call Connect on the lightning API, and the peer does not exists, what error do you send back?

Note that such error should be parseable by the client (say having a specific code in the JSON) as the client may want to treat such error differently.

Kukks: It should be a `400` status code with a  `ProblemDetails` ( not `ValidationProblemDetails`, its parent class ). 
The `ProblemDetails` and `ValidationProblemDetails` are great for our use-case as it allows us to add error codes too

NicolasDorier: I see, we need to fix the API before next release then.

Kukks:
#1628 does this

docs/greenfield-development.md
Kukks: #1628 does this

docs/greenfield-development.md
```diff
@@ -0,0 +1,72 @@
+
+# GreenField API Development Documentation
+## Adding new API endpoints
+
+* Always document all endpoints and model schemas in swagger. OpenAPI 3.0 is used as a specification, in JSON formatting, and is written manually. The specification is split to a file per controller and then merged by the server through a controller action at  `/swagger/v1/swagger.json`.
+* All `JsonConverter` usage should be registered through attributes within the model itself.
+* `decimal` and `long` and other similar types should be serialized to a string and able to deserialize from the original type and a string.
```
NicolasDorier, line 7: No, this depends. If you know the data will not have precision or overflow issue, it should use decimal and long.

docs/greenfield-development.md
```diff
@@ -0,0 +1,72 @@
...
+* the property is required.
+* the property has no default value.
+
+#### Solutions
+*  Check if the payload has the property present. If not, either set to the default value (in the case of a`POST`) or set to the model's current value. See [Detecting missing properties in a JSON model](#missing-properties-detect) for how to achieve this.
+
+#### Alternatives considered
+* Bump the version of the endpoint.
+* Assume the property is always sent and let the value be set to the default if not ( in the case of nullable types, this may be problematic when calling update endpoints). 
+* Use [`[JsonExtensionData]AdditionalData`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonExtensionDataAttribute.htm) so that clients receive the full payload even after updating only the server. This is problematic as it only fixes clients which implement this opinionated flow (this is not a standard or common way of doing API calls) .
```
NicolasDorier, line 63: Note that this must be ALWAYS used for endpoints where the user is expected to do a GET modify the data, followed by a POST.

NicolasDorier: Note that javascript implicitely do that.

Kukks: I really, really disagree with this. I thought we agreed that we would handle new optional properties through json prop detection and setting to default if not there?

NicolasDorier: Yeah, this would mean that javascript clients will behave differently than .NET or Java client who forgot `AdditionalData`

NicolasDorier: By default .NET and Java client remove json properties they don't know. Not Javascript.

Kukks: Yes, I'm still with @rockstardev  for removing `AdditionalData` though. It brings uncertainty to the mix. If they want to edit a new property in a new version, they should update their client

Kukks: That's fine, we should let each client deal with it on its own. We will have special server logic for when we add new properties without wanting to version bump the endpoint (detect missing json prop, fill with default or current value)

Kukks: > 
> 
> By default .NET and Java client remove json properties they don't know. Not Javascript.

Most times, for any client, the payload is reconstructed because of some user-interaction:
* Load data from GET
* create UI view model from data
* user edits view model
* build data model
* POST data

This is agnostic of language implementation details

rockstardev: Just to chime in to say I agree with @Kukks on this. I wouldn't get into `AdditionalData` on client because it's conceptually completely different than what Javascript is doing. 

If you can emulate the way Javascript does it on client - then go for it... have `AdditionalData` on client and make it generate JSON which includes property from `AdditionalData`. But definitely do not send `AdditionalData` over to server.

Kukks:
Let's try and get this merged before any more API work @NicolasDorier

Kukks:
@NicolasDorier pingeroo

britttttk:
Any resolution on this? Awesome doc :) ping @Zaxounette we should link this somewhere in Contribute docs once it's merged.

Kukks:
@NicolasDorier  pingaboo

PR #1628, opened 2020-06-03
GreenField: Handle status codes, error models consistently

Kukks:


BTCPayServer/Controllers/GreenField/LightningNodeApiController.cs
```diff
@@ -204,17 +199,13 @@ public virtual async Task<IActionResult> PayInvoice(string cryptoCode, PayLightn
             var result = await lightningClient.Pay(lightningInvoice.Invoice);
             switch (result.Result)
             {
-                case PayResult.Ok:
-                    return Ok();
                 case PayResult.CouldNotFindRoute:
-                    ModelState.AddModelError(nameof(lightningInvoice.Invoice), "Could not find route");
-                    break;
+                    return this.GetGeneralErrorResponse("Could not find route");
```
NicolasDorier, line 203: Please keep the enum, else the client will have to do some string parsing.

PR #1630, opened 2020-06-03
Pay endpoint: if JsonResponse is true, send error in json too

### LINKED ISSUE #1532: Invoice API returns HTTP 200 for creation failure
**Describe the problem/bug**
I have a script that creates invoices by making a POST request to my btcpayserver at /api/v1/invoices

I noticed that no invoices had been created in a long time, but didn't see errors in any logs.

Upon digging into the responses from btcpayserver the problem appears to be that the error is only returned in the HTML, not as a part of the HTTP response.

The root cause of the problem was a failed exchange rate provider. Perhaps consider using multiple fallback exchange rate APIs to eliminate this as a single point of failure?

**Your environment**
* Version of BTCPay Server: v1.0.4.2
* Deployment method: Docker
* Other relevant environment details: Ubuntu

**Logs (if applicable)**
No server log entries generated.

**Setup Parameters**
Nothing noteworthy

**To Reproduce**
1. Set store's exchange rate provider to NDAX
2. Make a POST request to create an invoice
3. Note that the server returns a 200 OK response

**Expected behavior**
An HTTP error code response; perhaps a 500?

**Actual behavior**
Server returns an HTTP 200 OK and an HTML response to my POST request that includes:

```
<section>
    <div class="container">
        <div class="row">
            <div class="col-lg-12">
                <h2>Pay Button request failed</h2>
                Please fix following errors:
                <ul>
                        <li>4/30/2020 7:16:33 PM: Creation of invoice starting&#xA;4/30/2020 7:16:33 PM: BTC_USD: The rating rule is ndax(BTC_USD)&#xA;4/30/2020 7:16:33 PM: BTC_USD: The evaluated rating rule is ERR_RATE_UNAVAILABLE(ndax, BTC_USD)&#xA;4/30/2020 7:16:33 PM: BTC_USD: Rate rule error (RateUnavailable)&#xA;</li>
                </ul>
            </div>
        </div>
    </div>
</section>
```

Kukks:
fixes #1532

### DIFF
--- a/BTCPayServer/Controllers/PublicController.cs
+++ b/BTCPayServer/Controllers/PublicController.cs
@@ -75,6 +75,11 @@ namespace BTCPayServer.Controllers
             catch (BitpayHttpException e)
             {
                 ModelState.AddModelError("Store", e.Message);
+                if (model.JsonResponse)
+                {
+                    return BadRequest(ModelState);
+                }
+
                 return View();
             }
             


NicolasDorier:
This is not a bug, this is a not an API but a HTML page.

Kukks:
> This is not a bug, this is a not an API but a HTML page.

Yes, but there is already a flag to return Json as a result, it was just not respected for errors.

PR #1633, opened 2020-06-04
Add Onion-Location HTTP header

### LINKED ISSUE #1626: Add Onion-Location HTTP header
Starting with the release of [Tor Browser 9.5](https://blog.torproject.org/new-release-tor-browser-95), websites can have their alternate .onion addresses advertised to Tor desktop users who have the 'Onion Location' option enabled.

Sites that add the .onion address advertisement HTTP header can prompt their visitors to switch to a version delivered using the Onion service for improved security.

Would be nice to add the `Onion-Location` HTTP header and automatically propose .onion version of BTCPay dashboard/store/checkout

dennisreimann:
Closes #1626.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -3607,6 +3607,36 @@ normal:
             }
         }
 
+        [Fact(Timeout = TestTimeout)]
+        [Trait("Integration", "Integration")]
+        public async void CheckOnionlocationForNonOnionHtmlRequests()
+        {
+            using (var tester = ServerTester.Create())
+            {
+                await tester.StartAsync();
+
+                var url = tester.PayTester.ServerUri.AbsoluteUri;
+                HttpClient client = new HttpClient();
+
+                // check onion location is present for HTML page request
+                using var htmlRequest = new HttpRequestMessage(HttpMethod.Get, new Uri(url));
+                htmlRequest.Headers.TryAddWithoutValidation("Accept", "text/html,*/*");
+
+                var htmlResponse = await client.SendAsync(htmlRequest);
+                htmlResponse.EnsureSuccessStatusCode();
+                Assert.True(htmlResponse.Headers.TryGetValues("Onion-Location", out var onionLocation));
+                Assert.StartsWith("http://wsaxew3qa5ljfuenfebmaf3m5ykgatct3p6zjrqwoouj3foererde3id.onion", onionLocation.FirstOrDefault() ?? "no-onion-location-header");
+
+                // no onion location for other mime types
+                using var otherRequest = new HttpRequestMessage(HttpMethod.Get, new Uri(url));
+                otherRequest.Headers.TryAddWithoutValidation("Accept", "*/*");
+
+                var otherResponse = await client.SendAsync(otherRequest);
+                otherResponse.EnsureSuccessStatusCode();
+                Assert.False(otherResponse.Headers.Contains("Onion-Location"));
+            }
+        }
+
         private static bool IsMapped(Invoice invoice, ApplicationDbContext ctx)
         {
             var h = BitcoinAddress.Create(invoice.BitcoinAddress, Network.RegTest).ScriptPubKey.Hash.ToString();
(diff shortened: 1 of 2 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -3607,6 +3607,36 @@ public async Task CanLoginWithNoSecondaryAuthSystemsOrRequestItWhenAdded()
...
+                HttpClient client = new HttpClient();
+
+                // check onion location is present for HTML page request
+                using var htmlRequest = new HttpRequestMessage(HttpMethod.Get, new Uri(url));
+                htmlRequest.Headers.TryAddWithoutValidation("Accept", "text/html,*/*");
+
+                var htmlResponse = await client.SendAsync(htmlRequest);
+                htmlResponse.EnsureSuccessStatusCode();
+                Assert.True(htmlResponse.Headers.TryGetValues("Onion-Location", out var onionLocation));
+                Assert.StartsWith("http://wsaxew3qa5ljfuenfebmaf3m5ykgatct3p6zjrqwoouj3foererde3id.onion", onionLocation.FirstOrDefault() ?? "no-onion-location-header");
```
RiccardoMasutti, line 3628:
```suggestion
                Assert.StartsWith("http://wsaxew3qa5ljfuenfebmaf3m5ykgatct3p6zjrqwoouj3foererde3id.onion", onionLocation.FirstOrDefault() ?? "no-onion-location-header");
```

What link is it?

RiccardoMasutti, line 3628: Where does this address come from?

dennisreimann: It is the onion address of the dev/test server when it is started locally.

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -3607,6 +3607,36 @@ public async Task CanLoginWithNoSecondaryAuthSystemsOrRequestItWhenAdded()
...
+        [Fact(Timeout = TestTimeout)]
+        [Trait("Integration", "Integration")]
+        public async void CheckOnionlocationForNonOnionHtmlRequests()
+        {
+            using (var tester = ServerTester.Create())
+            {
+                await tester.StartAsync();
+
+                var url = tester.PayTester.ServerUri.AbsoluteUri;
+                HttpClient client = new HttpClient();
```
NicolasDorier, line 3619: Next time use `tester.PayTester.HttpClient` instead. It set the url automatically, and make sure to not exhaust tcp ports.

dennisreimann: Thanks for the feedback! Addressed with #1634.

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -3607,6 +3607,36 @@ public async Task CanLoginWithNoSecondaryAuthSystemsOrRequestItWhenAdded()
...
+                Assert.True(htmlResponse.Headers.TryGetValues("Onion-Location", out var onionLocation));
+                Assert.StartsWith("http://wsaxew3qa5ljfuenfebmaf3m5ykgatct3p6zjrqwoouj3foererde3id.onion", onionLocation.FirstOrDefault() ?? "no-onion-location-header");
+
+                // no onion location for other mime types
+                using var otherRequest = new HttpRequestMessage(HttpMethod.Get, new Uri(url));
+                otherRequest.Headers.TryAddWithoutValidation("Accept", "*/*");
+
+                var otherResponse = await client.SendAsync(otherRequest);
+                otherResponse.EnsureSuccessStatusCode();
+                Assert.False(otherResponse.Headers.Contains("Onion-Location"));
```
Kixunil, line 3636: Could you please explain the reason behind this check? It's not mentioned in the [official Onion-Location tutorial](https://community.torproject.org/onion-services/advanced/onion-location/)

Kukks: In reality, the tor browser only needs to care for the html page being loaded, where it will detect and transition you to the hidden service instead . If a sub resource is loaded with an onion header too, I would be afraid of crazy side-effects hat could come: resources being loaded twice, crazy redirect loops, etc

Kixunil: I don't think so (unless you actually found those problems, which then would be nice to report). It's supposed to be used only when user explicitly chooses to redirect, so the behavior should be clean. They recommend nginx configuration that sends the header unconditionally. I just tried it and didn't notice any problem.

BTW, I'm not saying there's something wrong with this code nor do I suggest to change it. I only wanted to know if I'm missing something before actually using nginx-based configuration on my server (for other non-btcpay stuff as well).

Thanks for explaining!

PR #1643, opened 2020-06-08
Fix lightning implementation, docs and tests

NicolasDorier:


BTCPayServer.Client/Models/LightningNodeInformationData.cs
```diff
@@ -9,7 +9,8 @@ namespace BTCPayServer.Client.Models
 {
     public class LightningNodeInformationData
     {
-        public IEnumerable<string> NodeInfoList { get; set; }
+        [JsonProperty("nodeURIs")]
+        public IEnumerable<string> NodeURIs { get; set; }
```
Kukks, line 13: Is this possible to ` IEnumerable<NodeInfo> ` and then decorate with `[JsonConverter(typeof(NodeUriJsonConverter))]`?

BTCPayServer.Client/Models/GreenfieldValidationError.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            if (path == null)
+                throw new ArgumentNullException(nameof(path));
+            if (message == null)
+                throw new ArgumentNullException(nameof(message));
+            Path = path;
+            Message = message;
+        }
+
+        public string Path { get; set; }
```
Kukks, line 23: Should we include `Code` here too or overkill?

BTCPayServer/wwwroot/swagger/v1/swagger.template.lightning.internal.json
```diff
@@ -290,13 +254,14 @@
...
                             "application/json": {
                                 "schema": {
-                                    "type": "string"
+                                    "type": "string",
+                                    "description": "A bitcoin address belonging to the lightning node"
                                 }
                             }
                         }
                     },
-                    "403": {
```
Kukks, line 298: 403 is still possible when you have insufficient permissions on the api key

NicolasDorier: I think we should not mention it in the doc, as this is true for every endpoints.

BTCPayServer.Client/Models/LightningNodeInformationData.cs
NicolasDorier: yes, will do

BTCPayServer.Client/Models/GreenfieldValidationError.cs
NicolasDorier: overkill

PR #1647, opened 2020-06-10
Use Tor Http Proxy instead of our internal hack one

NicolasDorier:
I wanted to get rid of our internal proxy and instead use the built in http proxy of Tor.

Sadly, https://github.com/dotnet/runtime/issues/37690 stops me from doing so :(

I did not see any hack available that is easier than just reimplementing the proxy ourselves.


MaiHistoryCompanyLimited:
{#1647}

NicolasDorier:
We might be able to revisit this in .NET 6.0

NicolasDorier:
This has been addressed thanks to the .net6.0 release!

Issue #1542, opened 2020-05-03
Notification system


NicolasDorier:
## Goal

Haing a notification system twitter and facebook like to notify users of several events such as:

* New version of btcpayserver available
* New payment received
* New pending refunds

## Technical design

Say a notification is for a store. A controller would call `notifcations.Broadcast(storeId, notification)` then the notifications instance would be responsible to fetch all the current users of that storeId and save a notification for them.

Table would be
```json
{
    "Created": "1234-56-78",
    "NoticeForUserId": "GUID",
    "Blob": {
    },
    "NoticeType": "",
    "Archived": false
}
```

This mean that if a notification target 2000 users, there will be 2000 dups of notification inserted.

We could avoid duplication by creating a table mapping user to notifications instead of embedding NoticeForUserId. 

I think we should not really care about that, BTCPay Server is not meant to be facebook level.
Also, such table would still require 2000 insert to map the users to the notification (say via a `UserNotification` table)

For the notifcation type, some example:
```json
{
    "Created": "1234-56-78",
    "NoticeForUserId": "GUID",
    "Blob": {
        "version": "1.0.5.0"
    },
    "NoticeType": "NewVersion",
    "Archived": false
}
```
```json
{
    "Created": "1234-56-78",
    "NoticeForUserId": "GUID",
    "Blob": {
        "invoiceId": "abc",
        "paymentType": "BTC",
        "amount": "10000",
    },
    "NoticeType": "NewPayment",
    "Archived": false
}
```
The renderer is responsible to create a `BodyText` and the `ActionLink` advantage are:

* We can later add localization easily
* This is API friendly (events can be consumed via greenfield) (@woutersamaey  [wants it](https://github.com/btcpayserver/btcpayserver/issues/775#issuecomment-523324249))
* This is extensible we can easily make new type as we want

All of that without making db migrations.

## Alternatives

Rockstardev does not like blobs, but he failed to convince me blob are evil. 
Before that we spoke about having `BodyText` and `ActionLink` in the column, or on blobs

on blobs => Under the presented design this can still be implemented without DB migration change if we really need it (though I am not sure we will)

On columns => I think this is not really flexible, and invite lot's of DB headache like collation, magic max length column size etc... Any new feature like localization or adding an imagePath will require more and more columns.

## Complaints

[...]

PR #1651, opened 2020-06-11
Notification system

rockstardev:
Resolves: #1542

notification first version

For start only `NewVersionNotification` is implemented. New types of notifications can be added simply by inheriting `NotificationEventBase` - they will be automatically subscribed and saved to database.

Converting notification to ViewModel is done through abstract method on new class. Generic invoker that connects database classes is implemented in `BTCPayServer.Models.NotificationViewModels.NotificationViewModelExt`.

Finally, invoking these new notifications is done through `EventAggregator`, there is `BTCPayServer.Events.Notifications.NotificationsHelperExt` class that provides static utility extension methods.

TODOs after concept is ACKed / PR merged:
 - Remove "Generate Test Notification" button
 - Tweak design
 - Add more notification types


BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -0,0 +1,105 @@
...
+                var notif = _db.Notifications.Single(a => a.Id == id && a.ApplicationUserId == claimWithId.Value);
+                notif.Seen = !notif.Seen;
+                await _db.SaveChangesAsync();
+            }
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> MassAction(string command, string csvGuids)
```
NicolasDorier, line 75: You don't have to do this kind of hack with csv.
The solution is receiving an array of viewModel with a property `SampleViewModel.IsSelected` and bind the checkbox to this.

I can refactor that after merging.

BTCPayServer/Views/Notifications/Index.cshtml
```diff
@@ -0,0 +1,109 @@
+﻿@model BTCPayServer.Models.NotificationViewModels.IndexViewModel
+@{
+    ViewData["Title"] = "Notifications";
+}
+@section HeadScripts {
+    <script src="~/modal/btcpay.js" asp-append-version="true"></script>
+}
+<section>
+    <div class="container">
+        @if (TempData.HasStatusMessage())
```
Kukks, line 10: There is a partial for this

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public NotificationManager(ApplicationDbContext db)
+        {
+            _db = db;
+        }
+
+        public NotificationSummaryViewModel GetSummaryNotifications(ClaimsPrincipal user)
+        {
+            var resp = new NotificationSummaryViewModel();
+            var claimWithId = user.Claims.SingleOrDefault(a => a.Type == ClaimTypes.NameIdentifier);
```
Kukks, line 86: Inject `UserManager` and use its `GetUserId` method instead

BTCPayServer/Events/Notifications/NotificationEventBase.cs
```diff
@@ -0,0 +1,31 @@
...
+    public abstract class NotificationEventBase
+    {
+        public NotificationData ToData()
+        {
+            var obj = JsonConvert.SerializeObject(this);
+
+            var data = new NotificationData
+            {
+                Created = DateTimeOffset.UtcNow,
+                NotificationType = GetType().Name,
```
Kukks, line 22: probably better to have a `virtual string NotificationType {get { return GetType().Name; }} ` to be able to override this if need be

BTCPayServer/Events/Notifications/NewVersionNotification.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace BTCPayServer.Events.Notifications
+{
+    public class NewVersionNotification : NotificationEventBase
+    {
+        public string Version { get; set; }
+
+        public override NotificationViewModel ToViewModel(NotificationData data)
+        {
+            var casted = JsonConvert.DeserializeObject<NewVersionNotification>(data.Blob.ToStringFromUTF8());
```
Kukks, line 19: Should probably use the `ZipUtils` like we use everywhere else for blobs to stay consistent

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
```diff
@@ -0,0 +1,41 @@
...
+        public bool Seen { get; set; }
+    }
+
+    public static class NotificationViewModelExt
+    {
+        public static NotificationViewModel ViewModel(this NotificationData data)
+        {
+            var baseType = typeof(NotificationEventBase);
+
+            var typeName = baseType.FullName.Replace(nameof(NotificationEventBase), data.NotificationType, StringComparison.OrdinalIgnoreCase);
```
Kukks, line 35: This can be error prone and dangerous, especially if you refactor and change type names or namespaces.. How about you add an interface `INotificationDataParser` to `NotificationEventBase` for its `ToViewModel(NotificationData data)`, register them all to DI and then in `NotificationManager` when constructing the vms, you match implementation against `NotificationType`

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -0,0 +1,105 @@
...
+                    .Where(a => a.ApplicationUserId == userId)
+                    .Select(a => a.ViewModel())
+                    .ToList()
+            };
+
+            return View(model);
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> Generate()
```
Kukks, line 53: testing method that needs to be removed? 
Version notification needs some state saved somewhere + a hosted service to routinely query github for new tagged versions to publish the event

BTCPayServer/Events/Notifications/NotificationsHelperExt.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BTCPayServer.Events.Notifications
+{
+    public static class NotificationsHelperExt
+    {
+        public static void NoticeNewVersion(this EventAggregator aggr, string version)
+        {
+            aggr.Publish(new NewVersionNotification
```
Kukks, line 13: This should only be for server admins, so you will need `UserManager`

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,116 @@
...
+        // we need publicly accessible method for reflection invoke
+        public void SubscribeHelper<T>() => base.Subscribe<T>();
+
+        protected override async Task ProcessEvent(object evt, CancellationToken cancellationToken)
+        {
+            if (evt is NotificationEventBase)
+            {
+                var data = (evt as NotificationEventBase).ToData();
+
+                var admins = await _UserManager.GetUsersInRoleAsync(Roles.ServerAdmin);
```
Kukks, line 56: notifications are role agnostic, this should be handles on a notification type level

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,116 @@
...
+            resp.UnseenCount = _db.Notifications
+                .Where(a => a.ApplicationUserId == claimWithId.Value && !a.Seen)
+                .Count();
+
+            if (resp.UnseenCount > 0)
+            {
+                resp.Last5 = _db.Notifications
+                    .Where(a => a.ApplicationUserId == claimWithId.Value && !a.Seen)
+                    .OrderByDescending(a => a.Created)
+                    .Take(5)
```
Kukks, line 98: Should be an option that defaults to 5 not hardcoded

pavlenex:
UI-wise don't have any remarks, all looks good. Obviously needs more notifications type, but once the code review guys are happy, tACK from my end.

BTCPayServer.Data/Data/ApplicationDbContext.cs
```diff
@@ -37,90 +37,27 @@ public DbSet<InvoiceData> Invoices
...
+        public DbSet<StoreData> Stores { get; set; }
+        public DbSet<UserStore> UserStore { get; set; }
+        public DbSet<AddressInvoiceData> AddressInvoices { get; set; }
+        public DbSet<SettingData> Settings { get; set; }
+        public DbSet<PairingCodeData> PairingCodes { get; set; }
+        public DbSet<PairedSINData> PairedSINData { get; set; }
+        public DbSet<APIKeyData> ApiKeys { get; set; }
+        public DbSet<StoredFile> Files { get; set; }
+        public DbSet<U2FDevice> U2FDevices { get; set; }
+        public DbSet<NotificationData> Notifications { get; set; }
```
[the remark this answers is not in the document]
rockstardev: Yes

BTCPayServer.Data/Data/ApplicationDbContext.cs
```diff
@@ -131,6 +68,9 @@ protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         protected override void OnModelCreating(ModelBuilder builder)
         {
             base.OnModelCreating(builder);
+            NotificationData.OnModelCreating(builder);
```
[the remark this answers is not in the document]
rockstardev: Glad you like it, mentioned it to @NicolasDorier; I'll be tying relations to specific classes when new version is release and I start with cleanup.

BTCPayServer/Views/Shared/_Layout.cshtml
```diff
@@ -74,9 +79,42 @@
                         <li class="nav-item">
                             <a asp-area="" asp-controller="Manage" asp-action="Index" title="My settings" class="nav-link js-scroll-trigger" id="MySettings"><i class="fa fa-user"></i></a>
                         </li>
+
+                        var notificationModel = notificationManager.GetSummaryNotifications(User);
```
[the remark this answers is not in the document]
rockstardev: All right: 11e91e47393ace2e22a577c784a1cdbfa3fcff24

BTCPayServer/Views/Notifications/Index.cshtml
```diff
@@ -0,0 +1,109 @@
...
+                            <th width="80px">&nbsp;</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        @foreach (var item in Model.Items)
+                        {
+                            @* TODO: Multiselect akin to Gmail *@
+                            <tr data-guid="@item.Id">
+                                <td><input type="checkbox" class="selector" data-guid="@item.Id" /></td>
+                                <td onclick="rowseen(this)" class="cursor-pointer @(item.Seen ? "": "font-weight-bold")">@item.Created.ToBrowserDate()</td>
```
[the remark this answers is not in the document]
rockstardev: Not sure I follow... since there is also need to update the styling of rows. You are talking about handling the case where JS is _not_ enabled?

BTCPayServer/Views/Notifications/Index.cshtml
```diff
@@ -0,0 +1,109 @@
+﻿@model BTCPayServer.Models.NotificationViewModels.IndexViewModel
+@{
+    ViewData["Title"] = "Notifications";
+}
+@section HeadScripts {
+    <script src="~/modal/btcpay.js" asp-append-version="true"></script>
```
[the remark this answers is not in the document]
rockstardev: 95d72fa89e04a52a2a15a3c2c4695f6afde0d14b

BTCPayServer/Views/Notifications/Index.cshtml
rockstardev: Where? Considering you have this similar check throughout bunch of other files. Example:

https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Account/Register.cshtml#L7

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,116 @@
...
+                    ApplicationDbContextFactory contextFactory,
+                    EventAggregator eventAggregator) : base(eventAggregator)
+        {
+            _UserManager = userManager;
+            _ContextFactory = contextFactory;
+        }
+
+        protected override void SubscribeToEvents()
+        {
+            SubscribeAllChildrenOfNotificationEventBase();
```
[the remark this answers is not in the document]
rockstardev: Base class doesn't work... I've actually tried it like that first. Opened issue: #1657 ... let me know if you want to handle it and I can then rebase PR on top of that... or I can implement it as part of PR.

cc @NicolasDorier

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: This `UserManager<ApplicationUser>` is so terrible... we should wrap all that behind some consistent user logic class... then either having it as context for base classes or inject it in user dependent classes.

5b9d4c9b901df163cb06bf96d6f3cb55294e3d09

BTCPayServer/Events/Notifications/NotificationEventBase.cs
rockstardev: ad99b8b48ccfb4567005d56afc967e19f1038420

BTCPayServer/Events/Notifications/NewVersionNotification.cs
rockstardev: Hahaha... we are pkziping blobs; true rockstar mode.

91f63d77ce224936d54e280a253d3cc30e7e6daa

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
rockstardev: If you look at code previously it already had direct compare:

https://github.com/btcpayserver/btcpayserver/blob/70075f73ef0991cb9d462f3ba6ef5e50431ecf62/BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs#L32

The problem for me is that now every time new Notification type is added to the solution we would need to register it to this repository of knowledge. If you already have that we could simply change this method to:

    public static NotificationViewModel ViewModel(this NotificationData data)
    {
        if (data.NotificationType == nameof(NewVersionNotification))
            return new NewVersionNotification().ToViewModel(data);
        else if (...)
            // keep adding
    }

What I've done for now is made NotificationEventBase `internal`... when we are done with adding all Notification types I guess we can revisit this and see how to refactor it with some kind of DI registration.

rockstardev: 2ff608ba8e8892e217fb6bc9e973a25968120fe8

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: Yes... and I thought version checking to be separate PR, once we get Notification System in.

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -200,6 +200,9 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
 
             services.AddSingleton<ChangellyClientProvider>();
 
+            services.AddSingleton<IHostedService, NotificationDbSaver>();
+            services.AddScoped<NotificationManager>();
```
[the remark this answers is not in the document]
NicolasDorier: I am also suggesting to use same class for both NotificationDbSaver and NotificationManager. This is non blocking for merge though.

BTCPayServer.Data/Data/NotificationData.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+
+namespace BTCPayServer.Data
+{
+    public class NotificationData
+    {
+        public string Id { get; set; }
```
NicolasDorier, line 12: should use maxlegth

BTCPayServer.Data/Data/NotificationData.cs
```diff
@@ -0,0 +1,28 @@
...
+        public DateTimeOffset Created { get; set; }
+        public string ApplicationUserId { get; set; }
+        public ApplicationUser ApplicationUser { get; set; }
+        public string NotificationType { get; set; }
+        public bool Seen { get; set; }
+        public byte[] Blob { get; set; }
+
+        internal static void OnModelCreating(ModelBuilder builder)
```
NicolasDorier, line 20: good idea to do like that, we should clean other classes after next release

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,131 @@
...
+            base.SubscribeToEvents();
+        }
+
+        // subscribe all children of NotificationEventBase
+        public void SubscribeAllChildrenOfNotificationEventBase()
+        {
+            var method = this.GetType().GetMethod(nameof(SubscribeHelper));
+            var notificationTypes = this.GetType().Assembly.GetTypes().Where(a => typeof(NotificationEventBase).IsAssignableFrom(a));
```
NicolasDorier, line 39: ugh, I would prefer we fix the subscribe instead.

BTCPayServer/Events/Notifications/NotificationEventBase.cs
```diff
@@ -0,0 +1,28 @@
...
+using BTCPayServer.Models.NotificationViewModels;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Events.Notifications
+{
+    public abstract class NotificationEventBase
+    {
+        public virtual string NotificationType { get { return GetType().Name; } }
+
+        public NotificationData ToData()
```
NicolasDorier, line 12: NotificationData are not necessarily serializable.
We should do some custom serialization mapping instead of assuming it.

NicolasDorier:
My main problem in this PR is that you should not assume that the events in the aggregator are serializable. Also, you should not use serialization as NotificationType.

Prefer custom mapping with custom notification type, this make it harder to shoot ourselves in the foot.
Such custom notification type should have a Scope (we don't actually need to serialize this), that we can use to select which users are concerned by the notification. (In this PR there will be a single scope: Server Admin)

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: #1657

BTCPayServer/Hosting/BTCPayServerServices.cs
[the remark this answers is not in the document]
rockstardev: They are really conceptually different, I've refactored it to both handle singleton and caching.

d5d6cd3511b31cb22696d653d74908b61b2bad76

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+        }
+    }
+
+    public class NotificationManager
+    {
+        private readonly ApplicationDbContext _db;
+
+        public NotificationManager(ApplicationDbContext db)
```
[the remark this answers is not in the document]
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/1651#discussion_r439207515

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: Nah... we'll get to this after we handle design with known maximum of notifications... as @pavlenex mentions here: https://github.com/btcpayserver/btcpayserver/pull/1651#issuecomment-642722154

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: Refactored: 90336d020523f774d998e7d18e5b96e27e68597a

BTCPayServer/Events/Notifications/NotificationsHelperExt.cs
rockstardev: Refactored: 90336d020523f774d998e7d18e5b96e27e68597a

BTCPayServer.Data/Data/NotificationData.cs
rockstardev: We're not adding `MaxLength` on any other `public string Id` entity. Worse yet, we don't have it on related entities (like `ApplicationUser`).

But if you'll let me introduce limits, then I'm gonna add all possible ones: 0659d8ecdcb4908db634689b379845d07af9dc50

BTCPayServer/Events/Notifications/NotificationEventBase.cs
rockstardev: See refactor: 90336d0

If you want 100% control over serialization `NotificationEvent` can be passing `NotificationData` instead of `NotificationBase` and then `NotificationDbSaver` would be essentially just saving whatever `NotificationSender` generates.

rockstardev:
I've turned on rockstar mode and handled all of the relevant complaints... maybe you @Kukks should really stop coding and just review PRs of @NicolasDorier... we could end up working faster.

Leftovers before merged:
 - Me and Kukks discussing 2 unresolved issues
 - @NicolasDorier taking a look at refactored code and deciding if it likes it better now vs 3 hours ago. This PR is no longer dependent on #1657 now that I'm using `NotificationEvent`. 
 - Refactoring MassActions to use server model instead of csvGuids.

Leftovers during / after merged:
 - Taking a look at design with @pavlenex 
 - New notification types
 - Implementing configurable HostedService that makes use of `NewVersionNotification`

BTCPayServer/Views/Notifications/Index.cshtml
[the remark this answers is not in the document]
Kukks: Yeah, basically do it as if it was in no-js mode, then add jquery on top to override and send it via ajax instead

BTCPayServer.Data/Migrations/20200612051342_LimitingNotificationStringFields.cs
```diff
@@ -0,0 +1,21 @@
+﻿using BTCPayServer.Data;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace BTCPayServer.Migrations
+{
+    [DbContext(typeof(ApplicationDbContext))]
+    [Migration("20200612051342_LimitingNotificationStringFields")]
+    public class LimitingNotificationStringFields : Migration
```
NicolasDorier, line 9: this migration is completely bug free.

BTCPayServer.Data/Data/NotificationData.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            builder.Entity<NotificationData>()
+                .HasOne(o => o.ApplicationUser)
+                .WithMany(n => n.Notifications)
+                .HasForeignKey(k => k.ApplicationUserId);
```
NicolasDorier, line 29: Don't forget the DeleteOnCascade

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -200,6 +201,10 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
 
             services.AddSingleton<ChangellyClientProvider>();
 
+            services.AddSingleton<IHostedService, NotificationDbSaver>();
+            services.AddSingleton<NotificationManager>();
+            services.AddScoped<NotificationSender>();
```
NicolasDorier, line 206: I think those three class should be merged into one called `NotificationService`.

BTCPayServer/Events/Notifications/NewVersionNotification.cs
```diff
@@ -0,0 +1,26 @@
...
+using BTCPayServer.Models.NotificationViewModels;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Events.Notifications
+{
+    internal class NewVersionNotification : NotificationBase
+    {
+        public string Version { get; set; }
+
+        public override NotificationViewModel ToViewModel(NotificationData data)
```
NicolasDorier, line 11: This is super strange design. This is a instance method, but actually ignore all field of the instance.

What I propose is the following:

Run this one, from the class's caller. (You can use reflection if you want)
```csharp
var casted = JsonConvert.DeserializeObject<NewVersionNotification>(ZipUtils.Unzip(data.Blob));
```

Then instead of having a 

```diff
- public override NotificationViewModel ToViewModel(NotificationData data)
+ public override void FillViewModel(NotificationViewModel viewModel)
```
And make `FillViewModel` just use the instance fields.

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
```diff
@@ -0,0 +1,40 @@
...
+        public bool Seen { get; set; }
+    }
+
+    public static class NotificationViewModelExt
+    {
+        public static NotificationViewModel ViewModel(this NotificationData data)
+        {
+            var baseType = typeof(NotificationBase);
+
+            var typeName = baseType.FullName.Replace(nameof(NotificationBase), data.NotificationType, StringComparison.OrdinalIgnoreCase);
```
NicolasDorier, line 34: `NotificationType` is actually virtual, so not guaranteed to be a type name.
I would suggest instead of `NotificationType` property, having a `[Notifcation("simpleName")]` and use the `simpleName` int he NotificationType column.

Because I am a bit worried that we break code because of simple class renaming/refactoring. (say we rename a class, and thus the column's NotificationType do not map to an actual type anymore). 

Worse of it is that it would break in production.

BTCPayServer/Events/Notifications/NewVersionNotification.cs
rockstardev: It is "super strange design" because of blobs... the whole idea of this abstract method is to force developer to add logic for mapping Data to ViewModel when new notification class is added.

FillViewModel method wouldn't work because you need NotificationData in order to get Id, Created, Seen... and all those properties that are not on BlobParsedClass.

So look at the method as "force dev to define mapping between data and viewmodel"; don't look at it as method that directly operates on instance.

https://stackoverflow.com/a/7945206/237858

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
rockstardev: We are already saving simple name and not full class name (Full.Namespace.Plus.ClassName).

The real problem is that we need to be able to define conventions on all new Notifications. What we essentially need is "conventions on types", while interfaces and abstracts classes work with adding "conventions on interfaces". I could dig deeper and define additional factory classes, but then you would end up with way more code and also be required to define dictionary with mapping like:

`"NotificationType" => TypeWithLogicToConvertDataOfNotificationTypeInView`

I think what we currently have is great start - we'll keep adding new Notification classes into `BTCPayServer.Events.Notifications` namespace and things will work. If we really expand the system so much that we end up needing to create mapping between `NotificationType` and class that converts that Entity from DB with that type into ViewModel - we can dig into that.

P.S. I've added explicit `NotificationType` on `NewVersionNotification`: 1acf628c127de62c9b13e75f2e5526b5382266fd

BTCPayServer/Hosting/BTCPayServerServices.cs
rockstardev: Nah, they are all conceptually different - NotificationDbSaver is hosted service that listens for events and saves them to db, NotificationManager is used in UI, and NotificationSender will be used by classes that want to pump notifications into the system.

BTCPayServer.Data/Data/NotificationData.cs
rockstardev: :+1: 5267cbd84fbee052858f865fd6f5eca930fa9075

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: 34d593ea573929c8c7ee015db4eea821ed675f01

BTCPayServer/Views/Notifications/Index.cshtml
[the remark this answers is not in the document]
rockstardev: I'll let you do it after merge, especially since now I have form in there so I implement: https://github.com/btcpayserver/btcpayserver/pull/1651#discussion_r438610802

BTCPayServer/Views/Notifications/Index.cshtml
Kukks: i r wrong

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (_cache.TryGetValue<NotificationSummaryViewModel>(userId, out var obj))
+                return obj;
+            
+            var resp = fetchNotificationsFromDb(userId);
+            _cache.Set(userId, resp, new MemoryCacheEntryOptions().SetAbsoluteExpiration(TimeSpan.FromMilliseconds(_cacheExpiryMs)));
+
+            return resp;
+        }
+
+        private NotificationSummaryViewModel fetchNotificationsFromDb(string userId)
```
Kukks, line 73:
```suggestion
        private NotificationSummaryViewModel FetchNotificationsFromDb(string userId)
```

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,119 @@
...
+        private readonly ApplicationDbContextFactory _factory;
+        private readonly UserManager<ApplicationUser> _userManager;
+
+        public NotificationManager(ApplicationDbContextFactory factory, UserManager<ApplicationUser> userManager)
+        {
+            _factory = factory;
+            _userManager = userManager;
+        }
+
+        private static MemoryCache _cache = new MemoryCache(new MemoryCacheOptions());
```
Kukks, line 58: There is a service in DI for this, `IMemoryCacheProvider` i think

BTCPayServer/HostedServices/NotificationDbSaver.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        protected override void SubscribeToEvents()
+        {
+            Subscribe<NotificationEvent>();
+            base.SubscribeToEvents();
+        }
+
+        protected override async Task ProcessEvent(object evt, CancellationToken cancellationToken)
+        {
+            var casted = evt as NotificationEvent;
```
Kukks, line 33: either direct cast  to type or do a null guard to return out of method to avoid possible null exceptions

BTCPayServer/Controllers/NotificationsController.cs
Kukks: Ok, let's move this test somewhere else then maybe and create an actual test notification type. Would actually be good to have it in the tests project namespace if possible

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -0,0 +1,100 @@
...
+                    await _db.SaveChangesAsync();
+
+                    return RedirectToAction(nameof(Index));
+                }
+            }
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        private bool validUserClaim(out string userId)
```
Kukks, line 94:
```suggestion
        private bool ValidUserClaim(out string userId)
```

BTCPayServer/Events/Notifications/NotificationBase.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using BTCPayServer.Data;
+using BTCPayServer.Models.NotificationViewModels;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Events.Notifications
+{
+    // Make sure to keep all NotificationEventBase classes in same namespace
```
Kukks, line 8: do not like this requirement at all, wish we did not rely on reflection for dealing with type parsing and vm conversion :(

BTCPayServer/Events/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using BTCPayServer.Data;
+using Microsoft.AspNetCore.Identity;
+
+namespace BTCPayServer.Events.Notifications
+{
+    public class NotificationSender
```
Kukks, line 8: This is just a helper class right? I don't see it particularly useful in this PR except for testing

BTCPayServer/Views/Notifications/Index.cshtml
```diff
@@ -0,0 +1,108 @@
...
+                            </tr>
+                        </thead>
+                        <tbody>
+                            @foreach (var item in Model.Items)
+                            {
+                                <tr data-guid="@item.Id">
+                                    <td>
+                                        <input name="selectedItems" type="checkbox" class="selector" value="@item.Id" /> 
+                                    </td>
+                                    <td onclick="rowseen(this)" class="cursor-pointer @(item.Seen ? "": "font-weight-bold")">@item.Created.ToBrowserDate()</td>
```
Kukks, line 56: seems like I lost my prev comments somehow.. We should embed a form here in case js is not available

BTCPayServer/Events/Notifications/NewVersionNotification.cs
Kukks: It's not too bad, just a little weird. In transmuter we define something like 
```csharp
interface NotificationHandler<INotification>{
  bool CanHandleNotificatioNtype(string type);
  NotificationViewModel ToViewModel(NotificationData data);
  NotificationData[]  ConstructData(INotification notification);
}
```

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -0,0 +1,100 @@
...
+                var notif = _db.Notifications.Single(a => a.Id == id && a.ApplicationUserId == userId);
+                notif.Seen = !notif.Seen;
+                await _db.SaveChangesAsync();
+            }
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> MassAction(string command, string[] selectedItems)
```
Kukks, line 77: You can just use this action to embed the form in the html where there is a js requirement

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        [HttpGet]
+        public async Task<IActionResult> Index(int skip = 0, int count = 50, int timezoneOffset = 0)
+        {
+            if (!validUserClaim(out var userId))
+                return RedirectToAction("Index", "Home");
+
+            var model = new IndexViewModel()
+            {
+                Items = _db.Notifications
```
Kukks, line 43: Shouldnt we move this to the service and introduce a `NotificationQuery` model like we have elsehwere?

BTCPayServer/Events/Notifications/NewVersionNotification.cs
NicolasDorier: What about

`public override void FillViewModel(NotificationViewModel viewModel, NotificationData data)` ?

NicolasDorier: > method wouldn't work because you need NotificationData in order to get Id, Created, Seen

Actually this can be done at the caller's level, does not have to be done in the FillViewModel as it is something that ALL viewmodel would have to do.

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: ec876ce11555acaaf285d7cdf9c6b2deba3430e9

BTCPayServer/Events/Notifications/NotificationBase.cs
rockstardev: Well, I do not like being forced to register every new Notification class, but if you and @NicolasDorier insist, then after merge just refactor it that way.

BTCPayServer/Events/Notifications/NotificationSender.cs
rockstardev: My idea is that class to be a helper that you inject in classes whenever you want to send helper for sending notifications. So it'll be useful beyond testing.

BTCPayServer/Views/Notifications/Index.cshtml
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/1651#discussion_r439782585

BTCPayServer/Events/Notifications/NewVersionNotification.cs
rockstardev: e81db94be7c117ee1abae2d5b3be568c2a7af7dd

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: For extra interface - yeah, I was saying we can be more verbose, but that I preferred to keep it simple and with 0 registrations. As I said - if you prefer it with registrations - will be easy to go in that direction after merging PR.

For js action: https://github.com/btcpayserver/btcpayserver/pull/1651#discussion_r439920641

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: 25df4d07dd1620e11b823cd431c4f774bb87af26

BTCPayServer/HostedServices/NotificationDbSaver.cs
rockstardev: 762f7b94f1a95e8656e5411c2c4748ad61f4a74d

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: I've prepared the foundation with 0dab96f0a682b39ad8efd214a95cef492efcfdbc and then gave it a try... but man, it feels so fake and unnecessary.

I've done paging instead (f072ec3a8c4656cce0b12aaa6f69b56c9e0a39f2)... am thinking to maybe give it a try again with `NotificationQuery` and `NotificationRepository` when I abstract paging view models. But for now, I would definitely keep it like this, especially since we need `ApplicationDbContext` in that class for other stuff. 

If @NicolasDorier also wants repository I can refactor that after PR is merged (and fully remove `ApplicationDbContext` from controller).

rockstardev:
Hahaha... wait for a moment there @NicolasDorier

NicolasDorier:
sorry, got pumped up

rockstardev:
Done now, off to answer @Kukks comment and celebrate with going to sleep ;)

BTCPayServer/Controllers/NotificationsController.cs
rockstardev: I've removed the button, but for now left method for local testing - so dev can hit `/Generate?version=1.0.4.4` and have notification generated.

I'll be adding new notifications (like for partial payments) and once that's done I'll remove this test method and recode test to directly use `NotificationSender`.

PR #1652, opened 2020-06-11
Change the seed saved in NBXplorer to hot wallet

### LINKED ISSUE #1641: Rename " the seed saved in NBXplorer" to "Sign with a hot wallet"
~~I think it's high time we rename `sign with vault (preview)` to `Sign with a hardware wallet`~~

~~I don't think people know what vault is and preview sound sketchy but our HWI works great and I think we tested it enough~~

- [ ] Rename `the valut (preview)` to `a hardware wallet`

vault

pavlenex:
Closes #1641 

a hot wallet

Term may not be technically 100% correct but since we have "Is a hot wallet" in store settings, it's easy to connect the two, plus I think people would easier understand a hot wallet.

### DIFF
--- a/BTCPayServer/Views/Wallets/WalletSigningMenu.cshtml
+++ b/BTCPayServer/Views/Wallets/WalletSigningMenu.cshtml
@@ -14,7 +14,7 @@
         <button name="command" type="submit" class="dropdown-item" value="analyze-psbt">... a wallet supporting PSBT</button>
         @if (Model.NBXSeedAvailable)
         {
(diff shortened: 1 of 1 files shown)


Kukks:
maybe "the hot wallet" makes more sense?

pavlenex:
@Kukks Indeed, thanks!

NicolasDorier:
should be "the hotwallet" rather than "a hotwallet" though.

Kukks:
> 
> 
> should be "the hotwallet" rather than "a hotwallet" though.

late to the party, already done :P

PR #1654, opened 2020-06-11
POC, main nav sub menus

Kukks:
image
image
image
image

image

image
image
image


pavlenex:
Good concept imo. `Merchant` may in a way limit those who aren't merchants, but Kukks had a valid argument on that

> on the contrary, it groups all merchant related tools under one item, leaving the rest to be unrelated to merchant stuff

I think this can indeed be further improved. What worries me is that useers need quick access buttons to invoices and stores, so putting them in a sub menu may not be the best option, however once we have a dashboard we can add quick access buttons there.

Kukks:
no lov

PR #1659, opened 2020-06-12
Add AlwaysIncludeNonWitnessUTXO to Wallet Send

Kukks:
Band-aid fix for trezor support. if a BTCPay wallet is restored and the bitcoind node is not running txindex=1, this will fail either way on trezor

We will need to document this


NicolasDorier:
Can you add an info link to https://medium.com/@jmacato/wasabi-wallets-advisory-for-trezor-users-7d942c727f92 ?

PR #1665, opened 2020-06-14
Attempt to embed configurator

Kukks:
Experimental mess of trying to combine 2 separate asp.net core apps into one, either by loading the controllers into one WebHost or loading two separate webhosts with separate urls


NicolasDorier:
Will wait for next release for this one.

Kukks:
Never figured this out, will keep on backburner for now

PR #1666, opened 2020-06-15
Small refactoring of Notifications

NicolasDorier:
* Introduce the concept of Scope
* Make the NotificationSender more abstract
* Using attributes for establishing the mapping between dotnet type and the notification type
* Remove the event generation test endpoint if not regtest

For the mapping, this make sense because when we deserialize from the DB we don't have an actual object on which we can use the virtual NotificationType property. To fix this, I use attributes instead.


BTCPayServer/Views/Account/Login.cshtml
```diff
@@ -2,7 +2,6 @@
 @inject BTCPayServer.HostedServices.CssThemeManager themeManager
 @{
     ViewData["Title"] = "Log in";
-    Layout = "_LayoutWelcome";
```
Kukks, line 5: are you sure?

BTCPayServer/Services/Notifications/Blobs/NotificationAttribute.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace BTCPayServer.Services.Notifications.Blobs
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class NotificationAttribute : Attribute
```
Kukks, line 9: It's better to just have an abstract getter for `NotificationType` in `BaseNotification` so that it's not forgotten

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -57,9 +61,9 @@ public async Task<IActionResult> Index(int skip = 0, int count = 50, int timezon
         [HttpGet]
         public async Task<IActionResult> Generate(string version)
         {
-            await _notificationSender.NoticeNewVersionAsync(version);
-            // waiting for event handler to catch up
-            await Task.Delay(500);
+            if (_env.NetworkType != NBitcoin.NetworkType.Regtest)
```
Kukks, line 64: This action is only temporary until we add some test notification in the tests csproj

BTCPayServer/Services/Notifications/NotificationSender.cs
```diff
@@ -20,32 +24,58 @@ public NotificationSender(ApplicationDbContextFactory contextFactory, UserManage
             _eventAggregator = eventAggregator;
         }
 
-        internal async Task NoticeNewVersionAsync(string version)
+        public async Task SendNotification(NotificationScope scope, BaseNotification notification)
```
Kukks, line 27: I think these 2 params should be combined: `BaseNotification ` should have a `abstract NotificationScope  {get;}`.

Notification types and its scopes should be coupled and well known

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
```diff
@@ -30,22 +30,26 @@ public class NotificationViewModel
...
     {
-        public static NotificationViewModel ViewModel(this NotificationData data)
+        static Dictionary<string, Type> _NotificationTypes;
+        static NotificationViewModelExt()
         {
-            var baseType = typeof(BaseNotification);
-
-            var fullTypeName = baseType.FullName.Replace(nameof(BaseNotification), data.NotificationType, StringComparison.OrdinalIgnoreCase);
-            var parsedType = baseType.Assembly.GetType(fullTypeName);
+            _NotificationTypes = Assembly.GetExecutingAssembly()
```
Kukks, line 36: We should use DI for this but in separate PR

BTCPayServer/Views/Account/Login.cshtml
NicolasDorier: wtf I thought I did not commited that

BTCPayServer/Services/Notifications/Blobs/NotificationAttribute.cs
NicolasDorier: I don't understand, there is no NotificationType property on BaseNotification anymore.

BTCPayServer/Controllers/NotificationsController.cs
NicolasDorier: yeah I know, but if we forget, this can be used to ddos

BTCPayServer/Services/Notifications/NotificationSender.cs
NicolasDorier: I disagree, because this is the caller of `SendNotification` which know exactly who to send to, not the notification itself. (for example, which storeId)

Also there is no reason to save the scope in the DB.

BTCPayServer/Services/Notifications/Blobs/NotificationAttribute.cs
Kukks: Yes, I'm saying you should add it back in but instead of virtual make it `abstract`

BTCPayServer/Services/Notifications/NotificationSender.cs
Kukks: Ah I did not know that `BaseNotification` was the actual model that was jsonified and saved to the db

BTCPayServer/Models/NotificationViewModels/IndexViewModel.cs
Kukks: I guess it can stay, just feels it fits project better to just use DI for type discovery through regsitration

BTCPayServer/Services/Notifications/Blobs/NotificationAttribute.cs
NicolasDorier: It can't be forgotten or test will fail.

Putting abstract, don't mean the attribute will be forgotten either, worse, you would have to keep both the attribute and the abstract method in sync.... The attribute is needed, there is no workaround it.

NicolasDorier:
So there is two points @Kukks mentioned:

> The scope  should be saved in the event.

I disagree for two reason:
* Once the event is broadcast, we don't need the scope anymore
* If we did that, it means that the called of `SendNotification`, instead of passing two parameters, need to only pass the Notification object, but before that, set its Scope property. I don't see how that's improving things rather than passing the two parameters.

The about the `NotificationType` property.
> Yes, I'm saying you should add it back in but instead of virtual make it abstract
I don't understand what it is about, as the NotificationType property has been completely removed and we don't want to need both, an attribute AND a property override.
Moreover, the concern is "what if somebody forget the attribute". That's not really a concern, as his tests can't possibly pass by mistake.

rockstardev:
1. We meant that when you define notification you'll already be defining what will be the scope of notification. You'll never send `NewVersionNotification` to non-admins. But you say that you want flexibility on senders and to define that every time you are using`SendNotification` method - fine from me.

2. We were saying that having `abstract` would force developer to define `NotificationType`. Content of `NotificationType` can also be checked with tests. You are saying you want Attribute instead - we're fine with that... in that case you obviously don't need `abstract` property (and that's what @Kukks was saying - my reading of what he wrote doesn't imply that he wanted to "add property and keep attribute"... it was "add property, remove attribute").

Kukks:
>     1. We meant that when you define notification you'll already be defining what will be the scope of notification. You'll never send `NewVersionNotification` to non-admins. But you say that you want flexibility on senders and to define scope every time you are using `SendNotification` method - fine by me.
> 
>     2. We were saying that having `abstract` would force developer to define `NotificationType`. Content of `NotificationType` can also be checked with tests. You are saying you want Attribute instead - we're fine with that...
>        In that case you obviously don't need `abstract` property (and that's what @Kukks was saying - I understood his suggestion as: "add back property, remove attribute"... NOT "add property and keep attribute").
Ditto

NicolasDorier:
Will merge this then. Removing attribute is not possible since when deserializing we don't have any instance to get the property value.

PR #1670, opened 2020-06-16
Introduce INotificationHandler

NicolasDorier:
Tasting my own dog food told me that https://github.com/btcpayserver/btcpayserver/pull/1666 was not a good solution.

While adding notifications to the payout in the commit of my pull request https://github.com/btcpayserver/btcpayserver/pull/1639/commits/7652026731a07276965db96719459ef425f63565 I found out that FillViewModel needed some additional dependencies, namely, `LinkGenerator` and `PathBase` to correctly generate action links.

It turns out passing new args to all notification types every times there is a new need is not very maintainable. I expect new notification types will need different dependencies.

As such, I introduce the `INotificationHandler` which is DI friendly. This solves the chicken and eggs problem of having NotificationType property in the Notification type itself, while making sure specific notifications can pull dependencies they need.

It also solves the issue that you were complaining of forgetting the attribute.

It also allow us later to move the notification in the Greenfield client project when we expose notifications via the API, while keeping the handlers in the main project!


NicolasDorier:
better! https://github.com/btcpayserver/btcpayserver/pull/1639/commits/6b898cc3680a917d98ba4994259e74802609d8e1

PR #1671, opened 2020-06-17
Support wasabi file format input

Kukks:
This adds support to importing the wasabi file format which seems to be used more for imports by Wasabi,ColdCard,BlueWallet & Cobo Vault (and way better than electrum anyway)


Kukks:
Tested on mainnet and seems to work  correctly btw

NicolasDorier:
Add a test for it like I did for electrum.

Kukks:
@NicolasDorier  done

NicolasDorier:
awesome, need to add to changelog

PR #1672, opened 2020-06-17
Adapt payjoin for BIP78

NicolasDorier:
Adapting payjoin to latest change of bip78

### DIFF
--- a/BTCPayServer/Services/PayjoinClient.cs
+++ b/BTCPayServer/Services/PayjoinClient.cs
@@ -7,11 +7,16 @@ using System.Net.Http;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using BTCPayServer.Logging;
 using BTCPayServer.Payments.Changelly.Models;
 using Google.Apis.Http;
+using Microsoft.Extensions.Logging;
 using NBitcoin;
+using NBitcoin.Payment;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using NUglify.Helpers;
+using TwentyTwenty.Storage;
 using IHttpClientFactory = System.Net.Http.IHttpClientFactory;
 
 namespace BTCPayServer.Services
@@ -21,7 +26,7 @@ namespace BTCPayServer.Services
     {
         public static ScriptPubKeyType? GetInputsScriptPubKeyType(this PSBT psbt)
         {
-            if (!psbt.IsAllFinalized() || psbt.Inputs.Any(i => i.WitnessUtxo == null))
+            if (!psbt.IsAllFinalized())
                 throw new InvalidOperationException("The psbt should be finalized with witness information");
             var coinsPerTypes = psbt.Inputs.Select(i =>
             {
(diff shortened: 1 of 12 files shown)


BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -77,185 +87,232 @@ public PayjoinClient(ExplorerClientProvider explorerClientProvider, IHttpClientF
...
+            if (signedPSBT == null)
+                throw new ArgumentNullException(nameof(signedPSBT));
+            if (signedPSBT.IsAllFinalized())
                 throw new InvalidOperationException("The original PSBT should not be finalized.");
-            var clientParameters = new PayjoinClientParameters();
-            var type = derivationSchemeSettings.AccountDerivation.ScriptPubKeyType();
-            if (!SupportedFormats.Contains(type))
+            var optionalParameters = new PayjoinClientParameters();
+            var inputScriptType = derivationSchemeSettings.AccountDerivation.ScriptPubKeyType();
+            if (!SupportedFormats.Contains(inputScriptType))
```
Kukks, line 105: `SupportedFormats` is not longer an issue anymore afaik?  We should remove.

BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -77,185 +87,232 @@ public PayjoinClient(ExplorerClientProvider explorerClientProvider, IHttpClientF
...
+                throw new PayjoinSenderException("GlobalXPubs should not be included in the receiver's PSBT");
             }
+            ////////////
 
-            foreach (var output in cloned.Outputs)
-            {
-                output.HDKeyPaths.Clear();
-            }
+            if (proposal.CheckSanity() is List<PSBTError> errors && errors.Count > 0)
+                throw new PayjoinSenderException($"The proposal PSBT is not insance ({errors[0]})");
```
Kukks, line 139:
```suggestion
                throw new PayjoinSenderException($"The proposal PSBT is not sane ({errors[0]})");
```

BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -77,185 +87,232 @@ public PayjoinClient(ExplorerClientProvider explorerClientProvider, IHttpClientF
...
+            foreach (var proposedPSBTInput in proposal.Inputs)
             {
-                var errorStr = await bpuresponse.Content.ReadAsStringAsync();
-                try
+                if (proposedPSBTInput.HDKeyPaths.Count != 0)
+                    throw new PayjoinSenderException("The receiver added keypaths to an input");
+                if (proposedPSBTInput.PartialSigs.Count != 0)
+                    throw new PayjoinSenderException("The receiver added partial signatures to an input");
+                var proposedTxin = proposalGlobalTx.Inputs.FindIndexedInput(proposedPSBTInput.PrevOut).TxIn;
+                bool isOurInput = ourPrevouts.Count > 0 && ourPrevouts.Peek() == proposedPSBTInput.PrevOut;
```
Kukks, line 157: why the Peek, should you not check all items in collection in case the order changed?

BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -77,185 +87,232 @@ public PayjoinClient(ExplorerClientProvider explorerClientProvider, IHttpClientF
...
-                    throw new PayjoinReceiverException(error["errorCode"].Value<string>(),
-                        error["message"].Value<string>());
+                    var inputPrevout = ourPrevouts.Dequeue();
+                    var originalTxin = originalGlobalTx.Inputs.FindIndexedInput(inputPrevout).TxIn;
+                    var originalPSBTInput = originalPSBT.Inputs.FindIndexedInput(inputPrevout);
+                    // Verify that sequence is unchanged.
+                    if (originalTxin.Sequence != proposedTxin.Sequence)
+                        throw new PayjoinSenderException("The receiver modified the sequence of one of our inputs");
+                    // Verify the PSBT input is not finalized
+                    if (proposedPSBTInput.IsFinalized())
```
Kukks, line 168: Does this matter? If it was finalized, then sender == receiver

BTCPayServer/Services/PayjoinClient.cs
```diff
@@ -77,185 +87,232 @@ public PayjoinClient(ExplorerClientProvider explorerClientProvider, IHttpClientF
...
-            if (newPSBT.CheckSanity() is IList<PSBTError> errors2 && errors2.Count != 0)
+            return proposal;
+        }
+
+        private int GetVirtualSize(ScriptPubKeyType? scriptPubKeyType)
+        {
+            switch (scriptPubKeyType)
             {
-                throw new PayjoinSenderException($"The PSBT of the receiver is insane ({errors2[0]})");
+                case ScriptPubKeyType.Legacy:
```
Kukks, line 260: might want to link to source calculation of these numbers

BTCPayServer/Services/PayjoinClient.cs
NicolasDorier: Because the spec changed, the receiver can't shuffle anymore, only insert randomly. This make implementation easier and "streamable" which is nice for hardware wallets.

BTCPayServer/Services/PayjoinClient.cs
NicolasDorier: Documented in the BIP now, this is basically the virtual size of an additional input.

BTCPayServer/Services/PayjoinClient.cs
NicolasDorier: I prefer being strict over what the sender is accepting. PSBT make it easy to shoot yourself in the foot, so the stricter we are the better.

PR #1673, opened 2020-06-17
Maintain invoice list params on navigation

### LINKED ISSUE #1498: Invoices default view
It would be nice to have setting when one can change default view of list of invoices. For example show all invoices except expired ones (I don't care about those ones).

dennisreimann:
This addresses and potentially closes #1498. As discussed in the issue this addition maintains the list route params on navigation to the invoice details view. It appends the params to the Invoices item in the top navigation so that one can navigate back to the previous list view.

To be discussed: As the params are not namespaced and I'm grabbing them from the `HttpContext.Request.Query` they are also set on the Payment Requests list view. Not critical, but may lead to unexpected results when navigating to the Invoices list after e.g. previously paginating the payment requests. Appreciate feedback and better ideas to implement this – keeping it as a draft PR for now.

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -586,12 +586,32 @@ namespace BTCPayServer.Controllers
         [BitpayAPIConstraint(false)]
         public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0, bool unfiltered = false)
         {
-            if (!unfiltered)
+            if (!unfiltered && searchTerm == null && timezoneOffset == 0 &&
+                Request.Cookies.TryGetValue(_lastQueryCookieName, out var cookieValue))
             {
-                var queryCookie = HttpContext.GetCookieValue(_lastQueryCookieName);
-                var queryCookieParams = JsonConvert.DeserializeObject<Dictionary<string, string>>(queryCookie);
-                if (searchTerm == null) searchTerm = queryCookieParams["searchTerm"];
-                if (timezoneOffset == 0) timezoneOffset = Int32.Parse(queryCookieParams["timezoneOffset"], NumberFormatInfo.InvariantInfo);
+                var storedParams = cookieValue.ToLowerInvariant().Trim('?')
+                    .Split("&", StringSplitOptions.RemoveEmptyEntries)
+                    .ToDictionary(s => s.Split("=")[0], s => s.Split("=")[1]);
+
+                if (storedParams.ContainsKey(nameof(searchTerm).ToLowerInvariant()))
+                {
+                    searchTerm = storedParams[nameof(searchTerm).ToLowerInvariant()];
+                }
+
+                if (storedParams.ContainsKey(nameof(timezoneOffset).ToLowerInvariant()))
+                {
+                    timezoneOffset = int.Parse(storedParams[nameof(timezoneOffset).ToLowerInvariant()],
+                        CultureInfo.InvariantCulture);
+                }
+            }
+            else if (unfiltered && Request.Cookies.ContainsKey(_lastQueryCookieName))
+            {
+                Response.Cookies.Delete(_lastQueryCookieName);
+            }
+            else if (!unfiltered)
+            {
+                Response.Cookies.Delete(_lastQueryCookieName);
+                HttpContext.Response.Cookies.Append(_lastQueryCookieName, Request.QueryString.ToString());
             }
 
             var fs = new SearchString(searchTerm);
@@ -606,13 +626,6 @@ namespace BTCPayServer.Controllers
                 TimezoneOffset = timezoneOffset
             };
 
-            var invoicesLastQuery = new Dictionary<string, string>
-            {
-                { "searchTerm", searchTerm },
-                { "timezoneOffset", timezoneOffset.ToString(NumberFormatInfo.InvariantInfo) }
-            };
-            var invoicesQueryJson = invoicesLastQuery.ToJson();
-            HttpContext.Response.Cookies.Append(_lastQueryCookieName, invoicesQueryJson);
 
             InvoiceQuery invoiceQuery = GetInvoiceQuery(searchTerm, timezoneOffset);
             var counting = _InvoiceRepository.GetInvoicesTotal(invoiceQuery);
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -599,8 +599,6 @@ namespace BTCPayServer.Controllers
             var invoicesLastQuery = new Dictionary<string, string>
             {
                 { "SearchTerm", searchTerm },
-                { "Skip", skip.ToString(NumberFormatInfo.InvariantInfo) },
-                { "Count", count.ToString(NumberFormatInfo.InvariantInfo) },
                 { "TimezoneOffset", timezoneOffset.ToString(NumberFormatInfo.InvariantInfo) }
             };
             var invoicesQueryJson = invoicesLastQuery.ToJson();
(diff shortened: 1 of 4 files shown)


dennisreimann:
@rockstardev good point, way better!

BTCPayServer/Views/Shared/_Layout.cshtml
```diff
@@ -1,16 +1,15 @@
...
 @inject BTCPayServer.HostedServices.CssThemeManager themeManager
 
 @{
-    var invoicesRouteParams = new Dictionary<string, string> {
-        { "searchTerm", ViewContext.HttpContext.Request.Query["searchTerm"] },
-        { "skip", ViewContext.HttpContext.Request.Query["skip"] },
-        { "count", ViewContext.HttpContext.Request.Query["count"] },
-        { "timezoneOffset", ViewContext.HttpContext.Request.Query["timezoneOffset"] }
-    };
+    var invoicesLastQuery = (string)(ViewData["InvoicesLastQuery"]) ?? HttpContextAccessor.HttpContext.Request.Cookies["InvoicesLastQuery"] ?? "{}";
```
rockstardev, line 11: You don't need `HttpContextAccessor`, you can instead just use `Context.Request.Cookies["InvoicesLastQuery"]`

NicolasDorier:
This is using a mix of parameters and cookie.

IMHO, cookie are a good place for saving preferences, and you don't need to circulate via parameters at all.

rockstardev:
@NicolasDorier, Denis explained the problem:

> Passing the data on via the ViewData (for the invoices list view) is easier than trying to get the value from the response as the IResponseCookies does not offer read access.

I guess if you insist on keeping it all with cookies, we can do this: https://stackoverflow.com/a/61929474/237858

NicolasDorier:
I don't understand why you need to get anything from Cookie.Response.

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -441,6 +442,18 @@ public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip
                 StoreIds = storeIds,
                 TimezoneOffset = timezoneOffset
             };
+
+            var invoicesLastQuery = new Dictionary<string, string>
+            {
+                { "SearchTerm", searchTerm },
+                { "Skip", skip.ToString(NumberFormatInfo.InvariantInfo) },
+                { "Count", count.ToString(NumberFormatInfo.InvariantInfo) },
```
NicolasDorier, line 450: unsure if it is a good idea to save Skip and Count. I would expect to be on the first page if I click on Invoices.

dennisreimann:
> I don't understand why you need to get anything from Cookie.Response.

It's for the case rockstar described in his initial comment: In case one changes the values on the list invoices page, you'd still have the old setting in the nav item, because the new ones haven't been applied yet and hence aren't part of the request cookie.

BTCPayServer/Controllers/InvoiceController.UI.cs
dennisreimann: Good point, I'll remove them. 👍

dennisreimann:
Removed the `ViewData` hack and now the params are always taken from the cookie – either [from the response or request cookie](https://github.com/btcpayserver/btcpayserver/pull/1673/commits/26fe49693873bae50e8edad0a2d4c97e24db4d16#diff-c1fbc84028a12825021d0f2ada6e397eR480) 🍪😋

dennisreimann:
`CheckNoDeadLink` failing, should be unrelated.

dennisreimann:
There has to be some way of distinguishing the cases "user has not specified anything so show the last query" and "user wants to reset the filter to see everything". I introduced the `unfiltered` param as passing an empty search (`searchTerm=`) didn't seem attractive and also has its quirks, because without a Binding Model there is no way to distinguish the empty string vs. `null` case.

The current version seems like the sleekest option to me, because it keeps 90% of the code in the controller action and it does not involve the layout and other actions (like the redirect after create). Maybe we can come up with a better way to distinguishing the case above and I'll be happy to refactor it.

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -581,8 +584,36 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody]Upda
...
         [BitpayAPIConstraint(false)]
-        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0)
+        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0, bool unfiltered = false)
         {
+            if (!unfiltered && searchTerm == null && timezoneOffset == 0 &&
+                Request.Cookies.TryGetValue(_lastQueryCookieName, out var cookieValue))
+            {
+                var storedParams = cookieValue.ToLowerInvariant().Trim('?')
+                    .Split("&", StringSplitOptions.RemoveEmptyEntries)
+                    .ToDictionary(s => s.Split("=")[0], s => s.Split("=")[1]);
```
NicolasDorier, line 594: Will not work if some characters in the search query is url escaped (say ' ' replaced by '%20%`)

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -581,8 +584,36 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody]Upda
...
+                }
+            }
+            else if (unfiltered && Request.Cookies.ContainsKey(_lastQueryCookieName))
+            {
+                Response.Cookies.Delete(_lastQueryCookieName);
+            }
+            else if (!unfiltered)
+            {
+                Response.Cookies.Delete(_lastQueryCookieName);
+                HttpContext.Response.Cookies.Append(_lastQueryCookieName, Request.QueryString.ToString());
```
NicolasDorier, line 614: can you just save the search ?

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -581,8 +585,36 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody]Upda
...
         [BitpayAPIConstraint(false)]
-        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0)
+        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0, bool unfiltered = false)
         {
+            if (!unfiltered && searchTerm == null && timezoneOffset == 0 &&
+                Request.Cookies.TryGetValue(_lastQueryCookieName, out var cookieValue))
+            {
+                var storedParams = cookieValue.ToLowerInvariant().Trim('?')
+                    .Split("&", StringSplitOptions.RemoveEmptyEntries)
+                    .ToDictionary(s => s.Split("=")[0], s => HttpUtility.HtmlDecode(s.Split("=")[1]));
```
NicolasDorier, line 595: I don't think this is `System.Web.HttpUtility.UrlDecode()` (don't use NBitcoin one)

NicolasDorier:
hey @pavlenex can you test that?

dennisreimann:
Let‘s exclude it from the upcoming release. I‘m not sure that the recent commits simplified it.

NicolasDorier:
Agreed, I don't really like such tricky code, there should have a way to do it clean. Let's fix for next release, will take a look later.

dennisreimann:
In general I liked the way it was before @Kukks refactoring, especially regarding that we might want to use this elsewhere (e.g. payment requests) so imho the `GetCookieValue` extension should be kept.

NicolasDorier:
@dennisreimann let me run a bunch of post release cleanup in the code and will come back to this one. I think more than a single piece of data, we should probably have an object that span several peferences in the cookie.

I don't like the `unflitered` parameter and I don't like all the crazy cookie hack happening in the current form, will try to see if I can do something about it.

BTCPayServer/Controllers/InvoiceController.UI.cs
rockstardev: 2ea8e84

rockstardev:
Continuing this in new PR, moved commits there and rebased on top of the latest master.

Issue #1319, opened 2020-02-04
Checkout page to show when transaction is confirmed.


Transisto:
Most services will only trigger an action when the transaction is confirmed and it's usually a customer service nightmare to respond to people asking if their transaction is confirmed or not.

bolatovumar:
@Transisto could you clarify what you are suggesting? BTCPay already has settings which allow you tweak when the invoice is considered paid:

![Screen Shot 2020-02-04 at 8 46 30 PM](https://user-images.githubusercontent.com/1934678/73811875-92fa6400-478f-11ea-833c-1fe1c29ef5c9.png)

You can also choose if you want the user to be redirected after invoice is paid:
![Screen Shot 2020-02-04 at 8 46 45 PM](https://user-images.githubusercontent.com/1934678/73811894-a9082480-478f-11ea-9595-7c134a5e62c1.png)

Transisto:
I mean the business has to be the one to let the customer know when the
invoice is actually paid, because btcpay invoice show "paid" the second it
sees something in the mempool.

pavlenex:
There's a difference between [paid and confirmed status](https://docs.btcpayserver.org/features/invoices#invoice-statuses).

Alternatively, you can send follow up emails based on different status either via your CMS or use [BTC-Transmuter](https://github.com/btcpayserver/btcTransmuter).

Transisto:
@pavlenex, for example our website monitors the invoice status charge to
trigger the service but now we have to build a UI component and callback to
it so the customer can monitor the progression of it's transaction on chain.

The generated customers invoices themselves could show status change when
they are considered paid from the merchant standpoint instead of just the
mempool standpoint.

pavlenex:
No, we can't conflict statuses like that but I see what you would want. 

You'd want when invoice is paid (but not confirmed) to show XYZ message, for example "Your invoice has been paid and now awaiting confirmation on the blockchain". And then when completed "Your invoice has been confirmed on the blockchain." I still don't quite understand why we'd want to do this since people paying with BTC would already know how to check transaction from their wallets, it's more likely even that they will check it their wallets than in the invoice URL.

Please alter your initial post if you want this as a feature request and clearly explain what is it that you're suggesting by following the [Feature Request template ](https://github.com/btcpayserver/btcpayserver/blob/master/.github/ISSUE_TEMPLATE/feature_request.md)so we can easier understand.

[...]

PR #1674, opened 2020-06-19
Add invoice notifications

Kukks:
pav needs to play around and see which ones fit and what needs to be edited

### DIFF
--- a/BTCPayServer/Services/Notifications/Blobs/InvoiceEventNotification.cs
+++ b/BTCPayServer/Services/Notifications/Blobs/InvoiceEventNotification.cs
@@ -25,12 +25,12 @@ namespace BTCPayServer.Services.Notifications.Blobs
 
             internal static Dictionary<string, string> TextMapping = new Dictionary<string, string>()
             {
-                {InvoiceEvent.PaidInFull, "was fully paid."},
+                // {InvoiceEvent.PaidInFull, "was fully paid."},
                 {InvoiceEvent.PaidAfterExpiration, "was paid after expiration."},
                 {InvoiceEvent.ExpiredPaidPartial, "expired with partial payments."},
                 {InvoiceEvent.FailedToConfirm, "has payments that failed to confirm on time."},
-                {InvoiceEvent.ReceivedPayment, "received a payment."},
-                {InvoiceEvent.Confirmed, "was confirmed."}
+                // {InvoiceEvent.ReceivedPayment, "received a payment."},
+                {InvoiceEvent.Confirmed, "was confirmed paid."}
             };
             
             protected override void FillViewModel(InvoiceEventNotification notification,
--- a/BTCPayServer/Services/Notifications/Blobs/InvoiceEventNotification.cs
+++ b/BTCPayServer/Services/Notifications/Blobs/InvoiceEventNotification.cs
@@ -37,7 +37,7 @@ namespace BTCPayServer.Services.Notifications.Blobs
                 NotificationViewModel vm)
             {
                 var baseStr = $"Invoice {notification.InvoiceId.Substring(0, 5)}..";
-                vm.Body = $"{TextMapping[notification.Event]} {baseStr}";
+                vm.Body = $"{baseStr} {TextMapping[notification.Event]}";
                 vm.ActionLink = _linkGenerator.GetPathByAction(nameof(InvoiceController.Invoice),
                     "Invoice",
                     new {invoiceId = notification.InvoiceId}, _options.RootPath);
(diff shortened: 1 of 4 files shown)


BTCPayServer/Services/Notifications/Blobs/InvoiceEventNotification.cs
```diff
@@ -0,0 +1,65 @@
...
+                {InvoiceEvent.FailedToConfirm, "has payments that failed to confirm on time."},
+                {InvoiceEvent.ReceivedPayment, "received a payment."},
+                {InvoiceEvent.Confirmed, "was confirmed."}
+            };
+            
+            protected override void FillViewModel(InvoiceEventNotification notification,
+                NotificationViewModel vm)
+            {
+                var baseStr = $"Invoice {notification.InvoiceId.Substring(0, 5)}..";
+                vm.Body = $"{TextMapping[notification.Event]} {baseStr}";
```
rockstardev, line 40: You've missed the ordering of vars, should be `$"{baseStr} {TextMapping[notification.Event]}"`. 

I would invite @pavlenex to double check and verify strings.

Kukks:
> 
> 
> PR looks great. Design wise, we need to get exact look that would work for all notifications - taking into consideration that most of them will be textual and not necessary have icon involved. You @pavlenex can tell us if we can tap designers that already worked with project, or 3 of us will put our design hat on and sketch it.
> 
> On colors for certain events - I guess we should leverage styling already present in themes. If we need to go beyond that - again - it's design effort to specify how those colors map to notifications, and can then add that info to `TextMapping` variable.

I was thinking we could maybe add a `BodyClass` property for the VM. Then for specific things, we could try and use `alert-warning`, etc as our pav overlord deems fit

rockstardev:
And I think this is good change: "Instead of date use Ago like we do for invoices (I think ago is more relevant for notifications)" 

@Kukks you can just switch from `ToBrowserDate()` to `ToTimeAgo()` in NotificationNavItem.cshtml#22... I would do it but don't have your fork as remote registered.

rockstardev:
> I was thinking we could maybe add a BodyClass property for the VM. Then for specific things, we could try and use alert-warning, etc as our pav overlord deems fit

I think this is why "we need to get exact look that would work for all notifications". It would allow us to clear unknowns and know if we need `BodyClass` or we'll require even a bigger weapon (even more control over resulting HTML).

pavlenex:
> @pavlenex can tell us if we can tap designers that already worked with project, or 3 of us will put our design hat on and sketch it.

I'm afraid we won't have the time to involve designers for the next release, looking at the progress we had in past on design, it always took few weeks. If you guys want, we can improve the coloring or icons or whatever in the next release, but for now, we can focus on creating a clean, simple design which we can expand in the future. However if you want to tackle design, I think this one is quite [solid](https://user-images.githubusercontent.com/36959754/85150574-fe326780-b252-11ea-9630-b2caf654edaa.PNG). 

One reason I want to have different coloring involved is because not many people especially when they begin with BTCPay know the difference between paid and completed. They see paid and assume it's okay, if we can tackle that one textually somehow, for now I'd be good with it.

pavlenex:
Testing the PR after 5f2c5da it now looks solid. So the two concerns we should address before merging this:
- Adjusting notification appearance for themes.
- Distinguishing paid and complete

Capture

@rockstardev @Kukks how would you like to distinguish the `paid` and `completed` status? If we want to do it textually for now, I suggest we add something suggested in #1319
Perhaps:

`Invoice paid, pending confirmation`
`Invoice completed` or `Invoice paid and confirmed` or `Invoice fully paid and confirmed`

NicolasDorier:
This is too spammy for users. Please for now, can you just put a notification when it get paid?

Kukks:
> 
> 
> This is too spammy for users. Please for now, can you just put a notification when it get paid?

What about dropping `ReceivedPayment` and `PaidInFull` and leaving the rest

This is how it looks on a an invoice being paid and confirmed on time:
image


The rest only show in invalid situations where a merchant needs to take action

pavlenex:
> 
> 
> This is too spammy for users. Please for now, can you just put a notification when it get paid?

What is spammy? Complete and Paid or also invalid and partially paid? I think we can go with `complete` but I would not drop `invalid` and `partiallypaid` it's important to be notified of those and they don't happen that often anyways.

NicolasDorier:
Sorry, actually seems fine. But I would remove "confirmed" or "fully paid".

Kukks:
> 
> 
> Sorry, actually seems fine. But I would remove "confirmed" or "fully paid".

I removed ReceivedPayment and PaidInFull

This way, you know that an invoice has been paid and confirmed correctly

NicolasDorier:
@Kukks anyway they will know by checking the invoice. What they want to know via notification is "wow I got money in"

pavlenex:
I agree that initial "wow I got money" is important, especially in retail, but for online businesses that use integration it won't matter. I guess once we add notification preferences it'll make more sense for certain type of users.

PR #1676, opened 2020-06-20
Change generate wallet passphrase to text input to fix autocomplete

Kukks:


pavlenex:
tACK.

@NicolasDorier I think we should merge this for the upcoming release, fixes important problem imo.

PR #1682, opened 2020-06-23
Fix 404 link

RiccardoMasutti:


NicolasDorier:
the tests should catch those kind of error... will need to take a look

pavlenex:
No because it's not a 404 link.

PR #1685, opened 2020-06-23
Check and fix broken anchor links.

dennisreimann:
In addition to #1682, this also checks URL fragments so that we also get to know about broken anchors.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -85,7 +85,7 @@ namespace BTCPayServer.Tests
             var views = Path.Combine(TestUtils.TryGetSolutionDirectoryInfo().FullName, "BTCPayServer", "Views");
             var viewFiles = Directory.EnumerateFiles(views, "*.cshtml", SearchOption.AllDirectories).ToArray();
             Assert.NotEmpty(viewFiles);
-            Regex regex = new Regex("href=\"(http.*?)[\"#]");
+            Regex regex = new Regex("href=\"(http.*?)\"");
             var httpClient = new HttpClient();
             List<Task> checkLinks = new List<Task>();
             foreach (var file in viewFiles)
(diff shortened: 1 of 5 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -155,16 +155,33 @@ private static async Task CheckLinks(Regex regex, HttpClient httpClient, string
 
         private static async Task AssertLinkNotDead(HttpClient httpClient, string url, string file)
         {
+            var uri = new Uri(url);
```
Kukks, line 158: Could be a malformed url that causes the Uri constructor to crash, so declare it out here and set inside the try block

Kukks, line 158:
```suggestion
            Uri uri = null;
```

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -155,16 +155,33 @@ private static async Task CheckLinks(Regex regex, HttpClient httpClient, string
...
+
             try
             {
-                using var request = new HttpRequestMessage(HttpMethod.Get, new Uri(url));
+                using var request = new HttpRequestMessage(HttpMethod.Get, uri);
```
Kukks, line 162:
```suggestion
                uri = new Uri(url);
                using var request = new HttpRequestMessage(HttpMethod.Get, uri);
```

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -155,16 +155,33 @@ private static async Task CheckLinks(Regex regex, HttpClient httpClient, string
...
+                    var contents = await response.Content.ReadAsStringAsync();
+                    Assert.Matches($"id=\"{fragment}\"", contents);
+                }
+
                 Logs.Tester.LogInformation($"OK: {url} ({file})");
             }
+            catch (Exception ex) when (ex is MatchesException)
+            {
+                var details = ex.Message;
+                Logs.Tester.LogInformation($"FAILED: {url} ({file}) – anchor not found: {uri.Fragment}");
```
Kukks, line 181:
```suggestion
                Logs.Tester.LogInformation($"FAILED: {url} ({file}) – anchor not found: {(uri?.Fragment??string.Empty)}");
```

BTCPayServer.Tests/UnitTest1.cs
NicolasDorier: That's fine if it crash, because it SHOULD crash.

NicolasDorier:
Merging this, I think that the point of crash @Kukks  mentioned are not problematic, as it will crash with proper error message.

Issue #890, opened 2019-06-15
Invoice Refund Option


pavlenex:
**Is your feature request related to a problem? Please describe.**
When a merchant wants to issue a refund to the customer (especially if the merchant is not auto-converting fiat, but hodling) issuing a refund is very difficult, especially if the service/product is priced in fiat.

Use case: Order total is $100, however one item is defective and merchants needs to issue a refund. The price for the item is $25.

To issue a refund currently, the merchant (if he's hodling) needs to refund $25 at the rate of the time of placing the order. This means that the merchant needs to calculate how much BTC at a certain rate he needs to refund, then go to the wallet and refund it. 

For me personally, this process takes 10-15min.

**Describe the solution you'd like**
Have a refund button in the invoice, which opens a pop-up, where merchant enters the amount in fiat/btc selects the rate from the drop down (current rate, invoice rate or custom rate) and customers refund address. When the refund pop-up is filled, proceed to wallet button auto-fills the wallet where a refund can be issued.

It would be nice if we'd have a refund field in the invoice itself that would auto-populate if the refund is issued.

**Provide examples**
![Screen Shot 2019-06-15 at 10 41 28](https://user-images.githubusercontent.com/36959754/59549198-404b0e00-8f5a-11e9-99a0-312038941611.png)

woutersamaey:
Can we also have this as an API call? This seams like a more advanced eCommerce case that I’d drive from the online store software (Magento in my case).

pavlenex:
@woutersamaey Sure, but the merchant would use their BTCPay internal wallet to send the funds, then we'll need API calls for that, but not sure how you can fit all of that into Magento's UI. To refund merchant would need to go to BTCPay find an invoice, sign it with a wallet and broadcast.

woutersamaey:
@pavlenex it would make sense to have a refund API call, which queues the refund for later signing.
In a larger organisation, a supervisor could sign all pending transactions at the end of the day while the API was used to queue these.

pavlenex:
Yes that's a great example, I agree, was not thinking big on this one.

Kidist-Abraham:
Is this feature added yet?

pavlenex:
No, otherwise it would be closed.

Kukks:
> It's just a bit tricky to handle it as the flow is a bit interactive
* User contacts merchant requesting refund/merchant decides to cancel order and refund
[...]

PR #1689, opened 2020-06-24
Implement invoice refund

NicolasDorier:
Build on top of #1687. This give the ability to the merchant to create a refund when the payment has been confirmed.

We should have more option in the future, right now you can select the payment method and the rate.

Ping @pavlenex 

Need unit tests.

image

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.PullPayments.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.PullPayments.cs
@@ -54,5 +54,10 @@ namespace BTCPayServer.Client
             var response = await _httpClient.SendAsync(CreateHttpRequest($"api/v1/stores/{HttpUtility.UrlEncode(storeId)}/payouts/{HttpUtility.UrlEncode(payoutId)}", method: HttpMethod.Delete), cancellationToken);
             await HandleResponse(response);
         }
+        public async Task<PayoutData> ApprovePayout(string storeId, string payoutId, ApprovePayoutRequest request, CancellationToken cancellationToken = default)
+        {
+            var response = await _httpClient.SendAsync(CreateHttpRequest($"api/v1/stores/{HttpUtility.UrlEncode(storeId)}/payouts/{HttpUtility.UrlEncode(payoutId)}", bodyPayload: request, method: HttpMethod.Post), cancellationToken);
+            return await HandleResponse<PayoutData>(response);
+        }
     }
 }
--- a/BTCPayServer.Client/Models/ApprovePayoutRequest.cs
+++ b/BTCPayServer.Client/Models/ApprovePayoutRequest.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BTCPayServer.Client.Models
+{
+    public class ApprovePayoutRequest
+    {
+        public int Revision { get; set; }
+        public string RateRule { get; set; }
+    }
+}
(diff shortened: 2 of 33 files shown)


pavlenex:
Impressive work.

### Bugs

- When creating a refund, it creates the amount of a `rate` for first two choices.
- When creating a refund in fiat, the amount is rounded 118.65 becomes 120.

### Notes

Few notes as I tested

- In invoice information, a link `Refund link` should be named `Issue refund` and possibly made as a button or at least add something like `fa-undo`
- We need UI for partial refunds. Something like [this ](https://help.pixieset.com/hc/article_attachments/115003678592/How_do_I_refund_my_client_customer_via_Stripe__3.jpg)`
- I wanted to suggest you rename the `Create refund` to `Issue refund` but technically create is appropriate term, thoughts?

### Ideas for the future

- (In the future) would be good to have a quick action refund from the invoice page, I think what Kukks did for notifications, that sort of UI would be great, multiple select invoices.

### Questions:
- How do you handle refund for `invalid invoices`?
- How is tx fee calculated into refunds? Is it on top of the amount or? May be a good idea to add `deduct fee tickbox`

pavlenex:
Closes #890

NicolasDorier:
* [x] We should show the refund button even if not confirmed, just disabled and with a tooltip explanation "You can only issue refunds on confirmed invoices"
*  ~Refund screen: we should have useful invoice details available on screen such as payment amounts, rates at payment time, etc~
Comment: This one will go away when we replace the pull payment page by something integrated with the checkout UI.

* Refund screen: Rename "What to refund?" to "Issue Refund" and then add a sub section header "Which refund amount calculation shoul be used?"
* [x] Refund screen: fiat option should be renamed to "the invoice currency, at the rate when the refund will be sent."
* ~Refund screen: when multi-currency payments are detected, add a back after after selection the refund payment method~
Comment: Let's do that at the same time we add partial refund later

* Payout screen: if you have a payout in a payment method that BTCPay cant handle (for example, MoneroLike or a BitcoinLike with a network that has `ReadonlyWallet` set to true, do not allow it to be processed and instead display a message to do the transaction manually ( we should be able to tell btcpay that the payout was handled manually)( we should track the payout address in nbxplorer individually and update the payout data based on that subscription's events).
* ~Pull payments really does not belong inside wallet, it is incrediby confusing when you have multiple payment methods and you are in the wrong wallet when trying to approve a payout of another payment method.~
Comment: Next release

* ~Pull Paymennts List: Payouts list should appear under each item since they may be in different wallets.~
Comment: Next release

* ~Payouts list screen: Next to Amount, also show the actual payment method amnount that will be sent~
Comment: Next release, but this is only decided after the payout has been "approved", which happen in the UI only when you click "Pay the selected payouts".

* ~Refund screen: Does not handle paying for tx fee. Usually customer is respnsible for paying refund fees (or should be an option).~
Comment: Next release, go with adding partial refund

* ~Need a way to do refunds in multiple iterations, for example if a customer paid in 2 payment methods, allow the merchant to refund both individually.~
Comment: The back end allow pull payment to be in different payment methods. Not the UI, anyway, we need to integrate the UI of refund to the checkout UI.

* [x] Optionally allow payment methods for refund payouts that are available but werent used by customer. (Maybe they paid with lightning, but want to get the refund on chain instead as their node is down)

About the Bug: For the time being, I will disallow refunds in currencies with read only wallets. Let's change this next release.

NicolasDorier:
@Kukks can you open an issue with the remaining problem unfixed you pointed out?

PR #1700, opened 2020-06-27
Set noindex, nofollow to ViewPullPayment.cshtml

RiccardoMasutti:
We shouldn't make Google (or other search engines) index and cache these pages on SERPs.

This exposes a number of security and privacy concerns:
1. Receiver's addresses and transactions are publicly visible
2. Sender's (pull payment) balance is publicly visible
3. An attacker could take advantage of Google's dork (eg. `inurl:/pull-payments/`) to receive a list of all the websites with indexed pull payments, and attack the latter by continuously requesting fake payments.
This could cause the sender to make a mistake and approve the wrong payment.

### DIFF
--- a/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
+++ b/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
@@ -12,6 +12,7 @@
 <head>
     <title>@Model.Title</title>
     <meta charset="utf-8" />
+    <meta name="robots" content="noindex, nofollow">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="apple-mobile-web-app-capable" content="yes">
     <link href="@Context.Request.GetRelativePathOrAbsolute(themeManager.BootstrapUri)" rel="stylesheet" />


pavlenex:
ACK, but note this is not a bulletproof solution at all, with `noindex, nofollow` we're just kindly asking search engines not to index it. There needs to be a better protection for pull payments for people who are concerned about the attack vector.

BTW sender's balance is never visible, it's the balance of a pull payment.

RiccardoMasutti:
> ACK, but note this is not a bulletproof solution at all, with `noindex, nofollow` we're just kindly asking search engines not to index it. There needs to be a better protection for pull payments for people who are concerned about the attack vector.
> 
> BTW sender's balance is never visible, it's the balance of a pull payment.

Yes, one solution would be to protect pages with a password.

pavlenex:
> Yes, one solution would be to protect pages with a password.
There was a similar discussion on this for [payment requests](https://github.com/btcpayserver/btcpayserver/pull/1126#issuecomment-550967479), curious if @NicolasDorier still has a strong stance against it.

PR #1707, opened 2020-06-30
Enable CORS and fix small doc error

Kukks:


dergigi:
Haven't tested it, but looks good to me :+1:

Issue #1710, opened 2020-07-02
Login page re-design


pavlenex:
**Is your feature request related to a problem? Please describe.**
I know we've recently re-designed our login page, but I think we can still do a better job on making sure it is consistent with our [design guidelines ](design.btcpayserver.org/).

**Describe the solution you'd like**
The login page on BTCPay Server is the front page user sees before logging in. It appears even before user creates an account. The goal of this page is to:

- Have a clear CTA (user needs to create an account / log in to access BTCPay)
- Welcome user and explain concisely what BTCPay Server is
- Show prominent supporters of the BTCPay Server Foundation and link for the rest
- If Tor is enabled, we'd like to show .onion access link (currently the way we show it may not be clear enough for regular folks at all, luckily if user already accesses btcpay over Tor [we'll notify them via Tor Browser](https://pbs.twimg.com/media/Ebj7QWNVcAAziAD?format=png&name=medium))

Problems with current login page:
- If user does not have an account, `create account` is not that prominently displayed (I've seen users asking questions about this one)
- Tor .onion url may seem confusing for those unfamiliar with it
- forgot password page does not look consistent with login page
- Page looks all over the place, it  could use some consistency and cleanup (imo)

**Current login page design**
![login1](https://user-images.githubusercontent.com/36959754/86332497-e4c5ee00-bc4a-11ea-82b2-3fb06b2b7714.PNG)
![login2](https://user-images.githubusercontent.com/36959754/86332740-379fa580-bc4b-11ea-8b89-a0c86076431f.PNG)
![forgot your password](https://user-images.githubusercontent.com/36959754/86332765-3e2e1d00-bc4b-11ea-93e8-35d5c2e1b19e.PNG)

**Provide examples**
There are mockups of different [iterations of the login page on Figma. ](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY)

![login](https://user-images.githubusercontent.com/36959754/86332060-3c178e80-bc4a-11ea-87c1-d3cc73fa8c07.PNG)
![login2](https://user-images.githubusercontent.com/36959754/86332063-3d48bb80-bc4a-11ea-8b95-f31ce01fe8b6.PNG)

Another thing, is that login page is available for several themes, while we don't need design for several themes, it's good to keep that in mind.

![casa](https://user-images.githubusercontent.com/36959754/86332894-6fa6e880-bc4b-11ea-870a-db105a0cc354.PNG)
[...]

PR #1712, opened 2020-07-02
Redesign login and register / refactor create user

dennisreimann:
Addresses #1710. While working on this I came across two more things I chose to refactor – want to get your opinion on this as these things are a bit more involved and might need some more work. So I'll start with the refactorings:

1. Unified the welcome and error layouts to a single LayoutSimple: They were almost identical and hence there was a lot of duplicated code I decided to dry down.
2. Dissecting the register (user facing) and create user (admin area) workflows, giving them individual actions and views. This part might need more work, I just wanted to get it done as quickly as possible, because there might have objections here. IMHO it makes sense to split them up, because these are two separate use cases – which was hinted at with variables like `returnUrl`, `useBasicLayout` and `logon`. But I don't know the reasoning behind this and maybe you think it's better to keep it that way.

Alright, with that out of the way, here's what it looks like now.

High-level changes as proposed in the issue:

- One column layout to provide a clear structure 
- Register option is now a secondary button (might make sense for login on register page too)
- Make Tor-Link less prominent ([reasoning](https://github.com/btcpayserver/btcpayserver/issues/1710#issuecomment-652969599))
- Keep supporters on the page but move them to the bottom

## Mobile
mobile

## Desktop

desktop

## Admin: New user

create-user

new-user

## Error Pages

 The error pages still look the same as before, hence I'm leaving them out here.


Eskyee:
Looking good 🔥

pavlenex:
Nice progress. 
Supporters are still below the fold for me.
I suggest we also wait for feedback from designers in #1712 before we decide to further implement this.

BTCPayServer/Controllers/ServerController.cs
```diff
@@ -303,6 +305,40 @@ public new async Task<IActionResult> User(string userId, UserViewModel viewModel
             return RedirectToAction(nameof(User), new { userId = userId });
         }
 
+        [Route("server/users/new")]
+        [HttpGet]
+        public IActionResult NewUser()
```
Kukks, line 310: You could just name this the same as the post action `CreateUser` (and also keep the same route)

dennisreimann: Thanks, done in 0235624c6424ec281a7e8970e84381f9e753dcf4.

dennisreimann:
Rebased and incorporated the feeedback. 

- Login / Sign Up are now secondary buttons below the form
- Tor URL is now a button, which should be familiar to Tor users
- As we cannot open the onion URL directly in Tor browseer, I'bve made the button copy the URL to the clipboard
- Decreased the general page margin a bit, so that supporters should be visible above the fold

register-dark

login-light

login-dark

pavlenex:
@dennisreimann Not a big fan of that Tor button, it looks like a main CTA. I again suggest waiting for feedback in #1710 regarding login page before proceeding further. We'd have to swap that font for example.  Then `BTCPAYSERVER SUPPORTERS` 😠 at the bottom should be changed. No need to have 2 links in form of tool tip, etc.

dennisreimann:
After sleeping over it I agree regarind the position. Curious for the design feeeback in the issue, so that wee can finish this up :)

pavlenex:
It'll take some time to get the ball rolling in there as there's only @dstrukt working on it. Will try to get him some help.

dstrukt:
I agree @dennisreimann about the tor button feeling like a prominent CTA, which is why I opted for the icon + text to push it back in importance.

bolatovumar:
Are we limiting the number of sponsors that can appear on the login page at once to six?

dstrukt:
I think we should, but have them cycle through (assuming we have as many as 12 or 18 at any given time).

I'll provide an animation.

pavlenex:
> 
> 
> Are we limiting the number of sponsors that can appear on the login page at once to six?

Only a certain tier of supporters is displayed on the front page. So we're not limit it.

dennisreimann:
> I think we should, but have them cycle through (assuming we have as many as 12 or 18 at any given time).

Yeah, maybe, but let's not overcomplicate it and we can do this in another PR. Let's get this one finished up and improve it once there's a need for this.

dstrukt:
> > I think we should, but have them cycle through (assuming we have as many as 12 or 18 at any given time).
> 
> Yeah, maybe, but let's not overcomplicate it and we can do this in another PR. Let's get this one finished up and improve it once there's a need for this.

Agreed, it's a scaling solution, if there's no need at this moment, I support this course of action.

dennisreimann:
@NicolasDorier this isn't completely finished yet, but I'll incorporate the feedback in another PR.

NicolasDorier:
1. Kukks approved
2. No WIP
3. Awesome
4. No @pavlenex bitching
5. Tests pass

Instant merge!

pavlenex:
Kukks's approved it code-wise we were in the middle of discussing the design and UI.

NicolasDorier:
Sorry boss! Github should not make the green button so enticing.

Kukks:
Maybe we should start using the draft PR option more :P

Issue #1406, opened 2020-03-22
Errors in dev env on MacOS in launchSettings.json


pavlenex:
**Describe the problem/bug**
When a dev enviroment is set up on MacOS, there is an issue with `launchSettings.json`.

When you build a project, it will automatically alter your `launchSettings.json` and then it's considered as a change in the PR

**Your environment** 
* Version of BTCPay Server: Dev env
* Deployment method: Docker, v3.1.102 SDK, GitHub Desktop

**To Reproduce**
1. Build a project on Mac.
2. See that it changes parameters in `launchSettings.json` that never happens on Windows.

**Expected behavior**
Project should build without any changes.

**Actual behavior**
Building project changes file parameters.

**Screenshots/Links**
![Screenshot 2020-03-22 at 18 00 04](https://user-images.githubusercontent.com/36959754/77255348-fc163800-6c66-11ea-8e1d-bbc05a03ec76.png)

![Screenshot 2020-03-22 at 18 08 22](https://user-images.githubusercontent.com/36959754/77255484-261c2a00-6c68-11ea-8334-227595d54303.png)

**Additional context**
Affected files from the screenshot above:

```
@ -50,4 +50,7 @@
    <ProjectReference Include="..\BTCPayServer\BTCPayServer.csproj" />
  </ItemGroup>

  <ItemGroup>
    <Content Remove="Properties\launchSettings.json" />
  </ItemGroup>
</Project>
```
```
  <PropertyGroup Condition=" '$(RunConfiguration)' == 'Docker-Regtest' ">
    <StartAction>Project</StartAction>
    <EnvironmentVariables>
      <Variable name="BTCPAY_NETWORK" value="regtest" />
      <Variable name="BTCPAY_LAUNCHSETTINGS" value="true" />
      <Variable name="BTCPAY_BUNDLEJSCSS" value="false" />
      <Variable name="BTCPAY_LTCEXPLORERURL" value="http://127.0.0.1:32838/" />
      <Variable name="BTCPAY_BTCLIGHTNING" value="type=clightning;server=tcp://127.0.0.1:30993" />
      <Variable name="BTCPAY_BTCEXTERNALLNDGRPC" value="type=lnd-grpc;server=https://lnd:lnd@127.0.0.1:53280/;allowinsecure=true" />
[...]

PR #1717, opened 2020-07-07
Update BTCPayServer.csproj

Eskyee:
this solves the Old Visual Studio Bug Problem in a Mac <Content Remove="Properties\launchSettings.json" /> 
plus EnvironmentVariables like Variable name=
could you check this please for me @britttttk on a Mac and could confirm if this is correct and has not introduced any errors or bugs.  @NicolasDorier 
this is tested on a fresh copy of Mac install with VS Studio, and BTCPay Server running in my environment , with no problems.
happy if this can solve that old Mac VS bug....  in launchSettings.json & which @pavlenex @rockstardev could reproduce this old error.

https://github.com/btcpayserver/btcpayserver/issues/1406

https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Properties/launchSettings.json

Screenshot 2020-07-07 at 03 15 02

Screenshot 2020-07-07 at 03 20 54

Screenshot 2020-07-07 at 04 09 40

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Web">
   <Import Project="../Build/Version.csproj" Condition="Exists('../Build/Version.csproj')" />
   <Import Project="../Build/Common.csproj" />
   <PropertyGroup Condition="'$(Configuration)' == 'Debug' And '$(RazorCompileOnBuild)' != 'true'">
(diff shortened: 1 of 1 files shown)


Eskyee:
Ok yes this hardcoded profile you talk about is from the btcpayserver property’s folder, the file is called launchsetting.json , ok this no problem for this try at my attempting to fix our problem, even though my builds was clean showing no errors!! have you ever tried to do developing on a Mac with Visual Studio,  ?? Because then you would understand our simple old bug, and help find a solution because this is not standard to have this error in a project... from big company’s like Microsoft and Apple for so long.

Yes this does not affect Visual studio code or Rider... so can you explain... Might ?? Affect other IDE’s 
I can give more reference to this launch json bug if needed..

Developing on a Mac with free Visual studio shouldn’t be such a pain... 
Rider cost money.. but show no bug 😂 with launch setting json ..

How can we get past this bug error on Visual studio correctly... I made a fresh wipe just to test this out.. 

Cheers

Eskyee:
@rockstardev 🙌 any ideas on this bug problem.. it’s a nightmare to live with

rockstardev:
@Eskyee can you try checking out branch `fix/json-indent` in PR #1731?

If that fixes problem for you I think it may be better approach and we can keep that 2 space as convention going forward.

Eskyee:
after researching more I can't find anything to say we can't have this choice.
https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-build-the-same-source-files-with-different-options?view=vs-2019

https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-concepts?view=vs-2019

out of 140 local tests about 20 failed, which was lightning etc. etc, no biggie, going to run more tests with original config, and test against my idea to fix this bug, see where this error hides.

I will keep trying to narrow down. the problem.

Screenshot 2020-07-11 at 21 26 02


yes I can test in a while @rockstardev

rockstardev:
We will resolve this as part of #1731

Eskyee:
thanks for the fix @rockstardev

PR #1734, opened 2020-07-12
Maintain invoice list params on navigation, part 2

rockstardev:
Continuation of: #1673

Resolves: #1498

This will just setup Invoice link at the top as was the original suggestion. If @NicolasDorier doesn't have any complaints I'll apply this pattern to other grids. Optionally we can make this configurable in Server Policies (or maybe provide new, per-user setting page).


rockstardev:
Tests finally passing.

NicolasDorier:
let me give a try. I don't like having to change all links to the `invoices/` page like you do in layout.

rockstardev:
Obsoleted by: #1749

PR #1742, opened 2020-07-15
Allow mass-archiving of invoices

### LINKED ISSUE #1709: Multiple selection and batch action of invoices on the invoice page
In 1.0.5.0 we've added a way to select multiple notifications on the notifications page and perform a batch action. I think that this can be handy thing to add for invoices as well.

- Add a way to select multiple invoices 
- Add a way to perform a batch action
- Add archived invoices as an action

**Provide examples**
invoices

**Additional context**
Personally I couldn't remember any actions besides `archive` that would be useful for this, perhaps a `refund` but at this point it's probably too complex and I'd let refunds get a bit more feedback from users first.

bolatovumar:
closes #1709

Screen Shot 2020-07-14 at 8 00 51 PM

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -290,6 +290,24 @@ namespace BTCPayServer.Controllers
             return RedirectToAction(nameof(invoice), new { invoiceId });
         }
 
+        [HttpPost]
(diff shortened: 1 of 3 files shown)


NicolasDorier:
dope. I think we should have consistent UX. Pull Payments mass actions and notifications page should have the same kind of option.

NicolasDorier:
who be cool to have a selenium unit test for that

Issue #1645, opened 2020-06-09
Show recovery seed in a better way when creating a new wallet


pavlenex:
**Is your feature request related to a problem? Please describe.**
Currently, when you create a new wallet, recovery seed is shown on top of the Store Settings page, which can easily be disregarded. I've noticed that due to wallet incompatibilities, more users are now using `create a wallet`, that said we should make sure to educate them properly on the risks and create an UI that will make it clear that backup is crucial and that unless a hot wallet is used their seed is wiped out.

![Capture](https://user-images.githubusercontent.com/36959754/84124185-2643ee80-aa3b-11ea-8475-ca24bc505062.PNG)

**Describe the solution you'd like**
Create a separate page which will show first of all a short message like we have in [docs](https://docs.btcpayserver.org/CreateWallet/)

```WARNING
Securely back up your seed by writing it down and keep it in a safe place. Do not photograph it or store it in a digital format. When creating a wallet, the recovery seed will be shown only once and wiped from the server. If you loose your recovery seed, you'll lose the access to all of your funds.
```

**Provide examples**
![Seed](https://user-images.githubusercontent.com/36959754/84123840-ae75c400-aa3a-11ea-9e75-954e782c2230.png)

**Additional context**
I'd like to open this issue for a discussion. Perhaps we should consider having a seed confirm page to make sure users have properly backed up their mnemonic but also warn them about passphrases.

The separate page idea is because of the the changes in #1055 we plan to do eventually.
Opening this one for the future.

Eskyee:
Perhaps we should consider having a seed confirm page to make sure users have properly backed up their mnemonic but also warn them about passphrases.

Yes I agree @pavlenex 
Very important for noobs to understand this seed confirm point. as they could lose their money in the future, if they forgotten.  
mnemonic, passphrases.

dennisreimann:
I agree with @Eskyee and think having a separate page as proposed above is the way to go. This way we'd focus the users attention on the matter and would also have the attention and space to provide some more information around the topic (plus link to further information in the docs).

pavlenex:
I agree with having a separate page for this one. Actually it seems this relates to #1703. 
[...]

PR #1745, opened 2020-07-15
Recovery seed

dennisreimann:
Addresses #1645 by using a separate page for the recovery seed.

Also sprinkeled some UI improvements on the UpdateStore, DerivationScheme and Wallet Settings views.
I got some more ideas on how we could improve the UI/flow further, but we can tackle that in another PR after this one.

Things that should be discussed:
- Does this belong in `HomeController`? Because of the required policies I couldn't make it work in the `WalletsController`.
- Should the passphrase also be shown?

light

dark


Eskyee:
very nice @dennisreimann

bolatovumar:
Couple of minor comments:

1. Regarding "non air-gapped digital format" phrase. Novice users may not understand what this means at all. Probably safer to use really simple language here like "do not store this phrase on an internet-connected device" or something like that.
2. Because the distance between columns is larger than between rows, my eye wants to read from top to bottom and then to left (see below) but the numbers (1, 2, 3, etc...) go across, not down.

87590541-38264a80-c6e7-11ea-8018-aa201c71066d

See how Trezor does it, for example:

Seed_card_example

dennisreimann:
@bolatovumar good points. I'm with you on both, so let's see if we can improve the explanation and I'll tinker with the word list.

NicolasDorier:
so happy about this one.

pavlenex:
I would add, if wallet is created and `not a hot wallet`, let user know that the seed will be wiped from the server. This is really important.

dennisreimann:
> I would add, if wallet is created and not a hot wallet, let user know that the seed will be wiped from the server. This is really important.

yep, this is already [part of the template](https://github.com/btcpayserver/btcpayserver/pull/1745/files#diff-aa5e1e39fdec52aebd9d6e5a047d53d1R23).

GBKS:
Looking good. I attached a rough mockup with some suggestions:

1. Remove the top banner (if the extra statement about the wallet being generated is important, maybe implement it into the body text?)
2. Replace warning/error icon with an info icon (nothing is broken on this page, but there's important information)
3. Center the layout consistently

btcpay-seed-200716

I also like the suggestions about simplifying the language and flipping rows and columns.

pavlenex:
On a side note, I noticed this tackles derivation scheme import. Two things on that:
Please do not change the position of the button (it's already been documented) and such change is not bringing anything useful in terms of UX, and if possile make it more prominent, right now it seems like a secondary button. (grey)

dennisreimann:
@pavlenex You mean the screenshots like on [this page](https://docs.btcpayserver.org/CreateWallet/), right? 
We can separate out the changes unrelated to the workflow from this PR – as i said I got some more ideas for improvements. 

All in all, the terms and explanations on the wallet-related screens are too technical imho. There's hardly something that feels self explanatory or doesn't need further information. I think it's also a smell that we have to rely so heavily on screenshots in the docs. I'd love to go through this and clean it up as best as we can, because using so many screenshots will also prevent us from doing incremental improvements there.

pavlenex:
Yes. It's well documented for now and changes like this barely bring anything in terms of UX improvement. Our long-term goal is to completely re-work the workflow of setting up a store/wallet, and I'd leave these changes for that time.

Let's focus on recovery seed for now. Wallet page is another story that again first needs designers and UX view before it can be tackled.

For this particular PR, it seems the consensus on the modifications is:  (if I missed anything anybody feel free to correct me)

- Improve the text on the main body to be more human-friendly without too much technical terms.
- Reduce padding between columns 
- Use info icon instead of warring
- Center the layout

Optional suggestions:
- Use write down instead of save in the headline
- Use write down instead of save in the body
- Top banner modification (I actually think it's okay to leave it here because it makes the transition more natural and user is aware that his action is complete so my suggestion is we leave top banner for now)

dennisreimann:
> Our long-term goal is to completely re-work the workflow of setting up a store/wallet, and I'd leave these changes for that time.

Ok, I removed the commit with general changes and added another one that incorporates the suggested changes. Thanks for the valuable feedback everyone! 👍

The word list columns depend on the different screensizes – I've made it so that there's enough space to get that this is meant to be read in columns. See the screenshots below. (Mobile is just one column, left it out here)

> We'll probably need two versions of paragraph 2. When hot wallet is enabled and when it's not.

This last screenshot shows how it looks in casee the seed isn't stored. There's a separate paragraph in that case.

---



---



---

pavlenex:
@dennisreimann I think we need a better distinction between a hot wallet vs not a hot wallet messages, will try to think of something tomorrow. As mentioned, avoid air-gaped nobody knows what that means, it's hard to translate once we add localization, simple language is better.

pavlenex:
Posting screenshots again, since Dennis, double-posted the same one. Others are my suggestions.

## Current state 
### Wallet Generation when hot wallet is disabled
recoveryseed
### Wallet Generation when hot wallet is enabled
recoveryseed2

## My suggestions: 

In general, it's great progress so far. I really like the design and in particular I'm found of us having 2 separate messages for hot wallet vs non hot wallet, because I think it's essential piece of info. I feel like we maybe have way too much text or at lest we should better position it so that mnemonic words are in the center of the attention.

### Wallet Generation when hot wallet is disabled
- **I think the biggest mistake people can make with wallet generation is generate a wallet, do not back up or lose backup, receive payments, - they are rekt. We need to prevent this with very precise message.** So calling it a backup on this page may be a mistake. 
The text I suggest is:
```
The recovery seed allows you to access and restore your wallet. 
If you lose it or write it down incorrectly, you will permanently lose access to your funds. 
Do not photograph the recovery seed and do not store it digitally.

The recovery phrase will permanently be erased from the server.
```

The emphasis is on "**access** and restore your wallet".

- Change `The recovery phrase will only be shown before being wiped from the server.` to `The recovery seed will be erased from the server after you click Done`. If you do not back up properly, you'll lose access to your funds permanently. 
1recoveryseed


### Wallet Generation when hot wallet is enabled
The text I suggest is:
```
The recovery phrase is a backup that allows you to restore your wallet in case of a server crash.
If you lose it or write it down incorrectly, you may permanently lose access to your funds.
Do not photograph it. Do not store it digitally.

The recovery phrase will also be stored in a server as a hot wallet.
```
- Here I wanted to emphasize the word "backup" so people are aware that if their server goes down or crashes they MAY loose their funds if they don't have a backup.
2recoveryseed

As you can see my suggested layout re-positions text and modifies text slightly to better fit each use case. It's not perfect so any feedback regarding text and how to better convey a message is really appreciated, I think we really need to think careful about this, since user-funds are at risk if we don't coney the importance of each use-case.

dennisreimann:
@pavlenex I like the repositioning and think the text is good as it is. If there's no other feedback I'd implement it as suggested and we're good to go in my opinion. 

All in all this is a major improvement on what we currently have.

dennisreimann:
Updated and pushed.

## Hot Wallet enabled
hot

## Hot Wallet disabled


non-hot

pavlenex:
tACK. This looks good to me. I'll leave it to others to submit feedback if any. Great work so far everyone! 🚀

dennisreimann:
ok, let's first get #1752 done, then I'll take care of the tests and rebase here. I think they are failing because some layout stuff changed and it'll be easier to fix this one the login PR is merged. (It contains updates for the same layout files involved here)

dennisreimann:
Updated the views, which fixed the tests – nevertheless, let's first merge the login PR, then rebase this one and we're good to go.

NicolasDorier:
no rebase yolo merge

PR #1749, opened 2020-07-17
Save the latest search on invoices

NicolasDorier:
@rockstardev @dennisreimann 

Alternative to #1734

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -29,6 +29,7 @@ using Microsoft.EntityFrameworkCore;
 using NBitcoin;
 using NBitpayClient;
 using NBXplorer;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using StoreData = BTCPayServer.Data.StoreData;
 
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -597,12 +598,40 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody] Upd
...
+            public int? TimezoneOffset { get; set; }
+            public string SearchTerm { get; set; }
+        }
+
         [HttpGet]
         [Route("invoices")]
         [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie)]
         [BitpayAPIConstraint(false)]
-        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int timezoneOffset = 0)
+        public async Task<IActionResult> ListInvoices(string searchTerm = null, int skip = 0, int count = 50, int? timezoneOffset = null)
```
rockstardev, line 611: I don't like nullable here considering it results in all extra checks. I think you really added it to because of InvoicePreference.TimezoneOffset to save space in serialized JSON... I would rather move it logic there... create overload constructor and pass SearchTerm and TimeZoneOffset in, setting it to null if TimeZoneOffset is zero.

rockstardev: If you agree in principle (or don't have strong opinion) - I'll do that as part of refactor: https://github.com/btcpayserver/btcpayserver/pull/1749#pullrequestreview-450515311

PR #1752, opened 2020-07-17
Login/register view improvements

dennisreimann:
Continuation for #1712 – also addresses #1710.

Consistently aligned the texts centered and moved the Tor URL button below the form. I went with a button with a light outline, because copying the URL is a button action rather than a link.

ping @pavlenex @dstrukt 

light

dark

This is what the buttons looks like highlighted.

button


pavlenex:
Good idea about the button coloring, less focused and better. 👍 

We still need to clean up bottom section. It's BTCPay Server not `BTCPAYSERVER SUPPORTERS` and that tooltip makes no sense to me anymore.

@dstrukt What's your view on:
- Tor Button (position and colors)
- Supporters section, how can we clean it up
- We really need to change that font and perhaps iron out the text, the font looks really bad imo.

dennisreimann:
This is what it'd look like using our default font Open Sans.

login

Kukks:
I would prefer the tor button not taking up the premium vertical spacing
all on its own. I preferred it in the login box aligned with something
else. The new color scheme makes it less of a hardcore CTA so it could work

dennisreimann:
@kukks I agree as it would also be a better position in regards to the reading and action flow, becausee you'd want to switch to Tor before  maybe filling out the form and finding the button below.

I'd make it a small button and leave out the border then, to make it not too prominent though (see bordered version below). I think I like this best.

Sidenote: These screenshots also contain the non-uppercased version of the headline, so that we can discuss this too.

nb-dark

nb-light

Highlighted:

highlighted

---

Alternativ: Button with border (not as good imho)

login

NicolasDorier:
fuck 

image

dennisreimann:
@NicolasDorier Pardon, sir?

pavlenex:
Draft is dorieryolomerge protection mechanism. 😄

dennisreimann:
Yes, we learned our lessons here 😂

pavlenex:
@dennisreimann Can you push latest changes to the PR?
I like first two examples you mentioned [here ](https://github.com/btcpayserver/btcpayserver/pull/1752#issuecomment-660442084)and would like to preview them in dev env to get a better feeling.

One note from viewing the screenshots:
- If we're okay with Tor button in the top right of the form, I'd center align the "don't have an account" and "sign up" button as well as `forgot your password link" to make it consistent with the layout we have.

dennisreimann:
@pavlenex I can push this later today, wanted to wait for @dstrukt's feedback to get to something that's almost final ;)

dennisreimann:
Here's the centered version. I'd also propose making the supporter logos a tad bit smaller (60 vs. 75 px height), because otherwisee they weigh too heavy imho. Don't know if that's ok though?!

centered

Zaxounette:
I haven't seen anyone comment on this particular part in either of the issues or this PR, so i'll go with it:

The `Is Administrator` checkbox on the `Create Account` page.

If I recall correctly, the very first account created after server setup is automatically an admin and for futur administrators, the first on can create the accounts in `Server Settings`, So why the checkbox ?

Am I missing something ?

dstrukt:
@Zaxounette The "is Admin" is only on this mock AFAIA, it's not what is live to your average person.

---

Suggestions:

- Remove the green dash above and below. It's not add much in terms of visual division, it's just adding noise (imo) and height to the view.

- Change "BTCPay Server Supporters" to "Our Supporters" ..it's cleaner, we already know the product is from the title, no need to be redundant.

- Reduce the wording for the "Don't have an account?" and "Login" bits (refer to mock)

- The borders around and in the form elements are too dark (will edit to provide colors). It makes the elements look a bit "aged", less modern.

- Needs more padding (we're using a spacing system, correct? I would double it) between the sponsor product logo and the title of the sponsor company.

- Reduce the logos to 50 x 50px (or 48 x 48 .. not sure what sizing system we're working with, mine are 50 in the mock), and add more spacing in between them so that the entire component stays centered. They're currently too large.

- View all supporters should probably be a green link or at least green on hover.

---

Opinions:

- The copy Tor URL button looks better without the outline. I understand that there is an action behind this without that additional visual information, and it's a fairly common pattern. You can make the cursor a pointer element, and could even change the text to "Copied URL" when pressed to indicate button action.

- Similarly, the "sign up" button could easily be a text link on the create account page. That's a very common pattern and would actually look better (imo) and cleaner. 

- I know we also list the description in both views, but does it make sense to keep it in the Login view? If the user has signed up and created the account they probably already have an idea of what this product is...unless we're trying to beat it into their head.

--- 

I started to mock up the login page last night and update the sign up screen to show a really clean view, but it's been a busy week and I'm not 100% done incorporating everything -- the Tor element is an outlier.

This is WIP, revisiting a few things today, but I hope my suggestions can help push some of this further.


Login01
Login02
Signup01
Signup02

pavlenex:
@Zaxounette that is only for dev env.
Spot on feedback as always @dstrukt 🔥 I agree with everything pointed out, way cleaner, except for 5 supporters. We currently have 6 and can't really limit it to a fixed number.

dennisreimann:
Updated and pushed.

## Login

login-light

login-dark

## Register

register-light

register-dark

## Error pages

Also adapted them, as they are using the same layout.

error-light

error-dark

pavlenex:
tACK 🚀

dstrukt:
Damn, this is looking great @dennisreimann! Nice work!

Only small nit-picky things from me would be:

- the left and right padding around the form element
- the text under the main CTA (signup, etc..) for the secondary action is a bit small, might want to step it up to an equivalent value (from 14pt to 16pt?)

Otherwise, looks good to me, excited to see this live! 

Might have some additional nit-picky things then too :)

dennisreimann:
Thanks for the feedback – done :)

So fresh, so nice :D

register

login

dennisreimann:
@NicolasDorier It's all yours now, smash that button 😆

bolatovumar:
Tested it locally on my end and looks good.

NicolasDorier:
FINALLY

PR #1753, opened 2020-07-19
Allow sorting apps by store, name or app type

### LINKED ISSUE #1568: Apps - Sort by name (Store-Name-Apptype)
In the /apps page, add the possibility to sort apps by :

- Store name
- App name
- App type

by adding a 🔽 before or after : Store - App name - App type 

I think you can also do it on the /stores page with the Name & Website menus.

Thank you.

bolatovumar:
See in action:

Jul-19-2020 14-14-25

Also has helpful tooltips on hover:

Screen Shot 2020-07-19 at 2 08 07 PM
Screen Shot 2020-07-19 at 2 08 29 PM

I modelled the behavior after the macOS sort functionality in Finder:

Screen Shot 2020-07-19 at 2 25 26 PM

close #1568

### DIFF
--- a/BTCPayServer/Controllers/AppsController.cs
+++ b/BTCPayServer/Controllers/AppsController.cs
@@ -60,11 +60,11 @@ namespace BTCPayServer.Controllers
                     {
                         switch (sortOrderColumn)
                         {
-                            case "Name":
+                            case nameof(app.AppName):
                                 return app.AppName;
-                            case "Store":
+                            case nameof(app.StoreName):
                                 return app.StoreName;
-                            case "AppType":
+                            case nameof(app.AppType):
                                 return app.AppType;
                             default:
                                 return app.Id;
--- a/BTCPayServer/Controllers/AppsController.cs
+++ b/BTCPayServer/Controllers/AppsController.cs
@@ -47,9 +47,42 @@ namespace BTCPayServer.Controllers
 
         public string CreatedAppId { get; set; }
 
-        public async Task<IActionResult> ListApps()
+        public async Task<IActionResult> ListApps(
+            string sortOrder = null,
+            string sortOrderColumn = null
+        )
         {
             var apps = await _AppService.GetAllApps(GetUserId());
+
+            if (sortOrder != null && sortOrderColumn != null) 
+            {
+                apps = apps.OrderByDescending(app => 
+                    {
+                        switch (sortOrderColumn)
+                        {
+                            case "Name":
+                                return app.AppName;
+                            case "Store":
+                                return app.StoreName;
+                            case "AppType":
+                                return app.AppType;
+                            default:
+                                return app.Id;
+                        }
+                    }).ToArray();
+
+                switch (sortOrder)
+                {
+                    case "desc":
+                        ViewData[$"{sortOrderColumn}SortOrder"] = "asc";
+                        break;
+                    case "asc":
+                        apps = apps.Reverse().ToArray();
+                        ViewData[$"{sortOrderColumn}SortOrder"] = "desc";
+                        break;
+                }
+            }
+            
             return View(new ListAppsViewModel()
             {
                 Apps = apps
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/AppsController.cs
```diff
@@ -47,9 +47,42 @@ public partial class AppsController : Controller
...
         {
             var apps = await _AppService.GetAllApps(GetUserId());
+
+            if (sortOrder != null && sortOrderColumn != null) 
+            {
+                apps = apps.OrderByDescending(app => 
+                    {
+                        switch (sortOrderColumn)
+                        {
+                            case "Name":
```
Kukks, line 63: better to use `nameof(app.AppName;)` for example so that if the prop is renamed or removed, it would not compile and make you fix it before it gets merged

BTCPayServer/Controllers/AppsController.cs
```diff
@@ -47,9 +47,42 @@ public partial class AppsController : Controller
...
+                                return app.AppType;
+                            default:
+                                return app.Id;
+                        }
+                    }).ToArray();
+
+                switch (sortOrder)
+                {
+                    case "desc":
+                        ViewData[$"{sortOrderColumn}SortOrder"] = "asc";
```
Kukks, line 77: Maybe an enum fits better here?

dennisreimann:
I think it makes sense to use this pattern everywhere, if @rockstardev can properly standardize it. (and I have no doubt that he can, because why wouldn't he? He's our rockstar.)
Being consistent would not just make it easier to understand from a users perspective, but it'd also make the code easier to grasp when all those list action use the same patterns.

bolatovumar:
As far as I see it there are two places where you would want to have sorting:

1. Apps page
2. Invoices page

Other pages (like "Stores", for example) probably won't have enough items to warrant sorting.

BTCPayServer/Controllers/AppsController.cs
bolatovumar: @Kukks addressed here: 25527ec1dd99712d438353c4cb9cc9684514f24e

BTCPayServer/Controllers/AppsController.cs
bolatovumar: @Kukks you mean enum for `"asc"` and `"desc"` or for `$"{sortOrderColumn}SortOrder"`?

rockstardev:
I'm so touched by your belief in me @dennisreimann & @bolatovumar. I promise not to let you down.

rockstardev:
@bolatovumar I'm waiting on @NicolasDorier to review #1754 and merge. Once he does that I'll just rebase this on top of that and take a look. To me it definitely makes sense to at least extract sorting logic out of controllers... generalization or no generalization.

And same for the `<a>` elements - typing all that stuff for every column is kinda cumbersome - would be much better to abstract it one level above... because everything that's different is `Name` and `NameSortOrder` key between elements.

rockstardev:
And yeah - when I was saying "generalization" - I was more meaning in the direction of having a standard UI pattern (what @dennisreimann is saying).

bolatovumar:
@rockstardev cool, makes sense.

NicolasDorier:
I am not fan of generalization either.

As then they may decrease the time we spend changing something, but require more effort of communication to be on the same page, increase the knowledge needed to fix something and increase the chance of collateral damage on something else.

For generalization of UI pattern, I agree on simple things. Maybe some specific stuff can make better use of ASP.NET Core components for better reuse. (For example, the pager could be a component with a specific viewmodel that we reuse everywhere)

NicolasDorier:
@bolatovumar I am not fan of using ViewData for that, I think it should go on a ViewModel that we may be able to reuse elsewhere for controller side sorting. 

Though I will merge, can try to fix that after.

bolatovumar:
> Also maybe for some table, a sorting at the JS level may make more sense than at the controller level, so it is far from easy to generalize.

I was initially planning on doing this in JS but as @Kukks said in the original issue for this PR (https://github.com/btcpayserver/btcpayserver/issues/1568#issuecomment-629965636) we have to do it on the backend since we have pagination here and we can't do sorting across multiple pages with JS (well, without creating an API endpoint to return data and whatnot, at least).

PR #1754, opened 2020-07-19
Cookie Preferences Generalization, applying same style to ListInvoices and Payment Requests

rockstardev:
Any page that needs preservation of `searchTerm` now can leverage that logic with unique cookie key. Here is how it looks for Payment Requests now:

New Payment Requests look

I've also extracted paging table footer into partial, now usable on any page that has model inheriting from `BasePagingViewModel`

Resolves: #1775


BTCPayServer/Views/PaymentRequest/GetPaymentRequests.cshtml
```diff
@@ -10,18 +10,43 @@
...
         }
         <div class="row">
             <div class="col-lg-12 section-heading">
                 <h2>Payment Requests</h2>
                 <hr class="primary">
+                <p>Create, search or pay an payment request.</p>
             </div>
         </div>
+
+        <form asp-action="GetPaymentRequests" method="get" style="float:right;">
```
Kukks, line 25:
```suggestion
        <form asp-action="GetPaymentRequests" method="get" class="pull-right">
```

BTCPayServer/Views/PaymentRequest/GetPaymentRequests.cshtml
```diff
@@ -10,18 +10,43 @@
...
+                        <a class="dropdown-item" href="/payment-requests?SearchTerm=">Unfiltered</a>
+                    </div>
+                </div>
+            </div>
+            <span asp-validation-for="SearchTerm" class="text-danger"></span>
+        </form>
+
         <div class="row button-row">
-            <div class="col-lg-12">
+            <div class="col-lg-24">
```
Kukks, line 49: bootstrap grid only handles 12 columns

rockstardev: @Kukks we have this on ListInvoices.cshtml as well (https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Invoice/ListInvoices.cshtml#L91), resulting in button that's left aligned with title. 

If you know how to destroy col padding in better way let me know (or @dennisreimann) otherwise I'll waste my precious rockstar time to Google for proper way to say `no-left-padding` with bootstrap class.

BTCPayServer/Views/PaymentRequest/GetPaymentRequests.cshtml
rockstardev: 1b9bdb8c1740e4b012d42951ef8a4a16dfc8b60a

BTCPayServer/Views/PaymentRequest/GetPaymentRequests.cshtml
dennisreimann: That's just `pl-0` 😀

rockstardev: I'll try to remember that.

2da3c96

NicolasDorier:
Instead of using one different cookie per preference, I would prefer that we design instead a single cookie which wrap a *glup* json blob that anybody can edit to put their own preferences.

BTCPayServer/Controllers/Logic/ListCookiePreference.cs
```diff
@@ -0,0 +1,109 @@
...
+        InvoicesQuery, PaymentRequestsQuery
+    }
+
+    public class UserPrefsCookie
+    {
+        public ListQueryDataHolder InvoicesQuery { get; set; }
+
+        public ListQueryDataHolder PaymentRequestsQuery { get; set; }
+
+        internal ListQueryDataHolder GetSection(UserPrefCookieKeys key)
```
NicolasDorier, line 69: Why having an enum and a fancy Get/SetSection instead of just setting properties directly?

BTCPayServer/Controllers/Logic/ListCookiePreference.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Controllers.Logic
+{
+    // Classes here remember users preferences on certain pages and store them in unified blob cookie "UserPreferCookie"
+    public class ListCookiePreference
+    {
+        public static void Parse(ControllerBase ctrl, UserPrefCookieKeys key,
```
NicolasDorier, line 14: just make a Extension method for ControllerBase. With a Get/Set for UserPrefCookieKeys

BTCPayServer/Controllers/Logic/ListCookiePreference.cs
rockstardev: 19ffd03

BTCPayServer/Controllers/Logic/ListCookiePreference.cs
rockstardev: 19ffd03

NicolasDorier:
why are you making it so complicated with reflection and shit. One sec, refactoring it.

NicolasDorier:
Ah ok I know why now....

NicolasDorier:
Fuck those hacks, I have cleaner way by using a view model instead of several properties. It should work.

NicolasDorier:
The magic sauce is `[DisplayFormat(ConvertEmptyStringToNull = false)]`

NicolasDorier:
@rockstardev I prefer my version, but feel free to revert my changes with 

```csharp
 internal static void Parse(ControllerBase ctrl, Func<CookiePrefs, ListQueryDataHolder> getQueryDataHolder,
            ref string searchTerm, ref int? timezoneOffset)
```

So at least we don't use reflection.

The `CookiePrefs` need to be public as it will be reused by other things.

The pattern I use in my code is necessary to repeat everytime you want to set some preferences.
Sure you could remove those 3 lines just for list preferences, but this is not really intuitive that List preferences will use a different pattern than other type of preferences.

I also prevent the duplication of having 4 parameters, always named the same, with always same default, for every list page.

rockstardev:
As we agreed, I'm force pushing final version and will also take time to explain the reasoning behind the code being the way it is for sake of posterity.

My goal was to allow us to say in code "set searchTerm+timezoneOffset to cookie if searchTerm is null and save them to cookie otherwise". This is essentially what following line does:

`model = this.ParseListQuery(model ?? new InvoicesModel());`

By doing this I've completely abstracted all implementation details and moved complexity behind these calls. All logic of storage, manipulation, etc is behind that method call.

The problem we are having here is that in order to implement algorithm which initializes specific `BasePagingViewModel` from specific cookie property - we need to hold reference to the property. If we are OK with that, then it's simple, we define bindings:

 - InvoicesModel -> UserPrefsCookie.InvoicesQuery
 - ListPaymentRequestsViewModel -> UserPrefsCookie.PaymentRequestsQuery

And are able to code algorithm that:

 - gets userPrefsCookie
 - if `model.SearchTerm` is null, fetch and set the value from cookie using specific property depending on model
 - else, save value in specific cookie property and update the cookie for user

Because we need the reference to property of the cookie - there is no way around, but to use reflection. If we really wanted to evade reflection, alternative would be to have cookie as Dictionary of keys where we String/Enum Key -> ViewModel relation. But that would result in not-as-strong-typing; right now you can directly see which properties UserPrefsCookie holds from it's definition:

    class UserPrefsCookie
    {
        public ListQueryDataHolder InvoicesQuery { get; set; }

        public ListQueryDataHolder PaymentRequestsQuery { get; set; }
    }

    class ListQueryDataHolder
    {
        public ListQueryDataHolder() { }

        public ListQueryDataHolder(string searchTerm, int? timezoneOffset)
        {
            SearchTerm = searchTerm;
            TimezoneOffset = timezoneOffset;
        }

        public int? TimezoneOffset { get; set; }
        public string SearchTerm { get; set; }
    }

If anybody has idea on how to create algorithm that abstract cookie manipulation operations and extracting of appropriate ViewModel without using reflection - feel free to chime in. I simply don't see way around this... if we restrict ourselves from using reflection as tool then we would be required to explicitly define mappings in every controller; i.e. copy-paste code similar to:

    var prefCookie = ListCookiePreference.GetPrefCookie(this);
    if (model.SearchTerm is null)
    {
        var p = prefCookie.Prop;
        model.SearchTerm = p.SearchTerm;
        model.TimezoneOffset = p.TimezoneOffset;
    }
    else
    {
        prefCookie.Prop = new ListQueryDataHolder(model.SearchTerm, model.TimezoneOffset);
        ListCookiePreference.SetPrefCookie(prefCookie);
    }

To me this is not ideal. Because when I am coding, I always aim to break down complexity and hide it behind strongly typed interfaces. That way I get behaviors - this specific one being: initialize ViewModel from cookie if needed, save the updated cookie otherwise.

Another thing I consider is how much I want to embed certain behavior into infrastructure. Usually when behavior is new, I prefer to keep it disconnected and independent - hence static methods. If I wanted to make this behavior part of infrastructure then the first step would be to code this behavior as class that accepts dependencies in constructor and returns populated model (making all methods non-static).

Another step from there would be defining specific interface like `IamInitilizedFromCookie` (or attribute) and requiring each implementing ViewModel to specify to which cookie property it's being saved.

If we do all this, then we could evade reflection... if we expand interface to return resulting cookie and indiscriminately save it after we initialize ViewModel as part of HTTP pipeline.

I however am choosing not to do this - because I don't want to change infrastructure. For now all I want is to abstract behavior behind strongly defined class. Which is why there is also this code:

    public static T ParseListQuery<T>(ControllerBase ctrl, T model) where T : BasePagingViewModel
    {
        PropertyInfo prop;
        if (model is InvoicesModel)
            prop = typeof(UserPrefsCookie).GetProperty(nameof(UserPrefsCookie.InvoicesQuery));
        else if (model is ListPaymentRequestsViewModel)
            prop = typeof(UserPrefsCookie).GetProperty(nameof(UserPrefsCookie.PaymentRequestsQuery));
        else
            throw new Exception("Unsupported BasePagingViewModel for cookie user preferences saving");

        return ProcessParse(ctrl, model, prop);
    }

Looking at this code someone may propose to move binding to specific `BasePagingViewModel`s. To me, at the current state of codebase, that would absolutely be the wrong choice because it leaks abstraction. With this definition being placed in the `ListCookiePreference` - it's exactly where it should be... 

Spilling these bindings into ViewModels would weaken them by introducing reference to UserPrefsCookie and that shouldn't be done without enriching ViewModels with interfaces (which we are not doing because of previously discussed infrastructure choices).

PR #1760, opened 2020-07-22
GreenField: Invoice API

### LINKED ISSUE #189: New Invoice API
Get a cleaner invoice API not burdened by bitpay legacy.

Kukks:
closes #189 

Current swagger doc for new endpoints:
https://redocly.github.io/redoc/?url=https://gist.githubusercontent.com/Kukks/be7b6597acec868be3943ce68611d4bb/raw/91747abdbae70d6f3c04f4f4ae3fa8974ea860ae/swagger.json#tag/Invoices


BTCPayServer.Client/Models/CreateInvoiceRequest.cs
```diff
@@ -0,0 +1,142 @@
...
+namespace BTCPayServer.Client.Models
+{
+    public class CreateInvoiceRequest
+    {
+        [JsonProperty(ItemConverterType = typeof(DecimalDoubleStringJsonConverter))]
+        public decimal Amount
+        {
+            get;
+            set;
+        }
```
dennisreimann, line 15: Formatting: The props should all be consistenly formatted, hence:

```suggestion
        public decimal Amount { get; set; }
```

BTCPayServer.Client/Models/InvoiceData.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                [JsonProperty(ItemConverterType = typeof(DecimalDoubleStringJsonConverter))]
+                public decimal Fee { get; set; }
+
+                [JsonConverter(typeof(StringEnumConverter))]
+                public PaymentStatus Status { get; set; }
+
+                public string Destination { get; set; }
+        
+
```
dennisreimann, line 57:
```suggestion
```

BTCPayServer/Controllers/GreenField/InvoiceController.cs
```diff
@@ -0,0 +1,225 @@
...
+                Metadata =
+                    new CreateInvoiceRequest.ProductInformation()
+                    {
+                        Physical = entity.ProductInformation.Physical,
+                        ItemCode = entity.ProductInformation.ItemCode,
+                        ItemDesc = entity.ProductInformation.ItemDesc,
+                        OrderId = entity.OrderId,
+                        PosData = entity.PosData,
+                        TaxIncluded = entity.ProductInformation.TaxIncluded
+                    },
```
dennisreimann, line 119:
```suggestion
                Metadata = new CreateInvoiceRequest.ProductInformation()
                {
                    Physical = entity.ProductInformation.Physical,
                    ItemCode = entity.ProductInformation.ItemCode,
                    ItemDesc = entity.ProductInformation.ItemDesc,
                    OrderId = entity.OrderId,
                    PosData = entity.PosData,
                    TaxIncluded = entity.ProductInformation.TaxIncluded
                },
```

Kixunil:
Thoughts:

* Would be nice to have a websoket for listening for events (invoices added, paid...)
* I guess `Get invoices` will become slow once there are many of them. Some kind of filtering/pagination would be helpful.

Kukks:
> 
> 
> Thoughts:
> 
>     * Would be nice to have a websoket for listening for events (invoices added, paid...)
> 
>     * I guess `Get invoices` will become slow once there are many of them. Some kind of filtering/pagination would be helpful.

Separate thing for events. Need to plan that better before we venture out into 2 way server communication such as websockets.

Will try and plan some consistent paging mechanism for all endpoints as a separate PR

NicolasDorier:
Nice.

Though I think we should remove `customer` and `metadata` completely.
Instead, we would let the user define his own JSON object for that whatever data they want to present.

NicolasDorier:
About the status: I think we should remove the "completed" state for the new API.

NicolasDorier:
Remove the "Update Invoice" instead we should have specific domain specific actions.

NicolasDorier:
why test fails?

Kukks:
> 
> 
> Nice.
> 
> Though I think we should remove `customer` and `metadata` completely.
> Instead, we would let the user define his own JSON object for that whatever data they want to present.
> 
> Basically have "metadata" be a "JsonObject".

I agree, though this could break the requirements of existing users who have been using these fields and would break compatibility of their existing integrations. We would also need to keep at least the following from the top of my head: `itemDesc`, `itemCode`, `email`, with everything else being comnolidated to use the `posData` field.

Kukks:
> 
> 
> About the status: I think we should remove the "completed" state for the new API.

It's a bit useless but I wouldn't do this as we still need to keep it internally for the other APIs..

Kukks:
> 
> 
> Remove the "Update Invoice" instead we should have specific domain specific actions. (Archive or Mark Invalid/Mark Completed)

Yeah, will do this though I will make the mark status one call

NicolasDorier:
> I agree, though this could break the requirements of existing users who have been using these fields and would break compatibility of their existing integrations

I don't understand. If they use the new API they don't have existing integrations.

Kukks:
> 
> 
> > I agree, though this could break the requirements of existing users who have been using these fields and would break compatibility of their existing integrations
> 
> I don't understand. If they use the new API they don't have existing integrations. And if we really cared, we can just consider that if the metadata json object property has a specific name, like "buyerName" then we copy the properties value to the invoice entity.

No worries, I did the change. can you check it now? 
Also, no clue why LN tests are failing.. any idea?

NicolasDorier:
why `NumericStringJsonConverter`?

NicolasDorier:
Argh, can you not make refactoring change at the same time? It makes it hard to review. Like the status and converters.

NicolasDorier:
meh that's ok but prefer for future PR you don't cleanup things at the same time.

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -0,0 +1,844 @@
...
+                        "format": "email",
+                        "nullable": true,
+                        "description": "The email of the customer. If the store is configured to ask for a refund email, the checkout UI will prompt for one when not provided."
+                    },
+                    "currency": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "The currency the invoice will use"
+                    },
+                    "metadata": {
```
NicolasDorier, line 674: should not be a string but an untyped JSON object

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -0,0 +1,844 @@
...
+                        "oneOf": [
+                            {
+                                "$ref": "#/components/schemas/CheckoutOptions"
+                            }
+                        ],
+                        "description": "Additional settings to customize the checkout flow"
+                    }
+                }
+            },
+            "ProductInformation": {
```
NicolasDorier, line 690: should be removed, all of that can be in metadata

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -0,0 +1,844 @@
...
+                    },
+                    "taxIncluded": {
+                        "type": "string",
+                        "format": "decimal",
+                        "nullable": true,
+                        "description": "How much tax is included in the invoice amount"
+                    }
+                }
+            },
+            "BuyerInformation": {
```
NicolasDorier, line 726: should be removed, all of that can be in metadata

NicolasDorier:
@Kukks about the completed state, why do we need it exactly? As far as the merchant is concerned they only care if he can ship, which he can if confirmed or completed. There is nothing he would want to do only in completed and not confirmed.

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
Kukks: It should be a string, as the user may just want to throw in a string in there and not json.. Just like how we handle the posData field in the BTCPay UI

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
Kukks: They are not used directly, but i'M T

> 
> 
> why `NumericStringJsonConverter`?

Because I want to use the same converter for more types such  as `double`

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
Kukks: > 
> 
> why `NumericStringJsonConverter`?

Because I want to use the same converter for more types such  as `double`

Kukks:
> 
> 
> why `NumericStringJsonConverter`?



> 
> 
> why `NumericStringJsonConverter`?

Because I want to use the same converter for more types such  as `double`

Kukks:
> 
> 
> @Kukks about the completed state, why do we need it exactly? As far as the merchant is concerned they only care if he can ship, which he can if confirmed or completed. There is nothing he would want to do only in completed and not confirmed.
> 
> I am talking about removing it at the API level, not internally.

I wouldn't though. People are now used to this behavior and they would be confused seeing complete in the UI  but not the API

NicolasDorier:
IMHO we should remove the distinction between confirmed and complete in the UI this is completely useless. @pavlenex what do you think?

pavlenex:
Tough one, as Kukks said people may already be used to certain type of invoice behavior. This has always be a bit over the top for me. In most cases, merchant wants to know when something is paid and when they can ship which from what I understood is complete state.

Capture

> distinction between confirmed and complete in the UI
@NicolasDorier by distinction you mean completely remove one of the statuses?

NicolasDorier:
> when something is paid and when they can ship which from what I understood is complete state.

No, this is the confirmed state. Complete state is after 6 confirmation, this is just for bitpay's compatibility.

NicolasDorier:
Rebased

BTCPayServer.Client/Models/CreateInvoiceRequest.cs
```diff
@@ -0,0 +1,36 @@
...
+        public string Currency { get; set; }
+        public JObject Metadata { get; set; }
+        public CheckoutOptions Checkout { get; set; } = new CheckoutOptions();
+
+        public class CheckoutOptions
+        {
+            [JsonConverter(typeof(StringEnumConverter))]
+            public SpeedPolicy? SpeedPolicy { get; set; }
+
+            public string[] PaymentMethods { get; set; }
```
Kukks, line 23: Maybe we should switch to using PaymentMethodId directly and then use the JsonConverter?

Issue #1719, opened 2020-07-07
Re-design Payment Request page and Pull payment page


pavlenex:
**Is your feature request related to a problem? Please describe.**
This issue is about re-designing payment request and pull payments page to look a bit more user-friendly and consistent with our design guidelines. These pages in current form aren't bad, but it would be cool if we can enhance them even further. Layout redesign is also acceptable course of action, not just aesthetics.

### Payment Request

In BTCPay Server, there are a few ways ways to request a payment. One is traditionally by to manually create an invoice. Second is to use one of the apps we have like Point of Sale or Crowdfunding. Payment requests is yet another way to request a payment, but with a difference that it always provides a current rate to avoid volatility problems. 

Think of it as a payment link, where a freelancer can send to a person who hired them and get paid in BTC. Payer can then click on that link, generate an invoice at current rate and pay. 

You can learn more about payment requests if you read [docs ](https://docs.btcpayserver.org/PaymentRequests/)or better yet [watch a video](https://youtu.be/j6CvwDPvfzQ).

![paymentrequest](https://user-images.githubusercontent.com/36959754/86791069-6feb1d80-c069-11ea-96ee-3119ab683d23.PNG)

### Pull Payment

Pull payment is the opposite concept of payment requests in a way. It's made for the receiver to provide information for the payment (amount + address) so that merchant can claim a payout to them (refund).

![PULLPAYMENT](https://user-images.githubusercontent.com/36959754/86791087-77122b80-c069-11ea-9442-e0735e4d4a76.PNG)

You can learn more about pull payments requests if you read [docs ](https://docs.btcpayserver.org/PullPayments/).

**Describe the solution you'd like**

1. Discuss the layout and if we can improve it
2. Create a mockup in figma
3. Based on mockup developers will be able to improve the pages

**Additional context**
Unfortunately the naming of these two concept is confusing probably. Furthermore the UX for the Pull Payments is still new and we're trying to perfect it. If anybody has ideas on how to improve the workflow please open up another issue, for this issue in particular it would be good to focus on a smaller chunk - design of the pages themselves where people could pay or provide info to get paid.

dstrukt:
Few suggestions moving forward for making UX/UI tickets for more complex views (even though this is relatively simple)

1. If it's possible to do a screen capture of the interaction (via QuickTime on Mac, etc..) to capture more than a static mock. I know it's a slight bit more work, but I think it will help push discussion further :)
[...]

PR #1764, opened 2020-07-23
Improve Pull Payment view

dennisreimann:
Making sure the different constellations work properly, especially in terms of different breakpoints.

## Mobile

mobile-1

mobile-2

mobile-3

## Medium

medium-1

medium-2

## Large

large-1

large-2

### DIFF
--- a/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
+++ b/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
@@ -1,4 +1,5 @@
-﻿@model BTCPayServer.Models.ViewPullPaymentModel
+﻿@using NUglify.Helpers
+@model BTCPayServer.Models.ViewPullPaymentModel
 
 @addTagHelper *, BundlerMinifier.TagHelpers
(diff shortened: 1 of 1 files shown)


pavlenex:
I appreciate the influx of PR's but it would be good to focus on stuff we agreed on. For example this will most likely be a wasted effort because of the #1719. Don't have anything against the changes but would be good if we can collaborate a bit better to avoid overlaps and wasted efforts.

Kukks:
Yeah, I think what we should do is merge this one and then consolidate both views to have a common UI framework to reduce code duplication. Easier to be consistent when we use the same "layout"

dennisreimann:
@pavlenex sorry, I must have overlooked the issue. Did these spontaneously as I came across that page testing stuff. Will take a better look at the list next time.

pavlenex:
No worries, this is still a good PR, just may get obsoleted quick, don't want your efforts to be in vain.😺 
I still think we'll get something good out of it, once merged, we can consolidate the layouts and then proceed with #1719 😉

PR #1767, opened 2020-07-24
GreenField: Store set payment methods

Kukks:


Kukks:
After thinking about it,I need to make this API a bit more independent for future work around decoupling wallets from stores.

PR #1769, opened 2020-07-24
Check in VS Code debug configurations

msafi:
When setting up BTCPay on VS Code IDE, VS Code will ask

image

When you click "Yes", you'll be presented with two options

image

If you select `BTCPayServer.Tests`, the C# extension will generate broken debug configurations. So pressing F5 will not work. At this point, to fix the issue, you have to delete `.vscode/launch.json` and `.vscode/tasks.json` then restart VS Code. It will prompt you again to add the missing assets and this time you can select `BTCPayServer` to get things to work.

To avoid this situation altogether, this PR checks-in the correct `.vscode/launch.json` and `.vscode/tasks.json`. With that, VS Code users would not be prompted to add missing assets. Pressing F5 should just work.

But I think there's some cruft in those files (see comments on the code below)


.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
...
+   // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+   "version": "0.2.0",
+   "configurations": [
+        {
+            "name": ".NET Core Launch (web)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
+            // If you have changed target frameworks, make sure to update the program path.
+            "program": "${workspaceFolder}/BTCPayServer/bin/Debug/netcoreapp3.1/BTCPayServer.dll",
```
msafi, line 13: .NET Core Launch (web) runs `${workspaceFolder}/BTCPayServer/bin/Debug/netcoreapp3.1/BTCPayServer.dll` binary with no arguments.

Is the name ".NET Core Launch (web)" correct? Should it be renamed to `Docker-Regtest-https` to match Visual Studio?

.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
...
+            "sourceFileMap": {
+                "/Views": "${workspaceFolder}/Views"
+            }
+        },
+        {
+            "name": ".NET Core Attach",
+            "type": "coreclr",
+            "request": "attach",
+            "processId": "${command:pickProcess}"
+        }
```
msafi, line 34: I'm not sure if this configuration object is necessary? I tried running it and it doesn't seem to be doing anything useful. Any objections to deleting it?

.vscode/tasks.json
```diff
@@ -0,0 +1,42 @@
...
+            "args": [
+                "watch",
+                "run",
+                "${workspaceFolder}/BTCPayServer/BTCPayServer.csproj",
+                "/property:GenerateFullPaths=true",
+                "/consoleloggerparameters:NoSummary"
+            ],
+            "problemMatcher": "$msCompile"
+        }
+    ]
```
msafi, line 41: I don't see the `publish` and `watch` tasks being used by any of the launch configurations. Do developers run these tasks directly? If not, perhaps they should also be removed.

Eskyee: `{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (web)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceFolder}/BTCPayServer/bin/Debug/netcoreapp3.1/BTCPayServer.dll",
            "args": [],
            "cwd": "${workspaceFolder}/BTCPayServer",
            "stopAtEntry": false,
            "serverReadyAction": {
                "action": "openExternally",
                "pattern": "\\bNow listening on:\\s+(https?://\\S+)"
            },
            "env": {
                "ASPNETCORE_ENVIRONMENT": "Development"
            },
            "sourceFileMap": {
                "/Views": "${workspaceFolder}/Views"
            }
        },
    ]
}`

in the hidden .vscode folder.. this is my launch.json settings used on vs code Mac, the tasks.json you have committed are the same as mine. @msafi

Eskyee: my .vscode hidden files on a Mac
 Screenshot 2020-07-24 at 19 07 31

Screenshot 2020-07-24 at 19 07 52

.vscode/launch.json
Eskyee: .NET Core Attach not needed.

.vscode/launch.json
Eskyee: .NET Core Launch (web) is ok for VS Code  
VS Studio & VS Code behave very differently. with launch settings.
as a example .. in VS Studio  hidden folder the path is .vs btcpayserver xs   btcpayserver/BTCPayServer/Properties/launchSettings.json
and the hidden folder is different than In vs code .hidden folder for launch.
Screenshot 2020-07-24 at 19 19 34

hope that helps with your commit.

Eskyee:
Links to documentation supporting these rule changes: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run

.vscode/launch.json
```diff
@@ -0,0 +1,39 @@
...
+                "pattern": "\\bNow listening on:\\s+(https?://\\S+)"
+            },
+            "env": {
+                "ASPNETCORE_ENVIRONMENT": "Development"
+            },
+            "sourceFileMap": {
+                "/Views": "${workspaceFolder}/Views"
+            },
+            "logging": {
+                "moduleLoad": false
```
msafi, line 29: According to [this video](https://www.youtube.com/watch?v=j486T_Rk-yw&feature=youtu.be), setting `logging.moduleLoad` to `false` will suppress unnecessary logs in the terminal.

Eskyee: Yes @rockstardev knows best . Depending on the level of logs needed in a development environment.

NicolasDorier:
@msafi so with this, doing F5 under visual studio code just work?

.vscode/launch.json
dennisreimann: imho it'd make sense to rename/unify it. This way one can get a better grasp of what's going on and what to run. I found it hard to get started in VS Code and have gladly switched to Rider since.

msafi:
@NicolasDorier 

> @msafi so with this, doing F5 under visual studio code just work?

If they have the C# extension for VS Code installed, then yes, it should just work.

VS Code will prompt the user to install the C# extension when they open a `cs` file. Or we can also check-in [`.vscode/extensions.json`](https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions) with the following content so the user will be prompted to install the extension as soon as they open VS Code

```json
{
  "recommendations": [
    "ms-dotnettools.csharp"
  ]
}
```

.vscode/tasks.json
rockstardev: @msafi as @Eskyee says, those are standard.

- `watch` gives you ability to automatically recompile and reload the solution as soon as you change the file. So you can have server running and it'll automatically recompile and run on file save. Example: https://www.youtube.com/watch?v=nYudqdL55us
- `publish` is for exporting your website to another directory / service. We don't use since our deployments are described in `btcpayserver-docker` repository. But you can leave it in.

.vscode/launch.json
rockstardev: You can delete .NET Core Attach... it's there for where you have BTCPayServer process already running and you want to attach to it for debugging through VSCode. If it doesn't work for you - definitely remove and then if anybody comes along that finds use for it, they can add it the way they use it.

rockstardev:
@msafi feel free to add `extensions.json` as well, that's a good suggestion.

@dennisreimann warnings are gone because of CanUseTorClient

.vscode/launch.json
msafi: @dennisreimann Thing is, I'm not sure `.NET Core Launch (web)` does exactly the same thing as `Docker-Regtest-https`, does it? If we're not sure, maybe we should just leave it as it is?

Eskyee:
just tested this latest pull request in visual studio code, and the builds are fine no problems detected, 
ps now a much better UX on a Mac @NicolasDorier @msafi  ;) thanks

.vscode/launch.json
Eskyee: > .NET Core Launch (web)

I would leave as is for now, until @Kukks gives the nod!, 
extra info left here  https://code.visualstudio.com/docs/containers/debug-common

Eskyee: > imho it'd make sense to rename/unify it. This way one can get a better grasp of what's going on and what to run. I found it hard to get started in VS Code and have gladly switched to Rider since. 

its in the name  !! Rider 
much more faster, & productive with development!!  makes it fun to dev on a Mac..

Eskyee:
@msafi once you finished here, could you test Visual Studio, and BTCPay Server, as a default, etc etc, because we had a few problems, developing on a Mac!! cheers 

their are no problems developing with Rider IDE on a Mac.  it checks out ok.

msafi:
@Eskyee I just tested the configurations in VS Code and I was able to start the app and set a breakpoint. It seems to work.

I think this PR is ready to merge.

Eskyee:
Cheers 🍻

PR #1782, opened 2020-07-28
Add Bitcoin Only Build

xpayserver:


NicolasDorier:
awesome work thanks reviewing

NicolasDorier:
I need to change some stuff, you don't use MSBuild properly. But that's the idea.

NicolasDorier:
I will merge it.

@Kukks I created new configuration build for visual studio (Altcoins-Release and Altcoins-Debug) which currently only connect to LTC and LBTC. I adapted the launchsettings.json to have only Bitcoin, Bitcoin-Https and Altcoins.

I created a new docker-compose: docker-compose.altcoins.yml  (this is what we run on CI) with monero, ltc and lbtc.

I think it would be useful that you add to the launch settings monero, so it connects to the monero wallet in altcoins.

Issue #1420, opened 2020-03-28
Introduce "Update available" alert 


Kukks:
So many users stay on very outdated versions. We should add a small alert that shows up when the version shown on BTCPay does not match the latest release tag on github/ docker hub. 

 This should be optional with a flag of some sort `BTCPAY_UPDATE_ALERT`, which is off by default.

In the docker repo, we should turn it on by default, UNLESS `BTCPAY_IMAGE` is specified, which will turn it off

pavlenex:
Not quite sold on this one. People who don't regularly maintain their server are the ones who connect it, and never visit it unless something breaks, in that case they'll update anyway. I think it's been discussed many times and general idea is that we should not urge users to update in any way, since it's a recipe for a mass breaking of servers in case of a bug. On the other hand, it can be good for critical bug-fixes and security patches. Really not sure.

If we decide this is a good idea however, I suggest we look into hooking it up with #775 as a way of displaying.

Kinda related to #1385

woutersamaey:
You can make this very small and put it next to the current version number in the admin.

Something like:
Current version is 1.2.3 - An update is available

Doesn't need to be a big popup.

Eskyee:
I agree with kukks their are a lot of users I see using old versions of BTCPayserver, given more version information details , might help those update to the latest version more often! & might help to cut down on support for old versions!!

pavlenex:
Related #775 #1542

PR #1788, opened 2020-07-31
Adding HostedService that checks for new BTCPayServer version on GitHub once a day

rockstardev:
Resolves: #1420 

Policy to check for new version needs to enabled (false by default)

image

The check is disabled when developing (`NetworkType == NetworkType.Regtest && Environment.IsDevelopment()`).

I've implemented it as part of Policy settings since to me setting this over there seemed more intuitive... if you prefer to rather have it as part of `BTCPayServerOptions`, let me know.

### DIFF
--- a/BTCPayServer/Configuration/DefaultConfiguration.cs
+++ b/BTCPayServer/Configuration/DefaultConfiguration.cs
@@ -38,6 +38,8 @@ namespace BTCPayServer.Configuration
             app.Option("--sshtrustedfingerprints", "SSH Host public key fingerprint or sha256 (default: empty, it will allow untrusted connections)", CommandOptionType.SingleValue);
             app.Option("--torrcfile", "Path to torrc file containing hidden services directories (default: empty)", CommandOptionType.SingleValue);
             app.Option("--socksendpoint", "Socks endpoint to connect to onion urls (default: empty)", CommandOptionType.SingleValue);
+            app.Option("--updatecheck", $"Enabling once a day check for new releases (default: false)", CommandOptionType.SingleValue);
+            app.Option("--updateurl", $"Url location used for updatecheck (default: https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest)", CommandOptionType.SingleValue);
             app.Option("--debuglog", "A rolling log file for debug messages.", CommandOptionType.SingleValue);
             app.Option("--debugloglevel", "The severity you log (default:information)", CommandOptionType.SingleValue);
             app.Option("--disable-registration", "Disables new user registrations (default:true)", CommandOptionType.SingleValue);
--- a/BTCPayServer/Configuration/DefaultConfiguration.cs
+++ b/BTCPayServer/Configuration/DefaultConfiguration.cs
@@ -38,8 +38,7 @@ namespace BTCPayServer.Configuration
             app.Option("--sshtrustedfingerprints", "SSH Host public key fingerprint or sha256 (default: empty, it will allow untrusted connections)", CommandOptionType.SingleValue);
             app.Option("--torrcfile", "Path to torrc file containing hidden services directories (default: empty)", CommandOptionType.SingleValue);
             app.Option("--socksendpoint", "Socks endpoint to connect to onion urls (default: empty)", CommandOptionType.SingleValue);
-            app.Option("--updatecheck", $"Enabling once a day check for new releases (default: false)", CommandOptionType.SingleValue);
-            app.Option("--updateurl", $"Url location used for updatecheck (default: https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest)", CommandOptionType.SingleValue);
+            app.Option("--updateurl", $"Url used for once a day new release version check. Check performed only if value is not empty (default: empty)", CommandOptionType.SingleValue);
             app.Option("--debuglog", "A rolling log file for debug messages.", CommandOptionType.SingleValue);
             app.Option("--debugloglevel", "The severity you log (default:information)", CommandOptionType.SingleValue);
             app.Option("--disable-registration", "Disables new user registrations (default:true)", CommandOptionType.SingleValue);
--- a/BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
+++ b/BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
@@ -94,6 +94,9 @@ namespace BTCPayServer.HostedServices
         private static readonly Regex _releaseVersionTag = new Regex("^(v[1-9]+(\\.[0-9]+)*(-[0-9]+)?)$");
         public async Task<string> Fetch(CancellationToken cancellation)
         {
+            if (_updateurl == null)
+                return null;
+
             using (var resp = await _httpClient.GetAsync(_updateurl, cancellation))
             {
                 var strResp = await resp.Content.ReadAsStringAsync();
--- a/BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
+++ b/BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
@@ -39,7 +39,7 @@ namespace BTCPayServer.HostedServices
         public async Task ProcessVersionCheck()
         {
             var policies = await _settingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
-            if (policies.CheckForNewVersions && !_env.IsDeveloping)
+            if (policies.CheckForNewVersions)
             {
                 var tag = await _versionFetcher.Fetch(Cancellation);
                 if (tag != null && tag != _env.Version)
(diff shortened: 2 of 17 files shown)


BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,95 @@
...
+        protected async Task LoopVersionCheck()
+        {
+            await ProcessVersionCheck();
+            await Task.Delay(TimeSpan.FromDays(1), Cancellation);
+        }
+
+        public async Task ProcessVersionCheck()
+        {
+            var policies = await _settingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
+            if (policies.CheckForNewVersions && !_env.IsDeveloping)
```
NicolasDorier, line 42: remove `!_env.IsDeveloping`.

`CheckForNewVersions` should be false by default, I prefer not pushing people to update.

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        Task<string> Fetch(CancellationToken cancellation);
+    }
+
+    public class GithubVersionFetcher : IVersionFetcher
+    {
+        private readonly HttpClient _httpClient;
+        public GithubVersionFetcher()
+        {
+            _httpClient = new HttpClient();
```
NicolasDorier, line 75: Use the IHttpClietnFactory.

rockstardev: d280617

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
rockstardev: d280617

pavlenex:
Shouldn't we have User preferences for notifications? I get this particular one makes sense in Server settings, but then they'd be scattered around?

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        private readonly HttpClient _httpClient;
+        public GithubVersionFetcher(IHttpClientFactory httpClientFactory)
+        {
+            _httpClient = httpClientFactory.CreateClient(nameof(GithubVersionFetcher));
+            _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
+            _httpClient.DefaultRequestHeaders.Add("User-Agent", "BTCPayServer/NewVersionChecker");
+        }
+
+        public void Dispose()
```
NicolasDorier, line 80: I don't think you call that anywhere. That said, you don't have to dispose HttpClient... this is one of those weird type that defy conventional wisdom...

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            _httpClient.Dispose();
+        }
+
+        public async Task<string> Fetch(CancellationToken cancellation)
+        {
+            const string url = "https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest";
+            var resp = await _httpClient.GetAsync(url, cancellation);
```
NicolasDorier, line 88: though I think you need to dispose the resp here!

NicolasDorier:
Later we will just do a user level way to filter notifications. That is orthogonal to this PR.

NicolasDorier:
@rockstardev outside my two nit comments, you can merge this.

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
Kukks: Yeah this is just a magical call that .net invokes itself somehow.

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        Task<string> Fetch(CancellationToken cancellation);
+    }
+
+    public class GithubVersionFetcher : IVersionFetcher, IDisposable
+    {
+        private readonly HttpClient _httpClient;
+        public GithubVersionFetcher(IHttpClientFactory httpClientFactory)
+        {
+            _httpClient = httpClientFactory.CreateClient(nameof(GithubVersionFetcher));
```
Kukks, line 75: @NicolasDorier  I was wondering if maybe we should use the tor http client if it is available to do this call to not ping a third party constantly? (Also think it should be an option on rates so maybe it should be a separate PR)

NicolasDorier:
I agree with @Kukks on `BTCPAY_UPSTREAM_GITHUB_REPO`

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
NicolasDorier: I don't think we should aim at using TOR for protecting the privacy of the server itself, this is too hard, untestable, and will be quite unreliable.

This is better handled by using a VPN or putting the whole network behind TOR.

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
```diff
@@ -0,0 +1,100 @@
...
+            await ProcessVersionCheck();
+            await Task.Delay(TimeSpan.FromDays(1), Cancellation);
+        }
+
+        public async Task ProcessVersionCheck()
+        {
+            var policies = await _settingsRepository.GetSettingAsync<PoliciesSettings>() ?? new PoliciesSettings();
+            if (policies.CheckForNewVersions)
+            {
+                var tag = await _versionFetcher.Fetch(Cancellation);
```
NicolasDorier, line 44: what if that throw? (ie, github unavailable)

This will probably execute `ProcessVersionCheck` in a tight loop and spamming the logs.

NicolasDorier:
I think https://github.com/btcpayserver/btcpayserver/pull/1788#discussion_r463445424 is important as well

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
Kukks: Yeah fair

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
Kukks: Yeah I guess a try catch with either waiting another day or giving it a 1hour delay?

rockstardev: 11c09ea

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
rockstardev: 11c09ea

BTCPayServer/HostedServices/NewVersionCheckerHostedService.cs
rockstardev: You two are ganging up on these code reviews of my PRs so much, that now there is a duplication of threads. 

image

Resolved: 11c09ea , same as below.

rockstardev:
> Can you remove the public async Task<IActionResult> Generate(string version) in NotificationsController now?

No, it's used in test of notification in UnitTest1.cs

> One concern I have is not being able to specify the url of the repo to ping for versions of forks. Personally, would add a new config option "BTCPAY_UPSTREAM_GITHUB_REPO" and set it default to https://github.com/btcpayserver/btcpayserver

I've added option to directly set path to JSON file so that there isn't any magic string appending. Thus if someone wants to serve their update JSON file somewhere other than GitHub - they can, it just needs to be in format `{ "tag_name": "v1.2.3.4" }`

> Another concern is what happens if we do a feature test release like we had for the vlndseedbackup thing? Would this grab it and announce it as a new version?

I've decided to use same regex we use for circleci

For user notifications preferences - I spoke with @pavlenex and will be doing page to allow each user to select which notifications it wants to receive. When that's done I'll initialize each admins update check default from BTCPayServerOptions (instead of server policy).

BTCPayServer/Configuration/DefaultConfiguration.cs
```diff
@@ -38,6 +38,8 @@ protected override CommandLineApplication CreateCommandLineApplicationCore()
             app.Option("--sshtrustedfingerprints", "SSH Host public key fingerprint or sha256 (default: empty, it will allow untrusted connections)", CommandOptionType.SingleValue);
             app.Option("--torrcfile", "Path to torrc file containing hidden services directories (default: empty)", CommandOptionType.SingleValue);
             app.Option("--socksendpoint", "Socks endpoint to connect to onion urls (default: empty)", CommandOptionType.SingleValue);
+            app.Option("--updatecheck", $"Enabling once a day check for new releases (default: false)", CommandOptionType.SingleValue);
+            app.Option("--updateurl", $"Url location used for updatecheck (default: https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest)", CommandOptionType.SingleValue);
```
NicolasDorier, line 42: can you make it only in one variable? If not specified or empty, then the check is disabled.

rockstardev: Ah, this is what I was considering @NicolasDorier... but what happens when person has update url other than default and doesn't want to configure it each time.

This would also require our users to copy-paste ` https://api.github.com/repos/btcpayserver/btcpayserver/releases/latest` every time they want to enable it. What would be the way we're going to deploy this with `btcpayserver-docker`? BTCPAY_UPDATEURL empty and then users required to populate it to enable?

NicolasDorier: The default should be empty. It is btcpayserver-docker job to pass the right URL.

rockstardev: 16eedf4

PR #1800, opened 2020-08-04
Add API Keys Application identifier + Redirect

Kukks:
This lets the authorize api key screen redirect to the defined url and provide it with the user id, permissions granted and the key.

This also allows apps to match existing api keys generated for it specifically using the application identifier, and if matched, presented with a confirmation page before redirection.

### DIFF
--- a/BTCPayServer/Controllers/ManageController.APIKeys.cs
+++ b/BTCPayServer/Controllers/ManageController.APIKeys.cs
@@ -7,7 +7,9 @@ using BTCPayServer.Client;
 using BTCPayServer.Data;
 using BTCPayServer.Models;
 using BTCPayServer.Security.GreenField;
+using ExchangeSharp;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using NBitcoin;
 using NBitcoin.DataEncoders;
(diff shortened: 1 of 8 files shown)


BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -94,14 +96,86 @@ public async Task<IActionResult> AddApiKey()
...
-            var parsedPermissions = Permission.ToPermissions(permissions).GroupBy(permission => permission.Policy);
+            var requestPermissions = Permission.ToPermissions(permissions);
+            if (!string.IsNullOrEmpty(applicationIdentifier) && redirect != null)
+            {
+                //check if there is an app identifier that matches and belongs to the current user
+                var keys = await _apiKeyRepository.GetKeys(new APIKeyRepository.APIKeyQuery()
+                {
+                    UserId = new[] {_userManager.GetUserId(User)}
+                });
+                if (keys.Any())
```
NicolasDorier, line 107: this if is useless

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -183,7 +256,13 @@ public async Task<IActionResult> AuthorizeAPIKey([FromForm] AuthorizeApiKeysView
                 case "no":
                     return RedirectToAction("APIKeys");
                 case "yes":
-                    var key = await CreateKey(viewModel);
+                    var key = await CreateKey(viewModel, (viewModel.ApplicationIdentifier, viewModel.RedirectUrl.Authority));
+
+                    if (viewModel.RedirectUrl != null)
```
NicolasDorier, line 261: `viewModel.RedirectUrl == null` can never be true, because else, the like just above this would throw NullReferenceException

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -195,6 +274,20 @@ public async Task<IActionResult> AuthorizeAPIKey([FromForm] AuthorizeApiKeysView
...
+        private string GetRedirectToApplicationUrl(Uri redirect, APIKeyData key)
+        {
+            var uri = new UriBuilder(redirect);
+            var permissions = key.GetBlob().Permissions;
+            uri.AppendPayloadToQuery(new Dictionary<string, object>()
+            {
+                {"key", key.Id}, {"permissions",permissions}, {"user", key.UserId}
+            });
+            //uri builder has bug around string[] params
+            return uri.Uri.ToStringInvariant().Replace("permissions=System.String%5B%5D",
```
NicolasDorier, line 286: have you tried to use `uri.Uri.AbsoluteUri` instead?

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -195,6 +274,20 @@ public async Task<IActionResult> AuthorizeAPIKey([FromForm] AuthorizeApiKeysView
...
+                string.Join("&", permissions.Select(s1 => $"permissions={s1}")), StringComparison.InvariantCulture);
```
NicolasDorier, line 287: Injection possible via `s1`, need to be escaped.

BTCPayServer/Controllers/ManageController.APIKeys.cs
NicolasDorier: You can probably also avoid this bug by having one `{"permissions",perm}` by `perm` in `permissions`

BTCPayServer/Controllers/ManageController.APIKeys.cs
Kukks: fixed

BTCPayServer/Controllers/ManageController.APIKeys.cs
Kukks: fixed

BTCPayServer/Controllers/ManageController.APIKeys.cs
Kukks: fixed

BTCPayServer/Controllers/ManageController.APIKeys.cs
Kukks: fixed

PR #1814, opened 2020-08-05
Remove confirm option for seed from wallet settings

### LINKED ISSUE #1813: Remove confirm option when viewing seed from wallet settings
If we view the seed from the wallet settings, it will ask you to confirm saving the seed, which is redundant as they had already confirmed it before.

dennisreimann:
Closes #1813.


dennisreimann:
@Kukks can you retrigger the tests? The failure looks unrelated

```
  X CanUsePoSApp [2s 752ms]
  Error Message:
   The collection was expected to contain a single element, but it contained more than one element.
```

PR #1815, opened 2020-08-05
Improve invoices list view

dennisreimann:
Sparked by [@Eskyee 's remarks](https://chat.btcpayserver.org/btcpayserver/pl/g3eqikm4r7ghmb7hm1bdxkpgoa) I took a look at the Invoices list and I'd like to propose the following changes.

Same would apply to the payment requests as well, which I'd also update if we decide all of this makes sense.

## Before

before-lg

before-sm

## After

As you can see there are some general layout cleanups and rearrangements.

Beyond that, these changes have been made:

- Separate the mass actions from the create button and make them secondary
- Prepend search help/filtering info to search field
- Make search button secondary

after-lg

With the help text expanded:

after-lg-help

Mobile view:

after-sm

### Empty state

Let's also clean up the empty state, as the search, actions and table display make no sense if there are no items.

no-invoices

### DIFF
--- a/BTCPayServer/Components/Pager/Default.cshtml
+++ b/BTCPayServer/Components/Pager/Default.cshtml
@@ -26,23 +26,36 @@
             </li>
         </ul>
     }
-    <ul class="pagination float-right">
-        <li class="page-item disabled">
-            <span class="page-link">Page Size:</span>
-        </li>
-        <li class="page-item @(Model.Count == 50 ? "active" : null)">
-            <a class="page-link" href="@NavigatePages(0, 50)">50</a>
-        </li>
-        <li class="page-item @(Model.Count == 100 ? "active" : null)">
-            <a class="page-link" href="@NavigatePages(0, 100)">100</a>
-        </li>
-        <li class="page-item @(Model.Count == 250 ? "active" : null)">
-            <a class="page-link" href="@NavigatePages(0, 250)">250</a>
-        </li>
-        <li class="page-item @(Model.Count == 500 ? "active" : null)">
-            <a class="page-link" href="@NavigatePages(0, 500)">500</a>
-        </li>
-    </ul>
+
+    @if (Model.Total >= 50)
+    {
+        <ul class="pagination float-right">
+            <li class="page-item disabled">
+                <span class="page-link">Page Size:</span>
+            </li>
+            <li class="page-item @(Model.Count == 50 ? "active" : null)">
+                <a class="page-link" href="@NavigatePages(0, 50)">50</a>
+            </li>
+            @if (Model.Total >= 100)
+            {
+                <li class="page-item @(Model.Count == 100 ? "active" : null)">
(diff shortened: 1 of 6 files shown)


pavlenex:
I prefer the inline, what's the goal of having the layout in two rows? Personally I prefer inline as it pushes invoices even more to the top.
 I like the grayed out buttons, but I'd suggest for the search split toggle button to add darker gray color to indicate it that there's another action (invoice filtering)
For example (chose color randomly)
```
.dropdown-toggle-split {
    background: #495057;
}
```
split

dennisreimann:
> I prefer the inline, what's the goal of having the layout in two rows? Personally I prefer inline as it pushes invoices even more to the top.

I think separating them makes sense, because creating is a different kind of action then the mass editing ones. As the latter belong to the table, they also need to be part of the form, which wraps the buttons and table. I guess that technical restriction was also why the previous version had some funky markup, which tried to work around this limitation.

As for the button divider: Good point, I'll add this.

pavlenex:
@britttttk can you test and check if this addresses the #1823 entirely?

britttttk:
@pavlenex @dennisreimann invoice page is aligned looks much better! For the payment request page other comment I had it is the same. I only noticed it by chance testing zooming in and out, I don't think it's a big issue.

dennisreimann:
I can take care of that and will adapt the payment requests overview as well.

dstrukt:
Can we change the wording on the table header "Expiry" to "Expiration", and on the actual element "No Expiry" -> "No expiration"?

Other thought:

- The grey buttons for the mass actions and search are using the "disabled" state color, and personally read this way, as opposed to secondary actions or buttons.

dennisreimann:
> The grey buttons for the mass actions and search are using the "disabled" state color, and personally read this way, as opposed to secondary actions or buttons.

What color do we want to use for the secondary buttons? [The secondary buttons on Figma](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY?node-id=17%3A100) look like the primary/default ones and I'd like to distinguish the cases somehow. But you are right, the gray might give the impression of the buttons being disabled.

Here's what they would look like as outlined:

outlined

britttttk:
>"Expiry" to "Expiration", and on the actual table entry "No Expiry"

@dstrukt this Expiry is used in a few places in BTCPay. I mentioned this previously but it seems Expiration is american terminology. So I don't think using it is a requirement personally if it's common elsewhere... (although as an american I have never seen Expiry, but I don't think it's unintuitive).

dstrukt:
@britttttk ahhh thanks for clarifying. agreed, I could definitely figure it out as an american as well - just looked odd to me but TIL, and yes, I'm only referring to the payment requests - I feel like I've also seen it in one other place, but where is escaping me right now.

@dennisreimann DM'd you directly to talk through a few things before I share more designs.

Eskyee:
Expire or Expired is very English. And preferred. Here

PR #1818, opened 2020-08-06
Add Better Feature Build Support

xpayserver:
This allow you to build specific builds based on feature for specific version such as `v1.0.0.88-lndseedbackup-1` (good for forks with custom features)


NicolasDorier:
I don't want to increase the number of build unless strictly necessary.

PR #1819, opened 2020-08-06
Liquid -Fix Payment Data saving the blinded as a payment destination when it is unblinded

Kukks:


NicolasDorier:
If the merchant provide a blinded address and a payment is unblinded, the current behavior, not detecting the payment, is the right behavior. Is it what you try to fix?

Kukks:
This fixed an issue where a payment to an unblinded invoice address was showing the blinded address in the invoice payments list, because only the scriptpubkey is matched. Thinking about it, this might not be the right way to go.

Kukks:
Let me think about this a bit more later

PR #1821, opened 2020-08-06
Self-host webfonts

dennisreimann:
This inlines all webfonts we are using throughout the app, instead of loading them from Google Fonts. There are several benefits to this approach:
- Better performance
- We don't have any problems in regards to potential Google Fonts downtimes
- It protects our users privacy, as they don't have to request those fonts from a third-party

It sets up individual css files per font-face, so that they are easier to include and require on a per use-case basis.


dennisreimann:
@bolatovumar good point, done 👍

PR #1834, opened 2020-08-10
Refactor and decouple Payout logic to support other methods and to track payouts from external wallets

### LINKED ISSUE #1779: Payouts - Track offline transfer
If you pay a payout request with an external wallet outside of BTCPay, the payout system does not detect the transaction. 
When a payout is added, tell NBXplorer to track the address and make the NBX service detect it. Additionally, we should add a button "Mark payout as completed" so that edge cases are handled (and future support for lightning and other payment options)

Just the button above could work too but it's smoother with detecting it since we have the capability.

Kukks:
fixes #1779 
So that we can support lightning + external payout payments

This PR moves classes to separate folders and files and starts separating all the logic via a new `IPayoutHandler` which will be responsible for parsing Claims and tracking claim payments.


BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -0,0 +1,291 @@
...
+        _explorerClientProvider = explorerClientProvider;
+        _jsonSerializerSettings = jsonSerializerSettings;
+        _dbContextFactory = dbContextFactory;
+        _eventAggregator = eventAggregator;
+    }
+
+    public bool CanHandle(PaymentMethodId paymentMethod)
+    {
+        return paymentMethod.PaymentType == BitcoinPaymentType.Instance &&
+               _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethod.CryptoCode)?.ReadonlyWallet is false;
```
NicolasDorier, line 44: only readonly wallet can handle?

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    public async Task TrackClaim(PaymentMethodId paymentMethodId, IClaimDestination claimDestination)
+    {
+        if (!CanHandle(paymentMethodId))
```
NicolasDorier, line 49: Why is is not checked by the caller?

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -0,0 +1,291 @@
...
+        {
+            return;
+        }
+
+        var network = _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethodId.CryptoCode);
+        var explorerClient = _explorerClientProvider.GetExplorerClient(network);
+        if (claimDestination is IBitcoinLikeClaimDestination bitcoinLikeClaimDestination)
+            await explorerClient.TrackAsync(TrackedSource.Create(bitcoinLikeClaimDestination.Address));
```
NicolasDorier, line 57: I don't think we should do that. Because it means an attacker can falsely make the merchant believe he has been paid by sending money to himself. This would waste time of the merchant.

Better to make sure that the merchant paid it,  and if it is from an alternative wallet, we should just allow the merchant to manually consider it paid.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -0,0 +1,291 @@
...
+
+        if (o is NewBlockEvent || o is NewOnChainTransactionEvent)
+        {
+            await UpdatePayoutsInProgress();
+        }
+    }
+
+    public Task<decimal> GetMinimumPayoutAmount(PaymentMethodId paymentMethodId, IClaimDestination claimDestination)
+    {
+        if (!CanHandle(paymentMethodId))
```
NicolasDorier, line 115: the caller should have checked this.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: I don't think I understand, I don't see an attack vector? The person paying the claim would only be paying money to himself which only cheats himself. If a merchant sees a claim as paid, they don't do anything.

NicolasDorier: Imagine some action are being done by the merchant automatically as a result of this. For example, update of the books, removing it from his list of "things to do".

Imagine that the receiver did not actually wanted to provoke the payment. Maybe his wallet does not allow the generation of a new address until actually used and somebody else sent him money before the merchant could reimburse.

Now you have a merchant confused unable to fix the state of the pull payment. And a user that may not receive the refund because the merchant think it has been already done.

Kukks: cc @pavlenex

Kukks: I don't know then. I don't really agree, if I did a payment to a payout address from an external address, my instinct told me it should be marked and recorded now. 

What if it still records the transaction but need a "confirmation" from merchant that it was him? This way it is not forgotten and the merchant verifies too.

NicolasDorier: > if I did a payment to a payout address from an external address, my instinct told me it should be marked and recorded now.

This happen if you pay from the hot wallet.
If you pay from external wallet, I think this is good enough to just allow the merchant to manually mark it as paid.

Kukks: Thoughts are for the weak, updated branch to handle this!

Kukks:
Closing for now, #2046  replaces this PR for now as it is a subset of this (without the external wallet detection)

Issue #924, opened 2019-07-15
Defer onchain invoice creation until user selects the option


Sjors:
**Is your feature request related to a problem? Please describe.**
When Lightning payments are the default, BTCPay server still creates an onchain invoice immediately. This causes a Bitcoin address to be reserved, which can cause problems for wallets that don't scan ahead very far (e.g. BIP32 gap limit is 20, Bitcoin Core uses a keypool of 1000 at a time).

<img width="711" alt="Schermafbeelding 2019-07-15 om 22 07 12" src="https://user-images.githubusercontent.com/10217/61245496-eb630900-a74c-11e9-8e59-fd7b5fd3ebe1.png">

**Describe the solution you'd like**
Don't reserve the Bitcoin address until the user selects on-chain.

NicolasDorier:
Yes, but sadly, we can't change it without breaking the API.
Consumer of the API expect to be able to fetch the address after a call to create Invoice.

Sjors:
What do you mean by "after"? Is the address returned in the same call, or in a separate call? If the latter, then you could use just-in-time derivation, right?

NicolasDorier:
@Sjors basically you have one `Create invoice` call, and this must return all data for paying in all the different different payment methods. I can't change this without breaking API and clients.

NicolasDorier:
Though I think, we can do in the new, non backward compatible `GreenField` API that @Kukks is working on.

pavlenex:
Done in #1843

Issue #1728, opened 2020-07-10
Fix gap limit issues


Giszmo:
**Is your feature request related to a problem? Please describe.**
As [documented here](https://docs.btcpayserver.org/FAQ/FAQ-Wallet/#missing-payments-in-my-software-or-hardware-wallet), BTCPayServer does not play well with many light wallets and suggests to expose the seed to other software and potentially other hardware to resolve the issue.

**Describe the solution you'd like**
Please provide an option, maybe even default, that introduces a tiny bit of address reuse for the sake of trust in not having lost donations. Please resort to address reuse if the 20 address gap limit is soon to be exceeded and long expired invoices are available.

**Describe alternatives you've considered**
Please show address indices and addresses the user would have to send funds to, to help his wallet find the funds. The transaction list only shows txids but not address index or full list of addresses.

Kukks:
There are no lost donations, only wallet software that cannot see them.
Reusing address causing many dangerous side effects, such as having 2
customers being able to lay claim that a payment was made to their own
individual invoice or allows for a malicious user to spam invoices and
expose previous payment data to them.

I think we can start displaying indexes, and also a tool to generate the
specific private key if you input a seed.

Giszmo:
I understand your concerns and see how my solution 1 won't be the default ever but funds received to a non-standard address are lost to the user and need to be addressed somehow. As laid out in the documentation, wallet software usually follows the standard with a gap limit of 20 and while you claim that funds are not lost if you send beyond that, what if you send $1 to an address 120000 beyond that? It will cost more checking all the balances than that $1.

Does btcpayserver have any gap limit at all? I see good reason not to have one.
I bet this issue is something many people fall for and constantly comes up especially given [this issue](https://github.com/btcpayserver/btcpayserver/issues/924) which needlessly produces more gap for invoices that get paid. May I propose yet another solution?

BTCPayServer should send dust to addresses to make sure gaps can be bridged by standard BIP44 wallets. Should gaps build, show the user a red warning and assist him in filling gaps, ideally by bridging them with dust. The user might have to put a small amount into the btcpay hot wallet for that.

pavlenex:
@Giszmo That's been considered, but to do this, a hot wallet needs to be enabled. To use a hot wallet, you need to generate a wallet inside BTCPay so it defeats a whole purpose, since gap limit is only the issue with external wallets.

Giszmo:
[...]

Issue #863, opened 2019-05-26
BTCPay Forms (Form Builder) - Customizable checkout invoice 


pavlenex:
**Is your feature request related to a problem? Please describe.**

@Kukks and I brainstormed an idea a while ago about having customizable checkout pages. Leaving our ideas here for further discussion.

BTCPay Forms or Form Builder would be the app built-in into BTCPay that would allow users to create customizable checkouts and landing pages and gather information beyond default "refund email" that willl be parsed into invoice.

The use case:

1. E-commerce site uses pay button in store, (because they don't use CMS) and needs a more configurable way to ask for buyer address #505 and other relevant information in order to process the order.
Our invoices are not customizable, in order to change something, a user needs to fork btcpay and perform a bunch of technical issues. Instead store owner can use btcpay forms and customize checkout steps and fields that would be offered before the invoice is shown. 
2. The organization accepts the donation and needs customizable text fields at the checkout where the donor can give personal info, address, notes, instructions
3. People can create highly-customizable landing pages to accept donations, tips, etc.

**Describe the solution you'd like**
An app that would allow people to create customizable drag & drop forms, add fields that will parse information into an invoice.

**Provide examples**
![canvas](https://user-images.githubusercontent.com/36959754/58379703-84746f80-7fa7-11e9-895a-908e95986beb.png)

Realated issues #671 #505 #862

Liongrass:
I'd love something like that!

royserpa:
> 
> 
> I'd love something like that!
> 
> I'm mainly using html forms to create simple payment buttons, and I'd love to be able to easily add a line like ` <input type="text" name="Name" placeholder="Your Name">` to the `<form>` field
> and see it populate the name field in 'buyer information' in mybtcpay.com/invoices

Currently, all data gathering is being done on my website. I don't think it's needed to fork an instance as is.

Sure, it would be "easier" to implement, but with the help of the url notification and email notification methods, one can create a nice funnel and use your site's databases to store client data.

I think the btcpayserver server should stay only for btcpayserver info only. And in your website, you store all client data as you wish.

Liongrass:
With BTCPay Apps I can already run a store _without_ a separate website or database, which is really powerful. But I'm limited to the email address as the sole identifier. With just a two more field (name, company) I could use BTCPay properly for ticketing for events, and with just a few more (address) I could use it as a fully grown store.

JMadisonTP:
[...]

Issue #434, opened 2018-12-04
UI of cryptocurrency selection in the invoice


hubiktomas:
We were discussing with @bitcoinshirt that many customers do not get, how they can select the payment currency in the invoice. Maybe the tech people get that the dotted underline means that it is clickable, but many other customers do not get it and even though they know that there are other payment methods, they keep asking, how to select them.

NicolasDorier:
pign @rockstardev

astupidmoose:
I agree with this, we are still getting some customers asking "how to use this", or even worse some of them are simply sending LTC to our BTC addresses (This might not be fixable, not sure you can fix stupid lol).

I think maybe the currency in the "Pay with" needs to be formatted as a dropdown box, similar to the styling of the conversion tab on changelly. 

![image](https://user-images.githubusercontent.com/3916318/49715422-288a7a00-fc05-11e8-86fe-ebff3a6d80f8.png)

This boxed button approach gives a bit better hint at "You can click this and see more".

bitcoinshirt:
@astupidmoose  If you remember in past we had coin icons one next to the other, imo it's much better UX, but I think there was a technical problem with this setup. Especially if someone decides to have 10+ coins...

mariodian:
Why not just start with a screen that forces user to select a currency? When clicked, the currency list would roll up, the currency option would be added to the "Choose Currency" section (which also explicitly tells user that he can choose anytime) and the QR code would be displayed. That way user should be able to assume that when he clicks it again the currency list would roll down and he can choose again.

See the pic:

![checkout](https://user-images.githubusercontent.com/3636406/49867562-db491b00-fe45-11e8-80df-f0523a6acb33.png)

bitcoinshirt:
@mariodian This is good concept.

We discussed this in #22 already and exactly the idea that I had.
Two problems with this:
1. Adds additional step for the end-user on the checkout
2. In BTCPay user can pay invoice in multiple coins. You can pay 0.4 in BTC and 0.2 in LTC. How would end user switch currencies in that case? We probably should have a drop-down in the invoice for those.
I'm not sure if this technically creates any trouble on the back-end.

mariodian:
> Adds additional step for the end-user on the checkout

True, but if he doesn't want to pay with the predefined currency, he will need to perform an additional step anyway. If he's forced to choose a currency first at least he knows how to do it.

> In BTCPay user can pay invoice in multiple coins.

[...]

PR #1843, opened 2020-08-12
Make Invoice Create Faster And Fix Gap Limit Issue

xpayserver:
This make address reserve only when user "activate" payment method in ui. It is optional setting in store checkout ui so no difference existing store. 

This way BTC wallet only use address customer intend pay so it reduce gap limit issue almost entirely. In ETH case, it make it more likely for ETH and ERC to end in same wallet. Lightning no effect.

Also this make invoice creation very fast almost 10x faster


NicolasDorier:
I think @pavlenex will like that one.

@xpayserver I like the principle. However, rather than having this I would prefer a different model where we have two steps to the checkout process rather than hack around the interface.

1. Select your crypto currency
2. Show the current UX

Maybe @dennisreimann  might be interested to work on the UX to make that possible.

IMHO this mode should be the default.
But for the Bitpay's legacy API level, we will keep the current behavior, as to not break consumer of the API who expects finding the payment methods as soon as invoice created.

Ping @Kukks I think this is worth to discuss for the next invoice API. That it takes into account this option.

NicolasDorier:
related discussion https://github.com/btcpayserver/btcpayserver/issues/434

Kukks:
> I think @pavlenex will like that one.
> 
> @xpayserver I like the principle. However, rather than having this I would prefer a different model where we have two steps to the checkout process rather than hack around the interface.
> 
> 1. Select your crypto currency
> 2. Show the current UX
> 
> Maybe @dennisreimann might be interested to work on the UX to make that possible.
> 
> IMHO this mode should be the default.
> But for the Bitpay's legacy API level, we will keep the current behavior, as to not break consumer of the API who expects finding the payment methods as soon as invoice created.
> 
> Ping @Kukks I think this is worth to discuss for the next invoice API. That it takes into account this option.

I prefer this flow as choosing the invoice currency in advance means you cannot switch mid-way in the invoice. This flow will also be possible with #863 where you create an optiopn to choose your intended payment option
For Invoice API #1760, we can just add the "activated" property on the payment method array object (and create the post endpoint to activate like in here)

pavlenex:
Just a 🤖 linking issues, don't mind me.
Related #1728 #434
Closes #924

Kukks:
cc @astupidmoose I feel like we've talked about this before

astupidmoose:
Yea, I think this was an old issue which I just learned to work around. 

https://github.com/btcpayserver/btcpayserver/issues/22

rockstardev:
+1 on keeping current flow rather than choosing currency in advance and then not being able to switch midway. Ability to select while still on invoice is one of the biggest advantages we have. All the other checkouts suck because their engineering teams were too lazy to implement this advanced scenario.

NicolasDorier:
ok, I think I will merge this before next release then.

NicolasDorier:
@xpayserver I made some breaking change, can you rebase? It should not be hard.
You may want to do a hack I did for Monero in PaymentMethodId as well.

xpayserver:
> @xpayserver I made some breaking change, can you rebase? It should not be hard.
> You may want to do a hack I did for Monero in PaymentMethodId as well.

I have rebase

mariodian:
tACK - it really speeds things up

pavlenex:
@xpayserver do you mind resolving remaining conflicts, I promise we'll expedite the merge, since it's been approved a few times already.

Kukks:
@xpayserver can you rebase?

wallopthecat:
Is this being blocked by something?

Kukks:
> Is this being blocked by something?

We needed the author to rebase but seems like they did and did not get a notification. Let me rebase it myself and see

Kukks:
Rebased and fixed a number of things along with added tests. I think it is good to go now, will merge if there are no objections

rockstardev:
@Kukks can you push commit on @xpayserver repo, I don't have access:

```
From fba22a981048a2ac005de1f1c6a0e1fae2c72f49 Mon Sep 17 00:00:00 2001
From: rockstardev <rockstardev@users.noreply.github.com>
Date: Mon, 4 Jan 2021 23:19:22 -0600
Subject: [PATCH] Rearranging lazy payments position in UI to be near dependent
 Unified QR code

---
 BTCPayServer.Client/Models/StoreBaseData.cs          |  3 +--
 .../Controllers/GreenField/StoresController.cs       |  2 +-
 BTCPayServer/Controllers/StoresController.cs         |  2 +-
 BTCPayServer/Data/StoreBlob.cs                       |  3 +--
 .../StoreViewModels/CheckoutExperienceViewModel.cs   |  6 +++---
 BTCPayServer/Views/Stores/CheckoutExperience.cshtml  | 12 ++++++------
 6 files changed, 13 insertions(+), 15 deletions(-)

diff --git a/BTCPayServer.Client/Models/StoreBaseData.cs b/BTCPayServer.Client/Models/StoreBaseData.cs
index 7e0e76c5..ff4b9be6 100644
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -35,6 +35,7 @@ namespace BTCPayServer.Client.Models
         public bool LightningAmountInSatoshi { get; set; }
         public bool LightningPrivateRouteHints { get; set; }
         public bool OnChainWithLnInvoiceFallback { get; set; }
+        public bool LazyPaymentMethods { get; set; }
         public bool RedirectAutomatically { get; set; }
 
@@ -53,8 +54,6 @@ namespace BTCPayServer.Client.Models
 
         public string HtmlTitle { get; set; }
 
-        public bool LazyPaymentMethods { get; set; }
-
         [JsonConverter(typeof(StringEnumConverter))]
         public NetworkFeeMode NetworkFeeMode { get; set; } = NetworkFeeMode.Never;
diff --git a/BTCPayServer/Controllers/GreenField/StoresController.cs b/BTCPayServer/Controllers/GreenField/StoresController.cs
index 321d994d..e3f98c85 100644
--- a/BTCPayServer/Controllers/GreenField/StoresController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoresController.cs
@@ -168,8 +168,8 @@ namespace BTCPayServer.Controllers.GreenField
             blob.LightningAmountInSatoshi = restModel.LightningAmountInSatoshi;
             blob.LightningPrivateRouteHints = restModel.LightningPrivateRouteHints;
             blob.OnChainWithLnInvoiceFallback = restModel.OnChainWithLnInvoiceFallback;
-            blob.RedirectAutomatically = restModel.RedirectAutomatically;
             blob.LazyPaymentMethods = restModel.LazyPaymentMethods;
+            blob.RedirectAutomatically = restModel.RedirectAutomatically;
             blob.ShowRecommendedFee = restModel.ShowRecommendedFee;
             blob.RecommendedFeeBlockTarget = restModel.RecommendedFeeBlockTarget;
             blob.DefaultLang = restModel.DefaultLang;
diff --git a/BTCPayServer/Controllers/StoresController.cs b/BTCPayServer/Controllers/StoresController.cs
index 62892a51..79970f79 100644
--- a/BTCPayServer/Controllers/StoresController.cs
+++ b/BTCPayServer/Controllers/StoresController.cs
@@ -402,10 +402,10 @@ namespace BTCPayServer.Controllers
                 }
             }).ToList();
             vm.RequiresRefundEmail = storeBlob.RequiresRefundEmail;
-            vm.LazyPaymentMethods = storeBlob.LazyPaymentMethods;
             vm.LightningAmountInSatoshi = storeBlob.LightningAmountInSatoshi;
             vm.LightningPrivateRouteHints = storeBlob.LightningPrivateRouteHints;
             vm.OnChainWithLnInvoiceFallback = storeBlob.OnChainWithLnInvoiceFallback;
+            vm.LazyPaymentMethods = storeBlob.LazyPaymentMethods;
             vm.RedirectAutomatically = storeBlob.RedirectAutomatically;
             vm.ShowRecommendedFee = storeBlob.ShowRecommendedFee;
             vm.RecommendedFeeBlockTarget = storeBlob.RecommendedFeeBlockTarget;
diff --git a/BTCPayServer/Data/StoreBlob.cs b/BTCPayServer/Data/StoreBlob.cs
index eb593199..ea40ea7e 100644
--- a/BTCPayServer/Data/StoreBlob.cs
+++ b/BTCPayServer/Data/StoreBlob.cs
@@ -34,12 +34,11 @@ namespace BTCPayServer.Data
         [JsonConverter(typeof(Newtonsoft.Json.Converters.StringEnumConverter))]
         public NetworkFeeMode NetworkFeeMode { get; set; }
 
-        public bool LazyPaymentMethods { get; set; } = false;
-
         public bool RequiresRefundEmail { get; set; }
         public bool LightningAmountInSatoshi { get; set; }
         public bool LightningPrivateRouteHints { get; set; }
         public bool OnChainWithLnInvoiceFallback { get; set; }
+        public bool LazyPaymentMethods { get; set; }
         public bool RedirectAutomatically { get; set; }
         public bool ShowRecommendedFee { get; set; }
         public int RecommendedFeeBlockTarget { get; set; }
diff --git a/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs b/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
index 95aad703..2e758e5a 100644
--- a/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
+++ b/BTCPayServer/Models/StoreViewModels/CheckoutExperienceViewModel.cs
@@ -44,6 +44,9 @@ namespace BTCPayServer.Models.StoreViewModels
         [Display(Name = "Include lightning invoice fallback to on-chain BIP21 payment url")]
         public bool OnChainWithLnInvoiceFallback { get; set; }
 
+        [Display(Name = "Only enable the payment method after user explicitly activates it")]
+        public bool LazyPaymentMethods { get; set; }
+
         [Display(Name = "Redirect invoice to redirect url automatically after paid")]
         public bool RedirectAutomatically { get; set; }
 
@@ -66,9 +69,6 @@ namespace BTCPayServer.Models.StoreViewModels
         [Display(Name = "Custom HTML title to display on Checkout page")]
         public string HtmlTitle { get; set; }
 
-        [Display(Name = "Only enable the payment method after user explicitly activates it")]
-        public bool LazyPaymentMethods { get; set; }
-
         public List<PaymentMethodCriteriaViewModel> PaymentMethodCriteria { get; set; }
     }
 
diff --git a/BTCPayServer/Views/Stores/CheckoutExperience.cshtml b/BTCPayServer/Views/Stores/CheckoutExperience.cshtml
index f89783e9..29ab7497 100644
--- a/BTCPayServer/Views/Stores/CheckoutExperience.cshtml
+++ b/BTCPayServer/Views/Stores/CheckoutExperience.cshtml
@@ -62,12 +62,6 @@
                     <label asp-for="RequiresRefundEmail" class="form-check-label"></label>
                 </div>
             </div>
-            <div class="form-group">
-                <div class="form-check">
-                    <input asp-for="LazyPaymentMethods" type="checkbox" class="form-check-input"/>
-                    <label asp-for="LazyPaymentMethods" class="form-check-label"></label>
-                </div>
-            </div>
             <div class="form-group">
                 <div class="form-check">
                     <input asp-for="LightningAmountInSatoshi" type="checkbox" class="form-check-input" />
@@ -86,6 +80,12 @@
                     <label asp-for="OnChainWithLnInvoiceFallback" class="form-check-label"></label>
                 </div>
             </div>
+                <div class="form-check">
+                    <input asp-for="LazyPaymentMethods" type="checkbox" class="form-check-input" />
+                    <label asp-for="LazyPaymentMethods" class="form-check-label"></label>
+                </div>
+            </div>
             <div class="form-group">
                 <div class="form-check">
                     <input asp-for="RedirectAutomatically" type="checkbox" class="form-check-input" />
-- 
2.26.2.windows.1

```

rockstardev (approved):
LGTM... I'm kidna torn on the fact user needs to click on button to activate. But I guess we can always change it later - activate as soon as method is selected.

Unified QR Code is not working initially, but starts working if both onchain and offchain methods are activated. 

I would leave it like this... provided patch to move settings closer so it's more obvious to store owner - if they want lazy payment methods for now, they won't work combined with unified QR code.

https://user-images.githubusercontent.com/5191402/103610069-dbad2680-4ee4-11eb-980e-c8999540d3b3.mp4

pavlenex:
I'm with @rockstardev on this one, `Activate payment method` button is very confusing UX-wise. What does it even mean to a regular user and why are we adding an unnecessary step for the checkout here?

Since it's a Server Setting Policy, I can live with it, but would prefer payment method to be activated when currency is selected instead.

wallopthecat:
> 
> 
> I'm with @rockstardev on this one, `Activate payment method` button is very confusing UX-wise. What does it even mean to a regular user and why are we adding an unnecessary step for the checkout here?
> 
> Since it's a Server Setting Policy, I can live with it, but would prefer payment method to be activated when currency is selected instead.

How about changing the wording to "choose/use this currency" or "click to use CURRENCYNAME" - it prevents people who click on the other currencies so they get the total amount (to see if they can afford it) from generating addresses they don't intend to use

astupidmoose:
> I'm with @rockstardev on this one, `Activate payment method` button is very confusing UX-wise. What does it even mean to a regular user and why are we adding an unnecessary step for the checkout here?
> 
> Since it's a Server Setting Policy, I can live with it, but would prefer payment method to be activated when currency is selected instead.

Agreed. I can only imagine the number of customers who will open tickets asking "How do I activate a payment account" or something similar. I hope this is optional.

Kukks:
@xpayserver  @rockstardev  @pavlenex @astupidmoose  I have rebased and updated this PR so that it automatically activates when you load or change a payment method. There is no visible difference from the standard checkout now

Kukks:
@xpayserver For some reason I cannot rerun the tests on CircleCI because they are running on your account or something. Can you re run them

pavlenex:
This now LGTM, though I haven't tested it with altcoins, but it's good improvement that methods are swapped automatically. @astupidmoose can you test it perhaps?

NicolasDorier:
I think that later we should rather allow merchants to set which payments to activate by default.
But let's see how the feedback go for now.

pavlenex:
@NicolasDorier I agree this should be default if the feedback we get is good, good idea.

Issue #1805, opened 2020-08-04
Remove or redesign and re-position onion icon in the navigation menu


pavlenex:
**Is your feature request related to a problem? Please describe.**
![onion](https://user-images.githubusercontent.com/36959754/89325196-bba0df00-d688-11ea-8beb-833a8e2a780c.PNG)
The way Tor icon is positioned has always caused anxiety and endless tears when alone. First of all it's not really clear what's its purpose is. Majority of users would click on it and get a dead link. It doesn't provide any context whatsoever on what user should do. Not to mention that the alignment is .... bad.

We recently re-designed our login page to have "Copy Tor URL" that gives a bit more context. With #1626 in my opinion there's no really need to have onion icon in the header anymore. User familiar with Tor would be informed on the login page that it's supported and if they use Tor they'll be notified via tor browser.

I'd like to propose that we remove the onion icon it entirely.

If people don't agree with the removal of this icon, then maybe we can move it to footer and use the same CTA we use for the login page and re-position it in the footer.

**Describe alternatives you've considered**
- Remove onion icon entirely
- Re-position the onion CTA to footer and re-designed it to "Copy Tor URL"
- Do nothing and let Pavlenex cry

**Additional context**
People that know what Tor is and use Tor browser will already be notified, so for them the onion icon is not that useful. Even if they're unaware of ability to use Tor they'll see it on the login page.
![tor url](https://user-images.githubusercontent.com/36959754/89325334-eee36e00-d688-11ea-8ea8-caf5eb715cd7.PNG)

And if they already use Tor, they'll either automatically be redirected (depending on settings) or notified.
![onion2](https://user-images.githubusercontent.com/36959754/89325082-97450280-d688-11ea-9fcc-74a25311dd85.PNG)

That's why I think it's reasonable to remove it.

bolatovumar:
I agree with removing it.

dennisreimann:
I second removing it: 

* People using Tor are made aware by the Onion-Location header
* Those unfamiliar with Tor don't get anything out of clicking the onion (just an error page in their regular browser)
* Those who would like to switch to Tor can find and copy the onion URL on the login page or in the settings

Bosch-0:
Yep I agree with removing it from the top menu. 

Having it in plaintext in the footer is my suggestion. It's one of those advanced options that people who want to use it will find it.

PR #1849, opened 2020-08-18
Remove Tor link from navbar

NicolasDorier:
Removing the onion link in navbar 

Before:
image

Two reasons: 

1. Now with Tor Browser, it is easy to find the Tor link in the URL Bar.
2. People who clicks on it without Tor Browser leak DNS query.
3. It is ugly

@pavlenex  @dennisreimann


dennisreimann:
Yep! Closes #1805.

PR #1851, opened 2020-08-19
Allow auto approval of claims for pull payments

### LINKED ISSUE #1780: Pull Payments- Support auto approval of payouts
You currently need to manually approve payouts. When we add automated payouts, we need an option to have payouts automatically approved in order to be sent.

Just a checkbox when creating the pull payment to enable this.

Kukks:
closes #1780

### DIFF
--- a/BTCPayServer/HostedServices/PullPaymentHostedService.cs
+++ b/BTCPayServer/HostedServices/PullPaymentHostedService.cs
@@ -497,7 +497,7 @@ namespace BTCPayServer.HostedServices
                     await ctx.SaveChangesAsync();
                     if (req.ClaimRequest.PreApprove.GetValueOrDefault(ppBlob?.AutoApproveClaims is true) )
                     {
-                        payout.PullPaymentData.StoreData = await ctx.Stores.FindAsync(payout.PullPaymentData.StoreId);
+                        payout.StoreData = await ctx.Stores.FindAsync(payout.StoreDataId);
                         var rateResult = await GetRate(payout, null, CancellationToken.None);
                         if (rateResult.BidAsk != null)
                         {
--- a/BTCPayServer/HostedServices/PullPaymentHostedService.cs
+++ b/BTCPayServer/HostedServices/PullPaymentHostedService.cs
@@ -35,6 +35,7 @@ namespace BTCPayServer.HostedServices
         public string EmbeddedCSS { get; set; }
         public PaymentMethodId[] PaymentMethodIds { get; set; }
         public TimeSpan? Period { get; set; }
+        public bool AutoApproveClaims { get; set; }
         public TimeSpan? BOLT11Expiration { get; set; }
     }
 
@@ -117,6 +118,7 @@ namespace BTCPayServer.HostedServices
                 Limit = create.Amount,
                 Period = o.Period is long periodSeconds ? (TimeSpan?)TimeSpan.FromSeconds(periodSeconds) : null,
                 SupportedPaymentMethods = create.PaymentMethodIds,
+                AutoApproveClaims = create.AutoApproveClaims,
                 View = new PullPaymentBlob.PullPaymentView()
                 {
                     Title = create.Name ?? string.Empty,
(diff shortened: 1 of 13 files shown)


britttttk:
- It gives the awaiting approval notification even if the box is checked. 
- On the send page, confirmed payouts using auto-approve do not have the payout amount.

Kukks:
> * It gives the awaiting approval notification even if the box is checked.
> * On the send page, confirmed payouts using auto-approve do not have the payout amount.
Thanks for that, fixed :)

BTCPayServer/HostedServices/PullPaymentHostedService.cs
```diff
@@ -352,13 +354,32 @@ private async Task HandleCreatePayout(PayoutRequest req)
                 try
                 {
                     await ctx.SaveChangesAsync();
+                    if (ppBlob.AutoApproveClaims)
```
NicolasDorier, line 357: really hate this hack. I would prefer you call the approve inside the `Claim` method. You can pass the AutoApproveClaims to the `ClaimsResponse`.

Kukks: Not sure what you mean. I tried a few other ways before this but there was too much logic at play that could go wrong if I tried to optimize..

britttttk:
tACK - thanks for fixes @Kukks 

If anyone wants to see it:

image

Kukks:
@NicolasDorier  can we get this merged so we can move on with pull payments? I'm not sure what you wanted me to change code-wise.

NicolasDorier:
What bother me in this PR is that users does not know what it mean to "approve a payout".

It actually mean, in this context, to lockdown the rate as soon as the payout is requested.
Merging it as is is a very confusing UX.

Kukks:
> What bother me in this PR is that users does not know what it mean to "approve a payout".
> 
> It actually mean, in this context, to lock the exchange rate as soon as the payout is requested. But why would it be useful for users?
> Merging it as is is a very confusing UX.
> It does not mean "signing automatically".
> Ping @pavlenex , I think will agree with me.
> 
> I think it is doing something:
> 
> 1. That is not useful for anybody
> 2. Will confuse everybody as it is not doing what a reasonable user would expect

I really disagree with all of that.  
Approving a payout means to me: that it is ready to be paid.  Locking the rate is a side-effect not the feature.

When we add transaction batching (either via payjoin or other ways like a timer or criteria), approving a payout would queue it to the next batch, where a transaction is created, signed and broadcasted.

NicolasDorier:
> Locking the rate is a side-effect not the feature.

A side effect which is not intended, and actually the only effect that "approving" does at the moment.

Kukks:
> > Locking the rate is a side-effect not the feature.
> 
> A side effect which is not intended, and actually the only effect that "approving" does at the moment.

Rome wasn't built in a day, there are many building blocks needed to make pull payments useful. This is just the first one.

Kukks:
> > Locking the rate is a side-effect not the feature.
> 
> A side effect which is not intended, and actually the only effect that "approving" does at the moment.

Actually it is intended, when creating the pull payment, you've already decided to either lock the rate or allow the rate to be decided on the time the user creates the payout request.

NicolasDorier:
@Kukks the only reason why there is an approval step before paying is for locking the rate. There is no other reason.
Then it should not be called "Auto approval" but "Lock the rate as soon as the payout is created."

Kukks:
> @Kukks the only reason why there is an approval step before paying is for locking the rate. There is no other reason.
> Then it should not be called "Auto approval" but "Lock the rate as soon as the payout is created."

Meh, the state is `AwaitingApproval`, `automatic approval` fits. Locking rates only appliesto pull payments where the pull payment currenc and the payout options currencies are not the same. If you use the approval step for other things, it's completely different. One example is using it as a way to vet the destination.

britttttk:
I'm not commenting on the exchange rate lock, but please let me know if we need to update [this warning in the docs](https://docs.btcpayserver.org/PullPayments/#additional-use-cases-for-the-pull-payments-feature) from the outcome of this.

For the UX, I agree @NicolasDorier auto approve sounds like it should auto-pay (send) from the merchant's perspective. 

However I also agree with @Kukks that auto-approval is useful to communicate.

For anyone who wants to also give feedback this is what it looks like: 

**Without auto-approval:**
image

From the merchant perspective, there is no way to approve it. Paying it is the only way to approve it. 

**With auto-approval**
image

image

They appear the same after payout from both the requester and payer after payment. 

I think that the the communication of awaiting approval and auto approval is a useful thing to communicate (waiting for work to be sent, etc.) and the ux is clear.

Kukks:
> For the UX, I agree @NicolasDorier auto approve sounds like it should auto-pay (send) from the merchant's perspective.

This is planned and the main end-goal for this PR for me

Kukks:
> They appear the same after payout from both the requester and payer after payment.

This is different in #1834

MaiHistoryCompanyLimited:
{#1851}

PR #1856, opened 2020-08-19
attempt mysql fix

### LINKED ISSUE #1836: MySQL support broken in recent versions #2
**Describe the problem/bug**
MySQL support is not working with recent versions of BTCPayserver. This issue is related to https://github.com/btcpayserver/btcpayserver/issues/1224 but not the same.

**Your environment**
BTCPayserver 1.0.5.4, but I think it affects earlier versions too

**Logs (if applicable)**
`MySql.Data.MySqlClient.MySqlException (0x80004005): BLOB/TEXT column 'PullPaymentDataId' used in key specification without a key length
`
next time I get:

`MySql.Data.MySqlClient.MySqlException (0x80004005): Unknown table 'btcpayserver.RefundAddresses'
` but I guess that is related to incomplete migration failing in previous run.

Probably caused by 20200625064111_refundnotificationpullpayments.cs. `PullPaymentDataId` should have a `maxLength`. It might be worth thinking about some sort of integration tests catching such bugs earlier (altho I am fully aware that only PostgreSQL is officially supported).

Kukks:
fixes #1836

### DIFF
--- a/BTCPayServer.Data/Migrations/20200625064111_refundnotificationpullpayments.cs
+++ b/BTCPayServer.Data/Migrations/20200625064111_refundnotificationpullpayments.cs
@@ -70,10 +70,10 @@ namespace BTCPayServer.Migrations
                 {
                     Id = table.Column<string>(maxLength: 30, nullable: false),
                     Date = table.Column<DateTimeOffset>(nullable: false),
-                    PullPaymentDataId = table.Column<string>(nullable: true),
+                    PullPaymentDataId = table.Column<string>(maxLength: 30, nullable: true),
                     State = table.Column<string>(maxLength: 20, nullable: false),
                     PaymentMethodId = table.Column<string>(maxLength: 20, nullable: false),
-                    Destination = table.Column<string>(nullable: true),
+                    Destination = table.Column<string>(maxLength: 1000, nullable: true),
                     Blob = table.Column<byte[]>(nullable: true),
                     Proof = table.Column<byte[]>(nullable: true)
                 },
--- a/BTCPayServer.Data/Migrations/20200625064111_refundnotificationpullpayments.cs
+++ b/BTCPayServer.Data/Migrations/20200625064111_refundnotificationpullpayments.cs
@@ -73,7 +73,7 @@ namespace BTCPayServer.Migrations
                     PullPaymentDataId = table.Column<string>(maxLength: 30, nullable: true),
                     State = table.Column<string>(maxLength: 20, nullable: false),
                     PaymentMethodId = table.Column<string>(maxLength: 20, nullable: false),
-                    Destination = table.Column<string>(maxLength: 1000, nullable: true),
+                    Destination = table.Column<string>(nullable: true),
                     Blob = table.Column<byte[]>(nullable: true),
                     Proof = table.Column<byte[]>(nullable: true)
                 },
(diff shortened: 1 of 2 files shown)


rockstardev:
I love non-deterministic nature of your commit message.

NicolasDorier:
@Kukks can you also apply the `[MaxLength]` to the properties?

Again more pain with strings. I like binary blobs.

BTCPayServer.Data/Migrations/20200625064111_refundnotificationpullpayments.cs
```diff
@@ -70,10 +70,10 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 {
                     Id = table.Column<string>(maxLength: 30, nullable: false),
                     Date = table.Column<DateTimeOffset>(nullable: false),
-                    PullPaymentDataId = table.Column<string>(nullable: true),
+                    PullPaymentDataId = table.Column<string>(maxLength: 30, nullable: true),
                     State = table.Column<string>(maxLength: 20, nullable: false),
                     PaymentMethodId = table.Column<string>(maxLength: 20, nullable: false),
-                    Destination = table.Column<string>(nullable: true),
+                    Destination = table.Column<string>(maxLength: 1000, nullable: true),
```
NicolasDorier, line 76: I don't you don't need to change this one.

Kukks: fixe

ketominer:
Hi, it's me again! :)

I have the "MySql.Data.MySqlClient.MySqlException (0x80004005): Unknown table 'btcpay.RefundAddresses'" issue when upgrading from 1.0.4.4 to 1.0.6.8 but I don't have the MaxLength related issue (since that's fixed).

After logging the queries during migration, I see that it aborts just after creating the Notifications table (although this create succeeds) and doesn't seem to try to run the next operation which is create the PullPayments table (there is no trace of this CREATE TABLE in the full query log nor in the sql errors files which should log failed queries).

I then restarts the migration over and over again and fails on dropping RefundAddresses which is normal since this table is dropped first thing in the 20200625064111_refundnotificationpullpayments.cs file.

Any idea what could be causing that?

ketominer:
ok scrap that, the PullPayments create doesn't go through because the table is still here from my previous attempts. After a full cleanup, we fail just after the Refunds create table. So this leaves us with the classic
2021-02-07 19:00:27 btcpay[btcpay] @ localhost [127.0.0.1] ERROR 1170: BLOB/TEXT column 'InvoiceDataId' used in key specification without a key length : CREATE TABLE `Refunds` (
    `InvoiceDataId` longtext CHARACTER SET utf8mb4 NOT NULL,
    `PullPaymentDataId` longtext CHARACTER SET utf8mb4 NOT NULL,
    CONSTRAINT `PK_Refunds` PRIMARY KEY (`InvoiceDataId`, `PullPaymentDataId`),
    CONSTRAINT `FK_Refunds_Invoices_InvoiceDataId` FOREIGN KEY (`InvoiceDataId`) REFERENCES `Invoices` (`Id`) ON DELETE CASCADE,
    CONSTRAINT `FK_Refunds_PullPayments_PullPaymentDataId` FOREIGN KEY (`PullPaymentDataId`) REFERENCES `PullPayments` (`Id`) ON DELETE CASCADE
)

which is again a classic case of 4 GB wide (longtext) columns in btcpay :)

ketominer:
incoming PR!

PR #1857, opened 2020-08-19
Set roles when authenticating via greenfield

### LINKED ISSUE #1855: Cannot use Internal Lightning Node from a Greenfield authenticated user, even if "btcpay.server.canuseinternallightningnode" permission is present
**Describe the bug**
Cannot use Internal Lightning Node from a Greenfield authenticated user, even if "btcpay.server.canuseinternallightningnode" permission is present

**To Reproduce**
1. Create a Greenfield API Key with btcpay.server.canuseinternallightningnod
2. Try to use the Greenfield method GetInfo

**Expected behavior**
I should get the info from the internal lightning node

- BTCPay Server Version v1.0.5.2

By debugging, I think the problem is in the following line of code: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/GreenField/LightningNodeApiController.cs#L291

That should be modified into something like this
`
            return (_btcPayServerEnvironment.IsDevelopping || User.IsInRole(Roles.ServerAdmin) ||
                    (_cssThemeManager.AllowLightningInternalNodeForAll && !doingAdminThings) ||
                    User.HasClaim(GreenFieldConstants.ClaimTypes.Permission, Policies.CanUseInternalLightningNode));
`

What do you think?

Kukks:
fixes #1855


NicolasDorier:
I don't understand, can you make a test that fail without this?

NicolasDorier:
So if an admin create a "create invoice" api key, then the claim "Admin" will be attached?

NicolasDorier:
Can you write a test which would have caught #1855 ? I am surprised there is no test about this. I am almost sure we tested it.

Kukks:
@NicolasDorier  I have fixed the test, it was not caught because `BTCPayServerEnvironment.IsDeveloping` was always true.

PR #1861, opened 2020-08-22
Add store sort

### LINKED ISSUE #1837: Default to Alphabetic Ordering of Stores List
It's not related to an existing problem.

I'd like to see the list of Stores on the `stores` pages ordered alphabetically.

There are two alternatives that I can think of:

1. Keep it as is, which appears to be ordered by creation date (although this is not shown as a column in the list).
2. Provide a facility to choose the ordering (i.e. alphabetically, by creation date, by current balance, by most recent transaction, etc.).

**Provide examples**
See the screenshot showing the current ordering.  Note that this is not alphabetically ordered, and stores with identical names but numeric postfixes are in a non-logical order (i.e. 01, 03, 02 and 07, 09, 08).

Stores Ordering

**Additional context**
Nothing further to add although I see that there is a default ordering that is not alphabetic on most other tables\lists on the site.

bolatovumar:
Aug-22-2020 15-32-26

close #1837

### DIFF
--- a/BTCPayServer/Controllers/UserStoresController.cs
+++ b/BTCPayServer/Controllers/UserStoresController.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Threading.Tasks;
 using BTCPayServer.Data;
 using BTCPayServer.Models;
@@ -70,10 +71,40 @@ namespace BTCPayServer.Controllers
         }
 
         [HttpGet]
-        public async Task<IActionResult> ListStores()
+        public async Task<IActionResult> ListStores(
+            string sortOrder = null,
+            string sortOrderColumn = null
+        )
         {
             StoresViewModel result = new StoresViewModel();
             var stores = await _Repo.GetStoresByUserId(GetUserId());
+            if (sortOrder != null && sortOrderColumn != null) 
+            {
+                stores = stores.OrderByDescending(store => 
+                    {
+                        switch (sortOrderColumn)
+                        {
+                            case nameof(store.StoreName):
+                                return store.StoreName;
+                            case nameof(store.StoreWebsite):
+                                return store.StoreWebsite;
+                            default:
+                                return store.Id;
+                        }
+                    }).ToArray();
+
+                switch (sortOrder)
+                {
+                    case "desc":
+                        ViewData[$"{sortOrderColumn}SortOrder"] = "asc";
+                        break;
+                    case "asc":
+                        stores = stores.Reverse().ToArray();
+                        ViewData[$"{sortOrderColumn}SortOrder"] = "desc";
+                        break;
+                }
+            }
+
             for (int i = 0; i < stores.Length; i++)
             {
                 var store = stores[i];
--- a/BTCPayServer/Views/Apps/ListApps.cshtml
+++ b/BTCPayServer/Views/Apps/ListApps.cshtml
@@ -1,8 +1,13 @@
-﻿@using BTCPayServer.Services.Apps
-@model ListAppsViewModel
+﻿@model ListAppsViewModel
 @{
     ViewData["Title"] = "Apps";
+    var storeNameSortOrder = (string)ViewData["StoreNameSortOrder"];
+    var appNameSortOrder = (string)ViewData["AppNameSortOrder"];
+    var appTypeSortOrder = (string)ViewData["AppTypeSortOrder"];
+    var sortByDesc = "Sort by descending...";
+    var sortByAsc = "Sort by ascending...";
 }
+
 <section>
     <div class="container">
         @if (TempData.HasStatusMessage())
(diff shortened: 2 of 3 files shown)


bolatovumar:
> Nice.
> 
> I'm having a bit of trouble with the icon though:
> 
> * it's too bold
> * ideally it should be displayed all the time to hint that there's sorting functionality
> 
> What about something like the [font awesome sort icon](https://fontawesome.com/icons/sort?style=solid), but with state display: initially/unsorted grayed out arrows, sorted direction highlighted.

Yeah, I can update the icon. We already use this icon on another page (apps) so I just used the same icon here.

bolatovumar:
@dennisreimann Looks like I can't use the dualtone icon to highlight the sort direction as it's only available in the pro version of Font Awesome. How about this? I added a regular sort icon for non-sorted state and an alphanumeric ascending/descending icon for the sorted states.

Aug-23-2020 16-08-01

dennisreimann:
Pinging @dstrukt here, because he might also have input :)

dstrukt:
I do :)

- Agreed with the original sentiment, icon was way too large.
- I would only show the sort icon on column that it's currently affecting, keeps it cleaner and easier to parse with a glance. 
- I would also recommended the two-tone, it's the most widely used and recognizable pattern.

But given the FA pro constraint...I had more questions and clarifications, which will be helpful for the design system documentation:

1. Are we leveraging custom icons anywhere in the project or is it all based on the basic font-awesome package?
2. If not, was there interest in implementing the ability to support custom icons / utilizing icomoon / fontello, etc...? As I know I'll have custom icons, etc.. for us in the near future.

Assuming nothing custom, etc.. is possible, I think utilizing the A-Z could also work, however the left spacing on the icon should be adjusted slightly (2 or 4px).

Might be more than needed at this moment, but I've also mocked up a minimal state functionality for tables as well -- trying to get us to this point with all assets / components, and potentially with animatics in the future

TableStates

bolatovumar:
> 1. Are we leveraging custom icons anywhere in the project or is it all based on the basic font-awesome package?

I believe it's mostly just the standard font-awesome icons. There are SVGs used for logos and whatnot.

> 2. If not, was there interest in implementing the ability to support custom icons / utilizing icomoon / fontello, etc...? As I know I'll have custom icons, etc.. for us in the near future.

I don't see why not. As long as it's not a PITA to use from the development perspective.

> Assuming nothing custom, etc.. is possible, I think utilizing the A-Z could also work, however the left spacing on the icon should be adjusted slightly (2 or 4px).

I can do this.

> * I would only show the sort icon on column that it's currently affecting, keeps it cleaner and easier to parse with a glance.

Do you mean that there should be no icon shown initially and show it only once you click on the table heading to initiate the sort? This is in conflict of the original feedback from @dennisreimann (if I understand everything correctly):

> ideally it should be displayed all the time to hint that there's sorting functionality

Once we have the above questions answered I can go ahead with the changes and also update the other page which has sorting (apps page).

dstrukt:
There's always at least one sort active (which I guess isn't completely clear in that mock -- should have shown it switching instead of not there, instead of non-existent), but only one of the sort options should be visible at any given time.

Corrected (my apologies):

Screen Shot 2020-08-24 at 7 54 12 PM


@dennisreimann Curious to hear your thinking as well.

dennisreimann:
I like the version @dstrukt proposed, though I could see the need to display the arrows even for unsorted columns so that people know they can sort them. 

In general I think we should solve this with the tools we have currently and introduce custom icons or update them during the [Bootstrap 5 migration](https://github.com/btcpayserver/btcpayserver/issues/1866).

bolatovumar:
> There's always at least one sort active (which isn't completely clear in that first mock -- should have shown it switching, instead of non-existent), but only one of the sort options should be visible at any given time.
> 
> Corrected (my apologies):
> 
> Screen Shot 2020-08-24 at 7 54 12 PM
> 
> @dennisreimann Curious to hear your thinking as well.
> 
> @bolatovumar I've used icomoon in the past and it was no different than incorporating FA into the project - but I would also lean on a/the developer to give me their final say (if they don't mind) - creating and delivering the icons is easy.

There is a bit of an issue with having the table headers be grey AND not have the sorting icon shown by default because then that looks no different from any other table which doesn't have sorting. With this style of headings users would not know that the table is sortable unless they click on the heading.

For now, I think I will leave it at this version and we can update the styling if needed after the Bootstap 5 migration that @dennisreimann mentioned above.

bolatovumar:
Updated both pages which have table sorting to have the same look:

Screen Shot 2020-08-25 at 7 27 03 PM

Screen Shot 2020-08-25 at 7 27 21 PM

dennisreimann (approved):

pavlenex (approved):
tACK

PR #1870, opened 2020-08-27
Update AddDerivationSchemes_HardwareWalletDialogs.cshtml

lucasmoten:
Minor change for Electrum where menu choice is `Save backup`, not `Save Copy`


Kukks:
Seems like this is a change in the 4.x release. Wonder if we should add both `Save Copy / Save backup`

PR #1871, opened 2020-08-27
Add Value Criteria For Payment Method

xpayserver:
I upgrade lightning max/bitcoin min to support better control in store. Now can have setting only enable specific payment method only if value high/low I think make code simple more too and backward compatible with old feature
image


BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -207,11 +208,13 @@ internal async Task<InvoiceEntity> CreateInvoiceCoreRaw(InvoiceEntity entity, St
...
             {
                 currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, entity.Currency));
-                //TODO: abstract
-                if (storeBlob.LightningMaxValue != null)
-                    currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, storeBlob.LightningMaxValue.Currency));
-                if (storeBlob.OnChainMinValue != null)
-                    currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, storeBlob.OnChainMinValue.Currency));
+                foreach (var paymentMethodCriteria in store.GetPaymentMethodCriteria(_NetworkProvider, storeBlob))
+                {
+                    if (paymentMethodCriteria.Value != null)
```
Kukks, line 213: better to put in foreach with a `.Where`

BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -207,11 +208,13 @@ internal async Task<InvoiceEntity> CreateInvoiceCoreRaw(InvoiceEntity entity, St
...
+                    {
+                        currencyPairsToFetch.Add(new CurrencyPair(network.CryptoCode, paymentMethodCriteria.Value.Currency));
```
Kukks, line 215: you should make sure we dont have this storing duplicate rates to not ping rate providers too much

Kukks (approved):
LGTM, only small nits from my end and glad to be getting rid of that extra method on the payment method handler implementations!

BTCPayServer/Controllers/InvoiceController.cs
NicolasDorier: That's not too much of a problem, under the hood, the call actually don't reach exchange but get from cache.

NicolasDorier:
LGTM, @pavlenex you can test?

pavlenex:
To me, this looks a bit odd way to solve this  UX-wise. 

Maybe @dstrukt has an idea on how this can be solved in a better way

What we have now
PR1

What is proposed in this PR
PR

dstrukt:
@pavlenex will take a look today

dstrukt:
Few clarifications, but will use the mocks for reference:

1. We can clarify that the payment will be enabled if a there a minimum reached via invoice total, however, wouldn't we need the ability to disable it? Conversely, the ability to add another payment method easily?  (Unless there are other methods for doing this more easily / clearly?)

Screen Shot 2020-09-13 at 5 06 24 PM
Screen Shot 2020-09-13 at 5 06 46 PM

2. Stepping back from that specific feature, I think a few things on this screen need clarification and should broken out into respective sections... but I realize this pushes the scope of this ticket, so regardless of what we choose for #1 (which can be updated and reflected in the follow), here are some other ideas for discussion....

Explicitly some of the sections need better descriptors, and even in this mock they could be pushed a bit more, but overall, this is an initiative I'd like to push across the entire product.

Screen Shot 2020-09-13 at 5 04 15 PM

Feedback, need for clarification, etc.. welcome as always!

@pavlenex

dstrukt:
@Kukks and I chatted about the feature in MM. I had a few misunderstandings..made some changes.

An idea proposed by Kukks that gives the user the ability to select greater/less than..

The wording for the label could use a second pair of eyes, but I think it's clear enough.

Screen Shot 2020-09-15 at 12 21 06 AM

pavlenex:
IMO ^ is a perfect solution and quite clear. Good idea @dstrukt and @Kukks.

@xpayserver could you please make the change according to the proposal so that it's clearer on what's going on?

pavlenex:
tACK with a minor cosmetic change.

Looks great @xpayserver. I've tested it and here are a few minor notes (some of them may not be related to this PR, but perhaps are good to discuss them so we can improve stuff in the future).

I just suggest we change ` Enable payment methods only when amount is.. ` to `Enable payment method when the invoice total is` as originally suggested.

Unrelated to this PR, but an interesting edge-case.
1. Have 1 payment method, on-chain BTC for example.
2. Do not enable payment method if value is lower than 10 USD
3. Create an invoice at 9 USD
4. Infinite loop invoce creation called 
5. 
```
Exception thrown: 'BTCPayServer.BitpayHttpException' in BTCPayServer.dll
Exception thrown: 'BTCPayServer.BitpayHttpException' in System.Private.CoreLib.dll
Exception thrown: 'BTCPayServer.BitpayHttpException' in System.Private.CoreLib.dll
Loaded 'csmx3w15.vlp'. Cannot find or open the PDB file.
Exception thrown: 'BTCPayServer.BitpayHttpException' in System.Private.CoreLib.dll
Exception thrown: 'NBXplorer.Models.NBXplorerException' in System.Private.CoreLib.dll
Exception thrown: 'NBXplorer.Models.NBXplorerException' in System.Private.CoreLib.dll
Loaded 'c5vyykp5.rgn'. Cannot find or open the PDB file.
Loaded 'oqculfgf.d3s'. Cannot find or open the PDB file.
The thread 319301 has exited with code 0 (0x0).
Exception thrown: 'Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException' in System.Private.CoreLib.dll
```

NicolasDorier:
thinking merging that for next release.

> Infinite loop invoce creation called

This should not happen. It should say no payment method available.

BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -325,17 +328,29 @@ private async Task<PaymentMethod> CreatePaymentMethodAsync(Dictionary<CurrencyPa
...
+                        var limitValueCrypto = criteria.Value.Value / currentRateToCrypto.BidAsk.Bid;
+                        
+                        if (amount < limitValueCrypto && criteria.Above)
+                        {
+                            logs.Write($"{logPrefix} invoice amount below accepted value for payment method", InvoiceEventData.EventSeverity.Error);
+                            return null;
+                        }
+                        if (amount > limitValueCrypto && !criteria.Above)
+                        {
+                            logs.Write($"{logPrefix} invoice amount above accepted value for payment method", InvoiceEventData.EventSeverity.Error);
```
NicolasDorier, line 348: Should not be errors.

pavlenex:
> thinking merging that for next release.
> 
> > Infinite loop invoce creation called
> 
> This should not happen. It should say no payment method available.

@NicolasDorier Should I create a separate issue for this?

NicolasDorier:
no leave it here

NicolasDorier:
@pavlenex I think your "loops" came from your weird PC. Can you replicate again?

BTCPayServer/Controllers/InvoiceController.cs
Kukks: Looking at the diff, this was an error before

Kukks (approved):
Looks ok to me

PR #1872, opened 2020-08-28
Add invoice event severity

### LINKED ISSUE #1681: Mark Invoice Events as errors
We should add a new column to `InvoiceEvents` `string Severity` so that wen visualize error events in the invoice details page

Kukks:
closes #1681
image

### DIFF
--- a/BTCPayServer.Data/Data/InvoiceEventData.cs
+++ b/BTCPayServer.Data/Data/InvoiceEventData.cs
@@ -20,6 +20,7 @@ namespace BTCPayServer.Data
         }
 
         public string Message { get; set; }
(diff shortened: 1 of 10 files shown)


NicolasDorier:
awesome!

BTCPayServer.Data/Data/InvoiceEventData.cs
```diff
@@ -20,6 +20,7 @@ public DateTimeOffset Timestamp
         }
 
         public string Message { get; set; }
+        public EventSeverity? Severity { get; set; }
```
NicolasDorier, line 23: Can we remove the nullable part and have default to info?

PR #1875, opened 2020-08-28
Invoice notification email improvements

dennisreimann:
I find the invoice notification emails hard to grasp, because it's just an unformatted blob of JSON.

Here the most important information is displayed at the top, plus the option to toggle the full details JSON, which is now also formatted.
Also added the store name, so that one knows which store the invoice belongs to.

Notice: I've also deactivated email notifications for expired invoices. Maybe this needs to stay the way it was?

## Before

before

## After

after

### DIFF
--- a/BTCPayServer/HostedServices/InvoiceNotificationManager.cs
+++ b/BTCPayServer/HostedServices/InvoiceNotificationManager.cs
@@ -61,7 +61,7 @@ namespace BTCPayServer.HostedServices
             _StoreRepository = storeRepository;
         }
 
-        async void Notify(InvoiceEntity invoice, InvoiceEvent invoiceEvent, bool extendedNotification)
+        async Task Notify(InvoiceEntity invoice, InvoiceEvent invoiceEvent, bool extendedNotification)
         {
             var dto = invoice.EntityToDTO();
             var notification = new InvoicePaymentNotificationEventWrapper()
--- a/BTCPayServer/HostedServices/InvoiceNotificationManager.cs
+++ b/BTCPayServer/HostedServices/InvoiceNotificationManager.cs
@@ -13,6 +13,7 @@ using BTCPayServer.Payments;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Invoices;
 using BTCPayServer.Services.Mails;
+using BTCPayServer.Services.Stores;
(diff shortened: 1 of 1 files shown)


Kukks:
No bueno. This is a compatibility thing with bitpay and could break an
integration if we change from a raw Json. Some clients may expect to parse
the body directly

dennisreimann:
What a shame.

NicolasDorier:
If this is only changing the email, we don't give a shit about compat. People relying on email at a way to pass message between apps are nuts.

BTCPayServer/HostedServices/InvoiceNotificationManager.cs
```diff
@@ -40,23 +41,25 @@ public InvoicePaymentNotificationEventWrapper Notification
...
             _Client = httpClientFactory.CreateClient();
             _JobClient = jobClient;
             _EventAggregator = eventAggregator;
             _InvoiceRepository = invoiceRepository;
             _EmailSenderFactory = emailSenderFactory;
+            _StoreRepository = storeRepository;
         }
 
-        void Notify(InvoiceEntity invoice, InvoiceEvent invoiceEvent, bool extendedNotification)
+        async void Notify(InvoiceEntity invoice, InvoiceEvent invoiceEvent, bool extendedNotification)
```
NicolasDorier, line 62: never ever `async void` https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#async-void

NicolasDorier:
except the async void which should be async Task is seems good to me to merge.

BTCPayServer/HostedServices/InvoiceNotificationManager.cs
dennisreimann: Thanks for the pointer. Corrected with a15f5bef821ce8cc6d05472f9b8ed0c70b5e798d.

dennisreimann:
The test error looks unrelated: 

```
X CanUsePoSApp [1s 631ms]
  Error Message:
   The collection was expected to contain a single element, but it contained more than one element.
  Stack Trace:
     at BTCPayServer.Tests.AltcoinTests.CanUsePoSApp() in /source/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs:line 807
```

PR #1882, opened 2020-09-02
Add custom refund option

### LINKED ISSUE #1874: Add ability to generate a partial refund
When issuing refunds, we sometimes do not seek to refund the entire transaction. We'd like to instead specify a precise BTC or USD amount to refund. Currently, we manually create a new pull payment in order to accomplish this.

We'd like to have the ability to specify a custom refund amount after clicking the "issue refund" button on the order details page.

 It could look something like this. Would be cool if there was a way to enter the exact amount of BTC or USD; maybe a text entry field could appear below if selected.

CleanShot 2020-08-28 at 11 54 27@2x

The only alternative I could think of is to manually create the pull payment, which we currently do.

Kukks:
Allows you to specify an alternative refund amount and currency. This allows partial refunds and other negotiated terms
closes #1874

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -11,15 +11,18 @@ using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.Events;
 using BTCPayServer.Filters;
+using BTCPayServer.HostedServices;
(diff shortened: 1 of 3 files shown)


rockstardev (approved):
Nice usage of switches... tACK.

custom amount refund pull payment

NicolasDorier:
That's great

PR #1886, opened 2020-09-05
Allow admins to invite new users

Kukks:
* This refactors the email sending so that all the logic related to users and emails are now contained in one location.
* The Reset password screen has been updated from its ugly plain self to use the same layout as the login.
* An admin can now create a new account without specifying a password. A link is generated that can be given to the intended user to configure the password. If emails are configured, it also sends an email
* An admin can now create accounts that still require the user to verify their if the setting is enabled from the server settings. A link is generated that can be given to the intended user to configure the password. If emails are configured, it also sends an email.
* The above features can be used in conjunction: An email will have to verify their email through a link. Once verified, the user is redirected to setting the password.
* When an email has been verified OR a password has been set, users are now redirected to the login page with the email filled in and a success status message shown instead of a dedicated thank you page.


BTCPayServer/HostedServices/UserEventHostedService.cs
```diff
@@ -30,23 +31,60 @@ public class UserEventHostedService : EventHostedServiceBase
...
                     }
+                    else if (!await _userManager.HasPasswordAsync(userRegisteredEvent.User))
+                    {
+                        userPasswordResetRequestedEvent = new UserPasswordResetRequestedEvent()
+                        {
+                            CallbackUrlGenerated = userRegisteredEvent.CallbackUrlGenerated,
+                            User = userRegisteredEvent.User,
+                            RequestUri = userRegisteredEvent.RequestUri
+                        };
+                        goto passwordSetter;
```
NicolasDorier, line 66: ( ͡° ͜ʖ ͡°)

NicolasDorier: I shall pass it, but for that a proper method would have been more readable.

NicolasDorier:
ACK for me, but it needs a test.

NicolasDorier (changes requested):
Need a test

Kukks:
added test @NicolasDorier

Issue #1887, opened 2020-09-05
Wording/UX Fix: Word "Balance" should be "Transaction Amount" on wallet's list transactions page


maltokyo:
**Describe the bug**
On the wallet transaction list page, one of the columns is called `"Balance"`.  It should be called `"Transaction Amount"`

https://github.com/btcpayserver/btcpayserver/blob/eb0dee07d83db70d3f153fb7a6c2c12212e386ca/BTCPayServer/Views/Wallets/WalletTransactions.cshtml

**To Reproduce**
1. Go to Wallets in top bar
2. Click on "Manage" of one wallet
3. See image below, that column that says Balance is actually the transaction amount.

![image](https://user-images.githubusercontent.com/1388507/92313245-e785ec00-efc9-11ea-8651-a8008e6ca0bc.png)

I tried to fix myself, but somehow it said I had to make a branch, but I could only make a fork.  Can someone teach me how to "branch" it?

Zaxounette:
Hi @maltokyo !

You can follow this doc for your branch issue: https://docs.btcpayserver.org/Contribute/ContributeWrite/WriteSoftware/

On the string change, it's a nice find, gratz.

Zax.

maltokyo:
Thanks @Zaxounette !  Done!

maltokyo:
Closing as https://github.com/btcpayserver/btcpayserver/pull/1888 will fix this

Zaxounette:
@maltokyo FYI, a tip for your next issue and pull request, when signaling in a PR `Closes #ISSUE-NUMBER` like you did in #1888 the PR will automatically close that issue, so you don't need to manually close it.

maltokyo:
Ah, that's cool, TIL.  Thanks again for your help and for teaching me.

PR #1888, opened 2020-09-05
Wording/UX Fix on Wallet Tx Page

maltokyo:
Change Balance Column title to "Transaction Amount" to fix https://github.com/btcpayserver/btcpayserver/issues/1887

### DIFF
--- a/BTCPayServer/Views/Wallets/WalletTransactions.cshtml
+++ b/BTCPayServer/Views/Wallets/WalletTransactions.cshtml
@@ -90,7 +90,7 @@
                     </th>
                     <th class="text-left">Label</th>
                     <th>Transaction Id</th>
-                    <th class="text-right">Transaction Amount</th>
+                    <th class="text-right">Amount</th>
                     <th class="text-right"></th>
                 </tr>
             </thead>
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/Wallets/WalletTransactions.cshtml
```diff
@@ -90,7 +90,7 @@
                     </th>
                     <th class="text-left">Label</th>
                     <th>Transaction Id</th>
-                    <th class="text-right">Balance</th>
+                    <th class="text-right">Transaction Amount</th>
```
Zaxounette, line 93:
```suggestion
                    <th class="text-right">Amount</th>
```

The previous string being `Transaction ID`, keeping this string as `Amount` only will delete repetition, which is implied.

Zaxounette (approved):

BTCPayServer/Views/Wallets/WalletTransactions.cshtml
maltokyo: Done, thanks!

bolatovumar (approved):

BTCPayServer/Views/Wallets/WalletTransactions.cshtml
rockstardev: 0744e23

PR #1891, opened 2020-09-08
Allow entry in wallet send via fiat

### LINKED ISSUE #1890: Add ability to set fiat amount in send form
Add ability to set fiat amount in send form
image

here is a sendform sample with fiat amount/auto convert 
image

 If applicable provide examples, wireframes, sketches or images to better explain your idea.
image
provided from coinpayments.net send form

Kukks:
closes #1890

### DIFF
--- a/BTCPayServer/Controllers/WalletsController.cs
+++ b/BTCPayServer/Controllers/WalletsController.cs
@@ -471,6 +471,8 @@ namespace BTCPayServer.Controllers
 
             model.FeeSatoshiPerByte = model.RecommendedSatoshiPerByte.LastOrDefault()?.FeeRate;
             model.SupportRBF = network.SupportRBF;
+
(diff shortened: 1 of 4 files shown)


Zaxounette:
@Kukks Wouldn't this need rounding to the closest sat ? Seems like a lot of numbers after the comma.

Kukks:
> @Kukks Wouldn't this need rounding to the closest sat ? Seems like a lot of numbers after the comma.

Good call, fixed now

pavlenex (approved):

Kukks:
> tACK.
> 
> @Kukks is there a reasonable way to scale this to other currencies and avoid locking it down to the USD in the future?

Apparently, it currently works by fetching the currency that's tied to the region of your store's default language lol. 

We can change this later so that the user can update currency and then have it saved for the next time too

PR #1896, opened 2020-09-09
App: Simple Point Of Sale

### LINKED ISSUE #481: Quick Invoice (Keypad for POS)
Is it possible to implement a small keypad instead of  shopping chart? I need to use a smartphone with a small screen and I would like to insert only the requested amount, without images.
Here are two examples I've made

screenshot_20181129-181153
screenshot_20181207-185943

mariodian:
resolves #481 

Simple POS that lets you add a final amount in the selected currency. 

The store's logo can be displayed in the header if set (should we use a separate logo per each POS instance instead)?

The amount gets smaller as the number grows bigger.

There is no arithmetics feature because there's a possibly big list of situations that would need to be taken care of (javascript's float issue, negative amounts etc).

Screen Shot 2020-09-09 at 5 42 58 PM

### DIFF
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -351,6 +351,92 @@ namespace BTCPayServer.Controllers
 
         }
 
+        [HttpGet]
+        [Route("/apps/{appId}/simple-pos")]
+        [XFrameOptionsAttribute(XFrameOptionsAttribute.XFrameOptions.AllowAll)]
+        public async Task<IActionResult> ViewSimplePointOfSale(string appId)
+        {
+            var app = await _AppService.GetApp(appId, AppType.SimplePointOfSale);
+            if (app == null)
+                return NotFound();
+            var settings = app.GetSettings<SimplePointOfSaleSettings>();
+            var numberFormatInfo = _AppService.Currencies.GetNumberFormatInfo(settings.Currency) ?? _AppService.Currencies.GetNumberFormatInfo("USD");
+            double step = Math.Pow(10, -(numberFormatInfo.CurrencyDecimalDigits));
+            var store = await _AppService.GetStore(app);
+            var storeBlob = store.GetStoreBlob();
+
+            return View("ViewSimplePointOfSale", new ViewSimplePointOfSaleViewModel()
+            {
+                Title = settings.Title,
+                Currency = settings.Currency,
+                Step = step.ToString(CultureInfo.InvariantCulture),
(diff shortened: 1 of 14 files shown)


BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -351,6 +351,93 @@ public async Task<IActionResult> ContributeToCrowdfund(string appId, ContributeT
...
+                Currency = settings.Currency,
+                Step = step.ToString(CultureInfo.InvariantCulture),
+                CustomCSSLink = settings.CustomCSSLink,
+                CustomLogoLink = storeBlob.CustomLogo,
+                AppId = appId,
+                EmbeddedCSS = settings.EmbeddedCSS
+            });
+        }
+
+        [HttpPost]
```
Kukks, line 380: I think we can reuse the same submission endpoint as the original POS.

BTCPayServer/Services/Apps/AppType.cs
```diff
@@ -3,7 +3,8 @@ namespace BTCPayServer.Services.Apps
     public enum AppType
     {
         PointOfSale,
-        Crowdfund
+        Crowdfund,
+        SimplePointOfSale
```
Kukks, line 7: This should not be a separate app

Kukks (changes requested):
I see you created this as an entirely new app, but it should only be a new view for the existing POS app. In the past few months, we've split the POS to support loading different views for it (static vs cart) which should allow extending for this new mode. 

A lot of the code here can be removed and reduced to mostly a new cshtml view + css/js stack

mariodian:
I did it as a new app, because the old one is in hacky javacsript and this one is in Vue. Didn't want to mix two different frontends within a single app. 

My goal is to rewrite the old PoS into Vue at some point and then perhaps merge these two apps?

mariodian:
Also, this PoS has no inventory, unlike the old one. Not sure if mixing them together is a good idea.

Kukks:
> I did it as a new app, because the old one is in hacky javacsript and this one is in Vue. Didn't want to mix two different frontends within a single app.
> 
> My goal is to rewrite the old PoS into Vue at some point and then perhaps merge these two apps?

The views are separate so the hacky version can live standalone while the new one can be in Vue.



> Also, this PoS has no inventory, unlike the old one. Not sure if mixing them together is a good idea.

That's fine, as long as no `itemCode` is passed in the submission, the inventory system won't be used at all.

mariodian:
Ok, I’ll add it under the existing app and  (hopefully) rewrite the existing pos into vue later.

PR #1898, opened 2020-09-09
Greenfield: Rename API key redirect params; switch to POST body

dennisreimann:
This way it is clearer what to expect and it also make the parameter easier to consume.

### DIFF
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -335,7 +335,7 @@ namespace BTCPayServer.Tests
         {
             var source = driver.FindElement(By.TagName("body")).Text;
             var json = JObject.Parse(source);
-            return json.GetValue("key")!.Value<string>();
+            return json.GetValue("apiKey")!.Value<string>();
         }
     }
 }
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
@@ -11,6 +10,7 @@ using BTCPayServer.Security.GreenField;
 using BTCPayServer.Tests.Logging;
 using BTCPayServer.Views.Manage;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using OpenQA.Selenium;
 using Xunit;
 using Xunit.Abstractions;
@@ -109,7 +109,6 @@ namespace BTCPayServer.Tests
                         tester.PayTester.HttpClient);
                 });
 
-
                 //let's test the authorized screen now
                 //options for authorize are:
                 //applicationName
@@ -120,8 +119,9 @@ namespace BTCPayServer.Tests
                 //redirect
                 //appidentifier
                 var appidentifier = "testapp";
+                var callbackUrl = tester.PayTester.ServerUri + "postredirect-callback-test";
                 var authUrl = BTCPayServerClient.GenerateAuthorizeUri(tester.PayTester.ServerUri,
-                    new[] { Policies.CanModifyStoreSettings, Policies.CanModifyServerSettings }, applicationDetails: (appidentifier, new Uri("https://local.local/callback"))).ToString();
+                    new[] { Policies.CanModifyStoreSettings, Policies.CanModifyServerSettings }, applicationDetails: (appidentifier, new Uri(callbackUrl))).ToString();
                 s.Driver.Navigate().GoToUrl(authUrl);
                 Assert.Contains(appidentifier, s.Driver.PageSource);
                 Assert.Equal("hidden", s.Driver.FindElement(By.Id("btcpay.store.canmodifystoresettings")).GetAttribute("type").ToLowerInvariant());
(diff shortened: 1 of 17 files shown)


Kukks (changes requested):
Since it is a post (which I completely forgot), we should actually send these props in the body

dennisreimann:
Boy, that escalated quickly

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -157,16 +156,17 @@ public async Task<IActionResult> AddApiKey()
                     }
 
                     //we have a key that is sufficient, redirect to a page to confirm that it's ok to provide this key to the app.
-                    return View("Confirm",
-                        new ConfirmModel()
+                    return View("ConfirmAPIKey",
```
Kukks, line 159: Why a new view?

dennisreimann: 1. we need to pass through the params in the form body vs. the URL query
2. preparation for adding the „create a new separate key“ button in addition to the confirm („reuse existing key“) button

NicolasDorier:
tests do not pass

dennisreimann:
The integration test failures seeem unrelated (rate providers), but I'm still unhappy about how this turned out:
https://github.com/btcpayserver/btcpayserver/blob/a37b442c9607986ab397101526573abc9a4c52c8/BTCPayServer/Views/Shared/PostRedirect.cshtml

I couldn't find a way to apply attributes conditionally and Razor wouldn't let me use the form `action` in addition to the `asp-` attributes.  Couldn't just `if` the opening tag either, so it ended up like this.

BTCPayServer.Tests/ApiKeysTests.cs
```diff
@@ -132,13 +132,13 @@ public async Task CanCreateApiKeys()
...
-                var apiKeyRepo = s.Server.PayTester.GetService<APIKeyRepository>();
-
-                await TestApiAgainstAccessToken(results.Single(pair => pair.Key == "key").Value, tester, user,
-                    (await apiKeyRepo.GetKey(results.Single(pair => pair.Key == "key").Value)).GetBlob().Permissions);
+                // var apiKeyRepo = s.Server.PayTester.GetService<APIKeyRepository>();
+                //
+                // IEnumerable<KeyValuePair<string, string>> results = url.Split("?").Last().Split("&")
+                //     .Select(s1 => new KeyValuePair<string, string>(s1.Split("=")[0], s1.Split("=")[1]));
+                // await TestApiAgainstAccessToken(results.Single(pair => pair.Key == "key").Value, tester, user,
+                //     (await apiKeyRepo.GetKey(results.Single(pair => pair.Key == "key").Value)).GetBlob().Permissions);
```
dennisreimann, line 141: Not sure how to test thosee now that they are part of the POST body and we land on an external page.

BTCPayServer.Tests/SeleniumTester.cs
```diff
@@ -308,8 +308,9 @@ public string CreateInvoice(string storeName, decimal amount = 100, string curre
             currencyEl.Clear();
             currencyEl.SendKeys(currency);
             Driver.FindElement(By.Id("BuyerEmail")).SendKeys(refundEmail);
-            Driver.FindElement(By.Name("StoreId")).SendKeys(storeName + Keys.Enter);
-            Driver.FindElement(By.Id("Create")).ForceClick();
+            Driver.FindElement(By.Name("StoreId")).SendKeys(storeName);
+            Driver.FindElement(By.Id("Create")).Click();
+
```
dennisreimann, line 313: This is related to the ChromeDriver upgrade.

BTCPayServer.Tests/ApiKeysTests.cs
dennisreimann: Re-added with 3fe7515bcae457e6f9eec5c3dd6655bf46ccd13d.

dennisreimann:
Now the PR can be reviewed. The main goal was to update two things regarding the callback the consumer app receives:
- send parameters as POST body instead of as part of the URL query string
- renaming the parameters to be more clear: `user` became `userId`, `key` became `apiKey`.

While working on this I also made some minor UI updates to the confirm view:

confirm

The separate confirm view got introduced to prepare the addition of a third button for the "create a new api key" button use case, which I'd like to tackle in a separate PR.

Also added a non-javascript version for the post redirect view:

nojs

BTCPayServer/Controllers/HomeController.cs
```diff
@@ -199,6 +200,18 @@ public IActionResult RecoverySeedBackup(RecoverySeedBackupViewModel vm)
...
+        public ActionResult PostRedirectCallbackTestpage(IFormCollection data)
+        {
+            var list = data.Keys.Aggregate(new Dictionary<string, string>(), (res, key) =>
+            {
+                res.Add(key, data[key]);
+                return res;
+            });
+            return Json(list);
+        }
+
```
dennisreimann, line 214: I saw that there is the `#if DEBUG` conditional. If this also works for the tests or there is something similar for those, wee could use this here too.

Kukks: you can use it there for tests

BTCPayServer/BTCPayServer.csproj
```diff
@@ -57,6 +57,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Selenium.WebDriver.ChromeDriver" Version="85.0.4183.8700" />
```
Kukks, line 60: remove please

BTCPayServer/wwwroot/swagger/v1/swagger.template.authorization.json
```diff
@@ -86,7 +86,7 @@
                         }
                     },
                     "307": {
-                        "description": "Redirects to the specified url in `redirect` with query string values for `key` (the api key created or matched), `permissions` (the permissions the user consented to), and `user` (the id of the user that consented) upon consent"
+                        "description": "Redirects to the specified url in `redirect` with query string values for `apiKey` (the api key created or matched), `permissions` (the permissions the user consented to), and `userId` (the id of the user that consented) upon consent"
```
Kukks, line 89:
```suggestion
                        "description": "Makes browser do an HTTP POST request to the specified url in `redirect` with a JSON body consisting of `apiKey` (the api key created or matched), `permissions` (the permissions the user consented to), and `userId` (the id of the user that consented) upon consent"
```

Kukks (changes requested):

BTCPayServer/Controllers/HomeController.cs
dennisreimann: Added, thanks!

Kukks (approved):

PR #1902, opened 2020-09-11
GreenField: Invoices: Add Refund Email endpoint

Kukks:
This reverts commit 709b06baa330fc670b2e4358dc69ecdb7b6aa4e5.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.Invoices.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Invoices.cs
@@ -47,6 +47,17 @@ namespace BTCPayServer.Client
             return await HandleResponse<InvoiceData>(response);
         }
 
+        public virtual async Task<InvoiceData> AddCustomerEmailToInvoice(string storeId, string invoiceId,
+            AddCustomerEmailRequest request, CancellationToken token = default)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+            var response = await _httpClient.SendAsync(
+                CreateHttpRequest($"api/v1/stores/{storeId}/invoices/{invoiceId}/email", bodyPayload: request,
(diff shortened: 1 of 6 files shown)


NicolasDorier:
The reason why we removed that is that I don't want to consider Refund Email in a different way than any data that can be in the Metadata.
I would prefer having a method to change Metadata in general.

Kukks:
> The reason why we removed that is that I don't want to consider Refund Email in a different way than any data that can be in the Metadata.
> I would prefer having a method to change Metadata in general.

Hmmm, I am scared of the implications of making the metadata mutable. Thoughts?

NicolasDorier:
I am not scared about that, this is just a dumb data store for merchant.
What I am scared of is that the metadata is inside the blob, and this blob might get updated at same time of the metadata change, losing one of the change.

I need to check better what is the risk.

Kukks:
> I am not scared about that, this is just a dumb data store for merchant.
> What I am scared of is that the metadata is inside the blob, and this blob might get updated at same time of the metadata change, losing one of the change.
> 
> I need to check better what is the risk.

We should probably do some sort of queue service like we have for pull payments. I think it is an excellent way to avoid concurrency issues.

Kukks:
Hmm, would it make sense to:
* Add an invoice log when the metadata changes
* Create an IPN when data changes
* Store a history of all states the metadata had?

Kukks:
Keep in mind that OrderId is also stored there, changing that midway through an invoice payment could have consequences of integrations with ecommerce plugins.

Kukks:
Moreover is that the metata invoice update would require authentication. The current email update doesnt need any auth in the Checkout UI. This would mean you cannot recreate the checkout UI entirely elsewhere as is?

Kukks:
No consensus on this. closing.

PR #1913, opened 2020-09-15
Ensure new bolt11 invoice on partial payment to btcpay invoice

Kukks:


mariodian:
The LN payment after the partial onchain payment doesn't seem to be noticed by the system. 

So the flow was like this:

- I created a 0.00074143 BTC invoice. 
- Paid 0.00034143 BTC on chain. 
- Switched to LN payment
- The checkout page showed the correct due amount
- Paid the rest via CLI
- The output from the CLI seemed correct (including the right amount, and `"status": "complete"`)
- The checkout page still wants me to finish the payment (I can keep paying the LN invoice as much as I want)
- The invoice in the administration only sees the partial onchain payment

Tested on regtest.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -47,8 +47,17 @@ public class LightningLikePaymentHandler : PaymentMethodHandlerBase<LightningSup
...
             var invoice = paymentMethod.ParentEntity;
-            var due = Extensions.RoundUp(invoice.Price / paymentMethod.Rate, network.Divisibility);
+            decimal due = Extensions.RoundUp(invoice.Price / paymentMethod.Rate, network.Divisibility);
+            try
+            {
+                due = paymentMethod.Calculate().Due.ToDecimal(MoneyUnit.BTC);
+            }
+            catch (Exception)
+            {
+                // ignored
```
NicolasDorier, line 59: why would it throws?

Kukks: this throws during invoice creation but does not if you are updating

Kukks: can probably do a  more efficient check

BTCPayServer/Payments/Lightning/LightningListener.cs
```diff
@@ -134,6 +143,51 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+                {
+                    var invoice = await _InvoiceRepository.GetInvoice(inv.InvoiceId);
+                    var paymentMethods = invoice.GetPaymentMethods()
+                        .Where(method => method.GetId().PaymentType == PaymentTypes.LightningLike).ToArray();
+                    var store = await _storeRepository.FindStore(invoice.StoreId);
+                    if (paymentMethods.Any())
+                    {
+                        var logs = new InvoiceLogs();
+                        logs.Write("Partial payment detected, attempting to update all lightning payment methods with new bolt11 with correct due amount.", InvoiceEventData.EventSeverity.Info);
+                        foreach (var paymentMethod in paymentMethods)
```
NicolasDorier, line 159: there is one listener per shitcoin, so you need to filter that per network.

Kukks: need to update every single one as the amount due has changed and each needs a new bolt11

NicolasDorier: yes, but this code is running for all listener.
If you have say LTC + BTC + GRS, you will end up creating 3 invoice per crypto currency

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
Kukks: this throws during invoice creation but does not if you are updating

BTCPayServer/Payments/Lightning/LightningListener.cs
Kukks: No, this code runs on `LightningListener` not `LightningInstanceListener` so it will only run once per event

PR #1916, opened 2020-09-17
Add more labels (payment requests, apps, better payout label)

Kukks:


Kukks:
Color palette of all existing and new label colors: https://coolors.co/cedc21-51b13e-489d77-3f88af-5093b6

Issue #1912, opened 2020-09-15
Custom redirect_url for PoS


mariodian:
When PoS is used as a donation page (even Nicolas uses it that way), it would be nice if user could be redirected to a custom page after the payment, eg "thank you" page on your own website like [this](https://freedomnode.com/thanks/).

Currently, when user pays an invoice, they get redirected back to the PoS which in the case of a donation page looks like you're a greedy bastard asking for more donations :)

I think it'd make sense to either set the settings for the store and use that (similar to redirectAutomatically option) or define the redirect_url per app. 

This would also help those that use Cart PoS as a simple eshop. Redirecting to a custom summary or a thank you page would make more sense, imo.

mariodian:
Btw I should be able to add that functionality if people are interested and if it gets approved. Seems like a very easy change.

bolatovumar:
For some reason I thought we already had this feature but I can't actually find a setting for it.

Kukks:
It should be an easy change, go for it. I think I will try and group some of the more  advanced options in the POS app to reduce bloat after this PR.

mariodian:
@kukks so should you set the URL per app (I prefer this) or per store?

Kukks:
per app

B369:
@mariodian Amazing that #1924 is coming. I'm about to use the  Pos App for invoicing customer. Classic PDF invoicing with bank information but going to add a Bitcoin accepted bottom with a hyperlink to the Pos. 
#1924 was one missing feature.

Another missing  feature would be :
Currently I tick the "add your email for refund option "that i could identify who paid it. 
Wondering if that would be a huge thing to have instead add your email -> add your invoice number. That number could show up on INVOICES - Product information section -> Item code

I feel on the end it could be a complete new App like ...PointOfSales, Crowdfund, Invoice

mariodian:
@B369 please open a **new issue** for that one. 

P.S. I'm not affiliated with BTCPay so I don't know what the roadmap is and whether if such feature is even desirable. Let's see what others have to say about it.

PR #1924, opened 2020-09-23
Custom redirectUrl for PoS

mariodian:
solves #1912

Should I also add `redirectUrl` in the UI where you manually create an invoice?


Kukks:
No don't add it there, pretty sure we removed it because it confused people

britttttk:
tACK

>pretty sure we removed it because it confused people

I will document these redirects after this is merged

bolatovumar (approved):

Kukks (approved):

PR #1925, opened 2020-09-23
BTCPay Server Extensions

Kukks:
demo


Zaxounette:
Nicolas, you need to give him harder challenges ! :D

Eskyee:
Looking dope kukks 🙌

mariodian:
After you install an extension, the "Install" link should disappear IMO. 

Screen Shot 2020-10-04 at 3 47 14 PM

Also, installing/uninstall the extension crashes the whole server on the next request (so for example after you click `install` and then click `Server -> Maintenance` it will crash it). 

I don't see any error logs, just:

````
info: PayServer:      DelayedTransactionBroadcasterHostedService successfully exited...
info: PayServer:      Internal Socks HTTP Proxy closed
info: PayServer:      CssThemeManagerHostedService successfully exited...
info: PayServer:      CheckConfigurationHostedService successfully exited...
info: PayServer:      PullPaymentHostedService successfully exited...
info: PayServer:      NBXplorerListener successfully exited...
info: PayServer:      Disconnected from WebSocket of NBXplorer (BTC)
info: PayServer:      LightningListener successfully exited...
info: PayServer:      Stop watching invoices
info: PayServer:      RatesHostedService successfully exited...
info: PayServer:      DynamicDnsHostedService successfully exited...
info: PayServer:      TorServicesHostedService successfully exited...
info: PayServer:      NewVersionCheckerHostedService successfully exited..
````

Kukks:
Good call, will be sure to parse the scheduled commands when loading that
page.
The way the system currently works is that when you click
install/update/uninstall, it will create a set of commands in a txt file to
run on the next startup. (install:app.btcpay, delete:extension, etc)
And for a quick demo, i made it wait 2 seconds and then kill the server
after you schedule one of those actions

mariodian:
> The way the system currently works is that when you click
install/update/uninstall, it will create a set of commands in a txt file to
run on the next startup. (install:app.btcpay, delete:extension, etc)
And for a quick demo, i made it wait 2 seconds and then kill the server
after you schedule one of those actions

Oh ok, I see the file. Shouldn't it restart the server though? (at least that's what it says was gonna happen in the green alert after you install the plugin).

mariodian:
I think the difference between how remote and local plugins are displayed is a bit confusing (see the pic below).

Screen Shot 2020-10-04 at 5 32 04 PM

The locally installed plugin should have the gray card footer with "some kind of info" too. 

Either:

````
Current version: 1.0.0.0
Remote version: n/a
````

or explaining that this plugin can't be updated remotely. 

Also, it's missing the "Uninstall" link.

Kukks:
> > The way the system currently works is that when you click
> > install/update/uninstall, it will create a set of commands in a txt file to
> > run on the next startup. (install:app.btcpay, delete:extension, etc)
> > And for a quick demo, i made it wait 2 seconds and then kill the server
> > after you schedule one of those actions
> 
> Oh ok, I see the file. Shouldn't it restart the server though? (at least that's what it says was gonna happen in the green alert after you install the plugin).

Yeah, mine is just assuming that you are running the dockerised version which will auto restart the container if the process stops. Will need a bit more trickery to build it in to make it restart itself

mariodian:
> Will need a bit more trickery to build it in to make it restart itself

That may be difficult because you may run it directly, via systemd or any other way.

Kukks:
> > Will need a bit more trickery to build it in to make it restart itself
> 
> That may be difficult because you may run it directly, via systemd or any other way.

Yeah, so maybe I can add a new config option: `BTCPAY_EXTENSIONS_AUTOMATICSHUTDOWN:"true"`

mariodian:
Maybe just the green alert could tell you to manually restart the server when you don’t run via docker? Not sure if that’s possible.

Kukks:
> Maybe just the green alert could tell you to manually restart the server when you don’t run via docker? Not sure if that’s possible.

I think I will not automatically try to restart the server as it may disrupt checkouts if you do this during a busy period. Instead, I'll show an alert and if running the docker version, provide a link to "restart"

pavlenex:
@NicolasDorier Could you please take a look at this so that we can determine if it's something Kukks can continue to code on for the next release?

NicolasDorier:
tests fails

Kukks:
> tests fails

fixed

Kukks:
@mariodian I updated the UI a bit. Can you check it out now?

Kukks:
https://i.imgur.com/4uDsChe.png
https://i.imgur.com/D0BbYJl.png
https://i.imgur.com/hYPrgmH.png
https://i.imgur.com/LVqfPhI.png

mariodian:
@kukks this hasn't been addressed: https://github.com/btcpayserver/btcpayserver/pull/1925#issuecomment-703229113

Also the `.card-footer` has different paddings for "pending action", "install" and "uninstall". I'd remove `.py-0` from it entirely.

Kukks:
@mariodian check out the #1981 branch

PR #1931, opened 2020-09-30
Add QR code scan/show for PSBT  + Import wallet via QR

Kukks:
This PR introduces support to show and read PSBTs in BC-UR format via animated QR codes.  This allows you to use BTCPay with HW devices such as Cobo Vault and Blue wallet to sign transactions without ever exposing the keys outside of that device.
Spec: https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md
I've also bumped the QR code library we sue as it had a bug with large datasets.


NicolasDorier:
I think we will need some better tooling in PSBT.
Now PSBT Update will include previous transaction, but there is no option to turn it off contrary to the Send wallet, which may be problematic for those wallet.

I start thinking that whether or not including previous tx in the PSBT should be set in the wallet settings rather than Send.

Kukks:
We can leave the settings for the next release, but we should add it to the settings and when you set seomthing via the advanced settings in the wallet send screen, it would update the wallet settings.

For example, toggling coin selection should have it enabled for the next time. 
Toggling utxo change should NOT set it for next time though.
And then we can also add the optional payjoin parameters such as output subtitution toggle

BTCPayServer/Controllers/StoresController.BTCLike.cs
```diff
@@ -112,6 +112,18 @@ private DerivationSchemeSettings GetExistingDerivationStrategy(string cryptoCode
                     vm.Confirmation = false;
                     return View(nameof(AddDerivationScheme), vm);
                 }
+            }else if (!string.IsNullOrEmpty(vm.WalletFileContent))
```
dennisreimann, line 115: nit: formatting

BTCPayServer/Models/WalletViewModels/WalletPSBTViewModel.cs
```diff
@@ -13,6 +13,7 @@ public class WalletPSBTViewModel
         public string CryptoCode { get; set; }
         public string Decoded { get; set; }
         string _FileName;
+        public string PSBTHex{ get; set; }
```
dennisreimann, line 16: nit: formatting (missing spoace before `{`)

BTCPayServer/Views/Shared/CameraScanner.cshtml
```diff
@@ -0,0 +1,177 @@
...
+            </div>
+        </div>
+    </div>
+</div>
+<style>
+    .pending-action {
+        position: absolute;
+        width: 100%;
+        height: 100%;
+
```
dennisreimann, line 40: nit: formatting (remove blank line)

BTCPayServer/Views/Shared/CameraScanner.cshtml
```diff
@@ -0,0 +1,177 @@
...
+                this.workload = [];
+                this.errorMessage = "";
+            },
+            close: function() {
+                this.loaded = false;
+                this.data = "";
+                this.workload = [];
+                this.errorMessage = "";
+            },
+            onDecode(content) {
```
dennisreimann, line 102: We should be consistent in how functions are defined, I'd favor this function declaration style as it is more concise.

BTCPayServer/Views/Shared/CameraScanner.cshtml
```diff
@@ -0,0 +1,177 @@
...
+                        } else if (error.name === 'NotSupportedError') {
+                            self.errorMessage = 'This page is served in non-secure context (HTTPS, localhost or file://)'
+                        } else if (error.name === 'NotReadableError') {
+                            self.errorMessage = 'Couldn\'t access your camera. Is it already in use?'
+                        } else if (error.name === 'OverconstrainedError') {
+                            self.errorMessage = 'Constraints don\'t match any installed camera.'
+                        } else {
+                            self.errorMessage = 'UNKNOWN ERROR: ' + error.message
+                        }
+                    })
```
dennisreimann, line 172: nit: indentation

BTCPayServer/Views/Stores/AddDerivationScheme.cshtml
```diff
@@ -216,7 +217,20 @@
     <script src="~/js/StoreAddDerivationScheme.js" type="text/javascript" defer="defer" asp-append-version="true"></script>
     <script src="~/js/vaultbridge.js" type="text/javascript" defer="defer" asp-append-version="true"></script>
     <script src="~/js/vaultbridge.ui.js" type="text/javascript" defer="defer" asp-append-version="true"></script>
+    
+    <script src="~/vendor/vuejs/vue.min.js" asp-append-version="true"></script>
+    <script src="~/vendor/vue-qrcode/vue-qrcode.min.js" asp-append-version="true"></script>
+    <script src="~/vendor/bc-ur/web-bundle.js" asp-append-version="true"></script>
+    <script src="~/vendor/vue-qrcode-reader/vue-qrcode-reader.browser.js" asp-append-version="true"></script>
+    <link href="~/vendor/vue-qrcode-reader/vue-qrcode-reader.css" rel="stylesheet" asp-append-version="true"/>
```
dennisreimann, line 225: Would bundling make sense here? We are reusing the same combination in WalletPSBT.cshtml

PR #1934, opened 2020-10-03
Add Created date to user, add verified column in list and make user l…

Kukks:
…ist use same model as modern lists


NicolasDorier:
every test fail

PR #1938, opened 2020-10-05
Do not show password in clear text in email configuration (Fix #1790)

### LINKED ISSUE #1790: SMTP Passwords in Cleartext
When I was setting up SMTP, I noticed that the SMTP password was accessible in the clear. I see that this was discussed in a previous issue to change it from `type=password` to `type=text`: https://github.com/btcpayserver/btcpayserver/pull/954

Screen Shot 2020-07-30 at 4 47 45 PM

I would be concerned from a security perspective if someone got access to my BTCPay Server web portal and used my GSuite / Office365 credentials within to pivot further.

**Suggestion** 

Here's an example of how Grafana handles the same thing:

Screen Shot 2020-07-30 at 4 32 23 PM

You're still able to update and test a new password, but you're not able to read what the previous password was unless you actually look in the database. Essentially the password field in Grafana is populated with a placeholder if you go to view it within the web UI. 

In our BTCPay scenario, I suppose someone with access could just add their SSH key but I would hope that a security-conscious administrator would harden or restrict SSH access before going live to production.

NicolasDorier:
image

Fix #1790

### DIFF
--- a/BTCPayServer/Views/Shared/EmailsBody.cshtml
+++ b/BTCPayServer/Views/Shared/EmailsBody.cshtml
@@ -1,4 +1,4 @@
-﻿@model BTCPayServer.Models.ServerViewModels.EmailsViewModel
+@model BTCPayServer.Models.ServerViewModels.EmailsViewModel
 
 
 <partial name="_StatusMessage" />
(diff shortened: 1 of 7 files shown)


BTCPayServer/Views/Shared/EmailsBody.cshtml
```diff
@@ -27,83 +27,98 @@
...
+                    <label asp-for="Settings.Password"></label>
+                    <input asp-for="Settings.Password" type="password" value="@Model.Settings.Password" class="form-control" />
+                    <span asp-validation-for="Settings.Password" class="text-danger"></span>
+                }
+                else
+                {
+                    <label asp-for="Settings.Password"></label>
+                    <div class="input-group">
+                        <input value="Configured" type="text" readonly class="form-control" />
+                        <div class="input-group-append fiat-value">
```
Kukks, line 109:
```suggestion
                        <div class="input-group-append">
```

BTCPayServer/Views/Shared/EmailsBody.cshtml
```diff
@@ -27,83 +27,98 @@
...
             </div>
             <div class="form-group">
-                <label asp-for="Settings.Password"></label>
-                <input asp-for="Settings.Password" value="@Model.Settings.Password" class="form-control"/>
-                <span asp-validation-for="Settings.Password" class="text-danger"></span>
+                @if (!Model.PasswordSet)
+                {
+
+                    <label asp-for="Settings.Password"></label>
+                    <input asp-for="Settings.Password" type="password" value="@Model.Settings.Password" class="form-control" />
```
Kukks, line 101:
```suggestion
                    <input asp-for="Settings.Password" type="password" class="form-control" />
```

Kukks (changes requested):

NicolasDorier:
added test

Kukks (approved):

PR #1946, opened 2020-10-06
Fix modal overlay

### LINKED ISSUE #1941: When collapsed, the sync window blocks the version text in the footer
**Describe the bug**
New collapsible sync window is awesome, but we should just make sure it doesn't block the version in the footer.
While you can still the version via various hacks, sometimes a first-time user whoes node is syncing may have a question or a bug report that would require their version to be futher troubleshoot. 

**To Reproduce**
1. Minimize sync window while it's syncing
2. Scroll to the bottom to check version
3. It's hard to see the version

**Expected behavior**
A person should be able to see the version at a bottom even though the node is syncing.

**Screenshots**
Screenshot 2020-10-05 at 18 41 14

- BTCPay Server Version [available in the right bottom corner of footer] Woooot Sir my window is blocking a version 😄  jk it's v1.0.5.6
- Deployment Method: [e.g. Docker, Manual, Third-Party-hoist] Docker
 - Browser [e.g. chrome, safari] - Chrome, 13" MacAir

Kukks:
fixes #1941

### DIFF
--- a/BTCPayServer/Views/Shared/LayoutPartials/SyncModal.cshtml
+++ b/BTCPayServer/Views/Shared/LayoutPartials/SyncModal.cshtml
@@ -28,6 +28,7 @@
     </div>
     <script>
         var syncModalContent = $('#syncModalContent');
+        var syncModal = $('#syncModal');
         if (localStorage.getItem("sync-collapsed")){
             syncModalContent.removeClass("show");
         }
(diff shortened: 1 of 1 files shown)


Kukks:
oops @dennisreimann your fix will do it but this one adds that it "stickies" and gives you your original styling based on scroll position.

Too much or?

dennisreimann:
Performance-wise this is very expensive, because for every pixel scrolled it'll `getBoundingClientRect` and do the calculation. The alternative would be to use the `IntersectionObserver` API. Given that this only affects pages during sync I think it's fine either way, but the easiest would be to accept the fact that it's just a little further away from bottom and be done with it.

PR #1947, opened 2020-10-06
Fix onion location not always working

### LINKED ISSUE #1881: Checkout URL incompatible with .onion
Having issue with .onion address for website. 
User loads a cart in tor browser and proceeds to checkout then our btcpayserver is selected. 
URL loads with invoice?id=
User selects the .onion redirect and 404 error is produced. 
When manually pulling the invoice up from the server side it uses a URL with /i/ in place of above. 
Copying this into tor brower and selecting .onion correctly displays the invoice.

Kukks:
closes #1881


dennisreimann (approved):
Works, thanks for fixing, kukks! 🧅

Before:

before

After:

after

PR #1948, opened 2020-10-06
Rename access tokens to Bitpay in stores

Kukks:
A lot of users are confusing it with greenfield api keys so this should help a little bit.
IMO, we should remove from the nav and nest it under integrations later on.


NicolasDorier:
mmmh I really don't like ot mention Bitpay in a page that is often viewed by people. @pavlenex

pavlenex:
These were suggested:
`Legacy API`
`Obsoleted API`

I personally like legacy but you guys told me it's not possible for technical reason, I'd call it like that in the menu at least where it's visible, not a fan of it being mentioned either.

britttttk:
>don't like ot mention Bitpay in a page that is often viewed by people 

I agree, I definitely NACK using bitpay name. Btcpay and Bitpay is already confusing enough for newbies... Also it seems like a good thing to avoid for some trademark reasons, just my advice.

Kukks:
Hmm ok, what should we do then? We need to rename from "Access Tokens" as people seem to think it is for GreenField. 

My suggestion would be to nest the link under the new "Integrations" somewhere.

britttttk:
@Kukks tooltip to docs? :D

NicolasDorier:
I think moving to integration is not a good idea either. Because it kind of  break lot's of our doc, tutorial and video.
Maybe we could add better text in the access token page.

PR #1964, opened 2020-10-08
Decouple payment type icon from UI

Kukks:
We should put an effort to not couple payment methods/types together after spending so much time decoupling things.


NicolasDorier:
Not the right place to complain from it, but I thought the badge  🔗 to be misleading.

When I saw it in the invoice list, I thought it was a link I could click on.

PR #1977, opened 2020-10-13
Improve payment print styles

### LINKED ISSUE #1957: Add invoice like print-view for the payment-request
Once the user pays (or gets paid) a payment request, he may need to print it out and keep for the reference in a more of a formal invoice format.

While for the web, users don't care about the format, for documentation/paperwork it would be nice if our print-view is a bit of more suited for that use-case.

- Create an invoice-like print-view for payment requests that would show all the relevant data an invoice would usually have
- Some optional fields in the payment request UI might be needed for this?

Builds on aftet #1930 is merged.

dennisreimann:
Allows for export as invoice PDF to be used in accounting. Closes #1957.

The main option for providing additional information and styling it is the description field. This isn't available for the pull payment, that's why that PDF is rather sparse.

### DIFF
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -235,7 +235,7 @@ namespace BTCPayServer.Controllers
 
             var statusesAllowedToDisplay = new List<InvoiceStatus>() { InvoiceStatus.New };
             var validInvoice = result.Invoices.FirstOrDefault(invoice =>
-                Enum.TryParse<InvoiceStatus>(invoice.Status, true, out var status) &&
+                Enum.TryParse<InvoiceStatus>(invoice.StatusFormatted, true, out var status) &&
                 statusesAllowedToDisplay.Contains(status));
 
             if (validInvoice != null)
@@ -297,7 +297,7 @@ namespace BTCPayServer.Controllers
             }
 
             var invoices = result.Invoices.Where(requestInvoice =>
-                requestInvoice.Status.Equals(InvoiceState.ToString(InvoiceStatus.New),
+                requestInvoice.StatusFormatted.Equals(InvoiceState.ToString(InvoiceStatus.New),
                     StringComparison.InvariantCulture) && !requestInvoice.Payments.Any());
 
             if (!invoices.Any())
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -234,9 +234,7 @@ namespace BTCPayServer.Controllers
             }
 
             var statusesAllowedToDisplay = new List<InvoiceStatus>() { InvoiceStatus.New };
-            var validInvoice = result.Invoices.FirstOrDefault(invoice =>
-                Enum.TryParse<InvoiceStatus>(invoice.StatusFormatted, true, out var status) &&
-                statusesAllowedToDisplay.Contains(status));
+            var validInvoice = result.Invoices.FirstOrDefault(invoice => statusesAllowedToDisplay.Contains(invoice.Status));
 
             if (validInvoice != null)
             {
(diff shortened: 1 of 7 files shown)


britttttk:
tACK - My only comment is that I feel the abbreviation TX should be the complete word transaction, unless it is preferable on small screens to keep it short. 

image

pavlenex:
Second Brtit's suggestion for Transaction ID.
Haven't tested, but a minor UI issue that would confuse my accountant if I was to forward them this invoice, it would lead them into thinking I was paid way more than the original amount. 

Why do we show a full amount for underpaid invoice? Is it to early for me and my usual lack of coffee rant, or is something odd here?
Capture

dennisreimann:
@pavlenex This is indeeed confusing, good you bring it up. The price column displays the total amount of the invoice – the actual paid value is displayed in the transactions list on the right … from what I've seen we might have the data to display the actual paid price (`InvoiceEntity` has a paid property in addition to the amount), which would be the right thing here. Will try to get a hold of that property here …

pavlenex:
Okay, I just wanted to know if you want a separate issue to solve that since it's not really related to a print view or do you prefer to do it here 😄

dennisreimann:
Ready for review now :)

history

- [payment-request.pdf](https://github.com/btcpayserver/btcpayserver/files/5404101/payment-request.pdf)
- [pull-payment.pdf](https://github.com/btcpayserver/btcpayserver/files/5404102/pull-payment.pdf)

pavlenex:
tACK. Incredible work on details here Dennis. Love the rates column and the color statuses you've added. 
This is such a wonderful view right now that it's hard to keep the excitement. 
TACK
TACK2

If I'd had to nit-pick, the date we have in the payment-requests sometimes would appear as `Today at 9PM` which may be a weird format for an invoice, but it's not really related to the PR, but the way we display the date in general.

dennisreimann:
You are right about the date – I can take care of that and show the full date in the print view.

pavlenex (approved):
This now works great in print view, tACK!

dennisreimann:
I've rearranged the information in the payments history table and added the receive date for transactions.

Think it works better this way and the PR is complete now imho.

[payment-request.pdf](https://github.com/btcpayserver/btcpayserver/files/5417808/payment-request.pdf)

payments

dennisreimann:
Rebased it.

BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -235,7 +235,7 @@ public async Task<IActionResult> ViewPaymentRequest(string id)
 
             var statusesAllowedToDisplay = new List<InvoiceStatus>() { InvoiceStatus.New };
             var validInvoice = result.Invoices.FirstOrDefault(invoice =>
-                Enum.TryParse<InvoiceStatus>(invoice.Status, true, out var status) &&
+                Enum.TryParse<InvoiceStatus>(invoice.StatusFormatted, true, out var status) &&
```
NicolasDorier, line 238: this is wrong. StatusFormatted might not map to a specific InvoiceStatus

dennisreimann: I had to [rename `Status to StatusFormatted`](https://github.com/btcpayserver/btcpayserver/pull/1977/files#diff-db9aee17fb1347e6816037980ff6a3fb18a8b2bc06b718b26097c340ba4bcb39L142), because I also needed the `InvoiceState State` and `InvoiceStatus Status` properties.

NicolasDorier: I don't understand this logic here.
`StatusFormatted` is actually `StateFormatted` the state is status + exceptionstatus

dennisreimann: You are right, so this would be correct here:

```csharp
var validInvoice = result.Invoices.FirstOrDefault(invoice =>  statusesAllowedToDisplay.Contains(invoice.Status));
```

dennisreimann: Fixed with 1f8b746d7b62cab8c77b599320f58b63e0a09e54.

PR #1985, opened 2020-10-14
Use base65 instead of hex for BIP78

### LINKED ISSUE #1984: Payjoin client incompatibility with BIP78
Current implementation of Payjoin client in BTCPay isn't compatible with BIP78, as it encodes request in hex, not base64. Tested with JoinMarket payjoin receiver and it fails.

Likely this line needs to be changed, but I don't have test environment for that right now (I used testnet.demo.btcpayserver.org) - https://github.com/btcpayserver/btcpayserver/blob/f88c02cccd9b0c1fb25c4633049f4df79b4db522/BTCPayServer/Services/PayjoinClient.cs#L317.

See https://github.com/zkSNACKs/WalletWasabi/pull/4427.

Kukks:
fixes #1984

### DIFF
--- a/BTCPayServer/Services/PayjoinClient.cs
+++ b/BTCPayServer/Services/PayjoinClient.cs
@@ -314,7 +314,7 @@ namespace BTCPayServer.Services
             using (HttpClient client = CreateHttpClient(endpoint))
             {
                 var bpuresponse = await client.PostAsync(endpoint,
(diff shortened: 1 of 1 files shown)


NicolasDorier:
base65 is probaby better than 64.

Kukks:
only because base66 comes next

PR #1987, opened 2020-10-15
Remove SQLite as default option

Kukks:
IF MERGED, STORES WITH NO DB CONFIG WILL NOT START UNTIL CONFIG IS UPDATED WITH `sqlitefile=sqlite.db`

### DIFF
--- a/BTCPayServer/Configuration/DefaultConfiguration.cs
+++ b/BTCPayServer/Configuration/DefaultConfiguration.cs
@@ -27,7 +27,6 @@ namespace BTCPayServer.Configuration
             app.Option("--chains | -c", $"Chains to support as a comma separated (default: btc; available: {chains})", CommandOptionType.SingleValue);
             app.Option("--postgres", $"Connection string to a PostgreSQL database", CommandOptionType.SingleValue);
             app.Option("--mysql", $"Connection string to a MySQL database", CommandOptionType.SingleValue);
-            app.Option("--sqlite", $"Connection string to a SQLite database", CommandOptionType.SingleValue);
             app.Option("--sqlitefile", $"File name to an SQLite database file inside the data directory", CommandOptionType.SingleValue);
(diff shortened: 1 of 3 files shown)


BTCPayServer/Configuration/DefaultConfiguration.cs
```diff
@@ -25,8 +25,10 @@ protected override CommandLineApplication CreateCommandLineApplicationCore()
...
             app.Option("--allow-admin-registration", $"For debug only, will show a checkbox when a new user register to add himself as admin. (default: false)", CommandOptionType.BoolValue);
             app.Option("--chains | -c", $"Chains to support as a comma separated (default: btc; available: {chains})", CommandOptionType.SingleValue);
-            app.Option("--postgres", $"Connection string to a PostgreSQL database (default: SQLite)", CommandOptionType.SingleValue);
-            app.Option("--mysql", $"Connection string to a MySQL database (default: SQLite)", CommandOptionType.SingleValue);
+            app.Option("--postgres", $"Connection string to a PostgreSQL database", CommandOptionType.SingleValue);
+            app.Option("--mysql", $"Connection string to a MySQL database", CommandOptionType.SingleValue);
+            app.Option("--sqlite", $"Connection string to a SQLite database", CommandOptionType.SingleValue);
+            app.Option("--sqlitefile", $"File name to an SQLite database file inside the data directory", CommandOptionType.SingleValue);
```
NicolasDorier, line 31: this is confusing to have two parameters. Best is to keep only `--sqlite` and put an example in the description.

Kukks: that one requires that you have a filepath which could break if you change your data dir and your db was in there.
I would either keep sqlitefile only or both

NicolasDorier: then let's keep sqlitefile

Kukks: done

Issue #1959, opened 2020-10-07
When creating a store redirect to a wallet creation page


pavlenex:
**Is your feature request related to a problem? Please describe.**
When people create a store, in most cases they create it and forget that they need a wallet. Store without a wallet is not complete and pretty much useless.

This issue builds on continues improvements we offered in #1290 and should direct people to #1703 for overall better experience.

**Describe the solution you'd like**
The workflow is as following. While a store is created we should automatically redirect users to #1703 where they can configure their wallet. We can offer either a checkbox in create a store page or simply do it behind the scenes so when a user creates a store they'll be moved to a wallet page.

This issue is just a remind for a workflow but heavily relies on what we implement in #1703

@rockstardev may be able to expand more on technicalities. Also @NicolasDorier had some ideas.

rockstardev:
What I had in mind is that you just redirect to Store page, but:

1. Rearrange sections so that you have General, Wallet, and then Payment
2. Have persistent warning in Wallet section to notify user Derivation Scheme (XPUB) should be defined

So, small steps for start... if you like this direction I can open PR to start going in that direction: 

![image](https://user-images.githubusercontent.com/5191402/95504032-cec67880-0971-11eb-9f67-ce357d0a8f75.png)

rockstardev:
We can also improve descriptions as part of this PR, for Lightning paragraph we could go with something like:

```
If you want to receive payments over Lightning you will need to define a connection to a Lightning Node which will generate Lightning Network (LN) invoices. Keep in mind that Lightning Node requires a hot wallet to operate, as well as channels with incoming liquidity.
```

pavlenex:
@rockstardev your suggestion above seems to be what I had in mind for #1290. I just opened this issue to remind us that once we merge in #1703 we can use this issue to set up some sort of redirection when user creates a store to the wallet page, is that what you were thinking as well?

For lightning ACK, it makes sense. But I also think that process there could be improved, having to click on a `click here` link is not the best UX.

rockstardev:
No, my thinking was along the lines of not doing any extra redirection after store creation... just informing the user directly on the store page what needs to be done.

I really think we shouldn't make this too complicated with wizards - because if you wanted to enforce certain steps on the store you would do all that on "Create Store" page and do small wizard like:
 1. Pick a store name
 2. Define a wallet
 3. Define lightning connecting if you want it
 4. Define rate

[...]

PR #1990, opened 2020-10-16
Store hints/warnings

rockstardev:
Resolves: #1959

WIP

Illustration of new warnings:

image

Each store that doesn't have wallet setup / hint is not dismissed now has warning icon:

image

Hint also displayed on Rates page:

image

I assume the biggest discussion will be on where we are persisting these hints. I've done first version with saving hints in StoreBlob. But could see this ending on client side... or maybe in UserPrefsCookie. Will see after additional discussion.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -228,7 +228,11 @@ namespace BTCPayServer.Tests
             Driver.FindElement(By.Id("Email")).SendKeys(user);
             Driver.FindElement(By.Id("Password")).SendKeys(password);
             Driver.FindElement(By.Id("LoginButton")).Click();
+        }
 
+        public void GoToStores()
+        {
+            Driver.FindElement(By.Id("Stores")).Click();
         }
 
         public void GoToStore(string storeId, StoreNavPages storeNavPage = StoreNavPages.Index)
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -42,18 +42,19 @@ namespace BTCPayServer.Tests
         {
             await Server.StartAsync();
             ChromeOptions options = new ChromeOptions();
-            var isDebug = !Server.PayTester.InContainer;
+            if (Server.PayTester.InContainer)
+            {
+                // this must be first option https://stackoverflow.com/questions/53073411/selenium-webdriverexceptionchrome-failed-to-start-crashed-as-google-chrome-is#comment102570662_53073789
+                options.AddArgument("no-sandbox");
+            }
 
+            var isDebug = !Server.PayTester.InContainer;
             if (!isDebug)
             {
                 options.AddArguments("headless"); // Comment to view browser
                 options.AddArguments("window-size=1200x1000"); // Comment to view browser
             }
             options.AddArgument("shm-size=2g");
-            if (Server.PayTester.InContainer)
-            {
-                options.AddArgument("no-sandbox");
-            }
             Driver = new ChromeDriver(Server.PayTester.InContainer ? "/usr/bin" : Directory.GetCurrentDirectory(), options);
             if (isDebug)
             {
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -303,7 +303,7 @@ namespace BTCPayServer.Tests
         public string CreateInvoice(string storeName, decimal amount = 100, string currency = "USD", string refundEmail = "")
         {
             GoToInvoices();
-            Driver.FindElement(By.Id("CreateNewInvoice")).Click();
+            Driver.FindElement(By.Id("CreateNewInvoice")).Click(); // ocassionally gets stuck for some reason, tried force click and wait for element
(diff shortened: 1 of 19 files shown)


dstrukt:
Hell yeah!

Few thoughts: 

1. Personally, I don't think all of the explanation text underneath Wallet or Lightning is necessary. I think a clear tooltip explanation as to why you need to set up your derivation scheme to accept funds is all we need, until we update this section with a simple "Setup Wallet" CTA. 

If we really wanted to include more information it should be in the form of an expandable tooltip (or something along those lines) to help keep the focus clear. Ideally everything we're building is clear enough to navigate without a bunch of explanation, and we educate as we step through the process in bite-sized contextual pieces.

2. The wording for the wallet tooltips could a little clearer.

"A store requires a wallet to receive funds. Please provide a derivation scheme to set up your wallet."
&
"A connection to a lightning node is required to generate invoices and receive payments."

Perhaps?

3. As for the persistent, until complete, tooltip in the table rows, I imagine it would also be clickable and take you directly to the tab where you set up the wallet?

4. I still prefer the 

- General
- Wallet
- Lightning

list order.

I'm curious to see how this UX pattern feels when launched. I do like the simplicity of it... However, I still feel the CTA could be clearer and presented at the top of the page with a direct action vs. a series of warnings / hints. 

Maybe I'm wrong about this...shrug

Some visuals to correspond this long winded rant .. eventually the derivation scheme table would simplified with a singular CTA or a completed simplified component ... realizing I want to explore that..

🍻 

Screen Shot 2020-10-16 at 12 38 13 AM
Screen Shot 2020-10-16 at 12 37 27 AM

rockstardev:
OK, we addressed all this feedback during Design call and it was decided to proceed with merging this as foundation, and then going for more changes after.

@NicolasDorier the biggest thing in here is you giving ACK to having Hints as part of Store Blob - see how I've implemented it. I do think this way should stick long term - this will give user ability to dismiss hints in case they don't want the hint sticking (for whatever reason). Different than driving hints from say `DerivationScheme` collection (where you don't have ability to permanently dismiss).

Alternative storage is to persist this in cookies, downside being that it won't work across browsers... it's really part of store. 
Also another consideration in case of cookies is that we would need to iterate cookie storage when displaying store listing.

However, no critical downside in any case... so checkout branch and let us know your feedback.

BTCPayServer/Models/StoreViewModels/StoreViewModel.cs
```diff
@@ -90,6 +91,9 @@ public NetworkFeeMode NetworkFeeMode
         [Display(Name = "Enable Payjoin/P2EP")]
         public bool PayJoinEnabled { get; set; }
 
+        public bool HintWallet { get; set; }
+        public bool HintLightning { get; set; }
```
NicolasDorier, line 95: I think just exposing `StoreHints` instead would prevent from maintaining the mapping. That said, not really important.

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -182,6 +182,13 @@ public async Task CreateStore(string ownerId, StoreData storeData)
         public async Task<StoreData> CreateStore(string ownerId, string name)
         {
             var store = new StoreData() { StoreName = name };
+            var blob = store.GetStoreBlob();
+            blob.Hints = new Data.StoreBlob.StoreHints
+            {
+                Wallet = true,
+                Lightning = true
+            };
```
NicolasDorier, line 190: would be better in ctor I would say, bu thats ok

NicolasDorier:
seems great, though I think you should highjack one of the selenium test to make sure the warning show up when the store is just created.

BTCPayServer/Models/StoreViewModels/StoreViewModel.cs
rockstardev: I did it this way because view model wraps other properties from blob... this will work nicely if we add more hints to store and we don't want them on general page / in this view model.

BTCPayServer/Services/Stores/StoreRepository.cs
rockstardev: I wanted to leave default constructor empty because we need it from json deserialization and sometimes we want these props set to false, sometimes to true... found it easier like this.

rockstardev:
OK, since @NicolasDorier signed off on this approach, I've tweaked the rest of the code - so hints are dismissed when wallet / lightning connection are set and added additional checks to Selenium test.

Merging it later today.

BTCPayServer.Tests/SeleniumTester.cs
```diff
@@ -42,18 +42,19 @@ public async Task StartAsync()
         {
             await Server.StartAsync();
             ChromeOptions options = new ChromeOptions();
-            var isDebug = !Server.PayTester.InContainer;
+            if (Server.PayTester.InContainer)
+            {
+                // this must be first option https://stackoverflow.com/questions/53073411/selenium-webdriverexceptionchrome-failed-to-start-crashed-as-google-chrome-is#comment102570662_53073789
+                options.AddArgument("no-sandbox");
```
NicolasDorier, line 48: If this does not break the build, it is ok, but I think this mean some test might fail because chrome does not start test in a new profile.

NicolasDorier (approved):

pavlenex:
Hell yeaaaaah! 🚀 🚀 🚀

PR #1991, opened 2020-10-16
Allow disabling notifications per user and disabling specific notifications per user

### LINKED ISSUE #1974: Disable notifications for a store or user
The particular store has about 200 active users (admins and guests). With a large number of new invoices a very large number of notifications are sent. Which, in turn, leads to the fact that the amount of data in the database increases (users do not delete notifications). Also increases the load on the web sockets.

Introduce the ability to enable/disable notifications for a specific store or a specific user.

Add a filter for notifications of invoice events for a specific store or a specific user.

Kukks:
closes #1974

Someone can prettify this UI afterwards. The reason that you disable instead of enable specific notifications is because if we add new notifications, users would not get them by default.

image


pavlenex:
Nice. Concept ACK, but the UI needs some checkboxes love, (like we have for policies, notifications, etc) if @dennisreimann is willing to halp and improve it ✅

NicolasDorier:
I would also really prefer checkboxes.

britttttk:
@Kukks what is Junk in this list for?

Kukks:
> @Kukks what is Junk in this list for?

It's a way to test notifications when in Debug mode

britttttk:
Will the checkboxes fix this or it's a bug? I'm not sure
image

Kukks:
> Will the checkboxes fix this or it's a bug? I'm not sure
> image

That was a bug, fixed now. Thanks!

Kukks:
image


here you go

britttttk:
tACK - They are all working now (note: I didn't test version disable since it already existed in admin policy).

pavlenex:
@Kukks Can `Junk` be renamed? It's not really clear what it does, renaming it to something more sensible would really be helpful.
As Brit mentioned, since we do have "new update" in Server Settings > Policies, I don't think we really need them for user-settings?

Kukks:
It doesn't matter, it doesn't get included in releases at all.

Kukks:
> As Brit mentioned, since we do have "new update" in Server Settings > Policies, I don't think we really need them for user-settings?

The two things are unrelated tech wise. One issues version notifications, the other is a user rejecting those types of notifications.

It makes sense to have both as you can have multiple admins where >=1 admins would still want to know about it

dennisreimann (commented):
Would it be an option to extract this inline script to a separate file, which could be included in the main bundle?

BTCPayServer/Views/Manage/NotificationSettings.cshtml
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+    }
+    else
+    {
+        var items = new SelectList(NotificationHandlers.SelectMany(handler => handler.Meta.Select(tuple => new SelectListItem(tuple.name, tuple.identifier))), nameof(SelectListItem.Value), nameof(SelectListItem.Text));
+        
+        <div class="form-group">
+            <label asp-for="DisabledNotifications">Do not receive notifications for</label>
+
+            <select size="@items.Count()" class="form-control hide-when-js" multiple asp-for="DisabledNotifications"
```
dennisreimann, line 24: Tdchnically/HTML-wise: Why is this a multiselect instead of regular checkboxes?

Kukks:
> Would it be an option to extract this inline script to a separate file, which could be included in the main bundle?

Yes, but I would wait until we need to reuse it elsewhere

BTCPayServer/Views/Manage/NotificationSettings.cshtml
Kukks: because the data model is a string array vs an object with `id, selected` props

dennisreimann: Ok, but we could transform it for the view, right? 
I'd find this to be cleaner than having different approaches for JS/Non-JS, plus we'd get rid of the extra bootstrap-multiselect.js

pavlenex:
We can re-use this in Server Settings > Policies perhaps?

dennisreimann:
> Yes, but I would wait until we need to reuse it elsewhere

My concern is more to save those extra bytes per each request where the snippet is included. Currently it's part of each request instead of once in the bundle. Yes, it's not much, but we could optimize it.

Kukks:
> > Yes, but I would wait until we need to reuse it elsewhere
> 
> My concern is more to save those extra bytes per each request where the snippet is included. Currently it's part of each request instead of once in the bundle. Yes, it's not much, but we could optimize it.

Removed the js dependency, it was definitely a lazy move from my end when i added it in

BTCPayServer/Views/Manage/NotificationSettings.cshtml
Kukks: done

dennisreimann (approved):

PR #2000, opened 2020-10-20
Override Block Explorer Links

### LINKED ISSUE #1953: Allow users to select block explorers from a list or specify their own URL
It would be nice to have a list of popular block explorers to choose from, as well as a user provided custom field, instead of simply hard-coding blockstream.info

Here is the list that Bisq uses: https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/user/Preferences.java#L81

Kukks:
closes #1953
This one adds it to policies page with no extension stuff..

### DIFF
--- a/BTCPayServer.Common/BTCPayNetwork.cs
+++ b/BTCPayServer.Common/BTCPayNetwork.cs
@@ -130,9 +130,25 @@ namespace BTCPayServer
 
     public abstract class BTCPayNetworkBase
     {
+        private string _blockExplorerLink;
         public bool ShowSyncSummary { get; set; } = true;
         public string CryptoCode { get; internal set; }
-        public string BlockExplorerLink { get; internal set; }
+
+        public string BlockExplorerLink
+        {
+            get => _blockExplorerLink;
+            set
+            {
+                if (string.IsNullOrEmpty(BlockExplorerLinkDefault))
+                {
+                    BlockExplorerLinkDefault = value;
+                }
+
+                _blockExplorerLink = value;
+            }
+        }
+
+        public string BlockExplorerLinkDefault { get; internal set; }
         public string DisplayName { get; set; }
         public int Divisibility { get; set; } = 8;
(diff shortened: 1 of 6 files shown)


pavlenex:
@Kukks is there a reason why this wouldn't work in dev enviroment? I just want to be certain it's not because of the testnet. 
Here's what I did:
1. I added blockchain.com in the field
2. Saved
3. Went to an invoice clicked on TXID link and it defaulted to blockstream.info
4. I then headed back thinking I did something wrong and clicked on the default button but it didn't return anything to default.

blockexplorerPNG

Was having a default option (which I think is really useful but didn't work for me,) the reason why you opted for this type of form. We don't really use this type of form elsewhere.

Kukks:
> @Kukks is there a reason why this wouldn't work in dev enviroment? I just want to be certain it's not because of the testnet.
> Here's what I did:
> 
> 1. I added blockchain.com in the field
> 2. Saved
> 3. Went to an invoice clicked on TXID link and it defaulted to blockstream.info
> 4. I then headed back thinking I did something wrong and clicked on the default button but it didn't return anything to default.
> 
> blockexplorerPNG
> 
> Here's a console if it means anything.
> console
> 
> Was having a **revert to default option** (which I think is really useful but didn't work for me,) the reason why you opted for this type of form? We don't really use this type of form elsewhere, I don't really mind it, it's just different so I'm curious.

Can you try it again now? It was only loading them after a restart before, now it will load them after save too.


This form is more or less the same form used in the new notifications section, the domain mapping, etc
The default option is important because once you override, you'll never be able to see what the original value was.

pavlenex:
@Kukks Okay, switching links works now, but the default button still won't return stuff to default.

Kukks:
> @Kukks Okay, switching links works now, but the default button still won't return stuff to default.

fixed

pavlenex:
tACK, works now! 🚀

pavlenex (approved):

PR #2001, opened 2020-10-20
BTCPay Extensions Part 2

Kukks:
This PR cleans up the extension system a bit in that:
 * It renames the test extension to a more uniform name
 * Allows yo uto have system extensions, which are extensions but bundled by default with the release (and cannot be removed)
 * Adds a tool to help you generate an extension package from a csproj
 * Refactors the UI extension points to a view component
 * Moves some more interfaces to the Abstractions csproj


dennisreimann:
Just checked, install + uninstall work fine. I'll update the UI a bit once this gets merged – looking forward to this!

PR #2011, opened 2020-10-23
Store setup restructuring

dennisreimann:
Continuation of #1990 to go towards [what @dstrukt proposed](https://chat.btcpayserver.org/btcpayserver/pl/zxorhqasxidtpc8by45iyxt8ch) recently.

## Idea

We want to trim down the information on the general store settings page to the most important things and put everything else on the setup/modification pages of the individual sections (e.g. wallet and lightning). Screenshot of the idea: 

Screen Shot 2020-10-20 at 12 42 15 AM

## Basics

The first commit cleans up the markup and includes some wording improvements. For instance we name the actions "Setup" instead of "Modify" in case the function isn't set up yet.

Also moves the LN invoice description template out of the way – ideally we can move this to the LN setup/modify page.

Posting this as a draft to invite feedback and thoughts about what might not work about this. @rockstardev already mentioned in our dev call that it might be difficult to separate/split up things, so here we go … 

Screenshot of the basic cleanup: 

store-setup-1

## Wallets display

The second commit alters the wallet display to fit the proposed style … I think having the logos would be a great addition here.

disabled

lightning-available

enabled-dark

enabled-light


dennisreimann:
Rebased and improved the display.

light

dark

dennisreimann:
I think this is good to go – @dstrukt mentioned he wanted to update the "warnings", which curreently look like this:

dark

light

pavlenex:
Tested this a bit, and it looks like a really good improvement. I'm still thinking about two things:

1. Do we need to make `Setup` CTA a bit clearer, or is it clear enough?
2. I'm not a fan on the text in the wallet on-chain section, mostly because I don't think majority of people knows what on-chain payment is, where as if they've heard about the LN they will use that. My point is, let's try to keep things simple in the `on-chain` section. `A store requires a wallet to receive payments. Set up your wallet` I just feel there's too much text that people won't understand, those who need LN will most likely understand it without explanation above imo. 

Thoughts?

dstrukt:
1. It's moving in the right direction, and I agree it could be pushed a little bit further, and will drop some more explorations on the Mattermost later today for discussion to wrap this task up. Completely agree on keeping the explanation text simple as well!

2. I always for simplifying the interface, so I would be happy removing the on/off-chain text, but I can understand the case to keep it as well... but I don't think it's entirely necessary.

rockstardev:
Just to chime in and say that UI @dennisreimann did looks great. 

Just please don't make it too simple folks... we have smarties in Bitcoin community... if something verbose fits better - so be it.

dennisreimann:
Rebased again – let's get this thing merged soon :)

Shortened the explanations, improved the setup CTA and modify button.

## Dark

warn-dark

#### Setup / Modify

dark

## Light

warn-light

#### Setup / Modify

light

NicolasDorier:
Now he complains I don't merge fast enough!

dstrukt:
Shared on MM, but reposting for documentation's sake.

Per the latest iteration:

I've been considering a few options to keep our new more minimal direction, but still keep things consistent. My favorite iteration is the third variation in the first screenshot.

From a UX perspective, we should try to remain as consistent as possible with CTA colors and cues.

e.g. A solid green button or green text means "you can act or click upon this, reinforcing action and consistency in the product / brand.

In the example below with the larger green button (primary), we're pushing people to set-up, but once they know how, we can settle back on the tertiary (text only) green button for the "modify" function as the user now understands this interaction. I tried the secondary (green outline, green text), but I liked the first/tertiary combinations best (posted for example). This type of consistency and clean up is something I've been working towards across the whole product in the updated mocks.

As for the text, I'm not sure we need dismissible text if it's so short and concise, it could remain even after the wallet has been set up (or it could disappear if we really feel it's unnecessary), but I think it's text that should be permanent, at lest until set up is complete for the respective method. As for the warning icon .. I've always felt we need something a little more subtle, or different, to indicate it's a next step, not a warning, but I think with it on top, a clear and concise set of instructions and a nice bold CTA... and the other efforts (via eventual dashboard, etc..) to push people to this screen, I think we're probably good without it... the icon makes more sense on the store listing page.

The blue vs. grey (or black is an option) color, I'm a little less opinionated on and could go either way, the blue does place a bit more emphasis visually on the input value, without being too distracting.

I think we could also removed the divider between the currency and the public key / address, but keep it everywhere else, might help save that extra bit of space...shrug.

--

I also tried some other experiments that I don't care for as much, but for the sake of sharing exploration.

As always, open to any and all feedback!

Screen Shot 2020-11-27 at 12 27 02 PM

Screen Shot 2020-11-27 at 2 29 43 PM

Screen Shot 2020-11-27 at 12 27 02 PM

Screen Shot 2020-11-27 at 12 27 06 PM

dennisreimann:
> My favorite iteration is the third variation in the first screenshot.

Second that – it's lightweight and intuitive, I really like it.

PR #2021, opened 2020-10-26
Remove QuadrigaCX references

astupidmoose:
Quadriga has been dead for years. I've removed the references to them and replaced them with a different Canadian exchange (ndax) for example. 

The whole set of instructions could probably use an overhaul, but for now, at least let's get rid of Gerry's presence.


Kukks (approved):

bolatovumar (approved):

Kukks:
Just pushed a commit on your branch to also replace it in our tests

PR #2024, opened 2020-10-28
Add warning to configure e-mail server

### LINKED ISSUE #1989: Requires a confirmation mail for registering should not be enabled unless an SMTP is configured
**Describe the bug**
Currently a server that doesn't have SMTP configured can enable `Requires a confirmation mail for registering`. This can lead user into thinking that they can send out confirmation emails which is not the case.

**To Reproduce**
1. Go to Server Settings > Policies
2. Check on `Requires a confirmation mail for registering` and `Save`
3. Email registration won't be sent

**Expected behavior**
We should provide a warning. When user enables this show a warning `Your email server has not been configured. Please configure it first`.

**Screenshots**

- BTCPay Server Version [available in the right bottom corner of footer]  v1.0.5.7
- Deployment Method: [e.g. Docker, Manual, Third-Party-hoist] Docker

bolatovumar:
Adds a warning to configure the e-mail server before "Requires a confirmation mail for registering" checkbox can be checked if e-mail server is not configured.

Screen Shot 2020-10-27 at 9 34 36 PM

close #1989

### DIFF
--- a/BTCPayServer/Views/Server/Policies.cshtml
+++ b/BTCPayServer/Views/Server/Policies.cshtml
@@ -1,5 +1,7 @@
 @using BTCPayServer.Services
+@using BTCPayServer.Services.Mails;
 @model BTCPayServer.Services.PoliciesSettings
+@inject BTCPayServer.Services.SettingsRepository _SettingsRepository
 @{
     ViewData.SetActivePageAndTitle(ServerNavPages.Policies);
 }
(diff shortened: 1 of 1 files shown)


BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -150,6 +150,7 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
             services.TryAddSingleton<AppService>();
             services.AddSingleton<PluginService>();
             services.TryAddTransient<Safe>();
+            services.TryAddTransient<EmailSettings>();
```
Kukks, line 153: No, you should be using the Settings Repo to fetch the email settings.

BTCPayServer/Views/Server/Policies.cshtml
```diff
@@ -1,5 +1,6 @@
 @using BTCPayServer.Services
 @model BTCPayServer.Services.PoliciesSettings
+@inject BTCPayServer.Services.Mails.EmailSettings EmailSettings
```
Kukks, line 3: You should inject SettingsRepository, and then use something like `await _SettingsRepository.GetSettingAsync<EmailSettings>()` to get the settings.

Kukks (changes requested):

BTCPayServer/Views/Server/Policies.cshtml
bolatovumar: Thanks. Will take a look.

BTCPayServer/Hosting/BTCPayServerServices.cs
bolatovumar: Removed.

BTCPayServer/Views/Server/Policies.cshtml
bolatovumar: Done.

BTCPayServer/Views/Server/Policies.cshtml
```diff
@@ -14,9 +16,18 @@
 <form method="post">
     <div class="form-group">
         <div class="form-check">
-            <input asp-for="RequiresConfirmedEmail" type="checkbox" class="form-check-input"/>
+            @{
+                var emailSettings = (await _SettingsRepository.GetSettingAsync<EmailSettings>()) ?? new EmailSettings();
+                var isEmailNotConfigured = emailSettings.IsComplete();
+            }
+            <input asp-for="RequiresConfirmedEmail" type="checkbox" class="form-check-input" disabled="@isEmailNotConfigured" />
```
Kukks, line 23: You should enable it if  `emailSettings.IsComplete() || Model.RequiresConfirmedEmail` so that if someone enabled it before this is released, they can disable

Kukks (changes requested):

BTCPayServer/Views/Server/Policies.cshtml
bolatovumar: Right. Updated.

Kukks (approved):

PR #2025, opened 2020-10-28
GreenField Webhooks

Kukks:
Introduces a more versatile webhooks system, indepedent of the legacy BitPay IPN compatibility. 

The new webhooks system is agnostic of invoices, in that we can now create events for any type of action within BTCPay Server. 

There is also a verification system, that utilizes the amemessage signing system from Bitcoin Core. A private key is stored inside BTCPay (multiple ones depending on groups of events, for example, there is 1 private key for store related events) that is used to sign the event payload and the signature provided along with the event. 

Events can be queued up in specific groups, so that events are sent sequentially in those groups. For example, having a webhook event `X` to url `Y` will queue any of those events under `X_Y` and sent after each other. Other events are sent in parallel.

We use the same retry mechanism that we have for the Bitpay IPN system. BTCPay retries up to 6 times, with 10 minutes intervals between each try and a failure is defined as either a non successful HTTP status code or a 1 minute timeout.

This PR introduces only one event: Invoice status changes but new events can very easily be added.

There is  support for webhooks over Tor if the webhook url is an onion url.

In Store-> Integrations, there is a UI to add webhooks on a store level to receive events on all invoices created on that store.  

image

image

image


BTCPayServer.Client/Events/BaseEvent.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class GreenFieldEvent<T>
+    {
+        public string EventType { get; set; }
+        public T Payload { get; set; }
+        public string Signature { get; set; }
+        
+        public void SetSignature(string url, Key key)
+        {
+            Signature = key.SignMessage($"{url}_{Payload.GetHashCode()}");
```
Kukks, line 16: I need to get rid of `GetHashCode` and replace with a hash of the payload in json

NicolasDorier:
I like this approach, as we talked about: The authentication of callback can be check of key + signature, which seems to be the best practice (in github included).

We need a way to expose the public key of the server though.

NicolasDorier:
The way to check the signature and create it should not depends on any "normalization" of the json. This is difficult to get right, and will end up with nobody checking the signatures.

Instead the hash should be calculated on the payload bytes of the HTTP request, in the same way we do with bitpay's API request authentication.

Kukks:
> The way to check the signature and create it should not depends on any "normalization" of the json. This is difficult to get right, and will end up with nobody checking the signatures.
> 
> Instead the hash should be calculated on the payload bytes of the HTTP request, in the same way we do with bitpay's API request authentication.
> 
> We can reuse their BitId stuff.

Let's not. This is ugly and non-intuitive for any web dev. A simple normalized string of url + body is the best way and is what others do.

Kukks:
> I like this approach, as we talked about: The authentication of callback can be check of key + signature, which seems to be the best practice (in github included).
> 
> We need a way to expose the public key of the server though.

The key for the store is already exposed in the Integrations page and through the Greenfield Stores GET

Kukks:
Updated redoc and now we can even document webhooks through it!
image

NicolasDorier:
> A simple normalized string of url + body is the best way and is what others do.

What is the difference with BitId?

Kukks:
> > A simple normalized string of url + body is the best way and is what others do.
> 
> What is the difference with BitId?

I updated the code to do similar to BitID. It now only uses the SHA256 hash instead of the full Bitcoin Core signing compat.
Docs in swagger about it:

NicolasDorier:
Concept ACK, will review tomorrow.

dennisreimann:
I think we can clean up the UI a bit, similar to what I did with #1993. But I can take care of that after this once gets merged :)

NicolasDorier:
So for the UX, I checked what github does and I think we should follow their example.

image
image

Their authentication is using HMAC encoded with a shared secret in the webhook edit form, we should do the same.

https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/securing-your-webhooks

Notice their signature verification and signing is done at the edge, on the actual body bytes when coming in and before going out. The way you are checking now does not do this and needlessly couple newtonsoft json into the calculation of the hash of the message.

NicolasDorier:
The new conflicts are related to my PR above. This is basically your idea ported on a separate PR so we can simplify the webhooks PRs.

Kukks:
Sigh

PR #2028, opened 2020-10-29
Fix link from app label

### LINKED ISSUE #1997: Automatic app label gives 404
**Describe the bug**
When you click on an `i` tooltip for the automatic labels added in #1916 for the `app` it gives 404, because the URL goes to 
apps/xyz instead apps/xyz/pos for example, so app type is not included which leads to 404.

**To Reproduce**
1. Receive a payment through the pos app
2. Go to Wallets and click on the `i` for the app
4. See 404 page

**Expected behavior**
It should open a link from the app.

**Screenshots**
image

- BTCPay Server Version [available in the right bottom corner of footer]  1.0.5.8
- Deployment Method: [e.g. Docker, Manual, Third-Party-hoist]: Docker

**Logs (if applicable)**
Basic logs can be found in Server Settings > Logs. More logs https://docs.btcpayserver.org/Troubleshooting/#2-looking-through-the-logs

**Additional context**
I'm not sure what's expected behavior here. The `x` (remove the label) doesn't work (if the expected behavior is to not remove this then we shouldn't have the `x`.

Kukks:
fixes #1997

### DIFF
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -41,12 +41,13 @@ namespace BTCPayServer.Controllers
         [HttpGet("/apps/{appId}")]
         public async Task<IActionResult> RedirectToApp(string appId)
         {
-            switch (await _AppService.GetAppInfo(appId))
+           
+            switch ((await _AppService.GetApp(appId, null)).AppType)
             {
-                case ViewCrowdfundViewModel  _:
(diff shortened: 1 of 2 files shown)


pavlenex:
#1997 ?

pavlenex (approved):
tACK, though tests fail for some reason.

Kukks:
> tACK, though tests fail for some reason.

nothing to do with this PR

PR #2029, opened 2020-10-29
Remove only dependency on Dbriize (TextSearch in new invoice column)

Kukks:


NicolasDorier:
So I took time to read the doc of postgres.

Doing ` query = query.Where(i => i.TextSearch.Contains(queryObject.TextSearch));` is out of question.

It is a full table scan, will take a while for old store, longer and longer.
What we can do though, on postgres, is to have a `tsvector` column, and just putting there all the words to index.
Then put an index on this column, and query via a `tsquery`. This will be fast and do what is intended.

Downside is that it works only on postgres.
For other DB maybe we can just keep dbreeze?

rockstardev (commented):
@Kukks was kind enough to help me review this PR over screenshare - and the alternative proposal we came up with was to create `InvoicesIndex` table in database that would have two columns:

1. InvoiceId
2. IndexValue

You would then create a row for each value currently saved in the text file:
 - InvoiceTime.ToString(CultureInfo.InvariantCulture));
 - invoice.Price.ToString(CultureInfo.InvariantCulture));
 - invoice.Metadata.OrderId
 - ToString(invoice.Metadata, null)
 - invoice.StoreId
 - textSearch.Add(invoice.Metadata.BuyerEmail);

This way searches for Invoice would be simple SQL and work across all databases; allowing us to fully dump Breeze. As @Kukks - I'm against having different implementations depending on DB - if we're dumping Breeze we should fully dump it.

Also this format would be easy to extend in future... and have easy dump/rebuild index functionality.

NicolasDorier:
I think that's good idea to do it like that. If there is an index on the column, should be fine.
so `InvoiceId` would be foreign key with cascade delete, index on `IndexValue`.
For the Primary key, would be nice to have to have auto increment int, but I am not sure it is possible. As the key would take more space than the actual indexed data.

Or maybe is it possible to use `(IndexValue, InvoiceId)` as primary key? if that is possible, then we don't even have to put an index on index value.

NicolasDorier:
If I remember @rockstardev told me you both talked about having a table `(IndexValue, InvoiceId)`.
This would work.

Kukks:
Changed to a table but unsure if it's done correctly

BTCPayServer.Data/Data/InvoiceSearchData.cs
```diff
@@ -0,0 +1,28 @@
+using System.ComponentModel.DataAnnotations.Schema;
+using Microsoft.EntityFrameworkCore;
+
+namespace BTCPayServer.Data
+{
+    public class InvoiceSearchData
+    {
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public string Id { get; set; }
```
rockstardev, line 9: This should be `int` or `long`?

BTCPayServer.Data/Migrations/20201208150236_InvoiceSearchTable.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            int? maxLength = this.IsMySql(migrationBuilder.ActiveProvider) ? (int?)255 : null;
+            migrationBuilder.CreateTable(
+                name: "InvoiceSearchDatas",
+                columns: table => new
+                {
+                    Id = table.Column<string>(nullable: false, maxLength:maxLength),
+                    InvoiceDataId = table.Column<string>(nullable: true, maxLength:maxLength),
+                    Value = table.Column<string>(nullable: true)
```
rockstardev, line 20: I assume `maxLength` should only go on `InvoiceDataId`?

BTCPayServer.Data/Data/InvoiceSearchData.cs
```diff
@@ -0,0 +1,28 @@
...
+        internal static void OnModelCreating(ModelBuilder builder)
+        {
+            
+            builder.Entity<InvoiceSearchData>()
+                .HasOne(o => o.InvoiceData)
+                .WithMany(a => a.InvoiceSearchData)
+                .OnDelete(DeleteBehavior.Cascade);
+
+            builder.Entity<InvoiceSearchData>().HasIndex(data => new {data.InvoiceDataId, data.Value});
+            builder.Entity<InvoiceSearchData>().HasIndex(data => data.Value);
```
rockstardev, line 25: @NicolasDorier: you need to make subjective decision here of whether we want `Id` on this object. Tradeoff:

Removing `Id` we could switch to having less data and composite key:
```
builder.Entity<InvoiceSearchData>().HasKey(data => new
{
    data.InvoiceDataId,
    data.Value
});
```

We already do `Distinct()` calls on adding new entries in this table:
 - https://github.com/btcpayserver/btcpayserver/pull/2029/files#diff-f761bdfbdb17f885259fd38c8d601bec65e40d59a8278dbb73cab94804668c56R278
 - https://github.com/btcpayserver/btcpayserver/pull/2029/files#diff-c545b4806714f17ef9b042626a9db7614f96ba9abac22993a00f7ebc7d6398f9R363
 
 The problem would be for updates... where you would need to check presence of unique pair in database.
 
 So if we are deciding to keep `Id` - I would just make it `int` (or `long`) - that way it'll use less space.

BTCPayServer.Data/Migrations/20201208150236_InvoiceSearchTable.cs
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/2029/files#r546401885

BTCPayServer.Data/Data/InvoiceSearchData.cs
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/2029/files#r546401885

BTCPayServer.Data/Data/InvoiceSearchData.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        internal static void OnModelCreating(ModelBuilder builder)
+        {
+            
+            builder.Entity<InvoiceSearchData>()
+                .HasOne(o => o.InvoiceData)
+                .WithMany(a => a.InvoiceSearchData)
+                .OnDelete(DeleteBehavior.Cascade);
+
+            builder.Entity<InvoiceSearchData>().HasIndex(data => new {data.InvoiceDataId, data.Value});
```
rockstardev, line 24: @NicolasDorier: Decide if you want this additional index... or if usage will be mostly to search `WHERE Value LIKE '%lala%'` then we don't need this composite index.

rockstardev (commented):
Requires subjective decisions on two open comments.

@NicolasDorier when ready: make decisions and merge it.

EDIT: Changing to `Request Changes` since DBriizeEngine output is garbage; we get ~650 entries per invoice that are useless. 

I've consulted with @Kukks during the day on potentially switching to re-indexing from `InvoiceData` table... and he's OK with that approach as well. So I guess that's third subjective decision - will influence the other two.

rockstardev (changes requested):
Updated original comment

BTCPayServer.Data/Data/InvoiceSearchData.cs
NicolasDorier: we don't need this one.
Though I am unsure if `WHERE InoivceDataId = && Value =` would use the other index. I know there is weird rules in SQL where even the order of clauses in WHERE impact whether or not an index is use. So need to test it.

NicolasDorier:
Reindexing on the `InvoiceData` table is indeed interesting. That way we can just completely drop DBRiize.
Don't need to reindex the whole thing, just OrderId, addresses, InvoiceId and emails...

BTCPayServer.Data/Data/InvoiceSearchData.cs
rockstardev: I've tested this extensively... `WHERE InoivceDataId = && Value =` would work with Postgresql with just that `data.Value` index.

BTCPayServer.Data/Data/InvoiceSearchData.cs
rockstardev: Decided to go with this data structure: https://github.com/btcpayserver/btcpayserver/pull/2029/commits/2d96a1b0aaf05b939c0a8e4579737950fc3c0282#diff-45444102c608c30496ffadec8c0a632b65bc271671e6c58dab1f91ce3e49f03c

rockstardev:
Just moving of data migration to HostedService left.

Kukks:
> Just moving of data migration to HostedService left.

I think it should be in the migration startup task to ensure its done before btcpay starts? 
Unless you intend it to index in the background?

rockstardev:
@Kukks: I understood @NicolasDorier wants this done this indexing in the background so big instances don't hang on the startup.

I plan on doing batches of 1000 invoices and saving results... so if indexing stops in the middle it can recover. That foreign key gave us index on InvoiceDataId so it shouldn't be a big problem to select by "not exists" if needed.

Kukks:
> @Kukks: I understood @NicolasDorier wants this done this indexing in the background so big instances don't hang on the startup.
> 
> I plan on doing batches of 1000 invoices and saving results... so if indexing stops in the middle it can recover. That foreign key gave us index on InvoiceDataId so it shouldn't be a big problem to select by "not exists" if needed. Or I may just save "last invoiceid indexed" somewhere and go by that.

Sounds good. Will you use a Transaction to guarantee the batch successfully? And will you store the progress in the MigrationSetttings through the settings repo?

rockstardev:
@Kukks 

image

See `DbMigrationsHostedService` and let me know if you think something is missing in data migration.

@NicolasDorier ready for final review.

BTCPayServer/HostedServices/DbMigrationsHostedService.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                await MigratedInvoiceTextSearchToDb(settings.MigratedInvoiceTextSearchPages.Value);
+            }
+
+            // Refresh settings since these operations may run for very long time
+        }
+
+        private async Task MigratedInvoiceTextSearchToDb(int startFromPage)
+        {
+            using var ctx = _dbContextFactory.CreateContext();
```
Kukks, line 46: You should use a transaction to ensure all the data is saved without issue
```suggestion
            using var ctx = _dbContextFactory.CreateContext();
            using var transaction = ctx.Database.BeginTransaction();
```

BTCPayServer/HostedServices/DbMigrationsHostedService.cs
```diff
@@ -0,0 +1,107 @@
...
+                else
+                {
+                    // during final pass we set int.MaxValue so migration doesn't run again
+                    settings.MigratedInvoiceTextSearchPages = int.MaxValue;
+                }
+
+                // this call triggers update; we're sure that MigrationSettings is already initialized in db 
+                // because of logic executed in MigrationStartupTask.cs
+                _settingsRepository.UpdateSettingInContext(ctx, settings);
+                await ctx.SaveChangesAsync();
```
Kukks, line 103:
```suggestion
                await ctx.SaveChangesAsync();
                transaction.Commit();

```

Kukks (commented):
@rockstardev  added some comments

BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -259,6 +261,32 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.ToTable("InvoiceEvents");
                 });
 
+            modelBuilder.Entity("BTCPayServer.Data.InvoiceSearchData", b =>
+                {
+                    b.Property<int>("Id")
+                        .ValueGeneratedOnAdd()
```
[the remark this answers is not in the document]
rockstardev: Nice catch Kukkie, I've dealt with it in: 0f7869c23a8d5c9b928d0ae4181e6eb070b3533b

Tested with generating new migration, we are good to go now:  

https://user-images.githubusercontent.com/5191402/103175916-692cae80-4833-11eb-90ed-cb9ada700878.mp4

BTCPayServer/HostedServices/DbMigrationsHostedService.cs
rockstardev: From what I understand `SaveChangesAsync` already creates transaction. 

And what I did with `UpdateSettingInContext` is to ensure settings update doesn't trigger it's own `SaveChangesAsync`, but rather it's all unified in this one call on line 103.

BTCPayServer/HostedServices/DbMigrationsHostedService.cs
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/2029#discussion_r549137609

Kukks:
Good from my end
@NicolasDorier  MERGE MERGE MERGE

NicolasDorier:
checking now

BTCPayServer/HostedServices/DbMigrationsHostedService.cs
```diff
@@ -0,0 +1,107 @@
...
+            var totalPages = Math.Ceiling(totalCount * 1.0m / PAGE_SIZE);
+            for (int i = startFromPage; i < totalPages; i++)
+            {
+                invoiceQuery.Skip = i * PAGE_SIZE;
+                invoiceQuery.Take = PAGE_SIZE;
+                var invoices = await _invoiceRepository.GetInvoices(invoiceQuery);
+
+                foreach (var invoice in invoices)
+                {
+                    var textSearch = new List<string>();
```
NicolasDorier, line 60: Can you verify the CancellationToken here so that people can restart the servers and it does not hang here until migration is done?

NicolasDorier: Did it

PR #2030, opened 2020-10-29
Batch saving Invoice Log events every 5 seconds

Kukks:


NicolasDorier:
NACK, there is no reason we log so much that we need batching for inserting them.

PR #2038, opened 2020-11-02
Add warning for zero-conf accept option

### LINKED ISSUE #2003: Add a warning when Unconfirmed (0-comf) is selected as an option in the Store Settings
It's an increasing trend in the wallets (like [Electrum ](https://github.com/spesmilo/electrum/pull/6641)and [Bluewallet](https://bluewallet.io/features/)) to add a `Cancel` button. Their reasoning for it is different, but on merchant end, it provides a big problem when the UX to double-spend a transaction gets easier.

On our end, we do handle the 0-confirmations properly, but in this issue I'd like to propose that we add a little warring, just as a bit of education on why a merchant may NOT want to do this. 

I suggest we add a little yellowish `warning` when somebody selects this, just so we're sure they understood the implications.

Add a yellow warning box when user selects `Unconfirmed`. `Choosing to accept an unconfirmed invoice can lead to double-spending and is strongly discouraged.`

Capture

bolatovumar:
close #2003

Screenshot:
Screen Shot 2020-11-01 at 5 18 03 PM

### DIFF
--- a/BTCPayServer/Views/Stores/UpdateStore.cshtml
+++ b/BTCPayServer/Views/Stores/UpdateStore.cshtml
@@ -246,12 +246,20 @@
                             <span class="fa fa-question-circle-o" title="More information..."></span>
                         </a>
                     </div>
-                    <select asp-for="SpeedPolicy" class="form-control w-auto">
+                    <script>
+                        function handleSpeedPolicyChange(select) {
+                            unconfirmedWarning.hidden = select.value !== '0';
+                        }
+                    </script>
+                    <select asp-for="SpeedPolicy" class="form-control w-auto" onchange="handleSpeedPolicyChange(this)">
                         <option value="0">Is unconfirmed</option>
                         <option value="1">Has at least 1 confirmation</option>
                         <option value="3">Has at least 2 confirmations</option>
                         <option value="2">Has at least 6 confirmations</option>
                     </select>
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/Stores/UpdateStore.cshtml
```diff
@@ -246,12 +246,20 @@
                             <span class="fa fa-question-circle-o" title="More information..."></span>
                         </a>
                     </div>
-                    <select asp-for="SpeedPolicy" class="form-control w-auto">
+                    <script>
+                        function handleSpeedPolicyChange(select) {
+                            unconfirmedWarning.hidden = select.value !== '0';
```
Kukks, line 251: Where is `unconfirmedWarning` defined?

Kukks (changes requested):

BTCPayServer/Views/Stores/UpdateStore.cshtml
dennisreimann, line 251:
```suggestion
                            document.getElementById('unconfirmedWarning').hidden = select.value !== '0';
```

bolatovumar: All elements with and id attribute are automatically global variable lol (true story). I can update it to explicitly grab the element by its id if that's clearer.

bolatovumar: Alright, updated here: e88b88a5a30e3348e79c48a3f9548cbd77f71a4b

Kukks (approved):

PR #2042, opened 2020-11-03
Center pay button content

### LINKED ISSUE #1999: Payment Button Improvements for Custom Amount/Simple Input Style
The button I like most for squarespace sites and wix sites and non e commerce optimized sites where you just want to place an invoice button is the Custom Amount, Simple Input Style. Two things I've noticed that make using this button less than optimal are the 0 is not centered and when clicking into the box I've often seen users have trouble erasing the 0 in order to type over it the value that they want to pay.

- Maybe the "0" would be better as a "-"
- Centering the "0"
- Is it possible to have only a single click be necessary to highlight the zero and type over it instead of a double click?

It's not mission critical, just wanted to have a discussion.

**Additional context**

Here's how it looks on the site I'm using

Screen Shot 2020-10-19 at 11 24 11 AM

bolatovumar:
Generated code:
Screen Shot 2020-11-02 at 7 14 17 PM

close #1999

### DIFF
--- a/BTCPayServer/Views/Stores/PayButton.cshtml
+++ b/BTCPayServer/Views/Stores/PayButton.cshtml
@@ -375,6 +375,11 @@
             line-height: 35px;
             background: #fff;
         }
+        #btcpay-input-price::-webkit-outer-spin-button,
+        #btcpay-input-price::-webkit-inner-spin-button {
+          -webkit-appearance: none;
+          margin: 0;
+        }
     </style>
 </script>
 


dennisreimann (changes requested):
This looks more complicated than it needs to be. Why not just add the styles to the already existing style block? (line 1 of the generated output)

bolatovumar:
> This looks more complicated than it needs to be. Why not just add the styles to the already existing style block? (line 1 of the generated output)

Yeah, I could do that.

dennisreimann (approved):

dennisreimann:
Ah wait, the following part is missing now:

```css
        #btcpay-input-price {
          -moz-appearance: textfield;
        }
```

bolatovumar:
> Ah wait, the following part is missing now:
> 
> ```css
> #btcpay-input-price {
>   -moz-appearance: textfield;
> }
> ```

It's already present in the statements above in default styles. Also, setting `-moz-appearance: textfield;` or `-moz-appearance: none;` doesn't actually seem to do anything in Firefox.

Screen Shot 2020-11-07 at 7 06 29 PM

dennisreimann:
Ok, got it. Good to go then 👍

rockstardev (approved):

PR #2043, opened 2020-11-03
Automatically generate permissions docs for GreenField

Kukks:
This automatically documents all available permissions for Api keys in the swagger docs
@dennisreimann The docs builder might need a new way to generate docs 😓


dennisreimann:
We'll find a way :)

From what I see at first glance, the information is now extracted from thee source, so that it's not available in the initial JSON anymore, correct?

Right now we are doing some [funky jq magic](https://github.com/btcpayserver/btcpayserver-doc/blob/master/setup-deps.sh#L47), which resembles what the `public async Task<IActionResult> Swagger()` did before. This won't be fully possible anymore, as it lacks the programmatic introspection part …

Maybe it'd be best to get rid of the jq foo and export the docs JSON on build. This way the `swagger/v1/swagger.json` endpoint would not dynamically generate it on the fly, but just load the previously generated JSON. The docs build could then also consume the generated JSON and be done.

NicolasDorier:
@dennisreimann @Kukks in this case I would prefer that instead of doing what this PR does, we add a test checking we don't forget to put permission in the doc.

dennisreimann:
I can see where this is coming from though. In general I'd say if we can infer something from the code, we should do so – not just because I'm lazy, but because this way the information will be accurate. There's always a chance for forgetting/mixing up stuff, especially since the swagger JSON files tend to be big and have many nestings …

NicolasDorier:
@dennisreimann if there is just a test that make sure we don't forget that, this is enough for not forgetting it.

dennisreimann:
Sure, but e.g. #2052 shows that it is easy to mix things up if you manage this manually. However, there‘s no way to fully automate it either … just wanted to say that I get @kukks approach.

Kukks:
> @dennisreimann @Kukks in this case I would prefer that instead of doing what this PR does, we add a test checking we don't forget to put permission in the doc.
> 
> I am not too fan either of making the swagger generation magic.

I disagree. if anything I wished we had our automatic swagger generator that I had added in the beginning. All this manual API doc writing is very painful and always error prone.

NicolasDorier:
I saved lot's of headache by not having the library, no need to fight against the library, just writing JSON is easy.
What is hard is writing documentation, not writing in JSON.

Automatic generation is a pain, as you need to make model fit the generator, potentially add attributes on our models that would make users needing to also reference the library. I prefer no smart ass code, and keeping it simple. I hate having to fight with libraries similar to what happened with the openid and with EF. It makes things that should be simple complicated, just for the sake of saving some JSON (SQL queries in the case of EF) that anybody could write without passing hours on stackoverflow and github issue on the library they use.

And the thing is: You can actually automatically generate the doc manually, then copy/paste the resulting JSON.
Similar to how we use the swagger client generator on LND library: Most of it is automatically generated, with manual changes on top of it to accomodate the fact that the library need to maintain backward compatibility, even if LND is not.

dennisreimann (approved):

Kukks:
I switched to a test @NicolasDorier

PR #2020, opened 2020-10-26
Plugins: Load plugins by order, aesthetic plugin dependency system

Kukks:
Introduces plugins loading in order of installation, BTCPay itself shows up as a system plugin, and that plugins can define other plugins as dependencies and also use semver syntax.

PR #2048, opened 2020-11-05
Plugins UI improvements

dennisreimann:
Some visual polishing based on #2020.

dependencies

light

toggle

### DIFF
--- a/BTCPayServer/Views/Server/ListPlugins.cshtml
+++ b/BTCPayServer/Views/Server/ListPlugins.cshtml
@@ -83,7 +83,6 @@
             {
                 return false;
             }
-            ;
         }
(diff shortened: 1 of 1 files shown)


Zaxounette:
This might not be strictly UI only, but wouldn't it be a good idea to have an `author` string ?

Kukks:
> This might not be strictly UI only, but wouldn't it be a good idea to have an `author` string ?
If it becomes popular enough to need it,we can add it in later. They can use the description for now ( and display their name in plugin settings if needed)

dennisreimann:
I see this turning into a full blown app store one day 😉

Kukks (approved):

PR #2055, opened 2020-11-07
GreenField: Notifications API

Kukks:
This refactors notifications so that we dont have a bunch of duplicated direct access to db contexts in controllers and then introduces new endpoints to fetch/toggle seen/remove  notifications of the current user.


BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -89,6 +89,11 @@ public class GreenFieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
                         success = true;
                     }
                     break;
+                
+                case Policies.CanViewNotificationsForUser:
+                    success = context.HasPermission(Permission.Create(policy), requiredUnscoped) || context.HasPermission(Permission.Create(Policies.CanManageNotificationsForUser), requiredUnscoped);
```
NicolasDorier, line 94: You should use `Permission.ContainsPolicy`

NicolasDorier:
@Kukks `EnsureSwaggerPermissionsDocumented` failed.

Issue #1064, opened 2019-09-30
Feature Request: Merge BTC and Lightning Addresses into a Single QR


astupidmoose:
According to my understanding of the article below, lightning allows for a fallback BTC address to be included in the lightning generated QR code. 

https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields

TravelbyBit builds this QR into their lightning QR, which allows wallets to scan either a BTC address or lightning address in a single display option. 

https://twitter.com/VNumeris/status/1095826765111451648?s=19

I think it would be a cool way to merge the BTC and Lightning options into 1 qr code. 

Downside would be possible confusion to customers who are used to clicking between the two. 

Putting this here for discussion.

pavlenex:
How would you handle the `copy` part of the invoice?
It may end up being very confusing if you try to distinguish it all in one UI.
* `Copy LN`
* `Copy on-chain`
How would a noob know a difference?

astupidmoose:
Agreed, it might be confusing and needs more thought. Unfortunately a precedent has been set that they are different. Copy would be kind of messy based on its current layout. I need to think about this more. 

However that being said, I think ALL lightning invoices should have a built in BTC QR. I think this is a brilliant feature and eliminates the need to differentiate. From a UX design, I think it's great to have the wallet decide the BTC or lightning flow and have both displayed at once server side. 

:)

bolatovumar:
@astupidmoose I agree about having on-chain backup addresses for lightning invoices. I believe Zap wallet does this by default already (or will allow this in a future version).

pavlenex:
Ok, so instead of merging all into one, your proposal is to have a fallback on LN QR code, is that correct? I think that's a good compromise since it won't invade the current workflow but still provides value.

Do you guys know if this will cause any compatibility issues among Lightning Network wallets? I assume it won't but would be good to think about that as well.

astupidmoose:
Yea, that might be a good idea. 

Incase someone doesn't really understand lightning, then clicks that option and tries to pay they still get an useable QR. 

I wouldn't think it would mess any lightning wallets up as it's built into the standard. But that doesn't really mean anything lol. 

The backup BTC address should also be the BTC address in the BTC tab.

bitbug42:
I really like that idea, would greatly simplify the QR-scanning flow.

A good way to do it IMO would be to use a BIP21 URL like this:
[...]

PR #2060, opened 2020-11-09
Unified QR code, BIP21 with lightning fallback

rockstardev:
Resolves: #1064 

Here is feature in action:

bip21 fallback

As part of PR I've also changed lightning invoices to be lowercase by default, this one is long overdue.

### DIFF
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -30,14 +30,22 @@ namespace BTCPayServer.Client.Models
         public double PaymentTolerance { get; set; } = 0;
         public bool AnyoneCanCreateInvoice { get; set; }
 
+
+        //
+        public bool RequiresRefundEmail { get; set; }
+        public bool LightningAmountInSatoshi { get; set; }
+        public bool LightningPrivateRouteHints { get; set; }
+        public bool OnChainWithLnInvoiceFallback { get; set; }
+        public bool RedirectAutomatically { get; set; }
+
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public bool ShowRecommendedFee { get; set; } = true;
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public int RecommendedFeeBlockTarget { get; set; } = 1;
 
+        //
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public string DefaultLang { get; set; } = "en";
-        public bool LightningAmountInSatoshi { get; set; }
 
         public string CustomLogo { get; set; }
 
@@ -45,16 +53,13 @@ namespace BTCPayServer.Client.Models
 
         public string HtmlTitle { get; set; }
 
-        public bool RedirectAutomatically { get; set; }
 
-        public bool RequiresRefundEmail { get; set; }
 
         [JsonConverter(typeof(StringEnumConverter))]
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public NetworkFeeMode NetworkFeeMode { get; set; } = NetworkFeeMode.Never;
 
         public bool PayJoinEnabled { get; set; }
-        public bool LightningPrivateRouteHints { get; set; }
 
 
         [JsonExtensionData]
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -31,7 +31,6 @@ namespace BTCPayServer.Client.Models
         public bool AnyoneCanCreateInvoice { get; set; }
 
 
-        //
         public bool RequiresRefundEmail { get; set; }
         public bool LightningAmountInSatoshi { get; set; }
         public bool LightningPrivateRouteHints { get; set; }
@@ -43,7 +42,7 @@ namespace BTCPayServer.Client.Models
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public int RecommendedFeeBlockTarget { get; set; } = 1;
 
-        //
+        
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public string DefaultLang { get; set; } = "en";
 
--- a/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
@@ -46,8 +46,7 @@ namespace BTCPayServer.Payments.Bitcoin
             public Task<KeyPathInformation> ReserveAddress;
         }
 
-        public override void PreparePaymentModel(PaymentModel model, InvoiceResponse invoiceResponse,
-            StoreBlob storeBlob)
+        public override void PreparePaymentModel(PaymentModel model, InvoiceResponse invoiceResponse, StoreBlob storeBlob)
         {
             var paymentMethodId = new PaymentMethodId(model.CryptoCode, PaymentTypes.BTCLike);
 
@@ -55,8 +54,19 @@ namespace BTCPayServer.Payments.Bitcoin
             var network = _networkProvider.GetNetwork<BTCPayNetwork>(model.CryptoCode);
             model.IsLightning = false;
             model.PaymentMethodName = GetPaymentMethodName(network);
-            model.InvoiceBitcoinUrl = cryptoInfo.PaymentUrls.BIP21;
-            model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls.BIP21;
+
+            var invoiceBitcoinUrl = cryptoInfo.PaymentUrls.BIP21;
+
+            if (storeBlob.OnChainWithLnInvoiceFallback)
+            {
+                var lightningInfo = invoiceResponse.CryptoInfo.FirstOrDefault(a =>
+                    a.GetpaymentMethodId() == new PaymentMethodId(model.CryptoCode, PaymentTypes.LightningLike));
+                if (!String.IsNullOrEmpty(lightningInfo?.PaymentUrls?.BOLT11))
+                    invoiceBitcoinUrl += "&" + lightningInfo.PaymentUrls.BOLT11.Replace("lightning:", "lightning=", StringComparison.OrdinalIgnoreCase);
+            }
+
+            model.InvoiceBitcoinUrl = invoiceBitcoinUrl;
+            model.InvoiceBitcoinUrlQR = invoiceBitcoinUrl;
         }
 
         public override string GetCryptoImage(PaymentMethodId paymentMethodId)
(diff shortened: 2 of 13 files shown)


shesek:
Assuming the QR library you're using supports automatic mixed modes (I don't know if it does), making the bolt11 part of the URI in all-uppercase will take advantage of the more efficient alphanumeric encoding mode, making the QR smaller, less dense and easier to scan from a distance.

There's some information on this on Bitcoin Optech, in the context of bech32 QRs:
https://bitcoinops.org/en/bech32-sending-support/#creating-more-efficient-qr-codes-with-bech32-addresses

> By convention ln invoices are usually lowercase

They usually are when presented as text, but if you're aware of wallets that use lowercase in QRs, it would be good to give them a nudge and tell them to uppercase them :)

Kukks (approved):

rockstardev:
Thanks @shesek, I've left lowercase as default now, but tweaked our code to make lninvoice uppercased for QR code. Example QR:

image

`bitcoin:bcrt1qudfg7xmh47s33gja2mgrpxkt0c2yujyvm6v82e?amount=0.00032589&lightning=LNBCRT325890N1P063ASVPP57NSJTMJUNK09AJAVQ593DHMRFDTHN3QSCGYU3XXA22MQN28U03QQDP62PSKJEPQW3HJQ4MPD3KX2APQVDEX2CT5D9HKUGPGFAEXGETJYPY5GW3Q9YXQZUPCQP2SP5JFUTDSL7Z2S7C2N9TY4YQDY0FFXGHMDF0G47UP402PFYM08ET6NS9QY9QSQRUHPGHUZ66FDKUJ0KK27C50TENA7MDSVV8RMJN4M0KR8SNTDNMHS53NL0HR4DRQ47Z2XSFXT54NRLSF0AH90YWWQQECRGKE6Q687KCQQQE4YJ3`

I was thinking about making `LIGHTNING=` part of querystring uppercase, but I'll wait for wallet devs (@bitbug42 mentioned Eclair supports this lightning fallback) to confirm uppercase would work. Same for `bitcoin:` over `BITCOIN:`

shesek:
Have you tried comparing the QRs to see if the uppercase really enables the alphanumeric mode and makes them smaller/sparser?

Upper-casing the query-string parameter name is indeed problematic, but using `BITCOIN:` as the schema should be working, I've been using it myself with no issues. It's also recommended on the Bitcoin Optech article linked above:

> Although the preferred form of BIP21’s scheme identifier `bitcoin:` is lowercase, it can also be uppercased as allowed by both BIP21 and [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.1). This also produces a less complex image.

You should also upper-case the bech32 address to take full advantage of the alphanumeric encoding mode (it'll use multiple segments and switch modes between them - an alphanumeric segment for the schema+address, followed by a binary segment for the query, followed by another alphanumeric segment for the bolt11).

Kukks (dismissed):
Code is good, would have been better to do the cleanup as a separate PR but it's ok. I think I will propose to move the actual feature code to some system plugin in the future.

You need to add the new prop to the swagger docs

rockstardev:
@shesek: OK, we've previously removed uppercasing schema in QR Code because some readers were detecting it like `bitcoIN:` for some reason. Here is example QR:

image

I've also added uppercasing of receiving address if it starts with bech32 prefixes. Let's see if @NicolasDorier is OK with merging all this.

@kukks: I still can't believe we're manually editing swagger files ;). Added it.

Kukks:
> @shesek: OK, we've previously removed uppercasing schema in QR Code because some readers were detecting it like `bitcoIN:` for some reason. Here is example QR:
> 
> image
> 
> I've also added uppercasing of receiving address if it starts with bech32 prefixes. Let's see if @NicolasDorier is OK with merging all this.
> 
> @Kukks: I still can't believe we're manually editing swagger files ;). Added it.

Me too. Feel free to voice your swagger feedback at  #2043

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -46,18 +46,41 @@ class Prepare
...
+
+            if (bech32Prefixes.Any(a => model.BtcAddress.StartsWith(a, StringComparison.OrdinalIgnoreCase)))
+            {
+                model.InvoiceBitcoinUrlQR = model.InvoiceBitcoinUrlQR.Replace(
+                    $"BITCOIN:{model.BtcAddress}", $"BITCOIN:{model.BtcAddress.ToUpperInvariant()}", 
+                    StringComparison.OrdinalIgnoreCase
+                );
+            }
         }
+        private static string[] bech32Prefixes = new[] { "bc1", "tb1", "bcrt1" };
```
Kukks, line 83: We should really not have this hardcoded in here.  You should get this based on the current network using soemthing like `NBitcoin.Network.Main.GetBech32Encoder(Bech32Type.WITNESS_PUBKEY_ADDRESS, false).HumanReadablePart`

Kukks (changes requested):

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -46,18 +46,41 @@ class Prepare
...
+                    a.GetpaymentMethodId() == new PaymentMethodId(model.CryptoCode, PaymentTypes.LightningLike));
+                if (!String.IsNullOrEmpty(lightningInfo?.PaymentUrls?.BOLT11))
+                    lightningFallback = "&" + lightningInfo.PaymentUrls.BOLT11.Replace("lightning:", "lightning=", StringComparison.OrdinalIgnoreCase);
+            }
+
+            model.InvoiceBitcoinUrl = cryptoInfo.PaymentUrls.BIP21 + lightningFallback;
+            // We're trying to make as many characters uppercase to make QR smaller
+            // Ref: https://github.com/btcpayserver/btcpayserver/pull/2060#issuecomment-723828348
+            model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls.BIP21
+                .Replace("bitcoin:", "BITCOIN:", StringComparison.OrdinalIgnoreCase)
```
NicolasDorier, line 72: I prefer not doing this, as I don't know if various wallet support it.
Lightning is different, because everybody using lightning did it from beginning.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
NicolasDorier: Can we remove this uppercase stuff?

https://github.com/btcpayserver/btcpayserver/pull/2060/files?w=1#r520458592

I am not comfortable with it. Or maybe we can just activate it for people checking `OnChainWithLnInvoiceFallback` so we can get feedback from early users.

shesek:
> @shesek: OK, we've previously removed uppercasing schema in QR Code because some readers were detecting it like bitcoIN: for some reason.

This could show up like this in some QRs readers, but this shouldn't affect the ability to scan it in any way.

> I prefer not doing this, as I don't know if various wallet support it. Lightning is different, because everybody using lightning did it from beginning.

Well, FWIW, I've been using uppercase bech32 addresses in a bunch of my projects with no issues, including in Esplora and Spark Wallet.

And as mentioned before, using an upper case schema is allowed by BIP21, RFC3986, and is recommended by Optech.

BTCPayServer.Client/Models/StoreBaseData.cs
```diff
@@ -30,31 +30,36 @@ public abstract class StoreBaseData
         public double PaymentTolerance { get; set; } = 0;
         public bool AnyoneCanCreateInvoice { get; set; }
 
+
+        //
```
bolatovumar, line 34: Did you mean to add a comment here but forgot or something?

BTCPayServer.Client/Models/StoreBaseData.cs
```diff
@@ -30,31 +30,36 @@ public abstract class StoreBaseData
...
+        public bool LightningPrivateRouteHints { get; set; }
+        public bool OnChainWithLnInvoiceFallback { get; set; }
+        public bool RedirectAutomatically { get; set; }
+
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public bool ShowRecommendedFee { get; set; } = true;
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public int RecommendedFeeBlockTarget { get; set; } = 1;
 
+        //
```
bolatovumar, line 46: Same here.

rockstardev:
Two open issues:

 1. "`NBitcoin.Network.Main.GetBech32Encoder(Bech32Type.WITNESS_PUBKEY_ADDRESS, false).HumanReadablePart` instead of my l33t specification of exact strings" -> I'm OK with changing that
 2. "I am not comfortable with it. Or maybe we can just activate it for people checking OnChainWithLnInvoiceFallback so we can get feedback from early users" -> we already had `BITCOIN:` schema previously, but removed it because of QR Code readers would sometimes mess up casing. In the light of all the information @shesek is presenting I think it's best we go back to `BITCOIN:`. Same for uppercase BECH32 address in QR. it's all lowercase otherwise.

If you @NicolasDorier  and @Kukks are giving me yes on both - then I'll make the changes and merge PR so we can get this to @pavlenex for release.

NicolasDorier:
@rockstardev ah yeah you are right. Actually now I am surprised we have gone back to lowercase, thinking about it I would have sworn we were lower case for bech32 already...

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
rockstardev: @Kukks these three values are highly unlikely to change, and all the extra code and converting byte[] to string seems unnecessary to me. 

We could do three calls to fetch prefixes from Network class... but this is also why I've added test - so if bech32 prefixes change and world starts to end, we'll get notified that QR codes in BtcPayServer are not properly uppercased ;)

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/2060#issuecomment-727144821

BTCPayServer.Client/Models/StoreBaseData.cs
rockstardev: Sections no longer separated visually ;( 38ca0accde6cf57b40b1bd0c97e8abaab5dfcf38

BTCPayServer.Client/Models/StoreBaseData.cs
rockstardev: 38ca0accde6cf57b40b1bd0c97e8abaab5dfcf38

rockstardev:
Merging after addressing all discussion points. Feel free to further tweak in master if you want, but keep in mind that it's rockstar code, so it's already perfect.

pavlenex:
Eh uppercase schema `BITCOIN`apparently broke some things #2099
Who should we throw tomatoes at?

rockstardev:
@shesek, obviously 😉

shesek:
lol, okay, I can accept that... though I think that at least one or two should be thrown at CashApp's direction too :p

Are you certain that the uppercase protocol is indeed the cause? I would wait for #2099's OP (or some other CashApp user) to reply to https://github.com/btcpayserver/btcpayserver/issues/2099#issuecomment-738456041 before concluding this.

If CashApp indeed doesn't support this, I think that its worth trying to reach out them and ask them to fix it. It's causing them to be incompatible with other wallet software too.

Also, the fact that there's only one uppercase-related bug report for CashApp suggests that other wallets do mostly handle this properly.

pavlenex:
User was using P2SH btw.

PR #2062, opened 2020-11-11
Ensure campaign status is shown correctly in minimal crowdfund app

bolatovumar:
Before:
Screen Shot 2020-11-11 at 3 47 44 PM

After:
Screen Shot 2020-11-11 at 3 44 08 PM

With JS enabled:
Screen Shot 2020-11-11 at 3 44 22 PM

### DIFF
--- a/BTCPayServer/Views/AppsPublic/Crowdfund/MinimalCrowdfund.cshtml
+++ b/BTCPayServer/Views/AppsPublic/Crowdfund/MinimalCrowdfund.cshtml
@@ -87,7 +87,7 @@
                         </h5>
                         <h5 class="text-muted">Contributors</h5>
                     </div>
-                    @if (Model.Started && !Model.Ended)
+                    @if (Model.Started && !Model.Ended && Model.EndDate.HasValue)
                     {
                         <div class="col-sm">
                             <h5>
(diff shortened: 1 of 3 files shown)


Kukks (approved):

BTCPayServer/Views/AppsPublic/Crowdfund/MinimalCrowdfund.cshtml
```diff
@@ -87,7 +87,7 @@
                         </h5>
                         <h5 class="text-muted">Contributors</h5>
                     </div>
-                    @if (Model.Started && !Model.Ended)
+                    @if (Model.Started && !Model.Ended && Model.EndDate.HasValue)
```
Kukks, line 90:
```suggestion
                    @if (Model.Started && Model.Ended is false)
```

Kukks (changes requested):

BTCPayServer/Views/AppsPublic/Crowdfund/MinimalCrowdfund.cshtml
bolatovumar: @Kukks if I just check for `Model.Ended is false` and don't check for `Model.EndDate.HasValue` I will get an error because `Model.Ended` can be true even when `Model.EndDate` doesn't have a value. Did you mean that I should do `Model.Ended is false && Model.EndDate.HasValue` instead of `!Model.Ended && Model.EndDate.HasValue`?

bolatovumar: ping @Kukks

PR #2065, opened 2020-11-16
Introduce Additional Data to Store Blob and move obsolete props to migration

Kukks:


NicolasDorier:
test fail

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -93,6 +111,23 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
             }
         }
 
+        private async Task TransitionToStoreBlobAdditionalData()
```
NicolasDorier, line 114: Why is it necessary? we can just forget them.

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -163,21 +207,76 @@ private async Task ConvertCrowdfundOldSettings()
             }
         }
 
+        private async Task MigratePaymentMethodCriteria()
+        {
+            using (var ctx = _DBContextFactory.CreateContext())
+            {
+                foreach (var store in await ctx.Stores.AsQueryable().ToArrayAsync())
```
NicolasDorier, line 214: I was wondering, is there a way to disable tracking and allow async foreach here?
This would prevent problem if there is many store. Practically, I don't think this is a problem, but if we can do it I think we should.

NicolasDorier:
good to cleanup a bit.

BTCPayServer/Hosting/MigrationStartupTask.cs
Kukks: master plan for plugins

BTCPayServer/Hosting/MigrationStartupTask.cs
Kukks: I think I remember an option to explicitly turn off tracking for an IQueryable, let me see

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -190,24 +289,43 @@ private async Task ConvertMultiplierToSpread()
                 foreach (var store in await ctx.Stores.AsQueryable().ToArrayAsync())
                 {
                     var blob = store.GetStoreBlob();
-#pragma warning disable CS0612 // Type or member is obsolete
                     decimal multiplier = 1.0m;
-                    if (blob.RateRules != null && blob.RateRules.Count != 0)
+                    if (blob.AdditionalData.TryGetValue("rateRules", out var rateRulesJToken))
                     {
-                        foreach (var rule in blob.RateRules)
+                        var rateRules = rateRulesJToken.ToObject<List<RateRule_Obsolete>>();
```
NicolasDorier, line 295: probably does not work. The initial code is deserializing this through NBX serializer, which, among many things, parse in camelCase the properties. While by default Newtonsoft use PascalCase.

NicolasDorier:
Can you at least add tests that pass through all of this complex code? Because if that's bugged it can brick an install.

BTCPayServer/Hosting/MigrationStartupTask.cs
Kukks: It is using the NBX serializer, but as you can see from the screenshot below, the props are saved in camel case
image

Kukks:
> Can you at least add tests that pass through all of this complex code? Because if that's bugged it can brick an install.

Added tests

NicolasDorier:
Seems good to me.

I am a bit worried about the `Stores.ToArrayAsync`. I have like 9000+ stores on the demo instance.
Will take a look later see if we can't improve this a bit, and add some test cover.

PR #2066, opened 2020-11-16
Improve Lightning node info view

dennisreimann:
Added the store name, which I really missed :)

Cleaned up the UI and also the HTML – this unifies the JS/non-JS version.

Include minor code cleanups as well.

dark

light

### DIFF
--- a/BTCPayServer/Controllers/PublicLightningNodeInfoController.cs
+++ b/BTCPayServer/Controllers/PublicLightningNodeInfoController.cs
@@ -41,10 +41,10 @@ namespace BTCPayServer.Controllers
                 var paymentMethodDetails = GetExistingLightningSupportedPaymentMethod(cryptoCode, store);
                 var network = _BtcPayNetworkProvider.GetNetwork<BTCPayNetwork>(cryptoCode);
                 var nodeInfo =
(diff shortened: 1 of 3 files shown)


Kukks (changes requested):
Is the Vue App needed at all in this UI? I think we can handle most of the UI stuff with just the razor syntax  + the clipboard utility js script

dennisreimann:
The QR code is a Vue component, I guess that's the main point here. In LNbank I'm [generating the QR codes server-side](https://github.com/dennisreimann/btcpayserver-lnbank/blob/master/LNbank/Controllers/QrcodeController.cs) – I think we should switch to that as well so that we can offer a good fallback for non-JS cases.

Kukks:
> The QR code is a Vue component, I guess that's the main point here. In LNbank I'm [generating the QR codes server-side](https://github.com/dennisreimann/btcpayserver-lnbank/blob/master/LNbank/Controllers/QrcodeController.cs) – I think we should switch to that as well so that we can offer a good fallback for non-JS cases.

We should make a view(not vue) component for QR codes

Kukks:
Will merge this for now as it's a whole other thing to switch to that

PR #2077, opened 2020-11-22
Fix bug with already used colors being used first for new tx labels

### LINKED ISSUE #2072: Unique wallet tx label colors
The default labels (app, payment-request, etc.) use certain colors for labels. When I make custom labels, it uses those colors again before creating new unique label colors. 

When creating a new custom label for a tx, I would like it to use a color that is not in use for another label.

image

**Additional context**
In the image above, I only get a new color on the third custom label. I wish it was on the first.

bolatovumar:
Close #2072.

# Before

When creating a new label sometimes an already used color is picked first before a new color is picked.

# After

New and previously unused color is always picked when creating a new label if it's available.

### DIFF
--- a/BTCPayServer/Controllers/WalletsController.cs
+++ b/BTCPayServer/Controllers/WalletsController.cs
@@ -107,15 +107,21 @@ namespace BTCPayServer.Controllers
             _pullPaymentService = pullPaymentService;
         }
 
-        // Borrowed from https://github.com/ManageIQ/guides/blob/master/labels.md
+        // Borrowed from https://github.com/ManageIQ/guides/blob/master/labels.md and https://sashamaps.net/docs/resources/20-colors/
         readonly string[] LabelColorScheme = new string[]
         {
             "#fbca04",
             "#0e8a16",
             "#ff7619",
             "#84b6eb",
+            "#4363d8",
+            "#fabed4",
+            "#42d4f4",
+            "#9a6324",
+            "#800000",
+            "#808000",
             "#5319e7",
-            "#000000",
+            "#cdcdcd",
             "#cc317c",
         };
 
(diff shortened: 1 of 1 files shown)


Kukks:
I'm not sure if adding more hardcoded colors is the way to go. I wanted to try to generate a color deterministically from the label text but not sure if it would work in the real world

pavlenex:
> 
> 
> I'm not sure if adding more hardcoded colors is the way to go. I wanted to try to generate a color deterministically from the label text but not sure if it would work in the real world

That's sounds overly complicated and not really necessary, if you guys really want different colors, adding a way to add color hex or color picker is way more simple and effective. The way GitHub labels work is good balance between usability and simplicity imo.

bolatovumar:
> > I'm not sure if adding more hardcoded colors is the way to go. I wanted to try to generate a color deterministically from the label text but not sure if it would work in the real world
> 
> That's sounds overly complicated and not really necessary, if you guys really want different colors, adding a way to add color hex or color picker is way more simple and effective. The way GitHub labels work is good balance between usability and simplicity imo.

The way I extended the color list was to search for a list of the most contrasting colors and add the ones we do not already have. I made sure not to add any colors which didn't contrast enough with existing colors and the color of the text and the "x" icon.

dennisreimann:
Regarding contrast: We need to ensure it also works in dark mode and, at least 8 and 9 might be a problem there. Another thing is the text color on light background: "invoice", 6 and 12 could use a dark font color instead.

britttttk:
> I'm not sure if adding more hardcoded colors is the way to go.

I agree, I think that random would be better

bolatovumar:
Alright, I removed the new colors. The only color change I kept is updating `#000000` to `#cdcdcd`.
Now all this PR does is fix the issue @britttttk was having.

dennisreimann (approved):

Kukks (approved):

PR #2086, opened 2020-11-30
U2F: Update lib and improve error display

### LINKED ISSUE #2085: U2F on Mobile
When logging in with a smartphone I did not get any further than the 2FA request. I have added a U2F USB stick but this was not requested on the smartphone. 

**To Reproduce**
1. Add U2F Key
2. Try to log in with a smartphone

**Expected behavior**
Nothing happens and you wont be able to log in.

- BTCPay Server Version:  v1.0.6.0 (altcoins)
- Deployment Method: [e.g. Docker, Manual, Third-Party-host]
- Browser: Android 10 Firefox 83.1.0

dennisreimann:
The [previously used U2F library](https://github.com/fido-alliance/google-u2f-ref-code/) has been deprecated. The [new one](https://github.com/grantila/u2f-api) does not override the browsers `window.u2f` functionality if it is natively supported. It also displays the appropriate errors and falls back nicely in case the browser does not support U2F.

On mobile Safari I got a browser alert before, which told me about a resource not being able to load. It was triggered by the lack of U2F support, but the message was kind of misleading. With the updates I get this instead:

IMG_6679

Does not close the issue #2085, but it is related to it and should at least improve the error message.


bolatovumar (approved):

pavlenex:
Nice. Has anybody tested this with already registered keys? I assume it won't lock people out.

I've tested on Windows 10 and Firefox, and the experience is slightly different, since it skips through our U2F prompt page, in which case we may not need it, though I'm not quite certain on how it'll work with other OS.

dennisreimann:
Ah ok, there's a short blip after using the device – for me on macOS I land on the page and get the dialog. There was a [`setTimeout`](https://github.com/btcpayserver/btcpayserver/pull/2086/files#diff-41b0ce794f15d7864a1e326286f90ba6284d82b8f5a52f10e85b143065e25da5L44), which might have been placd there on purpose because of the different behaviours – right, @Kukks? If so, I can add it back in.

Kukks:
> Ah ok, there's a short blip after using the device – for me on macOS I land on the page and get the dialog. There was a [`setTimeout`](https://github.com/btcpayserver/btcpayserver/pull/2086/files#diff-41b0ce794f15d7864a1e326286f90ba6284d82b8f5a52f10e85b143065e25da5L44), which might have been placd there on purpose because of the different behaviours – right, @Kukks? If so, I can add it back in.

Best to add it back in, we need the user to be on that page to see the errors

dennisreimann:
`CheckNoDeadLink` fails, otherwise good. @pavlenex can you check again?

Kukks:
> `CheckNoDeadLink` fails, otherwise good. @pavlenex can you check again?

The tests available dont cover the actual UI U2F part. We should NOT merge until we create a test deploy with this somewhere

Kukks:
@pavlenex  I've built a docker image with this branch. The ideas is that you deploy to an existing server that has users using U2F already.

 You can deploy on a server with: 
```bash
export BTCPAY_IMAGE="btcpayserver/btcpayserver:u2ftest"
. ./btcpay-setup.sh -i
```

pavlenex:
Thanks Kukks, I'll start testing this out today.

NicolasDorier:
@pavlenex  how is the test? I want to include this to the bug fix release

pavlenex:
@NicolasDorier I am sorry, I'm being late on this one. I'll test now as well.

On your btcpay registed a U2F key, such as Yubikey or Ledger U2F.

Then do this as per Kukk's instructions:

```
export BTCPAY_IMAGE="btcpayserver/btcpayserver:u2ftest"
. ./btcpay-setup.sh -i
```

Then try to log in and see if you're not locked out.

pavlenex (approved):
@NicolasDorier Just tested and it works on my end.

tACK.

PR #2087, opened 2020-11-30
UI: Store setup finetuning

dennisreimann:
Incorporate @dstrukt's improvement feedback from [this comment](https://github.com/btcpayserver/btcpayserver/pull/2011#issuecomment-735909927).

Also improved the mobile version, which did not wrap.

## Light

light-1

light-3

## Dark

dark-1

dark-2

## Mobile

light-2

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -312,7 +312,7 @@ namespace BTCPayServer.Tests
                 await s.StartAsync();
                 var alice = s.RegisterNewUser();
                 var storeData = s.CreateNewStore();
-                var onchainHint = "A store requires a wallet to receive payments. Set up your wallet.";
+                var onchainHint = "Set up your wallet to receive payments at your store.";
                 var offchainHint = "A connection to a Lightning node is required to receive Lightning payments.";
 
                 // verify that hints are displayed on the store page
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -324,12 +324,13 @@ namespace BTCPayServer.Tests
                     "Warning hint on list not present");
 
                 s.GoToStore(storeData.storeId);
+                Assert.True(s.Driver.PageSource.Contains(onchainHint), "Wallet hint should be present at this point");
+                Assert.True(s.Driver.PageSource.Contains(offchainHint), "Lightning hint should be present at this point");
+
                 s.AddDerivationScheme(); // wallet hint should be dismissed
                 s.Driver.AssertNoError();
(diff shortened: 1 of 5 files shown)


bolatovumar (approved):

Kukks (approved):
Did you remove the hints completely? If so, we should clean the Blob objects from it

dennisreimann:
The hints are still presented as long as the wallet or lightning isn't set up yet. On the store setup page they aren't displayed as warnings anymore though – see e.g. the wallet hint in the first screenshot.

The warning sign gets displayed on the [list view](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UserStoresController.cs#L138). This still relies in the `Hints` in the blob being present, but only checks for the on chain wallet.

I now tried to remove the hints from the blob and switch the checks to use the actual data like this:

```csharp
vm.HintWallet = vm.DerivationSchemes.Any(ds => ds.Enabled) == false;
vm.HintLightning = vm.LightningNodes.Any(ln => ln.Enabled) == false;
```

That works well on the store details page as the `DerivationSchemes` and `LightningNodes` data is already present. That's not the case for the list view, where this would need to be gathered. 

How do we want to proceed?

dstrukt:
Looking good! Will get an updated mobile version (hadn't planned for the additional details).

pavlenex (approved):
tACK.

Kukks:
> The hints are still presented as long as the wallet or lightning isn't set up yet. On the store setup page they aren't displayed as warnings anymore though – see e.g. the wallet hint in the first screenshot.
> 
> The warning sign gets displayed on the [list view](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UserStoresController.cs#L138). This still relies in the `Hints` in the blob being present, but only checks for the on chain wallet.
> 
> I now tried to remove the hints from the blob and switch the checks to use the actual data like this:
> 
> ```cs
> vm.HintWallet = vm.DerivationSchemes.Any(ds => ds.Enabled) == false;
> vm.HintLightning = vm.LightningNodes.Any(ln => ln.Enabled) == false;
> ```
> 
> That works well on the store details page as the `DerivationSchemes` and `LightningNodes` data is already present. That's not the case for the list view, where this would need to be gathered.
> 
> How do we want to proceed?

Created a new issue from that comment. Let's merge

Kukks (approved):

dstrukt:
(Also posted on the MM)

We should determine how important including the "Wallet" text on mobile is.

Here are some variations, the first being probably the most informative and least disruptive, second more minimal, third with everything included, but I don't love it, and if we decide this direction is the one to go in, I want to explore one last thing.

Screen Shot 2020-12-02 at 5 58 14 PM

PR #2091, opened 2020-12-01
Removing word Bitcoin from forks

rockstardev:
We [got word](https://www.reddit.com/r/btc/comments/k3ya7k/the_people_behind_btcpay_server_are_a_bunch_of/) that we're inconsistently treating Bitcoin forks, so I've fixed the naming of methods and variables.


Kukks (approved):

kristapsk:
Concept ACK

bolatovumar (approved):

BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.BPlus.cs
```diff
@@ -5,22 +5,22 @@ namespace BTCPayServer
...
             var nbxplorerNetwork = NBXplorerNetworkProvider.GetFromCryptoCode("XBC");
             Add(new BTCPayNetwork()
             {
                 CryptoCode = nbxplorerNetwork.CryptoCode,
-                DisplayName = "Bitcoinplus",
+                DisplayName = "BPlus",
                 BlockExplorerLink = NetworkType == NetworkType.Mainnet ? "https://chainz.cryptoid.info/xbc/tx.dws?{0}" : "https://chainz.cryptoid.info/xbc/tx.dws?{0}",
                 NBXplorerNetwork = nbxplorerNetwork,
-                UriScheme = "bitcoinplus",
+                UriScheme = "bplus-fix-it",
```
Kukks, line 17:
```suggestion
                UriScheme = "bitcoinplus",
```

just their uri scheme, let it be

Kukks (commented):
leave the uri scheme as is so we don't break any usage

BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.BPlus.cs
Kukks: Apparently it is not even registered and used, so resolving this

PR #2095, opened 2020-12-02
GreenField: Update invoice metadata

### LINKED ISSUE #1975: GreenField API: Remaining Invoice work
We have released an incomplete invoice API a few versions ago. 
Notable missing features include:
* a notification system to be alerted on a status change and payment
* an endpoint to fetch existing payments #1901
* an endpoint to fetch payment method details ( current address, payment link, etc) #1901
* an endpoint to set the user email to an existing invoice (if it has not yet been set) similar to how the Checkout UI does it. #1902

Kukks:
closes #1975


BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -441,6 +441,22 @@ public async Task ToggleInvoiceArchival(string invoiceId, bool archived, string
...
+            {
+                var invoiceData = await GetInvoiceRaw(invoiceId);
+                if (invoiceData == null || (storeId != null &&
+                                            !invoiceData.StoreDataId.Equals(storeId,
+                                                StringComparison.InvariantCultureIgnoreCase)))
+                    return null;
+                var blob = invoiceData.GetBlob(_Networks);
+                blob.Metadata =  InvoiceMetadata.FromJObject(metadata);
+                invoiceData.Blob = ToBytes(blob);
+                await context.SaveChangesAsync().ConfigureAwait(false);
```
NicolasDorier, line 456: you don't need configureawait, but ok

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -441,6 +441,22 @@ public async Task ToggleInvoiceArchival(string invoiceId, bool archived, string
...
+        {
+            using (var context = _ContextFactory.CreateContext())
+            {
+                var invoiceData = await GetInvoiceRaw(invoiceId);
+                if (invoiceData == null || (storeId != null &&
+                                            !invoiceData.StoreDataId.Equals(storeId,
+                                                StringComparison.InvariantCultureIgnoreCase)))
+                    return null;
+                var blob = invoiceData.GetBlob(_Networks);
+                blob.Metadata =  InvoiceMetadata.FromJObject(metadata);
```
NicolasDorier, line 454: In theory we should update the OrderId column in case it changes. In practice I don't think we need.

NicolasDorier:
In a followup PR we should allow update of the checkout experience as well.

Issue #2099, opened 2020-12-03
CashApp doesn't recognize invoice as "Bitcoin Wallet" when scanning QR code


pretyflaco:
**Describe the bug**
Customer reported that "when scanning the BTCPay QR code, I get an error that it's not a Bitcoin wallet"

The BTCPay instance version is v1.0.6.0.

**To Reproduce**
Don't have CashApp to reproduce. 

**Expected behavior**
Recognize either on-chain or off-chain invoice

pavlenex:
Hi, we need a bit more context to the issue.

What's the address format that you're using for your wallet? SegWit/Legacy/P2SH? Do you by any chance have the `
Include lightning invoice fallback to on-chain BIP21 payment url` enabled in Store > Checkout experience?

britttttk:
I can reproduce this @pretyflaco regardless of wallet type it will not work, it gives the error you mentioned. I tried with and without payment amounts specified in cash app. 

I can however send to the BTCPay Wallet's Receive page from cash app.

pavlenex:
Isn't this a cashapp issue then?

britttttk:
I can reproduce it does not work with or without: lightning/payjoin/bip21.

pretyflaco:
Address format is **P2SH** 
Include lightning invoice fallback to on-chain BIP21 payment url is **not** enabled

britttttk:
It works in version 1.0.5.9 but stopped in 1.0.6.0

pavlenex:
Ok so the issue is most likely that we've added uppercase for the QR in 1.0.6.0
`BITCOIN:BC1XYZ?amount=0.00000001` vs `bitcoin:bc1xyz?amount=0.00001`  So uppercase BITCOIN which should be according to BIP21 protocol isn't recognized by the CashApp. Would be good to test this with a few other wallets.

rockstardev:
> So uppercase BITCOIN: which should be according to BIP21 protocol isn't recognized by the CashApp.

I think it's likely that `BITCOIN:` is supported, but uppercased Bech32 address in QR code is not supported. @britttttk since you have CashApp can you go to:

https://www.the-qrcode-generator.com

And then try following combinations:

BITCOIN:BC1Q6APR55WRXTDTK3CJ6M69MXAYW99VVGTVDNVUNT?amount=0.00005142
BITCOIN:bc1q6apr55wrxtdtk3cj6m69mxayw99vvgtvdnvunt?amount=0.00005142
[...]

PR #2100, opened 2020-12-03
Reverting uppercasing of Bech32 addresses in QR code

rockstardev:
Resolves: #2099

Linked issue contains all the details and great discussion. For reasons on why we decided to uppercase bech32 address in QR codes in the first place see: https://github.com/btcpayserver/btcpayserver/pull/2060#issuecomment-723828348


bolatovumar (approved):

NicolasDorier:
I will make new release.

Kixunil:
Does this upper case *protocol* or *addresses*? #2099 says and according to the analysis it should be it's protocol, the title of this says addresses. Which is it?

PR #2101, opened 2020-12-04
Add "Mark all as seen" button to notification dropdown

bolatovumar:
## Motivation

I often find myself wanting to just clear all notifications after I have seen them. In order to do this I need to go to the "Notifications" page, mass-select them and then perform a batch action. This takes four clicks to do. I would like to be able to do this quicker.

## Solution

Added a "Mark all as seen" button in the notification dropdown which allows me to accomplish the goal of marking all notifications as seen in two clicks instead of four.

Screen Shot 2020-12-03 at 7 54 20 PM

### DIFF
--- a/BTCPayServer/Controllers/NotificationsController.cs
+++ b/BTCPayServer/Controllers/NotificationsController.cs
@@ -206,6 +206,17 @@ namespace BTCPayServer.Controllers
             return RedirectToAction(nameof(Index));
         }
 
+        [HttpPost]
+        public async Task<IActionResult> MarkAllAsSeen(string returnUrl)
+        {
+            if (!ValidUserClaim(out var userId))
+            {
+                return NotFound();
+            }
+            await _notificationManager.ToggleSeen(new NotificationsQuery() {Seen = false, UserId = userId}, true);
+            return Redirect(returnUrl);
+        }
+
         private bool ValidUserClaim(out string userId)
         {
             userId = _userManager.GetUserId(User);
--- a/BTCPayServer/Components/NotificationsDropdown/Default.cshtml
+++ b/BTCPayServer/Components/NotificationsDropdown/Default.cshtml
@@ -2,6 +2,7 @@
 @inject UserManager<ApplicationUser> UserManager
 @inject CssThemeManager CssThemeManager
 @using BTCPayServer.HostedServices
+@using Microsoft.AspNetCore.Http.Extensions
 @model BTCPayServer.Components.NotificationsDropdown.NotificationSummaryViewModel
 
 @if (Model.UnseenCount > 0)
@@ -24,6 +25,9 @@
                 </a>
             }
             <a class="dropdown-item text-secondary" asp-controller="Notifications" asp-action="Index">See All</a>
+            <form asp-controller="Notifications" asp-action="MarkAllAsSeen" asp-route-returnUrl="@Context.Request.GetDisplayUrl()" method="post">
+                <button class="dropdown-item text-secondary" type="submit"><i class="fa fa-eye"></i> Mark all as seen</button>
+            </form>
         </div>
     </li>
 }
--- a/BTCPayServer/Services/Notifications/NotificationSender.cs
+++ b/BTCPayServer/Services/Notifications/NotificationSender.cs
@@ -13,6 +13,10 @@ namespace BTCPayServer.Services.Notifications
     public class UserNotificationsUpdatedEvent
     {
         public string UserId { get; set; }
+        public override string ToString()
+        {
+            return string.Empty;
+        }
     }
     public class NotificationSender
     {


Kukks (changes requested):
Useful, but I would wait until #2055 is merged as there is a cleanup on the notification backend code which put most of the notification code in a service instead. 

In terms of style, IMO there should be an individual list item (border in between the 2 options). Maybe @pavlenex and @dennisreimann have input on this.

dennisreimann:
I agree with @kukks, drawing a line between those actions to let the appear separate would be good. Maybe we can have one action/meta row and fit both "See all" and "Mark all read" besides each other.

I'd also propose to tinker a bit with having a different bg/text color for this row and maybe put it up at the top of the flyout. I feel like this is the more usual position for the actions and it would work better with longer notification lists.

dstrukt:
Will mock out an updated view, should have something tonight or tomorrow morning.

dstrukt:
In true fashion, I'm presenting a few options :)

My personal favorite is Option B, the right (with more padding), all based on the 4px base (4, 8, 16, 24, etc...).

I realize this would be a bit more work, but this notification overlay was an area I wanted to address, so I thought I'd go ahead and share something a bit more ambitious with this addition, it's all cosmetic, but if it's out of scope, I understand.

As for the "eye" icon, I'm ambivalent, I could take it or leave it, with the addition of (potential) future icons however, it's cleaner not to have it, but for the current state .. shrug.

As usual, open to any and all feedback!

Notifications

dstrukt:
(reposting, the original export was incorrect)

In my true fashion, I'm presenting a few options :)

My personal favorite is Option B, the right (with more padding), all based on the 4px base (4, 8, 16, 24, etc...).

I realize this would be a bit more work, but this notification overlay was an area I wanted to address, so I thought I'd go ahead and share something a bit more ambitious with this addition, it's all cosmetic, but if it's out of scope, I understand.

As for the "eye" icon, I'm ambivalent, I could take it or leave it, with the addition of (potential) future icons however, it's cleaner not to have it, but for the current state .. shrug.

As usual, open to any and all feedback!

Notifications

dennisreimann:
I like the ones with less padding a bit better, imho Option A with the two rows per item works well too.

bolatovumar:
> (reposting, the original export was incorrect)
> 
> In my true fashion, I'm presenting a few options :)
> 
> My personal favorite is Option B, the right (with more padding), all based on the 4px base (4, 8, 16, 24, etc...).
> 
> I realize this would be a bit more work, but this notification overlay was an area I wanted to address, so I thought I'd go ahead and share something a bit more ambitious with this addition, it's all cosmetic, but if it's out of scope, I understand.
> 
> Option A, extending the current version with a similar pattern.
> 
> Option C, a modified version with everything on the bottom, which I suppose I could make an option D using the current text + date with everything on the bottom, but I'm not particularly a fan of this direction, so I didn't
> 

I'm OK with all options. Option A is probably the simplest as it's closest to what we currently have.

> As for the "eye" icon, I'm ambivalent, I could take it or leave it, with the addition of (potential) future icons however, it's cleaner not to have it, but for the current state .. shrug.
> 

I only added the "eye" icon because the Notifications page itself had it for the "Mark all as seen" option.

> As usual, open to any and all feedback!
> 
> Notifications

dstrukt:
Sounds like the consensus is the 16px padding over the 24px option to keep things a bit tighter vertically - sounds good! The more I look at these, I think you guys are right on this call.

Keeping the item / date the way it currently is also sounds good - easy enough change down the road if we want to adjust for any reason.

If we roll with Option A, I think #1 (left) is probably the way to go, consistent vertical padding with 16px, and 24px padding on left & right. Attached the reduced version with the multi-wrap text / icon treatment, and updated on the Figma.

Good to know on the "eye" icon. Included both (regardless of which icon) to comparison, happy to go with either, still ambivalent. 

https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY?node-id=2561%3A2310

Notifications0 2

dennisreimann:
> Good to know on the "eye" icon. Included both (regardless of which icon) to comparison, happy to go with either, still ambivalent.

I'd omit it, as it looks cleaner and the icon doesn't add much meaning or convey this clearer here.

bolatovumar:
> > Good to know on the "eye" icon. Included both (regardless of which icon) to comparison, happy to go with either, still ambivalent.
> 
> I'd omit it, as it looks cleaner and the icon doesn't add much meaning or convey this clearer here.

Yeah, I agree.

dstrukt:
Sounds good, and agreed, it doesn't add much value, and I always opt for cleaner, but try not

bolatovumar:
Do we want to start implementing this new design or wait until #2055 is merged?

pavlenex:
@Kukks Do we want to wait for notification API or can @bolatovumar continue to work on this one?

Kukks:
Let's wait

> @Kukks Do we want to wait for notification API or can @bolatovumar continue to work on this one?



> Do we want to start implementing this new design or wait until #2055 is merged?

NicolasDorier:
I would be for merging this for next release. @Kukks your notification API seems great, but I don't think it is conflicting at all with this PR. If anything we would need to encapsulate this code in a service, but I don't see that as necessary for your PR.

BTCPayServer/Controllers/NotificationsController.cs
```diff
@@ -219,6 +219,26 @@ public async Task<IActionResult> MassAction(string command, string[] selectedIte
...
 
+        [HttpPost]
+        public async Task<IActionResult> MarkAllAsSeen(string returnUrl)
+        {
+            if (!ValidUserClaim(out var userId))
+            {
+                return NotFound();
+            }
+
+            var items = _db.Notifications.Where(a => a.ApplicationUserId == userId && !a.Seen);
```
NicolasDorier, line 230: I am raging inside... in pure SQL this would be 1 simple query :/ fuck ORMs

Kukks:
> I would be for merging this for next release. @Kukks your notification API seems great, but I don't think it is conflicting at all with this PR. If anything we would need to encapsulate this code in a service, but I don't see that as necessary for your PR.

My PR already did the hard work and moving dB access code to a service. Merging this PR means I have to clean up stuff again after this is merged

Kukks:
@bolatovumar I force pushed to this branch and based it on my PR and changed the backend code to fit. Only need to update the UI now

bolatovumar:
> @bolatovumar I force pushed to this branch and based it on my PR and changed the backend code to fit. Only need to update the UI now

Thanks. I'm OK with waiting, it's not a rush.

NicolasDorier:
I will merge this for next release and open an issue for the redesign of notification popup.

bolatovumar:
Lol @NicolasDorier is trigger-happy as always.

PR #2103, opened 2020-12-04
Lowercasing the protocol in QR code because it's not widely supported

rockstardev:
Thanks to @britttttk and her testing: https://github.com/btcpayserver/btcpayserver/issues/2099#issuecomment-738550565


shesek:
I'm not sure if "not widely supported" is quite accurate... are you aware of any wallets besides CashApp that don't support this?

(this might still be the right call, even if its just one wallet, but upper case QRs are widely supported in almost all wallets to the best of my knowledge)

rockstardev:
I wanted to address this problem with zero doubt until next version release, so went with all lowercase in coordination with rest of the team @NicolasDorier @pavlenex.

This time if we're doing this I think it's best to first do a community wide testing... opened issue to get the effort started: #2110 ... feel free to contribute @shesek if you have a wallet on your devices.

PR #2106, opened 2020-12-04
Introduce Spam protection

### LINKED ISSUE #1958: Policy for a server admin to limit the accounts per API call
This issue builds on #1933 and proposes that certain cases of spam go through the API.

The proposal is to add a Server Setting > Policy which would put a limit on the amount of accounts that can be created per API call.

**Additional context**
In a way I believe both #1933 and this issue can be good protection from a different kind of spam.

@rockstardev @Kukks may be able to expand on the technicalities here. I just opened an issue so we don't forget about it.

Kukks:
fixes #1958

Adds 2 new options:
* Do not allow stores to use the email settings of the server. Instead, they would need to fill in the email settings in their own store
* Do not allow user creation through the API unless you are an admin.

Both are opt-in and turned off by default.

### DIFF
--- a/BTCPayServer/Services/Mails/EmailSenderFactory.cs
+++ b/BTCPayServer/Services/Mails/EmailSenderFactory.cs
@@ -1,28 +1,34 @@
+using BTCPayServer.HostedServices;
 using BTCPayServer.Services.Stores;
 
 namespace BTCPayServer.Services.Mails
 {
     public class EmailSenderFactory
     {
-        private readonly IBackgroundJobClient _JobClient;
-        private readonly SettingsRepository _Repository;
-        private readonly StoreRepository _StoreRepository;
+        private readonly IBackgroundJobClient _jobClient;
+        private readonly SettingsRepository _repository;
+        private readonly StoreRepository _storeRepository;
+        private readonly CssThemeManager _cssThemeManager;
 
         public EmailSenderFactory(IBackgroundJobClient jobClient,
             SettingsRepository repository,
-            StoreRepository storeRepository)
+            StoreRepository storeRepository,
+            CssThemeManager cssThemeManager)
         {
-            _JobClient = jobClient;
-            _Repository = repository;
-            _StoreRepository = storeRepository;
+            _jobClient = jobClient;
+            _repository = repository;
+            _storeRepository = storeRepository;
+            _cssThemeManager = cssThemeManager;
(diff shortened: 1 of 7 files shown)


pavlenex:
@rockstardev can you please review this one when you get time? It's related to API spam that third-party hosts like @Eskyee were having.

BTCPayServer/Services/Mails/EmailSenderFactory.cs
```diff
@@ -1,28 +1,34 @@
...
 
         public IEmailSender GetEmailSender(string storeId = null)
         {
-            var serverSender = new ServerEmailSender(_Repository, _JobClient);
+            var serverSender = new ServerEmailSender(_repository, _jobClient);
             if (string.IsNullOrEmpty(storeId))
                 return serverSender;
-            return new StoreEmailSender(_StoreRepository, serverSender, _JobClient, storeId);
+            return new StoreEmailSender(_storeRepository,
+                !_cssThemeManager.Policies.DisableStoresToUseServerEmailSettings ? serverSender : null, _jobClient,
```
rockstardev, line 30: `StoreEmailSender` class depends on FallbackSender not being `null`... otherwise `FallbackSender.GetEmailSettings();` will throw null exception:

https://github.com/btcpayserver/btcpayserver/blob/64ab8e3fbea1dac000a2d3d9a0870fdbe80469b9/BTCPayServer/Services/Mails/StoreEmailSender.cs#L34

rockstardev (changes requested):
Both options look solid to me... if we agree they should become long term part of Spam solution, I would encourage having tests as well.

NicolasDorier:
need tests

> Do not allow user creation through the API unless you are an admin

I really think this one is quite unecessary. If somebody wants to spam, they will even without the API.

Kukks:
> need tests
> 
> > Do not allow user creation through the API unless you are an admin
> 
> I really think this one is quite unecessary. If somebody wants to spam, they will even without the API.

This is mostly about reducing email spam. When combining both options together, it should make a difference

Kukks:
Added tests

BTCPayServer/Services/Mails/EmailSenderFactory.cs
Kukks: got it, thanks!

PR #2117, opened 2020-12-08
UI: Improve mobile store setup view

dennisreimann:
According to [this comment](https://github.com/btcpayserver/btcpayserver/pull/2087#issuecomment-737672180) by @dstrukt.

Screenshot from XS and SM viewport size.

## Not set up

setup-xs
setup-sm

## Set up, but disabled

setup-disabled-xs
setup-disabled-sm

## Enabled

enabled-xs
enabled-sm

### DIFF
--- a/BTCPayServer/Views/Stores/UpdateStore.cshtml
+++ b/BTCPayServer/Views/Stores/UpdateStore.cshtml
@@ -27,27 +27,24 @@
             <ul class="list-group mb-3">
                 @foreach (var scheme in Model.DerivationSchemes.OrderBy(scheme => scheme.Collapsed))
                 {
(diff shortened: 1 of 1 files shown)


dstrukt:
This looks excellent!

I think the only thing that looks a bit off are font weights.

The group titles are correct, the card title is correct, as is the xpub/type.

The status looks a little too heavy, both the status and "Modify" should have a "semibold" weight (same as the card title "BTC", if that's semibold).

@dennisreimann I saw your latest post on BS5, we should chat in MM to figure out how we want to approach those changes, etc.. as I would imagine that will be a nice +1 for many screens, etc..

PR #2118, opened 2020-12-09
UI: Improve mobile store setup view

dennisreimann:
Follow up to #2117, according to [this comment](https://github.com/btcpayserver/btcpayserver/pull/2117#issuecomment-741255465).

Improved font weights on the right side – semibold instead of bold:


dstrukt:
Send it!

I'll try to catch these last nit-picky things quicker, but you guys are merging faster than I can review..I love it!

PR #2122, opened 2020-12-10
Hide pagination & page size when not necessary 

### LINKED ISSUE #2121: Hide pagination & page size when not necessary
I find the pagination & page size noisy when there is only a small number of table entries, see attached.

I would like to propose that until the minimum # of entries is displayed (in this case 50, but could be dynamic depending on page / content), the respective UI components are hidden to keep the UI cleaner. Similarly for page size ("1 to 2 of 2").

These components would be visible when the minimum conditions are met.

One instance where this happens, and a proposal of how much cleaner it can be (WIP with additional newer design system / grid).
Screen Shot 2020-08-23 at 6 26
Screen Shot 2020-12-09 at 6 49 56 PM

No alternatives, but happy to consider them.

**Additional context**
None needed.

dennisreimann:
As proposed by @dstrukt in [this issue](https://github.com/btcpayserver/btcpayserver/issues/2121). Closes #2121.

I saw that the pager component wasn't used for the notifications list and wallet transactions. The former was straight forward to integrate – for the latter I had to introduce the `PaginationQuery` property in the `BasePagingViewModel`: This list has the custom `labelFilter` query param, which we need to pass through on the pagination links.

I've talked to @kukks about this and we decided that using a `Dictionary` here would be a good choice, because it allows for easy query customizability like in this case.

### DIFF
--- a/BTCPayServer/Components/Pager/Default.cshtml
+++ b/BTCPayServer/Components/Pager/Default.cshtml
@@ -4,7 +4,7 @@
     var pageSizeOptions = new [] { 50, 100, 250, 500 };
 }
 
-@if (Model.Total > pageSizeOptions[0])
+@if (Model.Total > pageSizeOptions.Min())
 {
     <nav aria-label="..." class="w-100">
         @if (Model.Total > Model.Count)
@@ -33,7 +33,7 @@
             </ul>
         }
 
-        @if (Model.Total >= pageSizeOptions[0])
+        @if (Model.Total >= pageSizeOptions.Min())
         {
             <ul class="pagination float-right">
                 <li class="page-item disabled">
--- a/BTCPayServer/Components/Pager/Default.cshtml
+++ b/BTCPayServer/Components/Pager/Default.cshtml
@@ -1,57 +1,52 @@
 @model BasePagingViewModel
 
-@if (Model.Total > Model.Count)
+@{
+    var pageSizeOptions = new [] { 50, 100, 250, 500 };
+}
+
+@if (Model.Total > pageSizeOptions[0])
 {
     <nav aria-label="..." class="w-100">
-        <ul class="pagination float-left">
-            <li class="page-item @(Model.Skip == 0 ? "disabled" : null)">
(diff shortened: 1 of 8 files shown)


BTCPayServer/Components/Pager/Default.cshtml
```diff
@@ -1,6 +1,6 @@
 @model BasePagingViewModel
 
-@if (Model.Total > 0)
+@if (Model.Total > Model.Count)
```
Kukks, line 3: This should probably be a different criteria. If you set the page size to something big, you'll never be able to unset it.
It should be compared to the smallest page size option instead

Kukks (changes requested):
+34 −157  awesome

Just got one change request for when to hide the pager

BTCPayServer/Components/Pager/Default.cshtml
dennisreimann: Yup, thanks! Addressed in 97a832e13e3c35649c2f14f5874abedd519950dc.

dennisreimann:
`CanUseWebhooks` fails, unrelated to the changes here.

BTCPayServer/Components/Pager/Default.cshtml
Kukks: Where is that commit? It's not applied to this PR.
also use `pageSizeOptions.Min()` instead

Kukks (changes requested):

BTCPayServer/Components/Pager/Default.cshtml
```diff
@@ -1,57 +1,52 @@
 @model BasePagingViewModel
 
-@if (Model.Total > 0)
+@{
+    var pageSizeOptions = new [] { 50, 100, 250, 500 };
+}
+
+@if (Model.Total > pageSizeOptions[0])
```
Kukks, line 7:
```suggestion
@if (Model.Total > pageSizeOptions.Min())
```

Kukks (changes requested):

dstrukt:
I love it.. absolutely knocking through these!

bolatovumar (approved):

PR #2123, opened 2020-12-10
Refactor label implementation (Fix #2090)

### LINKED ISSUE #2090: Automatic tx labels "remove" icon
**Describe the bug**

Transactions labels which are automatically assigned such as "invoice" or "payment-request" still have the "x" icon next to them even though they are not actually removable. I don't believe they are supposed to be removable (correct me if I'm wrong) as they are automatically assign and allow the user to filter through different kinds of transactions easily.

Screen Shot 2020-11-30 at 8 24 59 PM

These automatic labels have custom colors and configuration:

Screen Shot 2020-11-30 at 8 25 42 PM

**To Reproduce**
1. Generate an invoice
2. Pay an invoice
3. Notice you have a new transaction in your wallet
4. Notice it has an "invoice" label
5. Notice it has an "x" icon next to it but clicking it does nothing

**Expected behavior**

I assume the "x" icon should not show up for any of these automatically-assigned transaction labels?

- BTCPay Server Version: v1.0.5.9

NicolasDorier:
Refactoring of the Label system.

Renamed `Label` to `ColoredLabel`.
Before, we were manipulating label via `string` and parsing the string as needed to build `ColoredLabel`.
Instead, this code is now manipulating `Label`. A `Label` is the object model representation of a label, without color.
The `LabelFactory` goal is to transform a `Label` into a `ColoredLabel` ready to show on the screen.

This keep backward compatibility (need additional tests though)

### DIFF
--- a/BTCPayServer.Data/Data/WalletTransactionData.cs
+++ b/BTCPayServer.Data/Data/WalletTransactionData.cs
@@ -26,11 +26,4 @@ namespace BTCPayServer.Data
                 .WithMany(w => w.WalletTransactions).OnDelete(DeleteBehavior.Cascade);
         }
     }
-
-    public class WalletTransactionInfo
-    {
-        public string Comment { get; set; } = string.Empty;
-        [JsonIgnore]
-        public HashSet<string> Labels { get; set; } = new HashSet<string>();
-    }
 }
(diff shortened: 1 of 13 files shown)


Kukks (commented):
Does coin selection still work?

NicolasDorier:
Tests pass, so I would be surprised it does not work. Will check.

NicolasDorier:
there is a bug for coinselection. Fixing it... should have test there

NicolasDorier:
thanks, fixed.

PR #2124, opened 2020-12-10
Add checkoutLink and defaultLang to GreenField invoice

NicolasDorier:


BTCPayServer.Client/Models/CreateInvoiceRequest.cs
```diff
@@ -31,9 +31,9 @@ public class CheckoutOptions
             public TimeSpan? Monitoring { get; set; }
 
             public double? PaymentTolerance { get; set; }
-
             [JsonProperty("redirectURL")]
             public string RedirectURL { get; set; }
+            public string DefaultLang { get; set; }
```
Kukks, line 36: can we name this better now?
`
```suggestion
            public string DefaultLanguage { get; set; }
```

BTCPayServer.Client/Models/InvoiceData.cs
```diff
@@ -7,6 +7,7 @@ namespace BTCPayServer.Client.Models
     public class InvoiceData : CreateInvoiceRequest
     {
         public string Id { get; set; }
+        public string CheckoutLink { get; set; }
```
Kukks, line 10: ugh I would really prefer we don't do this. It is very opinionated and depends a lot on the calling environment (multiple domains, onion, reverse proxy, ssh tunnelling, etc)

Kukks (changes requested):

BTCPayServer.Client/Models/CreateInvoiceRequest.cs
NicolasDorier: Actually I started doing that, but "Lang" is commonly used. No strong pref though.

BTCPayServer.Client/Models/InvoiceData.cs
NicolasDorier: No, we fill that out from the HttpRequest which gives the link adapted. (ie. if you query on onion, you get onion link)

People should not have to build the URL by themselves.

NicolasDorier:
renamed

PR #2125, opened 2020-12-10
Introduce QR Code View component

### LINKED ISSUE #2067: QR code view component

As [discussed here](https://github.com/btcpayserver/btcpayserver/pull/2066#issuecomment-728745799), our current approach of generating QR codes relies on JavaScript being enabled. 

We should add a server-side version for generating QR codes, so that we have a good fallback. I've implemented the [basics](https://github.com/dennisreimann/btcpayserver-lnbank/blob/master/LNbank/Controllers/QrcodeController.cs) already and we should extend this with a custom view component, that encapsulated the functionality we need.

Kukks:
fixes #2067 

This adds a C# QR code generator and a view component to wrap it in. It removes the need for JS in both the ln node info and the wallet receive pages. Apparently, there is a QR spec that states there is a standardized padding amount for a QR code and this PR also changes it on effected pages to follow it.

### DIFF
--- a/BTCPayServer/Views/Wallets/WalletReceive.cshtml
+++ b/BTCPayServer/Views/Wallets/WalletReceive.cshtml
@@ -1,5 +1,4 @@
-﻿
-@addTagHelper *, BundlerMinifier.TagHelpers
+﻿@addTagHelper *, BundlerMinifier.TagHelpers
 @model BTCPayServer.Controllers.WalletReceiveViewModel
 @{
     Layout = "../Shared/_NavLayout.cshtml";
@@ -10,7 +9,7 @@
 {
     <div class="row">
         <div class="col-md-12 text-center">
-            <partial name="_StatusMessage" />
+            <partial name="_StatusMessage"/>
         </div>
     </div>
 }
@@ -20,7 +19,6 @@
             <div class="card-body">
                 @if (string.IsNullOrEmpty(Model.Address))
                 {
-
                     <h3 class="card-title mb-3">Receive @Model.CryptoCode</h3>
                     <button id="generateButton" class="btn btn-lg btn-primary" type="submit" name="command" value="generate-new-address">Generate @Model.CryptoCode address</button>
                 }
@@ -30,7 +28,7 @@
                     <noscript>
                         <div class="card-body m-sm-0 p-sm-0">
                             <div class="input-group">
-                                <input type="text" class="form-control " readonly="readonly" asp-for="Address" id="address" />
+                                <input type="text" class="form-control " readonly="readonly" asp-for="Address" id="address"/>
                                 <div class="input-group-append">
                                     <span class="input-group-text fa fa-copy"> </span>
                                 </div>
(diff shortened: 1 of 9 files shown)


BTCPayServer/Views/Wallets/WalletReceive.cshtml
```diff
@@ -47,12 +45,11 @@
...
                         <div class="qr-container mb-4">
-                            <img v-bind:src="srvModel.cryptoImage" class="qr-icon" />
-                            <qrcode v-bind:value="srvModel.address" :options="{ width: 256, margin: 1, color: {dark:'#000', light:'#f5f5f7'} }" tag="svg">
-                            </qrcode>
+                            <img src="@Model.CryptoImage" class="qr-icon"/>
+                            <vc:qr-code data="@Model.Address"/>
                         </div>
-                        <div class="input-group copy" data-clipboard-target="#vue-address">
-                            <input type="text" class=" form-control " readonly="readonly" :value="srvModel.address" id="vue-address" />
+                        <div class="input-group copy" data-clipboard="@Model.Address" data-clipboard-confirm="Copied Address ✔">
```
Kukks, line 51: Seems like there is a bug with `data-clipboard` when you use it on an input @dennisreimann

dennisreimann: Ok, this works: You have to replace the `input-group` div with this … can't add it as a suggestion here, but you'll figure it out ;)

```html
                        <div class="input-group copy" data-clipboard="@Model.Address" data-clipboard-confirm=".clipboard-confirmation">
                            <input type="text" class="form-control" readonly="readonly" value="@Model.Address" id="address"/>
                                <button type="button" class="btn btn-outline-secondary clipboard-confirmation" data-clipboard-confirm="Copied ✔">Copy address</button>
                            </div>
                        </div>
```

dennisreimann: I'll have to massage the clipboard js afterwards though, found another bug/inconvenience there …

Kukks: got it, thanks!

dennisreimann (approved):
Nice, let's use this everywhere next 👍

BTCPayServer/Views/Wallets/WalletReceive.cshtml
```diff
@@ -47,14 +45,13 @@
...
-                            <img v-bind:src="srvModel.cryptoImage" class="qr-icon" />
-                            <qrcode v-bind:value="srvModel.address" :options="{ width: 256, margin: 1, color: {dark:'#000', light:'#f5f5f7'} }" tag="svg">
-                            </qrcode>
+                            <img src="@Model.CryptoImage" class="qr-icon"/>
+                            <vc:qr-code data="@Model.Address"/>
                         </div>
-                        <div class="input-group copy" data-clipboard-target="#vue-address">
-                            <input type="text" class=" form-control " readonly="readonly" :value="srvModel.address" id="vue-address" />
+                        <div class="input-group" data-clipboard="@Model.Address">
+                            <input type="text" class="form-control" style="cursor: copy" readonly="readonly" value="@Model.Address" id="address"/>
```
dennisreimann, line 52: Tests currently fail, because the ID changed from `vue-address` to `address`.

bolatovumar (approved):

NicolasDorier:
Concept ACK.  Keeping that for next version. We actually have lot's of things for this one already.

Aquiles00770:
bc1q3t45wyjtaamswptsdecq4gee50z0lng539wdvl

Issue #1711, opened 2020-07-02
Dashboard (replace homepage)


pavlenex:
**Is your feature request related to a problem? Please describe.**
Currently once you log in into btcpay server, you're greeted with a homepage that's not helpful at all. It's already been discussed that when the time is right, we should replace this homepage with an actual **dashboard**. There are a lot of issues to close before we can get into implementing this one, but I'm opening it for community feedback and ideas on what we can have on a dashboard.

### Homepage

![homepage](https://user-images.githubusercontent.com/36959754/86341360-7c7d0980-bc56-11ea-8238-fe2fcb3fadf5.PNG)

### Describe the solution you'd like
Replace the homepage with a dashboard. Dashboard should provide insights into analytics and important stuff going on on the server. Dashboard should be available for both server admins and regular users, so it should display data from the stores you're owner/member of only. For server admins we can show total number of stores and some analytics there.

### Dashboard example
Dashboard should not infringe the current header navigation of btcpay server. It needs to stay consistent with our design guidelines and provide minimum impact to the rest of the pages.

![dashboard example](https://user-images.githubusercontent.com/36959754/86341767-062cd700-bc57-11ea-9851-5857ff52adc9.PNG)

You can preview the dashboard from above [here](https://preview.pro.ant.design/dashboard/analysis).

One more thing a dashboard should do is perhaps guide user through setting up the store and a wallet properly and provide information on what needs to be done next.

### To-do

- [ ] Determine dashboard design (for designers)
- [ ] Determine important data we want to show
- [ ] Determine the CTA we want to show
- [ ] Determine how we'll show aggregate info for stores

I've been searching for a good example that'll fit into ours for quite a while and the one I outlined above was the best imo. If anybody else has good examples of dashboards, feel free to share them, but be aware they need to provide minimum disruption to the rest of the server.

dstrukt:
I'm working through this and will post some work when I've refined a bit.

bolatovumar:
I agree. I imagine it would be nice for merchants to have a dashboard. Also would be nice to be able to customized it by including different data and being able to add or remove stuff. This doesn't have to be in the first version of course.

pavlenex:
@bolatovumar besides merchant's use-case we can add some data regarding node info, uptime, etc. I think other type of users would like to have such data, but as you said, we can build on it later on. Curious to hear if others have suggestions on what kind of data we can include.

Kukks:
[...]

Issue #545, opened 2019-01-21
btcpayserver in powershell is looking for regtest as the datadirectory


mattnook:
I keep getting this error when trying to set up btcpayserver (Bitcoin Core and NBXplorer are working fine)

`PS C:\Users\user\btcpayserver\BTCPayServer> dotnet run
C:\Program Files\dotnet\sdk\2.2.103\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets(153,5): warning NETSDK1071: A PackageReference to 'Microsoft.AspNetCore.App' specified a Version of `2.1.6`. Specifying the version of this package is not recommended. For more information, see https://aka.ms/sdkimplicitrefs [C:\Users\user\btcpayserver\BTCPayServer\BTCPayServer.csproj]
[33m[1m  No files matched the pattern wwwroot/crowdfund-admin/**/*.js[0m[39m[49m
info: Configuration:  Data Directory: C:\Users\user\AppData\Roaming\BTCPayServer\RegTest
info: Configuration:  Configuration File: C:\Users\user\AppData\Roaming\BTCPayServer\RegTest\settings.config
info: Configuration:  Creating configuration file
info: Configuration:  Network: Regtest
info: Configuration:  Supported chains: BTC,LTC
info: Configuration:  Network: Regtest
info: Configuration:  Supported chains: BTC,LTC
info: Configuration:  Root Path: /
info: Configuration:  Postgres DB used (User ID=postgres;Host=127.0.0.1;Port=39372;Database=btcpayserver)
Application startup exception: System.Net.Sockets.SocketException (10061): No connection could be made because the target machine actively refused it`

I don't know what the issue is but it seems like the DataDirectory is trying to go to regtest rather than Main. How do I fix this? The " No files matched the pattern wwwroot/crowdfund-admin" also seems weird. I am using default settings in settings.config.

NicolasDorier:
I advise you against on setting up BTCPay on windows. 
BTCPayServer is using Regtest by default, and you can mainnet

The environment variables known by BTCPay are here https://github.com/btcpayserver/btcpayserver-docker#environment-variables if you want mainnet, set `NBITCOIN_NETWORK` correctly.

NicolasDorier:
Manual deployment is not cheaper. It is just more time consuming.
Please use https://www.youtube.com/watch?v=oce_fvRdLLY

[...]

PR #2127, opened 2020-12-11
Add links to API documentation for better discovery

NicolasDorier:
image

### DIFF
--- a/BTCPayServer/Views/Home/SwaggerDocs.cshtml
+++ b/BTCPayServer/Views/Home/SwaggerDocs.cshtml
@@ -22,6 +22,8 @@
     </style>
 </head>
 <body>
+    @*Ignore this, this is for making the test ClickOnAllSideMenus happy*@
+    <div class="navbar-brand" style="visibility:collapse;"></div>
     <redoc spec-url="@Url.ActionLink("Swagger")"></redoc>
     <script src="https://cdn.jsdelivr.net/npm/redoc@2.0.0-rc.45/bundles/redoc.standalone.js" integrity="sha384-RC31+q3tyqdcilXYaU++ii/FAByqeZ+sjKUHMJ8hMzIY5k4kzNqi4Ett88EZ/4lq" crossorigin="anonymous"></script>
 </body>
--- a/BTCPayServer/Views/Manage/APIKeys.cshtml
+++ b/BTCPayServer/Views/Manage/APIKeys.cshtml
@@ -4,54 +4,55 @@
     ViewData.SetActivePageAndTitle(ManageNavPages.APIKeys, "Manage your API Keys");
 }
 
-<partial name="_StatusMessage"/>
+<partial name="_StatusMessage" />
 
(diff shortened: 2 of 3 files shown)


Kukks (changes requested):
I don't think the side navigation menu item fits, the other one is good though.

NicolasDorier:
@Kukks please suggest something else then, the docs is currently hidden, that's worse.

Kukks:
> @Kukks please suggest something else then, the docs is currently hidden, that's worse.

Maybe in the footer next to version number and also in Integrations page in the store?

Kukks:
@pavlenex @dennisreimann any thoughts?

NicolasDorier:
Nobody look at there... Ping @dstruck

NicolasDorier:
Ping @dstrukt thoughts?

Kukks:
> Nobody look at there...

Yeah but arguably, nobody will look at the User navigation to find the link for API docs. I think maybe under store navigation would work better if anything

dennisreimann:
The way we do it everywhere else is with the help/question mark icons …

Bildschirmfoto 2020-12-11 um 09 08 30

Bildschirmfoto 2020-12-11 um 09 07 56

Bildschirmfoto 2020-12-11 um 09 08 13

If we just want to link to the docs, I think we should be conssistent here.

I don't fully get the issue though: Is it just that we need to provide the link or should we be offering more guidance/context for the users here? What sparked this issue?

pavlenex:
The way to fix this imo is:

1. Have a tooltip - ACK.
2. Once we have a dashboard #1711 make it more prominent, I am not a fan of a sidebar display for a link.
3. We can display Integrations and API better in our docs on a homepage [#545](https://github.com/btcpayserver/btcpayserver-doc/pull/545) once we figure out how to improve search algo in docs but also on our website.

dennisreimann:
So the quickfix would be to do it like on the Payment Request list (first screenshot): Keep the text and link to the docs via the icon and move the botton up to sit between the intro text and the list. :)

NicolasDorier:
@dennisreimann @pavlenex I don't like the tooltip, this normally link to user readable documentation, not API. This would be confusing.

dennisreimann:
But then again, what kind of problem are we trying to solve? Doesn't sound like just providing the link helps in that regard. I mean, we'd be throwing people to the API docs where they might feel lost too, because that's likely too technical.

pavlenex:
@NicolasDorier Can we just make it more prominent in the API keys section and then add the links to website and docs. I assume people are more likely to find docs that way?

NicolasDorier:
@dennisreimann the point we want to solve is that currently nobody know the Greenfield API exists, because there is no link to it.

dennisreimann:
Ok, but then let's motivate the usage a bit more and offer an introduction paragraph, that explains the API and it's use case and links to the relevant docs:

The [BTCPay Server Greenfield API](https://docs.btcpayserver.org/API/Greenfield/v1/) offers programmatic access to your instance. You can manage your bTCPay Server (e.g. stores, invoices, users) as well as automate workflows and integrations (see some [use case examples](https://docs.btcpayserver.org/GreenFieldExample/). For that you need the API keys, which can be generated here. Find more information in the [API authentication docs](https://docs.btcpayserver.org/API/Greenfield/v1/#section/Authentication).

NicolasDorier:
Removed from the navbar, added your text. We can improve the rest later.

dstrukt:
I missed this one.

Will take note to make this more prominent in places where it makes sense as well, dashboard, etc..

Agree with the approach @dennisreimann recommended for the time being.

Issue #1923, opened 2020-09-22
Add Bitpay invoice translator/deobfuscator link


Zaxounette:
**Enhancement request:** Add Bitpay invoice translator to UI/UX

It appears there is no link or mention of the bitpay translator/deobfuscator ( www.YOURBTCPAYINSTANCE.com/translate) in the UI/UX anywhere.
As briefly stated [here](https://github.com/btcpayserver/btcpayserver-doc/pull/725) in the comments it could be a good idea to add it somewhere.

Place it could be mentioned:
* `Server Settings` > `Services` > `Other External Services`

Open for discussion.

NicolasDorier:
Server settings is only accessible to admin, so that's not what we want.
There is several of such context free tooling I want to add (for example a transaction broadcaster).

Ping @dstrukt . I really wished we had somewhere to put those kind of tools.

dstrukt:
Thanks for tagging.

Will keep this is mind as I'm building out the remaining screens!

pavlenex:
Translator is great example of what should be a plugin.

Zaxounette:
> Translator is great example of what should be a plugin.

It is indeed a good idea !

Zaxounette:
#2130 being merged and translator being confirmed obsolete (lol) from what I understand, should this be closed ?

@pavlenex @NicolasDorier

pavlenex:
Yup, closing.

Zaxounette:
> Should be removed from docs as well if we have it, not 100% certain we do.

I'll look into that, boss ! <3

PR #2128, opened 2020-12-11
Document miscallenous features in the swagger doc

NicolasDorier:
We have some discovery issues with some hidden feature in BTCPay Server.
I decided to document them in the swagger docs

image


pavlenex:
@NicolasDorier Just a note BitPay translator would be a good use for a plugin as discussed here #1923, in that case I'm not sure if it makes sense to add it now, just a thought.

NicolasDorier:
we can remove it later if needed.
I am sad there are some hidden feature for now nobody can discover.

pavlenex:
BitPay doesn't really force BIP70 so Translator itself may not be that widely used. Alright once we make it a plugin hopefully we can remove it, just wanted to tie the two issues so we don't loose context there.

NicolasDorier:
if they do not force it we can remove it entirely then.

NicolasDorier:
Will remove it entirely after this PR then

PR #1625, opened 2020-06-02
Refactor the Network Provider to use DI to get networks

Kukks:
Still experimenting

PR #2129, opened 2020-12-11
Plugins flexibility PR

Kukks:
* Makes the BTCPayServerOptions.LoadArgs async to support Plugin hooks and actions
* relax the private set modifiers in the BTCPayNetwork
* Separate IPluginHookService from PluginService to reduce dependencies on DI
* fix some small bugs around image path
* Fix bug with new dbreeze migration where data dir was incorrect


BTCPayServer/Extensions.cs
```diff
@@ -371,9 +373,11 @@ public static Uri GetAbsoluteUriNoPathBase(this HttpRequest request, Uri relativ
 
         public static IServiceCollection ConfigureBTCPayServer(this IServiceCollection services, IConfiguration conf)
         {
-            services.Configure<BTCPayServerOptions>(o =>
+            services.AddSingleton(provider =>
             {
-                o.LoadArgs(conf);
+                var options = new BTCPayServerOptions();
+                options.LoadArgs(conf, provider.GetService<IPluginHookService>()).GetAwaiter().GetResult();
+                return options;
```
rockstardev, line 380: Sync call, expand with comment

BTCPayServer/Program.cs
```diff
@@ -29,7 +29,7 @@ static void Main(string[] args)
                 if (conf == null)
                     return;
                 Logs.Configure(loggerFactory);
-                new BTCPayServerOptions().LoadArgs(conf);
+                new BTCPayServerOptions().LoadArgs(conf).GetAwaiter().GetResult();
```
rockstardev, line 32: This call is just to detect configuration/syntax errors from BTCPayServer... if configuration file got corrupted you'll get error here and BTCPayServer stops here... kukkie saved you from advanced debugging (you know config is wrong).

rockstardev: This needed to change since `LoadArgs` became async task, I've just added comment to document why we previously had this call here.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -15,21 +15,17 @@
 
 namespace BTCPayServer.Plugins
 {
-    public class PluginService: IPluginHookService
```
rockstardev, line 18: NOTE: Making `PluginService` more cohesive, now it's in charge just of displaying plugin information. `ApplyAction` and `ApplyFilter` methods moved to dedicated `PluginHookService`. 👍

BTCPayServer/Plugins/PluginHookService.cs
```diff
@@ -0,0 +1,60 @@
...
+                    await pluginHookFilter.Execute(args);
+                }
+                catch (Exception e)
+                {
+                    _logger.LogError(e, $"Action on hook {hook} failed");
+                }
+            }
+        }
+
+        public async Task<object> ApplyFilter(string hook, object args)
```
rockstardev, line 41:
```suggestion
        // Trigger hook for plugins that optionally performs some operations on args and returns modified object back
        public async Task<object> ApplyFilter(string hook, object args)
```

BTCPayServer/Plugins/PluginHookService.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public PluginHookService(IEnumerable<IPluginHookAction> actions, IEnumerable<IPluginHookFilter> filters,
+            ILogger<PluginHookService> logger)
+        {
+            _actions = actions;
+            _filters = filters;
+            _logger = logger;
+        }
+
+        public async Task ApplyAction(string hook, object args)
```
rockstardev, line 24:
```suggestion
        // Trigger simple action hook for plugins
        public async Task ApplyAction(string hook, object args)
```

BTCPayServer/Plugins/PluginHookService.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                try
+                {
+                    await pluginHookFilter.Execute(args);
+                }
+                catch (Exception e)
+                {
+                    _logger.LogError(e, $"Action on hook {hook} failed");
+                }
+            }
```
rockstardev, line 38: @Kukks I was thinking that you can optionally make this method non-async... and then in `foreach` you would collect tasks, in the end doing:

`return Task.WhenAll(ListofPluginHookFilterTasksIfGreaterThan0);`

This would allow for fire-and-forget calls and caller to control await. I don't know if we want this in architecture... or if you want to implement fire-and-forget as separate method call. Just a suggestion.

BTCPayServer.Abstractions/Models/BaseBTCPayServerPlugin.cs
```diff
@@ -21,7 +21,6 @@ public virtual Version Version
 
         public abstract string Description { get; }
         public bool SystemPlugin { get; set; }
-        public bool SystemExtension { get; set; }
```
rockstardev, line 24: @Kukks, I assume `SystemPlugin` replaces `SystemExtension` and that's why you removed, correct?

rockstardev (approved):
Great PR, is already validated with usage by:
https://github.com/Kukks/btcpayserver/commit/fec41611a36d5453562a030fde68988c9663d551

Most of my comments are just notes for Emperor @NicolasDorier to make his review and merge easier.

The only "general improvement" comment I have is that we consider enum on `ApplyAction` and `ApplyFilter` - would be easier to find references on where certain hooks are being used and avoid typos.

Kukks:
> Great PR, is already validated with usage by:
> [Kukks@fec4161](https://github.com/Kukks/btcpayserver/commit/fec41611a36d5453562a030fde68988c9663d551)
> 
> Most of my comments are just notes for Emperor @NicolasDorier to make his review and merge easier.
> 
> The only "general improvement" comment I have is that we consider enum on `ApplyAction` and `ApplyFilter` - would be easier to find references on where certain hooks are being used and avoid typos.

I considered enum but decided against it because I wanted to allow plugins to define their own actions/hooks. If there was an enum, plugins would not be able to extend other plugins.

BTCPayServer/Extensions.cs
NicolasDorier: This is indeed contentious.
If you need async stuff, use IStartupTasks.
No `GetAwaiter().GeResult()`.

BTCPayServer/Program.cs
NicolasDorier: No `GetAwaiter().GetResult()`.

BTCPayServer.Abstractions/Models/BaseBTCPayServerPlugin.cs
Kukks: Correct

BTCPayServer/Extensions.cs
rockstardev: @NicolasDorier can you link to `IStartupTasks` and how they would be implemented? I've only found `IStartupFilter`, but that interface is also sync, so implementation would again end up being async-over-sync.

The only downside of this approach is is following conventions; it's impossible to end up with deadlocks since pipelines didn't already start.

So I would encourage you to give it a try implementing this... it's the only thing left with the PR and I think it's better you tailor this to how you think is best and merge, rather than going back and forth.

BTCPayServer/Program.cs
rockstardev: Will be resolved through: https://github.com/btcpayserver/btcpayserver/pull/2129#discussion_r546509565

rockstardev:
> I considered enum but decided against it because I wanted to allow plugins to define their own actions/hooks. If there was an enum, plugins would not be able to extend other plugins.

Makes sense, thanks for explaining. I think we may end up defining string `const`s eventually... but agree that for now sticking with pure strings is the best option.

BTCPayServer/Extensions.cs
Kukks: IStartupTask does not work. 
For reference it is this interface
```
 public interface IStartupTask
    {
        Task ExecuteAsync(CancellationToken cancellationToken = default);
    }
```

I NEED a way to:
* modify the network provider before it hits the DI
* use an async task to modify it


These issues are caused by the bad design around the `BTCPayNetworkProvider` which I've tried to solve for 2 years but each PR (which I worked for days on) was shot down.  https://github.com/btcpayserver/btcpayserver/pull/1625 https://github.com/btcpayserver/btcpayserver/pull/912

rockstardev: I thought it's something embedded in .NET Core... now I see @NicolasDorier pretty much implemented this pattern from: 
https://stackoverflow.com/a/63323207

@NicolasDorier I would merge this PR as is and then you take it and modify it the way you see fit. I think extracting this into `IStartupTask` would be worse - it would still work since `BTCPayServerOptions` is singleton - but `BTCPayServerOptions.NetworkProvider` would be modified in two places; kinda tangled.

NicolasDorier:
So your problem is probably solvable without hacking and without the need to break everything.

Solving the most immediate problem:

1. Load configuration in the StartupTask from the DB
2. In the config hooks, just use the data you loaded earlier.

This way no problem.


More generally, I believe we should use the IOption pattern of .netcore rather than having to invent our own stuff that only you will understand. On top of this, this would be useful for any plugin who want to override some settings.
Basically we would have a `DatabaseConfigurationProvider` (probably can find somewhere), this provider would have a `Task Load` function which would fetch the options from the DB, and store it in memory. We would call this in IStartupTask.

On top of this, the option pattern of .NET allow the possibility to reload a configuration on the fly.  (see `IOptionMonitor` and `IOptionSnapshot`)

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1

I don't see why the `BTCPayNetworkProvider` could not be depends on a `IOptionMonitor` and dynamically change the list of network inside.

Kukks:
The main issue is that a lot of the services use the initial list of the
networks and then use that for the lifetime of the app.

I cannot use the dB to store the config for plugins that need to configure
stuff as early as the network provider creation. This is because the dB
context factory depends on Database Options which depends on BTCPayServer
Options.

So my approach after failing there was to write to a json file, but I
cannot determine the data directory of BTCPay until BTCPayServer Options
starts loading data based on the network type and other properties it
needs.  Since I need to load the plugin data by knowing the data dir, I
created two plugin hooks:
* one announces the data directory of BTCPay as soon as it is conputed
* one provides the list of unfiltered networks to the plugin system for
mutations/transformations.

The Options pattern would be good for many scenarios, but not this one just
because of how early it is in the lifetime of the application and what is
needed for it.

Just FYI, this hook pattern is VERY widely used. Most of the plugins in
WordPress are based on this core concept and a lot of other plugin systems
use the exact same architecture for extensibility

NicolasDorier:
I am not complaining about the hooks, but about the hack with the LoadArgs making it async, which should be fixable.
I'll take a look in a few hours

rockstardev:
Yeah... I'm with Kukkie on this one. You can't solve this with refactoring `BTCPayServerOptions` because call to hooks is async. So as long as you're having hooks modifying Network Providers... `LoadArgs` will need to be async.

Using `StartupTask` won't work because by the time it hits, all the other classes are already initialized within DI pipeline with Network Providers that were not touched by plugins = no go.

The only viable alternative (if you don't want to do what Kukks did in this PR) is that `BTCPayServerOptions` is initialized + plugins async called... and then that initialized instance is passed into DI pipeline. You'll still be doing sync-over-async, just not within DI pipeline. In that context your refactoring will be useful since it'll make `BTCPayServerOptions` easier to initialize without DI.

... 

One crazy alternative would be that plugins declare `object ApplyFilterSYNC(string hook, object args)` and then you call that instead of async version in situations where you're forced to be sync (DI pipeline). I would recommend this approach only if you're true rockstardev 😂 .

NicolasDorier:
@rockstardev this hack is specifically done because the BTCPayNetworkProvider is tied to the BTCPayServerOptions. This itself does not have to be that way.

@Kukks can you tell me if the commit where I decoupled things kind of fix this problem?
If so, as I said, the next step is to remove the dependency of BTCpaySErverOptions to BTCPayNetworkProvider, which I think can be done easily.

rockstardev:
@NicolasDorier it's this line:
https://github.com/btcpayserver/btcpayserver/pull/2129/files#diff-1a573e46b5ae9d0055313df6918cf8dd976b822c49de23b9fed78048f41dad0cR100

And sure, if you extract `BTCPayNetworkProvider`, then that class will need to be:
  1. Initialized before DI pipeline with sync-over-async and passed in
  2. Or moved to IStartupTask 
  
I think approach #1 is better - more cohesive code; no splitting of initialization.

Your decoupling PR can help with the other call: `await pluginHookService?.ApplyAction("datadir-configured", DataDir);`, here:
https://github.com/btcpayserver/btcpayserver/pull/2129/files#diff-1a573e46b5ae9d0055313df6918cf8dd976b822c49de23b9fed78048f41dad0cR86

Same situation as with `BTCPayNetworkProvider` though - you need to make a choice between approach 1 or 2.

Kukks:
I remove all the stuff we didn't agree on. I will merge because there is a bug fix that needs to be merged before any release is done.

PR #2145, opened 2020-12-16
Use NBitcoin.BIP78 for Payjoin

Kukks:
Refactoring payjoin to use a separate library. This will prep the receiver side to enable batching and substitution capabilities in the future


NicolasDorier:
don't break everything in 1 PR.
Preferrable approach would be to just focus on putting the client in a library in a first run, leaving all the rest as is.

Also, I would prefer if this library for now was under BTCPayServer org, like BTCPayServer.BIP78.

NicolasDorier:
tests fail

Kukks:
Will reopen when the sender part is merged in

MaiHistoryCompanyLimited:
{#2145}

Kukks:
Will revisit this later

PR #2151, opened 2020-12-18
UI improvements

dennisreimann:
Various things I came across, mostly self-explanatory from the commit messages. 

The invoice selection on the list was something I was missing and I think it makes ticking those checkboxes easier. Refactored the JS for that a bit.


Kukks (approved):
naice

Kukks (changes requested):
This PR breaks the payjoin e2e test because you changed an id
test code
`    var paymentValueRowColumn = s.Driver.FindElement(By.Id($"invoice_{invoiceId}"))
                        .FindElement(By.ClassName("payment-value"));`

Kukks (approved):

PR #2153, opened 2020-12-19
Rename UiExtensionPoint folder to fix build crash

pavlenex:
When building on Ubuntu I got the following error:

```
System.InvalidOperationException: The view 'Components/UiExtensionPoint/Default' was not found. The following locations were searched:
/Views/Home/Components/UiExtensionPoint/Default.cshtml
/Views/Shared/Components/UiExtensionPoint/Default.cshtml
/Pages/Shared/Components/UiExtensionPoint/Default.cshtml
/Components/UiExtensionPoint/Default.cshtml
   at Microsoft.AspNetCore.Mvc.ViewEngines.ViewEngineResult.EnsureSuccessful(IEnumerable`1 originalLocations)
   at Microsoft.AspNetCore.Mvc.ViewComponents.ViewViewComponentResult.ExecuteAsync(ViewComponentContext context)
   at Microsoft.AspNetCore.Mvc.ViewComponents.DefaultViewComponentInvoker.InvokeAsync(ViewComponentContext context)
   at Microsoft.AspNetCore.Mvc.ViewComponents.DefaultViewComponentHelper.InvokeCoreAsync(ViewComponentDescriptor descriptor, Object arguments)
   at AspNetCore.Views_Shared__Layout.__Generated__UiExtensionPointViewComponentTagHelper.ProcessAsync(TagHelperContext __context, TagHelperOutput __output)
   at Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperRunner.<RunAsync>g__Awaited|0_0(Task task, TagHelperExecutionContext executionContext, Int32 i, Int32 count)
   at AspNetCore.Views_Shared__Layout.<ExecuteAsync>b__50_1() in /home/pavle/Documents/GitHub/btcpayserver/BTCPayServer/Views/Shared/_Layout.cshtml:line 28
   at Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperExecutionContext.SetOutputContentAsync()
   at AspNetCore.Views_Shared__Layout.ExecuteAsync() in /home/pavle/Documents/GitHub/btcpayserver/BTCPayServer/Views/Shared/_Layout.cshtml:line 8
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageCoreAsync(IRazorPage page, ViewContext context)
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageAsync(IRazorPage page, ViewContext context, Boolean invokeViewStarts)
[...]

PR #2154, opened 2020-12-19
rename UIExtensionPoint folder

pavlenex:
Rename to avoid build crashes on Ubuntu.

Finally fixed my dev env so this is continuation of #2153

I've tested this on Ubuntu 20.04 and it works. Would be cool if someone with MacOS or Windows can just try to build and run this PR so we're sure it doesn't break those environments.


bolatovumar (approved):

bolatovumar (approved):
Built this branch on MacOS Catalina and encountered no errors.

Kukks (approved):

PR #2157, opened 2020-12-20
GreenField: Store OnChain Payment Method

Kukks:


BTCPayServer.Client/Models/OnChainPaymentMethodData.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Crypto code of the payment method
+        /// </summary>
+        public string CryptoCode { get; set; }
+
+        /// <summary>
+        /// The derivation scheme
+        /// </summary>
+        public string DerivationScheme { get; set; }
```
NicolasDorier, line 23: Missing KeyPath, Label, and Fingerprint. (Can combine Fingerprint and KeyPath in one field `RootedKeyPath` though, even if in theory they can be set apart, this is a legacy concern)

Can do in followup PR

NicolasDorier:
Seems good to me.  I think we also need to add a route later to do what the "Add a new/existing seed" option of the UI is doing.

PR #2158, opened 2020-12-21
Use library for Payjoin Sender

Kukks:
Also moved the lib to https://github.com/btcpayserver/NBitcoin.BIP78


BTCPayServer/Payments/PayJoin/Sender/PayjoinServerCommunicator.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using BIP78.Sender;
+using NBitcoin;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace BTCPayServer.Payments.PayJoin.Sender
+{
+    public class PayjoinServerCommunicator : IPayjoinServerCommunicator
```
NicolasDorier, line 13: Most of this logic is not specific to btcpayserver. And most of this logic is part of the protocol itself.

The only specific part is `CreateHttpClient`. So I think this class should actually be in the client library, with a way to override `CreateHttpClient`.

NicolasDorier:
Concept ACK though:

1. My point about `PayjoinServerCommunicator` should be addressed, different PR is fine.
2. I renamed the library to BTCPayServer.BIP78
3. The library itself must have better documentation in README.md, and more importantly: Tests

The 3. is important, because as soon as we separate both codebase, it becomes a pain in the ass to debug, and if we don't have a test environment on the project itself then it will just make maintenance more painful.

BTCPayServer/Payments/PayJoin/Sender/PayjoinServerCommunicator.cs
Kukks: I did this on purpose. For example, Wasabi has its own class to do these calls which are not reliant on HttpClient.

The payjoin spec also allows to do communication over non HTTP methods ( I am prototyping a qr only mode for example) . If we forced this, this lib would become useless.

Kukks:
> Concept ACK though:
> 
> 1. My point about `PayjoinServerCommunicator` should be addressed, different PR is fine.
> 2. I renamed the library to BTCPayServer.BIP78
> 3. The library itself must have better documentation in README.md, and more importantly: Tests
> 
> The 3. is important, because as soon as we separate both codebase, it becomes a pain in the ass to debug, and if we don't have a test environment on the project itself then it will just make maintenance more painful.

Yes I did not document or port over the tests to avoid wasting time if I did not get the ACK here. Will do it after this

BTCPayServer/Payments/PayJoin/PayJoinExtensions.cs
```diff
@@ -14,12 +16,10 @@ public static void AddPayJoinServices(this IServiceCollection services)
             services.AddSingleton<HostedServices.Socks5HttpProxyServer>();
             services.AddSingleton<IHostedService, HostedServices.Socks5HttpProxyServer>(s => s.GetRequiredService<Socks5HttpProxyServer>());
             services.AddSingleton<PayJoinRepository>();
+            services.AddSingleton<IPayjoinServerCommunicator, PayjoinServerCommunicator>();
             services.AddSingleton<PayjoinClient>();
             services.AddTransient<Socks5HttpClientHandler>();
-            services.AddHttpClient(PayjoinClient.PayjoinOnionNamedClient)
-                .ConfigureHttpClient(h => h.DefaultRequestHeaders.ConnectionClose = true)
```
NicolasDorier, line 20: don't remove `h.DefaultRequestHeaders.ConnectionClose `  this will break things...

BTCPayServer/Payments/PayJoin/PayJoinExtensions.cs
```diff
@@ -14,12 +16,10 @@ public static void AddPayJoinServices(this IServiceCollection services)
...
-                .ConfigurePrimaryHttpMessageHandler<Socks5HttpClientHandler>();
-            services.AddHttpClient(WebhookNotificationManager.OnionNamedClient)
```
NicolasDorier, line 22: why you removed `WebhookNotificationManager.OnionNamedClient` ?

Kukks: I didn't, only moved it outsidemof the Payjoin Extensions since it did not fit

NicolasDorier:
I think there is a problem in the PR, tests fail, tried to run again twice.

NicolasDorier:
maybe unrelated one sec.

NicolasDorier:
I think something is wrong with `CanUseCoinSelection ` in this PR.

NicolasDorier:
Mhh does not seem to be caused by this PR, merging this.

PR #2160, opened 2020-12-21
Fix bug where altcoins were not added

Sword-Smith:


Kukks (changes requested):
This is not correct.  We use `filtered` to determine if sub assets of those altcoins should be included ONLY if the main network coin is added through config/

Sword-Smith:
> 
> 
> This is not correct. We use `filtered` to determine if sub assets of those altcoins should be included ONLY if the main network coin is added through config/

Thanks for the reply. But doesn't the `InitEthereum()` function contain the configuration? Is there a README somewhere where I can see how to enable Ethereum payments on BTCPayServer?

Kukks:
> > This is not correct. We use `filtered` to determine if sub assets of those altcoins should be included ONLY if the main network coin is added through config/
> 
> Thanks for the reply. But doesn't the `InitEthereum()` function contain the configuration? Is there a README somewhere where I can see how to enable Ethereum payments on BTCPayServer?

https://docs.btcpayserver.org/FAQ/FAQ-Altcoin/#how-to-add-an-altcoin-in-btcpay
Altcoins are not officially supported by us. Their respective communities should document their altcoins setup and integration.

PR #2166, opened 2020-12-23
Add Label and accoutKeyPath to the on chain payment methods

NicolasDorier:
This add a way to specify some fields of the on-chain payment methods such as `Label` and `AccountKeyPath`.

I wanted to use non-primitive (`RootedKeyPath`) types to make the greenfield library easier to use.
However, this caused problem as the default `ObjectModelValidator` would try to visit all the members of the `RootKeyPath` to validate everything. But those were sending back exception.

So I create a new `SkippableObjectModelValidator` which is like the `DefaultObjectModelValidator` of MVC, except that it is possible to register types to skip validations.

I initially wanted to remove all validations, but we rely on it for the UI and there is no way to allow it only for UI and not for API.

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreOnChainPaymentMethodsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreOnChainPaymentMethodsController.cs
@@ -10,6 +10,7 @@ using BTCPayServer.Services.Stores;
 using BTCPayServer.Services.Wallets;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using NBitcoin;
 using NBXplorer.DerivationStrategy;
 using StoreData = BTCPayServer.Data.StoreData;
 
@@ -203,6 +204,18 @@ namespace BTCPayServer.Controllers.GreenField
                 var strategy = DerivationSchemeSettings.Parse(paymentMethodData.DerivationScheme, network);
                 if (strategy != null)
                     await wallet.TrackAsync(strategy.AccountDerivation);
+                strategy.Label = paymentMethodData.Label;
+                var signing = strategy.GetSigningAccountKeySettings();
+                if (paymentMethodData.AccountKeyPath is RootedKeyPath r)
+                {
+                    signing.AccountKeyPath = r.KeyPath;
+                    signing.RootFingerprint = r.MasterFingerprint;
+                }
+                else
+                {
+                    signing.AccountKeyPath = null;
+                    signing.RootFingerprint = null;
+                }
                 store.SetSupportedPaymentMethod(id, strategy);
                 storeBlob.SetExcluded(id, !paymentMethodData.Enabled);
                 store.SetStoreBlob(storeBlob);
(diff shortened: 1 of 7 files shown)


Kukks:
Are you sure about the label? I think it should be a read only property due to the deriv scheme parser converts the xpub input and the original becomes the label

NicolasDorier:
@Kukks this is not readonly. What you refer is the default behavior, but you can set any label in the wallet settings.

Kukks:
> @Kukks this is not readonly. What you refer is the default behavior, but you can set any label in the wallet settings.

TIL

BTCPayServer/Hosting/SkippableObjectValidatorProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.Options;
+using NBitcoin;
+
+namespace BTCPayServer.Hosting
+{
+    public class SkippableObjectValidatorProvider : ObjectModelValidator
```
Kukks, line 13: This is some hardcore stuff

Kukks (changes requested):
Maybe it will simplify the code a lot to simply do the key path parsing and validation inside the controller action instead of the NBitcoin Json converter? It is a lot of hoops to jump through in mvc with this approach

BTCPayServer/Hosting/SkippableObjectValidatorProvider.cs
NicolasDorier: Well this is possible, but this make the library a pain in the ass to use.
To be honest, I am not really fan of hardcore code like that either. But having to do work for parsing in the controller while we can get NewtonsoftJson do the work for us, making our life easier AND easier to use the client library.

I think for the other actions we'll do (create a hot wallet or querying UTXOs), we'll want to reuse this.

Kukks (approved):
Ok makes sense then.

BTCPayServer/Controllers/GreenField/StoreOnChainPaymentMethodsController.cs
```diff
@@ -203,6 +204,18 @@ public ActionResult<OnChainPaymentMethodData> GetOnChainPaymentMethod(string cry
                 var strategy = DerivationSchemeSettings.Parse(paymentMethodData.DerivationScheme, network);
                 if (strategy != null)
                     await wallet.TrackAsync(strategy.AccountDerivation);
+                strategy.Label = paymentMethodData.Label;
```
Kukks, line 207:
```suggestion
                if(!string.IsNullOrEmpty(paymentMethodData.Label)
                {
                    strategy.Label = paymentMethodData.Label;
                }
```

Kukks (changes requested):
Maybe this last change on the label setting actually?

BTCPayServer/Controllers/GreenField/StoreOnChainPaymentMethodsController.cs
NicolasDorier: what I did before, but actually you want to set to empty or null if you want to unset the label.

PR #2168, opened 2020-12-25
Remove DB and Directory settings out of the BTCPayServerOptions

NicolasDorier:
Decoupling datadirectories and DB from BTCPayServerOptions.

This will make both usable before the BTCPayServerOptions get instantiated. (Say, in the startup)

Needed to prevent LoadArgs hacks of 
https://github.com/btcpayserver/btcpayserver/pull/2129

### DIFF
--- a/BTCPayServer.Abstractions/Models/DatabaseOptions.cs
+++ b/BTCPayServer.Abstractions/Models/DatabaseOptions.cs
@@ -1,11 +1,41 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices.ComTypes;
+using Microsoft.Extensions.Configuration;
+
 namespace BTCPayServer.Abstractions.Models
 {
     public class DatabaseOptions
     {
-        public DatabaseOptions(DatabaseType type, string connString)
+        public DatabaseOptions(IConfiguration conf, string dataDir)
         {
-            DatabaseType = type;
-            ConnectionString = connString;
+            var postgresConnectionString = conf["postgres"];
+            var mySQLConnectionString = conf["mysql"];
+            var sqliteFileName = conf["sqlitefile"];
+
+            if (!string.IsNullOrEmpty(postgresConnectionString))
+            {
+                DatabaseType = DatabaseType.Postgres;
+                ConnectionString = postgresConnectionString;
+            }
+            else if (!string.IsNullOrEmpty(mySQLConnectionString))
+            {
+                DatabaseType = DatabaseType.MySQL;
+                ConnectionString = mySQLConnectionString;
+            }
+            else if (!string.IsNullOrEmpty(sqliteFileName))
+            {
+                var connStr = "Data Source=" + (Path.IsPathRooted(sqliteFileName)
+                    ? sqliteFileName
+                    : Path.Combine(dataDir, sqliteFileName));
+
+                DatabaseType = DatabaseType.Sqlite;
+                ConnectionString = sqliteFileName;
+            }
+            else
+            {
+                throw new InvalidOperationException("No database option was configured.");
+            }
         }
 
         public DatabaseType DatabaseType { get; set; }
--- a/BTCPayServer/Hosting/BTCPayServerServices.cs
+++ b/BTCPayServer/Hosting/BTCPayServerServices.cs
@@ -1,5 +1,7 @@
 using System;
 using System.IO;
+using System.Runtime.InteropServices.ComTypes;
+using System.Security.Cryptography;
 using System.Threading;
 using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Abstractions.Extensions;
@@ -65,6 +67,7 @@ namespace BTCPayServer.Hosting
         }
         public static IServiceCollection AddBTCPayServer(this IServiceCollection services, IConfiguration configuration)
         {
+            services.AddSingleton<DataDirectories>();
             services.AddSingleton<MvcNewtonsoftJsonOptions>(o => o.GetRequiredService<IOptions<MvcNewtonsoftJsonOptions>>().Value);
             services.AddDbContext<ApplicationDbContext>((provider, o) =>
             {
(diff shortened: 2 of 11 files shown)


BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,11 +1,41 @@
...
     public class DatabaseOptions
     {
-        public DatabaseOptions(DatabaseType type, string connString)
+        public DatabaseOptions(IConfiguration conf, string dataDir)
         {
-            DatabaseType = type;
-            ConnectionString = connString;
+            var postgresConnectionString = conf["postgres"];
+            var mySQLConnectionString = conf["mysql"];
+            var sQLiteFileName = conf["sqlitefile"];
```
rockstardev, line 14: `sqliteFileName`

BTCPayServer/Hosting/Startup.cs
```diff
@@ -48,7 +48,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddMemoryCache();
             services.AddDataProtection()
                 .SetApplicationName("BTCPay Server")
-                .PersistKeysToFileSystem(GetDataDir());
+                .PersistKeysToFileSystem(new DirectoryInfo(new DataDirectories(Configuration).DataDir));
```
Kukks, line 51: We should create a single instance of `DataDirectories` early on in `ConfigureServices` and pass that along everywhere (and maybe even add it to DI since it should not be changing)?

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -111,32 +113,33 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
...
-                }
-                else if (!string.IsNullOrEmpty(opts.SQLiteFileName))
-                {
-                    var connStr = "Data Source=" +(Path.IsPathRooted(opts.SQLiteFileName)
-                        ? opts.SQLiteFileName
-                        : Path.Combine(opts.DataDir, opts.SQLiteFileName));
-                    Logs.Configuration.LogInformation($"SQLite DB used");
-                    Logs.Configuration.LogWarning("SQLite is not widely tested and should be considered experimental, we advise you to use postgres instead.");
-                    return  new DatabaseOptions(DatabaseType.Sqlite, connStr);
+                    var dbOptions = new DatabaseOptions(o.GetRequiredService<IConfiguration>(),
```
Kukks, line 122: We can move all of this into the `DatabaseOptions` ctor now

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,11 +1,41 @@
...
+            {
+                DatabaseType = DatabaseType.Postgres;
+                ConnectionString = postgresConnectionString;
+            }
+            else if (!string.IsNullOrEmpty(mySQLConnectionString))
+            {
+                DatabaseType = DatabaseType.MySQL;
+                ConnectionString = mySQLConnectionString;
+            }
+            else if (!string.IsNullOrEmpty(sQLiteFileName))
```
Kukks, line 26:
```suggestion
            else if (!string.IsNullOrEmpty(sqliteFileName))
```

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,11 +1,41 @@
...
+            {
+                var connStr = "Data Source=" + (Path.IsPathRooted(sQLiteFileName)
```
Kukks, line 28:
```suggestion
                var connStr = "Data Source=" + (Path.IsPathRooted(sqliteFileName)
```

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,11 +1,41 @@
...
+                    ? sQLiteFileName
+                    : Path.Combine(dataDir, sQLiteFileName));
```
Kukks, line 30:
```suggestion
                    : Path.Combine(dataDir, sqliteFileName));
```

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,11 +1,41 @@
...
+
+                DatabaseType = DatabaseType.Sqlite;
+                ConnectionString = sQLiteFileName;
```
Kukks, line 33:
```suggestion
                ConnectionString = sqliteFileName;
```

BTCPayServer/Plugins/PluginManager.cs
```diff
@@ -27,7 +28,7 @@ public static class PluginManager
             IConfiguration config, ILoggerFactory loggerFactory)
         {
             _logger = loggerFactory.CreateLogger(typeof(PluginManager));
-            var pluginsFolder = config.GetPluginDir(DefaultConfiguration.GetNetworkType(config));
+            var pluginsFolder = new DataDirectories(config).PluginDir;
```
Kukks, line 31: We should create a single instance of DataDirectories early on in ConfigureServices and pass that along everywhere (and maybe even add it to DI since it should not be changing)?

Kukks (changes requested):
Concept ACK, just small requests from my end

BTCPayServer/Hosting/BTCPayServerServices.cs
NicolasDorier: no, because DatabaseOptions don't have access to Logs singleton.

BTCPayServer/Plugins/PluginManager.cs
NicolasDorier: We can add it to DI yes.
As for creating it in `ConfigureServices` I am rather against.
Long time ago, I think I tried to do this for the BTCpayServerOptions, ended up working against the framework, which made life eventually harder.

NicolasDorier: @Kukks the problem would be solved if plugin loading happened during the IStartupTask.

NicolasDorier:
I am injected the `DataDirectories`.
We can probably manage to decrease the role of BTCPayServerOptions over time.
I think that the BTCPayNetworkProvider might be entirely independant rather than being tied to it.

PR #2172, opened 2020-12-28
Refactoring db entities to follow conventions

rockstardev:
- Properties now declared on a single line
 - Removed unnecessary usings
 - Reorganized `ApplicationDbContext` and moved it with `ApplicationDbContextFactory` to BTCPayServer.Data root
 - Refreshed schema migration to have new `InvoiceSearches` table follow established conventions... `Data` suffix shouldn't go in the name of the table, but it's to signal class is DB entity. ~~If you want @NicolasDorier I can remove migration from #2029 and then squash this into single migration (since we still haven't deployed changes to production).~~


rockstardev:
@NicolasDorier seems squashing migrations is the only viable option:

image

https://app.circleci.com/pipelines/github/btcpayserver/btcpayserver/4449/workflows/f9a7b534-672f-4f55-b5ab-09f0b7f84352/jobs/12501

NicolasDorier:
@rockstardev the problem is about the latest search table? If that's the case, just squash the migrations. We did not released yet.

NicolasDorier:
Remove the `20201228225040_AddingInvoiceSearchesTable.Designer.cs`, move the attribute of the partial class to the migration file.

rockstardev:
I did it dear Emperor!

PR #2174, opened 2020-12-30
Add API Key QR scan + click to reveal api key in List view

Kukks:
I did not make use of the new BCUR animated QR standard since this is BTCPay specific and the data being set is of minimal length but I can easily switch it over.

@21isenough 
image

### DIFF
--- a/BTCPayServer/Views/Shared/ShowQR.cshtml
+++ b/BTCPayServer/Views/Shared/ShowQR.cshtml
@@ -39,7 +39,8 @@
  			data: data,
  			fragments: [],
  			active: false,
- 			modalId: modalId
+ 			modalId: modalId,
+ 			useBCUR: true
  		},
  		computed:
  		{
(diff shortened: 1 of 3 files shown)


21isenough:
Awesome :smiley: :smiling_face_with_three_hearts:

Kukks:
> Awesome 😃 🥰

Do you prefer this or a BCUR encoded QR code?

21isenough:
No, this is great. As there is not much data, this should suffice. 

Am I correct that user and password would only be needed for the "Basic HTTP authentication ", right? 
Once I have the API key acquired the key and the URL are enough, I assume. So this is perfect :+1:

Kukks:
> No, this is great. As there is not much data, this should suffice.
> 
> Am I correct that user and password would only be needed for the "Basic HTTP authentication ", right?
> Once I have the API key acquired the key and the URL are enough, I assume. So this is perfect 👍

Yeah, correct. There is also an endpoint to check that the api key you are using has sufficient permissions for you

BTCPayServer/Views/Shared/ShowQR.cshtml
```diff
@@ -78,7 +79,11 @@
  					this.fragments = [];
  					return;
  				}
- 				this.fragments = window.bcur.encodeUR(this.data.toString(), 200);
+ 				if (this.useBCUR){
```
NicolasDorier, line 82: what is BCUR?

Kukks: BCUR is the spec used for animated QR codes (we use it for psbt, wallet import, wallet export) 
https://github.com/BlockchainCommons/bc-ur

Not used in this PR

PR #2179, opened 2021-01-03
More Options refactoring

Kukks:
Continues refactoring config classes to use the proper Options pattern where possible.
DataDirectories and DatabaseOptions are now configured the Options pattern and the BTCPayOptions is now moved alongside the other config setup


BTCPayServer/Extensions.cs
```diff
@@ -497,6 +489,17 @@ public static BTCPayNetworkProvider ConfigureNetworkProvider(this IConfiguration
             return result;
         }
 
+        public static DataDirectories Configure(this DataDirectories dataDirectories, IConfiguration configuration)
```
Kukks, line 492: This is not inside the DataDirectories class due to having dependencies that don't make sense in the Abstractions project

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
```diff
@@ -1,17 +1,21 @@
...
 
 namespace BTCPayServer.Abstractions.Models
 {
     public class DatabaseOptions
     {
-        public DatabaseOptions(IConfiguration conf, string dataDir)
+        public DatabaseType DatabaseType { get; set; }
+        public string ConnectionString { get; set; }
+
+        public void Configure(IConfiguration configuration, IOptions<DataDirectories> dataDirectories)
```
NicolasDorier, line 14: why you changed that? the constructor is better...

BTCPayServer/Extensions.cs
NicolasDorier: Just use a constructor... for the dependency, it is only `Storage.Services.Providers.FileSystemStorage.FileSystemFileProviderService.LocalStorageDirectoryName`, just hardcode it and we are done with it. (or move the constant in DataDirectories)

BTCPayServer.Abstractions/Models/DatabaseOptions.cs
Kukks: So that we actually use the IOptions pattern consistently.

NicolasDorier: ok one sec

BTCPayServer/Extensions.cs
Kukks: Constructor means you cant use IOptions pattern. You need to allow it to be created via empty ctor by the DI and then you use the `Configure()` to fill it

NicolasDorier:
So as discussed in Mattermost, Concept ACK, by looking at what .NET do, this indeed follow their patterns.
The only thing to change: Don't put the configure inside the option's class, you should call configure directly in the `AddBTCPayServer`. I don't think we need to reuse anywhere, so I think you can just put there directly.

Issue #2195, opened 2021-01-13
Update invoice endpoint fails to update metadata field


elvece:
**Bug**:
PUT to [update an invoice](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Invoices_UpdateInvoice) does not update the metadata field with the http request body. The original data in metadata is persisted. 

Response is a 200. 
Content-type is application/json
Example body:

`{
    "metadata":  {
        "POSData": 12
    }
}`

**Expected Behavior**:
PUT request to update an invoice should properly update the metadata field, with either POSData populated, but ideally any arbitrary object field. 

**Server Details**:
- BTCPay Server Version: v1.0.6.5
- Deployment Method: Docker
- Browser: Postman

saliehendricks:
Hi @elvece 

I just tested with my fresh dev instance and updated the invoice without issue. I get a 200 response with the invoice returned and metadata object updated.

Perhaps try with an API with higher privileges? I am using an admin Auth key.

S

elvece:
Hi @saliehendricks - the returned data in the 200 response does show the metadata as "updated", but the value in the UI is unchanged. I will try with a more privileged API to gather more information, but I do not think this should be necessary since I would like to use the restricted API permission (`btcpay.store.canmodifystoresettings`) with my implementation.

saliehendricks:
Hi @elvece , I see what you mean now. After the PUT, I do a GET and the Metadata is different. Will do some digging.

pavlenex:
Fixed in #2197

PR #2197, opened 2021-01-15
2195 Invoice metadata now saves correctly

saliehendricks:
#2195 Fixes metadata not saving due to 2 separate db contexts being used.

Calling invoice PUT api call:
/api/v1/stores/{storeid}/invoices/{invoiceid}
{
    "Metadata": {
        "PosData": 333
    }
}

Should result in the UI updating immediately:
image

Should also show the result in the API GET call:
/api/v1/stores/{storeid}/invoices/{invoiceid}
(this looked like it was working simply due to the fact that the model was updated and returned to the API caller)

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceRepository.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceRepository.cs
@@ -413,7 +413,7 @@ namespace BTCPayServer.Services.Invoices
         {
             using (var context = _ContextFactory.CreateContext())
             {
-                var invoiceData = await GetInvoiceRaw(invoiceId);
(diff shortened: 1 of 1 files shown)


Kukks (approved):
LGTM, great job!

NicolasDorier:
There was no test on that? shit

Kukks:
> There was no test on that? shit

There was but the test needed additional checks

saliehendricks:
The test is and was likely passing because the controller just passes back the metadata object that was sent in. There was no integration test that checks the crud on update invoice that I can see.

NicolasDorier:
`Assert.True(a == a)`

PR #2208, opened 2021-01-18
GreenField API: Configure Store Lightning Payment Method v2

Kukks:


BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
```diff
@@ -0,0 +1,246 @@
...
+            {
+                ModelState.AddModelError(nameof(LightningNetworkPaymentMethodData.ConnectionString),
+                    "Missing connectionString");
+            }
+
+            if (!ModelState.IsValid)
+                return this.CreateValidationError(ModelState);
+
+
+            PaymentMethodId paymentMethodId = new PaymentMethodId(network.CryptoCode, PaymentTypes.LightningLike);
```
NicolasDorier, line 124: Reuse the `var id` above.

NicolasDorier:
The `/internal` is really confusing. Can you keep it out of this PR?

Also there is a big vulnerability: If an admin share an API key with modify store, they are giving access to the internal node.
This is not what is intended. Instead, you should look for `canuseinternallightningnode` as we did for other APIs.

The CanUseInternalNode should check if the passed API Key has `canuseinternallightningnode` permission using the Authorization manager (see how I did for the register user).

Kukks:
> The `/internal` is really confusing. Can you keep it out of this PR?
> 
> Also there is a big vulnerability: If an admin share an API key with modify store, they are giving access to the internal node.
> This is not what is intended. Instead, you should look for `canuseinternallightningnode` as we did for other APIs.
> 
> The CanUseInternalNode should check if the passed API Key has `canuseinternallightningnode` permission using the Authorization manager (see how I did for the register user).
> 
> So in summary two things:
> 
> 1. I would like to keep the `/internal` routes out for now, as this is confusing.
> 2. The CanUseInternalNode must using the Authorization Manager to check if there is permission to use internal node.

Done

NicolasDorier:
test fail

NicolasDorier:
Can you have a test that actually pass through the `CanUseInternalNode` ?

Kukks:
> Can you have a test that actually pass through the `CanUseInternalNode` ?
 The last few asserts of the test are specifically around the `CanUseInternalNode`

Issue #1978, opened 2020-10-13
Improve reset password email 


britttttk:
**Is your feature request related to a problem? Please describe.**
Reset password email is very basic copy and looks almost like a phishing email. Users are not able to change this copy (afaik). I think we should try to enhance the copy for users and make it more apparent it is an email from a BTCPay Server.

Currently all it says: 
Please reset your password by clicking here: **link** 

**Describe the solution you'd like**
Now that there is the invite new users link which will send the "reset password" email to this user, I think we should make some copy changes to this email to make it more general and apply to both cases:

- user is resetting password
- user is creating password for the first time

**Additional context**
Possibly something like this: using the word *update* instead of *reset*

Subject: Update Password

A request has been made for you to update the password for your BTCPay Server account. Please click here to update your password. 

Open to any ideas...

brainharrington:
Allowing users to customize the email would be awesome. What I'm already doing manually is basically writing:

Hi <Name>,

Just set up your account for you on BTC Pay Server. Go ahead and click the link below to finish setting up your password.

Eskyee:
Subject: Update Password

A request has been made for you to update the password for your BTCPay Server account. Please click here to update your password.

a better improvement for emails needed, so it doesn't look  like spam, 
I agree with @britttttk 100%

bolatovumar:
> Now that there is the invite new users link which will send the "reset password" email to this user, I think we should make some copy changes to this email to make it more general and apply to both cases:

I wonder if it would make sense to send a different e-mail for when a user is invited vs when a user explicitly requests a password reset. I haven't really looked into this issue so not sure how feasible this solution is.

PresidentNick:
@bolatovumar https://github.com/btcpayserver/btcpayserver/blob/83dd80ae8660115d2db81d90973ce92e5df537f5/BTCPayServer/Extensions/EmailSenderExtensions.cs

Currently, it just builds the email string based on if it's setting or resetting

saliehendricks:
Happy to take a look at this.

Had a quick look at the code. It is fairly trivial to update the copy as per @britttttk's suggestion. 

My change will look something like:

Subject:
Body:
```
A request has been made for you to update the password for your BTCPay Server account. Please click <a href='{link}'>here</a> to update your password. 

Alternatively copy and paste this URL into your browser: 
{link}
```
[...]

PR #2211, opened 2021-01-20
Improve password reset email copy

saliehendricks:
Fix for #1978 

Simple fix to update the password reset email copy:
image


britttttk:
ack 

@saliehendricks looks good. I'm not sure about the copy and paste link though. It looks kind of long and intimidating to non-tech people. Is that typical? @brainharrington @Eskyee any feedback? :)

Eskyee:
@saliehendricks looks good.  but I agree with @britttttk 
the copy & paste does looks intimidating to newbies & messy yes.



BTCPay Logo

Please confirm your password.  by clicking the button or link below. 

Button





I think this is the standard Forgot My Password reply. it might be nice just to have a single button 
"if we have a button?? I would remove the link word" and BTCPay Logo in header  included in forgotten email..  
would look more professional and clean.  I think!!

PresidentNick:
The previous string was setting the email copy on whether the user was creating a new password or resetting their existing password.

@saliehendricks Can you add a conditional string for when the user is setting their password for the first time, please? 

Although, I haven't seen what calls the email function with the new password boolean set to True. So maybe it doesn't make sense to have an extra string for when the user is setting a new password? When is this function called with True for newPassword argument?

saliehendricks:
@britttttk I have seen fallback urls in authentication emails, but i dont disagree its pretty techy and ugly tbh. Happy to remove that part.

@Eskyee Agree, a button would be nice. Will try to put something together. I wasn't even sure the email was html formatted, but it appears to be. BTCPayServer logo would be pretty cool. Might as well give it a try :)

@PresidentNick The initial issue that was set out did say it should just be 1 email regardless of it being for a new user. Lets see what I can do to smooth out the copy a little - maybe we can get it done with just 1 simple sentence and no conditionals.

saliehendricks:
I think we making progress. So far this is what I have:
image

Still need to confirm location/file/css regarding the styling & logo with some other devs before committing.

Eskyee:
That's sexi 💯 
Have a look around here for any design tips https://design.btcpayserver.org
Good job 👏

Eskyee:
@dennisreimann could fast track that 🙌

dennisreimann:
Thanks for the pointer @Eskyee, I'll review this tomorrow :)

saliehendricks:
Thanks @dennisreimann. 

My thinking is to create /email.css (currently I just have static strings which is not great). This requires plumbing in the schema:host:port in order to link to the css+images. Let me know if that sounds ok.

dennisreimann:
I would not get too fancy here, especially since styling email content is a pain. Most clients don't load external content and if we want to go with HTML emails we should

- inline css and images, so I think having a general email layout file the content gets wrapped with would be best
- keep it as streamlined as possible
- provide a plain text version too

britttttk:
I like the button @saliehendricks 💯, good idea @Eskyee

BTCPayServer/Extensions/EmailSenderExtensions.cs
```diff
@@ -14,8 +14,8 @@ public static void SendEmailConfirmation(this IEmailSender emailSender, string e
         public static void SendSetPasswordConfirmation(this IEmailSender emailSender, string email, string link, bool newPassword)
         {
             emailSender.SendEmail(email,
-                $"{(newPassword ? "Set" : "Reset")} Password",
-                $"Please {(newPassword ? "set" : "reset")} your password by clicking here: <a href='{link}'>link</a>");
+                $"Update Password",
```
Kukks, line 17: I don't think this should be a static "Update" password. If the user was created by an admin and no pass is set, it's more a "set" password as it's a creation action
```suggestion
                $"{(newPassword ? "Set" : "Update")}  Password",
```

Kukks (changes requested):
I suggest we stick to just the copy text improvement for the scope of this PR. We can do more advanced layout email changes separately to keep things moving

BTCPayServer/Extensions/EmailSenderExtensions.cs
saliehendricks: @Kukks @dennisreimann Cool cool, will keep it to just a copy update. Will adjust the subject as above.

A new feature request can be opened for html emails.

saliehendricks: @Kukks @dennisreimann I thought before I strip out the html parts we could review and discuss what we have so far. Kept it as simple as possible (no logo) just inline styles for a button and html font.

image

dennisreimann: As this is explicitely setting the font color – have you checked this works in dark mode too? (E.g. for user's that are using dark mode via system preferences)

saliehendricks: @dennisreimann I don't think dark mode affects emails?

NicolasDorier (approved):

NicolasDorier:
fine for me, ping @Kukks , if that's ok for you, you can merge.

Kukks (approved):
Looks good to me now. We can iterate with nicer HTML stuff later.

Good job @saliehendricks 👍

PR #2234, opened 2021-01-28
Safe browsing quick fixes

dennisreimann:
As discussed in #2139 these changes might address the issue.

### DIFF
--- a/BTCPayServer/Controllers/AccountController.cs
+++ b/BTCPayServer/Controllers/AccountController.cs
@@ -66,6 +66,7 @@ namespace BTCPayServer.Controllers
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/login")]
         public async Task<IActionResult> Login(string returnUrl = null, string email = null)
         {
 
@@ -89,6 +90,7 @@ namespace BTCPayServer.Controllers
 
         [HttpPost]
         [AllowAnonymous]
+        [Route("/login")]
         [ValidateAntiForgeryToken]
         [RateLimitsFilter(ZoneLimits.Login, Scope = RateLimitsScope.RemoteAddress)]
         public async Task<IActionResult> Login(LoginViewModel model, string returnUrl = null)
@@ -396,6 +398,7 @@ namespace BTCPayServer.Controllers
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/register")]
         [RateLimitsFilter(ZoneLimits.Register, Scope = RateLimitsScope.RemoteAddress)]
         public async Task<IActionResult> Register(string returnUrl = null, bool logon = true)
         {
@@ -413,6 +416,7 @@ namespace BTCPayServer.Controllers
 
         [HttpPost]
         [AllowAnonymous]
+        [Route("/register")]
         [ValidateAntiForgeryToken]
         public async Task<IActionResult> Register(RegisterViewModel model, string returnUrl = null, bool logon = true)
         {
@@ -479,6 +483,7 @@ namespace BTCPayServer.Controllers
         public bool RegisteredAdmin { get; set; }
 
         [HttpGet]
+        [Route("/logout")]
         public async Task<IActionResult> Logout()
         {
             await _signInManager.SignOutAsync();
@@ -527,6 +532,7 @@ namespace BTCPayServer.Controllers
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/forgot-password")]
         public IActionResult ForgotPassword()
         {
             return View();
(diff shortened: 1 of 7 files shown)


BTCPayServer/Controllers/AccountController.cs
```diff
@@ -66,6 +66,7 @@ public string ErrorMessage
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/login")]
```
Kukks, line 69:
```suggestion
        [Route("~/login")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -89,6 +90,7 @@ public async Task<IActionResult> Login(string returnUrl = null, string email = n
 
         [HttpPost]
         [AllowAnonymous]
+        [Route("/login")]
```
Kukks, line 93:
```suggestion
        [Route("~/login")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -396,6 +398,7 @@ public IActionResult Lockout()
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/register")]
```
Kukks, line 401:
```suggestion
        [Route("~/register")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -413,6 +416,7 @@ public async Task<IActionResult> Register(string returnUrl = null, bool logon =
 
         [HttpPost]
         [AllowAnonymous]
+        [Route("/register")]
```
Kukks, line 419:
```suggestion
        [Route("~/register")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -479,6 +483,7 @@ public async Task<IActionResult> Register(RegisterViewModel model, string return
         public bool RegisteredAdmin { get; set; }
 
         [HttpGet]
+        [Route("/logout")]
```
Kukks, line 486:
```suggestion
        [Route("~/logout")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -527,13 +532,15 @@ public async Task<IActionResult> ConfirmEmail(string userId, string code)
 
         [HttpGet]
         [AllowAnonymous]
+        [Route("/forgot-password")]
```
Kukks, line 535:
```suggestion
        [Route("~/forgot-password")]
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -527,13 +532,15 @@ public async Task<IActionResult> ConfirmEmail(string userId, string code)
...
         public IActionResult ForgotPassword()
         {
             return View();
         }
 
         [HttpPost]
         [AllowAnonymous]
+        [Route("/forgot-password")]
```
Kukks, line 543:
```suggestion
        [Route("~/forgot-password")]
```

BTCPayServer/wwwroot/modal/btcpay.js
```diff
@@ -39,6 +38,8 @@
     iframe.style.height = '100%';
     iframe.style.width = '100%';
     iframe.style.zIndex = '2000';
+    // Removed, see https://github.com/btcpayserver/btcpayserver/issues/2139#issuecomment-768223263
+    // iframe.setAttribute('allowtransparency', 'true');
```
Kukks, line 42: What are the implications for this? Is there potential for it to render badly?

BTCPayServer/Controllers/AccountController.cs
Kukks: We need to leave the old route too. If there are active password resets, their links will be broken once they update.

BTCPayServer/Controllers/AccountController.cs
Kukks: We need to leave the old route too. If there are active password resets, their links will be broken once they update.

Kukks (changes requested):
Please prepend `~` to the route as it ensures that the controller doesnt do any funny business and try to itself  as part of the route.

Also, the forgot password route change will break existing links

Meh, hate these blind changes but will hopefully give us some insight.. (I messaged netcraft but still no response btw)

BTCPayServer/wwwroot/modal/btcpay.js
dennisreimann: I haven't noticed any difference in Firefox.

NicolasDorier: I think @rockstardev wrote this, he might have an idea.

NicolasDorier:
Another commit I would like to see: Remove the Tor link from the login/register page. This is not needed anymore.

dennisreimann:
@NicolasDorier see 4984674b1d23b6978ddedacc21371313d983ba6f

Kukks (approved):
ack

NicolasDorier:
I will make a release with only those fixes.

PR #2240, opened 2021-01-30
Invoices list: Remove icon indicator for onchain

dennisreimann:
As [discussed on the chat](https://chat.btcpayserver.org/btcpayserver/pl/tcjy5ornhbd8i8jm4yj9y3maie) the icon feels unnecessary and isn't clear to users. Leaving it off and only indicating Lightning transactions avoids confusion.

### DIFF
--- a/BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
+++ b/BTCPayServer/Payments/PaymentTypes.Bitcoin.cs
@@ -17,7 +17,7 @@ namespace BTCPayServer.Payments
 
         public override string ToPrettyString() => "On-Chain";
         public override string GetId() => "BTCLike";
-        public override string GetBadge() => "🔗";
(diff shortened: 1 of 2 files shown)


Zaxounette (approved):

Kukks (approved):
code ACK

concept...meh? I don't know. Why should LN be given preferential treatment for a quick overview of payment method?

dennisreimann:
@Kukks There's a bit more context and discussion in the linked chat – TL;DR onchain is considered "normal", because that's what most merchants have enabled, whereas LN is regarded as an add-on. In addition to that the onchain icon/emoji wasn't very clear and rather confusing, that's why just omitting it seemed like the best option.

NicolasDorier:
I agree that the icon was confusing, it confused me, I thought it was a clickable link at first...

PR #2246, opened 2021-02-02
GreenField API: Wallet API

Kukks:
Still very rough, open for feedback


BTCPayServer.Client/Models/CreateOnChainTransactionRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public class CreateOnChainTransactionRequestDestination
+        {
+            public string Destination { get; set; }
+            public decimal? Amount { get; set; }
+            public bool SubtractFromAmount { get; set; }
+        }
+        [JsonConverter(typeof(FeeRateJsonConverter))]
+        public FeeRate FeeSatoshiPerByte { get; set; }
+        public bool ProceedWithPayjoin { get; set; }
```
NicolasDorier, line 19: ?

BTCPayServer.Client/Models/CreateOnChainTransactionRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+
+        public class CreateOnChainTransactionRequestDestination
+        {
+            public string Destination { get; set; }
+            public decimal? Amount { get; set; }
+            public bool SubtractFromAmount { get; set; }
+        }
+        [JsonConverter(typeof(FeeRateJsonConverter))]
+        public FeeRate FeeSatoshiPerByte { get; set; }
```
NicolasDorier, line 18: I think you can just call it `FeeRate`. Reason is that later we could always allow other way to pass the fee rate say `feeRate: "1000 sat/kb"`

NicolasDorier:
@pavlenex can you tag this one for next release?

NicolasDorier:
Seems quite good, aside from some nits, the point which bother me is `CreateOnChainTransaction` which create transaction AND broadcast. This should be split into 2 parts. Create the transaction, THEN broadcast it. This should be two different calls.

pavlenex:
@Kukks there's a conflict now, can you please resolve it?

NicolasDorier:
I think this PR is breaking tests, tried to run again, no luck

NicolasDorier:
tests still broken

BTCPayServer.Client/Models/OnChainWalletTransactionData.cs
```diff
@@ -0,0 +1,35 @@
...
+        public string Comment { get; set; }
+        public Dictionary<string, LabelData> Labels { get; set; } = new Dictionary<string, LabelData>();
+
+        [JsonConverter(typeof(NumericStringJsonConverter))]
+        public decimal Amount { get; set; }
+
+        [JsonConverter(typeof(UInt256JsonConverter))]
+        public uint256 BlockHash { get; set; }
+
+        [JsonConverter(typeof(NumericStringJsonConverter))]
```
NicolasDorier, line 24: it should not be a string.

BTCPayServer/Extensions/AuthorizationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace BTCPayServer
+{
+    public static class AuthorizationExtensions
+    {
+        public static async Task<(bool HotWallet, bool RPCImport)> CanUseHotWallet(
+            this IAuthorizationService authorizationService,
+            PoliciesSettings policiesSettings,
+            ClaimsPrincipal user)
+        {
```
NicolasDorier, line 17: This method is completely wrong.

If not greenfield, you only need to check `user.IsInRole(Roles.ServerAdmin)`.
If greenfield, you need to run `authorizationService.AuthorizeAsync`.

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -0,0 +1,543 @@
...
+        [HttpDelete("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address")]
+        public async Task<IActionResult> UnReserveOnChainWalletReceiveAddress(string storeId, string cryptoCode)
+        {
+            if (IsInvalidWalletRequest(cryptoCode, out BTCPayNetwork network,
+                out DerivationSchemeSettings derivationScheme, out IActionResult actionResult)) return actionResult;
+
+            var addr = await _walletReceiveService.UnReserveAddress(new WalletId(storeId, cryptoCode));
+            if (addr is null)
+            {
+                return NotFound();
```
NicolasDorier, line 118: No need to change for this PR, but I think making naked NotFound (no NBXplorerError) is bad idea, as the user may think the route does not exist rather than the wallet does not exists.

NicolasDorier:
tests still broken

BTCPayServer.Client/Models/OnChainWalletTransactionData.cs
Kukks: Changed

BTCPayServer/Extensions/AuthorizationExtensions.cs
Kukks: The issue I found was that we want the user to use the store-relevant permissions but hot wallet features depends on a config or if the user is actually an admin:

*  If current claim has `CanModifyServerSettings`,  user is authorized
*  If current claim does not have `CanModifyServerSettings`, but is in greenfield (and thus may not have been granted that permission), check if the user is actually an admin (via role check on claim that we always set after we authenticate the user)
* If user does not have `CanModifyServerSettings` and is not an admin, check if the policies allow this functionality for all

BTCPayServer/Controllers/GreenField/LightningNodeApiController.cs
```diff
@@ -192,7 +192,7 @@ public virtual async Task<IActionResult> GetDepositAddress(string cryptoCode)
                 return NotFound();
             }
 
-            return Ok(new JValue((await lightningClient.GetDepositAddress()).ToString()));
+            return Ok((await lightningClient.GetDepositAddress()).ToString());
```
NicolasDorier, line 195: I think it should be `Json("blah")`, else it will returns `blah` rather than `"blah"`

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -335,8 +335,16 @@ public async Task<IActionResult> GetOnChainWalletUTXOs(string storeId, string cr
                               new FeeRate(1.0m);
             if (request.FeeRate is null)
             {
-                var feeRate = await explorerClient.GetFeeRateAsync(1);
-                request.FeeRate = feeRate.FeeRate;
+                try
+                {
+                    var feeRate = await explorerClient.GetFeeRateAsync(1);
+                    request.FeeRate = feeRate.FeeRate;
```
NicolasDorier, line 341: remove this, if FeeRate is not set, let NBXplorer does the default by not setting FeeRate.

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -1481,9 +1481,19 @@ public async Task WalletAPITests()
...
-                createTxRequest, tester.ExplorerClient.Network.NBitcoinNetwork);
+            Transaction tx;
+            try
+            {
+                tx = await client.CreateOnChainTransactionButDoNotBroadcast(walletId.StoreId, walletId.CryptoCode,
+                    createTxRequest, tester.ExplorerClient.Network.NBitcoinNetwork);
+            }
+            catch (Exception e)
+            {
+                createTxRequest.FeeRate = new FeeRate(5m);
```
NicolasDorier, line 1492: just hard code this one in the tests.

BTCPayServer/Controllers/GreenField/LightningNodeApiController.cs
Kukks: Yes, I changed it on purpose. It's much clearer IMO

Kukks:
reverted the "string" joson thing as it was causing issues with other endpoints

BTCPayServer/Extensions/AuthorizationExtensions.cs
rockstardev: Nicolas, the way I understood Kukks here is that if the store owner has access to the hot wallet within his store - we shouldn't require him to need ServerAdmin API key. 

Rather with API key he can create on store level he should be able to access hot wallet if him as store owner enabled it.

PR #2248, opened 2021-02-03
Ensure "No" selection is maintained for custom price in POS app

### LINKED ISSUE #2247: Custom price display bug
**Describe the bug**
The 'custom price' does not stay on 'no' after changes have been saved. This is regarding 'products' of a 'store' in the 'apps' tab. The bug is also visible in the 'toggle raw editor' template area.

**To Reproduce the bug**
In apps>store settings>edit product>custom price>no>save changes>save settings

Now the products custom price option has been saved and the store has been saved.

Scroll down to and select  'toggle raw editor' to start. The given products 'custom' variable is 'true' (aka yes). Go back to the given product through the edit button and the 'custom price' option is set to 'yes'.

**Expected behavior**

After setting the custom price to 'no' and saving the changes, it should remain on 'no' for both the toggle raw editor 'custom' variable (aka false) and the 'custom price' option within the edit product tab.

**Screenshots**

N/A

- BTCPay Server Version: v1.0.6.5
- Deployment Method: btcpay jungle
- Browser: Tor and Firefox

**Logs (if applicable)**
N/A

**Setup Parameters**
N/A

**Additional context**
This does not actually affect a customers ability to set a custom price. If the user has set it to 'no' then it will behave that way but not display it correctly. This casts doubt in users' minds as to whether or not a costumer can or can't set a custom price.

bolatovumar:
fix #2247

https://user-images.githubusercontent.com/1934678/106697410-8885cc80-6593-11eb-8aa2-406b798acbe8.mov

### DIFF
--- a/BTCPayServer/Views/Apps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Apps/TemplateEditor.cshtml
@@ -84,8 +84,7 @@
                                 <div class="col-sm-3">
                                     <label>Custom price</label>
                                     <select class="form-control" v-model="editingItem.custom">
-                                        <option v-bind:value="false">No</option>
-                                        <option v-bind:value="true">Yes</option>
+                                        <option v-for="option in customPriceOptions" :value="option.value">{{option.text}}</option>
                                     </select>
                                 </div>
                             </div>
@@ -134,6 +133,10 @@ $(function() {
             errors: [],
             items: [],
             editingItem: null,
+            customPriceOptions: [
+                { text: 'No', value: false },
+                { text: 'Yes', value: true },
+            ],
             elementId: "@Model.templateId"
         },
         computed: {
@@ -210,7 +213,7 @@ $(function() {
                             image = productProperty.replace('image:', '').trim();
                         }
                         if (productProperty.indexOf('custom:') !== -1) {
-                            custom =productProperty.replace('custom:', '').trim();
+                            custom = productProperty.replace('custom:', '').trim();
                         }
                         if (productProperty.indexOf('inventory:') !== -1) {
                             inventory = parseInt(productProperty.replace('inventory:', '').trim(),10);
@@ -228,7 +231,7 @@ $(function() {
                             price: price,
                             image: image || null,
                             description: description || '',
-                            custom: Boolean(custom),
+                            custom: custom === "true",
                             inventory: isNaN(inventory)? null: inventory,
                             paymentMethods: paymentMethods
                         });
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -210,7 +213,7 @@ $(function() {
                             image = productProperty.replace('image:', '').trim();
                         }
                         if (productProperty.indexOf('custom:') !== -1) {
-                            custom =productProperty.replace('custom:', '').trim();
+                            custom = productProperty.replace('custom:', '').trim();
```
bolatovumar, line 216: Just added a space here.

BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -228,7 +231,7 @@ $(function() {
                             price: price,
                             image: image || null,
                             description: description || '',
-                            custom: Boolean(custom),
```
bolatovumar, line 231: Since `custom` is always a string (from above `custom = productProperty.replace('custom:', '').trim();`)  casting it to a boolean value using `Boolean(custom)` will always produce `true` unless `custom` happens to be an empty string. Checking for `custom === "true"` ensures the boolean value is set correctly.

BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -263,8 +266,8 @@ $(function() {
                         if (inventory) {
                             template += '  inventory: ' + inventory + '\n';
                         }
-                        if (custom) {
-                            template += '  custom: true\n';
+                        if (custom != null) {
```
bolatovumar, line 269: Make sure we don't just always set `custom` to `true` in the template.

BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -376,6 +379,11 @@ $(function() {
                 return item;
             },
             unEscapeKey : function(k){
+                // Without this check a `false` boolean value will always be returned as an empty string
+                if (k === false) {
```
bolatovumar, line 383: This is a special case for when we select "No" and the select value is now `false`. If we pass `false` through the below `$('<div/>').html(k).text()` expression it will return an empty string which doesn't match any of the available `<select>` options which will make it look as if "No" is not selected. Passing `true` through `$('<div/>').html(k).text()`, however, will give you back a string `"true"` which makes the `<select>` work correctly.

dennisreimann (approved):

HeadyWook:
Awesome, thank you!

Kukks (approved):
LGTM

bolatovumar:
The test that failed was for coin selection and is unrelated.

PR #2251, opened 2021-02-03
Fix empty GetLightningClient return value

dennisreimann:
Fixes an [issue brought up the chat](https://chat.btcpayserver.org/btcpayserver/pl/y5813qcodtgitnpa8qur18upew): In case the LN node of a store isn't configured, the endpoint returned an error `503` instead of `404`, because the code crashed with an `System.NullReferenceException` before reaching [the `null` check](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/GreenField/LightningNodeApiController.cs#L46).


Kukks (approved):
LGTM

PR #2285, opened 2021-02-15
Circleci project setup

dotmaquinna:


Kukks:
Quite certain this is not meant to be a contrib to the original btcpay org

PR #2286, opened 2021-02-15
Remove Max body request size

Kukks:
Upload limit by kestrel restricts plugin upload of 30mb+. this removes that limit


NicolasDorier:
It's ok but have you made a search for alternative so it can be limited to a single route for example?
Also have you actually tried it? because nginx is also limiting stuff.
This open some vulnerabilities, so I think a high value would be better.

Kukks:
> It's ok but have you made a search for alternative so it can be limited to a single route for example?
> Also have you actually tried it? because nginx is also limiting stuff.
> This open some vulnerabilities, so I think a high value would be better.

Yeah, the error was coming from kestrel since I encountered it even on dev env.

PR #2287, opened 2021-02-15
Fix payment request template

bumbummen99:
This will prevent the page from being vertically too small for the screen. Right now the footer will cut off and the screen will continue with the body bg on big screens. With this patch the main section will take grow vertically if space is available.

### DIFF
--- a/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
+++ b/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
@@ -63,8 +63,8 @@
         </style>
     </noscript>
 </head>
-<body class="h-100">
-    <div id="app" class="h-100 d-flex flex-column">
+<body>
+    <div id="app" class="min-vh-100 d-flex flex-column">
         <nav id="mainNav" class="navbar sticky-top py-3 py-lg-4 d-print-block">
             <div class="container">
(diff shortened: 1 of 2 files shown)


dennisreimann:
For context: This sticks the footer to the bottom instead of having it right after the last container

Before … 

before

After …

after

@pavlenex @dstrukt What are your thoughts? I'd be fine with this. If so, we should also do it like this in the Pull Payment view.

bumbummen99:
@dennisreimann It seems in your template the footer has no background. The default light theme has a green background on the footer.

pavlenex:
LGTM, don't mind the new way of displaying it.

NicolasDorier:
ping @dennisreimann should we merge this?

dennisreimann:
@NicolasDorier yeah we could, but it should be the same for the pull payment view then. 

@bumbummen99 Can you also update the pull payment view?

dstrukt:
I'm okay with the fixed "powered by" footer for the empty state, and agree, it should be consistent.

dennisreimann (changes requested):
Please rebase (test fixes) and update the pull payment view.

dennisreimann:
@bumbummen99 Can you give my access to your fork so that I can take over the missing pieces if you don't have the time? It's almost done :)

dennisreimann:
Thanks, rebased the PR and unified the templates – thanks for the swift reaction!

PR #2297, opened 2021-02-18
docs(invoices): fix mistaken LightningNetwork payment method

AryanJ-NYC:


Kukks:
Actually, the docs are correct but our internal implementation is not parsing it correctly. We wanted to make it more clear on greenfield in terms of payment methods since onchain/offchain is a little confusing and bitcoinlike/lightninglike is just weird.

Kukks:
Opened a PR that addresses the underlying issue at #2301. Thanks for the discovery!

AryanJ-NYC:
Awesome. Thanks!

PR #2299, opened 2021-02-18
PoS: Custom buy button text per product

dennisreimann:
Closes #2232.

I saw the bounty by @MaxHillebrand in the chat and did this – only realizing afterwards that @bolatovumar already discussed this with Max and wanted to take it on. As I didn't intend to get in the way I'd suggest Umar reviews this and we split the bounty if that's okay too.

## Editor

Works with regular text too, this example uses the `{0}` to be replaced by the amount, which is supported on a store level too. 

editor

## Static View

static

## Cart View

cart

### DIFF
--- a/BTCPayServer.Tests/README.md
+++ b/BTCPayServer.Tests/README.md
@@ -35,7 +35,7 @@ You can also generate blocks:
 LND Connection string example: 
 type=lnd-rest;server=https://127.0.0.1:8084/;macaroonfilepath="local path to admin.macaroon on your computer, without these quotes";allowinsecure=true
 
-Now you can create a lightning invoice on BTCPay Server regtest and make a payment through Polar.
+Now you can create a Lightning invoice on BTCPay Server regtest and make a payment through Polar.
 
 PLEASE NOTE: You may get an exception break in Visual Studio. You must quickly click "Continue" in VS so the invoice is generated.
 Or, uncheck the box that says, "Break when this exceptiontype is thrown".
--- a/BTCPayServer.Tests/README.md
+++ b/BTCPayServer.Tests/README.md
@@ -26,6 +26,17 @@ You can also generate blocks:
 .\docker-bitcoin-generate.ps1 3
 ```
 
+### Using Polar to test Lightning payments
+
+-Install and run [Polar](https://lightningpolar.com/). Setup a small network of nodes.
+-Goto your Store's General Settings and enable Lightning
+-Build you connection string using the Connect infomation in the Polar app
+
+LND Connection string example: 
+type=lnd-rest;server=https://127.0.0.1:8084/;macaroonfilepath="local path to admin.macaroon on your computer, without these quotes";allowinsecure=true
+
+Now you can create a lightning invoice on BTCPayserver regtest and make payment through polar
+PLEASE NOTE: 
 ### Using the test litecoin-cli
 
 Same as bitcoin-cli, but with `.\docker-litecoin-cli.ps1` and `.\docker-litecoin-cli.sh` instead.
--- a/BTCPayServer.Tests/README.md
+++ b/BTCPayServer.Tests/README.md
@@ -28,15 +28,19 @@ You can also generate blocks:
 
 ### Using Polar to test Lightning payments
 
--Install and run [Polar](https://lightningpolar.com/). Setup a small network of nodes.
--Goto your Store's General Settings and enable Lightning
--Build you connection string using the Connect infomation in the Polar app
+- Install and run [Polar](https://lightningpolar.com/). Setup a small network of nodes.
+- Go to your store's General Settings and enable Lightning.
+- Build your connection string using the Connect infomation in the Polar app.
 
 LND Connection string example: 
 type=lnd-rest;server=https://127.0.0.1:8084/;macaroonfilepath="local path to admin.macaroon on your computer, without these quotes";allowinsecure=true
 
-Now you can create a lightning invoice on BTCPayserver regtest and make payment through polar
-PLEASE NOTE: 
+Now you can create a lightning invoice on BTCPayserver regtest and make a payment through polar.
+
+PLEASE NOTE: You may get an exception break in Visual Studio. You must quickly click "Continue" in VS so the invoice is generated.
+Or, uncheck the box that says, "Break when this exceptiontype is thrown".
+
+
 ### Using the test litecoin-cli
 
 Same as bitcoin-cli, but with `.\docker-litecoin-cli.ps1` and `.\docker-litecoin-cli.sh` instead.
--- a/BTCPayServer.Tests/README.md
+++ b/BTCPayServer.Tests/README.md
@@ -35,7 +35,7 @@ You can also generate blocks:
 LND Connection string example: 
 type=lnd-rest;server=https://127.0.0.1:8084/;macaroonfilepath="local path to admin.macaroon on your computer, without these quotes";allowinsecure=true
 
-Now you can create a lightning invoice on BTCPayserver regtest and make a payment through polar.
+Now you can create a lightning invoice on BTCPay Server regtest and make a payment through Polar.
 
 PLEASE NOTE: You may get an exception break in Visual Studio. You must quickly click "Continue" in VS so the invoice is generated.
 Or, uncheck the box that says, "Break when this exceptiontype is thrown".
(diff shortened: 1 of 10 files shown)


Kukks (changes requested):
Need to take into account crowdfund app too as it uses the same templating

dennisreimann:
@Kukks Thanks for the pointer, added.

crowdfund

MaxHillebrand:
You're a treasure @dennisreimann, thanks for picking it up so fast :) 

Yes, I'm happy to include reviewers of the PR in the bounty.

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
```diff
@@ -17,6 +17,7 @@ public class ItemPrice
             public ItemPrice Price { get; set; }
             public string Title { get; set; }
             public bool Custom { get; set; }
+            public string BuyButtonText { get; set; }
```
Kukks, line 20: `BuyButtonText` or simply `ButtonText`?

Kukks (approved):
LGTM, just a small quyestion if the prop name should be simpler or not but nothing major

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
dennisreimann: I thought about this too and was torn: The upper level prop is called `CustomButtonText` – one might argue we should match this, but I wanted to avoid confusion and used the buy-prefix to be specific … so yeah, let's discuss this.

pavlenex (changes requested):
If `custom price` is set to `yes` the text you add in the button name field doesn't take effect. If it's desired behavior, we should probably hide the `Buy Button Text` field, if it's not desired, the name in the `Buy Button Text` should take effect even if custom price is set to `yes`.

BTCPayServer/Models/AppViewModels/ViewPointOfSaleViewModel.cs
bolatovumar: I think `BuyButtonText` is more accurate even though it's more verbose.

bolatovumar:
> If `custom price` is set to `yes` the text you add in the button name field doesn't take effect. If it's desired behavior, we should probably hide the `Buy Button Text` field, if it's not desired, the name in the `Buy Button Text` should take effect even if custom price is set to `yes`.
> 
> Screenshot from 2021-02-19 17-23-46
> 
> Screenshot from 2021-02-19 17-26-22

Yeah, I think if you specify custom button text it should apply no matter what you select for "Custom price". The user who puts in custom button text probably doesn't care what is selected in the "Custom price" dropdown when they set the button text, they just want this text to appear no matter what.

dennisreimann:
Fixed, also updated the display so that it works with long and short text …

custom-price

pavlenex (approved):
tACK.

bolatovumar (commented):
I tested this branch locally on my machine and noticed a couple of issues:

1. If I add a custom text to the first item in the default POS app (green tea) and then update it again right after all OTHER items get the old text for the first item after the second update (see screenshot). I can reproduce this consistently.
   - Screen Shot 2021-02-22 at 6 59 50 PM
 2. If I add text that's too long the button overspills. This one is probably not a big deal as I don't expect people will add long text in these buttons
    - Screen Shot 2021-02-22 at 7 01 36 PM

dennisreimann:
@bolatovumar good spot, it's fixed now. As far as I read it, we had the same problem before with other properties (price, title, description and custom) because they weren't reset before parsing a new item. Fixed in 32370545cbcc2d7a3141a6ab3fd19ecb94ccc8d6.

BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -478,12 +478,21 @@ public async Task CanCreateAppPoS()
                 s.Driver.FindElement(By.Id("SelectedStore")).SendKeys(storeName);
                 s.Driver.FindElement(By.Id("Create")).Click();
                 s.Driver.FindElement(By.Id("DefaultView")).SendKeys("Cart");
+                s.Driver.FindElement(By.CssSelector(".template-item:nth-of-type(1) .btn-primary")).Click();
```
bolatovumar, line 481: Wouldn't it be better to use some `test-id` attribute instead so if a new button is added or something you don't have to update this test?

BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -113,11 +113,17 @@
                                     <input type="text" required pattern="[^\*#]+"  class="form-control mb-2" v-model="editingItem.id" ref="txtId"/>
                                 </div>
                             </div>
+                            <div class="form-row">
+                                <div class="col">
+                                    <label>Buy Button Text</label>
+                                    <input type="text" id="BuyButtonText" class="form-control mb-2" v-model="editingItem.buyButtonText" ref="txtBuyButtonText"/>
```
bolatovumar, line 119: `ref` attribute seems weird. `txtBuyButtonText` <-- looks like `txt` and `Text` are meant to represent text and are repeated?

BTCPayServer/Views/AppsPublic/PointOfSale/Cart.cshtml
```diff
@@ -253,7 +253,7 @@
                         </div>
                         <div class="card-footer pt-0 bg-transparent border-0">
 
-                            <span class="text-muted small">@Model.ButtonText.Replace("{0}",item.Price.Formatted).Replace("{Price}",item.Price.Formatted)</span>
+                            <span class="text-muted small">@((item.BuyButtonText ?? Model.ButtonText).Replace("{0}",item.Price.Formatted).Replace("{Price}",item.Price.Formatted))</span>
```
bolatovumar, line 256: Can we confirm with @MaxHillebrand that we also want to modify the button text in the cart?

bolatovumar (approved):
Tested the updated branch and couldn't find any issues.

BTCPayServer/Views/AppsPublic/PointOfSale/Cart.cshtml
MaxHillebrand: Yes please. It makes sense to have this custom too, for the same reasons as the other buttons.

BTCPayServer.Tests/SeleniumTests.cs
dennisreimann: In general I agree. This one selects the first article in the editor – as it is a dynamic list I thought it'd be ok this way too.

BTCPayServer/Views/Apps/TemplateEditor.cshtml
dennisreimann: It's the same as for the other input fields – I think the `txt` part refers to the input being a text input. Can change it, but think we should then change the other field refs too to be consistent.

dennisreimann:
Fix for the Selenium test contained in #2310. Can rebase this one if the other PR gets merged first.

BTCPayServer.Tests/README.md
```diff
@@ -26,6 +26,21 @@ You can also generate blocks:
...
+### Using Polar to test Lightning payments
+
+- Install and run [Polar](https://lightningpolar.com/). Setup a small network of nodes.
+- Go to your store's General Settings and enable Lightning.
+- Build your connection string using the Connect infomation in the Polar app.
+
+LND Connection string example: 
+type=lnd-rest;server=https://127.0.0.1:8084/;macaroonfilepath="local path to admin.macaroon on your computer, without these quotes";allowinsecure=true
+
+Now you can create a lightning invoice on BTCPay Server regtest and make a payment through Polar.
```
MaxHillebrand, line 38:
```suggestion
Now you can create a Lightning invoice on BTCPay Server regtest and make a payment through Polar.
```

minor typo nit for consistency.

dennisreimann: Thanks for fixing! Not sure though why this slipped in there with my rebase … it's unrelated and present on master already.

MaxHillebrand:
Thank you @dennisreimann @bolatovumar @Kukks @pavlenex for all you past and present support of free software like BTCPay! Y'all are heroes!!

Some sats have magically found their way into your wallets. :black_heart: :fire:

PR #2308, opened 2021-02-23
Improve policies structure and wordings

dennisreimann:
Closes #2307.

policies

### DIFF
--- a/BTCPayServer/Views/Server/Policies.cshtml
+++ b/BTCPayServer/Views/Server/Policies.cshtml
@@ -1,93 +1,115 @@
 @using BTCPayServer.Services
 @using BTCPayServer.Services.Mails;
 @model BTCPayServer.Services.PoliciesSettings
-@inject BTCPayServer.Services.SettingsRepository _SettingsRepository
+@inject SettingsRepository _SettingsRepository
 @{
     ViewData.SetActivePageAndTitle(ServerNavPages.Policies);
 }
 
-
 <partial name="_StatusMessage"/>
-@if (!this.ViewContext.ModelState.IsValid)
+
+@if (!ViewContext.ModelState.IsValid)
 {
     <div asp-validation-summary="All" class="text-danger"></div>
 }
 
 <form method="post">
-    <div class="form-group">
-        <div class="form-check">
-            @{
-                var emailSettings = (await _SettingsRepository.GetSettingAsync<EmailSettings>()) ?? new EmailSettings();
-                /**
-                 * The "|| Model.RequiresConfirmedEmail" check is for the case when a user had checked
-                 * the checkbox without first configuring the e-mail settings so that they can uncheck it.
-                 **/
-                var isEmailConfigured = emailSettings.IsComplete() || Model.RequiresConfirmedEmail;
-            }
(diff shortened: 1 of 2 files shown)


dennisreimann:
@pavlenex Somehow I cannot add @britttttk and @Zaxounette as reviewers … can we add them to the list of potential reviewers?

Kukks (approved):
CODE ACK
Concept ACK

Zaxounette (changes requested):
* The "check for new version and notify" string is missing in your screenshot. Did you place it elsewhere ?

* Also, the `Display app on website root` dropdown clickable arrow has changed. 
I personally prefer the current arrow representation.

Current:
DaQ3sWAwQr

Proposed in this PR:
MXZVgDUDph

dennisreimann:
@Zaxounette I took the screenshot locally and [this check](https://github.com/btcpayserver/btcpayserver/pull/2308/files#diff-fb98632cfde120a1e821fce642b2a8ea5f615f6d284b84011aa2fbed07501be7L76) seems to prevent the option from showing in this case (maybe related to the dev mode).

Here's what it looks like without the check:

policies

As for the select dropdown: I've added the `w-auto` CSS class, which ensures it's not wider than necessary, but as far as I can see it doesn't affect the arrow display. I also checked my current live version, and it looks the same there too:

select

Zaxounette:
Thanks Dennis ! 

FYI, just checked on different browsers for the dropdown arrow:
* Firefox renders it as your screenshot.
* Chrome renders it as mine.

In any case, sorry for that.

Zaxounette (approved):
Thanks Dennis !

FYI, just checked on different browsers for the dropdown arrow:

* Firefox renders it as your screenshot.
* Chrome renders it as mine.

In any case, sorry for that.

dennisreimann:
No worries, thanks for pointing that out, @Zaxounette – might as well have been related to that change, so that feedback is welcome :)

britttttk:
Concept ACK, much more clear! @dennisreimann

pavlenex:
> @pavlenex Somehow I cannot add @britttttk and @Zaxounette as reviewers … can we add them to the list of potential reviewers?

I've invited both Brit and Zax to the organization, once they accept the invitation, we should be able to assign issues to them.

BTCPayServer/Views/Server/Policies.cshtml
```diff
@@ -135,40 +157,43 @@
...
-                            <button type="button" class="text-decoration-none btn btn-link revert-default fa fa-refresh " title="Revert to default"></button>
-                        </div>
-                    </div>
 
-                </li>
-            }
-        </ul>
+                @for (int i = 0; i < networks.Count(); i++)
+                {
+                    var network = networks.ElementAt(i);
```
NicolasDorier, line 175: I'll let it pass this time, but please don't put logic in cshtml, you should put this logic in the controller and changing the viewmodel

dennisreimann: This PR didn't change that part, but I'm happy to refactor it with the next pass. Thanks for the hint!

PR #2309, opened 2021-02-23
Allow invoice to be marked even when new

### LINKED ISSUE #2019: Allow transioning invoices into the "completed" state without going through "invalid"
Currently, you have to mark the invoice as "invalid" first, and only then the option to mark it as "completed" is made available. It would be nice if it was possible to change into "completed" directly.

Kukks:
fixes #2019


dennisreimann:
Fix for the failing Selenium test contained in #2310 – grab it while it's hot. 😬

Issue #1603, opened 2020-05-25
The internal Lightning connection string address is checked for HTTPS instead of the one entered via web UI


Kixunil:
**Describe the problem/bug**
Instead of checking the actual connection string, the check whether an address is localhost or onion checks against the `btc.lightning` setting from config file

**Your environment**
* Version of BTCPay Server: 1.0.4.2
* Deployment method: [Cryptoanarchy Debian Repository](https://https://github.com/Kixunil/cryptoanarchy-deb-repo-builder)
* Other relevant environment details: N/A

**Logs (if applicable)**
N/A

**To Reproduce**
1. Configure BTCPayServer without `btc.lightning` key in the config file
2. Go to Add Lightning node page
3. Add the connection string to that field using http and pointing to localhost or 127.0.0.1
4. *See BTCPay complaining about HTTP*
5. Add `btc.lightning` with the same string to btcpay config
6. Restart the service
7. Add the same string again
8. *Observe that the string is accepted*
9. Change the connection string to a non-local address
10. **See the connection string being accepted despite being dangerous**

**Expected behavior**
The setting of `btc.lightning` doesn't affect detection of HTTP. In particular:

* When `btc.lightning` is missing or is different from the specified string, the error message doesn't show for loclahost and onion addresses
* **When `btc.lightning` contains `https`, but the entered string does not, the user gets the error**

**Actual behavior**
The condition checks the internal node instead of the entered connection string.

**Screenshots/Links**
I believe screenshot is not needed here, will add if required.

**Additional context**
* [Probably offending line](https://github.com/btcpayserver/btcpayserver/blob/3425f84507ada2f527f7e735da6270e0b7eb86a0/BTCPayServer/Controllers/StoresController.LightningLike.cs#L94)
* This supersedes #1600

bolatovumar:
I'm not sure if this issue even applies anymore? The error referenced in this issue doesn't even exist anymore.

Kixunil:
Good point, I just tested it with recent version and it now works fine. I guess the refactorings fixed it. :tada:

PR #2314, opened 2021-02-26
Refactor how we handle and validate LN ConnectionStrings

NicolasDorier:
This PR is refactoring how we handle and validate connection strings for LN payment methods.

Because connection string can read the file system, we must be sure that a malicious party can't leverage lightning connection string to read and steal arbitrary file on the host.

While the original code was doing this successfully, it was hard to read and hard to understand.

I simplified the code in the following way: Now a Lightning Payment Method can be either an external connection string (as before), but some fields forbidden (such as macaroonfilepath).
If the user wants to use the internal node, then the connection string should be the string "Internal Node".

The nice thing about it is that if the user change his internal lightning implementation, he won't need to re-set the connection string to the new internal node.

I am also migrating all stores which used the internal node before to the new connection string "Internal Node".

long story short:
1. The connection strings are now strictly limited to a subset if you are not admin "IsSafe()"
2. For using internal node, you need to set the connection string to "Internal Node"
3. A migration has been done so connection strings that match internal node are converted to "Internal Node"
4. Some cleanup of old fields in the DB that we don't use anymore
5. The new design means that it is possible to set lightning network's payment method to internal in the greenfield API @Kukks

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -540,7 +540,7 @@ namespace BTCPayServer.Tests
             }
         }
 
-        private async Task AssertValidationError(string[] fields, Func<Task> act)
+        private async Task<GreenFieldValidationException> AssertValidationError(string[] fields, Func<Task> act)
         {
             var remainingFields = fields.ToHashSet();
             var ex = await Assert.ThrowsAsync<GreenFieldValidationException>(act);
@@ -550,6 +550,7 @@ namespace BTCPayServer.Tests
                 remainingFields.Remove(field);
             }
             Assert.Empty(remainingFields);
+            return ex;
         }
 
         private async Task AssertHttpError(int code, Func<Task> act)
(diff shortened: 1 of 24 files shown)


Kixunil:
This sounds reasonable, good point about reading arbitrary files!

I'd still love to see the ability to have multiple internal nodes. It doesn't need to be implemented right away. Maybe to make it easier connection string could be `InternalNode=LABEL`? The initial implementation can use fixed `InternalNode=main` value.

NicolasDorier:
That's good idea, after this refactoring it can be easily implemented later.

BTCPayServer/Views/Stores/AddLightningNode.cshtml
```diff
@@ -42,6 +42,14 @@
             <div class="form-group">
                 <p class="mb-2">The connection string encapsulates the configuration for connecting to your lightning node. BTCPay Server currently supports:</p>
                 <ul>
+                    <li class="mb-2">
+                        <strong>Internal node</strong>, if you are administrator of the server:
+                        <ul>
+                            <li>
+                                <code>Internal Node</code>
```
Kukks, line 49: Is it supposed to be with a space here too?

BTCPayServer/Payments/Lightning/LightningListener.cs
```diff
@@ -376,15 +394,26 @@ async Task Listen(CancellationToken cancellation)
...
+                Logs.PayServer.LogError(ex, $"{supportedPaymentMethod.CryptoCode} (Lightning): Error while contacting {GetLightningUrl().BaseUri}");
+                Logs.PayServer.LogInformation($"{supportedPaymentMethod.CryptoCode} (Lightning): Stop listening {GetLightningUrl().BaseUri}");
             }
             catch (OperationCanceledException) when (cancellation.IsCancellationRequested) { }
             if (_ListenedInvoices.IsEmpty)
-                Logs.PayServer.LogInformation($"{supportedPaymentMethod.CryptoCode} (Lightning): No more invoice to listen on {supportedPaymentMethod.GetLightningUrl().BaseUri}, releasing the connection.");
+                Logs.PayServer.LogInformation($"{supportedPaymentMethod.CryptoCode} (Lightning): No more invoice to listen on {GetLightningUrl().BaseUri}, releasing the connection.");
+        }
+
+        private LightningConnectionString GetLightningUrl()
```
Kukks, line 405: Duplicated code (add as static helper somewhere?)

BTCPayServer/Payments/Lightning/LightningListener.cs
```diff
@@ -240,6 +244,16 @@ private async Task CreateNewLNInvoiceForBTCPayInvoice(InvoiceEntity invoice)
 
         }
 
+        private LightningConnectionString GetLightningUrl(LightningSupportedPaymentMethod supportedMethod)
```
Kukks, line 247: Are you sure this is correct? 
It seems to try and see if payment method has a value and if not sets it to internal node.
What if a user finds a way to empty the conn string without deleting the method (non admin would gain access to internal node)?

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -135,6 +142,21 @@ public async Task<NodeInfo> GetNodeInfo(bool preferOnion, LightningSupportedPaym
             }
         }
 
+        private ILightningClient CreateLightningClient(LightningSupportedPaymentMethod supportedPaymentMethod, BTCPayNetwork network)
```
Kukks, line 145: duplicated method

BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -170,10 +143,9 @@ public async Task<IActionResult> AddLightningNode(string storeId, LightningNodeV
                     return View(vm);
             }
         }
-
         private bool CanUseInternalLightning()
         {
-            return (_BTCPayEnv.IsDeveloping || User.IsInRole(Roles.ServerAdmin) || _CssThemeManager.AllowLightningInternalNodeForAll);
+            return User.IsInRole(Roles.ServerAdmin) || _CssThemeManager.AllowLightningInternalNodeForAll;
```
Kukks, line 148: This might break tests?

BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -88,40 +92,9 @@ public async Task<IActionResult> AddLightningNode(string storeId, LightningNodeV
...
-                    {
-                        ModelState.AddModelError(nameof(vm.ConnectionString), "The macaroonfilepath should be fully rooted");
-                        return View(vm);
-                    }
-                }
-
-                if (isInternalNode && !CanUseInternalLightning())
-                {
-                    ModelState.AddModelError(nameof(vm.ConnectionString), "Unauthorized url");
+                    ModelState.AddModelError(nameof(vm.ConnectionString), $"You are not admin, so the connection string should not contains 'cookiefilepath', 'macaroondirectorypath', 'macaroonfilepath', and should not point to a local ip or to a dns name ending with '.internal', '.local', '.lan' or '.'.");
```
Kukks, line 97:
```suggestion
                    ModelState.AddModelError(nameof(vm.ConnectionString), $"You are not a server admin, so the connection string should not contain 'cookiefilepath', 'macaroondirectorypath', 'macaroonfilepath', and should not point to a local ip or to a dns name ending with '.internal', '.local', '.lan' or '.'.");
```

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -1287,33 +1288,85 @@ public async Task LightningNetworkPaymentMethodAPITests()
...
             {
-                await client.GetStoreOnChainPaymentMethod(store.Id, "BTC");
+                await adminClient.GetStoreOnChainPaymentMethod(store.Id, "BTC");
+            });
+
+
+            // Let's verify that the admin client can't change LN to unsafe connection strings without modify server settings rights
+            foreach (var forbidden in new string[]
+            {
+                "type=clightning;server=tcp://127.0.0.1",
```
Kukks, line 1327: what about 0.0.0.0, and even the ip of the server?

Kukks: or even ipv6 local ones?

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -114,6 +128,94 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
...
+                    });
+                    store.DerivationStrategies = jObject.ToString();
+                    store.DerivationStrategy = null;
+                }
+                if (string.IsNullOrEmpty(store.DerivationStrategies))
+                    continue;
+
+                var strats = JObject.Parse(store.DerivationStrategies);
+                bool updated = false;
+                foreach (var prop in strats.Properties().Where(p => p.Name.EndsWith("LightningLike", StringComparison.OrdinalIgnoreCase)))
```
Kukks, line 163: Should try and parse with PaymentMethodId to be safe

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -114,6 +128,94 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
...
+                {
+                    var method = ((JObject)prop.Value);
+                    var lightningCharge = method.Property("LightningChargeUrl", StringComparison.OrdinalIgnoreCase);
+                    var ln = method.Property("LightningConnectionString", StringComparison.OrdinalIgnoreCase);
+                    if (lightningCharge != null)
```
Kukks, line 168: why is charge so special?

Kukks (changes requested):
simplifies and cleans up all the hardcore logic around internal node checks

BTCPayServer/Controllers/StoresController.LightningLike.cs
NicolasDorier: I changed tests so the user is admin.

BTCPayServer.Tests/GreenfieldAPITests.cs
NicolasDorier: @Kukks yeah, both should be forbidden, I can add test for that.

BTCPayServer/Hosting/MigrationStartupTask.cs
NicolasDorier: I was thinking about it, but then we have a problem if somebody disabled shitcoins. (As payment id will not be parseable)

BTCPayServer/Hosting/MigrationStartupTask.cs
NicolasDorier: Charge was the first lightning implementation supported by BTCPay, so before there was 3 fields Username/Password/ChargeUrl rather than a single ConnectionString.

We probably changed that 1 or 2 years ago, but those fields were still there.

BTCPayServer/Payments/Lightning/LightningListener.cs
NicolasDorier: not possible, setting the connection string is only done through `void SetLightningUrl(LightningConnectionString connectionString)` which verify that: `connectionString` is not null.
`LightningConnectionString` make it sure it is not empty.

BTCPayServer/Views/Stores/AddLightningNode.cshtml
NicolasDorier: yes

Kixunil:
Since this changes connection string validation, I wonder how this affects #1603 
If it doesn't, would it make sense to bundle such change here?

NicolasDorier:
It probably does, will test if I can repro. I don't check http/https anymore in this PR. I just check that the external configuration string does not contains local stuff, so the checks are now simpler.

BTCPayServer/Payments/Lightning/LightningListener.cs
NicolasDorier: I will put a separated string instead. Make it harder to shoot in the foot.

BTCPayServer/Views/Stores/AddLightningNode.cshtml
Kixunil: I'm afraid this will lead to confusion if other ways are without space. Do you have some particular reason for it?

Kukks (approved):
ACCK

PR #2318, opened 2021-02-27
Fix direct URL for local storage with custom root path

### LINKED ISSUE #2315: LocalStorage URLs don't contain root path
**Describe the bug**
After uploading a file into local storage in server settings a URL is shown which points to `/LocalStorage/...` instead of `/root_paht/LocalStorage/...` and is thus broken. Manually prepending root path fixes the URL which then returns the appropriate file.

**To Reproduce the bug**
1. Setup BTCPayserver with non-empty root path, e.g. `/btcpay`
2. In Server settings -> Files upload any file
3. See the generated Direct URL contains `/LocalStorage/...` instead of `/btcpay/LocalStorage/...`
4. Click indirect URL - it will redirect to Direct URL which is broken
5. Fix the URL by prepending `/btcpay` to path - the request will succeed

**Expected behavior**

Direct URL contains `/btcpay` or whatever root path was set
Indirect URL redirects to URL containing the root path

**Screenshots**

btcpay-bug

- BTCPay Server Version:   1.0.6.7
- Deployment Method:        Debian package
- Browser:                  Firefox

**Logs (if applicable)**

I don't think they are needed

**Setup Parameters**

```
network=mainnet
rootPath=/btcpay
```

The rest should be irrelevant

**Additional context**

Same issue seems to be present in version 1.0.6.4

Cc @MaxHillebrand

bolatovumar:
Closes #2315 

Tested both with custom rooth path and without and seems to be working correctly for both.

Screen Shot 2021-02-26 at 7 19 10 PM

### DIFF
--- a/BTCPayServer/Storage/Services/Providers/FileSystemStorage/FileSystemFileProviderService.cs
+++ b/BTCPayServer/Storage/Services/Providers/FileSystemStorage/FileSystemFileProviderService.cs
@@ -37,7 +37,9 @@ namespace BTCPayServer.Storage.Services.Providers.FileSystemStorage
         public override async Task<string> GetFileUrl(Uri baseUri, StoredFile storedFile, StorageSettings configuration)
         {
             var baseResult = await base.GetFileUrl(baseUri, storedFile, configuration);
-            var url = new Uri(baseUri, LocalStorageDirectoryName);
+            // Set the relative URL to the directory name if the root path is default, otherwise add root path before the directory name
+            var relativeUrl = baseUri.AbsolutePath == "/" ? LocalStorageDirectoryName : $"{baseUri.AbsolutePath}/{LocalStorageDirectoryName}";
+            var url = new Uri(baseUri, relativeUrl);
             return baseResult.Replace(new DirectoryInfo(_datadirs.Value.StorageDir).FullName, url.AbsoluteUri,
                 StringComparison.InvariantCultureIgnoreCase);
         }
--- a/BTCPayServer/Views/Server/Files.cshtml
+++ b/BTCPayServer/Views/Server/Files.cshtml
@@ -64,7 +64,7 @@ else
                             @Url.Action("GetFile", "Storage", new
                             {
                                 fileId = Model.SelectedFileId
-                            }, Context.Request.Scheme, Context.Request.Host.ToString(), Context.Request.PathBase.ToString())
+                            }, Context.Request.Scheme, Context.Request.Host.ToString())
                         </a>
                     </div>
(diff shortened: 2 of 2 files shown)


BTCPayServer/Views/Server/Files.cshtml
```diff
@@ -70,7 +70,7 @@ else
                 </li>
                 <li class="list-group-item ">
                     <div class="row px-0 mx-0">
-                        <span class="col-sm-12 col-md-2">Direct Url </span>
+                        <span class="col-sm-12 col-md-2">Direct Url: </span>
```
bolatovumar, line 73: Not sure why there was not a `:` here previously. It there for the other line.

BTCPayServer/Storage/Services/Providers/FileSystemStorage/FileSystemFileProviderService.cs
```diff
@@ -37,7 +37,9 @@ protected override Task<IStorageProvider> GetStorageProvider(FileSystemStorageCo
         public override async Task<string> GetFileUrl(Uri baseUri, StoredFile storedFile, StorageSettings configuration)
         {
             var baseResult = await base.GetFileUrl(baseUri, storedFile, configuration);
-            var url = new Uri(baseUri, LocalStorageDirectoryName);
+            // Set the relative URL to the directory name if the root path is default, otherwise add root path before the directory name
+            var relativeUrl = baseUri.AbsolutePath == "/" ? LocalStorageDirectoryName : $"{baseUri.AbsolutePath}/{LocalStorageDirectoryName}";
```
Kukks, line 41: I think there is a helper extension method you can use that will compute the url correctly

Kukks (approved):
ACK

BTCPayServer/Storage/Services/Providers/FileSystemStorage/FileSystemFileProviderService.cs
bolatovumar: @Kukks happen to know what it would be?

Kukks:
image
What is this path at the end of the url?

bolatovumar:
> image
> What is this path at the end of the url?

No idea. It seems to always insert a `/#` at the end of the URL (even for default root path). I can take a look into it. It doesn't seem to actually affect anything and the link still works as expected.

bolatovumar:
@Kukks addressed here: 2d64ae31b8a6c010223d24497ce539d21589c766

Screen Shot 2021-02-28 at 12 40 24 PM

PR #2324, opened 2021-03-01
Fix Rates Test (Dash rate source switched to bitfinex)

### LINKED ISSUE #2323: Failing Tests - Bittrex not providing DASH_BTC rate
**Describe the bug**
As reported by @rockstardev in https://github.com/btcpayserver/btcpayserver/pull/2181#issuecomment-787512659 tests are failing because `Bittrex not providing DASH_BTC rate any more`.

Kukks:
fixes #2323

### DIFF
--- a/BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.Dash.cs
+++ b/BTCPayServer.Common/Altcoins/BTCPayNetworkProvider.Dash.cs
@@ -20,7 +20,7 @@ namespace BTCPayServer
                 DefaultRateRules = new[]
                     {
                         "DASH_X = DASH_BTC * BTC_X",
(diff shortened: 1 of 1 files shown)


NicolasDorier:
are you sure this is correct? I think the problem is that bitfinex just delisted DASH https://www.coindesk.com/bittrex-to-delist-privacy-coins-monero-dash-and-zcash

Kukks:
> are you sure this is correct? I think the problem is that bitfinex just delisted DASH https://www.coindesk.com/bittrex-to-delist-privacy-coins-monero-dash-and-zcash

Yes, I switched it to use bitfinex instead

rockstardev (approved):

PR #2325, opened 2021-03-01
Returning 0 for NextNetworkFee if it's null

rockstardev:
I've ran into this problem several times when testing locally, with NextNetworkFee throwing null:

image

This usually happens in the tests, specifically try running `UnitTest1.CanSetUnifiedQrCode` without this PR.

I've left comments on places of improvements in the code... how we can refactor initialization of `NextNetworkFee`.


Kukks (approved):

NicolasDorier:
Seems good, but I am curious about why it happen in the first place.

NicolasDorier:
The test `CanSetUnifiedQrCode` does not fail for me though

NicolasDorier:
Merging, even though I think there is another problem, I don't see how this can be null.

rockstardev:
@NicolasDorier there were improvements around this initialization... what's left was related to tests - and this null checks solves that finally.

PR #2334, opened 2021-03-07
Update invoice log row styling

### LINKED ISSUE #2328: Font color same as background color for invoice logs when mouse over
If there is a danger line in the invoice logs and you over with the mouse, you can't read the log.

image

image

bolatovumar:
**Before:**

Screen Shot 2021-03-06 at 10 08 09 PM

**After:**

Screen Shot 2021-03-06 at 9 58 23 PM

Because the `alert` CSS class is not intended to be used to style table rows it ends up looking weird in some cases when you hover over the table row and default table row hover styling is applied. Styling just the text in the table prevents this issue.

**Standard table (before):**

Screen Shot 2021-03-06 at 10 04 37 PM

**Standard table (after):**

Screen Shot 2021-03-06 at 10 01 51 PM

closes #2328

### DIFF
--- a/BTCPayServer/Views/Invoice/Invoice.cshtml
+++ b/BTCPayServer/Views/Invoice/Invoice.cshtml
@@ -290,7 +290,7 @@
                     <tbody>
                         @foreach (var evt in Model.Events)
                         {
-                            <tr class="alert alert-@evt.GetCssClass()">
+                            <tr class="text-@evt.GetCssClass()">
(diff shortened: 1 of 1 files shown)


dennisreimann:
There are also [contextual table classes](https://getbootstrap.com/docs/4.6/content/tables/#contextual-classes), which could solve the issue of the warning yellow being barely readable. Those also don't have the hover effect, which isn't needed here anyways.

PR #2337, opened 2021-03-07
Fix issue around new bech uppercase + vault supported flag

Kukks:
This adds a flag to the BTCPayNetwork to show it supports Vault. Additionally, it also fixes an issue in the bech32 prefix detection code if BTC is not available in the instance


Kukks:
@gruve-p Can you test it?

gruve-p:
tACK https://github.com/btcpayserver/btcpayserver/pull/2337/commits/d50151947e8fa7fbbb767ef33c5c26451d9ad70b on https://testnet.grspay.com/

PR #2339, opened 2021-03-07
Changed display date format on View Pull Payments screen

### LINKED ISSUE #1725: Pull Payment DateTime displayed
**Describe the problem/bug**
The DateTime displayed in the viewed pull payment is ahead of my local time. 

**Your environment**
* Version of BTCPay Server: BTCPayServer v1.0.5.2
* Deployment method: LunaNode
* Other relevant environment details: The DateTime is correct in the list of pull payments. 

**To Reproduce**
1. Go to Wallets
2. Click on Pull payments
3. Click view on the pull payment

**Expected behavior**
The time is was created for me is shown correctly as it is in the list of pull payments. 
In this case it should be 7/8/2020 11:06:16 PM

**Actual behavior**
When viewing the pull payment it instead shows as:
Updated 7/9/2020 5:11 AM

**Screenshots/Links**
Correct local time shown in the list of pull payments:
PullPaymentsCorrectTime

Incorrect local time shown in the display of the pull payment:
PullPaymentsTime

AlexGidge:
Closes #1725 by:

- Adding "Start Date" to View Pull Payments screen
- "Last Updated" relabelled to "Last Refreshed"
- "Last Refreshed" to display in local timezone

Note: Date times are formatted based on the server's date time culture info 

----
Before:
image

----
After:
image

### DIFF
--- a/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
+++ b/BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
@@ -89,9 +89,14 @@
                                 <h2 class="h4 mb-3">@Model.Title</h2>
                             }
                             <div class="d-flex align-items-center">
-                                <span class="text-muted text-nowrap">Last Updated</span>
+                                <span class="text-muted text-nowrap">Start Date</span>
                                 &nbsp;
-                                <span class="text-nowrap">@Model.LastUpdated.ToString("g")</span>
+                                <span class="text-nowrap">@Model.StartDate.ToBrowserDate()</span>
+                            </div>
+                            <div class="d-flex align-items-center">
+                                <span class="text-muted text-nowrap">Last Refreshed</span>
+                                &nbsp;
+                                <span class="text-nowrap">@Model.LastRefreshed.ToBrowserDate()</span>
                                 <button type="button" class="btn btn-link d-none d-lg-inline-block d-print-none border-0 p-0 ml-4 only-for-js" id="copyLink">
                                     Copy Link
                                 </button>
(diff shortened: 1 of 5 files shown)


britttttk:
tACK

BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
```diff
@@ -89,9 +89,14 @@
...
                             }
                             <div class="d-flex align-items-center">
-                                <span class="text-muted text-nowrap">Last Updated</span>
+                                <span class="text-muted text-nowrap">Start Date</span>
                                 &nbsp;
-                                <span class="text-nowrap">@Model.LastUpdated.ToString("g")</span>
+                                <span class="text-nowrap">@Model.StartDate.ToBrowserDate()</span>
+                            </div>
+                            <div class="d-flex align-items-center">
+                                <span class="text-muted text-nowrap">Last Refreshed</span>
```
Zaxounette, line 97:
```suggestion
                                <span class="text-muted text-nowrap">Last Updated</span>
```

Zaxounette (changes requested):
As a standalone string, `Last updated` indeed was confusing.
But with the addition of the string `Start date`, both together are clear.

In addition, `Last refreshed` can be confusing on the viewers standpoint, because it could refer to having refreshed the browser page, which has nothing to do with this.

pavlenex:
Is there a reason why we want to display `Last Refreshed` why does it matter to the user?

AlexGidge:
@Zaxounette I've reverted the field label back to `Last Updated` 

@pavlenex Does the date help the user if they look at the page and the pull payment is still showing an outdated amount/status?

NicolasDorier:
@AlexGidge the changes on ToBrowserDate seems to have broken something (`g` instead of `o`) seems to have broken date (does not appear as browser date anymore)

What was the rational for this change?

AlexGidge:
@NicolasDorier Sorry, that's my mistake - I didn't look into the formatting enough. I copied `g` from the existing HTML assuming that only impacted formatting not date localisation

I've reproduced the issue in my fork and confirmed the correct date is now displaying after your fix in 94acf60

PR #2340, opened 2021-03-08
Add DELETE action for api/v1/users endpoint

bolatovumar:
See discussion here for details: https://github.com/btcpayserver/btcpayserver/discussions/2304


BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +176,75 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
             return CreatedAtAction(string.Empty, model);
         }
 
+        [HttpDelete("~/api/v1/users/{userId}")]
+        [Authorize(Policy = Policies.CanCreateUser, AuthenticationSchemes = AuthenticationSchemes.GreenfieldAPIKeys)]
```
Kukks, line 180: This permission is definitely wrong.  I think we should create a new policy `CanDeleteUser` and then have the checks in there:
* If user id == current user && (API key permissions has `delete user`)
* if user is admin && (API key permissions has `modify server`)

Take a look at the AuthorizationHandler and accompanying files (follow the policies)

BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +176,75 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
...
+                return Forbid(AuthenticationSchemes.GreenfieldBasic);
+            }
+
+            // Ok, this user is an admin but there are other admins as well so safe to delete
+            await DeleteUserAndAssociatedData(userId, user);
+
+            return Ok();
+        }
+
+        private async Task DeleteUserAndAssociatedData(string userId, ApplicationUser user)
```
Kukks, line 218: I think there is already a delete action available somewhere in the Admin users list. We should make sure they use the same code (move it a service if need be and use said service here and in that page). That way we can add a delete user action in the account settings later

Kukks (changes requested):
Would also 
* need to be added to the swagger docs
* Need tests

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: Yeah, I wasn't sure if this was the right one to use. I will add a new one.

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: Yeah, this code replicates that other code. I can share the code so it's not repeated.

bolatovumar: @Kukks added a user service. Let me know if that's what you wanted or if I got something wrong.

bolatovumar:
> Would also
> 
> * need to be added to the swagger docs

Added docs.

> * Need tests

Will add tests once functionality is complete and @Kukks OKs it.

BTCPayServer/Controllers/GreenField/UsersController.cs
Kukks: Sorry missed this reply.

Yes, that's correct

BTCPayServer/Controllers/ManageController.APIKeys.cs
```diff
@@ -470,6 +470,7 @@ public class PermissionValueItem
                 {
                     {BTCPayServer.Client.Policies.Unrestricted, ("Unrestricted access", "The app will have unrestricted access to your account.")},
                     {BTCPayServer.Client.Policies.CanCreateUser, ("Create new users", "The app will be able to create new users on this server.")},
+                    {BTCPayServer.Client.Policies.CanDeleteUser, ("Delete users", "The app will be able to delete users on this server.")},
```
Kukks, line 473:
```suggestion
                    {BTCPayServer.Client.Policies.CanDeleteUser, ("Delete users", "The app will be able to delete users on this server (if an admin, else only your own user).")},
```

BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +179,60 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
...
+
+            var user = userId == null ? null : await _userManager.FindByIdAsync(userId);
+            if (user == null)
+            {
+                return NotFound();
+            }
+
+            var roles = await _userManager.GetRolesAsync(user);
+            // We can safely delete the user if it's not an admin user
+            if (!_userService.IsRoleAdmin(roles))
```
Kukks, line 201: I'm not 100% but i think you can get the Roles the user is in without and db rountrip.   The `User` object has the roles assigned to a user set in the greenfield authorization flow.  I think it's ` User.IsInRole()`

bolatovumar: This won't work because `User.IsInRole()` checks if current user is in a particular role but here we are checking if the user we are trying to delete is an admin.

BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
```diff
@@ -107,6 +107,48 @@
                     }
                 ]
             }
+        },
+        "/api/v1/users/{userId}": {
```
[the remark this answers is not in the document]
bolatovumar: @Kukks `"/api/v1/users/me"` already exists for `GET` requests up above. Or do you mean something else here?

BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
```diff
@@ -107,6 +107,48 @@
...
+                        "description": "Authorized but forbidden to delete the user. Can happen if you attempt to delete the only admin user."
+                    },
+                    "404": {
+                        "description": "User with provided ID was not found"
+                    }
+                },
+                "security": [
+                    {
+                        "API Key": [
+                            "btcpay.server.cancreateuser"
```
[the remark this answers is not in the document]
bolatovumar: @Kukks updated here: 0a7f52fa0fae7d4f93438e78f306011c7d326eae

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: @Kukks Ok, there is something weird going on here. If we want any non-admin user to be able to delete themselves then I need to update the permission from `"btcpay.server.candeleteuser"` to `"btcpay.user.candeleteuser"` as `"btcpay.server.*"` are forbidden for assignment to non-admin users (see here: https://github.com/btcpayserver/btcpayserver/blob/bbe1442c284f32a133aafb548756b403a144013e/BTCPayServer/Controllers/ManageController.APIKeys.cs#L439). However, if I update it to `"btcpay.user.candeleteuser"` then something like 
```C# 
var canDeleteUser = (await _authorizationService.AuthorizeAsync(User, null, new PolicyRequirement(Policies.CanDeleteUser))).Succeeded;
``` 
always returns `false` (doesn't matter if user is admin or not and whether the permission is assigned to the API key). I'm not sure why this is. Is there something special about `"btcpay.user.*"` permissions?

It seems that `var canDeleteUser = (await _authorizationService.AuthorizeAsync(User, null, new PolicyRequirement(Policies.CanDeleteUser))).Succeeded;` works for admin users (as in it can return `true` or `false` when appropriate) but not for non-admin users (always returns `false`).

Screen Shot 2021-04-04 at 5 09 12 PM
^^^ Always returns `false` for both permissions

Screen Shot 2021-04-04 at 5 10 19 PM
^^^ Returns correct values for both permissions

BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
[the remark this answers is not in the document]
Kukks: Yes, you can do a DELETE with the same route

BTCPayServer/Controllers/GreenField/UsersController.cs
Kukks: got it 👍

bolatovumar:
@Kukks does this look reasonable now? I updated the PR by adding the `DELETE` `"~/api/v1/users/me"` endpoint with the constraint of `CanDeleteUser` which any non-admin or admin user can use to delete themselves as long as they are not the only admin user on the service. Same for the `"~/api/v1/users/{userId}"` endpoint except it can only be used by admin users.

If this looks good I will add tests and whatnot.

BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +196,57 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
             return CreatedAtAction(string.Empty, model);
         }
 
+        [HttpDelete("~/api/v1/users/{userId}")]
+        [Authorize(AuthenticationSchemes = AuthenticationSchemes.GreenfieldAPIKeys)]
+        public async Task<ActionResult<ApplicationUserData>> DeleteUser(string userId)
+        {
+            // Only admins should be allowed to delete users
+            if (!User.IsInRole(Roles.ServerAdmin))
```
Kukks, line 204: use the policy for canmodifyserver to detect if it is an admin instead so that we are consistent with everywhere else

BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +196,57 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
...
+                return Forbid(AuthenticationSchemes.GreenfieldBasic);
+            }
+
+            // Ok, this user is an admin but there are other admins as well so safe to delete
+            await _userService.DeleteUserAndAssociatedData(user);
+
+            return Ok();
+        }
+
+        private async Task<Boolean> IsAdmin()
```
Kukks, line 235: Is this used?

Kukks (changes requested):
Got one final comment but yes it looks good to move forward with Client + tests

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: It's unused. Removed in this commit: ff363f99491050bcf7e4683107aa5ef09731f439

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: Addressed in this commit: 49f6eb3b1756fd03c0e6a69ea1a763976a949c10

bolatovumar:
> Got one final comment but yes it looks good to move forward with Client + tests

I added some basic tests for failure cases here: 08f9cf3a4e8e7ed0fd53a92b8a3a3d27edf36005 but for some reason I can't test the success case. 


This works:

```c#
[Fact(Timeout = TestTimeout)]
[Trait("Integration", "Integration")]
public async Task CanDeleteUsersViaApi()
{
    using (var tester = ServerTester.Create(newDb: true))
    {
        await tester.StartAsync();
        var unauthClient = new BTCPayServerClient(tester.PayTester.ServerUri);
        // Should not be authorized to perform this action
        await AssertHttpError(401,
            async () => await unauthClient.DeleteUser("lol user id"));

        var user = tester.NewAccount();
        user.GrantAccess();
        await user.MakeAdmin();
        var adminClient = await user.CreateClient(Policies.Unrestricted);
        // Should 404 if user doesn't exist
        await AssertHttpError(404,
            async () => await adminClient.DeleteUser("lol user id"));
    }
}
```

But as soon as I attempt to delete a user I get a **405 Method Not Allowed** status code for the final assertion which is strange:

```c#
[Fact(Timeout = TestTimeout)]
[Trait("Integration", "Integration")]
public async Task CanDeleteUsersViaApi()
{
    using (var tester = ServerTester.Create(newDb: true))
    {
        await tester.StartAsync();
        var unauthClient = new BTCPayServerClient(tester.PayTester.ServerUri);
        // Should not be authorized to perform this action
        await AssertHttpError(401,
            async () => await unauthClient.DeleteUser("lol user id"));

        var user = tester.NewAccount();
        user.GrantAccess();
        await user.MakeAdmin();
        var adminClient = await user.CreateClient(Policies.Unrestricted);
        // Should 404 if user doesn't exist
        await AssertHttpError(404,
            async () => await adminClient.DeleteUser("lol user id"));
        
        // Should be able to delete a user successfully
        var userToDelete = tester.NewAccount();
        var response = await adminClient.DeleteUser(userToDelete.UserId);
        Assert.IsType<OkResult>(response); 
    }
}
```

Screen Shot 2021-04-11 at 11 35 00 AM

Any idea why this might be?

BTCPayServer/Controllers/GreenField/UsersController.cs
```diff
@@ -165,6 +196,36 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
...
         }
 
+        [HttpDelete("~/api/v1/users/{userId}")]
+        [Authorize(Policy = Policies.CanModifyServerSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        public async Task<IActionResult> DeleteUser(string userId)
+        {
+            var user = await _userManager.FindByIdAsync(userId);
+            if (user == null)
+            {
+                return NotFound();
```
NicolasDorier, line 206: Do not use naked NotFound like this, see how I did in for example `InvoiceNotFound`

NicolasDorier:
@bolatovumar should we put this one in next release?

BTCPayServer/Controllers/GreenField/UsersController.cs
bolatovumar: @NicolasDorier done in this commit: 192cbc9541f1a77398c8530a43d55ea18b1d50f8

bolatovumar:
> @bolatovumar should we put this one in next release?

@NicolasDorier yeah, if it gets reviewed by then.

NicolasDorier:
Code seems OK. I don't really like the `UserService`, but maybe can't do better. Will think a bit, then merge.

bolatovumar:
@NicolasDorier alright, let me know if you want me to make any changes. I added `UserService` because @Kukks asked me to add it to reuse code. 

In the meantime I rebased the PR on top of latest `master` branch.

Kukks (changes requested):
Ok one last thing from my end: There is a missing documentation of the new permission that was added in swagger. If you run the test `EnsureSwaggerPermissionsDocumented ` locally, it will tell you how to fix the docs

Joker11156o:
tisdag 1 juni 2021 skrev Andrew Camilleri ***@***.***>:

bolatovumar:
@Kukks ok updated the description in ca3f97c and tests are passing now.

Kukks (approved):
Did some small refactoring and improved tests coverage

PR #2366, opened 2021-03-15
Fix POS item newline break

### LINKED ISSUE #2365: Multiple paragraphs in description breaks items in POS app
**Describe the bug**
When editing or adding an item to the POS app, if the description has more than 1 paragraph, it breaks the database layout. Effectively breaking the item and creating an item per paragraph with corrupted values.

**To Reproduce the bug**
1. Go to "Apps"
2. Click on `Settings` of a "Point of Sale"
3. Add or edit an item
4. Add or edit the description of the item with a text that has 3 paragraphs.
5. Click `Save Settings`

**Expected behavior**
The description of edited item has 3 paragraphs

**Actual Behavior**

2 new items with broken values appear.

**screenshot**
image

- BTCPay Server Version:    v1.0.6.2
- Deployment Method:  https://mynodebtc.com/
- Browser: Firefox

Kukks:
fixes #2365

### DIFF
--- a/BTCPayServer/Views/Apps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Apps/TemplateEditor.cshtml
@@ -31,7 +31,6 @@
                     <div v-if="anyImages" class="card-img-top border-bottom" v-bind:style="getImage(item)"></div>
                     <div class="card-body">
                         <h6 class="card-title" v-html="item.title"></h6>
-                        <p v-text="item.description" style="white-space: pre;"></p>
                         <button type="button" class="btn btn-primary" v-on:click="editItem(index)">Edit</button>
                         <button type="button" class="btn btn-danger" v-on:click="removeItem(index)">
                             <i class="fa fa-trash"></i>
--- a/BTCPayServer/Views/Apps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Apps/TemplateEditor.cshtml
@@ -214,7 +214,10 @@ $(function() {
                             if (description.startsWith('"') && description.endsWith('"')){
                                 description = description.substr(1, description.length-2);
                             }
-                            description.replaceAll("<br/>", "\n").replaceAll('\"', '"')
+                            description = description
+                            .replaceAll("<br>", "\n")
+                            .replaceAll("<br/>", "\n")
+                            .replaceAll('\\"', '"')
                         }
                         if (productProperty.indexOf('image:') !== -1) {
                             image = productProperty.replace('image:', '').trim();
--- a/BTCPayServer/Views/Apps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Apps/TemplateEditor.cshtml
@@ -211,9 +211,10 @@ $(function() {
                         }
                         if (productProperty.indexOf('description:') !== -1) {
                             description =productProperty.replace('description:', '').trim();
-                            if (description && description.startsWith('"') && description.endsWith('"')){
+                            if (description.startsWith('"') && description.endsWith('"')){
                                 description = description.substr(1, description.length-2);
                             }
+                            description.replaceAll("<br/>", "\n").replaceAll('\"', '"')
                         }
                         if (productProperty.indexOf('image:') !== -1) {
                             image = productProperty.replace('image:', '').trim();
@@ -239,7 +240,7 @@ $(function() {
                             title: title,
                             price: price,
                             image: image || null,
-                            description:  description? description.replaceAll("<br/>", "\n") : '',
+                            description:  description || '',
                             custom: custom === "true",
                             buyButtonText: buyButtonText,
                             inventory: isNaN(inventory)? null: inventory,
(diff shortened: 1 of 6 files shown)


BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -31,6 +31,7 @@
                     <div v-if="anyImages" class="card-img-top border-bottom" v-bind:style="getImage(item)"></div>
                     <div class="card-body">
                         <h6 class="card-title" v-html="item.title"></h6>
+                        <p v-text="item.description" style="white-space: pre;"></p>
```
dennisreimann, line 34: The `white-space: pre;` leads to strange side effect :)

pre

dennisreimann (changes requested):

BTCPayServer/Views/Apps/TemplateEditor.cshtml
Kukks: Removed it altogether, this was just scope creep :)

dennisreimann:
I think without proper YAML parsing we are setting us up for problems with edge cases here. For instance if the string contains `"`, it'll also break – a description like this will result in an invalid template:

```
This is "THE BEST TEA EVAR" someone said.

I tend to agree. 
```

With a YAML parser we could use [multiline YAML](https://yaml-multiline.info/) and get rid of the other hacks as well. Otherwise we need to at least patch this by escaping the quotes `\"` … not optimal.

Kukks:
> I think without proper YAML parsing we are setting us up for problems with edge cases here. For instance if the string contains `"`, it'll also break – a description like this will result in an invalid template:
> 
> ```
> This is "THE BEST TEA EVAR" someone said.
> 
> I tend to agree. 
> ```
> 
> With a YAML parser we could use [multiline YAML](https://yaml-multiline.info/) and get rid of the other hacks as well. Otherwise we need to at least patch this by escaping the quotes `\"` … not optimal.

If anything, we should get rid of the YAML parser and switch over to JSON. YAML is a superset of JSON anyway so it shouldn't be that hard. We're also already hiding the raw editor now so any benefits of yml readability has gone a long time ago.

Kukks:
> I think without proper YAML parsing we are setting us up for problems with edge cases here. For instance if the string contains `"`, it'll also break – a description like this will result in an invalid template:
> 
> ```
> This is "THE BEST TEA EVAR" someone said.
> 
> I tend to agree. 
> ```
> 
> With a YAML parser we could use [multiline YAML](https://yaml-multiline.info/) and get rid of the other hacks as well. Otherwise we need to at least patch this by escaping the quotes `\"` … not optimal.

I had actually initially added multiline yaml support but it ended up being messy and it affected many parts.

dennisreimann:
+ 1 for switching to JSON.

Kukks:
> I think without proper YAML parsing we are setting us up for problems with edge cases here. For instance if the string contains `"`, it'll also break – a description like this will result in an invalid template:
> 
> ```
> This is "THE BEST TEA EVAR" someone said.
> 
> I tend to agree. 
> ```
> 
> With a YAML parser we could use [multiline YAML](https://yaml-multiline.info/) and get rid of the other hacks as well. Otherwise we need to at least patch this by escaping the quotes `\"` … not optimal.

addressed this issue

dennisreimann (approved):

NicolasDorier:
do not use `@Safe.Raw` for things that are not raw html.

BTCPayServer/Views/AppsPublic/PointOfSale/Cart.cshtml
```diff
@@ -242,18 +242,18 @@
...
-                            @:<img class="card-img-top" src="@image" alt="Card image cap" asp-append-version="true">
+                            @:<img class="card-img-top" src="@Safe.Raw(image)" alt="Card image cap" asp-append-version="true">
                         }
                         <div class="card-body p-3">
-                            <h6 class="card-title mb-0">@item.Title</h6>
+                            <h6 class="card-title mb-0">@Safe.Raw(item.Title)</h6>
                             @if (!String.IsNullOrWhiteSpace(description))
                             {
-                                <p class="card-text">@description</p>
+                                <p class="card-text">@Safe.Raw(System.Net.WebUtility.HtmlDecode(description))</p>
```
NicolasDorier, line 251: Remove HtmlDecode, if the goal is to put HTML, just `Safe.Raw`

BTCPayServer/Views/AppsPublic/PointOfSale/Static.cshtml
```diff
@@ -21,17 +21,23 @@
...
                     {
-                        <img class="card-img-top" src="@item.Image" alt="Card image cap" asp-append-version="true">
+                        <img class="card-img-top" src="@Safe.Raw(item.Image)" alt="Card image cap" asp-append-version="true">
                     }
                     <div class="card-body pb-0">
-                        <h5 class="card-title">@item.Title</h5>
+                        <h5 class="card-title">@Safe.Raw(item.Title)</h5>
                         @if (!String.IsNullOrWhiteSpace(item.Description))
                         {
-                            <p class="card-text">@System.Net.WebUtility.HtmlDecode(item.Description)</p>
```
NicolasDorier, line 34: Remove HtmlDecode, if the goal is to put HTML, just Safe.Raw

NicolasDorier:
Do not use `Safe.Raw` for things that are not meant to be html

NicolasDorier:
I think it breaks a test `CanCreateAppPoS `

Kukks:
> I think it breaks a test `CanCreateAppPoS `

Fixed, it was actually an unrelated bug I uncovered...

PR #2375, opened 2021-03-17
GreenField: Add FeeRate To Wallets API

### LINKED ISSUE #1846: Extend API: feerate
As described on [Mattermost](https://chat.btcpayserver.org/btcpayserver/pl/xxs5hyt7zbypfgn3ndnx5oes6w) - I'm trying to enrich the checkout page with some fee information.

Like: "if you pay with On-Chain funds (from a segwith wallet): your transaction costs could on average be: X.YZ €"
or: "If you pay with On-Chain funds (from a legacy wallet): your transaction costs would be on average: ...
or: "If you use lightning expect to pay somewhere around 0-3% (depending on your routing and proximity to our node)"

I would need an /api/feerate endpoint. Which expects a store_id to send me back the current feerate per vB for the way my store is configured (i.e. estimates based on next block or next X blocks, etc.) - preferably without AUTH/public facade and as json output so I can include it in my frontend html/javascript.

I have checked the Greenfield and Legacy API docs and couldn't find any information related to feerate. If there is such a feature, please feel free to close this request.

Kukks:
closes #1846

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
@@ -44,6 +44,7 @@ namespace BTCPayServer.Controllers.GreenField
         private readonly DelayedTransactionBroadcaster _delayedTransactionBroadcaster;
         private readonly EventAggregator _eventAggregator;
         private readonly WalletReceiveService _walletReceiveService;
+        private readonly IFeeProviderFactory _feeProviderFactory;
 
         public StoreOnChainWalletsController(
             IAuthorizationService authorizationService,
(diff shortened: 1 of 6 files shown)


bolatovumar (approved):
The code looks good but I'm wondering if it would make more sense as its own endpoint? It's not immediately obvious that this endpoint would return a fee rate. If I were the user of the Greenfield API I would probably expect to find a separate fee rate endpoint.

NicolasDorier:
I agree with @bolatovumar, it should be its own endpoint with configurable target.

Kukks:
@bolatovumar  @NicolasDorier  switched to separate endpoint

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -81,12 +86,28 @@ public async Task<IActionResult> ShowOnChainWalletOverview(string storeId, strin
                 out DerivationSchemeSettings derivationScheme, out IActionResult actionResult)) return actionResult;
 
             var wallet = _btcPayWalletProvider.GetWallet(network);
+            var feeRateTarget = Store.GetStoreBlob().RecommendedFeeBlockTarget;
```
bolatovumar, line 89: It seems that this variable is unused?

Kukks: removed, thanks!

Kukks:
8

PR #2377, opened 2021-03-19
BetterFormFileHandling

Kukks:


NicolasDorier:
Argh sorry I was working on the same at the same time.
That said there is no tests, and this PR does not work! Surprisingly / or \ are not in the list of invalid chars on linux for filename.... I don't know the fuck why. Anyway I fixed it.

PR #2378, opened 2021-03-19
UI: Header and navigation improvements

dennisreimann:
Center aligning the main menu and giving the items and active state, as proposed by @dstrukt in #2364.

We can adjust the colors once we get the Bootstrap 5 migration done, as this will include the updated color palette.

mobile

light

dark

casa

### DIFF
--- a/BTCPayServer/Components/NotificationsDropdown/Default.cshtml
+++ b/BTCPayServer/Components/NotificationsDropdown/Default.cshtml
@@ -2,6 +2,7 @@
 @inject UserManager<ApplicationUser> UserManager
 @inject CssThemeManager CssThemeManager
 @using BTCPayServer.HostedServices
+@using BTCPayServer.Views.Notifications
 @using Microsoft.AspNetCore.Http.Extensions
 @model BTCPayServer.Components.NotificationsDropdown.NotificationSummaryViewModel
 @addTagHelper *, BundlerMinifier.TagHelpers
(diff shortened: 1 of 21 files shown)


dstrukt:
😍 Lookin good!

Should we add "server settings" to the end of the list in this instance? 

Additionally, I've always wondered if it should be an icon and moved to the right, but I figured there might be some context around that one.

I have some additional suggestions for our mobile view, but will mock those up for a later time.

dennisreimann:
I tend to agree, though I see the need for having it displayed prominently for the initial setup. Once that's done I can understand why having it as an icon (similar and next to the account settings) would make sense. Let's discuss this one :)

dstrukt:
Sounds good -- will add it to the list for the next design meeting.

pavlenex:
Concept ACK for the header menu being centered.

For Server Settings re-position/icon, I don't feel it's a good idea to change it at this time, as it would require a significant change in the documentation and simply will be odd for the users who're already used to it being first.

dstrukt:
I'm happy to leave it for now, but this might not be the last time you hear from me on that issue :)

pavlenex (approved):
tACK. Great change 🚀 🚀

dennisreimann:
I got #2362 almost done. As it builds on this: Do we want to include it here or first merge this and do it as a separate PR? I'd prefer the latter …

pavlenex:
Let's do a separate PR @dennisreimann since that's what you prefer 🏗️

dennisreimann:
Rebased and good to go :)

pavlenex (approved):
tACK

dennisreimann:
Rebased. Once this gets merged after the upcoming release, I'll send the PR for #2362, which is based on this one.

PR #2384, opened 2021-03-21
Abstract Store integrations

Kukks:
This decouples the shopify integration from the store and is the first step to converting it into a plugin + allows other plugins to show up as an available store integration.

image
image

No breaking changes, everything should stay working.


pavlenex:
Interesting approach, I like it. 

Concept ACK.

dstrukt:
I think the UI for plugins / integrations is definitely something we'll want to look at / develop a system for as we get a better sense of how everything will work, but it's great to see the current UI repurposed!

Always happy to share thoughts or give feedback in the meantime as well.

dennisreimann:
I think we should move the views from `Shared/` to the plugin's folder and add something like this to `Startup`, so that they can get resolved:

```csharp
.AddRazorOptions(o =>
            {
                // /Components/{View Component Name}/{View Name}.cshtml
                o.ViewLocationFormats.Add("/{0}.cshtml");
                o.PageViewLocationFormats.Add("/{0}.cshtml");
                o.ViewLocationFormats.Add("/Plugins/{0}.cshtml");
                o.PageViewLocationFormats.Add("/Plugins/{0}.cshtml");
            })
```

Tested this and it works. Having the views encapsulated with the rest of the plugin would be nice.

Kukks:
> I think we should move the views from `Shared/` to the plugin's folder and add something like this to `Startup`, so that they can get resolved:
> 
> ```cs
> .AddRazorOptions(o =>
>             {
>                 // /Components/{View Component Name}/{View Name}.cshtml
>                 o.ViewLocationFormats.Add("/{0}.cshtml");
>                 o.PageViewLocationFormats.Add("/{0}.cshtml");
>                 o.ViewLocationFormats.Add("/Plugins/{0}.cshtml");
>                 o.PageViewLocationFormats.Add("/Plugins/{0}.cshtml");
>             })
> ```
> 
> Tested this and it works. Having the views encapsulated with the rest of the plugin would be nice. The reference would then be
> 
> ```cs
> applicationBuilder.AddSingleton<IUIExtension>(new UIExtension("Shopify/Views/StoreIntegrationShopifyOption",
>                 "store-integrations-list"));
> ```

This is only short term, they will be moved to their own csproj in a separate PR

BTCPayServer/Plugins/Shopify/ShopifyController.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            return _cachedShopifyJavascript;
+        }
+
+        [AllowAnonymous]
+        [HttpGet("stores/{storeId}/integrations/shopify/shopify.js")]
+        public async Task<IActionResult> ShopifyJavascript(string storeId)
+        {
+            var jsFile =
+                $"var BTCPAYSERVER_URL = \"{Request.GetAbsoluteRoot()}\"; var STORE_ID = \"{storeId}\"; {await GetJavascript()}";
```
NicolasDorier, line 96: need to escape both `Request.GetAbsoluteRoot()` and `storeId` with htmlHelper.

NicolasDorier:
Merging this for now, we can improve later.

PR #2388, opened 2021-03-22
Refactor TorService

junderw:
This will make it easier to add the extra config elements,

I think we should make the config like:
`BTCPAY_TOR_SERVICES=serviceName1,onionHost1,virtualPort1:serviceName2,onionHost2,virtualPort2`
So it can be parsed by splitting by `:` then splitting by `,` and making sure each service has 3 elements and the 3rd element should be parsed from string to int.

This will make it easier to pass values for node management software that makes use of docker-compose.

See https://github.com/getumbrel/umbrel/issues/607

@Kukks I think this is a first step to what we were talking about.


Kukks:
I've pushed directly to this PR and added the functionality requested. Slightly different syntax than what you requested but  we already use this format for non-onion related tor services

junderw:
LGTM logic-wise

BTCPayServer/Services/TorServices.cs
```diff
@@ -2,110 +2,157 @@
...
-            network = _networks.GetNetwork<BTCPayNetworkBase>(splitted[0]);
-            return network != null;
+        }
+
+        internal override Task[] InitializeTasks()
+        {
+            return new[] {CreateLoopTask(Refresh)};
+        }
+
+        private Task LoadFromConfig()
```
NicolasDorier, line 145: 1. Why `Task`
2. Why ignoring `torcservices` if `torrcfile` is set?

NicolasDorier:
The `torservices` should be documented in the `DefaultConfiguration` and be parsed into the BTCPayServerOptions as well.

rockstardev:
@junderw did you allow maintainers to add commits to your PR (see screenshot for setting)?

junderw:
yes, that is why Kukks was able to push commits.

Kukks:
> yes, that is why Kukks was able to push commits.

Are you sure it's not because we are the same person with split personalities that activate based on timezone?

BTCPayServer/Services/TorServices.cs
Kukks: Removed task, was leftover from before I refactored.

Kukks: I ignore since they  might define configs that conflict with one another

junderw:
> Are you sure it's not because we are the same person with split personalities that activate based on timezone?

Or it could be your ssh key in my authorized_keys :thinking:

rockstardev:
OK, it was my bad... I've set everything up and pushed my commits @junderw & @Kukks 

Watch it all happen on: https://www.youtube.com/watch?v=R3WxF1NQsOU ~43 mins

junderw:
LGTM

utACK 5b9dac5c455ef7e01b1ebb2a3717669e6bc811b3

NicolasDorier:
Can we just add that the new config is mutually exclusive with `torrcfile` in the doc? And throw a proper exception if both are specified?

junderw:
Is the documentation addition a part of this repo? I couldn't figure out where the doc change needs to be made.

junderw:
test failures unrelated.

Kukks:
@NicolasDorier  https://www.youtube.com/watch?v=52vsExtoeC0

NicolasDorier:
wolololo

PR #2390, opened 2021-03-23
Seaparate CoinSwitch as a plugin

Kukks:
This partially decouples coinswitch  from BTCPay. There are some some dependencies which require the code to remain inside the main codebase, but we can abstract that in a later PR. This PR includes #2384 as it makes Coinswitch appear as an option in the Integrations page.


dstrukt:
😍

Eskyee:
💯

BTCPayServer.Tests/CoinSwitchTests.cs
```diff
@@ -1,91 +1,91 @@
...
-
-                store = await tester.PayTester.StoreRepository.FindStore(user.StoreId);
-
-                Assert.False(store.GetStoreBlob().CoinSwitchSettings.Enabled);
-            }
-        }
-
-    }
-}
+// using System.Threading.Tasks;
```
NicolasDorier, line 1: remove those?

pavlenex:
@Kukks What's the goal with this one?
Extract it as a system plugin, meaning I can't disable it at all?
I don't think this should be a system plugin, but I assume you're doing this to not break stuff, right?

Screenshot 2021-03-31 at 19 08 31

Kukks:
> ~@Kukks What's the goal with this one?~
> ~Extract it as a system plugin, meaning I can't disable it at all? I don't think this should be a system plugin, but I assume you're doing this to not break stuff, right?~
> 
> NVM. I now understand it. It's a cool concept, would be great to further abstract this in the future.
> 
> Screenshot 2021-03-31 at 19 08 31
> 
> Screenshot 2021-03-31 at 19 11 34
> 
> Concept ACK.

Yes, it's a system plugin until I can develop the plugin system further enough to completely make it opt in. Also need to think about how to migrate users already using this feature.

BTCPayServer.Tests/CoinSwitchTests.cs
rockstardev: ae93cfa

NicolasDorier:
Nice

PR #2391, opened 2021-03-24
Activate CSP

NicolasDorier:
Attempt to activate some limited CSP to prevent XSS vulnerabilities to escalate too much.
This also fix a bug with the previous root domain mapping, where HTTP headers were not properly set if using root domain mapping to a crowdfund or pos app.

### DIFF
--- a/BTCPayServer/Controllers/AppsPublicController.cs
+++ b/BTCPayServer/Controllers/AppsPublicController.cs
@@ -57,8 +57,10 @@ namespace BTCPayServer.Controllers
         }
         
         [HttpGet]
+        [Route("/")]
         [Route("/apps/{appId}/pos/{viewType?}")]
         [XFrameOptionsAttribute(XFrameOptionsAttribute.XFrameOptions.AllowAll)]
+        [DomainMappingConstraint(AppType.PointOfSale)]
         public async Task<IActionResult> ViewPointOfSale(string appId, PosViewType? viewType = null)
         {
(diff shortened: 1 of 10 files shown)


BTCPayServer/Controllers/AppsPublicController.cs
```diff
@@ -104,10 +107,13 @@ public async Task<IActionResult> ViewPointOfSale(string appId, PosViewType? view
         }
 
         [HttpPost]
+        [ContentSecurityPolicyAttribute(Enabled = false)]
+        [Route("/")]
```
Kukks, line 111: This route set on all actions might cause MVC to crash due to a collision

NicolasDorier: no it does not. Because we use `DomainMappingConstraint`

NicolasDorier: MVC select the route by enumerating all, and for each route calculating the constraints. Problem arise if after applying the constraints, there is more than one route.

In this case, this is not the case.

NicolasDorier:
I will not merge this, this does not protect agaisnt XSS and risk of breaking too many things.

NicolasDorier:
I will probably take parts of this PR though.

pavlenex:
@NicolasDorier is this one ready to be closed then?

NicolasDorier:
yeah... might revist later, but this PR might introduce bugs and does not really protect everything...

PR #2395, opened 2021-03-25
Remove GetAwaiter from tests

NicolasDorier:
This is probably the reason why tests are sometimes flaky.


NicolasDorier:
Will probably not merge that, still have timeout.

PR #2401, opened 2021-03-27
Document and handle Invoice Metadata better

Kukks:
PosData would crash in certain scenarios when created via API. Invoice metadata known fields were not completely documented

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -2,15 +2,12 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations.Schema;
 using System.Linq;
-using Amazon.Runtime.Internal.Util;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.JsonConverters;
 using BTCPayServer.Models;
 using BTCPayServer.Payments;
 using BTCPayServer.Payments.Bitcoin;
-using Microsoft.AspNetCore.Http.Extensions;
-using Microsoft.CodeAnalysis;
 using NBitcoin;
 using NBitcoin.DataEncoders;
 using NBitpayClient;
@@ -18,11 +15,10 @@ using NBXplorer;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Newtonsoft.Json.Serialization;
-using YamlDotNet.Core.Tokens;
-using YamlDotNet.Serialization.NamingConventions;
 
 namespace BTCPayServer.Services.Invoices
 {
+
     public class InvoiceMetadata
     {
         public static readonly JsonSerializer MetadataSerializer;
@@ -64,7 +60,30 @@ namespace BTCPayServer.Services.Invoices
 
         [JsonProperty(PropertyName = "taxIncluded", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public decimal? TaxIncluded { get; set; }
-        public string PosData { get; set; }
+
+        [JsonIgnore]
+        public string PosData
+        {
+            get
+            {
+                return PosRawData?.ToString();
+            }
+            set
+            {
+                try
+                {
+
+                    PosRawData = JToken.Parse(value);
+                }
+                catch (Exception )
+                {
+                    PosRawData = JToken.FromObject(value);
+                }
+            }
+        }
+
+        [JsonProperty(PropertyName = "posData")]
+        public JToken PosRawData { get; set; }
         [JsonExtensionData]
         public IDictionary<string, JToken> AdditionalData { get; set; }
 
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -70,15 +70,22 @@ namespace BTCPayServer.Services.Invoices
             }
             set
             {
-                try
+                if (value is null)
                 {
-
-                    PosRawData = JToken.Parse(value);
+                    PosRawData = JValue.CreateNull();
                 }
-                catch (Exception )
+                else
                 {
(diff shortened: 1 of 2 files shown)


bolatovumar (approved):

NicolasDorier:
I don't think we should document legacy stuff. Specially fields that does not impact any logic anywhere.

Kukks:
> I don't think we should document legacy stuff. Specially fields that does not impact any logic anywhere.

There were users that specifically could not switch to use the new api because they needed to keep the same data structure to not break their system. The feature is actually there but was not documented anywhere

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -64,7 +60,37 @@ static InvoiceMetadata()
...
+                    {
+                        PosRawData = JToken.FromObject(value);
+                    }
+                }
+               
+            }
+        }
+
+        [JsonProperty(PropertyName = "posData")]
+        public JToken PosRawData { get; set; }
```
NicolasDorier, line 93: why should we expose the fact it is a `JToken`?

NicolasDorier:
@Kukks can we keep PosData a string like it always was?

Also in this case if you insist of keeping the old fields, at least mention they are legacy and there is no logic associated with it.

Kukks:
The pos Data has a special feature of detecting Json and rendering it in the invoice details UI in a human readable way (not just Json ident but also attempts to generate a full UI)

NicolasDorier:
posData is not necessarily JSON, so this is probably a breaking change. What happen if you desererialize a posData which can't be pasrsed as a JToken? 

The fact that the UI attempt to parse it as a JSON is a UI concern, not a concern for how we store this data.

Kukks:
> posData is not necessarily JSON, so this is probably a breaking change. What happen if you desererialize a posData which can't be pasrsed as a JToken?

Before this PR, creating an invoice with:
```
metadata: {
  "posData" : {...}
}
```
threw an exception

with this PR, if the posData is set to a string, it will give you a JToken of type String

In the eyes of Bitpay API and UI, `PosData` property is still string, we just use an intermediary `PosRawData` to load the raw jtoken value






> The fact that the UI attempt to parse it as a JSON is a UI concern, not a concern for how we store this data.

Yes, I agree. but the GreenField API does not have restrictions on the format, and if we try to parse the metadata for a bitpay invoice api/ btcpay UI, it will crash due to a mismatch. 
Thinking about it, we should probably apply similar logic to all the other well-known metadata properties

NicolasDorier:
```
Before this PR, creating an invoice with:

metadata: {
  "posData" : {...}
}
```

Ah shit I did not thought about this. But this is a problem on all the other properties...

What about just changing all the properties to `JToken`? and just putting the burden on the consumer of those properties the priviledge to cast it correctly?

Should not be that hard, I think only OrderId and PosData would be impacted.

Kukks:
> ```
> Before this PR, creating an invoice with:
> 
> metadata: {
>   "posData" : {...}
> }
> ```
> 
> Ah shit I did not thought about this. But this is a problem on all the other properties...
> 
> What about just changing all the properties to `JToken`? and just putting the burden on the consumer of those properties the priviledge to cast it correctly?
> 
> Should not be that hard, I think only OrderId and PosData would be impacted.
> Can add extension method to JToken so you can do `OrderId.AsString()` which returns string if it is JValue string.

I think we can fix this gracefully. Will give it a try

ratacibernetica:
For completeness sake, in version 1.0.7.0 (comes with Umbrel as of today) this request throws a 500 Internal Server Error:

```json
{
    "amount": 10,
    "currency": "MXN",
    "itemDesc": "Derp product 10",
    "itemCode": "DERP010",
    "metadata": {
        "orderId": "210700397785",
        "posData":
        {
            "TransID": "210700397785",
            "Fecha": "2021-07-06 23:00:30",
            "Carrier": "Derp",
            "Telefono": "5555555505"
        }
    }
}
```

And this one works as expected:
```json
{
    "amount": 10,
    "currency": "MXN",
    "itemDesc": "Derp product 10",
    "itemCode": "DERP010",
    "metadata": {
        "orderId": "210700397785",
        "posRawData":
        {
            "TransID": "210700397785",
            "Fecha": "2021-07-06 23:00:30",
            "Carrier": "Derp",
            "Telefono": "5555555505"
        }
    }
}
```

the OpenAPI spec `v1`  says that `posData` could be either a `string` or a JSON object. But if you pass an Object, it crashes. Should use `posRawData` if storing an object is the intention.

image

Kukks:
> For completeness sake, in version 1.0.7.0 (comes with Umbrel as of today) this request throws a 500 Internal Server Error:
> 
> ```json
> POST api/v1/stores/{store_id}/invoices
> {
>     "amount": 10,
>     "currency": "MXN",
>     "itemDesc": "Derp product 10",
>     "itemCode": "DERP010",
>     "metadata": {
>         "orderId": "210700397785",
>         "posData":
>         {
>             "TransID": "210700397785",
>             "Fecha": "2021-07-06 23:00:30",
>             "Carrier": "Derp",
>             "Telefono": "5555555505"
>         }
>     }
> }
> ```
> 
> And this one works as expected:
> 
> ```json
> POST api/v1/stores/{store_id}/invoices
> {
>     "amount": 10,
>     "currency": "MXN",
>     "itemDesc": "Derp product 10",
>     "itemCode": "DERP010",
>     "metadata": {
>         "orderId": "210700397785",
>         "posRawData":
>         {
>             "TransID": "210700397785",
>             "Fecha": "2021-07-06 23:00:30",
>             "Carrier": "Derp",
>             "Telefono": "5555555505"
>         }
>     }
> }
> ```
> 
> the OpenAPI spec `v1` says that `posData` could be either a `string` or a JSON object. But if you pass an Object, it crashes. Should use `posRawData` if storing an object is the intention.
> 
> image

Yes, this has been resolved. Umbrel is urged to stay up to date as 1.0.7.0 is quite old now

PR #2425, opened 2021-04-03
Allow Payjoin for wallet receive addresses

Kukks:
This PR 
* enables payjoin support for wallet receive addresses not tied to invoices. 
* fixes a bug in Wallet Send where if a bip21 did not have an amount it would fail
* adds bip21 to wallet address greenfield api

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
@@ -23,7 +23,6 @@ using NBitcoin.Payment;
 using NBXplorer;
 using NBXplorer.Models;
 using Newtonsoft.Json.Linq;
-using Org.BouncyCastle.Ocsp;
 using StoreData = BTCPayServer.Data.StoreData;
 
 namespace BTCPayServer.Controllers.GreenField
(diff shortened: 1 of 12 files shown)


BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -22,6 +23,7 @@
 using NBXplorer;
 using NBXplorer.Models;
 using Newtonsoft.Json.Linq;
+using Org.BouncyCastle.Ocsp;
```
NicolasDorier, line 26: remove this

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -122,9 +124,18 @@ public async Task<IActionResult> GetOnChainWalletReceiveAddress(string storeId,
             {
                 return BadRequest();
             }
+
+            var bip21 = network.GenerateBIP21(kpi.Address.ToString(), null);
+            var allowedPayjoin = derivationScheme.IsHotWallet && Store.GetStoreBlob().PayJoinEnabled;
+            if (allowedPayjoin)
+            {
+               bip21 +=
+                   $"?{PayjoinClient.BIP21EndpointKey}={Request.GetAbsoluteUri(Url.Action("Submit", "PayJoinEndpoint", new {cryptoCode}))}";
```
NicolasDorier, line 133: Can you use `nameof()` for Submit?

NicolasDorier:
Nice, code review ACK, though I think you should add in the API a way to disable payjoin for the newly generated address.

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
Kukks: done

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
Kukks: done

MaiHistoryCompanyLimited:
{#2425}

Kukks:
> {#2425}

Stop spamming all out PRs.

Kukks:
> Nice, code review ACK, though I think you should add in the API a way to disable payjoin for the newly generated address. (and default to payjoin enabled)

We can do this if someone requests it IMO. It currently checks if the user has payjoin enabled in their wallet

PR #2432, opened 2021-04-05
Fix hot wallet docs link

dennisreimann:
Now that btcpayserver/btcpayserver-doc#884 has been merged the docs link changed.

### DIFF
--- a/BTCPayServer/Views/Server/Policies.cshtml
+++ b/BTCPayServer/Views/Server/Policies.cshtml
@@ -25,9 +25,7 @@
         <div class="form-check my-1">
             <input asp-for="AllowHotWalletForAll" type="checkbox" class="form-check-input"/>
             <label asp-for="AllowHotWalletForAll" class="form-check-label"></label>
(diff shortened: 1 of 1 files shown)


britttttk:
@dennisreimann I would still link to the section that was used before https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/CreateWallet.md#requirements-to-create-wallets

britttttk:
Also is there wallet docs section which can clarify the policy: `Allow non-admins to import hot wallets for their stores`

Is that just importing seed? Should we link to this or add info about this policy in this section?
https://github.com/btcpayserver/btcpayserver-doc/blob/master/docs/ConnectWallet.md#enter-wallet-seed

dennisreimann:
@britttttk I've updated the existing link to the one you proposed. For the other policy I think we should come up with a new docs part, because the Enter Seed one doesn't work for that case imho.

pavlenex (approved):

britttttk (approved):

PR #2439, opened 2021-04-07
Add button to copy API key to clipboard

bolatovumar:
Every time I create an API key I have to select the text manually to copy it. BORING! 

Now with this PR you will get a button which will copy the key for you. WOW!

https://user-images.githubusercontent.com/1934678/113802661-f1163480-970f-11eb-83b7-1fb0bf7053da.mov

### DIFF
--- a/BTCPayServer/Views/Manage/APIKeys.cshtml
+++ b/BTCPayServer/Views/Manage/APIKeys.cshtml
@@ -13,7 +13,7 @@
     <thead>
     <tr>
         <th>Label</th>
-        <th>Key</th>
+        <th class="w-125px">Key</th>
         <th>Permissions</th>
         <th class="text-right">Actions</th>
     </tr>
@@ -28,10 +28,13 @@
             <td>@keyData.Label</td>
             <td>
                 <code class="hide-when-js">@keyData.Id</code>
-                <code class="only-for-js cursor-pointer" data-reveal="@keyData.Id">Click to reveal</code>
-                <button class="btn btn-sm btn-light mt-2" hidden data-clipboard-confirm>
-                   Copy to clipboard
-                </button>
+                <button class="btn btn-sm btn-outline-primary only-for-js" data-reveal-btn>Click to reveal</button>
+                <div hidden>
+                    <code data-api-key>@keyData.Id</code>
+                    <button class="btn btn-sm btn-outline-primary mt-2" data-clipboard-confirm>
+                        Copy to clipboard
+                    </button>
+                </div>
             </td>
             <td>
                 @{
@@ -55,7 +58,7 @@
             </td>
             <td class="text-right">
                 <a asp-action="RemoveAPIKey" asp-route-id="@keyData.Id" asp-controller="Manage">Remove</a>
-                <span class="only-for-js">-</span>
+                <span>-</span>
                 <button type="button" class="btn btn-link only-for-js" data-qr="@index">Show QR</button>
             </td>
         </tr>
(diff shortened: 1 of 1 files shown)


dennisreimann:
Maybe we can reuse the UI pattern from the wallet receive page here for better consistency. It'd look like this:

apikey


If you take a look at that page you can also simplify the code a bit. It should work out of the box without the custom JS code introduced here, because there's a [global event handler for `data-clipboard`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/main/site.js#L83) already.

Something like this:

```html
<div class="input-group" data-clipboard="API_KEY">
    <input type="text" class="form-control" style="cursor: copy" readonly="readonly" id="apikey" value="API_KEY">
            <button type="button" class="btn btn-outline-secondary" data-clipboard-confirm="" data-clipboard-initial-text="Copy API key">Copy API key</button>
    </div>
</div>
```

bolatovumar:
@dennisreimann I tried it but it ends up looking weird because of width constrain. You can see the entire key before copying it.

Screen Shot 2021-04-08 at 8 16 48 PM

dennisreimann:
I think we should combine the Label and Key columns, because that label space is too small anyways. Let's make the heading "Key", have the label as a kind of headline and put the input below it. I think this could work.

dstrukt:
I like the direction @dennisreimann would take. I have 1 other suggestion, and it might help consolidate some of the table UI a bit across the board. Will post a visual of both options EOD.

dstrukt:
1. In regards to the iteration with the simple "Copy" text, I think it could be hidden, and show on hover, then change to "Copied" on click.

2. Another thought that crossed my mind was .. how necessary / convenient is it to show all permissions in the table view, vs. say tucking them behind an aside / overlay / etc.. where you can simply "+ add another permission" when you tap a Read / Write button within the shown table ... see last example.

API_Keys_V1 1

Thoughts?

bolatovumar:
> 1. In regards to the iteration with the simple "Copy" text, I think it could be hidden, and show on hover, then change to "Copied" on click. Removing the "Click to reveal", unless this is absolutely necessary...

I think the "click to reveal" was added so someone looking at your screen wouldn't be able to photograph your API keys or something. @Kukks do you have the background on this piece?

> 2. Another thought that crossed my mind was .. how necessary / convenient is it to show all permissions in the table view, vs. say tucking them behind an aside / overlay / etc.. where you can simply "+ add another permission" when you tap a Read / Write button within the shown table ... see last example.

Probably not terribly important but it's nice to have them all listed if you want to create a few different keys with different permissions and want to be able to search with just "Ctrl + F".

> Not a fan of the input style copy .. it works for stuff outside of a table, but within a table it feels clunky and crowded.
> 
> 1. The notion of tucking "actions" behind a dropdown has also been something that might be worth exploring as tables scale beyond 2-3 actions. Helps reduce noise, and we could easily add a decent click area for both mobile or desktop. Regardless, thought I'd toss it out there for discussion.

I like the idea of tucking actions away. I thought of this as well because the "Actions" section starts looking weird when other sections start taking up more space.

dstrukt:
> I think the "click to reveal" was added so someone looking at your screen wouldn't be able to photograph your API keys or something. @Kukks do you have the background on this piece?

Ahhh I see, I kind of figured there was some back story like that, and I suppose we could still always just do the original "Click to reveal" in place of any of that text .. but wanted to confirm that extra click was necessary. 

> Probably not terribly important but it's nice to have them all listed if you want to create a few different keys with different permissions and want to be able to search with just "Ctrl + F".

I'll mock up a quick example to show was could be possible / see if it actually makes sense to do this.

bolatovumar:
@dstrukt how about this design for now (this is what Github uses for personal tokens). It's simple and the icon makes it pretty obvious what it does. Also we already have a similar icon bundled with BTCPay.

Screen Shot 2021-04-16 at 4 13 12 PM

dstrukt:
I tend to opt for the more minimal (string + "Copy"), but yeah, that'd definitely work.

You've been in the functionality, which is the most important part, and we can always adjust the subtly of the design in the future if needed, or when we do more work on our tables. 

I don't want to delay getting this useful functionality out there.

bolatovumar:
> I tend to opt for the more minimal (string + "Copy"), but yeah, that'd definitely work.
> 
> You've been in the functionality, which is the most important part, and we can always adjust the subtly of the design in the future if needed, or when we do more work on our tables.
> 
> I don't want to delay getting this useful functionality out there.

Ok, I won't be making any design updates now then. We can adjust the design later if needed.

bolatovumar:
@dstrukt another example from Stripe. It's quite nice:

Screen Shot 2021-04-18 at 8 32 39 PM

dstrukt:
Ahh, that is nice and elegant. 

Looks like that functionality is more common than I thought. TIL.

If you wanted to implement something like that with blur, and a simple text / small button, feel free, I'm sure it'll look great!

bolatovumar:
@NicolasDorier I made the design updates I was going to make, I think it's OK to merge now.

PR #2448, opened 2021-04-08
do not crash invoice if wellknown metadata keys used with different e…

Kukks:
…xpected types

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -135,7 +135,7 @@ namespace BTCPayServer.Services.Invoices
             JToken data;
             if (value is null)
             {
-                data = JValue.CreateNull();
+                AdditionalData?.Remove(propName);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


NicolasDorier:
I don't really like we try to be too smart here. Let me give a try.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -30,70 +30,136 @@ static InvoiceMetadata()
...
+            }
+            else
+            {
+                try
                 {
-                    try
+                    if (value is string s)
                     {
-                        PosRawData = JToken.Parse(value);
+                        data = JToken.Parse(s);
```
NicolasDorier, line 146: this assume s is representing json, this is not the case. A String of `\"lol\"` should not be interpreted as json `"blah" :"lol"`

NicolasDorier:
@Kukks can you also add some stuff I did on the other PR which fix crash if there is duplicate field in the JSON in the posdata_parser (just replace .Add by TryAdd)

BTCPayServer/Services/Invoices/InvoiceEntity.cs
Kukks: No, it actually works this way.

for example, creating an invoice with metadata:
```
    "metadata": {
            "taxIncluded" : 1,
            "buyerName" : "\"hello\"",
        "posData" : {
            "test": [ "hello", "bye"]
        }
    }
```
results in following: 
image

Kukks:
> @Kukks can you also add some stuff I did on the other PR which fix crash if there is duplicate field in the JSON in the posdata_parser (just replace .Add by TryAdd)

Done

PR #2449, opened 2021-04-08
Self-host PoS sample images

dennisreimann:
As brought up in #2447 and previously been worked on in #1039.

Fixes a currently broken image reference for the black tea image, but the main thing here is that we should self-host these images anyway. As ketominer originally mentioned, we are otheriwse exposing the BTCPay host by deep linking to those images.


Kukks (commented):
One concern is if the image link will work when there is a tooth path defined (eg: domain.com/btcpay)

Eskyee:
I agree with @dennisreimann 
exposing the BTCPay host by deep linking to those images. the least footprints BTCPay exposing is better, 
but like @Kukks has explained, broken links may happen because of (eg: domain.com/btcpay)
lots to think about here  ??

Eskyee:
trying to understand IPFS some more @Kukks @dennisreimann and maybe we could add to a fragment,  in future 
because storing files, etc etc sound very interesting here, for the future of decentralised storage. 
Screenshot 2021-04-08 at 12 31 28
Screenshot 2021-04-08 at 12 32 27

Eskyee:
image

dennisreimann:
> One concern is if the image link will work when there is a tooth path defined (eg: domain.com/btcpay)

I also thought about this and tried it using the `~` prefix. This works in the customer facing view, but not in the template preview. The latter can be fixed  with a bit of JS in the [`getImage` function](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Apps/TemplateEditor.cshtml#L163). In that function I saw that the placeholder image also wouldn't work in that case and I thought it might be ok to also go with `/img/...`, because having the `~/img/` prefix might confuse users … but yeah, then let's fix this :)

Eskyee:
tested  your branch local dennisreimann:pos-sample-images
and got this.. https://cdn.pixabay.com/photo/2016/11/29/13/04/beverage-1869716__480.jpg

error still. @dennisreimann

dennisreimann:
> error still.

Did you create a new PoS app? It won't work for existing ones, because the image URLs are set on creation …

Eskyee:
Ok ✅ yes works via img ok..
Can confirm

Eskyee:
image

Kukks (approved):
I think the small hack in the editor is fine IMO. 
Code ack

NicolasDorier:
I prefer we do not self host, also I don't really like the hack with `~`.
What about hosting those images on github instead?

Those images are not so important, this is just an example, it should not be part of the repo.

dennisreimann:
The hack only relates to the sample images and to the editor. Also I think it's worth it for not making those requests to a third-party. I think the trade-off of hosting those six images ourselves is owrth not exposing the BTCPay host to a third-party.

Kukks:
How about this: we use a default placeholder image 

> The hack only relates to the sample images and to the editor. Also I think it's worth it for not making those requests to a third-party. I think the trade-off of hosting those six images ourselves is worth not exposing the BTCPay host to a third-party.

I agree with @dennisreimann. Maybe we can optimize by hosting one image instead, or using a placeholder svg?

dennisreimann:
There is a [placeholder svg](https://github.com/btcpayserver/btcpayserver/pull/2449/files#diff-1a9890a8b7102fa91230d0d78f24a0b189a466000d8c69a7595c868a079e7ad1R164) already. 

Those images make for a better example imho, but we can remove them from the sample pos if that's wanted.

BTCPayServer/Views/Apps/TemplateEditor.cshtml
```diff
@@ -161,8 +161,10 @@ $(function() {
                 this.getInputElement().parent().toggle();
             },
             getImage: function(item){
+                var image = this.unEscapeKey(item.image) || "~/img/img-placeholder.svg";
+                var url = image.startsWith("~") ? image.replace('~', window.location.pathname.substring(0, image.indexOf('/apps'))) : image;
```
NicolasDorier, line 165: You do not even need this hack, url support relative url already.

Which is even better to use as it obviously support RootPath

https://stackoverflow.com/questions/940451/using-relative-url-in-css-file-what-location-is-it-relative-to

bolatovumar (approved):

dennisreimann:
How do we want to proceed here? Remove the sample images and the hack?

NicolasDorier:
@dennisreimann if you remove the hack with the ~, does it work?

dennisreimann:
@NicolasDorier Not for instances that aren't deployed at the root path, e.g. under `/btcpay/`. The editor and public view have a different path structure, so that we cannot use a relative image URL. Using the `~` makes it work in the public view via the general ASP.NET mechanism – the hack is for the editor where we patch it for the JS/non-ASP.NET view.

dennisreimann:
Rebased and pushed the alternative version without sample images and removed path hack.

NicolasDorier:
I don't really like having relative path in image.
The reason for the default template is that it teach user how we expect them to write the template. And we expect them to put full absolute link in the image field.

What about just hosting those images on github and referencing it instead?

dennisreimann:
I don't like reaching out to a third-party host for those images, because it exposes the BTCPay instance. This was the reason for the PR in the first place. 

My take: the version with the self-hosted images and tiny hack for the editor is best, because it'd also show the user how it could be done with images uploaded to the instance itself. 

imho it also doesn't hurt to ship those six sample images as they give a better impression of how it is supposed to work.

NicolasDorier:
ok I give up, put back your hack

dennisreimann:
How can we proceed here? I think everything has been said and the different variants have been implemented -- let's decide how we want to fix it and then we can include it in the next release.

britttttk:
>The reason for the default template is that it teach user how we expect them to write the template. And we expect them to put full absolute link in the image field.

I agree this is helpful to users.

Kukks:
Seems like nicolas conceded to the hack and I am in favor of it. We can iterate on this later on but this base direction of this PR is good (self hosted images, reduce leakage)

PR #2450, opened 2021-04-08
Add payjoin option to hot wallet setup

dennisreimann:
Follow-up to the discussion in #2406.

Enables payjoin by default when creating a hot wallet and offers the user an opt-out.

If we enable it by default I'd recommend to show it immediately and not hide it in the advanced settings, though it feels like it would naturaly belong there.

Wording for the accompanying text tbd, the help icon links to the [PayJoin docs](https://docs.btcpayserver.org/Payjoin/).

This is currently only available when generating a new hot wallet. It could potentially also be offered on the seed import page, but there we have a toggle for first choosing it to be a hot wallet. I think it might be confusing to have the separate PayJoin toggle there as well, because having the hot wallet enabled is a prerequisite for offering the PayJoin toggle. 

hotwallet-payjoin

Let's continue the discussion around that here.

### DIFF
--- a/BTCPayServer/Controllers/StoresController.cs
+++ b/BTCPayServer/Controllers/StoresController.cs
@@ -466,7 +466,6 @@ namespace BTCPayServer.Controllers
                 }
             }
             
-
             if (!ModelState.IsValid)
             {
                 return View(model);
@@ -558,12 +557,9 @@ namespace BTCPayServer.Controllers
                 }
             }
         }
-
-
-
-        [HttpGet]
-        [Route("{storeId}")]
-        public IActionResult UpdateStore()
+        
+        [HttpGet("{storeId}")]
+        public async Task<IActionResult> UpdateStore()
         {
             var store = HttpContext.GetStoreData();
             if (store == null)
@@ -586,12 +582,19 @@ namespace BTCPayServer.Controllers
             vm.PayJoinEnabled = storeBlob.PayJoinEnabled;
             vm.HintWallet = storeBlob.Hints.Wallet;
             vm.HintLightning = storeBlob.Hints.Lightning;
+            
+            (bool canUseHotWallet, _) = await CanUseHotWallet();
+            vm.CanUsePayJoin = canUseHotWallet && store
+                .GetSupportedPaymentMethods(_NetworkProvider)
+                .OfType<DerivationSchemeSettings>()
+                .Any(settings => settings.Network.SupportPayJoin &&
+                                 !string.IsNullOrEmpty(_ExplorerProvider.GetExplorerClient(settings.Network)
+                                     .GetMetadata<string>(settings.AccountDerivation, WellknownMetadataKeys.Mnemonic)));
+            
             return View(vm);
         }
-
-
-        [HttpPost]
-        [Route("{storeId}")]
+        
+        [HttpPost("{storeId}")]
         public async Task<IActionResult> UpdateStore(StoreViewModel model, string command = null)
         {
             bool needUpdate = false;
--- a/BTCPayServer/Controllers/StoresController.cs
+++ b/BTCPayServer/Controllers/StoresController.cs
@@ -587,9 +587,7 @@ namespace BTCPayServer.Controllers
             vm.CanUsePayJoin = canUseHotWallet && store
                 .GetSupportedPaymentMethods(_NetworkProvider)
                 .OfType<DerivationSchemeSettings>()
-                .Any(settings => settings.Network.SupportPayJoin &&
-                                 !string.IsNullOrEmpty(_ExplorerProvider.GetExplorerClient(settings.Network)
-                                     .GetMetadata<string>(settings.AccountDerivation, WellknownMetadataKeys.Mnemonic)));
+                .Any(settings => settings.Network.SupportPayJoin && settings.IsHotWallet);
             
             return View(vm);
         }
@@ -638,11 +636,7 @@ namespace BTCPayServer.Controllers
                 {
                     var problematicPayjoinEnabledMethods = CurrentStore.GetSupportedPaymentMethods(_NetworkProvider)
                         .OfType<DerivationSchemeSettings>()
-                        .Where(settings =>
-                            settings.Network.SupportPayJoin &&
-                            string.IsNullOrEmpty(_ExplorerProvider.GetExplorerClient(settings.Network)
-                                .GetMetadata<string>(settings.AccountDerivation,
-                                    WellknownMetadataKeys.Mnemonic)))
+                        .Where(settings => settings.Network.SupportPayJoin && !settings.IsHotWallet)
                         .Select(settings => settings.PaymentId.CryptoCode)
                         .ToArray();
 
--- a/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
+++ b/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
@@ -77,7 +77,7 @@ namespace BTCPayServer.Tests
 
                 // Get enabled state from overview action
                 StoreViewModel storeModel;
-                response = controller.UpdateStore();
+                response = await controller.UpdateStore();
                 storeModel = (StoreViewModel)Assert.IsType<ViewResult>(response).Model;
                 var lnNode = storeModel.LightningNodes.Find(node => node.CryptoCode == cryptoCode);
                 Assert.NotNull(lnNode);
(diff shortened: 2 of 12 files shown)


dstrukt:
1. I'm also a bit conflicted on, should we tuck this behind "advanced settings" or not, as it does seem like that type of option ... but considering it's on by default, it seems like something that should be visible.

2. Not entirely sure how we should proceed on the seed import issue raised, we could hide/show depending on if the Hot Wallet option is enabled/disabled, and then have it be toggled on by default .. which leads me to .. 

after playing around with it, I have a functional request around the enable/disable switch + save button, that we should address when tackling the "Settings" page integration, but noting for the design meeting, as it's out of scope here (I think).

..and alternatively, I think breaking the recovery seed flow into multiple steps seems a bit overkill...

Will keep thinking on this one.

Zaxounette:
As stated in the previous [discussion](https://github.com/btcpayserver/btcpayserver/discussions/2406#discussioncomment-585319):
1°) I'd like a "`recommended`" added on the right of the "enabled/disabled" button for two reasons:
* Because we're actually recommending it by enabling it by default, so let's make it clear visually.
* Because some user types (me included) disable options that aren't recommended if they don't know what they are. 

2°) The wording you put is good, it explains briefly what the gain is of using that feature, and links to the docs for more informations. There's really nothing more to add there IMHO.

3°) Regarding the seed import and this default behavior:
Couldn't a trick be used to make the payjoin enabled option appear only if the hot wallet option is ticked, and disappear if untocked ?

Zaxounette:
@dstrukt 

> I have a functional request around the enable/disable switch + save button, that we should address when tackling the "Settings" page integration, but noting for the design meeting, as it's out of scope here (I think).

If it's even remotely related to design, I feel like it is in scope of the meetings. IMO you can bring it up next time.

NicolasDorier:
I think from UX point of view this is confusing. Because now we have `Enable payjoin` at store level and at wallet level.

This checkbox would be better placed in the Receive page of the wallet and/or in the wallet settings rather than in the setup wizard.

Kixunil:
@NicolasDorier would it make sense to make it per-wallet only? Per-store seems to be bloated.

Kukks:
> @NicolasDorier would it make sense to make it per-wallet only? Per-store seems to be bloated.

Maybe, it is an onchain-wallet specific setting after all.

Maybe we should move it from the Store Blob to the payment method blob

NicolasDorier:
@Kukks @Kixunil I don't think that make sense.
Most of the properties at the store level are about setting the default settings when creating a new invoices. Payjoin is not different.

MaiHistoryCompanyLimited:
{#2450}

Kixunil:
In that case maybe adding the checkbox into the store wallet setup flow could work?

Kukks:
> @Kukks @Kixunil I don't think that make sense.
> Most of the properties at the store level are about setting the default settings when creating a new invoices. Payjoin is not different.

Payjoin is now usable without an invoice (using wallet receive). Does that change your opinion on this @NicolasDorier?

NicolasDorier:
@Kukks I think the following:

* The enable payjoin checkbox should be `enable payjoin on every invoices by default`
* Activating payjoin for the wallet, should not be done as a checkbox like this, but as a checkbox in the receive wallet page. (eventually preference saving what user check via cookie)

dennisreimann:
We discussed this in [yesterdays design meeting](https://github.com/btcpayserver/organization/issues/33#issuecomment-829586745), here are the main points that were brought up:

- Consens is we want to make Payjoin the default option for hot wallets
- Thus we need to make the user aware and also offer a good way to opt-out
  - The user needs to understand what using Payjoin means in terms of …
  - Transaction handling: It might be confusing to see the usage of UTXOs in a way you would not expect
  - Regulatory/legal compliance (might be hypothetical, but I think we cannot simply write it off)  
- Mentioning this feature in a prominent and clear step of the wallet creation thus seems intuitive  
- Moving the Payjoin toggle to the Wallet Receive tab seems like we would bury the option
  - Users might not even know of the option, less think to disable it if they wish if it's placed in that tab.
- The Payjoin option is only useful for hot wallets, so it makes sense to have it in the hot wallet setup

Other than that two more points came up:

- Right now the Payjoin toggle is also offered for non-hot wallets, which does not make sense.
- For future reference the clarification of whether Payjoin is a store or wallet concept would make sense. As wallets are tight to a store right now this blends and we aren't distinguishing there, but it would help to better communicate this feature.

dennisreimann:
@Kukks @NicolasDorier What are your thoughts regarding the comments that came up in the design meeting?

Kukks:
> We discussed this in [yesterdays design meeting](https://github.com/btcpayserver/organization/issues/33#issuecomment-829586745), here are the main points that were brought up:
> 
> * Consens is we want to make Payjoin the default option for hot wallets

ACK

> * Thus we need to make the user aware and also offer a good way to opt-out
>   
>   * The user needs to understand what using Payjoin means in terms of …
>   * Transaction handling: It might be confusing to see the usage of UTXOs in a way you would not expect
>   * Regulatory/legal compliance (might be hypothetical, but I think we cannot simply write it off)

We should look to existing external docs instead

> * Mentioning this feature in a prominent and clear step of the wallet creation thus seems intuitive

ACK BUT the payjoin option is currently on a store wide level. We need to discuss to perhaps move the payjoin option to a wallet level then (eg. if payjoin is on right now  it is on for both BTC and LTC if it is configured). 

> * Moving the Payjoin toggle to the Wallet Receive tab seems like we would bury the option
>   

I disagree with putting the payjoin toggle there. 

>   * Users might not even know of the option, less think to disable it if they wish if it's placed in that tab.
> * The Payjoin option is only useful for hot wallets, so it makes sense to have it in the hot wallet setup
> 
> Other than that two more points came up:
> 
> * Right now the Payjoin toggle is also offered for non-hot wallets, which does not make sense.

Correct, if we move the payjoin option to a wallet level, we can handle this properly.

> * For future reference the clarification of whether Payjoin is a store or wallet concept would make sense. As wallets are tight to a store right now this blends and we aren't distinguishing there, but it would help to better communicate this feature.

IMO Payjoin = wallet feature.

dstrukt:
🎉

I'm excited to see we're all on the same page now, and this is getting pushed forward after so much discussion, etc..

dennisreimann:
Integrated @Kukks' feedback and made it so that the PayJoin enabling option is only show if one of the networks supports PayJoin and has a hot wallet associated.

@NicolasDorier Can you take a look again, let me know what you think and how to proceed here? Thanks!

BTCPayServer/Controllers/StoresController.cs
```diff
@@ -583,12 +579,19 @@ public IActionResult UpdateStore()
             vm.PayJoinEnabled = storeBlob.PayJoinEnabled;
             vm.HintWallet = storeBlob.Hints.Wallet;
             vm.HintLightning = storeBlob.Hints.Lightning;
+            
+            (bool canUseHotWallet, _) = await CanUseHotWallet();
+            vm.CanUsePayJoin = canUseHotWallet && store
+                .GetSupportedPaymentMethods(_NetworkProvider)
+                .OfType<DerivationSchemeSettings>()
+                .Any(settings => settings.Network.SupportPayJoin &&
+                                 !string.IsNullOrEmpty(_ExplorerProvider.GetExplorerClient(settings.Network)
```
NicolasDorier, line 588: Please, instead of fetching the mnemonic like this, use `settings.IsHotWallet`.

dennisreimann: Done in 9fe24aad2d12b714bd955d43d5e2a0c0bd70ed32, thanks for the pointer!

NicolasDorier:
@dennisreimann @dstrukt rather than removing the option in update store if the wallet isn't hot. What about make the box disabled with a tooltip explaining it is only for hotwallet?

dstrukt:
Hmm... is the intention simply to communicate that this functionality *is* possible with a hot wallet?

If it's really important to communicate, that's fine with me, but otherwise, I'd opt to remove it to reduce noise in the UI, as it will be enabled and visible by default if the wallet is hot.

..but perhaps I'm missing some additional context?

dennisreimann:
I'd second what @dstrukt just said. I think there'd be better places to educate the user about that functionality and presenting it as proposed isn't really actionable. The user would have to replace the existing wallet and I think that's rather unlikely. 

Maybe we should point out earlier in the setup flow that this feature exists and which route the user should take to enable it. Featuring it in the first (import vs. create) and second (how wallet vs. watch-only) branching pages would be a better option imho.

jmayes41 (commented):
new-release

PR #2453, opened 2021-04-11
Unify POS app card markup

bolatovumar:
See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/2431, particularly this comment: https://github.com/btcpayserver/btcpayserver/discussions/2431#discussioncomment-587869

**Before (notice the button looks different):**

Screen Shot 2021-04-10 at 6 33 42 PM

**After:**

Screen Shot 2021-04-10 at 6 33 22 PM

### DIFF
--- a/BTCPayServer/Views/AppsPublic/PointOfSale/Static.cshtml
+++ b/BTCPayServer/Views/AppsPublic/PointOfSale/Static.cshtml
@@ -19,7 +19,7 @@
             @for (int x = 0; x < Model.Items.Length; x++)
             {
                 var item = Model.Items[x];
-                var buttonText = (string.IsNullOrEmpty(item.BuyButtonText) ? 
+                var buttonText = (string.IsNullOrEmpty(item.BuyButtonText) ?
                     item.Custom?
                     Model.CustomButtonText :
                     Model.ButtonText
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/AppsPublic/PointOfSale/Static.cshtml
```diff
@@ -120,3 +90,31 @@
         </div>
     </div>
 </div>
+
+@functions {
+    private void PayFormInputContent(string buttonText, decimal? minPriceValue = null, decimal? priceValue = null)
```
bolatovumar, line 95: Split shared markup which was almost the same but slightly different previously into these two functions to ensure the mark-up is the same.

MaiHistoryCompanyLimited:
{#2453}

dennisreimann (approved):

BTCPayServer/Views/AppsPublic/PointOfSale/Static.cshtml
Kukks: Interesting take, you've recreated React in .net razor

dennisreimann: Is there a way to define inline view components? That might be a good alternative here …

bolatovumar: I couldn't find a way to do it inline. There are a few ways to do it not in line while still componentizing the mark-up and this one seemed to work best.

NicolasDorier: cool I did not know we could do that in razor.

NicolasDorier (approved):

PR #2456, opened 2021-04-12
Make the Metadata JToken, do not crash if posData has duplicate fields

NicolasDorier:
Alternative to #2448

I don't really like the fact that we are hacking around a class that should be relatively simple.
Instead here, I just use JToken everywhere, and use `AsString()`. Additionally, I improved the UI posData parser to accept a posData of JObject type.


Kukks (approved):
lgtm

Kukks (changes requested):
There a bunch of tests failing so the changes are not as graceful as #2448

NicolasDorier:
yeah, tried to fix it but this goes into a rabbit hole.

PR #11, opened 2017-11-06
Allowing user to invalidate paid invoice

lepipele:

PR #2457, opened 2021-04-12
Make CSSThemeManager really only focus on theme

Kukks:
The settings repo now handles caching itself. Since it is the only IO to settings records, it should work well without issue.

~~This also reduces the scope of the css theme manager to only deal with handling theme data. Realistically, we can also remove the css theme manager but let's leave that for another PR.~~

CSSThemeManager removed!


dennisreimann (approved):
Good change, the name change alone makes it worth it :)

Building on this I'd like to do another change we've decided on in the [Design Call #11](https://github.com/btcpayserver/organization/issues/40#issuecomment-863577103): If the custom theme isn't set deliberately (and the instance is using the default light or dark theme) we should move the theme choice to the user side and display a toggle like we do in the docs, the website, the blog etc. 

I think we should aslo get rid of the `BootstrapCssUri` and `CreativeStartCssUri`, because they are part of the legacy themeing approach.

Kukks:
> I think we should aslo get rid of the `BootstrapCssUri` and `CreativeStartCssUri`, because they are part of the legacy themeing approach.


Separate PR for later

NicolasDorier:
Please no abstract `ISettingsRepository` with extensions methods. Just a SettingsRepository with normal methods.

BTCPayServer/Views/Shared/_LayoutPos.cshtml
```diff
@@ -20,8 +22,9 @@
 
     <link rel="manifest" href="~/manifest.json">
 
-    <link href="@this.Context.Request.GetRelativePathOrAbsolute(themeManager.BootstrapUri)" rel="stylesheet" asp-append-version="true" />
-    <link href="@this.Context.Request.GetRelativePathOrAbsolute(themeManager.ThemeUri)" rel="stylesheet" asp-append-version="true" />
```
NicolasDorier, line 24: you removed `asp-append-version="true"` ? I think it is useful for some caching issues. @dennisreimann

NicolasDorier:
Concept ACK, just no abstract repo please :D

BTCPayServer/Views/Shared/_LayoutPos.cshtml
dennisreimann: Yes, please keep them. And while we are at it, let's add them to the references in `ViewPullPayment.cshtml `, `ViewPaymentRequest.cshtml `, `ViewCrowdfund.cshtml`, `ShowLightningNodeInfo.cshtml` as well.

Kukks: fixed

Kukks:
> Concept ACK, just no abstract repo please :D
> Code seems good as well.

abstract repo was there before and is used for plugins

PR #2462, opened 2021-04-14
Payouts: Detect External OnChain Payouts 

### LINKED ISSUE #1779: Payouts - Track offline transfer

Kukks:
closes #1779

### DIFF
--- a/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
+++ b/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using BTCPayServer;
+using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.Events;
--- a/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
+++ b/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
@@ -9,8 +10,13 @@ using BTCPayServer.Data;
 using BTCPayServer.Events;
 using BTCPayServer.HostedServices;
 using BTCPayServer.Logging;
+using BTCPayServer.Models;
+using BTCPayServer.Models.WalletViewModels;
 using BTCPayServer.Payments;
 using BTCPayServer.Services;
+using BTCPayServer.Services.Notifications;
+using BTCPayServer.Services.Notifications.Blobs;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
@@ -29,22 +35,32 @@ public class BitcoinLikePayoutHandler : IPayoutHandler
     private readonly BTCPayNetworkJsonSerializerSettings _jsonSerializerSettings;
     private readonly ApplicationDbContextFactory _dbContextFactory;
     private readonly EventAggregator _eventAggregator;
+    private readonly NotificationSender _notificationSender;
 
     public BitcoinLikePayoutHandler(BTCPayNetworkProvider btcPayNetworkProvider,
         ExplorerClientProvider explorerClientProvider, BTCPayNetworkJsonSerializerSettings jsonSerializerSettings,
-        ApplicationDbContextFactory dbContextFactory, EventAggregator eventAggregator)
+        ApplicationDbContextFactory dbContextFactory, EventAggregator eventAggregator, NotificationSender notificationSender)
     {
         _btcPayNetworkProvider = btcPayNetworkProvider;
         _explorerClientProvider = explorerClientProvider;
         _jsonSerializerSettings = jsonSerializerSettings;
         _dbContextFactory = dbContextFactory;
         _eventAggregator = eventAggregator;
+        _notificationSender = notificationSender;
     }
 
     public bool CanHandle(PaymentMethodId paymentMethod)
     {
         return paymentMethod.PaymentType == BitcoinPaymentType.Instance &&
                _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethod.CryptoCode)?.ReadonlyWallet is false;
+    }
+
+    public async Task TrackClaim(PaymentMethodId paymentMethodId, IClaimDestination claimDestination)
+    {
+        var network = _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethodId.CryptoCode);
+        var explorerClient = _explorerClientProvider.GetExplorerClient(network);
+        if (claimDestination is IBitcoinLikeClaimDestination bitcoinLikeClaimDestination)
+            await explorerClient.TrackAsync(TrackedSource.Create(bitcoinLikeClaimDestination.Address));
     }
 
     public Task<IClaimDestination> ParseClaimDestination(PaymentMethodId paymentMethodId, string destination)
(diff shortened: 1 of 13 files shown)


pavlenex:
@Kukks There are some conflicts here we should resolve in order to procceed.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -119,6 +136,78 @@ public Task<decimal> GetMinimumPayoutAmount(PaymentMethodId paymentMethodId, ICl
         return Task.FromResult(0m);
     }
 
+
+    public Dictionary<PayoutState, List<(string Action, string Text)>> GetPayoutSpecificActions()
```
NicolasDorier, line 140: just hard code it in the cshtml, no need to make it more dynamic that it need to be.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -260,6 +349,15 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
                             newTransaction.NewTransactionEvent.TransactionData.TransactionHash,	
                             UpdateTransactionLabel.PayoutTemplate(payout.Id,payout.PullPaymentDataId, walletId.ToString())));	
                     }
+                    else
```
NicolasDorier, line 352: The `Accounted` property is super misleading.
It should be renamed to `IsExtermalPayout`.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -47,6 +56,14 @@ public bool CanHandle(PaymentMethodId paymentMethod)
                _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethod.CryptoCode)?.ReadonlyWallet is false;
     }
 
+    public async Task TrackClaim(PaymentMethodId paymentMethodId, IClaimDestination claimDestination)
+    {
+        var network = _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(paymentMethodId.CryptoCode);
+        var explorerClient = _explorerClientProvider.GetExplorerClient(network);
+        if (claimDestination is IBitcoinLikeClaimDestination bitcoinLikeClaimDestination)
+            await explorerClient.TrackAsync(TrackedSource.Create(bitcoinLikeClaimDestination.Address));
```
NicolasDorier, line 64: So now, even for non external payout, we will track the individual address. We should not do this, since then we will receive two notifications from NBXplorer, and the payout handler will not know if it need to use the "NewTransactionAddress" or "NewTransactionDerivationScheme" event.

Kukks: > So now, even for non external payout, we will track the individual address. We should not do this, since then we will receive two notifications from NBXplorer, and the payout handler will not know if it need to use the "NewTransactionAddress" or "NewTransactionDerivationScheme" event.

This is handled in the Listener

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -241,30 +319,43 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
+                var derivationSchemeSettings = payout.PullPaymentData.StoreData
+                    .GetDerivationSchemeSettings(_btcPayNetworkProvider, newTransaction.CryptoCode).AccountDerivation;
+
+                var storeWalletMatched = (await _explorerClientProvider.GetExplorerClient(newTransaction.CryptoCode)
+                    .GetTransactionAsync(derivationSchemeSettings,
+                        newTransaction.NewTransactionEvent.TransactionData.TransactionHash));
+                var isInternal = storeWalletMatched is null? false:  !newTransaction.NewTransactionEvent.Outputs.All(output => storeWalletMatched.Outputs.Any(
+                    matchedOutput => matchedOutput.Index == output.Index && matchedOutput.Value == output.Value &&
+                                     matchedOutput.ScriptPubKey == output.ScriptPubKey) ) && !newTransaction.NewTransactionEvent.Outputs.All(output => storeWalletMatched.Inputs.Any(
+                    matchedInput => matchedInput.Index == output.Index && matchedInput.Value == output.Value &&
```
NicolasDorier, line 332: I don't understand this condition, the `matchedOutput` isn't enough?

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -241,30 +319,43 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
+                if (!proof.Candidates.Add(txId)) continue;
+                if (proof.Accounted is true)
+                {
+                    payout.State = PayoutState.InProgress;
+                    var walletId = new WalletId(payout.PullPaymentData.StoreId, newTransaction.CryptoCode);
+                    _eventAggregator.Publish(new UpdateTransactionLabel(walletId,	
+                        newTransaction.NewTransactionEvent.TransactionData.TransactionHash,	
+                        UpdateTransactionLabel.PayoutTemplate(payout.Id,payout.PullPaymentDataId, walletId.ToString())));	
+                }
+                else
```
NicolasDorier, line 348: the fact that `proof.Accounted` is false, does not mean it is external. It may have been like that already in the previous proof. I think you should check `isInternal` rather than using `else`.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: will attempt to simplify after all payouts  related PRs are merged

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: fixed this

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: Because I wanted to make sure that there was no edge case around the store wallet being the destination or involved in some other way. Maybe I am being too paranoid?

NicolasDorier: What is the problem if that's the case? It should be considered paid.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -196,36 +285,25 @@ private async Task UpdatePayoutsInProgress()
...
-    private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent newTransaction)
+    private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent newTransaction,
+        AddressTrackedSource addressTrackedSource)
     {
         try
         {
             var network = _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(newTransaction.CryptoCode);
-            Dictionary<string, decimal> destinations;
-            if (newTransaction.NewTransactionEvent.TrackedSource is AddressTrackedSource addressTrackedSource)
+            Dictionary<string, decimal> destinations = new Dictionary<string, decimal>
```
NicolasDorier, line 294: I don't think we need dictionary anymore.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -241,30 +319,44 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
+
+                var derivationSchemeSettings = payout.PullPaymentData.StoreData
+                    .GetDerivationSchemeSettings(_btcPayNetworkProvider, newTransaction.CryptoCode).AccountDerivation;
+
+                var storeWalletMatched = (await _explorerClientProvider.GetExplorerClient(newTransaction.CryptoCode)
+                    .GetTransactionAsync(derivationSchemeSettings,
+                        newTransaction.NewTransactionEvent.TransactionData.TransactionHash));
+                //if the wallet related to the store related to the payout does not have the tx: it is external
+                //if the wallet has the tx but outputs that matched weren't the payout's output: it is external 
+                var isInternal = storeWalletMatched is { } && destination.Value == storeWalletMatched.Outputs
```
NicolasDorier, line 331: you only have to test if `storeWalletMatched` is not null. You can remove the other condition.
The other condition will always be true because we check it above

`destination.Value != BTCPayServer.Extensions.RoundUp(payoutBlob.CryptoAmount.Value, network.Divisibility)`

PR #2463, opened 2021-04-14
UI fixes

dennisreimann:
Two commits with simple UI fixes:

- The bootstrap.css contained a formatting error, which looks like it got introduced when resolving a merge conflict. I must have overlooked this when recently updating the bootstrap version.
- Two typos in views, which also look like something I introduced with the recent monster view update.

### DIFF
--- a/BTCPayServer/Views/Home/Home.cshtml
+++ b/BTCPayServer/Views/Home/Home.cshtml
@@ -8,7 +8,7 @@
         <div class="header-content-inner text-white">
             <h1>Welcome to BTCPay Server</h1>
             <hr />
-            <p class="font-weight-semibold">BTCPay Server is a free and open source server for merchants wanting to accept Bitcoin for their business.</p>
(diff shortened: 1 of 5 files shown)


BTCPayServer/Views/Home/Home.cshtml
```diff
@@ -8,7 +8,7 @@
         <div class="header-content-inner text-white">
             <h1>Welcome to BTCPay Server</h1>
             <hr />
-            <p class="font-weight-semibold">BTCPay Server is a free and open source server for merchants wanting to accept Bitcoin for their business.</p>
+            <p class="fw-semibold">BTCPay Server is a free and open source server for merchants wanting to accept Bitcoin for their business.</p>
```
bolatovumar, line 11: Unfortunately swapping out the class name here will break the appearance as there is a more specific CSS rule which will apply in this case so the text won't appear semibold.

Screen Shot 2021-04-14 at 8 19 14 PM

dennisreimann: You are right, thanks for the pointer! Fixed :)

PR #2469, opened 2021-04-16
Move payment method enabling toggle to store overview

dennisreimann:
As wished by @NicolasDorier this brings the enable/disable functionality to the store overview page.

@Zaxounette also proposed a confirmation view for disabling, indicating that having the node offline for an extensive time could result in the loss of funds. (not added yet, will do that next)


BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -128,6 +129,38 @@ public async Task<IActionResult> AddLightningNode(string storeId, LightningNodeV
...
+                TempData[WellKnownTempData.ErrorMessage] = $"Only the server admin can {(enabled ? "enable" : "disable")} the lightning node";
+            }
+            else
+            {
+                var paymentMethodId = new PaymentMethodId(network.CryptoCode, PaymentTypes.LightningLike);
+                var storeBlob = store.GetStoreBlob();
+                storeBlob.SetExcluded(paymentMethodId, !enabled);
+                store.SetStoreBlob(storeBlob);
+                await _Repo.UpdateStore(store);
+                TempData[WellKnownTempData.SuccessMessage] = $"{network.CryptoCode} Lightning payments are {(enabled ? "enabled" : "disabled")}.";
```
Zaxounette, line 158:
```suggestion
                TempData[WellKnownTempData.SuccessMessage] = $"{network.CryptoCode} Lightning payments are now {(enabled ? "enabled" : "disabled")} for this store.";
```

BTCPayServer/Controllers/StoresController.Onchain.cs
```diff
@@ -506,6 +492,40 @@ public async Task<IActionResult> DeleteWallet(string storeId, string cryptoCode)
...
+
+            var paymentMethodId = new PaymentMethodId(network.CryptoCode, PaymentTypes.BTCLike);
+            var storeBlob = store.GetStoreBlob();
+            storeBlob.SetExcluded(paymentMethodId, !enabled);
+            store.SetStoreBlob(storeBlob);
+            await _Repo.UpdateStore(store);
+            _EventAggregator.Publish(new WalletChangedEvent {WalletId = new WalletId(storeId, cryptoCode)});
+
+            TempData[WellKnownTempData.SuccessMessage] =
+                $"{network.CryptoCode} on-chain payments are {(enabled ? "enabled" : "disabled")}.";
```
Zaxounette, line 524:
```suggestion
                $"{network.CryptoCode} on-chain payments are now {(enabled ? "enabled" : "disabled")} for this store.";
```

Zaxounette (changes requested):
Wording nitpicks.

Sentence structure + adding `for this store` since these toggles are per-store based.
Avoids possible user confusion in thinking that it disables for any and every store on the instance/account.

NicolasDorier:
Cool, should we be worried about the user forgetting to click save? @dstrukt 
I guess not, since we already require it in the current way of doing.

dennisreimann:
The toggle is actually a button that submits the go to value.

BTCPayServer/Controllers/StoresController.LightningLike.cs
dennisreimann: Thanks, updated! 👍

dennisreimann:
Good from my side, open to confirm wording suggestions though. @Zaxounette @pavlenex 

Once this is merged I'll redo the LN setup page as proposed by @dstrukt.

Zaxounette (changes requested):
AAnother wording pass.

* First, an old string that can be changed since the file is being worked on. Change `succeeded` to `successful`. Make a bit more sense in the context.
* Two instances of changing `modified` to `updated`. It's semantics, but relates to what wording can be found in the majority of other softwares. Also sounds better :p

NicolasDorier:
I think this popup

image


Should not appear. Because this is unrelated to disabling payment methods. Even if you disable lightning, it does not mean your node stop monitoring, thus user would be confused.

Zaxounette:
> I think this popup should not appear. Because this is unrelated to disabling payment methods. Even if you disable lightning, it does not mean your node stop monitoring, thus user would be confused.

I guess you're right. That one is on me.

dennisreimann:
Yeah, again also a misunderstanding on my part – as this is on the store level it ain't necessary. Thanks for the clarification, I'll remove the extra commit with the confirmation.

dstrukt:
As I said on MM, this looks great, and fits the specifications exactly to what we discussed!

Will help clean up a some of UI on the other screens, which is great.

BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -128,6 +129,38 @@ public async Task<IActionResult> AddLightningNode(string storeId, LightningNodeV
...
+
+            var network = cryptoCode == null ? null : _ExplorerProvider.GetNetwork(cryptoCode);
+            if (network == null)
+                return NotFound();
+
+            var lightning = GetExistingLightningSupportedPaymentMethod(cryptoCode, store);
+            if (lightning == null)
+                return NotFound();
+
+            if (!User.IsInRole(Roles.ServerAdmin))
```
NicolasDorier, line 147: what? no, only the owner of the store can.

NicolasDorier: You can remove this whole block, the controller level filter check that the user is indeed admin.

dennisreimann: Done in 962c80f.

PR #2477, opened 2021-04-19
Lightning setup redesign

dennisreimann:
Based on #2469, this introduces the Lightning setup redisgn by @dstrukt.

##  Internal Node

This one looks and feels empty without the extra text assuring the user that there's nothing to configure. Open to suggestions on how to handle this. 

light-internal

dark-internal

## Custom Node

light-custom

dark-custom

## Internal Node not available for non-admins

The current state sets the custom node option in case the internal one isn't allowed. So this screen doesn't occur … nevertheless I wanted to bring it up to discuss how we want to handle it. Leaving out the switch options would also be a variant, though I think defaulting to custom and displaying "Internal node" as disabled makes most sense.

light-internal-disabled


dstrukt:
This looks fantastic! @dennisreimann 

I think the only bit that confuses me a little bit is the button. Every time I read it, I would do a double take back to the top to make sure I was on the correct setting, almost expecting it to toggle to the other node option .. but maybe that's just me.

I think I still prefer "Save" or something along those lines, but that's my 2 sats.

Otherwise, I think this is pretty damn close to ready - curious to see other feedback.

dennisreimann:
The test failure in `CanOnlyUseCorrectAddressFormatsForPayjoin` seems unrelated. 

@Zaxounette This is now all yours, feel free to go ahead and propose wording changes. 👍

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
```diff
@@ -0,0 +1,163 @@
...
+    <input asp-for="LightningNodeType" value="@LightningNodeType.Custom" type="radio" id="LightningNodeType-@LightningNodeType.Custom">
+    <label asp-for="LightningNodeType" for="@($"LightningNodeType-{LightningNodeType.Custom}")">Custom node</label>
+    <div id="InternalSetup" class="text-left">
+        @if (Model.CanUseInternalNode)
+        {
+            <p class="my-4">That's it, no need to configure anything else.</p>
+        }
+        else
+        {
+            <p class="my-4">Only administrators can use the internal Lightning node.</p>
```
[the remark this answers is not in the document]
Zaxounette: Also, regarding this, I feel it should be centered.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
```diff
@@ -0,0 +1,163 @@
...
+        }
+    </div>
+    <div id="CustomSetup" class="text-left">
+        <label asp-for="ConnectionString" class="mt-4">The connection string configuration for your custom Lightning node:</label>
```
Zaxounette, line 91:
```suggestion
        <label asp-for="ConnectionString" class="mt-4">Connect your custom Lightning Node with this BTCPay Server instance using the following string:</label>
```

Suggestion. Makes it more clear that the string is used to connect to a custom node.

Zaxounette (changes requested):
Wording pass.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
```diff
@@ -0,0 +1,163 @@
+@model LightningNodeViewModel
+@{
+    Layout = "_LayoutWalletSetup.cshtml";
+    ViewData.SetActivePageAndTitle(StoreNavPages.Index, "Configure your Lightning node", Context.GetStoreData().StoreName);
```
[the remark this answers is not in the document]
dennisreimann: imho this is confusing … there is no Lightning wallet, we are configuring the connection to a specific node here.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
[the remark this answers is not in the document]
dennisreimann: I tried centering it, it also looked a bit weird imho.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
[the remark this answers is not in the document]
Zaxounette: Another suggestion then: `Connect to a Lightning Node`.

It does indeed fit better than my first suggestion.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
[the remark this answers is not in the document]
Zaxounette: Sadge :'(

If centered looks weird then leave as is.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
dennisreimann: What about "Connection configuration for your custom Lightning node: (see examples below)" like this …

cs

Zaxounette: I like the hiding mechanism, but i'm not sure about the `see examples below` and `click to show examples`.

How about just leaving out the first one `see examples below` and just staying with the second one ? Would delete the kind of repetition I'm feeling.

They are in proximity of each other on the page and a user will see both at the same time without scrolling. This might probably be true also on mobile.

dennisreimann:
Alrigght, thanks for all the suggestions @Zaxounette, I applied them and this is the current state of the custom node section now … expanded c-lightning to give an impression of what it looks like.

custom-node

Zaxounette:
Nice ! 
The `that's it, no need to config anything else` sentence might still need brainstorming.
But I'd like other's opinions on that, it might just be me.

dennisreimann:
Sure, let's nail it down, I think this is very close and I want to get it included in the upcoming release.

@NicolasDorier @dstrukt @pavlenex 

For reference …

internal

dstrukt:
l like the addition of the choice to collapse / expand them.

However, I know we have the text to indicate, "click to show", that could be a little more clear, e.g. with an accordion carrot .. 

I'm not the biggest fan of the default bootstrap accordion, and I think we need to do a little bit of custom modification to make it look a bit nicer, but it's a more clear indication of usage - mocked this up real quick.

Zaxounette:
> mocked this up real quick.

I like this addition.

dennisreimann:
Done, thanks for the superb suggestion @dstrukt!

light

dark

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
```diff
@@ -0,0 +1,201 @@
...
+
+<form method="post" class="mt-n2 text-center">
+    <input asp-for="LightningNodeType" value="@LightningNodeType.Internal" type="radio" id="LightningNodeType-@LightningNodeType.Internal" disabled="@(!Model.CanUseInternalNode)">
+    <label asp-for="LightningNodeType" for="@($"LightningNodeType-{LightningNodeType.Internal}")">Use internal node</label>
+    <input asp-for="LightningNodeType" value="@LightningNodeType.Custom" type="radio" id="LightningNodeType-@LightningNodeType.Custom">
+    <label asp-for="LightningNodeType" for="@($"LightningNodeType-{LightningNodeType.Custom}")">Use custom node</label>
+    <div id="InternalSetup" class="text-left">
+        @if (Model.CanUseInternalNode)
+        {
+            <p class="my-4">That's it, no need to configure anything else.</p>
```
Zaxounette, line 106:
```suggestion
            <p class="my-4">Using the BTCPay Server internal node for this store requires no further configuration. Click the save button below to start accepting Bitcoin through the Lightning Network.</p>
```

Simple proposal.
A really high level sentence, aimed at the less technical of our users, to inform what needs to be done (nothing) and what they'll have once they save.

Zaxounette (changes requested):
Simple proposal for the `nothing more to do` sentence.

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
dennisreimann: Awesome, thank you! 👍

dstrukt: ACK

NicolasDorier:
Code ACK, anything to add?

dennisreimann:
@NicolasDorier We are ready for liftoff. 🚀

PR #2490, opened 2021-04-26
Bootstrap v5 migration

### LINKED ISSUE #1866: Bootstrap 5 migration
This issue serves to plan and discuss the migration to [Bootstrap 5](https://v5.getbootstrap.com/), which is in alpha currently. We use Bootstrap across our products and had to make [adjustments to the current version in the design system](https://design.btcpayserver.org/resources/css/). The v5 will be a big step forward, allowing us to hopefully drop the fork approach.

@dstrukt and @Bosch-0 are making great progress with the [design specifications and mockups on Figma](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/BTCPAY). These have a cleaner look/more reduced style and involve some larger changes. We should hold those off until Bootstrap 5 is released, because it will most likely make our lives easier approaching this bigger UI redesign.

The plan I'd propose would be to

1) Update to v5 once it's at least in RC stage ([Migration docs](https://v5.getbootstrap.com/docs/5.0/migration/))
2) Make the adaptions to v5 to fit the new style on Figma – see the [Bootstrap kitchensink page in the design system](https://design.btcpayserver.org/views/bootstrap/)
3) Integrate the new css files in BTCPay Server
4) Make necessary adaptions across the products 

The main reason for this issue is to ensure everyone is on the same page and we can point towards it for clarifications in other current issues – basically as an explainer why we are holding off larger changes and discuss the migration there.

dennisreimann:
[Migrating](https://getbootstrap.com/docs/5.0/migration/) the views to use Bootstrap v5.

WIP, first draft, but looking good already :)

Closes #1866.


Kukks:
Can you check the styling on pull payments? I think there are some issues around StatusMessage and also the general body

dennisreimann:
I'm stuck here and cannot figure out where the timeouts in the Selenium tests are coming from ­— especially as some of them pass. It looks like it's always the same tests that fail, so there seems to be a pattern to it and my guess is that it might be JS/Vue-related, however it doesn't make sense that these tests don't even start running properly.

Any ideas, @NicolasDorier @kukks?

dennisreimann:
Fuck yeah, finally the tests are running again. The problem was that the [Bootstrap 5 minification does not work with BundleMinifier](https://github.com/madskristensen/BundlerMinifier/issues/558). I'll need to clean up the recent commits a bit, then this is finally ready for review!

dennisreimann:
Besides the obvious markup-related changes for the migration, this PR also contains one structural change: Before the scripts where part of the `<head>`, whereas now they got moved to the end of the `<body>`. This was introduced to work around a bug in the final Bootstrap v5.0.0, which got fixed in v5.0.1. Nevertheless this is considered best practice (that's why the Bootstrap team overlooked the case where the scripts were put in the header) and we should keep it this way.

I've split styles and scripts in the views and created the different sections `PageHeadContent` and `PageFootContent` to include them where they should appear. Before we had a single `Scripts` section, which also included styles and which was part of the `<head>`.

In parts I've also started to replace custom jQuery-based scripts with vanilla JS, as I was trying to fix the tests. Refrained from that as it would bloat up the PR even more and turned out to be unnecessary for fixing the issue. Nevertheless we should work towards migrating away from jQuery, as Bootstrap also got rid of it and we can refactor custom JS parts whenever we come across them.

Looking forward to your feedback and seeing this merged! :D

NicolasDorier:
Before
image

Now:
image


Some nits that should be fixed later, maybe some of them unrelated to this PR:

The error/success alert is off

image

Some stuff are not centered anymore:

image

image

NicolasDorier:
I notice that if I change `col-lg-10 col-xl-9` to `col-lg-10`, then it looks better.

image

NicolasDorier:
Merging since this is a large PR.
I think there is other stuff to fix after though.

dennisreimann:
@NicolasDorier Thanks for the merge and pointers, I'll go through the details with @dstrukt now 👍

dstrukt:
YES!

Excited to see the progress on this.

Also need to go through and test everything .. will follow up with bugs, etc.. as well.

dennisreimann:
They introduced a new XXL breakpoint (1400px width) in addition to the upper limit of 1200px before, that's why the site is wider now. Removing that again … will open a new PR for the fixes.

PR #2491, opened 2021-04-26
Offer enabling toggle only for wallets that are set up

dennisreimann:
The recently introduced enable toggle should only be offered for wallets that are set up … otherwise this leads to a 404.

Example: Onchain is set up, so the toggle is displayed. Offchain isn't set up, hence the togggle isn't present.

store


Zaxounette:
ACK the idea, makes perfect sense.

Would also make sense to have this merged for 1.0.8.0.

Kukks (approved):

PR #2492, opened 2021-04-26
Set lightning invoice fallback in QR code as uppercase

bjarnemagnussen:
A lightning invoice is generally encoded as uppercase at checkout inside the QR code on the "Pay with Bitcoin (Lightning)" tab. However, this is not true if an invoice is added as a fallback option to the QR code on the "Pay with Bitcoin" tab.

I am not sure if adding the fallback invoice to the QR code is purposefully done in _lowercase_ for better compatibility. Otherwise, this PR fixes it by adding it in _uppercase_.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -2209,6 +2209,10 @@ namespace BTCPayServer.Tests
                 // Standard for all uppercase characters in QR codes is still not implemented in all wallets
                 // But we're proceeding with BECH32 being uppercase
                 Assert.True($"bitcoin:{paymentMethodSecond.BtcAddress.ToUpperInvariant()}" == split);
+
+                // Fallback lightning invoice should be uppercase inside the QR code.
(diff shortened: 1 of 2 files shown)


rockstardev (changes requested):
Thanks for PR @bjarnemagnussen 

Looks great, just would suggest two things:
1. Remove outdated comment `//    + lightningFallback.ToUpperInvariant().Replace("LIGHTNING=", "lightning=", StringComparison.OrdinalIgnoreCase);`
2. Set `lightningFallback` variable right away properly when it's initialized on line 71:

```
if (!string.IsNullOrEmpty(lightningInfo?.PaymentUrls?.BOLT11))
    lightningFallback = "&" + lightningInfo.PaymentUrls.BOLT11
        .Replace("lightning:", "lightning=", StringComparison.OrdinalIgnoreCase)
        .ToUpperInvariant();
```

Kukks (commented):
LGTM, applied the suggestions from @rockstardev  too.

Kukks (approved):

PR #2504, opened 2021-05-04
Sign in with other device (quick mobile login)

Kukks:
Lets you log into btcpay using an active session when your MFA is not available on the platform. 

Some scenarios: 
* Logging in on a device when FIDO2 has been configured for specific platforms only
* Logging in on another device while being next to another authenticated device


pavlenex:
@Kukks any instructions on how to test this? Does it even make sense to test locally?

Kukks:
yes, you can. Set 2FA or FIDO2 on your account, and open another btcpay instance in another browser or in private mode. Once you enter your password, you should see an option to enter a code or scan a qr. You may find the code in your profile section

BTCPayServer/Controllers/ManageController.cs
```diff
@@ -248,5 +250,18 @@ private void AddErrors(IdentityResult result)
...
         }
         #endregion
+
+        [HttpGet]
+        public async Task<IActionResult> LoginCodes()
+        {
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                throw new ApplicationException($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
```
NicolasDorier, line 260: return just notfound

BTCPayServer/Views/Account/SecondaryLogin.cshtml
```diff
@@ -34,6 +34,16 @@
                 </div>
             }
         </div>
+        
+         <div class="row  justify-content-center">
+                    <div class="col-12">
+                        <p class="text-secondary">
+                            Not on your secured device or with access to your securtity key?
```
britttttk, line 41:
```suggestion
                            Not on your secured device or with access to your security key?
```

BTCPayServer/Views/Account/LoginWithRecoveryCode.cshtml
```diff
@@ -4,33 +4,55 @@
...
 <section>
     <div class="container">
         <div class="row">
             <div class="col-lg-12 section-heading">
                 <h2>@ViewData["Title"]</h2>
                 <hr class="primary">
                 <p>
-                    You have requested to login with a recovery code. This login will not be remembered until you provide
-                    an authenticator app code at login or disable 2FA and login again.
+                    You have requested to login with a login or recovery code. This login will not be remembered. You can find a login code in your Account=>Login Codes page when logged in.
```
britttttk, line 14:
```suggestion
                    You have requested to login with a login or recovery code. This login will not be remembered. You can find a login code in your Account > Login Codes page when logged in.
```

britttttk:
> you should see an option to enter a code or scan a qr. You may find the code in your profile section

For windows on incognito I see `This security key doesn't look familiar. Please try a different one.` is that expected?

Kukks:
> > you should see an option to enter a code or scan a qr. You may find the code in your profile section
> 
> For windows on incognito I see `This security key doesn't look familiar. Please try a different one.` is that expected?

Yes, it's because it will still prompt you for FIDO2.

bolatovumar (commented):
I'm trying to test this wit 2FA using Google Authenticator app and can't get it to work.

This is what I'm doing:

1. Set up 2FA using Google Authenticator app on my phone
2. Log in using Chrome
3. Open Safari and Chrome in incognito mode and try to log-in
4. I don't see anything special, just a regular 2FA screen so I have to put in the code again
5. Also, clicking on the link on the bottom left gives me a 500 error

Screen Shot 2021-07-06 at 12 45 14 PM
Screen Shot 2021-07-06 at 12 45 35 PM

Kukks:
> I'm trying to test this wit 2FA using Google Authenticator app and can't get it to work.
> 
> This is what I'm doing:
> 
> 1. Set up 2FA using Google Authenticator app on my phone
> 2. Log in using Chrome
> 3. Open Safari and Chrome in incognito mode and try to log-in
> 4. I don't see anything special, just a regular 2FA screen so I have to put in the code again
> 5. Also, clicking on the link on the bottom left gives me a 500 error
> 
> Screen Shot 2021-07-06 at 12 45 14 PM
> Screen Shot 2021-07-06 at 12 45 35 PM

Thank you for the test. Fixed now, was an issue with a rebase

bolatovumar (approved):
Tested using 2FA and Google Authenticator app using Chrome/Safari/Firefox and works as expected. Couldn't test the FIDO2 locally because I believe it requires HTTPS so it doesn't work locally.

One thing I was wondering if the "camera" icon on the page which allows you to enter the login code is supposed to do anything on desktop browser? I'm clicking on it in Chrome and nothing is happening. No errors in the console either.

Kukks:
> Tested using 2FA and Google Authenticator app using Chrome/Safari/Firefox and works as expected. Couldn't test the FIDO2 locally because I believe it requires HTTPS so it doesn't work locally.
> 
> One thing I was wondering if the "camera" icon on the page which allows you to enter the login code is supposed to do anything on desktop browser? I'm clicking on it in Chrome and nothing is happening. No errors in the console either.

(we have an https profile in BTCPay to use https on localhost and fido2 allows self signed certs on localhost)

Kukks:
This sounds like a bug introduced with the bs5 migration, will fix.

NicolasDorier:
@bolatovumar normally you can run btcpay in test with https and local cert. Its the `Bitcoin-HTTPS` launch profile.

Kukks:
> Tested using 2FA and Google Authenticator app using Chrome/Safari/Firefox and works as expected. Couldn't test the FIDO2 locally because I believe it requires HTTPS so it doesn't work locally.
> 
> One thing I was wondering if the "camera" icon on the page which allows you to enter the login code is supposed to do anything on desktop browser? I'm clicking on it in Chrome and nothing is happening. No errors in the console either.

Fixed, was breaking change syntax from bootstrap5

bolatovumar:
> > Tested using 2FA and Google Authenticator app using Chrome/Safari/Firefox and works as expected. Couldn't test the FIDO2 locally because I believe it requires HTTPS so it doesn't work locally.
> > One thing I was wondering if the "camera" icon on the page which allows you to enter the login code is supposed to do anything on desktop browser? I'm clicking on it in Chrome and nothing is happening. No errors in the console either.
> 
> Fixed, was breaking change syntax from bootstrap5

@Kukks ok, tested the camera integration on my laptop and it's working as expected. Scans the QR code and allows me to log in and all.

bolatovumar:
> @bolatovumar normally you can run btcpay in test with https and local cert. Its the `Bitcoin-HTTPS` launch profile.

@NicolasDorier good to know but I couldn't figure out how to use it in VSCode lol.

BTCPayServer/Controllers/ManageController.cs
Kukks: This is consistent with the rest of the controllers. We can change them all later in a separate PR
image

pavlenex:
@Kukks Do we want to proceed with this one for the next release? Thinking about LNURL-Auth you mentioned it may fit in nicely, or is it too big of an undertaking?

Kukks:
> @Kukks Do we want to proceed with this one for the next release? Thinking about LNURL-Auth you mentioned it may fit in nicely, or is it too big of an undertaking?

We should proceed with this. LNURL Auth is potentially feasible

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -140,6 +144,14 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
                                 RememberMe = model.RememberMe
                             };
                         }
+                        else
+                        {
+                            var identity = new ClaimsIdentity(IdentityConstants.TwoFactorUserIdScheme);
+                            identity.AddClaim(new Claim(ClaimTypes.Name, user.Id));
+                            
+                            identity.AddClaim(new Claim(ClaimTypes.AuthenticationMethod, "FIDO2"));
+                            await HttpContext.SignInAsync(IdentityConstants.TwoFactorUserIdScheme, new ClaimsPrincipal(identity));
```
NicolasDorier, line 153: I don't understand what this does.

NicolasDorier:
The UX is a bit weird. This is a second factor auth, but it doesn't show in the two factor section of the profile. Should ask what @dstrukt think about it.

Also, I was expecting this QR code was an easy way to log in your btcpay on mobile. (so you don't have to type your login/pass on shitty mobile, and instead just scan the code and be connected)
If that was the case, it would make sense for this feature to not be "two factor".

Kukks:
We can probably incorporate it as a 2fa method for now and then do a separate pr to allow 2fa methods as a way to replace user/pass too

BTCPayServer/Controllers/AccountController.cs
Kukks: I dug into how .net identity handles 2FA and when 2fa is enabled, the 2fa method is stored in the identity

Kukks:
> The UX is a bit weird. This is a second factor auth, but it doesn't show in the two factor section of the profile. Should ask what @dstrukt think about it.
> 
> Also, this feature is useful right now only if you have a second factor, if you don't have set it up, it shouldn't even be exposed in the profile categories.
> 
> I was expecting this QR code was an easy way to log in your btcpay on mobile. (so you don't have to type your login/pass on shitty mobile, and instead just scan the code and be connected) If that was the case, it would make sense for this feature to not be "two factor".

Can  you try again?  I changed it to be closer to what you wanted

NicolasDorier:
@Kukks you did the reverse: You just have put this QR code in 2FA, but made it "first factor" since you can bypass the password.

If you want to make it first factor, I don't think it should be in second factor, and should be in the same category you specified in an earlier PR.
Also, if you want it to be first factor, you should include the username in the QR code, so the user can log-in without having to use shitty mobile keyboard.

NicolasDorier:
Need a test as well

BTCPayServer/Controllers/ManageController.LoginCodes.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Threading.Tasks;
+using BTCPayServer.Abstractions.Extensions;
+using BTCPayServer.Abstractions.Models;
+using Microsoft.AspNetCore.Mvc;
+
+namespace BTCPayServer.Controllers
+{
+    public partial class ManageController
+    {
+        [HttpGet]
```
NicolasDorier, line 11: should be post?

NicolasDorier: I think you should require generation explicitely in the UX, to remove the risks user use an outdated code, because they took their time.
Also since the qr code is confidential, should have proper warning before generating.

dennisreimann:
tACK. Works as intended. 

The only thing I'd suggest is auto submitting the form once the QR code got scanned and recognized. Right now the user has to confirm, which seems unnecessary in this case as there is nothing the user can really verify or check - either the code will work or it won't.

dennisreimann (approved):

Kukks:
> Need a test as well

Added Test

> @Kukks you did the reverse: You just have put this QR code in 2FA, but made it "first factor" since you can bypass the password.
> 
> If you want to make it first factor, I don't think it should be in second factor, and should be in the same category you specified earlier. Also, if you want it to be first factor, you should include the username in the QR code, so the user can log-in without having to use shitty mobile keyboard. Also, you removed the possibility to copy/paste. (Unsure we need it though, but if you remove, you should not allow to copy it in the first place)
> 
> Personally, I think this feature is best as a first factor.

Adapted PR to do this

NicolasDorier:
@Kukks can you add the username to the QR code so the user on mobile don't even have to use his keyboard to connect?

NicolasDorier:
It's also not apparent that the QR Code is expiring. I think adding a small timer and auto regenerate when around expiration many issues. (Same UX as google authenticator)

NicolasDorier:
Strange flickering when I click on regenerate.

https://user-images.githubusercontent.com/3020646/145159420-c0452deb-a315-40be-b26c-09d834b91aee.mp4

dennisreimann:
Added some UI improvements and a simple countdown timer, which refreshes the code after expiry.

code

NicolasDorier:
Seems to me ready. I don't like the term "Login Code", nobody can understand what it does just by reading. 

Should we rename it in another way? Maybe "Mobile Login"? @dstrukt  @Zaxounette

dstrukt:
Sorry, I must have missed this one.

Definitely prefer "Mobile Login" to "Login Codes".

Can't think of anything better atm, but will ping in MM with other suggestions if something comes to mind.

NicolasDorier:
@Kukks feel free to merge or to wait for the bikeshedding on naming.

PR #2512, opened 2021-05-06
Make sure BTCPayServer does not take the culture of the server

NicolasDorier:
@dennisreimann reported a bug, where, running btcpayserver on a linux with some local culture would make it impossible to send money from the wallet.

This PR not only make sure that the request is always CultureInvariant, so things are predictably formatted, but on top of this, I added our model binders as default, which would allow us to remove some of of the attribute spread around the code later on.


dennisreimann:
Can confirm this fixes the issue for me.

dennisreimann:
Can you retrigger the tests? The pipeline seems to have had a hickup. Would be good to merge it in case the tests pass.

PR #2517, opened 2021-05-10
Add Lightning payout support

Kukks:


dennisreimann (commented):
Tested it and pushed some UI-related updates [on my fork](https://github.com/dennisreimann/btcpayserver/tree/payout-ln).

Works and we can improve some UI and UX bits once this is merged:

- The confirmation screens could use some love
- The table on the payee side does not work well responsive and even on the desktop the invoice string is too massive
- The pull payment creation form view could be enhanced too, but that can also be done when we add a description there

BTCPayServer/Payments/Lightning/LightningExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+using BTCPayServer.Lightning;
+using BTCPayServer.Services;
+
+namespace BTCPayServer.Payments.Lightning
+{
+    public static  class LightningExtensions
+    {
+        
+
+        public static ILightningClient CreateLightningClient(this LightningSupportedPaymentMethod supportedPaymentMethod, BTCPayNetwork network, LightningNetworkOptions options,  LightningClientFactoryService lightningClientFactory)
```
NicolasDorier, line 11: why extension method and not just add to `LightningClientFactoryService`?

Kukks: no specific reason but there wasn't a need to couple the lightning client factory with a store payment method

NicolasDorier:
Need tests

dstrukt:
> Tested it and pushed some UI-related updates [on my fork](https://github.com/dennisreimann/btcpayserver/tree/payout-ln).
> 
> Works and we can improve some UI and UX bits once this is merged:
> 
> * The confirmation screens could use some love
> * The table on the payee side does not work well responsive and even on the desktop the invoice string is too massive
> * The pull payment creation form view could be enhanced too, but that can also be done when we add a description there

Mind posting some screenshots @dennisreimann? Would be happy to help out with this, if possible.

Kixunil:
Anyone tested what happens if BTCPay doesn't have permission to send? (union of readonly and invoice macaroon) I obviously don't expect it to work but would be nice to make sure it has proper error message instead of crashing mysteriously. :)

dennisreimann:
@dstrukt Here are a few screenshots, I'll also try to make it work to pay the payouts, right now my local LN setup is failing.

## Empty state

payouts-none

## With payouts
payouts-list

## Payout view for LN request
Screenshot 2021-08-02 at 13-11-34 Contributions

## Approval confirmation

payout-approval

## Approval feedback (here failed)

payout-failed

dennisreimann:
Having this under "Wallets" is kinda weird for the LN case, because it requires the store, which is associated with that wallet, to have LN enabled. This isn't really intuitive -- imho it might make more sense to move the pull payments over to the store, because that's what it seems tight to.

Kukks:
> Having this under "Wallets" is kinda weird for the LN case, because it requires the store, which is associated with that wallet, to have LN enabled. This isn't really intuitive -- imho it might make more sense to move the pull payments over to the store, because that's what it seems tight to.

Yes, I have been saying this from the start.. IMO we should move all entities which fall under a store into a submenu of some sort

dstrukt:
> Having this under "Wallets" is kinda weird for the LN case, because it requires the store, which is associated with that wallet, to have LN enabled. This isn't really intuitive -- imho it might make more sense to move the pull payments over to the store, because that's what it seems tight to.

Seconded. Felt that way even exploring an updated wallet detail view visually as well. Thanks for the screenshots!

> Yes, I have been saying this from the start.. IMO we should move all entities which fall under a store into a submenu of some sort

As in, the store is the parent entity, and everything is accessible via that parent store view? This was a pretty old wireframe, so it's ugly as hell and probably makes less sense, but I think it carries the basic idea until I have the time to revise something, if I'm understanding you correctly... I'm curious to hear more of your thinking here in terms of organization. 

Screen Shot 2021-08-02 at 7 38 51 AM

Zaxounette (changes requested):
@dennisreimann 
In your screenshot `Payout view for LN request`, the confirmation message being so big, with part of the string at the top and the rest at the bottom can make it daunting and not easily readable.
It works for on-chain because the address is short'ish, but for LN it's way too big IMO.

I couldn't find the actual string to propose a change directly, but something in the likes of:
> Your claim request of {AMOUNT} has been submitted and is awaiting approval.
> Payout address/channel: {PAYOUT_STRING}

Kixunil:
Suggestion: have another button "Open in external wallet" to allow paying from other wallet. This would also solve the situation with secure setups that don't give BTCPayServer permissions to pay invoices.

NicolasDorier:
Somehow, I can't run the tests, no permissions? wtf

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -64,22 +64,23 @@ public async Task TrackClaim(PaymentMethodId paymentMethodId, IClaimDestination
             await explorerClient.TrackAsync(TrackedSource.Create(bitcoinLikeClaimDestination.Address));
     }
 
-    public Task<IClaimDestination> ParseClaimDestination(PaymentMethodId paymentMethodId, string destination)
+    public Task<(IClaimDestination, decimal?)> ParseClaimDestination(PaymentMethodId paymentMethodId, string destination)
```
NicolasDorier, line 67: Can you remove this hack, and instead modify `IClaimDestination` to add a `Money? Amount` property (add nullable around it).

Kukks:
> Suggestion: have another button "Open in external wallet" to allow paying from other wallet. This would also solve the situation with secure setups that don't give BTCPayServer permissions to pay invoices.

I generally agree, but then you have to click on each one individually and make sure you note the result for each one.. 
This is already possible with another new feature we added (https://github.com/btcpayserver/btcpayserver/pull/2539) so your usecase is possible :)

BTCPayServer/Controllers/WalletsController.PullPayments.cs
```diff
@@ -48,8 +52,16 @@ public partial class WalletsController
             if (GetDerivationSchemeSettings(walletId) == null)
                 return NotFound();
 
+            var storeMethods = CurrentStore.GetSupportedPaymentMethods(NetworkProvider).Select(method => method.PaymentId).ToList();
+            var paymentMethodOptions = _payoutHandlers.SelectMany(handler => handler.GetSupportedPaymentMethods()).Where(id => storeMethods.Contains(id));
```
NicolasDorier, line 56: Can you use an extensions methods for this `_payoutHandlers.SelectMany(handler => handler.GetSupportedPaymentMethods()).Where(id => storeMethods.Contains(id)` I think you use it at least 3 place now.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: Tried Money but it does not work well when you take into account that each payment method has its own monetary format that may not play well with `Money`

BTCPayServer/Controllers/WalletsController.PullPayments.cs
Kukks: done

Kixunil:
> then you have to click on each one individually

Yes, this is a stopgap solution. I think proper solution would be to have batch-payment URI scheme. Maybe something for the future.

Kukks:
@NicolasDorier  ready for merge

NicolasDorier:
@Kukks is the UI tested?

Currently doesn't work

image

BTCPayServer/Views/PullPayment/ViewPullPayment.cshtml
```diff
@@ -63,7 +63,7 @@
 
                             <div class="col-12 mb-3 col-sm-6 mb-sm-0 col-lg-3">
                                 <div class="input-group">
-                                    <input type="number" class="form-control form-control-lg text-end hide-number-spin" asp-for="ClaimedAmount" max="@Model.AmountDue" min="@Model.MinimumClaim" step="any" placeholder="Amount" required>
+                                    <input type="number" class="form-control form-control-lg text-end hide-number-spin" asp-for="ClaimedAmount" max="@Model.AmountDue" min="@Model.MinimumClaim" step="any" placeholder="Amount">
```
Kixunil, line 66: Shouldn't it be conditional on whether invoice contains the amount or not?

Kukks: There are various things we need to improve here. For now, we'll remove the requirement and let the backend return validation messages.

NicolasDorier: it's bit complicated. If it is bolt11 or bitcoin url with amount, amount is not required.
I think for this release that's fine to cut corner

Marcin198 (approved):

pavlenex:
This one is waiting for #2804

Kukks:
> This one is waiting for #2804

No longer waiting on it. Being developed in parallel now.

PR #2518, opened 2021-05-11
Fix: Payout Transaction not matching when rate provided longer decima…

### LINKED ISSUE #2513: Pull payment payout still on awaiting payment list after payout sent.
**Describe the bug**
Pull payment payout still list at "Awaiting payment" list after sent out the payout.
**To Reproduce the bug**
1. Go to 'greenfield api create pull payment, login backend > manage payout >  select all > action >  send selected payout'
2. after sent payout it still show on awaiting payment list.

**Expected behavior**

**Screenshots**
https://ibb.co/vqG389L
https://ibb.co/Wy64Jtj
https://ibb.co/hYVrjDt

- BTCPay Server Version:  BTCPayServer v1.1.0.0
- Deployment Method:        Docker
- Browser:   Chrome

**Logs (if applicable)**

**Setup Parameters**

**Additional context**

Kukks:
…l precision

fixes #2513 most likely

Bug scenario:
Create Pull payment of 10 USD
Create payout of 10 USD
Approve & send => broadcast tx
Still at `Awaiting Payment`

The issue is the cryptoAmount to send was computed as 0.123456789022323 and BTC only sends up to 0.12345678

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -471,6 +471,25 @@ namespace BTCPayServer.Tests
                 {
                     Revision = payout.Revision
                 }));
+
+                // Create one pull payment with an amount of 9 decimals
+                var test3 = await client.CreatePullPayment(storeId, new Client.Models.CreatePullPaymentRequest()
+                {
+                    Name = "Test 2",
+                    Amount = 12.303228134m,
+                    Currency = "BTC",
+                    PaymentMethods = new[] { "BTC" }
+                });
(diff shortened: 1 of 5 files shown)


NicolasDorier:
This might still fail, depending on how the amount has been rounded.
We should just make sure the CryptoAmount  is correctly rounded from beginning rather than doing this hack.

Kukks:
> This might still fail, depending on how the amount has been rounded.
> We should just make sure the CryptoAmount is correctly rounded from beginning rather than doing this hack.

True, will need it in both ends though so that the existing payouts are able to be processed too

PR #2534, opened 2021-05-19
Bootstrap migration fixups

dennisreimann:
Continuing with the finetuning of #2490. 

Bootstrap v5 introduced a new XXL breakpoint (1400px width) in addition to the upper limit of 1200px before. Removed that as we currently don't need that.

Contains some form updates too. Will extend this with the feedback from @dstrukt and the test deployment.


dennisreimann:
Been through the feedback session with @dstrukt now and will take care of the results from the testers channel next. Afterwards this baby is good to go! 

Mostly this has been minor finetunings, but there are these three areas that visually changed a bit more. So here are before and after screenshots …

## Domain mapping

### Before

domain-mapping-before

### After

domain-mapping-after

## Checkout experience

### Before

checkout-experience-before

### After

checkout-experience-after

## Webhook

### Before

webhook-before

### After

webhook-after

dstrukt:
I want to mess around with the domain mapping a bit.

Should the fee confirmation be split into it's own section "On-chain" .. or does it make sense to keep it under "General". Regardless, I definitely like that the section is broken up to make it easier to digest.

Need to go through this one more time, but everything else is lookin' good.

dennisreimann:
@dstrukt yes, let's use this as an example where you can do the pixel pushing and we can have a session where I explain things and help if needed. 👍

dennisreimann:
Addressed the issues brought up via the Testers channel (see [Google doc](https://docs.google.com/document/d/1eWjcNRL8qoLjw5I3U-A3jGRsdxstJEs4ciFru8GdDy4/edit)) and rebased with current master. Ready for review 🎉

BTCPayServer/Views/Stores/ModifyWebhook.cshtml
```diff
@@ -4,76 +4,75 @@
...
             </div>
             <div class="form-group">
-                <div class="form-check">
-                    <input asp-for="AutomaticRedelivery" type="checkbox" class="form-check-input" />
-                    <label asp-for="AutomaticRedelivery" class="form-check-label">Automatic redelivery</label>
-                    <p class="text-muted small form-text">We will try to redeliver any failed delivery after 10 seconds, 1 minutes and up to 6 times after 10 minutes</p>
-                </div>
+                <label asp-for="AutomaticRedelivery" class="form-label">Automatic redelivery</label>
+                <input asp-for="AutomaticRedelivery" type="checkbox" class="btcpay-toggle ms-2" />                
+                <p class="text-muted small form-text">We will try to redeliver any failed delivery after 10 seconds, 1 minutes and up to 6 times after 10 minutes</p>
```
britttttk, line 37:
```suggestion
                <p class="text-muted small form-text">We will try to redeliver any failed delivery after 10 seconds, 1 minute and up to 6 times after 10 minutes</p>
```

dennisreimann: Thanks for spotting this! :+1:

dennisreimann:
This is good to go. 

@Kukks can you give it a review and merge? @dstrukt could use this for his finetuning PRs :)

PR #2539, opened 2021-05-21
Mark Payouts as Paid

Kukks:
This PR allows users to mark payouts as paid manually through the UI  and through the API. It also sets up the payout proof system to be able store a manual proof that will in a later PR allow you to specify a proof of payment (link or text)

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -490,6 +490,10 @@ namespace BTCPayServer.Tests
                 // The payout should round the value of the payment down to the network of the payment method
                 Assert.Equal(12.30322814m, payout.PaymentMethodAmount);
                 Assert.Equal(12.303228134m, payout.Amount);
+
+                payout = await client.MarkPayoutPaid(storeId, payout.Id);
+                Assert.Equal(PayoutState.Completed,payout.State);
+                await AssertAPIError("invalid-state", async () => await client.MarkPayoutPaid(storeId, payout.Id));
             }
         }
 
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -491,7 +491,8 @@ namespace BTCPayServer.Tests
                 Assert.Equal(12.30322814m, payout.PaymentMethodAmount);
                 Assert.Equal(12.303228134m, payout.Amount);
 
-                payout = await client.MarkPayoutPaid(storeId, payout.Id);
+                await client.MarkPayoutPaid(storeId, payout.Id);
+                payout = (await client.GetPayouts(payout.PullPaymentId)).First(data => data.Id == payout.Id);
                 Assert.Equal(PayoutState.Completed, payout.State);
                 await AssertAPIError("invalid-state", async () => await client.MarkPayoutPaid(storeId, payout.Id));
             }
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -492,7 +492,7 @@ namespace BTCPayServer.Tests
                 Assert.Equal(12.303228134m, payout.Amount);
 
                 payout = await client.MarkPayoutPaid(storeId, payout.Id);
-                Assert.Equal(PayoutState.Completed,payout.State);
+                Assert.Equal(PayoutState.Completed, payout.State);
                 await AssertAPIError("invalid-state", async () => await client.MarkPayoutPaid(storeId, payout.Id));
             }
         }
--- a/BTCPayServer/Controllers/GreenField/PullPaymentController.cs
+++ b/BTCPayServer/Controllers/GreenField/PullPaymentController.cs
@@ -180,10 +180,7 @@ namespace BTCPayServer.Controllers.GreenField
             var pp = await _pullPaymentService.GetPullPayment(pullPaymentId);
             if (pp is null)
                 return PullPaymentNotFound();
-            using var ctx = _dbContextFactory.CreateContext();
-            var payouts = await ctx.Payouts.Where(p => p.PullPaymentDataId == pullPaymentId)
-                        .Where(p => p.State != PayoutState.Cancelled || includeCancelled)
-                       .ToListAsync();
+            var payouts = pp.Payouts .Where(p => p.State != PayoutState.Cancelled || includeCancelled).ToList();
             var cd = _currencyNameTable.GetCurrencyData(pp.GetBlob().Currency, false);
             return base.Ok(payouts
                     .Select(p => ToModel(p, cd)).ToList());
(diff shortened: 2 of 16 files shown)


bolatovumar (approved):
Tested locally and works as advertised!

One weird thing that I noticed is that there are these buttons which look like they are supposed to do something but they don't actually do anything, they are just links to their respective sections.

Screen Shot 2021-05-23 at 3 04 29 PM

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
```diff
@@ -354,5 +368,34 @@ public async Task<IActionResult> ApprovePayout(string storeId, string payoutId,
                     throw new NotSupportedException();
             }
         }
+
+        [HttpPost("~/api/v1/stores/{storeId}/payouts/{payoutId}/mark-paid")]
```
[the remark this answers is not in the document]
Kukks: It's meant to be a specific action endpoint, soc not sure if there would ever be a purpose for a GET on it?

Kukks:
> Tested locally and works as advertised!
> 
> One weird thing that I noticed is that there are these buttons which look like they are supposed to do something but they don't actually do anything, they are just links to their respective sections.
> 
> Screen Shot 2021-05-23 at 3 04 29 PM

Thanks for the review @bolatovumar! The UI already exists in that current style and the reasoning is that there would be more options there if there were more payouts available in different states. I am focusing on the base functionality in the meantime to expand usage of this pull payments feature. I'm sure @dennisreimann and @dstrukt can find an optimal UX improvement later on

dennisreimann:
Yes, these work more like tabs. We can indeed improve the clarity of that once the second Bootstrap PR I'm currently working on is merged.

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
```diff
@@ -354,5 +368,34 @@ public async Task<IActionResult> ApprovePayout(string storeId, string payoutId,
...
+            });
+            var errorMessage = PayoutPaidRequest.GetErrorMessage(result);
+            switch (result)
+            {
+                case PayoutPaidRequest.PayoutPaidResult.Ok:
+                    return await GetPayout(null, payoutId);
+                case PayoutPaidRequest.PayoutPaidResult.InvalidState:
+                    return this.CreateAPIError("invalid-state", errorMessage);
+                case PayoutPaidRequest.PayoutPaidResult.NotFound:
+                    return NotFound();
```
NicolasDorier, line 395: Can you return a non generic error? Search `InvoiceNotFound`

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -490,6 +490,10 @@ await this.AssertAPIError("invalid-state", async () => await client.ApprovePayou
                 // The payout should round the value of the payment down to the network of the payment method
                 Assert.Equal(12.30322814m, payout.PaymentMethodAmount);
                 Assert.Equal(12.303228134m, payout.Amount);
+
+                payout = await client.MarkPayoutPaid(storeId, payout.Id);
+                Assert.Equal(PayoutState.Completed,payout.State);
```
britttttk, line 495:
```suggestion
                Assert.Equal(PayoutState.Completed, payout.State);
```

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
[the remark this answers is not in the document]
dennisreimann: Agreed, as this is a modifying endpoint we should be explicit here.

dennisreimann (commented):
Works for me. I know @dstrukt is currently working on the UI part -- one idea that came to my ming is whether or not the tabs are needed at all. IMHO it might be a better option to just display the lists/tables underneath each other and remove the tabs/nav pill completely.

Kukks:
> Works for me. I know @dstrukt is currently working on the UI part -- one idea that came to my ming is whether or not the tabs are needed at all. IMHO it might be a better option to just display the lists/tables underneath each other and remove the tabs/nav pills completely.

This becomes a bit of an issue when you try  to find stuff though :/ I am also working on a separate or to add serverside paging since there are users running into perf issues (this page loads ALL payouts in the db currently)

dennisreimann:
@Kukks yeah, my suggestion was based on the assumption that there aren't too many payouts and the lists would not grow too long, which might be a problem in terms of performance as well as UI. One might easily overlook items below the fold in that case, I agree.

Let's see what dstrukt comes up with.

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
```diff
@@ -189,6 +189,21 @@ public async Task<IActionResult> GetPayouts(string pullPaymentId, bool includeCa
...
+        [HttpGet("~/api/v1/pull-payments/{pullPaymentId}/payouts/{payoutId}")]
+        [AllowAnonymous]
+        public async Task<IActionResult> GetPayout(string pullPaymentId, string payoutId)
+        {
+            if (payoutId is null)
+                return PayoutNotFound();
+            await using var ctx = _dbContextFactory.CreateContext();
+            var payout = await ctx.Payouts.Include(data => data.PullPaymentData)
+                .FirstOrDefaultAsync(p => p.Id == payoutId && (string.IsNullOrEmpty(pullPaymentId) || p.PullPaymentDataId.Equals(pullPaymentId) ));
+            if(payout is null )
```
britttttk, line 201:
```suggestion
            if (payout is null)
```

britttttk (commented):
nit

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
```diff
@@ -189,6 +189,21 @@ public async Task<IActionResult> GetPayouts(string pullPaymentId, bool includeCa
                     .Select(p => ToModel(p, cd)).ToList());
         }
 
+        [HttpGet("~/api/v1/pull-payments/{pullPaymentId}/payouts/{payoutId}")]
+        [AllowAnonymous]
+        public async Task<IActionResult> GetPayout(string pullPaymentId, string payoutId)
+        {
+            if (payoutId is null)
```
NicolasDorier, line 196: just reuse GetPayouts logic.

BTCPayServer/Controllers/GreenField/PullPaymentController.cs
```diff
@@ -189,6 +189,21 @@ public async Task<IActionResult> GetPayouts(string pullPaymentId, bool includeCa
...
+                return PayoutNotFound();
+            await using var ctx = _dbContextFactory.CreateContext();
+            var payout = await ctx.Payouts.Include(data => data.PullPaymentData)
+                .FirstOrDefaultAsync(p => p.Id == payoutId && (string.IsNullOrEmpty(pullPaymentId) || p.PullPaymentDataId.Equals(pullPaymentId) ));
```
NicolasDorier, line 200: We should not allow PullPaymentId to be not null or empty.

NicolasDorier:
Left two comments to fix, then it is OK for me.

NicolasDorier:
I think one test is broken @Kukks  `CanUsePullPaymentViaAPI `

PR #2552, opened 2021-05-28
Auto-detect language on payment page

woutersamaey:
based on the requst Accept-Language header, which is the language you configured in your browser/OS and this 99.99% accurate

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -482,7 +482,7 @@ namespace BTCPayServer.Controllers
             return View(model);
         }
 
-        private async Task<PaymentModel?> GetInvoiceModel(string invoiceId, PaymentMethodId paymentMethodId, string lang)
+        private async Task<PaymentModel?> GetInvoiceModel(string invoiceId, PaymentMethodId paymentMethodId, string? lang)
         {
             var invoice = await _InvoiceRepository.GetInvoice(invoiceId);
             if (invoice == null)
@@ -532,14 +532,17 @@ namespace BTCPayServer.Controllers
 
             var divisibility = _CurrencyNameTable.GetNumberFormatInfo(paymentMethod.GetId().CryptoCode, false)?.CurrencyDecimalDigits;
 
-            if ("auto".Equals(lang, StringComparison.InvariantCulture) || (lang == null && storeBlob.AutoDetectLanguage))
+            switch (lang?.ToLowerInvariant())
             {
-                // Auto-detect the langauge
-                lang = _languageService.AutoDetectLanguageUsingHeader(null).Code;
-            }
-            else if(lang != null)
-            {
-                lang = _languageService.FindLanguage(lang)?.Code;
+                case "auto":
+                case null when storeBlob.AutoDetectLanguage:
+                    lang = _languageService.AutoDetectLanguageUsingHeader(HttpContext.Request.Headers, null).Code;
+                    break;
+                case { } langs when !string.IsNullOrEmpty(langs):
+                {
+                    lang = _languageService.FindLanguage(langs)?.Code;
+                    break;
+                }
             }
             lang ??= storeBlob.DefaultLang;
             
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -531,6 +531,21 @@ namespace BTCPayServer.Controllers
             var paymentMethodHandler = _paymentMethodHandlerDictionary[paymentMethodId];
 
             var divisibility = _CurrencyNameTable.GetNumberFormatInfo(paymentMethod.GetId().CryptoCode, false)?.CurrencyDecimalDigits;
+
+            if ("auto".Equals(lang, StringComparison.InvariantCulture) || (lang == null && storeBlob.AutoDetectLanguage))
+            {
+                // Auto-detect the langauge
+                lang = _languageService.AutoDetectLanguageUsingHeader(null).Code;
+            }
+            else if(lang != null)
+            {
+                lang = _languageService.FindLanguage(lang)?.Code;
+            }
+            else
+            {
+                lang = storeBlob.DefaultLang;
+            }
+            
             var model = new PaymentModel()
             {
                 Activated = paymentMethodDetails.Activated,
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -541,10 +541,7 @@ namespace BTCPayServer.Controllers
             {
                 lang = _languageService.FindLanguage(lang)?.Code;
             }
-            else
-            {
-                lang = storeBlob.DefaultLang;
-            }
+            lang ??= storeBlob.DefaultLang;
             
             var model = new PaymentModel()
             {
--- a/BTCPayServer/Controllers/InvoiceController.cs
+++ b/BTCPayServer/Controllers/InvoiceController.cs
@@ -14,6 +14,7 @@ using BTCPayServer.Models;
 using BTCPayServer.Payments;
 using BTCPayServer.Rating;
 using BTCPayServer.Security;
+using BTCPayServer.Services;
 using BTCPayServer.Services.Apps;
 using BTCPayServer.Services.Invoices;
 using BTCPayServer.Services.Rates;
@@ -44,6 +45,7 @@ namespace BTCPayServer.Controllers
         private readonly PaymentMethodHandlerDictionary _paymentMethodHandlerDictionary;
         private readonly ApplicationDbContextFactory _dbContextFactory;
         private readonly PullPaymentHostedService _paymentHostedService;
+        private readonly LanguageService _languageService;
 
         public WebhookNotificationManager WebhookNotificationManager { get; }
 
@@ -59,7 +61,8 @@ namespace BTCPayServer.Controllers
             PaymentMethodHandlerDictionary paymentMethodHandlerDictionary,
             ApplicationDbContextFactory dbContextFactory,
             PullPaymentHostedService paymentHostedService,
-            WebhookNotificationManager webhookNotificationManager)
+            WebhookNotificationManager webhookNotificationManager,
+            LanguageService languageService)
         {
             _CurrencyNameTable = currencyNameTable ?? throw new ArgumentNullException(nameof(currencyNameTable));
             _StoreRepository = storeRepository ?? throw new ArgumentNullException(nameof(storeRepository));
@@ -73,6 +76,7 @@ namespace BTCPayServer.Controllers
             _paymentHostedService = paymentHostedService;
             WebhookNotificationManager = webhookNotificationManager;
             _CSP = csp;
+            _languageService = languageService;
         }
 
 
--- a/BTCPayServer/Services/LanguageService.cs
+++ b/BTCPayServer/Services/LanguageService.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using BTCPayServer.Client.Models;
@@ -34,6 +35,7 @@ namespace BTCPayServer.Services
 
             _languages = result.ToArray();
         }
+
         public Language[] GetLanguages()
         {
             return _languages;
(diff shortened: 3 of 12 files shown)


BTCPayServer/Services/LanguageService.cs
```diff
@@ -26,9 +27,11 @@ public Language(string code, string displayName)
     public class LanguageService
     {
         private readonly Language[] _languages;
+        private readonly IHttpContextAccessor _httpContextAccessor;
 
-        public LanguageService(IWebHostEnvironment environment)
+        public LanguageService(IWebHostEnvironment environment,IHttpContextAccessor httpContextAccessor)
```
britttttk, line 32:
```suggestion
        public LanguageService(IWebHostEnvironment environment, IHttpContextAccessor httpContextAccessor)
```

britttttk (commented):
nit

BTCPayServer/Services/LanguageService.cs
```diff
@@ -51,6 +54,19 @@ public Language[] GetLanguages()
 
         public Language FindBestMatch(string defaultLang)
         {
+            var _request = _httpContextAccessor.HttpContext.Request;
```
Kukks, line 57:
```suggestion
```

Kukks (changes requested):
just some small changes 👍

woutersamaey:
Actually, I wanted to make some more improvements, but need to lookup the exact syntax. 

The Accept-Language header can contain multiple values with a weight score. Now the logic chooses the first language, but a fallback would be better if a language is not in the list of supported languages. 

Sorting an array by float value is something I need to figure out in C#. 

Shouldn't be that hard to find. 

Another thing is, should we now remove the "Default Language" setting? It won't be very useful anymore. 

The Accept-Language header is 99.99% accurate and supported by all browsers, even very old ones. 

What's your take @Kukks with regard to removing this setting now?

Kukks:
> Actually, I wanted to make some more improvements, but need to lookup the exact syntax.
> 
> The Accept-Language header can contain multiple values with a weight score. Now the logic chooses the first language, but a fallback would be better if a language is not in the list of supported languages.
> 
> Sorting an array by float value is something I need to figure out in C#.
> 
> Shouldn't be that hard to find.
> 
> Another thing is, should we now remove the "Default Language" setting? It won't be very useful anymore.
> 
> The Accept-Language header is 99.99% accurate and supported by all browsers, even very old ones.
> 
> What's your take @Kukks with regard to removing this setting now?

Concept ACK from my end but better to do it in a separate PR as there are some implications around this.

woutersamaey:
@Kukks I've improved the code by checking all locales in the header, sorted by their weighting and if we support the language.

If this is to your liking, I'll create a new PR to remove the "defaultLang" field and other cleanup.

It may also be a good idea to create a Unit test for various `Accept-Language` headers, but I've never created a test before in C# so I could use some help with this.

In short, these are some cases and expected results.
```
nl,fr;q=0.7,en;q=0.5 => result should be nl (because the default weight 1 for nl)
nl;q0.1,fr;q=0.7,en;q=0.5 => result should be fr (because 0.7 is the highest number)
xx,*;q=0.5 => result should be the default, because "xx" does not exist and * has no meaning
```

woutersamaey:
Can someone please take a look at my unit test? It won't run because it doesn't know how to inject the LanguageService.
How do I inject it / configure the DI?

woutersamaey:
My unit test succeeded:
  √ CanAutoDetectLanguage [482ms]

woutersamaey:
The test that failed might have something do to with my changes. I'm investigating.

Kukks (approved):
Fixed the service and test should pass now

NicolasDorier:
Nice concept ACK. I am wondering though if it should be always on (like this PR) OR configurable. And if configurable, whether it should default to `true` OR default to `false`.

Any idea on the UX @pavlenex @dstrukt @dennisreimann @britttttk @Zaxounette  ?

Also I am wondering if this PR is not breaking the case when somebody specify the specific language to use in the checkout url directly. (I will check this)

woutersamaey:
In my humble opinion as a web dev, this should be standard on as it is 99.99% accurate since it uses the browser's language. I can't really think of a use case where someone would be using a browser/OS in a language they don't understand.

All other language options should be removed, except maybe allow a querystring parameter ?lang=nl as a specific case override.

My greater concern is with the supported languages like nl-NL and de-DE. These should be renamed to just nl and de without the country attached. If specific dialects are needed, a variation can be added later, but this too seams very unlikely (with regards to these 2 example languages).

Zaxounette:
> I am wondering though if it should be always on (like this PR) OR configurable. And if configurable, whether it should default to `true` OR default to `false`.

@NicolasDorier
Since the Invoice checkout view is the main view everybody sees, whether they have setup BTCPay (merchant) or not (customer), i'd vote for consistency in the view all over instances. So no configuration.

> My greater concern is with the supported languages like nl-NL and de-DE. These should be renamed to just nl and de without the country attached.

@woutersamaey 
This is a standardized nomenclature defined by ISO 639 worldwide.

woutersamaey:
I'm not sure how ISO fits in here. The general line of thought goes like this. 

A locale can be in 2 formats:
"de" meaning "German"
"de-DE" meaning "German as spoken in Germany". 

In Belgium some people speak German too, but there is no support for de-BE with the current setup.

The current code allows for "de" to match "de-DE", but it's unusual to match like this.

If I visit the site with my Accept-Language header set to "de-BE", it should not match "de-DE" and fallback to the next language in my list. This is why most browsers send the same language multiple times from region specific to general language. A typical header would be in this case "de-BE" first, the second "de", but never "de-DE".

Renaming "de-DE" to just "de" would make the scope bigger as it will now apply to all German speakers regardless of there you are. 

Now assume there are a few specific words that differ between Germany' German and Belgian's German, you could create "de-DE" and "de-BE" at that time and it would only need to contain the differences.

I've never seen an app translate country specific first over language so I thought maybe this is unintentional and maybe a side effect of the translation tool(s) in use?

So not seeing how ISO fits in here. I believe regular "de" is very much valid.

NicolasDorier:
@woutersamaey renaming language can be done later, I think that the best match algorithm I did was smart enough to use `de-DE` if you request `de`.

Now I am thinking about it, this PR also mean we probably have an option in the store settings which is useless: The default language.

Sadly, we use it still to show the fiat amount in wallet send. (yeah, that's dirty, and I plan to change this...)
If we merge this PR, later we need to remove it.

NicolasDorier:
Also last concern I have:

If I was a customer, buying on a site in english, but redirected to a payment page in french (because respecting my browser settings), I would wonder if I am not being scammed by the payment website.

I am fine for merging this if you think that's not a problem, but maybe we should ask to a few big merchant like namecheap or coincards. @astupidmoose

astupidmoose:
I'd rather not have this feature right now, or have the ability to turn it off at least. 

We are an English only site, but many customers are French Canadian so their browser is set as french. It would look strange to have our checkout page in French when the customer was on an English only site. It would also give the sense that we potentially provide support in French, which we do not.

Zaxounette:
> I'd rather not have this feature right now, or have the ability to turn it off at least.
> 
> We are an English only site, but many customers are French Canadian so their browser is set as french. It would look strange to have our checkout page in French when the customer was on an English only site. It would also give the sense that we potentially provide support in French, which we do not.

An idea, if technically possible, would be to pull the language settings from the previous page (in your case, the cart) and replicate the same language?

astupidmoose:
> > I'd rather not have this feature right now, or have the ability to turn it off at least.
> > We are an English only site, but many customers are French Canadian so their browser is set as french. It would look strange to have our checkout page in French when the customer was on an English only site. It would also give the sense that we potentially provide support in French, which we do not.
> 
> An idea, if technically possible, would be to pull the language settings from the previous page (in your case, the cart) and replicate the same language?

As long as its handled automatically and I don't have to do anything? lol

woutersamaey:
I see 2 possible solutions:

- If using a specific language is so important, why not add a querysting parameter for the desired language? That would force the language to be what is passed in the URL.
- We could set the default language to "auto" by default. That would trigger this feature.

I understand the point of the user suddenly changing languages, but am not sure it's a disadvantage or would trigger caution. All depends on the individual. If a site were in a language I am not fluent in, having the payment page in my native language, may even be beneficial. So this argument can be made both ways.

Kukks:
> I see 2 possible solutions:
> 
> * If using a specific language is so important, why not add a querysting parameter for the desired language? That would force the language to be what is passed in the URL.
> * We could set the default language to "auto" by default. That would trigger this feature.
> 
> I understand the point of the user suddenly changing languages, but am not sure it's a disadvantage or would trigger caution. All depends on the individual. If a site were in a language I am not fluent in, having the payment page in my native language, may even be beneficial. So this argument can be made both ways.

This is already the case. You can specify a query string to override

woutersamaey:
That should solve it for @astupidmoose then 👍🏻

NicolasDorier:
btw I am surprised this PR does not break the explicit language selection. I guess it is because selenium is not using preferred languages. Will check later.

woutersamaey:
> There is also the case when the browser does not include lang. But is it happening often?

All browsers, including very old ones, send the `Accept-Language`. I've never encountered a case where it did not send it.
If, hypothetically, the header is empty, the page would still default to English because of how the JavaScript is written on the payment page.

woutersamaey:
Yes. You just need to change the URL where the customer is sent to to pay.

woutersamaey:
@NicolasDorier what’s next for this PR please?

Kukks:
> btw I am surprised this PR does not break the explicit language selection. I guess it is because selenium is not using preferred languages. Will check later. Waiting for namecheap to comment as well.

The explicit language selection happens in the javascript side and will override the lang set by C# code.

Kukks:
> @NicolasDorier what’s next for this PR please?

I would suggest we add a "auto language detect" configuration (not just an "auto" option in language list) to ease @astupidmoose 's worries.

* Query string explicit language selection
* If (auto language detect = true) and (browser has language header), use language header lang
* Default language of store

NicolasDorier:
@woutersamaey here is what I suggest: Adding a query parameter to the query string (we already document a few of them in the docs).

For example `lang` in https://docs.btcpayserver.org/API/Greenfield/v1/#tag/Miscalleneous/paths/~1i~1{invoiceId}/get .

If `lang` is set to `auto`, then automatically deduce the language from the browser, else fallback on the default language of the store.

In the store settings, have a `Detect the automatically the language` in `Checkout experience`. Default to `false`.
If `true`, and `lang` is not set, default `lang` to `auto`. Else default `lang` to the default language.

woutersamaey:
@NicolasDorier I finished the changes you requested. Should be good to go!

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -528,6 +528,21 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
...
+            }
+            else if(lang != null)
+            {
+                lang = _languageService.FindLanguage(lang)?.Code;
+            }
+            else
+            {
+                lang = storeBlob.DefaultLang;
+            }
+
```
NicolasDorier, line 545: Add `lang ??= storeBlob.DefaultLang` here. And remove the else clause.

BTCPayServer/Controllers/InvoiceController.cs
```diff
@@ -73,6 +76,7 @@ public partial class InvoiceController : Controller
             _paymentHostedService = paymentHostedService;
             WebhookNotificationManager = webhookNotificationManager;
             _CSP = csp;
+            _languageService = languageService;
```
NicolasDorier, line 79: you don't need this chane I think.

BTCPayServer/Services/LanguageService.cs
```diff
@@ -44,35 +46,102 @@ public LanguageService(IWebHostEnvironment environment)
...
             var countryEnd = "-" + country;
-            countryMatches =
-                countryMatches
-                .Where(l => l.Code.EndsWith(countryEnd, StringComparison.OrdinalIgnoreCase));
-            return countryMatches.FirstOrDefault() ?? langMatches.FirstOrDefault();
+            countryMatches = countryMatches.Where(l =>
+                l.Code.EndsWith(countryEnd, StringComparison.OrdinalIgnoreCase)).ToList();
+            var bestMatch = countryMatches.FirstOrDefault() ?? langMatches.FirstOrDefault();
+
+            if (bestMatch != null)
```
NicolasDorier, line 129: what is this code? Should just be `return bestMatch`

BTCPayServer/Services/LanguageService.cs
```diff
@@ -44,35 +46,102 @@ public LanguageService(IWebHostEnvironment environment)
...
+
+            return null;
+        }
+
+        /**
+         * Look for a supported language that matches the given locale (can be in different notations like "nl" or "nl-NL").
+         * Example: "nl" is not supported, but we do have "nl-NL"
+         */
+        public Language FindLanguage(string locale)
         {
```
NicolasDorier, line 111: Why did you changed implementation of `FindLanguage`? I think previous logic was working fine.

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -528,6 +528,21 @@ private async Task<PaymentModel> GetInvoiceModel(string invoiceId, PaymentMethod
             var paymentMethodHandler = _paymentMethodHandlerDictionary[paymentMethodId];
 
             var divisibility = _CurrencyNameTable.GetNumberFormatInfo(paymentMethod.GetId().CryptoCode, false)?.CurrencyDecimalDigits;
+
+            if ("auto".Equals(lang, StringComparison.InvariantCulture) || (lang == null && storeBlob.AutoDetectLanguage))
```
NicolasDorier, line 532: can you document it in the swagger? The chceckout page is already documented, just need to add a note about `auto`.

woutersamaey: I've added it to the docs.

BTCPayServer/Controllers/InvoiceController.UI.cs
woutersamaey: Didn't know this existed. Nice.

Kukks:
Changes have been done as requested and rebased

Issue #2545, opened 2021-05-27
Need an API endpoint to check if a store can actually be used to create invoices


woutersamaey:
As discussed with @Kukks on Mattermost:

woutersamaey
10:25 AM

> I just ran into the following error: {"code":"generic-error","message":"Warning: No wallet has been linked to your BTCPay Store.. Obvisouly something is not configured correctly. Can I use the API to check if the store is properly configured and usable? I want to warn the admin before actually creating an invoice. This error occurred when trying to create an invoice by the customer, without the admin realizing.
> Not sure how to check for this.
> A boolean "canCreateInvoices" in the store's data would be nice for this, or maybe a list "configurationErrors" ? Maybe in the future there would be other config requirements? That way we can be flexible and allow the store to be used as long as the list is empty or the boolean is true.

	
kukks
10:35 AM

> Maybe I can add a stores/id/payment-methods which would could return a list of configured payment methods

PR #2555, opened 2021-06-02
Add "stores/{storeId}/payment-methods" endpoint

bolatovumar:
address #2545

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
@@ -45,14 +45,12 @@ namespace BTCPayServer.Controllers.GreenField
             _cssThemeManager = cssThemeManager;
         }
 
-        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
-        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/LightningNetwork")]
-        public ActionResult<IEnumerable<LightningNetworkPaymentMethodData>> GetLightningPaymentMethods(
-            [FromQuery] bool enabledOnly = false)
+        public static IEnumerable<LightningNetworkPaymentMethodData> GetLightningPaymentMethods(StoreData store, BTCPayNetworkProvider networkProvider, bool enabledOnly = false)
         {
-            var blob = Store.GetStoreBlob();
+            var blob = store.GetStoreBlob();
             var excludedPaymentMethods = blob.GetExcludedPaymentMethods();
-            return Ok(Store.GetSupportedPaymentMethods(_btcPayNetworkProvider)
+
+            return store.GetSupportedPaymentMethods(networkProvider)
                 .Where((method) => method.PaymentId.PaymentType == PaymentTypes.LightningLike)
                 .OfType<LightningSupportedPaymentMethod>()
                 .Select(paymentMethod =>
@@ -63,8 +61,15 @@ namespace BTCPayServer.Controllers.GreenField
                     )
                 )
                 .Where((result) => !enabledOnly || result.Enabled)
-                .ToList()
-            );
+                .ToList();
+        }
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/LightningNetwork")]
+        public ActionResult<IEnumerable<LightningNetworkPaymentMethodData>> GetLightningPaymentMethods(
+            [FromQuery] bool enabledOnly = false)
+        {
+            return Ok(GetLightningPaymentMethods(Store, _btcPayNetworkProvider, enabledOnly));
         }
 
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
--- a/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
@@ -27,20 +27,17 @@ namespace BTCPayServer.Controllers.GreenField
         private StoreData Store => HttpContext.GetStoreData();
         private readonly StoreRepository _storeRepository;
         private readonly BTCPayNetworkProvider _btcPayNetworkProvider;
-        private readonly IOptions<LightningNetworkOptions> _lightningNetworkOptions;
         private readonly IAuthorizationService _authorizationService;
         private readonly CssThemeManager _cssThemeManager;
 
         public StoreLightningNetworkPaymentMethodsController(
             StoreRepository storeRepository,
             BTCPayNetworkProvider btcPayNetworkProvider,
-            IOptions<LightningNetworkOptions> lightningNetworkOptions,
             IAuthorizationService authorizationService,
             CssThemeManager cssThemeManager)
         {
             _storeRepository = storeRepository;
             _btcPayNetworkProvider = btcPayNetworkProvider;
-            _lightningNetworkOptions = lightningNetworkOptions;
             _authorizationService = authorizationService;
             _cssThemeManager = cssThemeManager;
         }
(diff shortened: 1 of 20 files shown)


BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
```diff
@@ -45,22 +46,31 @@ public class StoreLightningNetworkPaymentMethodsController : ControllerBase
             _cssThemeManager = cssThemeManager;
         }
 
-        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
-        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/LightningNetwork")]
-        public ActionResult<IEnumerable<LightningNetworkPaymentMethodData>> GetLightningPaymentMethods(
-            [FromQuery] bool enabledOnly = false)
+        public static IEnumerable<LightningNetworkPaymentMethodData> GetLightningPaymentMethods(StoreData store, BTCPayNetworkProvider networkProvider, bool enabledOnly = false)
```
bolatovumar, line 49: The change here is that I moved the code into a static method so it can be called from another class that's responsible for the endpoint which returns data both for on-chain and LN payments methods.

BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
```diff
@@ -45,22 +46,31 @@ public class StoreLightningNetworkPaymentMethodsController : ControllerBase
...
+
+            return store.GetSupportedPaymentMethods(networkProvider)
                 .Where((method) => method.PaymentId.PaymentType == PaymentTypes.LightningLike)
                 .OfType<LightningSupportedPaymentMethod>()
                 .Select(paymentMethod =>
-                    new LightningNetworkPaymentMethodData(paymentMethod.PaymentId.CryptoCode,
-                        paymentMethod.GetExternalLightningUrl().ToString(), !excludedPaymentMethods.Match(paymentMethod.PaymentId)))
+                    new LightningNetworkPaymentMethodData(
+                        paymentMethod.PaymentId.CryptoCode,
+                        paymentMethod.GetExternalLightningUrl()?.ToString() ?? paymentMethod.GetDisplayableConnectionString(),
```
bolatovumar, line 60: This code was copied almost one to one from the non-static method but this line is different. Previously because `paymentMethod.GetExternalLightningUrl()` can return `null` this would consistently result in a 500 error for me (both in local development and on my actual BTCPay instance deployed in the wild). I updated this line not to result in a null reference exception.

BTCPayServer/Controllers/GreenField/StoreOnChainPaymentMethodsController.cs
```diff
@@ -36,22 +36,27 @@ public class StoreOnChainPaymentMethodsController : ControllerBase
             _walletProvider = walletProvider;
         }
 
-        [Authorize(Policy = Policies.CanViewStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
-        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/onchain")]
-        public ActionResult<IEnumerable<OnChainPaymentMethodData>> GetOnChainPaymentMethods(
-            [FromQuery] bool enabledOnly = false)
+        public static IEnumerable<OnChainPaymentMethodData> GetOnChainPaymentMethods(StoreData store, BTCPayNetworkProvider networkProvider, bool enabledOnly = false)
```
bolatovumar, line 39: Same change as above, I just made this into a static method so I can call it from another class.

BTCPayServer/Controllers/GreenField/StorePaymentMethodsController.cs
```diff
@@ -0,0 +1,39 @@
...
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/payment-methods")]
+        public ActionResult<LightningNetworkPaymentMethodData> GetPaymentMethods(
+            [FromQuery] bool enabledOnly = false
+        )
+        {
+            var storeBlob = Store.GetStoreBlob();
+            var excludedPaymentMethods = storeBlob.GetExcludedPaymentMethods();
+
+            return Ok(new {
```
bolatovumar, line 33: This just combines the on-chain and LN calls and returns them as one object. Not sure if that's what you had in mind or something else.

BTCPayServer/Controllers/GreenField/StorePaymentMethodsController.cs
```diff
@@ -0,0 +1,39 @@
...
+        private StoreData Store => HttpContext.GetStoreData();
+        private readonly BTCPayNetworkProvider _btcPayNetworkProvider;
+
+        public StorePaymentMethodsController(BTCPayNetworkProvider btcPayNetworkProvider)
+        {
+            _btcPayNetworkProvider = btcPayNetworkProvider;
+        }
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/payment-methods")]
```
bolatovumar, line 25: This is the new endpoint which gives you configured payment methods for both on-chain and LN.

BTCPayServer/Payments/Lightning/LightningSupportedPaymentMethod.cs
```diff
@@ -1,24 +1,23 @@
+#nullable enable
```
bolatovumar, line 1: Added this as previously a bunch of the fields here could actually be `null` but it wasn't made obvious which resulted in a 500 error when `.ToString()` was called on the result of `GetExternalLightningUrl` which can return `null`.

bolatovumar (commented):
@Kukks not sure if this was what you meant in your comment in #2545 but I took a stab at it. Even if this was not what you meant it's probably useful to have this endpoint. If this looks good I will go ahead and add docs/tests.

BTCPayServer/Controllers/GreenField/StorePaymentMethodsController.cs
Kukks: I think this needs to change, perhaps to return a `Dictionary<string, object> ` where the key is the payment method and the object is either 
* the actual object. 
* a generic object with common data from each payment method (enabled, general description?)

Kukks (changes requested):
I think the actual result needs to be more generic. There are also other payment types available (such as xmr and eth), which while we wont handle, should be hookable into this in the future through plugins from anyone that cares

Will also:
* swagger docs
*  Greenfield client support
* need tests

bolatovumar:
> I think the actual result needs to be more generic. There are also other payment types available (such as xmr and eth), which while we wont handle, should be hookable into this in the future through plugins from anyone that cares

Are there example of plugins hooking into BTCPay functionality out there? I'm really familiar with how plugins work currently.

> Will also:
> 
> * swagger docs
> * Greenfield client support
> * need tests

Yeah, will add it once the functionality is good.

pavlenex:
@bolatovumar Can you please resolve the conflict?

bolatovumar:
> @bolatovumar Can you please resolve the conflict?

@pavlenex I don't think resolving the conflict will help here as the main thing preventing me from continuing is that I don't know how I can make it available for plugins to hook into for other chains like XMR or whatever as @Kukks said above:

> There are also other payment types available (such as xmr and eth), which while we wont handle, should be hookable into this in the future through plugins from anyone that cares

Kukks:
Hey @bolatovumar, rebased, fixed conflicts, reafactored, added docs, added tests.

Can you try it out and see if it's good for you?

bolatovumar (commented):
@Kukks tested and working as expected.

One thing I noticed: if I try to hit the API endpoint for a store which doesn't exist I get a 403 status. Should it be a 404? I placed a breakpoint and it looks like the corresponding method doesn't even get hit if you provide a non-existent store id so it must be getting checked somewhere even before it gets to the API method. It's probably unrelated to this PR though.

Kukks:
> @Kukks tested and working as expected.
> 
> One thing I noticed: if I try to hit the API endpoint for a store which doesn't exist I get a 403 status. Should it be a 404? I placed a breakpoint and it looks like the corresponding method doesn't even get hit if you provide a non-existent store id so it must be getting checked somewhere even before it gets to the API method. It's probably unrelated to this PR though.

Yeah, since the expected permission for these endpoints is to modify a store, we try to check if the current API key has access to that specific store id that is present in the route. Since this check happens during permissions, the error is automatically a 403

PR #2559, opened 2021-06-03
Wallet: Signing UI improvements

### LINKED ISSUE #2508: Improve the UI of signing (sign with)
It's been proposed recently that we should have a better signing UI than clicking on the `sign with` button which currently provides unclear options. 
Screenshot 2021-05-06 at 11 14 44

Perhaps a `Sign` button opens up a new page/view similar to what we have in the wallet setup.

Screenshot 2021-05-06 at 11 18 09

Opening this up for concrete suggestions.

dennisreimann:
See prior discussion in #2508. Needs some more input, but ready for discussion:

- [x] Wording of the titles and descriptions
- [x] Do we want to recomment/discourage use of specific options?
- [x] An additional icon for PSBT would be good. @dstrukt

signing

Closes #2508.


BTCPayServer/Views/Wallets/WalletSigningOptions.cshtml
```diff
@@ -0,0 +1,102 @@
...
+        <vc:icon symbol="back" />
+    </a>
+    <a asp-all-route-data="Model.RouteDataBack" class="cancel">
+        <vc:icon symbol="close" />
+    </a>
+}
+
+<header class="text-center">
+    <h1>Choose your signing method</h1>
+    <p class="lead text-secondary mt-3">You can sign the transaction using on of the following methods.</p>
```
britttttk, line 21:
```suggestion
    <p class="lead text-secondary mt-3">You can sign the transaction using one of the following methods.</p>
```

dennisreimann: Thanks for spotting! 👀

NicolasDorier:
I am wondering as well: Knowing that most likely, for a wallet, this choice 99% of the time will never change. Can we take advantage of this fact to "pre select" the last one selected?

We can use the preference cookie to store this.

NicolasDorier:
About the preference cookie: Forget about that for now, cookie have a size limit and I think we'll reach to it if we store too much there. I'll need to do something else. But at least in the UX for now it will be nice to think about it.

NicolasDorier:
I confirm: I will take care of wallet preferences later.  But for this PR it would be nice to think about how we can limit the number of choice the user has to make by taking advantage of the fact more of the settings does not change.

pavlenex:
@NicolasDorier We discussed this on the [design call #10](https://github.com/btcpayserver/organization/issues/38) The idea is to remove the `hot wallet`. If the hot wallet is created give user the ability to automatically sign with a hot wallet, without having to choose anything, just by clicking `send`. People use hot-wallet for the convenience of spending funds, and we should cater to that need.

I'm not sure if we can have other pre-sets. We should always give the option to users to sign with seed in case their hardware wallet gets lost or broken.

dennisreimann:
I agree with Pav and think the step here should be to introduce the logic to:
- not present this screen and immediately proceed with signing for the hot wallet case
- show the first three options for non-hot wallets

Everything beyond that can be implemented in a separate PR. Will proceed in that direction here.

dennisreimann:
Rebased and updated as described above: If it's a hot wallet, you directly sign with it and do not get to see the signing options page.

Also improved the wording and added a link to the wallet PSBT docs, which get updated in btcpayserver/btcpayserver-doc#916.

signing

Zaxounette (approved):
ACK the wording
ACK the `hot wallet` deletion that was talked about.

dennisreimann:
Tests should work fine, please nudge them :)

NicolasDorier:
cool make sense

dennisreimann:
@NicolasDorier I removed the form route params, but I think for the back case they still make sense -- the alternative would be using a `javascript:history.back()` link.

Also changed the PSBT option to go to the `decode` command, rather than analyse and came across a more general issue: I think we can improve the PSBT case, because the flow is kind of weird when you are coming from the send view… after Send > PSBT you get this:

psbt

Having "Sign transaction" as the primary action does not make sense for this case, as it would take you back to the signing options view you just saw. What you'd actually want to use is one of the "Other actions", most likely Download or Show QR (depending on your wallet). Once you've processed it with the signing wallet, you'll want to import it back into BTCPay and "Review" it.

I think we should separate the two cases depending on where one originates: When you start by importing a PSBT, having the "Sign transaction" option makes sense. For the Send flow I'd recommend we restructure this and pull the export actions out of the dropdown, make them primary and offer a clear way to afterwards import the resulting PSBT.

dstrukt:
ACK'd hot wallet decision as well (as per consensus on our last call)

Beautiful progress!

Can we add this bit to this week's design call, regardless of what progress is made by then? 

Would definitely be good to run through!

NicolasDorier:
cool will review one more time and merge that soon. I want ot use this to build the RBF/CPFP support.

dennisreimann:
@NicolasDorier Ok, if it's good code wise let's merge it in its current state and we'll figure out how to improve the case I described in a separate PR.

PR #2563, opened 2021-06-10
try move coinswitch as a fully separate plugin

Kukks:


pavlenex:
@Kukks Do you need any feedback/testing for this one at this stage?

What do we need to proceed with it?

Kukks:
> @Kukks Do you need any feedback/testing for this one at this stage?
> 
> What do we need to proceed with it?

Requires a bit more work, I am still experimenting and only put it up to share my progress of solution exploration

PR #2567, opened 2021-06-14
Fix bug with LN payment method API endpoint throwing 500

### LINKED ISSUE #2566: /api/v1/stores/{storeId}/payment-methods/LightningNetwork endpoint throws 500 error for stores with internal LN node
**Describe the bug**
If you try to hit the `/api/v1/stores/{storeId}/payment-methods/LightningNetwork` endpoint on a store with an internal LN node you will get a 500 error

**To Reproduce the bug**
1. Make a `GET` request to a `/api/v1/stores/{storeId}/payment-methods/LightningNetwork` endpoint
2. Observe that you get a 500 error

**Expected behavior**

Should not get a 500 error.

**Screenshots**

Screen Shot 2021-06-13 at 6 58 54 PM

- BTCPay Server Version:  1.1.2.0
- Deployment Method:   Docker
- Browser:  any client

bolatovumar:
close #2566

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -11,7 +12,6 @@ using BTCPayServer.Lightning;
 using BTCPayServer.Payments;
 using BTCPayServer.Payments.Lightning;
 using BTCPayServer.Security;
-using BTCPayServer.Services;
 using BTCPayServer.Services.Stores;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -56,8 +56,12 @@ namespace BTCPayServer.Controllers.GreenField
                 .Where((method) => method.PaymentId.PaymentType == PaymentTypes.LightningLike)
                 .OfType<LightningSupportedPaymentMethod>()
                 .Select(paymentMethod =>
-                    new LightningNetworkPaymentMethodData(paymentMethod.PaymentId.CryptoCode,
-                        paymentMethod.GetExternalLightningUrl().ToString(), !excludedPaymentMethods.Match(paymentMethod.PaymentId)))
+                    new LightningNetworkPaymentMethodData(
+                        paymentMethod.PaymentId.CryptoCode,
+                        paymentMethod.GetExternalLightningUrl()?.ToString() ?? paymentMethod.GetDisplayableConnectionString(), 
+                        !excludedPaymentMethods.Match(paymentMethod.PaymentId)
+                    )
+                )
                 .Where((result) => !enabledOnly || result.Enabled)
                 .ToList()
             );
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/GreenField/StoreLightningNetworkPaymentMethodsController.cs
```diff
@@ -56,8 +56,12 @@ public class StoreLightningNetworkPaymentMethodsController : ControllerBase
                 .Where((method) => method.PaymentId.PaymentType == PaymentTypes.LightningLike)
                 .OfType<LightningSupportedPaymentMethod>()
                 .Select(paymentMethod =>
-                    new LightningNetworkPaymentMethodData(paymentMethod.PaymentId.CryptoCode,
-                        paymentMethod.GetExternalLightningUrl().ToString(), !excludedPaymentMethods.Match(paymentMethod.PaymentId)))
+                    new LightningNetworkPaymentMethodData(
+                        paymentMethod.PaymentId.CryptoCode,
+                        paymentMethod.GetExternalLightningUrl()?.ToString() ?? paymentMethod.GetDisplayableConnectionString(),
```
bolatovumar, line 61: This line is the main change. Previously it could result in a 500 because `GetExternalLightningUrl` can return `null`.

BTCPayServer/Payments/Lightning/LightningSupportedPaymentMethod.cs
```diff
@@ -1,24 +1,23 @@
...
         [JsonIgnore]
         public PaymentMethodId PaymentId => new PaymentMethodId(CryptoCode, PaymentTypes.LightningLike);
 
         [Obsolete("Use Get/SetLightningUrl")]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public string LightningConnectionString { get; set; }
+        public string? LightningConnectionString { get; set; }
 
-        public LightningConnectionString GetExternalLightningUrl()
+        public LightningConnectionString? GetExternalLightningUrl()
```
bolatovumar, line 20: Marked this and other things as nullable for more type safety.

BTCPayServer/Payments/Lightning/LightningSupportedPaymentMethod.cs
```diff
@@ -1,24 +1,23 @@
...
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Payments.Lightning
 {
     public class LightningSupportedPaymentMethod : ISupportedPaymentMethod
     {
         public const string InternalNode = "Internal Node";
-        public string CryptoCode { get; set; }
+        public string? CryptoCode { get; set; }
```
NicolasDorier, line 11: are you sure it can be null? Maybe the way we initialize it, but this should not happen.

bolatovumar: @NicolasDorier C# says it can be `null` because it's not initialized with any value so it complains.

Kukks: Yeah #nullable will complain about most things but it's ok (I think you could do a flag that tells the compiler that it is NOT null though.. in typescript it was something like adding `!` )

Kukks (approved):
ACK

NicolasDorier:
if it's not nullable, but we say it is, it is even worse than if we said ntohgin, will change this.

PR #2572, opened 2021-06-16
Better webhook UI. Full URLs are visible.

woutersamaey:
Better webhook page that can now display the full URLs instead of cutting off at 300px.

# Screenshots before
## Desktop
Screenshot from 2021-06-16 08-59-06
## Mobile
Screenshot from 2021-06-16 08-55-52

# Screenshots after
## Desktop
Screenshot from 2021-06-16 08-54-23
## Mobile
Screenshot from 2021-06-16 08-55-33

### DIFF
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -20,10 +20,6 @@
     white-space: nowrap;
 }
 
-td.url{
-    word-break: break-word;
-}
-
 @media (min-width: 768px) {
     td.actions {
         white-space: nowrap;
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -21,7 +21,7 @@
 }
 
 @media (min-width: 768px) {
-    .sm\:text-nowrap {
+    .text-sm-nowrap {
         white-space: nowrap;
     }
 }
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -20,6 +20,16 @@
     white-space: nowrap;
 }
 
+td.url{
+    word-break: break-word;
+}
+
+@media (min-width: 768px) {
+    td.actions {
+        white-space: nowrap;
+    }
+}
+
 @media (min-width: 768px) {
     .wraptextAuto {
         max-width: 600px;
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -21,7 +21,7 @@
 }
 
 @media (min-width: 768px) {
-    td.actions {
+    .sm\:text-nowrap {
         white-space: nowrap;
     }
 }
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -21,7 +21,7 @@
 }
 
 @media (min-width: 768px) {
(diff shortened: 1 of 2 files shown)


woutersamaey:
@dennisreimann for your consideration...

dennisreimann:
I think @dstrukt also has ideas for that :)

dennisreimann (approved):

dennisreimann:
Yes, it's an improvement already, I agree.

BTCPayServer/wwwroot/main/site.css
```diff
@@ -20,6 +20,16 @@
     white-space: nowrap;
 }
 
+td.url{
```
Kukks, line 23: can we use existing bootstrap utilities @dennisreimann ?

woutersamaey: Yes we should use `.text-break`

woutersamaey: Added a new commit to fix this.

woutersamaey (commented):
Now using `.text-break`

BTCPayServer/wwwroot/main/site.css
```diff
@@ -20,6 +20,12 @@
     white-space: nowrap;
 }
 
+@media (min-width: 768px) {
+    td.actions {
```
Kukks, line 24: There is also the class `text-nowrap` for this

Kukks (changes requested):

BTCPayServer/wwwroot/main/site.css
woutersamaey: Well, I'm only wrapping for non-mobile screens, so it's not 100% the same.
If we wrap on mobile, the action column becomes very wide and the URLs are hard to read.
I did consider this, but as far as I know, Bootstrap utility classes cannot be used for 1 screen size and not another.
Other frameworks offer this using prefixes like `sm:text-nowrap` and `md:text-nowrap`, but I don't believe Bootstrap supports this.

Kukks: Ok, that makes sense. Can you make it less specific then? Maybe make the selector `.sm-txt-no-wrap`?

woutersamaey: Following other frameworks, I created `.sm:text-no-wrap`. This means it will be the same as `.text-no-wrap` on screens SM and bigger.
The prefix `sm:` makes it easy to read and less confusing.

dennisreimann: Using `sm:` doesn't follow how Bootstrap does responsive classes and I think we should not introduce a separate approach here. So that'd be `text-sm-nowrap` etc.

woutersamaey: Ok, so you mean like in the line of `col-sm-8` that `text-sm-...` would fit.

woutersamaey: I've made the change.

dennisreimann: Thanks,! One thing I noticed though: The SM breakpoint works from 576px on, yours is targeting MD (768px). Here's a reference for that: https://getbootstrap.com/docs/5.0/layout/breakpoints/

woutersamaey: You're right. It should be `md`. We could to `sm`, but `md` is better UI in this case.
I've renamed the class to reflect `md` and not `sm`

PR #2575, opened 2021-06-17
Cleanup some old code in wallet setup

NicolasDorier:
**Remove the possibility to enter a hint address in the address confirmation screen.** 

This is because this made only sense when the xpub is manually entered. 
Nowadays, there is almost no reason to enter a xpub manually, and we do not recommend doing this.

**Remove the DerivationSchemeViewmodel.Enabled property**

This property existed because originally the enable/disable checkbox of a payment method was inside the setup wallet flow.
This is not the case anymore, so this property is useless.

**Encrypt WalletSetupViewModel.Config**

This is to make sure a client can't inject arbitrary config, bypassing the checks we have.

### DIFF
--- a/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
+++ b/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
@@ -954,28 +954,6 @@ normal:
 
             result = testnetParser.Parse(tpub);
             Assert.Equal(tpub, result.ToString());
-            testnetParser.HintScriptPubKey = BitcoinAddress
-                .Create("tb1q4s33amqm8l7a07zdxcunqnn3gcsjcfz3xc573l", testnetParser.Network).ScriptPubKey;
-            result = testnetParser.Parse(tpub);
-            Assert.Equal(tpub, result.ToString());
-
-            testnetParser.HintScriptPubKey = BitcoinAddress
-                .Create("2N2humNio3YTApSfY6VztQ9hQwDnhDvaqFQ", testnetParser.Network).ScriptPubKey;
-            result = testnetParser.Parse(tpub);
-            Assert.Equal($"{tpub}-[p2sh]", result.ToString());
-
-            testnetParser.HintScriptPubKey = BitcoinAddress
-                .Create("mwD8bHS65cdgUf6rZUUSoVhi3wNQFu1Nfi", testnetParser.Network).ScriptPubKey;
-            result = testnetParser.Parse(tpub);
-            Assert.Equal($"{tpub}-[legacy]", result.ToString());
-
-            testnetParser.HintScriptPubKey = BitcoinAddress
-                .Create("2N2humNio3YTApSfY6VztQ9hQwDnhDvaqFQ", testnetParser.Network).ScriptPubKey;
-            result = testnetParser.Parse($"{tpub}-[legacy]");
-            Assert.Equal($"{tpub}-[p2sh]", result.ToString());
-
-            result = testnetParser.Parse(tpub);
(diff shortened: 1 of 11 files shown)


BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
```diff
@@ -179,6 +179,7 @@ public async Task CanSetupWallet()
                 DerivationSchemeSettings.TryParseFromWalletFile(content, onchainBTC.Network, out var expected);
                 Assert.Equal(expected.ToJson(), onchainBTC.ToJson());
 
+                await controller.SetWalletEnabled(storeId, cryptoCode, true);
```
NicolasDorier, line 182: Ping @dennisreimann my PR seems to have changed one behavior: If BTC payment method with a wallet set up is disabled, then replacing the wallet would automatically re-enable the payment method.

Is it expected?

dennisreimann: Yes, we did that on purpose, because most likely you'll want to use the new wallet.

pavlenex:
Concept ACK.

pavlenex:
cc @Zaxounette @britttttk this PR potentially requires doc changes.

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -136,6 +142,20 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
...
         }
 
+        private async Task MigrateHotwalletProperty()
+        {
+            await using var ctx = _DBContextFactory.CreateContext();
+            foreach (var store in await ctx.Stores.AsQueryable().ToArrayAsync())
+            {
+                foreach (var paymentMethod in store.GetSupportedPaymentMethods(_NetworkProvider).OfType<DerivationSchemeSettings>())
+                {
+                    paymentMethod.IsHotWallet = paymentMethod.Source == "NBXplorer";
```
dennisreimann, line 152: I'm not sure we can rely on this check alone. The `Source` is also `NBXplorer` for watch-only wallets that got generated after my wallet redesign refactoring. ([reference](https://github.com/btcpayserver/btcpayserver/pull/2575/files#diff-1aa28c96817f27566a9d2affe6e6c4be151c6ec314a2fb26b5a7b25e82170829L301))

NicolasDorier: I know. But this does not change current behavior.
Currently the IsHotWallet is true if the Source is NBXplorer.
We can't really make difference for those old wallet, so the best we can do is to at least keep current behavior for old wallets, and change the behavior for new wallet.

dennisreimann: I though maybe we could use the check for the mnemonic being present in the migration code, so that these cases would be properly convertable.

Zaxounette:
> cc @Zaxounette @britttttk this PR potentially requires doc changes.

[Issue created](https://github.com/btcpayserver/btcpayserver-doc/issues/928)

BTCPayServer/Hosting/MigrationStartupTask.cs
NicolasDorier: The problem is that currently it is possible to make a cold wallet with Source=NBXplorer.

However, if this cold wallet is a hot wallet on a separate store, and we checked if the mnemonic is present, then the coldwallet `IsHotWallet` would be `true`.

Note that having `IsHotWallet` to `true` is not dangerous if the mnemonic is not present. It will just crash later.
What we want to prevent is having `IsHotWallet` to `true` for somebody only knowing the xpub.
Sadly we can't detect that for old wallets.

PR #2578, opened 2021-06-18
Update WebhookDataBase Swagger definition

### LINKED ISSUE #2577: Greenfield API doc error. Creating a new webhook field "specificEvents" should be an array of strings, not a string.
The Greenfield API doc says that `specificEvents` should be a string when creating a new webhook, but it appears that it should be an ARRAY of strings instead.

So, looks like a documentation issue, but not 100% sure.

Did know if this issue belongs in this repo or in the docs repo, so I've created it here (for now).

**Screenshots**
image

bolatovumar:
close #2577

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
@@ -571,13 +571,16 @@
                         "description": "Which event should be received by this endpoint",
                         "properties": {
                             "everything": {
-                                "type": "string",
+                                "type": "boolean",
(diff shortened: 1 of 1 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -571,13 +571,13 @@
                         "description": "Which event should be received by this endpoint",
                         "properties": {
                             "everything": {
-                                "type": "string",
```
bolatovumar, line 574: This was also incorrect.

PR #2592, opened 2021-06-21
GreenField: Add StoreId to Invoice model

Kukks:
solves #2591


dennisreimann (approved):
As discussed previously, good addition.

PR #2595, opened 2021-06-21
Greenfield: Add CanModifyInvoices Permissions

Kukks:
solves #2594


dennisreimann (approved):

NicolasDorier:
@Kukks seems good to me but you need also to specify that this is a permission included in modifystore, it makes sense and we avoid breaking change.

Kukks:
> @Kukks seems good to me but you need also to specify that this is a permission included in modifystore, it makes sense and we avoid breaking change.

It is here: https://github.com/btcpayserver/btcpayserver/pull/2595/files#diff-76423341f3bef419f44b4f018e00f4586ff234d2cad4799dfefe1fc164212534R174

bolatovumar (approved):

PR #2598, opened 2021-06-21
Improved the names of the Invoice webhooks to reflect the event type in code + better descriptions

### LINKED ISSUE #2580: Greenfield doc issue: 6 Event Names do not match the 6 values of the `type` field
**Describe the bug**
Just realized some confusing shit in the Docs. The event names do NOT match the type field.

There are 6 events and 6 `type` values.

Payment Received = InvoiceReceivedPayment
InvoicePaidInFull = Invoice processing???
Invoice settled = InvoiceConfirmed ???
The other 3 are self-explanatory

**Screenshots**
image

Can't we name the events like the `type` value to avoid confusion?

This actually cost me quite some time :(

woutersamaey:
Fixes #2580

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
@@ -888,10 +888,10 @@
                 }
             }
         },
-        "InvoiceConfirmed": {
+        "InvoiceSettled": {
             "post": {
-                "summary": "InvoiceConfirmed",
-                "description": "Triggers when an invoice has enough confirmations on the blockchain according to your store's configuration, making the payment settled.",
+                "summary": "InvoiceSettled",
+                "description": "Triggers when an invoice is considered settled and the merchant can proceed with the order's delivery. The invoice now has enough confirmations on the blockchain according to your store's configuration.",
                 "parameters": [
                     {
                         "in": "header",
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
@@ -738,9 +738,9 @@
         }
     },
     "x-webhooks": {
-        "Invoice created": {
+        "InvoiceCreated": {
             "post": {
-                "summary": "Invoice created",
+                "summary": "InvoiceCreated",
                 "description": "A new invoice has been created",
                 "parameters": [
                     {
@@ -768,9 +768,9 @@
                 }
             }
         },
-        "Invoice expired": {
+        "InvoiceExpired": {
             "post": {
-                "summary": "Invoice expired",
+                "summary": "InvoiceExpired",
                 "description": "An invoice expired",
                 "parameters": [
                     {
@@ -798,9 +798,9 @@
                 }
             }
         },
-        "Payment received": {
+        "InvoiceReceivedPayment": {
             "post": {
-                "summary": "Payment received",
+                "summary": "InvoiceReceivedPayment",
                 "description": "An invoice received a payment",
                 "parameters": [
                     {
@@ -828,10 +828,10 @@
                 }
             }
         },
-        "Invoice processing": {
+        "InvoicePaidInFull": {
             "post": {
-                "summary": "Invoice processing",
-                "description": "An invoice has been fully paid",
+                "summary": "InvoicePaidInFull",
+                "description": "Triggers when an invoice is fully paid, but doesn't have the required amount of confirmations on the blockchain yet according to your store's settings.",
                 "parameters": [
                     {
                         "in": "header",
@@ -858,9 +858,9 @@
                 }
             }
         },
-        "Invoice invalid": {
+        "InvoiceInvalid": {
             "post": {
-                "summary": "Invoice invalid",
+                "summary": "InvoiceInvalid",
                 "description": "An invoice became invalid",
                 "parameters": [
                     {
@@ -888,10 +888,10 @@
                 }
             }
         },
-        "Invoice settled": {
+        "InvoiceConfirmed": {
             "post": {
-                "summary": "Invoice settled",
-                "description": "An invoice has been settled",
+                "summary": "InvoiceConfirmed",
+                "description": "Triggers when an invoice has enough confirmations on the blockchain according to your store's configuration, making the payment settled.",
                 "parameters": [
                     {
                         "in": "header",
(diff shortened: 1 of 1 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -888,10 +888,10 @@
                 }
             }
         },
-        "Invoice settled": {
+        "InvoiceConfirmed": {
```
Kukks, line 891: @NicolasDorier  I think we released webhooks with the wrong naming for this event... oh well

Kukks (approved):
I think it looks ok to me

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
Zaxounette: > oh well

Why not correct the name?
We'll just add a big ass disclaimer for the name change in the next release logs and blog post.

Kukks:
No, that is unacceptable as it is a breaking change, that will break any
install that updates that uses it without warning, which deteriorates trust
in developers resulting in less integrations overall.

woutersamaey:
What's the next step for this PR?

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
NicolasDorier: It should be `InvoiceSettled`

woutersamaey: Are you sure @NicolasDorier ?
I'm naming the events to match the`type` value. The doc says `InvoiceConfirmed`

image

NicolasDorier: I am certain it is `WebhookEventType.InvoiceSettled`

woutersamaey: The question is: How come the doc is wrong? I looked at the code and don't understand how this is possible.

NicolasDorier: What do you mean? The doc is relying on the type `WebhookEventType` and `WebhookEventType.InvoiceConfirmed` does not exists.

In the greenfield API there is no concept of invoice confirmed/complete instead both state are "settled" because there is no reason to treat them differently. I just fucked up when writing the doc.

woutersamaey: Okay, final try here with screenshots and all...

@NicolasDorier I mean: why does the doc say that the `type` field can have a value of `InvoiceConfirmed`  when this is a wrong / non-existing value for this field:
image

Looking at the code, the `type` field comes from this:
image

And the enum values are these:
image

There is no  `InvoiceConfirmed` enum value here, but the docs still says so... so we look at the docs!

The doc JSON talks about `type` being `InvoiceSettled` in the `description` (strange!), but the live docs have no `InvoiceSettled`, only `InvoiceConfirmed` (screenshot above vs code fragment below). The doc is generated by examining `#/components/schemas/WebhookInvoiceEvent` and as I just showed, this type does NOT have the enum printed in the live docs.
image

So, do we need to re-generate the docs? Or is there some other weirdness going on?

pavlenex:
@woutersamaey, @Kukks mentioned he needs to double-check a few things including if webhooks are right, so we'll see how to proceed afterward.

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
bolatovumar: @woutersamaey I think the docs are just outdated. I made a PR to fix them here: https://github.com/btcpayserver/btcpayserver/pull/2681

NicolasDorier: @woutersamaey  The doc is written by hands, so yes, there might have some differences and here the doc is wrong.
@bolatovumar fixed it.

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -888,10 +888,10 @@
...
             "post": {
-                "summary": "Invoice settled",
-                "description": "An invoice has been settled",
+                "summary": "InvoiceConfirmed",
+                "description": "Triggers when an invoice has enough confirmations on the blockchain according to your store's configuration, making the payment settled.",
```
NicolasDorier, line 894: This is too technical imho. I would rather say `When triggered, the invoice is considered settled and the merchant can proceed to the order's delivery.`. Maybe @pavlenex  or @Zaxounette have better idea?

woutersamaey: Your text is very good, but I'd still like to mention the confirmations as some more tech savvy merchants do wonder about this. I've had questions about this particular status by merchants. 

So we make a combination?

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
bolatovumar, line 891: As long as this is updated to `InvoiceSettled` this PR should be good to go.

woutersamaey: Changed it.

woutersamaey:
I believe we're now good to go?

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
woutersamaey: I have put in a mix of both texts.

bolatovumar (approved):

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -888,10 +888,10 @@
                 }
             }
         },
-        "Invoice settled": {
+        "InvoiceSettled": {
             "post": {
-                "summary": "Invoice settled",
-                "description": "An invoice has been settled",
+                "summary": "InvoiceSettled",
+                "description": "Triggers when an invoice is considered settled and the merchant can proceed with the order's delivery. The invoice now has enough confirmations on the blockchain according to your store's configuration.",
```
Kukks, line 894:
```suggestion
                "description": "Triggers when an invoice is considered settled and the merchant can proceed with the order's delivery. The invoice now has enough confirmations on the blockchain (if paid on-chain) according to your store's configuration.",
```

Kukks (commented):
confirmations are only a concept for a specific payment method

PR #2599, opened 2021-06-21
Add permission key to API page

woutersamaey:
Now we can see the permission code + Fixed a margin-bottom when selecting a specific store

Screenshot:
image

### DIFF
--- a/BTCPayServer/Views/Manage/AddApiKey.cshtml
+++ b/BTCPayServer/Views/Manage/AddApiKey.cshtml
@@ -49,18 +49,26 @@
                     {
                         <div class="list-group-item form-group py-3">
                             <div class="form-check">
-                                <input id="@Model.PermissionValues[i].Permission" type="checkbox" asp-for="PermissionValues[i].Value" class="form-check-input ms-n4" />
-                                <label for="@Model.PermissionValues[i].Permission" class="h5 form-check-label me-2 mb-1">@Model.PermissionValues[i].Title</label>
-                                <button type="submit" class="btn btn-link p-0 mb-1" name="command" value="@($"{Model.PermissionValues[i].Permission}:change-store-mode")">Select specific stores</button>
-                                <span asp-validation-for="PermissionValues[i].Value" class="text-danger"></span>
-                                <span class="form-text text-muted">@Model.PermissionValues[i].Description</span>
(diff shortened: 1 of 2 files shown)


dennisreimann:
Code-wise this looks good, but the UI could be better imho. I don't like that the options are completely separated -- also the separated list group style gives me the impression I'd be able to click the whole area (as e.g. in the wallet setup or when choosing the signing options).

I'd rather see the list group items combined or get rid of them completely -- the latter might not be good because the list group items help to refer to the additional specific store options.

woutersamaey:
@dennisreimann I agree.

It was not my intention of rebuilding the page, just have the permission key on the page as it's otherwise hard to know which permission to check when working with Greenfield.

Zaxounette:
Would it make sense to nest the permission names behind a clickable `!`, like when we have a clickable `?` for more info?

woutersamaey:
That makes finding the permissions you need harder as you need to try out every tooltip.
I was thinking more on the lines of a table-like structure. Not a table per se as it’s not responsive, but something like it.

Again, it was not my intention to rebuild the page, just add missing and essential information.

dennisreimann:
@woutersamaey Can you give me access to your fork so that I can push a commit here? I improved the list, especially the responsive view of the smaller breakpoints.

woutersamaey:
@dennisreimann I've sent you an invite. I wasn't sure how best to give you access, so I hope it did it the right way.

dennisreimann:
I'm not a maintainer here, just a regular pleb. 😂

Thanks, will push my updates!

NicolasDorier:
Ready to push?

dennisreimann:
I've added my update -- @woutersamaey can you confirm everything's ok and this is ready to merge?

woutersamaey:
It works, but I'm not happy with the experience on mobile. This isn't new so let's merge this and we can discuss on our next call.

The issue I have on mobile is that when you click "Select specific stores" you navigate to a new page and need to scroll back down. We can improve by using a `#hash` or by using JavaScript, but let's keep this for next time.

woutersamaey:
@NicolasDorier ready!

dennisreimann:
> The issue I have on mobile is that when you click "Select specific stores" you navigate to a new page and need to scroll back down. We can improve by using a #hash or by using JavaScript, but let's keep this for next time.

Good catch! Added it :)

PR #2601, opened 2021-06-21
Export all checked invoices. Not just invoices with payments.

### LINKED ISSUE #2579: Exporting invoices to CSV or JSON provides empty results when invoices that aren't complete are selected
**Describe the bug**
Exporting invoices to CSV or JSON doesn't work. Empty results.

**To Reproduce the bug**
1. Go to "Invoices" in the backoffice
2. Select some invoices or all, doesn't matter
3. Click Export
4. Choose "CSV" or "JSON"
5. You get an empty CSV or an empty JSON

**Expected behavior**
The invoice data should be there for the items I selected.

**Screenshots**
image

image

- BTCPay Server Version:   1.1.2.0
- Deployment Method:  Docker
- Browser:   Chrome latest

woutersamaey:
Fixes #2579

All checked invoices will be exported instead of only the paid ones.

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -408,8 +408,9 @@ namespace BTCPayServer.Controllers
         }
 
         [HttpPost]
-        public async Task<IActionResult> MassAction(string command, string[] selectedItems)
+        public async Task<IActionResult> MassAction(string command, string[] selectedItems, string selectedAll)
         {
+            // TODO can selectedAll be a bool? Tried it but doesn't work. What value should be submitted for the casting to "bool" ? Number "1" does not cast to bool :(
             if (selectedItems != null)
             {
                 switch (command)
@@ -418,6 +419,14 @@ namespace BTCPayServer.Controllers
                         await _InvoiceRepository.MassArchive(selectedItems);
                         TempData[WellKnownTempData.SuccessMessage] = $"{selectedItems.Length} invoice(s) archived.";
 
+                        break;
+                    case "export_json":
+                        // TODO consider selectedAll === '1' or not
+                        // TODO export JSON
+                        break;
+                    case "export_csv":
+                        // TODO consider selectedAll === '1' or not
+                        // TODO export CSV
                         break;
                 }
             }
@@ -752,6 +761,7 @@ namespace BTCPayServer.Controllers
         [BitpayAPIConstraint(false)]
         public async Task<IActionResult> Export(string format, string searchTerm = null, int timezoneOffset = 0)
         {
+            // TODO refactor
             var model = new InvoiceExport(_CurrencyNameTable);
 
             InvoiceQuery invoiceQuery = GetInvoiceQuery(searchTerm, timezoneOffset);
--- a/BTCPayServer/Services/Invoices/Export/InvoiceExport.cs
+++ b/BTCPayServer/Services/Invoices/Export/InvoiceExport.cs
@@ -58,52 +58,92 @@ namespace BTCPayServer.Services.Invoices.Export
             var exportList = new List<ExportInvoiceHolder>();
             var currency = Currencies.GetNumberFormatInfo(invoice.Currency, true);
             var invoiceDue = invoice.Price;
-            // in this first version we are only exporting invoices that were paid
-            foreach (var payment in invoice.GetPayments(true))
-            {
-                var cryptoCode = payment.GetPaymentMethodId().CryptoCode;
-                var pdata = payment.GetCryptoPaymentData();
 
-                var pmethod = invoice.GetPaymentMethod(payment.GetPaymentMethodId());
-                var paidAfterNetworkFees = pdata.GetValue() - payment.NetworkFee;
-                invoiceDue -= paidAfterNetworkFees * pmethod.Rate;
+            var payments = invoice.GetPayments(true);
+            if (payments.Count > 0)
+            {
+                foreach (var payment in payments)
+                {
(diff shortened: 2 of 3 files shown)


woutersamaey:
@Kukks @bolatovumar and @pavlenex should be okay for review now.

woutersamaey:
I totally missed the filter system on the right hand side, so that dropdown button I added has no use. I've taken that commit back.
This PR is now as it should be.

pavlenex:
This PR should export all `checked` invoices, otherwise, it should export the paid/completed ones as it was before.

When I tested this PR it exports all invoices regardless if they were selected or not.

To test, create a few unpaid invoices, go to export.

woutersamaey:
That doesn't sound good. I must've broken something...

woutersamaey:
@pavlenex turns out I didn't break anything. The current export function does not consider what is checked or not.
I agree that the checked items should be POSTed to the controller and then filtered out.
It gets more complicated when you use "select all". In this case a special value should be POSTed so the controller knows to take everything and not just the items that were checked on the current page (you may have pagination and thus some checkboxes are out of view / not checked).

woutersamaey:
Okay so I figured out what needs to happen and am 90% of the way.
Just could use some help with loading the invoices by ID and minor refactoring.

If someone could spare 15-30 mins of their time, I can finalize.

NicolasDorier:
The export button never exported invoices. It exports payments.

NicolasDorier:
We should not merge this IMHO.  The current behavior is expected: This is exporting payments, not invoices. Because this is meant for accounting, and accounting does not care about invoice without payments.

woutersamaey:
Then let's relabel the button to "Export All Payments"?

pavlenex:
@NicolasDorier Problem is, the ability for people to select invoices from the invoice view leads people into thinking they may export invoices.  Screenshot 2021-06-24 at 17 05 54

So either we improve the UI that makes it obvious what's exported or we proceed and allow people to export invoices they select regardless of their status, whereas payments are exported by default.

@woutersamaey is there any use case why someone may want to export an invoice that doesn't have a payment?

woutersamaey:
@pavlenex I guess a use case could be that you want to follow up on expired invoices because you missed out on some business.

pavlenex:
@woutersamaey I'm still not understanding it entirely, can you elaborate on "missed out on some business."

I'm really trying to think of a use-case where a merchant would export invoices from PayPal where the customer didn't have enough funding, that's not relevant for them as they got no payments IMO.

woutersamaey:
@pavlenex I guess you would not export from BTCPay Server but from your online store or other backoffice. If this is your reasoning, then sure, it's uncommon.
I'm not sure about the apps like the POS and Crowdfunding. Maybe someone left his or her email and did not pay. Wouldn't you want to send them an email if you could?
I'm still kinda new to this, so I may be wrong...

Zaxounette:
> @woutersamaey I'm still not understanding it entirely, can you elaborate on "missed out on some business."
>
> I'm really trying to think of a use-case where a merchant would export invoices from PayPal where the customer didn't have enough funding, that's not relevant for them as they got no payments IMO.

I'd translate that by saying it can be the equivalent of analysing "abandoned carts", but maybe in a more cluncky way since we only have non-paid invoices and no other customer data.

woutersamaey:
> we only have non-paid invoices and no other customer data

You do have email addresses in some cases, no?
But yes, abandoned carts is my line of thought.

pavlenex:
Ok, so you're talking about `abandoned orders, which indeed merchants could try to recover and get a payment from a customer. However, I am unsure if `export` is a proper tool to do this. If I was to do this manually, I'd do it from a CMS like Woocommerce, or perhaps use an invoice filter to filter out invoices...

But I guess your point can be valid for certain workflows, now that I'm thinking about it, if I was a merchant using BTCPay solely, I may want to reach out to people that didn't complete payments to see if I can make those abandoned orders into actual orders. To do this, it's quite possible that I may be going to invoice export to get their emails and contact them.

I guess when we designed `Export` we had `accounting` in mind, but there is def a use case to have someone export selected invoices. 

@NicolasDorier What do you say, is it worth refactoring this? If a user export's by default it exports only payment, but if they select invoices manually that want to export, we allow them to do that?

woutersamaey:
Yes, let's keep both features and make them less confusing.
1 Button to export all payments, top right, as a general action.
2 "Export invoices" under the Actions dropdown.

pavlenex:
@woutersamaey I'm not sure I'm in for separating buttons that eventually do the same thing. This would be a good topic to discuss at a design meeting cc @dstrukt @dennisreimann @Zaxounette.

Zaxounette:
Adding it to the list for design meetings.

NicolasDorier:
I am using the export function to analyse how much money I get via pivot table in excel.
This won't be possible with the PR.

I would rather that we rename the button for now to "export payments". And maybe add another one about invoices.

woutersamaey:
I would also like to suggest that we export to XLSX straight away as a 3rd option. Over the years I've had many clients struggle with CSV because under some conditions Excel does not open them nicely. Writing an XSLX is not hard so if we can, it would probably help noobies.

pavlenex:
> There should have two different buttons, this is fundamentally different kind of data for different kind of analysis.

This will be bad UX.

jennybhatt:
I'm so confused someone please call me and tell me how to withdraw funds and how I ended up here on this site and what to do PLEASE PLEASE HELP ME 7053458786 many thanks Jenny.

Spunsd:
Can I pay with an api?

pavlenex:
We can close this one as we've removed invoice export in favor of reporting.

Issue #614, opened 2019-02-26
Add VAT/Tax field to Buyer Information on invoice


nodecheck:
Would it be possible to add to the Buyer Information section for invoices a VAT ID/Tax ID field?  This would be ideal for people wanting to process company transactions where completing such information would be a requisite.

NicolasDorier:
ping @JeffVandrewJr any idea what can be done which would be perfect?

JeffVandrewJr:
Right now BTCPay optionally allows input of a `taxIncluded` field to separate collected tax from revenue. It doesn't currently separate different types of tax, the idea being that you should have separate sales tax software for that. In the USA, we don't have VAT, but for sales tax we have special vendors that specialize in sales tax software. 

If you wanted to drastically expand the POS app functionality, you could theoretically allow the user to input different tax names and rates and have the POS app automatically calculate it. For instance, you could have the user create a tax called "New Jersey Sales Tax" with a rate of 6.625%, and the user could select which items were subject to that tax. The user would need to be responsible for updating any rates over time, as BTCPay can't monitor tax rates in thousands of jurisdictions.

Kukks:
With that amount of data, it would make more sense to add it to `posData` when submitting it

NicolasDorier:
@Kukks yeah I think that is good idea.

nodecheck:
From our perspective in the EU, it's much simpler, since for example, the UK has VAT/Tax of 20% so it means each transaction would have this by default.  For other countries with more complex tax whereby they have differing levels, that's something completely different and difficult to manage obviously.

For me, I want to set the rate to 20%, but I'd like the customer to be able to put in during checkout, their VAT/Tax ID, so that it appears in the invoice.  Of course, their address should be able to appear there also on the invoice.  As at present from the PoS shop this doesn't appear on the generated invoice.

second2falcon:
Is there any update/change on this? Would be nice to see something where we might be able to enter the VAT for specific items (in the Netherlands, depending on the kind of product you pay different tax). Maybe in raw editor for the items in a PoS app.

Zoz92:
Can add tax on the receipt and détails of the card ?

pavlenex:
@Zoz92 Yes, we're working on this for 1.9.0 perhaps.

pavlenex:
@Zoz92 and @second2falcon You guys use our Point of sale app, and would like to have a tax field there as well, or is this about the API and custom integrations?

[...]

PR #2603, opened 2021-06-21
Create invoice now contains all fields

woutersamaey:
- Form in 2 columns
- Contains all fields
- Moved Store selection to the top (so you never overlook it)
- I have some more ideas to make the amount + currency more compact, but need help in #2602 

Screenshot on desktop
Screenshot from 2021-06-21 16-26-47

Screenshot on mobile
Screenshot from 2021-06-21 16-10-24

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -831,6 +831,13 @@ namespace BTCPayServer.Controllers
                     FullNotifications = true,
                     BuyerEmail = model.BuyerEmail,
                     BuyerName =  model.BuyerName,
+                    BuyerPhone =  model.BuyerPhone,
+                    BuyerAddress1 = model.BuyerAddress1,
+                    BuyerAddress2 = model.BuyerAddress2,
+                    BuyerZip = model.BuyerZip,
+                    BuyerCity = model.BuyerCity,
+                    BuyerState = model.BuyerState,
+                    BuyerCountry = model.BuyerCountry,
                     SupportedTransactionCurrencies = model.SupportedTransactionCurrencies?.ToDictionary(s => s, s => new InvoiceSupportedTransactionCurrency()
                     {
                         Enabled = true
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -830,6 +830,7 @@ namespace BTCPayServer.Controllers
                     ItemDesc = model.ItemDesc,
                     FullNotifications = true,
                     BuyerEmail = model.BuyerEmail,
+                    BuyerName =  model.BuyerName,
                     SupportedTransactionCurrencies = model.SupportedTransactionCurrencies?.ToDictionary(s => s, s => new InvoiceSupportedTransactionCurrency()
                     {
                         Enabled = true
--- a/BTCPayServer/Models/InvoicingModels/CreateInvoiceModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/CreateInvoiceModel.cs
@@ -63,6 +63,48 @@ namespace BTCPayServer.Models.InvoicingModels
         {
             get; set;
         }
+        
+        [DisplayName("Buyer Phone")]
+        public string BuyerPhone
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer Address 1")]
+        public string BuyerAddress1
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer Address 2")]
+        public string BuyerAddress2
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer City")]
+        public string BuyerCity
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer State")]
+        public string BuyerState
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer Country")]
+        public string BuyerCountry
+        {
+            get; set;
+        }
+        
+        [DisplayName("Buyer Zip")]
+        public string BuyerZip
+        {
+            get; set;
+        }
 
         [Uri]
         [DisplayName("Notification Url")]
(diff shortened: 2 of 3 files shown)


woutersamaey:
Besides this I also want to improve the "Notification Url" field. Maybe add a tooltip or a better name for it? Is this the URL the customer is redirected to after payment?
I understand this label comes from BitPay, but we can do better I believe.

dstrukt:
I have quite a few thoughts here, and haven't finished submitting my PR for this, and realizing I should probably submit a few smaller PRs addressing some of the basic UI issues.

I don't think we should go with the two column form, there are instances where it might make sense, think amount & currency, city & zip, etc.. but it's historically been found in multiple UX studies that this overwhelming and not preferred by people, and we see this in the most well-known products.
 
Additionally, I think there's more thought that could be put into the form(s), and their structure to make them more digest-able.

So in that regard, NACK.

As for the store at the very top, ACK, been putting that into the Figma mocks as well, and should be the first choice the user makes.

Will follow up on this later today.

woutersamaey:
> I don't think we should go with the two column form, there are instances where two columns of varied width (2/3 and 1/3), think amount & currency, city & zip, etc.. for the economy of space, but it's historically been found in many UX studies that this overwhelming, less intuitive, and generally not preferred by people, and we see this in the most well-known products.

I agree some fields aren't combined optimally, but we can work on that. There are a lot of fields so making them presentable is a challenge.

> Additionally, I think there's more thought that could be put into the form(s), one being simply making the width shorter, and adding proper headers, etc.. organizational and structural to make this page more digest-able.

Making the width shorter is an option.
You end up with forms that look and feel more like modals, but aren't modals in fact.
Maybe this page should be a modal then?

> As for the store at the very top, ACK, been putting that into the Figma mocks as well, and should be the first choice the user makes.

Actually I wanted the correct store to be pre-selected if the previous page was an invoice list for 1 store, but haven't gotten around to that.

> I like the buyer details, etc.. as well, and agree the Notification URL needs more context.
> 
> Which I think is everything else, ACK
> 
> Will follow up on this later today, hopefully with a mock we can discuss.

Great!

dennisreimann:
This form would also benefit from grouping the different areas (e.g. basics, buyer info, additional options). Some of these ideas are [in Figma](https://www.figma.com/file/C7Xyq0FlxgFW8vaBr8ht1z/btcpayserver?node-id=3445%3A104) already.

dstrukt:
> Actually I wanted the correct store to be pre-selected if the previous page was an invoice list for 1 store, but haven't gotten around to that.

This would be an excellent change, and probably one of the best UX changes we could make to this page.

> Making the width shorter is an option.
> You end up with forms that look and feel more like modals, but aren't modals in fact.
> Maybe this page should be a modal then?

The fields will definitely be shortened, but I don't necessarily think it should be a modal.

I think I'm getting sick, so my expectation for posting yesterday were a bit delayed, but still planning to follow up with a mock of this, I think it will hopefully frame and answer a few of these concerns and give us something concrete to discuss.

- 

Wanted to second @dennisreimann's comment, some of the grouping, etc.. is mocked up in Figma, and it would be a good place for us to all converge so we can discuss a mock, critique wording, and decide upon a consistent approach to all of this.

All that said, love that you're thinking about this with us and on-board helping us push this forward!

pavlenex:
Ok, let's put this one on hold until we see how we can make them look identical to @dstrukt's earlier proposed design.

@dstrukt Once you get well, please post some screens so that we know how to proceed further.

NicolasDorier:
@woutersamaey I would like to remove all reference to Buyer Address and all those kind of useless fields.

The new invoice model have a "Metadata" in it, which is an arbitrary JSON object attached to the invoice.
I would rather that you add this field rather than the old, deprecated fields.

woutersamaey:
Sure, but how would you visualize an arbitrary JSON to the backoffice user? You're rendering the JSON to a <table> like structure?

Also, out of curiosity, can the JSON field be queried? Like using the filter on the invoice page?
I.e. Show me all invoices with metadata.customerEmail like '%@somecompany.com' That would be awesome as well.
I don't know what PostgreSQL can handle with regards to JSON.

We would also need to migrate existing data. Move these fields to inside metadata? What if that overrides existing values inside metadata? Might get tricky.

pavlenex:
I wonder if we can add stuff requested in #614?

dennisreimann:
There's also #863 which seems related and could solve this in a more flexible way. Why was it closed, @pavlenex?

pavlenex:
@dennisreimann not closed, converted to a discussion for voting, as it's more of an external plugin, as we won't be expanding our apps (quite. the opposite the plan is to remove them as plugins long-term)

dennisreimann:
The mentioned discussion is #2623.

pavlenex:
@Kukks I brought up in a design meeting, your idea of a "hidden collapsable/flexible" way to display fields "table view?"

Could you possibly elaborate on the approach Nicolas and you agreed on so that we know how to proceed with these additional invoice field issues/pr's?   I assume that having something customizable may be good here. #614 #863 https://github.com/btcpayserver/btcpayserver/discussions/2630

Kukks (changes requested):
@NicolasDorier  and I agreed that for now, we can add these fields in here, on the condition that we wrap all non-required fields in some container that is hidden by default, and a link action that would "Show additional fields". In the future, we can add a more advanced dynamic way that allows you to customize the metadata property in a nice UX without hardcoding a million inputs in this UI.

pavlenex:
@woutersamaey Would you like to modify according to @Kukks suggestions?

bolatovumar:
@woutersamaey I can take this over and implement the suggestion from @Kukks above if you are not working on this anymore.

woutersamaey:
> @woutersamaey I can take this over and implement the suggestion from @Kukks above if you are not working on this anymore.

Yes. Please do. Tnx!

bolatovumar:
> @NicolasDorier and I agreed that for now, we can add these fields in here, on the condition that we wrap all non-required fields in some container that is hidden by default, and a link action that would "Show additional fields". In the future, we can add a more advanced dynamic way that allows you to customize the metadata property in a nice UX without hardcoding a million inputs in this UI.

@Kukks @pavlenex since w don't actually have any required fields for invoices anymore the `wrap all non-required fields in some container` doesn't really apply anymore so what I did is I wrapped the new "buyer details" field in an accordion item which can be expanded. I left the order of other fields more or less the same, just kept them in the two column layout same as in this PR. 

See new PR here: https://github.com/btcpayserver/btcpayserver/pull/3122 (I can't push to this branch because I don't have write access to @woutersamaey's repo so I had to open a new PR)

PR #2606, opened 2021-06-21
Minor UI improvements

woutersamaey:
- Added a checkout button on the invoice detail page (checkout link was only on the list page before)

image

- Added 1 "plus" icon to "Add User" button where it was missing. All other pages have buttons with an icon.

image

- Moved icons to the front of the button instead of the rear, which is never the case on other pages

image

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -129,6 +129,7 @@ namespace BTCPayServer.Controllers
                 PosData = PosDataParser.ParsePosData(invoice.Metadata.PosData),
                 Archived = invoice.Archived,
                 CanRefund = CanRefund(invoice.GetInvoiceState()),
+                ShowCheckout = invoice.Status == InvoiceStatusLegacy.New,
                 Deliveries = (await _InvoiceRepository.GetWebhookDeliveries(invoiceId))
                                     .Select(c => new Models.StoreViewModels.DeliveryViewModel(c))
                                     .ToList()
--- a/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
@@ -116,5 +116,6 @@ namespace BTCPayServer.Models.InvoicingModels
         public List<PaymentEntity> Payments { get; set; }
         public bool Archived { get; set; }
         public bool CanRefund { get; set; }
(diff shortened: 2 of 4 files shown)


dstrukt:
I like that we'd be working towards more consistent buttons.

However, I've been debating if we should be trying to deprecate the icons in our buttons, especially if they're not custom, to both reduce the horizontal sizings, and it's much cleaner without the icon. 

Thoughts? @dennisreimann @Zaxounette

woutersamaey:
Well, reading text is harder then "reading" icons, so having them is a plus.

I realize some designers believe icons are outdated, but they serve a function. I'll suggest to keep them. Less is not more in this case. They communicate visually and effectively. If you were to remove the text and only keep the icon, it'd still communicate.

The only thing I don't like on these screenshots is the grey disabled button. It doesn't look disabled to me. It just looks like a button.

We need a different disabled style.

Zaxounette:
> However, I've been debating if we should be trying to deprecate the icons in our buttons, especially if they're not custom, to both reduce the horizontal sizings, and it's much cleaner without the icon, but I see the value in keeping them too...
> 
> Thoughts? @dennisreimann @Zaxounette

This particular change would have to be justified in more details in a separate discussion. I feel like it would not find consensus.

Zaxounette (approved):
Concept ACK.

dennisreimann (approved):

dennisreimann:
As this also contains functional changes and consolidates the buttons, I'd say we get it merged and discuss the icons in buttons topic separately.

dstrukt:
> The only thing I don't like on these screenshots is the grey disabled button. It doesn't look disabled to me. It just looks like a button.

> We need a different disabled style.

Agreed, and many subtle adjustments to the colors, even the secondary buttons are being made .. they're mocked up, but just not coded yet, should happen within the next week or so.

> Well, reading text is harder then "reading" icons, so having them is a plus.

> I realize some designers believe icons are outdated, but they serve a function. I'll suggest to keep them. Less is not more in this case. They communicate visually and effectively. If you were to remove the text and only keep the icon, it'd still work out.

Yeah I don't disagree, icons definitely have their place, and have gone back and forth utilizing icons in buttons in the Figma mocks, so it's not a hard no, more of a consideration if the action, etc.. can be communicated effectively without them.

Don't mind keeping them in there, and maybe we work towards polishing them up in the future.

My only nitpick on this PR.

pavlenex:
Concept ACK.

PR #2608, opened 2021-06-21
Improves Create Store & Create App views

dstrukt:
I was planning to submit a larger PR that addressed many of the "create" views, and as a result, end up sitting on a larger PR, for the sake of making sure everything is done in one consistent edit ... but I was inspired by @woutersamaey's smaller PRs, and will likely submit a PR that addresses each individual view or a very small subset of views, depending on the complexity.

This PR improves the long row form for both the Create Store & App forms, makes the "back" link consistent with our other links (will make this consistent with the other create views), improves a bit of spacing, and re-orders the App form elements, placing the Store up top.

If smaller PRs aren't the move, LMK, and I will go back to doing a larger PR.

(not sure why tests failed)

Screen Shot 2021-06-21 at 2 47 10 PM

Screen Shot 2021-06-21 at 3 06 38 PM

### DIFF
--- a/BTCPayServer/Views/Apps/CreateApp.cshtml
+++ b/BTCPayServer/Views/Apps/CreateApp.cshtml
@@ -14,7 +14,7 @@
         <h2 class="mb-4">@ViewData["Title"]</h2>
 
         <div class="row">
-            <div class="col-lg-12">
+            <div class="col-lg-6">
                 <form asp-action="CreateApp">
                     <div asp-validation-summary="ModelOnly" class="text-danger"></div>
                     <div class="form-group">
@@ -30,9 +30,9 @@
                         <label asp-for="SelectedStore" class="form-label" data-required></label>
                         <select asp-for="SelectedStore" asp-items="Model.Stores" class="form-select"></select>
                     </div>
-                    <div class="form-group">
+                    <div class="form-group mt-4">
                         <input type="submit" value="Create" class="btn btn-primary" id="Create"/>
-                        <a asp-action="ListApps" class="text-muted ms-3">Back to list</a>
+                        <a asp-action="ListApps" class="btn btn-link px-0 ms-3">Back to list</a>
                     </div>
                 </form>
             </div>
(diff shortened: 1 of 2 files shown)


dennisreimann (approved):
tACK.

Thoughts and suggestions on the Create App form:

- I find it more intuitive to have the name go first
- imho having this on the container works better: `col-sm-8 col-md-6 col-lg-4`
- optional and up to you: Applying the `w-auto` class to the selects lets them take up just the space the content needs (see screenshot).

create

dennisreimann:
Also while we are at it let's also put an `autofocus` on the Name input, so that one can immediately start typing.

dstrukt:
Few thoughts:

1. Hmm... I've felt the opposite, I found picking the store it's associated to / where the money's headed (and I think wouter came to a similar conclusion about the Store up top), the type of app, then naming the app to make the most logical, linear thought process, but I suppose I can see it both ways .. 

2. I added the sizings on the child divs as opposed to the parent container, for the instances where we might have different sizes for sections .. e.g. Payment Requests

image

The top part could be "6", and the description/CSS section could be a "9" on md/lg (I'm just throwing out numbers, but you get the idea), and I didn't feel adding it to the parent container would make the most sense .. and while it might work best here, it wouldn't be consistent with the other views (if that's important)

Thoughts?

3. I thought about the w-auto, and it's good to know I would have used the right class, but I'm not a fan of the variable sizing, and would rather opt to keep them the same size.

ACK:
- autofocus .. should probably be something we're doing across all views, good call!

dennisreimann:
yup, ok, that's fine too and I just wanted to share my thoughts :)

dstrukt:
Glad you did, gives me a few more things to think about, will add 1 of them to the design review, as I think more discussion would be worth worthwhile.

pavlenex (approved):
tACK. Great work towards a much larger goal of improving the view.

Exciting to see @dstrukt coding like a boss 💯

PR #2614, opened 2021-06-23
Delete confirmation modals

dennisreimann:
This is a proof of concept for what's been proposed in #2589 by @dstrukt.

A progressive enhancement for the delete confirmations: If JavaScript is enabled, a modal in the current page context is used instead of a separate confirmation page. Both views use the shared modal code, the modal version gets the relevant data injected via JavaScript.
Depending on the context, the relevant buttons are shown (e.g. the modal has a close button in the top right corner, which isn't shown on the separate confirmation page).

The current state is to gather feedback. If we decide to progress with this, the rest of the `ConfirmModel` cases can also be added in a similar fashion.

## TODO

- [x] Incorporate styles dstrukt proposed [here](https://github.com/btcpayserver/btcpayserver/pull/2614#issuecomment-867116095)
- [x] Add confirmation text input

### DIFF
--- a/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
+++ b/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
@@ -40,7 +40,7 @@
 <div class="list-group">
     @if (Model.Is2faEnabled)
     {
-        <a asp-action="Disable2fa" class="list-group-item d-flex justify-content-between align-items-center list-group-item-action py-3" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-title="Disable two-factor authentication (2FA)" data-description="Disabling 2FA does not change the keys used in authenticator apps. If you wish to change the key used in an authenticator app you should reset your authenticator keys." data-confirm="Disable 2FA">
+        <a asp-action="Disable2fa" class="list-group-item d-flex justify-content-between align-items-center list-group-item-action py-3" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-title="Disable two-factor authentication (2FA)" data-description="Disabling 2FA does not change the keys used in the authenticator apps. If you wish to change the key used in an authenticator app you should reset your authenticator keys." data-confirm="Disable 2FA">
             <div>
                 <h5>Disable 2FA</h5>
                 <p class="mb-0 me-3">Disable two-factor authentication. Re-enabling will not require you to reconfigure your Authenticator app. </p>
@@ -81,5 +81,5 @@
     }
 </div>
 
-<partial name="_Confirm" model="@(new ConfirmModel("2FA", "Your account will no longer have this credential as an option for MFA.", "Remove"))" />
+<partial name="_Confirm" model="@(new ConfirmModel("Remove 2FA credential", "Your account will no longer have the credential as an option for multi-factor authentication.", "Remove"))" />
 
--- a/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
+++ b/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
@@ -81,24 +81,5 @@
     }
 </div>
 
-<div class="modal fade" id="ConfirmModal" tabindex="-1" aria-labelledby="ConfirmTitle" aria-hidden="true">
-    <partial name="ConfirmModal" model="@(new ConfirmModel("2FA", "Your account will no longer have this credential as an option for MFA.", "Remove"))" />
-</div>
+<partial name="_Confirm" model="@(new ConfirmModel("2FA", "Your account will no longer have this credential as an option for MFA.", "Remove"))" />
 
-@section PageFootContent {
-    <script>
-        const modal = document.getElementById('ConfirmModal')
-        modal.addEventListener('show.bs.modal', function (event) {
-            const link = event.relatedTarget
-            const titleEl = document.getElementById('ConfirmTitle')
-            const buttonEl = document.getElementById('ConfirmContinue')
-            const formEl = modal.querySelector('form')
-            const bodyEl = modal.querySelector('.modal-body')
-            const { title, description, confirm } = link.dataset
-            titleEl.textContent = title
-            bodyEl.textContent = description
-            buttonEl.textContent = confirm
-            formEl.setAttribute('action', link.getAttribute('href'))
-        });
-    </script>
-}
--- a/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
+++ b/BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
@@ -3,9 +3,9 @@
     ViewData.SetActivePageAndTitle(ManageNavPages.TwoFactorAuthentication, "Two-factor authentication");
 }
 
-@if(Model.Is2faEnabled)
+@if (Model.Is2faEnabled)
 {
-    if(Model.RecoveryCodesLeft == 0)
+    if (Model.RecoveryCodesLeft == 0)
     {
         <div class="alert alert-danger">
             <h4 class="alert-heading mb-3">
@@ -15,7 +15,7 @@
             <p class="mb-0">You must <a asp-action="GenerateRecoveryCodes" class="alert-link">generate a new set of recovery codes</a> before you can log in with a recovery code.</p>
         </div>
     }
-    else if(Model.RecoveryCodesLeft == 1)
+    else if (Model.RecoveryCodesLeft == 1)
     {
         <div class="alert alert-danger">
             <h4 class="alert-heading mb-3">
@@ -25,7 +25,7 @@
             <p class="mb-0">You can <a asp-action="GenerateRecoveryCodes" class="alert-link">generate a new set of recovery codes</a>.</p>
         </div>
     }
-    else if(Model.RecoveryCodesLeft <= 3)
+    else if (Model.RecoveryCodesLeft <= 3)
     {
         <div class="alert alert-warning">
             <h4 class="alert-heading mb-3">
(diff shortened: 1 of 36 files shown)


pavlenex:
A fantastic improvement to the UX IMO.

This is what we have now
Screenshot 2021-06-24 at 12 19 30

This is what this PR is trying to achieve
Screenshot 2021-06-24 at 12 19 18

Concept ACK on my end to proceed with implementation.

Much smoother UX.

NicolasDorier:
@dennisreimann I think that we should remove the old one. No need of fallback for people who does not have JS.

dennisreimann:
@NicolasDorier Ok, just to clarify, because we are doing progressive enhancement in several places: For the backend we can build JS-only versions of features?

NicolasDorier:
@dennisreimann I think it make sense to have no js callback for things like checkout to increase the reach. We don't want the merchant to drop customer because there is no JS.

But for the backend? I don't think we should care. Even the checkout page, I am unsure it is working correctly without JS.

dennisreimann:
@NicolasDorier Ok, full ACK and that makes our lives on the UI side a little easier!

Kukks:
> @dennisreimann I think that we should remove the old one. No need of fallback for people who does not have JS.
> This is the back end, not really something customer facing.

I disagree with this point. I always try to make things at least degrade but still work for non JS users. But if this PR moves forward I guess I'll open a PR to fix this after.

bolatovumar:
We do have some users who sometimes report bugs they observe when using admin pages with JS disabled so I assume there are people who do this? (use admin pages without JS)

dennisreimann:
Kukks and I decided to leave the non-JS code as a fallback and we'll do classic progressive enhancement here.

In the design meeting we also discussed to do client-side confirmation for delete actions: Like on GitHub the user has to confirm the action by typing/copying the items name into a textfield. This will be done in the model, only client-side and no addition of that in the backend code.

BTCPayServer/Views/UserStores/ListStores.cshtml
```diff
@@ -93,3 +93,21 @@
         </div>
     </div>
 </section>
+
+<div class="modal fade" id="ConfirmModal" tabindex="-1" aria-labelledby="ConfirmTitle" aria-hidden="true">
```
Kukks, line 97: This code seems duplicated on every page with the confirm dialog. Can we have a partial that contains it  all?

Kukks (changes requested):

pavlenex:
tACK.

Only thing I'd like to see improved is the text we use for the modals.
Screenshot 2021-09-01 at 12 22 45

`This app will be removed from this store` isn't helpful. _AppName_ app will be removed from the _StoreName_ store may be a tiny bit more contextual. (with storename and appname being visually different, bold or italic so we don't end up with `store store`.

dennisreimann:
@Kukks @pavlenex I think your feedback can be addressed in unison: What about if we simplify the headline a la "Confirm deletion/change/removal" (whatever matches the use case) and in the text refer to the actual item? 

Right now I just replicated what the separate page version of the confirmation stated, but I think the proposed way will work well with the upcoming "To confirm type ITEM_NAME" input we planned.

If we streamline it, it will also be easier to extract the snippet into a partial. Will proceed in that direction.

dennisreimann:
Refactored the code to use one single partial, this makes it way easier to maintain and integrate new confirmable actions.

I've also done some preparations for the confirmation input, which I'd now like to add in this PR as well.

@pavlenex I think we should also take the time to change the wordings here and there while we are added. You already noted that some of the texts aren't helpful. I also think we aren't consistent: Sometime we speak of deletion, othertimes of removal. 

@Zaxounette Could you maybe help with the wording?

BTCPayServer/Views/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -40,48 +40,46 @@
 <div class="list-group">
     @if (Model.Is2faEnabled)
     {
-        <a asp-action="Disable2faWarning" class="list-group-item d-flex justify-content-between align-items-center list-group-item-action py-3">
+        <a asp-action="Disable2fa" class="list-group-item d-flex justify-content-between align-items-center list-group-item-action py-3" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-title="Disable two-factor authentication (2FA)" data-description="Disabling 2FA does not change the keys used in authenticator apps. If you wish to change the key used in an authenticator app you should reset your authenticator keys." data-confirm="Disable 2FA">
```
Zaxounette, line 43:
```suggestion
        <a asp-action="Disable2fa" class="list-group-item d-flex justify-content-between align-items-center list-group-item-action py-3" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-title="Disable two-factor authentication (2FA)" data-description="Disabling 2FA does not change the keys used in the authenticator apps. If you wish to change the key used in an authenticator app you should reset your authenticator keys." data-confirm="Disable 2FA">
```

BTCPayServer/Views/Server/DynamicDnsServices.cshtml
```diff
@@ -3,51 +3,78 @@
...
             <p>
-                <span>
-                    Dynamic DNS service allows you to have a stable DNS name pointing to your server, even if your IP address change regulary. <br />
-                    This is recommended if you are hosting BTCPayServer at home and wish to have a clearnet HTTPS address to access your server.
-                </span>
+                Dynamic DNS allows you to have a stable DNS name pointing to your server, even if your IP address changes regulary.
+                This is recommended if you are hosting BTCPay Server at home and wish to have a clearnet domain to access your server.
+            </p>
+            <p>
+                Note that you need to properly configure your NAT and BTCPay Server install to get HTTPS certificate.
```
Zaxounette, line 28:
```suggestion
                Note that you need to properly configure your NAT and BTCPay Server installation to get the HTTPS certificate. 
```

BTCPayServer/Views/Server/LndSeedBackup.cshtml
```diff
@@ -30,16 +38,26 @@
...
                     </form>
                 </div>
             }
         </div>
     </div>
 }
 
+@if (!Model.Removed)
+{
+    <partial name="_Confirm" model="@(new ConfirmModel("Delete LND Seed", "Please make sure you made a backup of the seed and password before deleting the LND backup seed from the server!", "Delete"))"/>
```
Zaxounette, line 51:
```suggestion
    <partial name="_Confirm" model="@(new ConfirmModel("Delete LND Seed", "This action will permanently delete your LND seed and password. You will not be able to recover them if you don't have a backup. Are you sure?", "Delete"))"/>
```

BTCPayServer/Views/UserStores/ListStores.cshtml
```diff
@@ -93,3 +93,5 @@
         </div>
     </div>
 </section>
+
+<partial name="_Confirm" model="@(new ConfirmModel("Delete store", "This action is irreversible and will remove all information related to this store. (Invoices, Apps etc...)", "Delete"))" />
```
Zaxounette, line 97:
```suggestion
<partial name="_Confirm" model="@(new ConfirmModel("Delete store", "This action will permanently delete this store and its associated settings, apps, invoices, etc... Are you sure?", "Delete"))" />
```

Zaxounette (changes requested):
> Could you maybe help with the wording?

Aye captain!

> I also think we aren't consistent: Sometime we speak of deletion, other times of removal.

Removal and deletion have two different purposes:
* `Removal/Remove`: We can use this when another action can produce the opposite result without any loss. **Example:** Removing a user from a store <> Adding a user to a store.
* `Deletion/Deleting`: When an action is permanent and other actions can't get whatever you deleted back to it's previous state
  * We can also use `delete permanently` to emphasis that the action is permanent and it has to be thought of, not rushed.

I also added a bunch of "are you sure?" parts to the strings, to emphasis again to the user that he's taking an action that should be thought of, and not done in a hurry.

dennisreimann:
Thanks @Zaxounette, I've incorporated your wording changes. Also included @dstrukt's modal style changes and added the confirmation input, as you can see below:

delete-confim

More lightweight actions like archiving a pull request do not require the confirmation input, so that it does not interrupt the user's flow too much. Only destructive actions require it (in case the user has JavaScript enabled -- this is a purely client side feature).

@Kukks can you re-review? I like how this turned out, thanks to you and @pavlenex for the feedback!

dennisreimann:
The tests work locally, I think they might be flaky.

pavlenex:
@dennisreimann Can we just use `DELETE`
`LND seed` may confuse some people to type in their actual lnd seed.
Plus, we should pay attention to capitalization and use all caps like they do on github.

dennisreimann:
@pavlenex Good point, done!

pavlenex:
@dennisreimann Do we plan to expand this to the delete a user view, doesn't appear that the modal is implemented there?
When deleting a user
Screenshot 2021-09-03 at 17 00 10

dennisreimann:
@pavlenex The delete user case is a bit more [complicated](https://github.com/btcpayserver/btcpayserver/pull/2614/files#L205), because it involves server side checks on the user being an admin and if so, whether or not it's the last admin. I kept it this way, because it would be tricky to move this to the client-side.

pavlenex (approved):
Alright, makes sense. I've played around a bit with this PR and it LGTM.

NicolasDorier:
@dennisreimann it seems to me there can be only a single confirm modal per page? What if on the same page there is two different thing that can be deleted?

dennisreimann:
The modal gets configured via the data attributes of the opening link/button. See for instance the 2FA page, where three different actions are on one page – or the list views.

pavlenex:
@dennisreimann How about for wallet removal? I assume it's the same case as with the user removal?
Would be good to have confirmation there and modal if possible.

Screenshot 2021-09-05 at 13 17 13

Tested once again with and LGTM.

dennisreimann:
Thanks for spotting the inconsistency, will fix that. 

The wallet removal is similar to the user case, it has different conditions depending on whether or not it's a hot wallet. That might be just different texts though, will double check that

Kukks (approved):
Code looks good now :)

pavlenex (approved):
tACK.

Great work, let's get this baby merged. 🚀

PR #2627, opened 2021-06-27
Improve create payment request view

dstrukt:
In my attempt to improve some of the create views, I've moved on Payment Requests.

Few things to note:
- Added helper text under the Email
- Added section headers
- For the future, want to improve the currency / amount based on the dropdown selector we've discussed (@dennisreimann @woutersamaey)
- For the future, want to improve and clean up the calendar / time selector (need to mock up the component), as the input field color (should be white), currently looks disabled.
- Made the back to list, a link element, consistent with the others .. for now, til deprecated.
- Left the "create" heading, until we can move to the header bar.
- For the future, want to make sure the correct store is selected, would be based on the work @woutersamaey is doing on Invoices currently.
- Added the store selector to the top, as it should be the first thing the user considers where the money is going.
- Need to improve the default spacing for the entire app to be 8px, so the spacing between form elements is 24px (3x), not the 21px .. more on the design system / CSS side of things though.

There will be a subsequent PR for this view when more changes and improvements facilitate an improved view, but for now...

Current:
Screen Shot 2021-06-27 at 10 23 41 AM

Based on my mock, minus the obvious differences:
Screen Shot 2021-06-27 at 10 29 12 AM

@dennisreimann On my fork -- feel free to make any additional changes you think might be helpful, or potentially easy based on the notes above.

As always, any feedback, suggestions welcome.

### DIFF
--- a/BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
+++ b/BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
@@ -42,7 +42,7 @@
                         <input type="email" asp-for="Email" class="form-control" />
                         <span asp-validation-for="Email" class="text-danger"></span>
                         <p id="PaymentRequestEmailHelpBlock" class="form-text text-muted">
-                            Recieve updates for this payment request.
+                            Receive updates for this payment request.
                         </p>
                     </div>
                     <h4 class="mt-5 mb-4">Request Details</h4>
--- a/BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
+++ b/BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
@@ -87,7 +87,7 @@
                         <textarea asp-for="Description" class="form-control richtext"></textarea>
                         <span asp-validation-for="Description" class="text-danger"></span>
                     </div>
-                    <h4 class="mt-5 mb-4">Custom Styling</h4>
+                    <h4 class="mt-5 mb-4">Custom Appearance</h4>
                     <div class="form-group">
                         <label asp-for="CustomCSSLink" class="form-label"></label>
                         <a href="https://docs.btcpayserver.org/Theme/#2-bootstrap-themes" target="_blank">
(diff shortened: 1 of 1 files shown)


woutersamaey:
With regard to the calendar / datetime picker, the argument can be made that we should not use such a fancy component. But just rely on the HTML5 native input type datetime-local, as explained here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

Our company uses this a lot and have not encountered any issues with it. Also works great on mobile devices as the OS has its own native keyboard for it.

dstrukt:
Maybe there was a misunderstanding, I'll try to clarify.

Agreed, that's in line with what I'm thinking too, and I like what I'm seeing in the docs.

I'm happy to leverage the three way scroll / html 5 component over our current component, was designing / iterating on what we currently had to make the level of effort minimal - no intentions to be fancy or superfluous, but what we have currently can undoubtedly be improved.

BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
```diff
@@ -18,19 +18,40 @@
...
+                            <input type="text" class="form-control" value="@Model.Stores.Single(item => item.Value == Model.StoreId).Text" readonly />
+                        }
+                        <span asp-validation-for="StoreId" class="text-danger"></span>
+                    </div>
+                    <div class="form-group">
+                        <label asp-for="Email" class="form-label"></label>
+                        <input type="email" asp-for="Email" class="form-control" />
+                        <span asp-validation-for="Email" class="text-danger"></span>
+                        <p id="PaymentRequestEmailHelpBlock" class="form-text text-muted">
+                            Recieve updates for this payment request.
```
britttttk, line 45:
```suggestion
                            Receive updates for this payment request.
```

britttttk: the worst English word

NicolasDorier:
I think that's fine to use the current component if we are already doing it elsewhere anyway.

In a differnt PR later on, we can discuss if we could replace all of them with native one.

BTCPayServer/Views/PaymentRequest/EditPaymentRequest.cshtml
dstrukt: Great catch @britttttk! 

It really is, and despite my best efforts... muscle memory always lets me down.

PR #2628, opened 2021-06-27
Removes lightning setup & wallet setup back button

dstrukt:
I've noticed we have redundant back buttons on both the wallet (on the first page, not subsequent steps) and lightning setup (singular setup page) pages, and also "X" close buttons.

I've made this PR to remove both of those buttons.

If there's a better way to handle this in the code, feel free to make adjustments!

Lightning Network

Before:
Screen Shot 2021-06-27 at 11 36 26 AM

After:
Screen Shot 2021-06-27 at 12 31 56 PM

Wallet

Before:
Screen Shot 2021-06-27 at 12 39 36 PM
After:
Screen Shot 2021-06-27 at 1 18 10 PM

### DIFF
--- a/BTCPayServer/Views/Stores/SetupLightningNode.cshtml
+++ b/BTCPayServer/Views/Stores/SetupLightningNode.cshtml
@@ -4,9 +4,6 @@
     ViewData.SetActivePageAndTitle(StoreNavPages.Index, "Connect to a Lightning node", Context.GetStoreData().StoreName);
 }
 
-@section Navbar {
-}
-
 <header class="text-center">
     <h1>@ViewData["Title"]</h1>
     <div class="d-flex mt-4 mb-5">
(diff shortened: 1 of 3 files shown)


BTCPayServer/Views/Stores/SetupLightningNode.cshtml
```diff
@@ -5,9 +5,6 @@
 }
 
 @section Navbar {
-    <a asp-controller="Stores" asp-action="UpdateStore" asp-route-storeId="@Model.StoreId">
-        <vc:icon symbol="back" />
-    </a>
 }
```
dennisreimann, line 8: Can you remove the entire section here, as in the setup wallet view below?

dennisreimann (approved):

BTCPayServer/Views/Stores/SetupLightningNode.cshtml
dstrukt: Absolutely. 

Removed in latest commit.

dennisreimann:
Awesome. Reviewed and applied it to the modify wallet page as well, as this is the entry for the already set up case. Good to go, let's merge this baby!

PR #2644, opened 2021-06-30
Save paymentRequestId in Metadata when creating invoice for Payment Request

woutersamaey:
Improvements as discussed in #2643 

It bothered me that there is no PaymentRequestId in the invoice. Only an order ID that looks like it, but not very clean.

In the process I had to replace the `BitpayCreateInvoiceRequest` with a `CreateInvoiceRequest` so I could access the `metadata`.

@Kukks for your eyes.

So:
- Added Payment Request ID to the invoice Metadata
-  ~~Added Payment Request ID to invoice webhook payload~~ (removed as per discussion)
- Added Payment Request ID on invoice detail page + link to the Payment Request
- Added a new field `metadata.orderUrl` that will make the order ID clickable on the invoice detail page

image

### DIFF
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -264,18 +264,22 @@ namespace BTCPayServer.Controllers
             {
                 var redirectUrl = _linkGenerator.PaymentRequestLink(id, Request.Scheme, Request.Host, Request.PathBase);
 
+                var invoiceMetadata =
+                    new InvoiceMetadata
+                    {
+                        OrderId = PaymentRequestRepository.GetOrderIdForPaymentRequest(id),
+                        PaymentRequestId = id,
+                        BuyerEmail = result.Email
+                    };
 
-                var invoiceRequest = new CreateInvoiceRequest();
-                invoiceRequest.Metadata = new JObject();
-                invoiceRequest.Metadata["orderId"] = $"{PaymentRequestRepository.GetOrderIdForPaymentRequest(id)}";
-                invoiceRequest.Metadata["paymentRequestId"] = id;
-                invoiceRequest.Metadata["buyerEmail"] = result.Email;
-                invoiceRequest.Currency = blob.Currency;
-                invoiceRequest.Amount = amount.Value;
-                invoiceRequest.Checkout.RedirectURL = redirectUrl;
-                
-                // TODO what is this? This field was used in BitpayCreateInvoiceRequest but does not exist in CreateInvoiceRequest
-                // invoiceRequest.FullNotifications = true;
+                var invoiceRequest =
+                    new CreateInvoiceRequest
+                    {
+                        Metadata = invoiceMetadata.ToJObject(),
+                        Currency = blob.Currency,
+                        Amount = amount.Value,
+                        Checkout = {RedirectURL = redirectUrl}
+                    };
 
                 var additionalTags = new List<string>() {PaymentRequestRepository.GetInternalTag(id)};
                 var newInvoice = await _InvoiceController.CreateInvoiceCoreRaw(invoiceRequest,store, "/",additionalTags, cancellationToken);
--- a/BTCPayServer/Controllers/PaymentRequestController.cs
+++ b/BTCPayServer/Controllers/PaymentRequestController.cs
@@ -8,12 +8,9 @@ using BTCPayServer.Abstractions.Extensions;
 using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
-using BTCPayServer.Events;
 using BTCPayServer.Filters;
-using BTCPayServer.Models;
 using BTCPayServer.Models.PaymentRequestViewModels;
 using BTCPayServer.PaymentRequest;
-using BTCPayServer.Security;
 using BTCPayServer.Services.Invoices;
 using BTCPayServer.Services.PaymentRequests;
 using BTCPayServer.Services.Rates;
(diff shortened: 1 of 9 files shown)


BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -266,25 +263,29 @@ public async Task<IActionResult> ViewPaymentRequest(string id)
...
-                        new List<string>() { PaymentRequestRepository.GetInternalTag(id) },
-                        cancellationToken: cancellationToken))
-                    .Data.Id;
+
+
+                var invoiceRequest = new CreateInvoiceRequest();
+                invoiceRequest.Metadata = new JObject();
+                invoiceRequest.Metadata["orderId"] = $"{PaymentRequestRepository.GetOrderIdForPaymentRequest(id)}";
+                invoiceRequest.Metadata["paymentRequestId"] = id;
+                invoiceRequest.Metadata["buyerEmail"] = result.Email;
```
NicolasDorier, line 272: Please use `InvoiceMetadata`. It's a strongly typed representation of some fields. You can convert easily to/from JObject.
You could add the PaymentRequestId there.

BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -266,25 +263,29 @@ public async Task<IActionResult> ViewPaymentRequest(string id)
...
+                invoiceRequest.Currency = blob.Currency;
+                invoiceRequest.Amount = amount.Value;
+                invoiceRequest.Checkout.RedirectURL = redirectUrl;
+                
+                // TODO what is this? This field was used in BitpayCreateInvoiceRequest but does not exist in CreateInvoiceRequest
+                // invoiceRequest.FullNotifications = true;
```
NicolasDorier, line 278: I don't think we need this because this is only used for Bitpay's style IPN and since we do not set the NotificationUrl in the invoice, it would have been ignored anyway.

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -640,6 +640,11 @@
                                 "type": "string",
                                 "description": "The invoice id of the invoice's event",
                                 "nullable": false
+                            },
+                            "paymentRequestId": {
```
NicolasDorier, line 644: do not add this to the API. They can already get it by querying the invoice and read the metadata.

BTCPayServer/Views/Invoice/Invoice.cshtml
```diff
@@ -68,6 +68,10 @@
                         <th>Order Id</th>
                         <td>@Model.TypedMetadata.OrderId</td>
                     </tr>
+                    <tr>
+                        <th>Payment Request Id</th>
```
NicolasDorier, line 72: I don't think it make sense to just show a cryptic ID.
On top of this, we show already because we are showing the metadata and the paymentRequestId is in the metadata.

NicolasDorier: I propose something else: An `OrderLink` in the metadata. If set, then you put a small icon that the user can click to do to this link next to the order Id.

BTCPayServer/Controllers/PaymentRequestController.cs
woutersamaey: Changed it as requested.

BTCPayServer/Controllers/PaymentRequestController.cs
woutersamaey: Removed it as requested.

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
woutersamaey: Removed it from the API as requested.

BTCPayServer/Views/Invoice/Invoice.cshtml
woutersamaey: I like this suggestion. It's useful for other integrations as well.
Will make the changes + document.

woutersamaey: @NicolasDorier I have 1 issue with the `OrderLink`. In case of a Payment Request, we should not put the full URL here because if BTCPay Server moves to another domain, the link will break. Is it okay if I put in a relative URL or do you have or want a custom URL scheme?

What I mean with a custom schema is that you could use `paymentrequest://ID_COMES_HERE` to dynamically build the URL. For an invoice it could be `invoice://ID_COMES_HERE` etc. This is a flexible solution that allows for dynamic internal linking. There may be better alternatives that use the controller or route name so any route can be experessed too.

I'll also be naming the field `OrderUrl` instead of `OrderLink`. A Link is a `<a href>` while this is a URL and it sounds better, but this is nit-pick.

NicolasDorier: Ah that's a good point. What about: if `orderLink` is not set and `paymentRequestId` is set, you can calculate the orderLink directly from the controller on the fly.

woutersamaey: Yep. Will do this.

NicolasDorier:
@woutersamaey it seems that the woocommerce plugin already use some kind of `orderUrl` field. This is perfect!

image

woutersamaey:
> But since the metadata is `orderUrl`, we can properly make a clickable link for it next to the Order Id.

I made the Order ID a clickable link. No separate button. Was considering doing a separate button, but the other links (like email and store ID) are all regular links so it does not make sense to add a "View Order" button for this 1 thing and not the others.

Screenshot.
image

BTCPayServer/Views/Invoice/Invoice.cshtml
woutersamaey: Should be OK now

woutersamaey:
Cleaning up now...

woutersamaey:
@NicolasDorier should be OK now. Can you review please?

NicolasDorier:
yep sorry! checking now

PR #2646, opened 2021-07-01
Improves create point of sale view

dstrukt:
Adds section headers, re-orders sections, and tightens up some of the form elements.

First step in the direction of trying to clean up these create views. I didn't want to try to cover every possible improvement, if it was even possible, in a single PR.

TODO:
- Add the ability to hide/show certain form elements if checkbox has been checked (will discuss in tomorrow's design meeting) @dennisreimann @woutersamaey will likely need help with this too
- @pavlenex @Zaxounette Would be good to go through the section headers and critique the wording, and even potentially how I've grouped them in this iteration .. still trying to figure out how we want to handle some of this.

Before:
screencapture-localhost-14142-apps-361Xxgx6ZZnhFnRty5SBimMvpMhU-settings-pos-2021-06-30-19_31_05

After:
screencapture-localhost-14142-apps-361Xxgx6ZZnhFnRty5SBimMvpMhU-settings-pos-2021-06-30-20_04_58


dennisreimann:
@dstrukt very good changes, I like where this is going!

I've added examples for how to toggle additional settings in case a checkbox is checked. See f7e11006e99c714e89b4012610cf0564689faa9f

Some thoughts:

- We could leverage the toggle button for the enabling options, like we do in the wallet setup
- Maybe we can move the "Point of Sale Behaviour" and "Custom Appearance" settings into the "Additional Options"?

dennisreimann (commented):
Code-wise this looks good :)

dstrukt:
@dennisreimann Ooo I like that idea of utilizing the toggle switches .. will play around a bit with the Figma mock, as I think they'd look great, but want to verify to make sure.

Yeah I figured we could potentially discuss some wording real quick in today's call, I think it's listed as a line item, as I was a bit torn between really descriptive and broken up sections, or just kinda of lumping more into a few, and having less, but that could def work too! Will play around a bit with the order of things if we opt for the latter.

Glad to hear the code is looking good though!

dstrukt:
@dennisreimann Love the switches, definitely the move.

@Zaxounette I think we should refer some of the labels / helper text on the form elements to refine where possible. Will probably post my suggestions in a following screenshot and will be making them in a few follow up commits.

Screen Shot 2021-07-01 at 1 54 34 PM

dennisreimann:
Updates according to the new mockup and discussion in yesterdays design meeting. 

Let me know if something is unclear or you need help porting those things in #2647.

Zaxounette:
I didn't find the code relating to your proposed changes relating to the [screenshot](https://github.com/btcpayserver/btcpayserver/pull/2646#issuecomment-872544583), so here's a few proposals:

----
**Default View**
Label name proposed change from `Default View` to `Point of Sale Style`
In design slang, they are different views, but for layman, they could be better understood as different "styles" of POS pages, thus this proposal.

----

**This could use some context**
Context proposal: `Choose between three Point of Sale styles: Light (Keypad only), Static (Item list only) or Cart (Item list & cart)`

----

**Discounts**
`Could use an explanation` - A bit of context for you: This actually is self explanatory. When checked, anyone that views the POS page can input a discount in the total due.
* I view this option being used more in a situation where a waiter uses the POS view to enter the customers' order, then inputs the discount (if applicable) and confirms the checkout and only presents the tablet to the customer once the invoice is presented.
* I don't see this option being used in a self-checkout situation. Any customer could input a 99% discount for example.
* In a self-checkout situation, this feature could probably be expanded (or completely revamped) to enable input of discount codes configured in the POS settings and applying discounts based on the codes only, not a customer chosen percentage
  * We could also possibly have the actual functionality and the above proposed one sit as different discount methods that a user could choose between. Just an idea.

----

`Receive updates for this payment request` I expect these are left-overs from other mock-ups.

----

Other than that, nothing else for the time being.

dstrukt:
> I didn't find the code relating to your proposed changes relating to the screenshot, so here's a few proposals:

Not made yet :D 

> Default View
> Label name proposed change from Default View to Point of Sale Style
> In design slang, they are different views, but for layman, they could be better understood as different "styles" of POS pages, thus this proposal.

That was the default, but agreed it needs to be changed ... do you like "Type" more than "Style"?

> This could use some context
> Context proposal: Choose between three Point of Sale styles: Light (Keypad only), Static (Item list only) or Cart (Item list & cart)

Agreed, I like this.

> Discounts
> Could use an explanation - A bit of context for you: This actually is self explanatory. When checked, anyone that views the POS page can input a discount in the total due.

I follow the basic idea, but it's still a bit of a strange checkbox without all of the additional context you've provided, no? Maybe it's just me.

> In a self-checkout situation, this feature could probably be expanded (or completely revamped) to enable input of discount codes configured in the POS settings and applying discounts based on the codes only, not a customer chosen percentage

Agreed, the "checkout code" discount is what I think of when I consider this feature, but it's not clear at all that any of that is possible .. but not may not be solved with a quick wording edit...

Worth bringing up for future edits and improvements on this view / feature imo.

> We could also possibly have the current functionality and the above proposed one sit as different discount methods that a user could choose between. Just an idea.

I like it, worth kicking around to see what's possible.

> Receive updates for this payment request I expect these are left-overs from other mock-ups.

You are correct :)

Zaxounette:
> That was the default, but agreed it needs to be changed ... do you like "Type" more than "Style"?

No preference really, both are more understandable than "view" for layman.
 
> I follow the basic idea, but it's still a bit of a strange checkbox without all of the additional context you've provided, no? Maybe it's just me.

I agree, additional context for this is necessary, but can't input everything I just exposed in that section. A full on doc section just for this wouldn't be worth it.
Maybe just adding `Not recommended for customer self-checkout` could be sufficient, at least with the feature working as it does right now.

dstrukt:
I've address the multiple expanded accordion @dennisreimann.

I've updated the Default View label copy, and added helper text to the discount checkbox per your feedback @Zaxounette.

The only remaining bit I'm considering is the 

> Choose between three Point of Sale styles: Light (Keypad only), Static (Item list only) or Cart (Item list & cart)

It's making me wonder if we should just change the names of the dropdown select items:

Point of Sale Style:
- Keypad Only
- Item list Only
- Item List & Cart

Thoughts?

I think this is the last change before we can merge this one.

dstrukt:
Still a few things I think could be improved, and will likely follow up with another PR in the near future, but this a big improvement from the existing state of the POS creation view.

dennisreimann (approved):
Yes, let's merge the current state and do more iterations in new PRs. This one is good to go and a very solig improvement, very good work, @dstrukt!

Kukks (approved):

PR #2647, opened 2021-07-01
Improves create crowdfund view

dstrukt:
Similar to the Point of Sale improvements, this is an attempt to add section headers, reogranize, and clean up the Crowdfund view.

First step of a few steps in improving this view.

@Zaxounette @pavlenex More wording, section heading copy, etc... to discuss.

Note:
The heading update to the _partial in the POS update PR will update the "Perks" section to make it match the other headers.

Before:
screencapture-localhost-14142-apps-2LyhduaYtVwxfjgS9LyLPsLtp5y5-settings-crowdfund-2021-06-30-21_06_37

After:
screencapture-localhost-14142-apps-2LyhduaYtVwxfjgS9LyLPsLtp5y5-settings-crowdfund-2021-06-30-21_05_05


Zaxounette:
Items that can be discussed upon:
* "Reset goal every" section:
When "never" is selected, it could be nice to have the input box on the right hidden?
* "Allow crowdfund to be publicly visible" checkbox could be placed at the end, right above the save button.
It completely makes sense where you placed it, but could also make sense next to the save button since it could be easily missed in the Crowdfund behavior section.
* I'd disable sounds by default (opinionated)

Other than these nitpicks and opinions, much love <3

dstrukt:
- Agreed on the hidden input @dennisreimann. A change I might need help making.

- The "allow crowdfund to be publicly visible" should be a switch, like we're going to do to the P.O.S... Agree that it should be moved, but I'm thinking it should be much higher, one of the first few items, as it's an important toggle.. and we probably should improve the wording @Zaxounette want to take a stab at it? Similar to all the wording, if you see anything areas for improvement.

- All "enables" should be switches @dennisreimann, can do this here if you demonstrate on the P.O.S example.

- Agree with disabling sounds by default as well.

- Going to keep grouping things and try to improve organization based on feedback from today's design call.

Zaxounette:
> "allow crowdfund to be publicly visible" [...] we probably should improve the wording

1. `Enable Crowdfund`
2. `Activate Crowdfund`
3. `Hide crowdfund page from the public` (default checked)

On the top of my head.

dennisreimann:
Rebased. Similar to #2646 I think this is good to go. So unless @dstrukt has anything necessary to add I'd say we merge this and iterate in another PR.

dennisreimann (approved):

dstrukt:
@dennisreimann and I have a few other minor adjustments to make, but have decided we're going to open a PR in the future to address these.

This one is ready for review.

BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
```diff
@@ -29,7 +29,7 @@ public class UpdateCrowdfundViewModel
         public string NotificationUrl { get; set; }
 
         [Required]
-        [Display(Name = "Allow crowdfund to be publicly visible (still visible to you)")]
+        [Display(Name = "Make Crowdfund Public")]
```
Zaxounette, line 32: I'm okay with this wording, but could we have a descriptive sentence under this checkbox?

Something in the likes of "The crowdfund is only visible to you. To make it visible to anyone else, enable this option."

dennisreimann: Done in a90ad6e. 

helper

dennisreimann:
once-again

Zaxounette (approved):

BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
dstrukt: @Zaxounette It could definitely be improved, and potentially re-ordered, and made more important (as it is basically the switch to flip), but I think the consensus was to follow this PR up, as even this first iteration was such a good step in the right direction. Glad you followed up this though!

Issue #2631, opened 2021-06-28
Create a unique identifier for the invoice user for different use-cases (Tag, Customer ID, Product IT)


pavlenex:
One of our bigger users brought up to our attention that they'd like to have a better way of identifying (grouping) invoices by customer.

> Many customers use BTCPay for payments and frequently we have a need to find in BTCPay all the invoices related to a specific customer. 

> There is a Buyers info section which we could fill in, fill in "Name" and email for example,  and then try to search by buyers info. However, in case I have Users: Joe1 Joe12 Joe123 and search by Joe1, I will receive the info of all listed users

> MB it would be good to have Username field like the OrderID one, that is present in the grid and which could be used for the precise search by this field (mb consider search by full match).

### This has also been brought up by @dstrukt in https://github.com/btcpayserver/btcpayserver/discussions/2630

<div type='discussions-op-text'>

<sup>Originally posted by **dstrukt** June 28, 2021</sup>
One thing I've been considering after seeing all of the new customer fields being added to the "create invoice" view is the notion of adding a customer, so that the "customer" becomes a reusable, re-selectable entity when filling out an invoice.

It's tedious and frustrating if you have to re-enter this information multiple times on an invoice, and this seems like a reasonable means of addressing that problem.

However, there are potentially privacy concerns with this, but I suppose if that information is already being input and stored (?), would enabling this a distinct feature be much different?

There's definitely more to think about here, and what the implementation of this might look like more concretely, but I wanted to throw this out there to get some feedback / get some perspective on something I might not be considering.

Anyway, all feedback and thoughts welcome!</div>

pavlenex:
For this particular use case, I've suggested we create a `Customer ID` field, however, @NicolasDorier pointed out that grouping should also be done by different criteria, such as `Product ID`. Nicolas suggested we call this field `tags` so that it can allow flexibility. Personally, I am not 100% sold on `tags` though it's quite common wording, if `Customer ID` and `Product ID` are the only use-cases perhaps we can have those two fields. The main question is, would `tags` make sense for people?

dstrukt:
AFAIA, those would be the only two use-cases.

Need to think on the "tag" notion though.

pavlenex:
#2648

PR #2648, opened 2021-07-01
Greenfield: add text search terms to an invoice

NicolasDorier:
* Now it is possible to use `additionalSearchTerms` to the invoice creation via the greenfield API.
* It is possible to text search in the invoice list via greenfield API.

Some of our heavy user reported that their support team find it difficult to search for a specific invoice through the list.
By adding some context specific terms to the addtionalSearchTerms at invoice creation time, it is now possible to make search easier.

### DIFF
--- a/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
@@ -7,35 +7,8 @@ using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Client.Models
 {
-    public class CreateInvoiceRequest
+    public class CreateInvoiceRequest : InvoiceDataBase
     {
-        [JsonConverter(typeof(NumericStringJsonConverter))]
-        public decimal Amount { get; set; }
-        public string Currency { get; set; }
-        public JObject Metadata { get; set; }
-        public CheckoutOptions Checkout { get; set; } = new CheckoutOptions();
-
-        public class CheckoutOptions
-        {
-
-            [JsonConverter(typeof(StringEnumConverter))]
-            public SpeedPolicy? SpeedPolicy { get; set; }
-
-            public string[] PaymentMethods { get; set; }
-
-            [JsonConverter(typeof(TimeSpanJsonConverter.Minutes))]
-            [JsonProperty("expirationMinutes")]
-            public TimeSpan? Expiration { get; set; }
-            [JsonConverter(typeof(TimeSpanJsonConverter.Minutes))]
-            [JsonProperty("monitoringMinutes")]
-            public TimeSpan? Monitoring { get; set; }
-
-            public double? PaymentTolerance { get; set; }
-            [JsonProperty("redirectURL")]
-            public string RedirectURL { get; set; }
-
-            public bool? RedirectAutomatically { get; set; }
-            public string DefaultLanguage { get; set; }
-        }
+        public string[] AdditionalSearchTerms { get; set; }
     }
 }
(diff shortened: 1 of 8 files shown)


pavlenex:
#2631

NicolasDorier:
I gave up calling it "tags" because tags would implies other kind of UX concept. (like a way to group by tags all the invoices)
Here we just add the words to the search index.

BTCPayServer.Client/Models/CreateInvoiceRequest.cs
```diff
@@ -7,35 +7,8 @@
...
-            public TimeSpan? Monitoring { get; set; }
-
-            public double? PaymentTolerance { get; set; }
-            [JsonProperty("redirectURL")]
-            public string RedirectURL { get; set; }
-
-            public bool? RedirectAutomatically { get; set; }
-            public string DefaultLanguage { get; set; }
-        }
+        public string[] AdditionalSearchTerms { get; set; }
```
Kukks, line 12: I think we should make this property available also when invoices are fetched. Otherwise, integrations are blind in terms of what they have set it to.

Kukks (changes requested):
Just one comment.

BTCPayServer.Client/Models/CreateInvoiceRequest.cs
NicolasDorier: That's actually the big reason why I called it `addtionalSearchTerms` rather than `tags`.

If we expose this, then it begs the question of why we don't show other terms as well?

Kukks: > If we expose this, then it begs the question of why we don't show other terms as well?
Because they are predefined by existing data. We could just document what the default terms are.

Kukks: I think it makes sense to have both features, but both should be exposed

NicolasDorier: So you mean for each invoice we should expose all the indexed terms? That seem to me quite useless. It's not like any search engine is doing that.

Kukks: > So you mean for each invoice we should expose all the indexed terms? That seem to me quite useless. It's not like any search engine is doing that.
> 
> But if we do it, this is not be limited to AddtionalSearchTerms, but should be to all terms.

No I mean AddtionalSearchTerms is what should be exposed because everything else added to the search index is pre-determined and they know it just by looking at existing data (we can document it). All the indexed data is already available via API
```
foreach (var paymentMethod in invoice.GetPaymentMethods())
                    {
                        if (paymentMethod.Network != null)
                        {
                            var paymentDestination = paymentMethod.GetPaymentMethodDetails().GetPaymentDestination();
                            textSearch.Add(paymentDestination);
                            textSearch.Add(paymentMethod.Calculate().TotalDue.ToString());
                        }
                    }
                    // 
                    textSearch.Add(invoice.Id);
                    textSearch.Add(invoice.InvoiceTime.ToString(CultureInfo.InvariantCulture));
                    textSearch.Add(invoice.Price.ToString(CultureInfo.InvariantCulture));
                    textSearch.Add(invoice.Metadata.OrderId);
                    textSearch.Add(invoice.StoreId);
                    textSearch.Add(invoice.Metadata.BuyerEmail);
                    //
                    textSearch.Add(invoice.RefundMail);
```

NicolasDorier:
As discussed, merging this. If anybody ask requesting it, we can add.
Or maybe just add a real tag system later.

PR #2654, opened 2021-07-02
Add misc/permissions to document the hierarchical structure

NicolasDorier:
This document which permission include which other in the form (I renamed `subPermissions` to `included`)

@Kukks I am thinking, maybe in a later PR, also adding the description. Then getting rid of the human description of each permission in the doc, and rather referencing this operation directly.

image


BTCPayServer.Client/Models/PermissionMetadata.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class PermissionMetadata
+    {
+        static PermissionMetadata()
+        {
+            Dictionary<string, PermissionMetadata> nodes = new Dictionary<string, PermissionMetadata>();
+            foreach (var policy in Client.Policies.AllPolicies)
+            {
+                nodes.Add(policy, new PermissionMetadata() { PermissionName = policy });
+            }
+            foreach (var n in nodes)
```
Kukks, line 18: I think we need a recursive method here instead in case there are sub permissions more than 1 level deep

Kukks (changes requested):
Just one comment around a possible bug

BTCPayServer.Client/Models/PermissionMetadata.cs
NicolasDorier: I don't think we need, because we are looping on all permissions. The API is not hierarchical, it is just a list of the permission included in each permission.

The `Permission.Contains` would properly returns true whatever how deep the hierarchy is.

BTCPayServer/wwwroot/swagger/v1/swagger.template.misc.json
```diff
@@ -1,5 +1,43 @@
...
+                                    "items": {
+                                        "type": "object",
+                                        "properties": {
+                                            "name": {
+                                                "type": "string",
+                                                "description": "The permission id",
+                                                "nullable": false
+                                            },
+                                            "included": {
+                                                "type": "string",
```
bolatovumar, line 27: Wouldn't this be type `"array"`?

BTCPayServer/wwwroot/swagger/v1/swagger.template.misc.json
```diff
@@ -1,5 +1,43 @@
 {
     "paths": {
+        "/misc/permissions": {
+            "get": {
+                "tags": [
+                    "Miscalleneous"
```
bolatovumar, line 6: Typos here, should be `miscellaneous`

BTCPayServer/wwwroot/swagger/v1/swagger.template.json
```diff
@@ -76,7 +76,7 @@
...
... permissions applies to all stores of the user, you can limit to a specific store with the following format: `btcpay.store.cancreateinvoice:6HSHAEU4iYWtjxtyRs9KyPjM9GAQp8kw2T9VWbGG1FnZ`:\n\n* `btcpay.store.canmodifystoresettings`: Modify your stores\n* `btcpay.store.webhooks.canmodifywebhooks`: Modify stores webhooks\n* `btcpay.store.canviewstoresettings`: View your stores\n* `btcpay.store.cancreateinvoice`: Create an invoice\n* `btcpay.store.canviewinvoices`: View invoices\n* `btcpay.store.canmodifypaymentrequests`: Modify your payment requests\n* `btcpay.store.canviewpaymentrequests`: View your payment requests\n* `btcpay.store.canuselightningnode`: Use the lightning nodes associated with your stores\n* `btcpay.store.cancreatelightninginvoice`: Create invoices the lightning nodes associated with your stores\n\nNote that API Keys only limits permission of a user and can never expand it. If an API Key has the permission `btcpay.server.canmodifyserversettings` but that the user account creating this API Key is not administrator, the API Key will not be able to modify the server settings.\nSome permissions may include other permissions, see [this operation](#operation/langCodes).\n",
```
bolatovumar, line 79: `#operation/langCodes` pointing to the wrong ID?

NicolasDorier:
Fixed all

Issue #878, opened 2019-06-03
Use explorer hidden service URL instead of clearnet when relevant


stevenroose:
Self-reminder of @NicolasDorier: https://github.com/btcpayserver/btcpayserver/pull/869#issuecomment-497580440

Situations in which using the hidden service URL would be relevant:
- when the BTCPayServer is running as a hidden service as well
- when use of the Tor Browser is detected

rockstardev:
@pavlenex @woutersamaey close this issue?

woutersamaey:
I can't say. I'm not up to speed on this issue.

PR #2655, opened 2021-07-02
Hide referrer URL to hide our BTCPay Server URL

woutersamaey:
PRIVACY!
Hide our referer URL when linking to blockstream.info for transcation details (or other sites).
This is 1 case I found where clicking the link would tell the external site where our BTCPay Server is.

I need your help:
1) Are there more cases like this?
2) Should I use JavaScript to auto-add "noreferrer noopener" to all `<a href>` with external domains? I vote yes, but open to comments.
3) Are there cases where we use the referrer in a meaningful way?

### DIFF
--- a/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
+++ b/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
@@ -275,7 +275,7 @@
                                                                 <td class="ps-3 text-break">
                                                                     @if (!string.IsNullOrEmpty(payment.Link))
                                                                     {
-                                                                        <a href="@payment.Link" class="text-print-default" target="_blank">@payment.Id</a>
+                                                                        <a href="@payment.Link" class="text-print-default" rel="noreferrer noopener" target="_blank">@payment.Id</a>
                                                                     }
                                                                     else
                                                                     {


pavlenex:
I remember @NicolasDorier suggested something along those lines, not sure if he remembers. The issue is #878

ndeet:
Hi, agree this is a good idea on every outgoing link. I they might have a way to work around the referrer and can still track it if they try hard enough (maybe I'm wrong) but also because of disabling the `Window.opener` access on the linked site it makes total sense to do it on every outgoing link, imo.

Context:
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer

ndeet:
> I remember @NicolasDorier suggested something along those lines, not sure if he remembers. The issue is #878

I think the mentioned issue is unrelated to this. There the discussion is about using TOR instead of clearnet whenever possible. In this issue it is more about if you have to use clearnet anyway you should "harden" the outgoing link and prevent the external site from gathering additional information about where the request came from.

Kukks:
There is probably quite a big number of external links (all the docs for example). I think the JS way will catch them all but it feels a bit hacky, but may be acceptable. I don't think we use the referrer in a meaningful way (@NicolasDorier may know better though).

woutersamaey:
> I think they might have a way to work around the referrer and can still track it if they try hard enough (maybe I'm wrong)

@ndeet there shouldn't be a way around this. If the browser does not send the header, they cannot possibly know.

woutersamaey:
I've used the JS solution in the past to auto-add things like `target="_blank"` + an external site icon to all external links. That works and is dependable.
Only issue is that I don't want to break any feature that relies on the referrer. I mean an external site that would use the referrer back to BTCPay. Does seam highly unlikely.

woutersamaey:
One more question: Does BTCPay Server work without JavaScript enabled. Should we consider paranoid users like this?

Kukks:
> One more question: Does BTCPay Server work without JavaScript enabled? Should we consider paranoid users like this?
> Example: Tor Browser with max security settings enabled.

It works for the most part. I always consider at least some basic way to use any feature without JS enabled, even if it looks ugly (see the coin selection system which degrades into a fugly giant multi-select)

We can add the JS solution as a quickfix and then gradually fix all external links properly (and add a test to detect these, we already have a test that checks for dead links)

dennisreimann:
> Does BTCPay Server work without JavaScript enabled? Should we consider paranoid users like this?
> Example: Tor Browser with max security settings enabled.

We recently talked about the JS-requirement for the backend [here](https://github.com/btcpayserver/btcpayserver/pull/2614#issuecomment-867596281) and decided it would be OK to build backoffice related things without an explicit non-JS fallback.

As for the general approach: I think using `rel="noopener noreferrer"` for external links can be considered best practice and we should do this. I would discourage jumping through hoops and going with a JS-based solution here, the browser should take care of that and using those attributes is the standard way of achieving this.

Kukks:
> > Does BTCPay Server work without JavaScript enabled? Should we consider paranoid users like this?
> > Example: Tor Browser with max security settings enabled.
> 
> We recently talked about the JS-requirement for the backend [here](https://github.com/btcpayserver/btcpayserver/pull/2614#issuecomment-867596281) and decided it would be OK to build backoffice related things without an explicit non-JS fallback.
> 
> As for the general approach: I think using `rel="noopener noreferrer"` for external links can be considered best practice and we should do this. I would discourage jumping through hoops and going with a JS-based solution here, the browser should take care of that and using those attributes is the standard way of achieving this.

It is still hacky though. I'd rather that it is a short-term hack then the "defacto" way to do things. We server render the HTML so it doesn't make sense to start observing the whole UI in JS to patch things after load..

woutersamaey:
This is what I'm gonna do:
- Use JavaScript to scan for external links and form submits (yes this too!) on all pages for missing `rel="noreferrer noopener"`. And if detected, show a warning `<p class="alert alert-danger">` message on the page. I also considered `alert()` or put it in `console.log` but that would either be to intrusive or not intrusive enough.
- Change as many `<a href="">` as I can find. The message on screen will help out here + provide a safety check for the future so we would never forget this is an issue.

NicolasDorier:
We should not over think it.

One idea is to just add a unit test which scan for `<a` and check the `norefrrer` in all views. (I do something like this to check dead links)

But just this fix, is solving 80% of the issue.

PR #2668, opened 2021-07-06
New unit test to scan for external links/forms and if they have rel="noreferrer noopener"

woutersamaey:
@NicolasDorier this is the continuation of previous PR #2655 
- New unit test
- Fixed all templates
- URLs that are inside a variable cannot be identified as internal or external. So, I added "noreferrer noopener" to them anyway as it does no harm. This is like 1-2% of cases. Should not be an issue.
- Renamed the original method `CheckLinks` to `CheckDeadLinks` since this test is also checking links and to avoid confusion.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -206,10 +206,77 @@ namespace BTCPayServer.Tests
             List<Task> checkLinks = new List<Task>();
             foreach (var file in viewFiles)
             {
-                checkLinks.Add(CheckLinks(regex, httpClient, file));
+                checkLinks.Add(CheckDeadLinks(regex, httpClient, file));
             }
 
             await Task.WhenAll(checkLinks);
+        }
+
+        [Fact]
+        [Trait("Fast", "Fast")]
+        public async Task CheckExternalNoReferrerLinks()
+        {
+            var views = Path.Combine(TestUtils.TryGetSolutionDirectoryInfo().FullName, "BTCPayServer", "Views");
+            var viewFiles = Directory.EnumerateFiles(views, "*.cshtml", SearchOption.AllDirectories).ToArray();
+            Assert.NotEmpty(viewFiles);
+
+            foreach (var file in viewFiles)
+            {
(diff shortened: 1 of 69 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -206,12 +206,79 @@ public async Task CheckNoDeadLink()
...
+            foreach (var file in viewFiles)
+            {
+                var html = await File.ReadAllTextAsync(file);
+
+                checkHtmlNodesForReferrer(file, html, "a", "href");
+                checkHtmlNodesForReferrer(file, html, "form", "action");
+            }
+        }
+
+        private void checkHtmlNodesForReferrer(string filePath, string html, string tagName, string attribute)
```
Kukks, line 232: nit: PascalCase
```suggestion
        private void CheckHtmlNodesForReferrer(string filePath, string html, string tagName, string attribute)
```

BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -206,12 +206,79 @@ public async Task CheckNoDeadLink()
...
+                            var relWords = rel.Split(" ");
+                            Assert.Contains("noreferrer", relWords);
+                            Assert.Contains("noopener", relWords);
+                        }
+                    }
+                }
+            }
+        }
+
+        private String getAttributeValue(String nodeHtml, string attribute)
```
Kukks, line 270: nit:PascalCase
```suggestion
        private String GetAttributeValue(String nodeHtml, string attribute)
```

Kukks (approved):

woutersamaey:
Thanks for your approval @Kukks

Kukks (changes requested):
Seems like there may be a malformed HTML tag in this or in Notifications/Index

woutersamaey:
Possibly a bad merge. WIll check.

Kukks:
(you'll also need to update the calls with the new pascal casing)

woutersamaey:
@Kukks should be OK now.

PR #2669, opened 2021-07-06
Implement topup invoices

NicolasDorier:
Implement idea on https://github.com/btcpayserver/btcpayserver/discussions/2659

There is still a bug on top-up invoices with payjoin that I need to solve. I also need to add a test for this.

An invoice with an amount of 0 now has a special meaning: The first payment to it will change its amount to the amount of the payment.

### DIFF
--- a/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
@@ -10,7 +10,7 @@ namespace BTCPayServer.Client.Models
     public class CreateInvoiceRequest
     {
         [JsonConverter(typeof(NumericStringJsonConverter))]
-        public decimal Amount { get; set; }
+        public decimal? Amount { get; set; }
         public string Currency { get; set; }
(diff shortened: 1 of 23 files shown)


Kukks:
> Implement idea on #2659
> 
> There is still a bug on top-up invoices with payjoin that I need to solve. I also need to add a test for this.
> 
> An invoice with an amount of 0 now has a special meaning: The first payment to it will change its amount to the amount of the payment.

Please change it from amount `0` to `null`. There are other use-cases for 0 amount invoices today

Kukks (commented):
I am surprised that this did not break all sorts of things in the system

NicolasDorier:
> Please change it from amount 0 to null. There are other use-cases for 0 amount invoices today

What use case? a 0 invoice is instantly confirmed.

NicolasDorier:
There is a use case for 0 invoice: To implement "free product". So I need to use `-1`, or do it properly with a flag.

NicolasDorier:
supersede by #2730

Issue #917, opened 2019-07-15
Test mode so we can test payment flows without actually paying with BTC or tBTC


woutersamaey:
I am testing BTCPay Server integration with Magento 2 and was wondering if I can put BTCPay Server in "test mode" where I don't actually have to pay with BTC on the payment page.

Some payment providers allow you to put your store in "test mode" and a few buttons will appear on the payment page:
- Simulate full payment
- Simulate partial payment (not common for traditional credit card payments, but might be useful for crypto)
- Simulate payment failed

If you click on "Simulate full payment", the payment is considered OK and you proceed to the next step.
The order steps can be similar, but depends on what the payment page should do.

This is only for integration testing.

Another idea that might be handy is that instead of putting the store in test mode, the payment request itself indicates that it is a test request so in automated testing or end-to-end testing situations, one does not need to keep enabling and disabling test mode all the time.

Or, there could be a "test user" who is always in test mode while the rest is in production mode.

Not sure if this is a feature request or something is already present.

pavlenex:
Hey @woutersamaey are you familiar with testnet and regtest?
Basically, you can run environment variable `export NBITCOIN_NETWORK="testnet"`
It will launch your server on testnet network which will look like [this](https://testnet.demo.btcpayserver.org/).

1. Set testnet network.
2. Get a testnet wallet and get some testnet coins via faucets online.
3. Pay and test
Small info on Testnet can be found [here](https://docs.btcpayserver.org/btcpay-basics/tryitout).

Not completely sure that's what you're looking for.

woutersamaey:
@pavlenex I am familiar, but:
- It requires running a second BTCPay Server instance (test and prod) on another server (can't use Docker for running 2 instances on the same machine)
- Testers will need tBTC. In a larger organisation, testers may not be educated, will run out of tBTC, will need to share passwords, lose access to test wallets, etc. Additional overhead.
- Automated testing is not possible.

Kukks:
How about just using Lightning to pay? creating 1sat invoices on mainnet will not break the bank and are instant anyway :)

woutersamaey:
@Kukks for this we need to have test products set up to cost just 1 sat or 0,01 USD. Also takes time to change the prices, again the wallet problem, etc.

pavlenex:
When you put it in that perspective, I see why you need it, I usually use LN as Kukks said, or just mark invoices as completed manually. Will mark this as a feature request, if you wish to expand more on it feel free to do so.

[...]

PR #2410, opened 2021-03-31
Local Greenfield Client for Plugins

Kukks:
This introduces the GreenField client to the internal codebase, so that we can dogfood the API directly inside btcpay! This also allows plugins to have immediate access to many internal features that are needed.

PR #2672, opened 2021-07-06
Easier payment testing

woutersamaey:
As discussed in #917 we could use an easier way to test payments:
- Non-technical people need to test too
- Large organizations have multiple testers
- Need to test & cover all scenarios like under and over payment, multiple payments, late payments etc...
- No easy end-to-end testing (selenium) possible for merchants

This PR fixes this.

I still need help with a few TODOs in the code. This is my first attempt at dealing with bitcoin payments, so be gentle :D

### DIFF
--- a/BTCPayServer.Client/Models/FakePaymentRequest.cs
+++ b/BTCPayServer.Client/Models/FakePaymentRequest.cs
@@ -1,0 +1,9 @@
+using System;
+
+namespace BTCPayServer.Client.Models
+{
+    public class FakePaymentRequest
+    {
+        public Decimal amount { get; set; }
+    }
+}
--- a/BTCPayServer.Client/Models/FakePaymentResponse.cs
+++ b/BTCPayServer.Client/Models/FakePaymentResponse.cs
@@ -1,0 +1,12 @@
+using System;
+
+namespace BTCPayServer.Client.Models
+{
+    public class FakePaymentResponse
+    {
+        public Decimal amountRemaining { get; set; }
+        public String txid { get; set; }
+        public String successMessage { get; set; }
+        public String errorMessage { get; set; }
+    }
+}
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -32,6 +32,7 @@ using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.EntityFrameworkCore;
 using NBitcoin;
+using NBitcoin.RPC;
 using NBitpayClient;
 using NBXplorer;
 using Newtonsoft.Json.Linq;
@@ -463,6 +464,61 @@ namespace BTCPayServer.Controllers
 
             return View(nameof(Checkout), model);
         }
+        
+        [HttpPost]
+        [Route("i/{invoiceId}/fake-payment")]
+        public async Task<IActionResult> TestPayment(string invoiceId, FakePaymentRequest request)
+        {
+            var invoice = await _InvoiceRepository.GetInvoice(invoiceId);
+            var store = await _StoreRepository.FindStore(invoice.StoreId);
+            
+            // TODO security: only allow this controller to work on REGTEST!
+            
+            var NetworkProvider = new BTCPayNetworkProvider(ChainName.Regtest);
+            
+            // TODO do we need to use GetEnvironment() ?? 
+            //var credentialString = GetEnvironment("TESTS_BTCRPCCONNECTION", "server=http://127.0.0.1:43782;ceiwHEbqWI83:DwubwWsoo3");
+            var credentialString = "server=http://127.0.0.1:43782;ceiwHEbqWI83:DwubwWsoo3";
+
+            
+            
+            // TODO support altcoins, not just bitcoin
+            //var network = invoice.Networks.GetNetwork(invoice.Currency);
+            var network = NetworkProvider.GetNetwork<BTCPayNetwork>("BTC");
+            var ExplorerNode = new RPCClient(RPCCredentialString.Parse(credentialString), network.NBitcoinNetwork);
+            //ExplorerNode.ScanRPCCapabilities();
+            var paymentMethodId = store.GetDefaultPaymentId(_NetworkProvider);
+
+            //var network = NetworkProvider.GetNetwork<BTCPayNetwork>("BTC");
+            var bitcoinAddressString = invoice.GetPaymentMethod(paymentMethodId).GetPaymentMethodDetails().GetPaymentDestination();
+            
+            var bitcoinAddressObj = BitcoinAddress.Create(bitcoinAddressString, network.NBitcoinNetwork);
+
+            // TODO support fiat amounts, not just crypto
+            //var btcAmount = invoice.Price;
+            var btcAmount = request.amount;
+
+            var FakePaymentResponse = new FakePaymentResponse();
+
+            try
+            {
+                FakePaymentResponse.txid = ExplorerNode.SendToAddress(bitcoinAddressObj, new Money(btcAmount, MoneyUnit.BTC)).ToString();
+                FakePaymentResponse.amountRemaining = invoice.Price - btcAmount;
+                FakePaymentResponse.successMessage = "Created transaction " + FakePaymentResponse.txid;
+            }
+            catch (Exception e)
+            {
+                FakePaymentResponse.errorMessage = e.Message;
+                FakePaymentResponse.amountRemaining = invoice.Price;
+            }
+
+            if (FakePaymentResponse.txid != null)
+            {
+                return Ok(FakePaymentResponse);                
+            }
+            return BadRequest(FakePaymentResponse);
+        }
+        
 
         [HttpGet]
         [Route("invoice-noscript")]
--- a/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout.cshtml
@@ -17,6 +17,20 @@
                         <span>{{$t("Open in wallet")}}</span>
                     </a>
                 </div>
+                
+                <!-- TODO only show for REGTEST! -->
+                <form id="fake-payment" action="/i/@Model.InvoiceId/fake-payment" method="post">
+                    <label for="fake-payment-crypto-code" class="form-label">Fake a @Model.CryptoCode payment for testing</label>
+                    <p class="alert alert-danger" style="display: none;"></p>
+                    <p class="alert alert-success" style="display: none;"></p>
+                    <div class="input-group mb-3">
+                      <input id="fake-payment-amount" name="amount" type="number" step="any" class="form-control" placeholder="Amount" value="@Model.BtcDue" aria-label="Amount" aria-describedby="fake-payment-crypto-code">
+                      <span class="input-group-text" id="fake-payment-crypto-code">@Model.CryptoCode</span>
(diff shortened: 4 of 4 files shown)


NicolasDorier:
@woutersamaey why don't you just send money from the store's wallet to the checkout?

woutersamaey:
> @woutersamaey why don't you just send money from the store's wallet to the checkout? There is an API for this in greenfield

I didn't know about this. I read the docs on how to test payments and this is what the docs told me how to test.

NicolasDorier:
This doc is meant for BTCPay Server developer testing, not for testing an external component interaction with btcpay. (though you could still use it)

woutersamaey:
I wanna refactor to use your suggestion, but can you point me in the right direction? Maybe also advise on some of the TODOs? Most are really simple but time costly or hard for me to find.

In `BTCPayServer.Client` there is a Greenfield API client, but it's only used in `BTCPayServer.Tests` so it doesn't look wired in the main app (let alone API key / internal auth). Not sure if this is the right track to the interal Greenfield client.

danielcharrua:
> @woutersamaey why don't you just send money from the store's wallet to the checkout? There is an API for this in greenfield

@NicolasDorier Can you please elaborate more on this? please give an example. Thanks

Kukks:
> I wanna refactor to use your suggestion, but can you point me in the right direction? Maybe also advise on some of the TODOs? Most are really simple, but time costly for me or hard to find on my own.
> 
> In `BTCPayServer.Client` there is a Greenfield API client, but it's only used in `BTCPayServer.Tests` so it doesn't look wired in the main app (let alone API key / internal auth). Not sure if this is the right track to the interal Greenfield client.
> 
> I remember reading about this, but not sure if Greenfield is already wired to use internally or is that still work in progress?

It is a work in progress: #2410

There is also another way to extend checkout UIs without burdening them with new features: UI Extension points.

For example in `BitcoinLikeMethodCheckout.cshtml` there is this line:
`@await Component.InvokeAsync("UiExtensionPoint" ,  new { location="checkout-bitcoin-post-content"})`

How to use:
* Create a new Partial in `Views/Shared` `BitcoinRegtestDevPayment.cshtml` 
* Add some code to the Startup: `services.AddSingleton<IUIExtension>(new UIExtension("BitcoinRegtestDevPayment", "checkout-bitcoin-post-content"));`

That partial will now always be rendered inside the bitcoin payment method checkout body.

Kukks:
> I wanna refactor to use your suggestion, but can you point me in the right direction? Maybe also advise on some of the TODOs? Most are really simple, but time costly for me or hard to find on my own.
> 
> In `BTCPayServer.Client` there is a Greenfield API client, but it's only used in `BTCPayServer.Tests` so it doesn't look wired in the main app (let alone API key / internal auth). Not sure if this is the right track to the interal Greenfield client.
> 
> I remember reading about this, but not sure if Greenfield is already wired to use internally or is that still work in progress?

It is a work in progress: #2410

There is also another way to extend checkout UIs without burdening them with new features: UI Extension points.

For example in `BitcoinLikeMethodCheckout.cshtml` there is this line:
`@await Component.InvokeAsync("UiExtensionPoint" ,  new { location="checkout-bitcoin-post-content"})`

How to use:
* Create a new Partial in `Views/Shared` `BitcoinRegtestDevPayment.cshtml` 
* Add some code to the Startup: `services.AddSingleton<IUIExtension>(new UIExtension("TestExtensionNavExtension", "header-nav"));`


> > @woutersamaey why don't you just send money from the store's wallet to the checkout? There is an API for this in greenfield
> 
> @NicolasDorier Can you please elaborate more on this? please give an example. Thanks

In your integration, when using GreenField API, you can create an onchain transaction using your store wallet that pays the invoice. https://docs.btcpayserver.org/API/Greenfield/v1/#operation/StoreOnChainWallets_CreateOnChainTransaction

woutersamaey:
> > > @woutersamaey why don't you just send money from the store's wallet to the checkout? There is an API for this in greenfield
> > 
> > 
> > @NicolasDorier Can you please elaborate more on this? please give an example. Thanks
> 
> In your integration, when using GreenField API, you can create an onchain transaction using your store wallet that pays the invoice. https://docs.btcpayserver.org/API/Greenfield/v1/#operation/StoreOnChainWallets_CreateOnChainTransaction

I know this call exists, but how to call it from inside BTCPay C# code is the question. Would you just call your own API? What API key should be used then? I don't want to create an API key for internal use.

Maybe I should wait for #2410 to be finished?

woutersamaey:
> That partial will now always be rendered inside the bitcoin payment method checkout body.

How to make it appear on regtest only?

Kukks:
> > That partial will now always be rendered inside the bitcoin payment method checkout body.
> 
> How to make it appear on regtest only?

```
@inject BTCPayNetworkProvider BTCPayNetworkProvider
@if (BTCPayNetworkProvider.NetworkType == ChainName.Regtest)
{
    <button>Make payment</button>
}
```

woutersamaey:
If we use the Greenfield API to use the store wallet to pay our own invoices, would that not look weird in the backoffice? Just checking, haven't tried it yet.

Kukks:
Yes, in the tx list, the balance change will only reflect a negative
balance difference, which would only be the miner fee. But it doesn't
really matter since these are only for testing purposes.

woutersamaey:
> Yes, in the tx list, the balance change will only reflect a negative balance difference, which would only be the miner fee. But it doesn't really matter since these are only for testing purposes.

It would not matter to the dev / technical person, but imagine a larger company where an accountant is involved to give the OK. Would they think of this as a good test? I'd prefer to have everything align 100% with real world scenarios.

Say that we would wanna do this, we can just transfer from another wallet or test store (assuming they have the same crypto enabled)? Or do you see a better option?

Easy solution would be to let the tester select the store wallet as source of the test payment. If they want an exact test, they just create a 2nd store as source of the test payments.

Kukks:
> > Yes, in the tx list, the balance change will only reflect a negative balance difference, which would only be the miner fee. But it doesn't really matter since these are only for testing purposes.
> 
> It would not matter to the dev / technical person, but imagine a larger company where an accountant is involved to give the OK. Would they think of this as a good test? I'd prefer to have everything align 100% with real world scenarios.
> 
> Say that we would wanna do this, we can just transfer from another wallet or test store (assuming they have the same crypto enabled)? Or do you see a better option?
> 
> Easy solution would be to let the tester select the store wallet as source of the test payment. If they want an exact test, they just create a 2nd store as source of the test payments. Does this sound good?

Sure, but now it's getting to the point where it is over-engineered and no longer simple to do. IMO this is a bit much for simple tests. Larger companies would happily cough up the cost for a few onchain txs to be safe

woutersamaey:
Good point! I do believe I can make it easy and presentable. Will think on it a bit and try out a few options.

woutersamaey:
Will also try to add a button to `Expire now` or in 1 or 2 sec to allow the websocket and backend logic to pick up on it. Let's see how this goes.

Since we don't have input groups, I'll also change the HTML to regular form fields so it at least looks presentable.

BTCPayServer.Client/Models/FakePaymentRequest.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace BTCPayServer.Client.Models
+{
+    public class FakePaymentRequest
+    {
+        public Decimal amount { get; set; }
```
bolatovumar, line 7: I believe the convention is to have Title-cased field names (e.g. `Amount`)?

BTCPayServer.Client/Models/FakePaymentResponse.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace BTCPayServer.Client.Models
+{
+    public class FakePaymentResponse
+    {
+        public Decimal amountRemaining { get; set; }
```
bolatovumar, line 7: Same here.

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -463,6 +464,61 @@ public async Task<IActionResult> MassAction(string command, string[] selectedIte
...
         }
+        
+        [HttpPost]
+        [Route("i/{invoiceId}/fake-payment")]
+        public async Task<IActionResult> TestPayment(string invoiceId, FakePaymentRequest request)
+        {
+            var invoice = await _InvoiceRepository.GetInvoice(invoiceId);
+            var store = await _StoreRepository.FindStore(invoice.StoreId);
+            
+            // TODO security: only allow this controller to work on REGTEST!
```
bolatovumar, line 475: Looks like all of these are still in `TODO` state?

bolatovumar (commented):
This functionality would be really nice to have! Easier than having to run a CLI command every time.

I think it would be nice to be able to still create a non-"fake" checkout so that the real UI can be tested without the stuff which won't show up in prod deployments.

BTCPayServer/Controllers/InvoiceController.UI.cs
woutersamaey: Yeah like I said at the top, I need some help. I'm new to the project and C# coding (but have my expertise in other languages of course).

BTCPayServer.Client/Models/FakePaymentRequest.cs
woutersamaey: Sure, no biggie.

BTCPayServer/Controllers/InvoiceController.UI.cs
woutersamaey: I'm putting TODOs for everything in my head so if someone likes to help me, they know where help is needed ;-)

woutersamaey:
@dennisreimann is laid some of the groundwork we discussed, but still need help. If you can help, there are TODOs in the code.
I implemented the fake payment as a 3rd tab like we discussed, but this won't work if we want to keep the payment form visible after payment or after expiry like @Zaxounette suggested we do to simulate late and over payments.

dennisreimann:
Hey @woutersamaey, I started tinkering with this and ended up moving the testing UI out of the tabs. I think this simplifies things quite a bit from the code side and the testing form also stays in place once the invoice is paid or expired. Let me know what you think.

I've also moved the testing-related controller actions into a separate partial controller class. This is analog to the Checkout-Testing partial I've moved the UI to.

testing

NicolasDorier:
1. Please use a MVC filter to remove access to test method instead
2. Add another boolean in the BTCPayEnvironment
3. Add a configuration settings (don't test on regtest)

The reason I want to do this, is that later I'd like to provide a public test instance on signet or testnet for testers so they can make their own tests with selenium IDE or even manual tests.

woutersamaey:
@NicolasDorier just so that I understand correctly: You want to have a config setting to turn this part of the UI incl. the associated controllers on/off?

Just FYI: This is implemented by calling the bitcoin node methods `sendtoaddress`, `getnewaddress` and `generatetoaddress` using NBitcoin, so it works the same as is suggested in the docs. We are NOT using Greenfield to pay from 1 wallet to another or using your own wallet to pay yourself. I discussed this with @dennisreimann and we settled on this to avoid further complexities.

Pros are that the transactions and balance looks natural in the store (might be important for realistic testing).
Cons are we can only do BTC OnChain and probably LN, but no altcoins. Personally, I don't care much for altcoins and am leaving this out of scope on purpose. If someone wants to refactor and do altcoin support at a later date that would be fine of course.

Is this something you can agree on before I put in more time?

woutersamaey:
I would very much like to continue work on this, but indeed let's get @NicolasDorier take...

NicolasDorier:
Coded my suggestion on https://github.com/btcpayserver/btcpayserver/pull/2965

danielcharrua:
Hey guys, this is documented somehow? It would be great to have this on documentation or guides for us to start testing payments. In my case I would like to use it with WooCommerce.

Issue #2639, opened 2021-06-29
Crowdfunding eats lot of CPU in open tab


Kixunil:
**Describe the bug**
When crowdfunding app is opened in browser and the tab is selected, it causes high CPU usage. My system shows 20-50% CPUusage and the fan spins violently. The problem occurs in logged-in session as well as in non-logged in session in at least two quite different deployments.

**To Reproduce the bug**
Steps to reproduce the reported bug:
1. Create a crowdfunding app
2. Open the crowdfunding app
3. Wait a few seconds
4. Check system monitor - see high usage
5. Switch to other tab
6. Wait a bit - see lower usage
7. Switch back (usage will rise again)
8. Switch window or desktop (on Linux)
9. CPU usage is still high
10. Go back, kill the tab
11. CPU usage drops again

**Expected behavior**

CPU usage is not (significantly) higher than usual.

**Your BTCPay Environment (please complete the following information):**

I've tested this in two environments:

Debian

- BTCPay Server Version:    1.1.1
- Deployment Method:        Debian package
- Browser:                  Firefox 78.11esr

Docker

- BTCPay Server Version:    1.1.2
- Deployment Method:        official docker with altcoins
- Browser:                  Firefox 78.11esr

**Logs (if applicable)**

<details>
<summary>Browser console from Debian deployment</summary>

```
UPDATED 
Object { hubPath: Getter & Setter, storeId: Getter & Setter, appId: Getter & Setter, title: Getter & Setter, description: Getter & Setter, mainImageUrl: Getter & Setter, embeddedCSS: Getter & Setter, customCSSLink: Getter & Setter, startDate: Getter & Setter, endDate: Getter & Setter, … }
 
Arguments { 0: {…}, … }
crowdfund-bundle-1.min.js:39:76032
UPDATED 
Object { hubPath: Getter & Setter, storeId: Getter & Setter, appId: Getter & Setter, title: Getter & Setter, description: Getter & Setter, mainImageUrl: Getter & Setter, embeddedCSS: Getter & Setter, customCSSLink: Getter & Setter, startDate: Getter & Setter, endDate: Getter & Setter, … }
 
Arguments { 0: {…}, … }
crowdfund-bundle-1.min.js:39:76032
UPDATED 
[...]

PR #2673, opened 2021-07-06
remove unused code in fireworks.js

bolatovumar:
relates to #2639


BTCPayServer/wwwroot/crowdfund/services/fireworks.js
```diff
@@ -144,7 +138,7 @@ Circle.prototype.draw = function() {
     ctx.globalAlpha = 1;
 }
 
-var animate = anime({
```
bolatovumar, line 147: Removed the variable declaration as it makes it a bit confusing since I expect the variable to be used somewhere but it's not.

PR #2674, opened 2021-07-06
update InvoiceStatusMark Swagger definition

### LINKED ISSUE #2597: Greenfield API: Cannot set an invoice status to `Complete`. Should it be `Settled` instead? Documentation error.
**Describe the bug**
The Greenfield API tells me that I can set an invoice to `Complete`, using the call "Mark Invoice Status".
However, it appears there is no status `Complete` in the C# code. There is `Settled` though. No `Complete`.

The error I get is
```
Error converting value "Complete" to type 'BTCPayServer.Client.Models.InvoiceStatus'. 
```

See docs: https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Invoices_MarkInvoiceStatus

It appears the docs are wrong and `Complete` should instead be `Settled` ?
Setting an invoice as `Invalid` works though, so I believe this is likely a documentation issue.

bolatovumar:
close #2597

The definition for the invoice status was updated in this PR: https://github.com/btcpayserver/btcpayserver/pull/2079/files but the docs have not been updated accordingly. This PR updates the docs and improves formatting.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.Invoices.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Invoices.cs
@@ -90,7 +90,7 @@ namespace BTCPayServer.Client
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
-            if (request.Status!= InvoiceStatus.Settled && request.Status!= InvoiceStatus.Invalid)
+            if (request.Status != InvoiceStatus.Settled && request.Status != InvoiceStatus.Invalid)
                 throw new ArgumentOutOfRangeException(nameof(request.Status), "Status can only be Invalid or Complete");
             var response = await _httpClient.SendAsync(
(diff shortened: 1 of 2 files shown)


BTCPayServer.Client/BTCPayServerClient.Invoices.cs
```diff
@@ -90,7 +90,7 @@ public partial class BTCPayServerClient
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
-            if (request.Status!= InvoiceStatus.Settled && request.Status!= InvoiceStatus.Invalid)
+            if (request.Status != InvoiceStatus.Settled && request.Status != InvoiceStatus.Invalid)
```
bolatovumar, line 93: This is the code that actually checks that we pass either `Settled` or `Invalid` and not something else. I just improved the formatting here.

PR #2679, opened 2021-07-07
Add webhook delivery status indicator

bolatovumar:
As discussed in: https://github.com/btcpayserver/btcpayserver/discussions/2616

Added status indicator for each webhook for a store. Besides just the visual indicator you also get a tooltip on hover with some additional info like when the last delivery occurred and what error message we have if any.

Screen Shot 2021-07-07 at 1 19 30 PM

### DIFF
--- a/BTCPayServer/Views/Stores/Webhooks.cshtml
+++ b/BTCPayServer/Views/Stores/Webhooks.cshtml
@@ -30,17 +30,26 @@
                 <td class="align-baseline">
                     @if (wh.LastDeliverySuccessful)
                     {
-                        <span class="d-flex align-items-center fa fa-check text-success" title="@(wh.LastDeliveryTimeStamp == null ? "No deliveries for this webhook yet" : "Last delivery on: " + @wh.LastDeliveryTimeStamp)"></span>
+                        <span
+                            class="fa fa-check text-success"
+                            id="delivery-status-icon"
+                            data-bs-toggle="tooltip"
+                            title="@(wh.LastDeliveryTimeStamp == null ? "No deliveries for this webhook yet" : "Last delivery " + @wh.LastDeliveryTimeStamp?.ToTimeAgo())"></span>
                     }
                     else
                     {
-                        <span class="d-flex align-items-center fa fa-times text-danger" title="Error: @wh.LastDeliveryErrorMessage. Delivery last attempted on: @wh.LastDeliveryTimeStamp"></span>
+                        <span
+                            class="fa fa-times text-danger"
+                            id="delivery-status-icon"
+                            data-bs-toggle="tooltip"
+                            data-bs-html="true"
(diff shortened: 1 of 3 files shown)


BTCPayServer/Views/Stores/Webhooks.cshtml
```diff
@@ -17,15 +17,26 @@
...
         {
             <tr>
+                <td class="align-baseline">
+                    @if (wh.LastDeliverySuccessful)
+                    {
+                        <span class="d-flex align-items-center fa fa-check text-success" title="@(wh.LastDeliveryTimeStamp == null ? "No deliveries for this webhook yet" : "Last delivery on: " + @wh.LastDeliveryTimeStamp)"></span>
+                    }
+                    else
+                    {
+                        <span class="d-flex align-items-center fa fa-times text-danger" title="Error: @wh.LastDeliveryErrorMessage. Delivery last attempted on: @wh.LastDeliveryTimeStamp"></span>
```
NicolasDorier, line 37: you need to add `.ToBrowserDate()` this will convert the timestamp to datetime of the browser.

NicolasDorier: Argh, you can't because it is in a tooltip... I think we should not show the timestamp then, because the user will be confused. Thinking it is in local time, but it is UTC.

bolatovumar: I can take a look into localizing  the time here. I think it's useful to know when the last delivery occurred but it's not strictly necessary.

bolatovumar: @NicolasDorier it's actually localized already as is:


https://user-images.githubusercontent.com/1934678/124977222-16536100-dfe5-11eb-8501-063c6b8051e5.mov

NicolasDorier: how is it even possible? My guess is that it is the UTC time of the server.

NicolasDorier: @bolatovumar actually there is another idea (if that's the server's date): Rather than put the date, you can put the time spent since the delivery happened.

bolatovumar: @NicolasDorier yeah, it's the timezone of the server. I will look into the "time spent since the delivery happened" part.

bolatovumar: @NicolasDorier OK, updated to use relative time in this commit: https://github.com/btcpayserver/btcpayserver/pull/2679/commits/678b8835e035d83ae70d4d49975aac285515ddb6

Also improved tooltips by using the Bootstrap tooltips:

Screen Shot 2021-07-11 at 12 07 37 PM
Screen Shot 2021-07-11 at 12 07 47 PM

PR #2680, opened 2021-07-07
Auto-select store when creating a new invoice

bolatovumar:
It was annoying to me that when I go to "Stores", then click on "Invoices" for a store I get the invoices for that store but then if I proceed on clicking on "Create an invoice" button on the invoices page, my store is not pre-selected so I can end up creating an invoice for a store that I didn't intend to create an invoice for.

This PR will use the existing store id from the search term to pre-select the right store when user is on the "Create an invoice" page. If store id is not present in the search term, it will work as it did previously (i.e., the first item in the list will be selected).

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -434,8 +434,8 @@ namespace BTCPayServer.Controllers
         [AcceptMediaTypeConstraint("application/bitcoin-paymentrequest", false)]
         [XFrameOptionsAttribute(null)]
         [ReferrerPolicyAttribute("origin")]
-        public async Task<IActionResult> Checkout(string invoiceId, string id = null, string paymentMethodId = null,
-            [FromQuery] string view = null, [FromQuery] string lang = null)
+        public async Task<IActionResult> Checkout(string? invoiceId, string? id = null, string? paymentMethodId = null,
+            [FromQuery] string? view = null, [FromQuery] string? lang = null)
         {
             //Keep compatibility with Bitpay
             invoiceId = invoiceId ?? id;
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -792,9 +793,14 @@ private SelectList GetPaymentMethodsSelectList()
...
         [BitpayAPIConstraint(false)]
-        public async Task<IActionResult> CreateInvoice()
+        public async Task<IActionResult> CreateInvoice(InvoicesModel? model = null)
         {
-            var stores = new SelectList(await _StoreRepository.GetStoresByUserId(GetUserId()), nameof(StoreData.Id), nameof(StoreData.StoreName), null);
+            var stores = new SelectList(
+                await _StoreRepository.GetStoresByUserId(GetUserId()),
+                nameof(StoreData.Id),
+                nameof(StoreData.StoreName),
+                new SearchString(model?.SearchTerm).GetFilterArray("storeid")?.ToArray().FirstOrDefault()
```
bolatovumar, line 802: This is actually the line that makes it tick.

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -433,8 +434,8 @@ public async Task<IActionResult> MassAction(string command, string[] selectedIte
         [AcceptMediaTypeConstraint("application/bitcoin-paymentrequest", false)]
         [XFrameOptionsAttribute(null)]
         [ReferrerPolicyAttribute("origin")]
-        public async Task<IActionResult> Checkout(string invoiceId, string id = null, string paymentMethodId = null,
-            [FromQuery] string view = null, [FromQuery] string lang = null)
+        public async Task<IActionResult> Checkout(string? invoiceId, string? id = null, string? paymentMethodId = null,
```
dennisreimann, line 437: Why does this affect the Checkout actions?

bolatovumar: @dennisreimann it doesn't. I enabled strict null checking by adding `#nullable enable` at the top of the file and C# started complaining about these values being `null` yet their type `string` or whatever. I marked them as nullable because they can be `null`. I could remove these changes, they are not strictly necessary and have no effect other than just making the types more accurate.

dennisreimann: Ok, thanks for the clarification!

dennisreimann (approved):
LGTM :)

dstrukt:
Concept ACK!

Excited to see this live.

satwo:
Tested ACK.

PR #2688, opened 2021-07-09
Add "skip" and "limit" params for onchain txs API endpoint

bolatovumar:
Discussed here: https://github.com/btcpayserver/btcpayserver/discussions/2667

### DIFF
--- a/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.Globalization;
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
```diff
@@ -183,7 +189,7 @@ public async Task<IActionResult> UnReserveOnChainWalletReceiveAddress(string sto
                 filteredFlatList.AddRange(txs.ReplacedTransactions.Transactions);
             }
 
-            var result = filteredFlatList.Select(information =>
+            var result = filteredFlatList.Skip(skip).Take(limit).Select(information =>
```
bolatovumar, line 192: I'm not sure if it's enough to just add `Skip` and `Take` here or if it should also be added to `await _walletRepository.GetWalletTransactionsInfo` above. I didn't do that because I think we can end up in a situation where we may not have the transaction which we need when we get tx details below because the tx list and `filteredFlatList` are not necessarily ordered the same.

NicolasDorier: NBX does not allow filtering, so that's the right place

NicolasDorier:
seems good to me, still draft?

BTCPayServer/Controllers/GreenField/StoreOnChainWalletsController.cs
bolatovumar: @NicolasDorier yeah, I looked into it and couldn't find a way to filter there so did it here.

bolatovumar:
> seems good to me, still draft?

@NicolasDorier if it looks good to you, i will take it out of draft mode.

PR #2700, opened 2021-07-16
Fix: Payment Request status does not update on invoice marked events or when pr amount is changed

### LINKED ISSUE #2689: Payment Requests stay open even after the underpaid invoice is marked settled manually
**Describe the bug**
Payment Requests stay open even after the underpaid invoice is marked settled manually

**To Reproduce the bug**
1. Create a Payment Request for a fiat amount (guess it can be in BTC too)
2. Underpay the invoice
3. Mark the invoice as settled manually even though it was underpaid.
4. Now the Payment Request remains open and believes there is still an amount due

**Expected behavior**
Close the Payment Request along with the manually settled invoice.

**Screenshots**
- Notice the amount of 2500 USD is marked complete
- Yet the Payment Request is still open and asking for 35.20 USD
image

- BTCPay Server Version:    1.1.2.0
- Deployment Method:       docker
- Browser:                Does not matter

Kukks:
fixes #2689

### DIFF
--- a/BTCPayServer/PaymentRequest/PaymentRequestHub.cs
+++ b/BTCPayServer/PaymentRequest/PaymentRequestHub.cs
@@ -150,17 +150,20 @@ namespace BTCPayServer.PaymentRequest
             {
                 foreach (var paymentId in PaymentRequestRepository.GetPaymentIdsFromInternalTags(invoiceEvent.Invoice))
                 {
-                    if (invoiceEvent.Name == InvoiceEvent.ReceivedPayment)
+                    if (invoiceEvent.Name == InvoiceEvent.ReceivedPayment || invoiceEvent.Name == InvoiceEvent.MarkedCompleted  || invoiceEvent.Name == InvoiceEvent.MarkedInvalid)
                     {
                         await _PaymentRequestService.UpdatePaymentRequestStateIfNeeded(paymentId);
-                        var data = invoiceEvent.Payment.GetCryptoPaymentData();
-                        await _HubContext.Clients.Group(paymentId).SendCoreAsync(PaymentRequestHub.PaymentReceived,
-                            new object[]
-                            {
-                            data.GetValue(),
-                            invoiceEvent.Payment.GetCryptoCode(),
-                            invoiceEvent.Payment.GetPaymentMethodId()?.PaymentType?.ToString()
-                            });
+                        var data = invoiceEvent.Payment?.GetCryptoPaymentData();
+                        if (data != null)
+                        {
+                            await _HubContext.Clients.Group(paymentId).SendCoreAsync(PaymentRequestHub.PaymentReceived,
+                                new object[]
+                                {
+                                    data.GetValue(),
+                                    invoiceEvent.Payment.GetCryptoCode(),
+                                    invoiceEvent.Payment.GetPaymentMethodId()?.PaymentType?.ToString()
+                                }, cancellationToken);
+                        }
                     }
 
                     await InfoUpdated(paymentId);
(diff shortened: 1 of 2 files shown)


bolatovumar:
Testing this branch and it doesn't quite work for me (idk, it could be that I'm doing something wrong).

What I'm doing:

1. Created a payment request for 0.2 BTC
2. Paid 0.1 BTC
3. Marked the invoice from step 1 as complete manually
4. Generated a few more blocks (12) just in case
5. Still see the payment request in pending state

Screen Shot 2021-07-18 at 5 05 03 PM

Screen Shot 2021-07-18 at 5 05 26 PM

Kukks:
> Testing this branch and it doesn't quite work for me (idk, it could be that I'm doing something wrong).
> 
> What I'm doing:
> 
> 1. Created a payment request for 0.2 BTC
> 2. Paid 0.1 BTC
> 3. Marked the invoice from step 1 as complete manually
> 4. Generated a few more blocks (12) just in case
> 5. Still see the payment request in pending state
> 
> Screen Shot 2021-07-18 at 5 05 03 PM
> 
> Screen Shot 2021-07-18 at 5 05 26 PM

This is an interesting scenario and seems to be by design: We only count the full invoice amount for a "marked complete" invoice  IF it has no payments attached to it. Perhaps this needs to be changed.

Kukks:
@NicolasDorier We currently have the following logic for handling summing of contributions to an app:
```
                    // If the user get a donation via other mean, he can register an invoice manually for such amount
                    // then mark the invoice as complete
                    var payments = p.GetPayments(true);
                    if (payments.Count == 0 &&
                        p.ExceptionStatus == InvoiceExceptionStatus.Marked &&
                        p.Status == InvoiceStatusLegacy.Complete)
                        return new[] { contribution };

                    contribution.CurrencyValue = 0m;
                    contribution.Value = 0m;

                    // If an invoice has been marked invalid, remove the contribution
                    if (p.ExceptionStatus == InvoiceExceptionStatus.Marked &&
                        p.Status == InvoiceStatusLegacy.Invalid)
                        return new[] { contribution };
```

Does it make sense to switch it to be handled in the same way as marked invalid? 

cc @pavlenex (too much code, didn't read: If an invoice attached to an app is marked complete but has no payment, we add the invoice full amount to the total contributed. If an invoice attached to an app is marked complete but has a payment, we only add the payment amount to the total contributed. If an invoice attached to an app is marked invalid, we count none of the payment amount to the total contributed.)

bolatovumar:
@Kukks it was my understanding that the scenario I tested was the scenario that @woutersamaey expected to work (i.e., if a user underpays the expected amount but payment is manually marked as complete the payment request would be in complete state).

NicolasDorier:
that's quite tricky... @woutersamaey scenario seems good to me.

NicolasDorier:
Another idea: If we had the "external payments" features, then unstucking @woutersamaey could be done in creating an external 
 (fake) payment to the invoice.

Kukks:
> Another idea: If we had the "external payments" features, then unstucking @woutersamaey could be done in creating an external
> (fake) payment to the invoice.
> 
> It would be easier to explain that what @Kukks said, and no edge cases.

ACK. Regardless, this is still bug fix unrelated to the partial payments on marked invoices issue

NicolasDorier:
@Kukks if I understood, what you fixed is: if invoice has no payment and marked complete, it counts in contributions?
If that's the case, then I don't think it should be done. Because external payments would be better.

Kukks:
> @Kukks if I understood, what you fixed is: if invoice has no payment and marked complete, it counts in contributions?
> If that's the case, then I don't think it should be done. Because external payments would be better.

No the fix is for 2 issues:
* If an invoice relating to a payment request has been marked, the payment request status is never updated. (eg a complete invoice is marked invalid, or the other way around)
* If the amount of the payment request was updated, the payment request status is never updated

PR #2702, opened 2021-07-19
Allow camera switching + support multiple QR display modes

Kukks:


BTCPayServer/Views/Shared/CameraScanner.cshtml
```diff
@@ -62,7 +66,7 @@ function initCameraScanningApp(title, onDataSubmit, modalId) {
 		}
   	});
 
-	new Vue({
+    var app = new Vue({
```
woutersamaey, line 69: Use `let` or `const` over `var`!

Kukks:
@dennisreimann If you have some time to give the UX some life after I messed with it, I'd be grateful.

dennisreimann:
I changed it to start with the camera turned off, because otherwise when you go to the wallet send tab it immediately asks for permission.

PR #2703, opened 2021-07-19
Payment Settled Webhook event

### LINKED ISSUE #2640: Greenfield API please allow confirmations after timer expires
Can we get btcpayserver to allow for required confirmations after timer expires runs out via Greenfield api? 

At the moment btcpayserver completely ignores payments after timer expires runs out. Many times, especially when Blockchain is busy, a payment MAY get delayed or detected later or sometimes a client might even send it AFTER timer has run out. It all currently gets ignored with no confirmations.

Seems like only FULLY PAID and OVERPAID gets the selected confirmations needed and only if detected before the timer runs out expires. Anything detected after the timer expires gets no confirmations.

UNDER/PARTIAL also gets none.

Kukks:
resolves #2691
resolves #2640

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceRepository.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceRepository.cs
@@ -810,16 +810,7 @@ namespace BTCPayServer.Services.Invoices
             }
             catch (DbUpdateException) { alreadyExists = true; }
 
-            if (alreadyExists)
-            {
-                return null;
-            }
-
-            if (paymentData.PaymentConfirmed(entity, invoiceEntity.SpeedPolicy))
-            {
-                _eventAggregator.Publish(new InvoiceEvent(invoiceEntity, InvoiceEvent.PaymentSettled) { Payment = entity });
-            }
-            return entity;
+            return alreadyExists ? null : entity;
         }
 
         public async Task UpdatePayments(List<PaymentEntity> payments)
--- a/BTCPayServer/Services/Invoices/InvoiceRepository.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceRepository.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Data.OleDb;
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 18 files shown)


BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -763,75 +763,95 @@ private string NormalizeExceptionStatus(string status)
...
         {
             if (payments.Count == 0)
                 return;
-            using (var context = _ContextFactory.CreateContext())
-            {
-                foreach (var payment in payments)
+            await using var context = _ContextFactory.CreateContext();
+            var paymentsDict = payments
+                .Select(entity => (entity, entity.GetCryptoPaymentData()))
+                .ToDictionary(tuple => tuple.Item2.GetPaymentId());
```
NicolasDorier, line 832: This will crash if there are two payments using the same payment id

Kukks: Can there ever be payments using the same id?

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -763,75 +763,95 @@ private string NormalizeExceptionStatus(string status)
         /// <returns>The PaymentEntity or null if already added</returns>
         public async Task<PaymentEntity> AddPayment(string invoiceId, DateTimeOffset date, CryptoPaymentData paymentData, BTCPayNetworkBase network, bool accounted = false)
```
NicolasDorier, line 764: The repository is meant only to modify the DB.

NOT having logic like sending events.
You should easily be able to refactor this: If the AddPayment failed because the payment is already there, you get null. Else it get inserted. Thus you can do something like 

```csharp
var entity = AddPayment();
if (entity != null)
{
 fireEvents(entity);
}
```

Kukks: I disagree since the Repository is doing so much more than DB actions at this point but ok

NicolasDorier:
Another emperor taste which will greatly simplify things:

Rather than returning `(PaymentEntity paymentEntity, CryptoPaymentData oldData, CryptoPaymentData newData, string invoiceId)`, can you just return `InvoiceEvent`? That avoid cryptic return type and a mapping that can go wrong.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Data.OleDb;
```
NicolasDorier, line 3: image

NicolasDorier, line 3: image

Kukks: removed it, was some stray unused import

Kukks:
Replaced with #2944

Issue #899, opened 2019-06-26
Allow bulk upload of images


thecrypt0hunter:
**Is your feature request related to a problem? No.

> Currently uploading an image to BTCPayServer can only be done one at a time 

**Describe the solution you'd like** 

> It would save a bit of time if I could upload multiple images in one go. 

**Describe alternatives you've considered**

>The current way is just fine, until you have a bunch of files to upload.

woutersamaey:
Where are you uploading these images? Is this for the POS app?
I’d suggest implementing plupload which is a very nice tool for multiple uploads at the same time, has drag and drop, progress bar, etc...

pavlenex:
@woutersamaey No, he's referring to the Server Settings > Files. Eventually, that sort of uploading should be added to apps.

cypherbeerus:
Hi! I'm looking into this issue. Can you clarify why would someone upload files to the server in the first place?

rockstardev:
@cypherbeerus so you could reference them in various scenarios... from POS app for example (https://www.youtube.com/watch?v=Mz4uajz0z9Y).

Or you just want to upload files to reference them externally, outside of BTCPayServer.

cypherbeerus:
Earlier, when a file was uploaded, a confirmation card, a card containing links for this SelectedFileId was rendered.
I've changed it such that when multiple files(>1) are uploaded, files are added but there isn't any confirmation. 
Any thoughts on if/how we can provide confirmation that files were uploaded.

rockstardev:
@cypherbeerus for any additional work I would provide a new issue and link this one. Maybe coordinate with our UX experts @dstrukt & @dennisreimann.

Great job on getting 1st PR merged!

rockstardev:
OK, I see the work is already happening on: https://github.com/btcpayserver/btcpayserver/issues/2707 ... game on!

PR #2705, opened 2021-07-20
Add support for uploading multiple files to server

cypherbeerus:
#899 
Changed file upload object to a list, which allows multiple files to be uploaded at once.

### DIFF
--- a/BTCPayServer/Controllers/ServerController.Storage.cs
+++ b/BTCPayServer/Controllers/ServerController.Storage.cs
@@ -146,27 +146,50 @@ namespace BTCPayServer.Controllers
             public bool IsDownload { get; set; }
         }
 
-
+        
         [HttpPost("server/files/upload")]
-        public async Task<IActionResult> CreateFile(List<IFormFile> files)
+        public async Task<IActionResult> CreateFiles(List<IFormFile> files)
         {
-            if (files.Count > 0)
+            if (files != null && files.Count > 0)
             {
+                int invalidFileNameCount = 0;
                 List<string> fileIds = new List<string>();
                 foreach (IFormFile file in files)
                 {
                     if (!file.FileName.IsValidFileName())
                     {
-                        this.TempData.SetStatusMessageModel(new StatusMessageModel()
-                        {
-                            Message = "Invalid file name",
-                            Severity = StatusMessageModel.StatusSeverity.Error
-                        });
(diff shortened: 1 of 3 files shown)


BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -148,23 +148,42 @@ public enum TmpFileTimeType
...
+                List<string> fileIds = new List<string>();
+                foreach (IFormFile file in files)
                 {
-                    Message = "Invalid file name",
-                    Severity = StatusMessageModel.StatusSeverity.Error
-                });
-                return RedirectToAction(nameof(Files));
+                    if (!file.FileName.IsValidFileName())
+                    {
+                        this.TempData.SetStatusMessageModel(new StatusMessageModel()
```
Kukks, line 160: You should probably do a counter of invalid file names  and set the status message based on the counter result (if counter ==0 show "Files added", else if counter >0 & counter< files.length show "only x files added due to invalid files" else show "files could not be added"

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -148,23 +148,42 @@ public enum TmpFileTimeType
...
+                    }
+                    var newFile = await _FileService.AddFile(file, GetUserId());
+                    fileIds.Add(newFile.Id);
+                }
+                if (fileIds.Count == 1)
+                {
+                    return RedirectToAction(nameof(Files), new
+                    {
+                        statusMessage = "File added!",
+                        fileId = fileIds[0]
```
Kukks, line 175: Perhaps due to this change, we should enhance the `Files` action to accept an array of `fileIds` instead. We can do this in a separate PR.

Kukks (changes requested):
Left a comment around the setting of the status message, and tests need to be fixed.

BTCPayServer/Controllers/ServerController.Storage.cs
cypherbeerus: Yes, this makes sense. I'll create a separate issue later for this.

Kukks (changes requested):
Looks great now, good job on your first contribution!

Kukks (approved):
Oops, my evil twin wanted to reject this.  I'll approve :)

PR #2701, opened 2021-07-19
Add LNbank plugin

dennisreimann:
Based on #2410 and WIP -- most of the general app migration is done, but there are still some errors and things to be discussed.

PR #2716, opened 2021-07-27
Refactoring: Allow ViewsRazor extension to be used by plugins

dennisreimann:
Moves the `ViewsRazor` extension into Abstractions, so that it can be used by plugins.

Separated out of #2701, prerequisite for the LNbank plugin integration.


Kukks (approved):

dennisreimann:
@NicolasDorier What do you think?

PR #2723, opened 2021-07-30
Use PaymentUrlBuilder for ensuring proper formatting of BIP21 addresses

NicolasDorier:
It is part of https://github.com/btcpayserver/btcpayserver/pull/2669

This is essential, as topup invoice doesn't have amount, so the code to properly generate bip21 by just handling strings was getting too complex.


Kukks (approved):
ACK

PR #2724, opened 2021-07-30
Properly clip taxIncluded and invoice's amount

NicolasDorier:
Fix a bug where invoice's amount and taxIncluded may not be properly rounded to the right decimal.
This was previously only enforced for the bitpay API.

It was previously implemented in https://github.com/btcpayserver/btcpayserver/pull/2669


Kukks (approved):
ACK

PR #2725, opened 2021-07-30
Add Baillie Gifford to list of supporters

Zaxounette:


Zaxounette:
Oh, not good?
Was it supposed to be updated somewhere else?

pavlenex:
wanted to do it on a single PR, the readme and the login page since they're in a group of "larger supporters" but I guess I forgot to tell you that ,doing a PR soon

Zaxounette:
Ohhhh yes I forgot about the login page completely ...

PR #2730, opened 2021-08-01
Implement topup invoices

NicolasDorier:
Implement idea on #2659

I bite the bullet and propagate the nullable decimal up to the `InvoiceEntity`. As a result, there is more changes, but those are relatively simple.

I am not fan than now we need to handle the case where price can be null everywhere.

I updated the pay button to support it.

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceRepository.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceRepository.cs
@@ -206,7 +206,8 @@ namespace BTCPayServer.Services.Invoices
 
                 textSearch.Add(invoice.Id);
                 textSearch.Add(invoice.InvoiceTime.ToString(CultureInfo.InvariantCulture));
-                textSearch.Add(invoice.Price.ToString(CultureInfo.InvariantCulture));
+                if (!invoice.IsUnsetTopUp())
+                    textSearch.Add(invoice.Price.ToString(CultureInfo.InvariantCulture));
                 textSearch.Add(invoice.Metadata.OrderId);
                 textSearch.Add(invoice.StoreId);
                 textSearch.Add(invoice.Metadata.BuyerEmail);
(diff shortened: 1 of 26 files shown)


NicolasDorier:
ping @Kukks @dennisreimann

NicolasDorier:
I am really on the fence on the nullable things.

1. This might be a breaking change. If you use a client which expect to always have an amount on an invoice (if calling GetInvoices for example)
2. Now we need to check if the amount is null all the time.

That said, I don't see other good alternative.

Kukks:
> I am really on the fence on the nullable things.
> 
> 1. This might be a breaking change. If you use a client which expect to always have an amount on an invoice (if calling GetInvoices for example)
> 2. Now we need to check if the amount is null all the time.
> 
> That said, I don't see other good alternative.

The alternative is to have 0 amount + isTopup

NicolasDorier:
@Kukks there are advantage to this actually. Since the amount change for the topup, right now we can't know if an invoice was topup when it has been paid.

NicolasDorier:
will implement this
```
	
kukks
7:45 PM



I say we add type to invoice  where value can be Standard and Topup. 
If: 

creating using Bitpay: Type is ALWAYS standard no matter what
creating using Greenfield: You specify type but if amount is set >0. the validation error. if amount is null the type is implicitly toptip
fetching topup invoice using Bitpay: I am inclined to say it should return not found. But we can also just send it back with 0 amount
fetching topup invoice using Greenfield: send it back with 0 amount. The type property will help identify it and the status not being complete will protect the flow for existing integrations
```

NicolasDorier:
@Kukks  ended up following your advice, except I allow the creation of topup via bitpay API.
It turns out it is actually simpler code wise. However, I do not expose the type of invoice in their API.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -425,6 +426,20 @@ public async Task UpdateInvoiceStatus(string invoiceId, InvoiceState invoiceStat
                 await context.SaveChangesAsync().ConfigureAwait(false);
             }
         }
+        internal async Task UpdateInvoicePrice(string invoiceId, InvoiceEntity invoice)
+        {
+            using (var context = _ContextFactory.CreateContext())
+            {
+                var invoiceData = await context.FindAsync<Data.InvoiceData>(invoiceId).ConfigureAwait(false);
+                if (invoiceData == null)
```
Kukks, line 434: We should check if invoice type is topup

Kukks (changes requested):

Issue #2656, opened 2021-07-02
Immature coinbase UTXOs counted towards balance


Kixunil:
**Describe the bug**
If BTCPay is funded directly from `bitcoin-cli -chain=regtest generate 6 ADDRESS_FROM_BTCPAY`, the coinbase UTXOs didn't reach maturity but the UI says balance is 300 BTC. Originally I thought sending is broken but realized that it's actually that message.

**To Reproduce the bug**
Steps to reproduce the reported bug:
1. Setup regtest BTCPayServer instance
2. Setup hot wallet
3. Generate a deposit address for hot wallet
4. run `bitcoin-cli -chain=regtest generate 6 ADDRESS_FROM_BTCPAY`
5. Attempt to spend portion (e.g. 0.0042) of generated sats (300) - will say insufficient amount
6. Generate 100 more blocks to a different address
7. Attempt to spend all coins - will work

**Expected behavior**

The UI either reports balance without immature UTXOs and maybe additionally notes that there are immature UTXOs.

**Screenshots**

![300-btc-is-not-enough](https://user-images.githubusercontent.com/1178779/124269817-aacc4800-db3b-11eb-936e-b360b402f565.png)

**Your BTCPay Environment (please complete the following information):**
- BTCPay Server Version:   1.1.2
- Deployment Method:        Debian package
- Browser:                 Firefox 78.11 ESR

pavlenex:
@NicolasDorier can you check this out, this reminds me a bit of that wallet spending bug we've been trying to track, though I'm not sure, it doesn't look like it.

sageprogrammer:
@NicolasDorier @Kukks I think this has to be fixed from NBxplorer side. When calling GetBalance from nbxplorer client in BTCPay, immature coinbase txn are also counted into confirmed balance, rather than unconfirmed oen.

Kixunil:
@sageprogrammer the thing is immature is not unconfirmed either, it's just immature. So basically there should be three fields: confirmed, immature, unconfirmed.

sageprogrammer:
> @sageprogrammer the thing is immature is not unconfirmed either, it's just immature. So basically there should be three fields: confirmed, immature, unconfirmed.

Correct, The GetBalance function in nbxplorer client will have to be changed first so that it returns confirmed, immature, unconfirmed.
[...]

PR #2731, opened 2021-08-01
Show Immature Balance in walletsend page

sageprogrammer:
**A wallet with Immature txn shows as**
20210801_15h54m16s_grim

**A wallet without any immature txn shows as**
20210801_15h54m35s_grim

I also thought of putting the msg into the tooltip, but decided against that.

Closes: #2656

### DIFF
--- a/BTCPayServer/Views/Wallets/WalletSend.cshtml
+++ b/BTCPayServer/Views/Wallets/WalletSend.cshtml
@@ -84,7 +84,7 @@
                         <button type="button" class="crypto-balance-link btn btn-link p-0 align-baseline">@Model.CurrentBalance</button> <span>@Model.CryptoCode</span>.
                         @if (Model.ImmatureBalance>0)
                         {
-                            <span>@Model.ImmatureBalance @Model.CryptoCode are still immature and require additional confirmations.</span>
+                            <span><br>Note: @Model.ImmatureBalance @Model.CryptoCode are still immature and require additional confirmations.</span>
                         }
                     </p>
                 </div>
@@ -116,7 +116,7 @@
                                             <button type="button" class="crypto-balance-link btn btn-link p-0 align-baseline">@Model.CurrentBalance</button> <span>@Model.CryptoCode</span>.
                                             @if (Model.ImmatureBalance > 0)
                                             {
-                                                <span>@Model.ImmatureBalance @Model.CryptoCode are still immature and require additional confirmations.</span>
+                                                <span><br>Note: @Model.ImmatureBalance @Model.CryptoCode are still immature and require additional confirmations.</span>
                                             }
                                         </p>
                                         <span asp-validation-for="Outputs[index].Amount" class="text-danger"></span>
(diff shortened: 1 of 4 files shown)


BTCPayServer/Views/Wallets/WalletSend.cshtml
```diff
@@ -109,6 +114,10 @@
                                         <p class="form-text text-secondary crypto-info mb-2">
                                             Your current balance is
                                             <button type="button" class="crypto-balance-link btn btn-link p-0 align-baseline">@Model.CurrentBalance</button> <span>@Model.CryptoCode</span>.
+                                            @if (Model.ImmatureBalance > 0)
+                                            {
+                                                <span><br>Note: @Model.ImmatureBalance @Model.CryptoCode are still immature and require additional confirmations.</span>
```
Zaxounette, line 119:
```suggestion
                                                <span><br>Your unconfirmed balance is: @Model.ImmatureBalance @Model.CryptoCode </span>
```

Does `Model.ImmatureBalance` include unconfirmed? Or solely immature?

If i's immature + unconfirmed, we could best use the sentence above, and then add a tooltip.
For a normie/non-tech user, both sentences together make the differentiation between what is spendable and not, clear.

Then we can effectively use a tooltip for a more-info/educational purpose with a sentence in the likes of "Your unconfirmed balance includes transactions that have not yet confirmed on the blockchain, or that are immature" to indicated what is accounted for in this balance.

Zaxounette (commented):
> I also thought of putting the msg into the tooltip, but decided against that.

I'm curious as to what made you decide against?

BTCPayServer/Views/Wallets/WalletSend.cshtml
NicolasDorier: @Zaxounette immature is not unconfirmed. Immature are money that is unspendable received from a coinbase (if you are a miner).

This is a pretty niche case, but people making tests on regtest are confused.

NicolasDorier:
Seems good to me @Zaxounette I think the UX don't have to be perfect here. This is to fix a problem in test environment, when people use `generatetoaddress` to fund a wallet.

In prod this would happen only if a miner is using BTCPay as wallet.

BTCPayServer/Views/Wallets/WalletSend.cshtml
sageprogrammer: > @Zaxounette immature is not unconfirmed. Immature are money that is unspendable received from a coinbase (if you are a miner) because you need to wait 100 conf.
> 
> This is a pretty niche case, but people making tests on regtest are confused.

Pretty much this

NicolasDorier:
will add a small warning sign

pavlenex:
@sageprogrammer killing it 🔥 🔥 🔥

BTCPayServer/Views/Wallets/WalletSend.cshtml
Zaxounette: I know what an immature transaction is.

I was asking if `Model.ImmatureBalance` counts immature alone, or immature + unconfirmed for the balance that is shown on the UI ...

If it's immature alone and the message doesn't appear if none are present, then agreed, niche case and no need to modify anything.
If it counts unconfirmed as well, it absolutely will confuse people.

sageprogrammer: Yes,`Model.ImmatureBalance` only counts the immature txns.

NicolasDorier:
Changed UI a little bit

https://github.com/btcpayserver/btcpayserver/pull/2732

BTCPayServer/Views/Wallets/WalletSend.cshtml
Zaxounette: Thanks =)

PR #2732, opened 2021-08-01
Improve UI of immature balance, show available balance in wallet list

NicolasDorier:
Improved a bit the design, and also made sure we show the available balance in the wallet list rather than the total balance.

@sageprogrammer 

image


sageprogrammer:
Yes, This looks better

pavlenex:
Concept ACK.

PR #2735, opened 2021-08-01
Enhance files actions to accept arrays of fileids

### LINKED ISSUE #2707: Enhance the `Files` action to accept an array of `fileIds` instead.
Perhaps due to this change, we should enhance the `Files` action to accept an array of `fileIds` instead. We can do this in a separate PR.

_Originally posted by @Kukks in https://github.com/btcpayserver/btcpayserver/pull/2705#r673682464_

cypherbeerus:
Closes #2707 
Following the conversation in #2705 and #2707, I've updated the Files action model.
- Changed the SelectedFileId and DirectFileUrl properties in the ViewFilesViewModel from string to a list of strings
- Updated the UI to show a list of cards corresponding to the list of SelectedFileIds
- The changes made are backward compatible, if one passes a string containing a single file Id, it won't break. All methods which calls the Files view like Delete, CreateTempUrl are still working.
- Updated the tests
- Checks whether all files passed exist or not

### DIFF
--- a/BTCPayServer/Controllers/ServerController.Storage.cs
+++ b/BTCPayServer/Controllers/ServerController.Storage.cs
@@ -29,35 +29,21 @@ namespace BTCPayServer.Controllers
 {
     public partial class ServerController
     {
-        [HttpGet("server/files/{fileId?}")]
-        public async Task<IActionResult> Files(string fileId = null, bool multiple = false)
+        [HttpGet("server/files")]
+        public async Task<IActionResult> Files([FromQuery] string[] fileIds = null)
         {
-            if (string.IsNullOrEmpty(fileId))
+            var model = new ViewFilesViewModel()
             {
-                var model = new ViewFilesViewModel()
-                {
-                    Files = await _StoredFileRepository.GetFiles(),
-                    SelectedFileIds = null,
-                    DirectFileUrls = null,
-                    StorageConfigured = (await _SettingsRepository.GetSettingAsync<StorageSettings>()) != null
-                };
-                return View(model);
-            }
-            else
-            {
-                List<string> fileIds;
-                if (multiple)
-                {
-                    fileIds = JsonConvert.DeserializeObject<List<string>>(fileId);
-                }
-                else
-                {
-                    fileIds = new List<string>();
-                    fileIds.Add(fileId);
-                }
+                Files = await _StoredFileRepository.GetFiles(),
+                SelectedFileIds = null,
+                DirectUrlByFiles = null,
+                StorageConfigured = (await _SettingsRepository.GetSettingAsync<StorageSettings>()) != null
+            };
 
-                List<string> fileUrlList = (fileIds == null || fileIds.Count == 0) ? null : new List<string>();
+            if (fileIds != null && fileIds.Length > 0)
+            {
                 bool allFilesExist = true;
+                Dictionary<string, string> directUrlByFiles = new Dictionary<string, string>();
                 foreach (string filename in fileIds)
                 {
                     string fileUrl = await _FileService.GetFileUrl(Request.GetAbsoluteRootUri(), filename);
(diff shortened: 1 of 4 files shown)


cypherbeerus:
@kukks @rockstardev 
Build failing on test "EnsureNewLightningInvoiceOnPartialPayment". 
Unable to figure out the reason.

rockstardev:
@Kukks `EnsureNewLightningInvoiceOnPartialPayment` test has failed twice in a row before finally passing:

https://app.circleci.com/pipelines/github/btcpayserver/btcpayserver?branch=pull%2F2735

image

NicolasDorier:
Should we merge this for this release @rockstardev ?

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -22,25 +22,75 @@
...
+                        {
+                            Files = await _StoredFileRepository.GetFiles(),
+                            SelectedFileIds = null,
+                            DirectFileUrls = null,
+                            StorageConfigured = (await _SettingsRepository.GetSettingAsync<StorageSettings>()) != null
+                        }
+                    );
+                }
+
+                var model = new ViewFilesViewModel()
```
Kukks, line 85: We cna probably get away with having this one instantiation of the view model for the entre method with some fine-tuning of the above logic (better to avoid 3 different ways of creating the same VM that you want to return)

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -190,18 +240,11 @@ public async Task<IActionResult> CreateFiles(List<IFormFile> files)
...
-                        fileId = fileIds[0]
-                    });
-                }
-                else
-                {
-                    return RedirectToAction(nameof(Files));
-                }
+                return RedirectToAction(nameof(Files), new
+                { 
+                    fileId = JsonConvert.SerializeObject(fileIds),
```
Kukks, line 245: As mentioned above, make fileId a string array and just pass the fileIds to it instead of jsonifying it into a string and needing a multiple flag.

BTCPayServer/Models/ServerViewModels/ViewFilesViewModel.cs
```diff
@@ -6,8 +6,8 @@ namespace BTCPayServer.Models.ServerViewModels
     public class ViewFilesViewModel
     {
         public List<StoredFile> Files { get; set; }
-        public string DirectFileUrl { get; set; }
-        public string SelectedFileId { get; set; }
+        public List<string> DirectFileUrls { get; set; }
```
Kukks, line 9: Personally I would prefer a dedicated property that holds both of these items. Perhaps a `Dictionary<string,string> directFileUrlForSelectedFiles`  where the key is the selected file id and the value is the direct file url

Kukks (changes requested):
Suggested some changes for simplification of the logic

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -22,25 +22,75 @@
...
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Controllers
 {
     public partial class ServerController
     {
         [HttpGet("server/files/{fileId?}")]
-        public async Task<IActionResult> Files(string fileId = null)
+        public async Task<IActionResult> Files(string fileId = null, bool multiple = false)
```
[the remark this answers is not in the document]
cypherbeerus: fileId is an optional route parameter. How can that be a list?

dennisreimann:
Not directly related to this PR, but we should clean up the UI once this is merged. Also the UX can be improved as having the file details separately below the files table is rather unintuitive.

BTCPayServer/Controllers/ServerController.Storage.cs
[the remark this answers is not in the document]
cypherbeerus: Hi! I tried sorting this out. 
fileId is a route parameter. I'm not sure if there's a way other than serializing the list, or converting fileId to a query parameter.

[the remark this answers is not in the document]
cypherbeerus: Are we fine with converting fileId to a query parameter?
The API calls will now look like
GET /server/files?fileId=f1&fileId=f2

BTCPayServer/Models/ServerViewModels/ViewFilesViewModel.cs
cypherbeerus: Will update this. Agreed that the mapping would be better.

BTCPayServer/Controllers/ServerController.Storage.cs
```diff
@@ -22,25 +22,75 @@
...
+                    if (fileUrl == null)
+                    {
+                        allFilesExist = false;
+                        break;
+                    }
+                    fileUrlList.Add(fileUrl);
+                }
+
+                if (!allFilesExist)
+                {
```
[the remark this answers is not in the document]
cypherbeerus: Done

BTCPayServer/Controllers/ServerController.Storage.cs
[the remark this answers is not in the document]
cypherbeerus: Added a new commit.

dstrukt:
Second @dennisreimann's call, but great work, excited to see this functionality get worked into btcpay!

BTCPayServer/Models/ServerViewModels/ViewFilesViewModel.cs
```diff
@@ -6,8 +6,8 @@ namespace BTCPayServer.Models.ServerViewModels
     public class ViewFilesViewModel
     {
         public List<StoredFile> Files { get; set; }
-        public string DirectFileUrl { get; set; }
-        public string SelectedFileId { get; set; }
+        public Dictionary<string, string> DirectUrlByFiles { get; set; }
+        public string[] SelectedFileIds { get; set; }
```
Kukks, line 10: This is not needed as `DirectUrlByFiles.Keys` should yield the same list

Kukks (changes requested):

BTCPayServer/Models/ServerViewModels/ViewFilesViewModel.cs
cypherbeerus: done

pavlenex:
tACK.

PR #2327, opened 2021-03-04
Fix Ethereum issue and add easy sweep feature

xpayserver:
hi i fix #2186 and add beta sweep feature 
it group all fund token and eth by wallet and attempt compute most economical path for sweep. when not enough eth it ask send transfer to address in sweep list

PR #2739, opened 2021-08-03
Big integer

ImTiaan:
This is an attempted fix of #2327 with add beta sweep feature


Kukks:
Do you intend to finish this? We will remove all support otherwise.

PR #2748, opened 2021-08-03
Improve Payment Request view

### LINKED ISSUE #2747: Payment requests should show only paid addresses
Proposed by @NicolasDorier in #2734

Payment requests doesn't show the address which received the Bitcoins.
When I pay somebody with a payment request, I can't print a receipt with the address written on it.

This is problematic. If you pay 10 persons in a single transaction and print the payment request page, then the accountant is at loss to reconciliate the outputs of the transaction to the person it actually paid.

Right now it provides way too much information for an average user, perhaps we can just show the basics, what's really been paid and show the rest on expansion. IMO Payment history should contain payments, not failed payments, especially if we consider this a page that people could print out and hold as an invoice of some sort or a proof for the accountant.

dennisreimann:
Removes irrelevant invoices (invalid or without any payments) from the payment history and added the destination to the payment data.

Also changes the payment history to have one table per invoice, whereas before all invoices were displayed in one table. That made it hard to read, because of the various column labels. Now the invoices are cleanly separated and also have a margin, which makes it easier to distinguish which invoice a payment belongs to. 

Closes #2747.

payment-history

### DIFF
--- a/BTCPayServer/PaymentRequest/PaymentRequestService.cs
+++ b/BTCPayServer/PaymentRequest/PaymentRequestService.cs
@@ -136,7 +136,8 @@ namespace BTCPayServer.PaymentRequest
                         .Where(payment => payment != null)
                         .ToList();
 
-                    if (state.Status == InvoiceStatusLegacy.Invalid || !payments.Any())
+                    if (state.Status == InvoiceStatusLegacy.Invalid ||
+                        state.Status == InvoiceStatusLegacy.Expired && !payments.Any())
                         return null;
                     
                     return new ViewPaymentRequestViewModel.PaymentRequestInvoice
--- a/BTCPayServer/PaymentRequest/PaymentRequestService.cs
+++ b/BTCPayServer/PaymentRequest/PaymentRequestService.cs
@@ -102,6 +102,43 @@ namespace BTCPayServer.PaymentRequest
                 Invoices = invoices.Select(entity =>
                 {
                     var state = entity.GetInvoiceState();
+                    var payments = entity
+                        .GetPayments(true)
+                        .Select(paymentEntity =>
+                        {
+                            var paymentData = paymentEntity.GetCryptoPaymentData();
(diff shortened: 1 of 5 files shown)


BTCPayServer/PaymentRequest/PaymentRequestService.cs
```diff
@@ -102,6 +102,43 @@ public async Task<ViewPaymentRequestViewModel> GetPaymentRequest(string id, stri
...
+                                Link = link,
+                                Id = txId,
+                                Destination = paymentData.GetDestination()
+                            };
+                        })
+                        .Where(payment => payment != null)
+                        .ToList();
+
+                    if (state.Status == InvoiceStatusLegacy.Invalid || !payments.Any())
+                        return null;
```
dennisreimann, line 140: Hides the noise/irrelevant information.

dstrukt:
Love that you broke them into their respective tables within the card, definitely the original vision, but wasn't aware they weren't separate.

🚀

NicolasDorier:
tests fail

BTCPayServer/PaymentRequest/PaymentRequestService.cs
```diff
@@ -102,6 +102,44 @@ public async Task<ViewPaymentRequestViewModel> GetPaymentRequest(string id, stri
...
+                                Id = txId,
+                                Destination = paymentData.GetDestination()
+                            };
+                        })
+                        .Where(payment => payment != null)
+                        .ToList();
+
+                    if (state.Status == InvoiceStatusLegacy.Invalid ||
+                        state.Status == InvoiceStatusLegacy.Expired && !payments.Any())
+                        return null;
```
dennisreimann, line 141: This hides unnecessary information from the payment history.

dennisreimann:
@NicolasDorier Code and hence fixed -- it was an actual error in the condition I added, so good the tests catched that!

PR #2749, opened 2021-08-04
Show output descriptor in wallet settings

NicolasDorier:
Remove code duplication for creating the DerivationSchemeSettings from the parser and expose output descriptor in wallet settings.

image

I am on the fence to merge this. The UX sucks ([I explain here](https://github.com/bitcoin/bitcoin/issues/17190#issuecomment-892319777) why there isn't only one output descriptor but two)

Maybe users should just not copy/paste descriptors in the first place, this is confusing.

### DIFF
--- a/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
+++ b/BTCPayServer.Tests/AltcoinTests/AltcoinTests.cs
@@ -997,21 +997,21 @@ normal:
             //standard legacy hd wallet
             var parsedDescriptor = mainnetParser.ParseOutputDescriptor(
                 "pkh([d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/0/*)");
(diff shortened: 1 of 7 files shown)


NicolasDorier:
@Kukks what do you think? This seems it will confuse average users, so I'm not sure it is a good idea to expose it like this.
Maybe hide it in an "advanced information" part or something.

Kukks:
Let's wait for next release...this new ux is def super confusing :(

NicolasDorier:
noting that for later, for marginal improvement idea:
image

dennisreimann:
Kool, I'll work on the whole wallet section including settings in the next release. Looking forward to incorporate the output descriptors there :)

NicolasDorier:
ok Let's shelf this PR until the wallet setting stuff is done then.

dennisreimann:
@NicolasDorier my idea would be to do it the other way around. However, this'll be good :)

dstrukt:
@Kukks The current UX (with the modify wallet & wallet "settings") will be cleaned up in the coming week or so, there were a few items from the wallet > settings view that required a bit more refactoring / consideration when building out the wallet setup ... but completely agreed with your sentiment.

NicolasDorier:
Closing this PR for now. @pavlenex is stressed when there is sleeping PRs. Will revisit this after the wallet settings redesign.

PR #2759, opened 2021-08-09
Add example for orderId query string param in /api/v1/stores/{storeId}/invoices

### LINKED ISSUE #2736: Greenfield API: Get invoices with Order ID filter as array does not work.
**Describe the bug**
The docs tell me that I can load invoices by orderId as an array. This does not work. It works if `orderId` is a string, but not an array.

image

**To Reproduce the bug**
1. Just run the API call with `?orderId[]=1000&orderId[]=1001&orderId[]=1002` assuming 1000,1001 and 1002 are order IDs you have in `metadata`. You will get ALL invoices returned instead of a filtered set.
2. Calling the API with just `?orderId=1000` as a string (not array) and you will get the invoices with this order ID (but only 1 order ID can be given).

**Expected behavior**
- Either doc needs to be fixed or API call should work.
- The call should return an error if a querystring parameter is not understood. Like in my case I'm trying to filter using multiple order IDs, but I get all invoices. An error like "invalid filter parameter" would make sense instead of giving me all invoices which I was certainly not expecting.
- Maybe more and other filter options would be nice, but that's not the issue here.

- BTCPay Server Version:   Latest
- Deployment Method:      Docker
- Browser:              N/A

bolatovumar:
close #2736 

Adds an example for `orderId` query string param in ` /api/v1/stores/{storeId}/invoices` to make it clear how it's supposed to be used. There are different conventions for using arrays in query string parameter values so those not familiar with .NET conventions may get confused without an example provided.

Screen Shot 2021-08-08 at 8 23 19 PM

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
@@ -26,7 +26,8 @@
                             "items": {
                                 "type": "string"
                             }
-                        }
+                        },
+                        "example": "1000&orderId=1001&orderId=1002"
                     },
                     {
                         "name": "status",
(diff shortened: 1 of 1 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -983,7 +984,7 @@
                         "allOf": [ {"$ref": "#/components/schemas/TimeSpanMinutes"}]
                     },
                     "monitoringMinutes": {
-                        "type": "integer",
+                        "type": "number",
```
bolatovumar, line 987: This is not related to the issue at hand but I saw a warning in the console for this and corrected it:

Screen Shot 2021-08-08 at 8 14 57 PM

PR #2773, opened 2021-08-13
Match Lightning payment based on payment hash if BOLT11 is not the same.

Kukks:


Kukks:
@NicolasDorier  This is required for LNURL support

PR #2777, opened 2021-08-16
Support multi currency contributions

Kukks:
Previously, if crowdfund was set to count all invoices on the store towards the goal, and invoices were in different currencies, they would not be counted unless the invoice currency matched the crowdfund currency. This fixes that UNLESS the store rate source cannot find an exchange rate for between the invoice current and the crowdfund currency

### DIFF
--- a/BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
@@ -2,7 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using BTCPayServer.Payments;
+using BTCPayServer.Rating;
 using BTCPayServer.Services.Rates;
 
 namespace BTCPayServer.Models.AppViewModels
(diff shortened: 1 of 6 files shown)


pavlenex:
Tested on an affected instance, and it works.

BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
```diff
@@ -58,14 +61,47 @@ public class Contribution
...
-            public Contributions(IEnumerable<KeyValuePair<PaymentMethodId, Contribution>> collection) : base(collection)
+            public Dictionary<string, Dictionary<PaymentMethodId, Contribution>> Collection { get; }
+
+            
+            public Contributions(Dictionary<string, Dictionary<PaymentMethodId, Contribution>> collection)
+            {
+                Collection = collection;
+            }
+
+            public async Task<decimal> GetTotalCurrency(string targetCurrency, RateRules rateRules, RateFetcher rateFetcher)
```
NicolasDorier, line 76: I don't feel good about it.

1. This will crash if the `rateRules` doesn't cover the conversion.
2. This will make the rate from the invoice different from the rate when we evaluate the contribution.

NicolasDorier:
I prefer we do not merge this, this will create more headache and corner case, for something that is almost never used.
See: https://github.com/btcpayserver/btcpayserver/pull/2777/files#r698203558

BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs
Kukks: 1. Does not crash as `FetchRates` catches and provides a "not available" response
2. If the invoice uses the same currency code as the app, it will use the same rate and not use this method.

pavlenex:
Ok, so we're closing this one, however if there are reports from peope in the future, we can re-open and re-visit.

Kukks also told me he things this is very edge-case and hacky, closing for now and if there are reports at least we can tell people how to fix it.

PR #2794, opened 2021-08-20
Refactor themes

dennisreimann:
In [design call #11](https://github.com/btcpayserver/organization/issues/40#issuecomment-863577103) we decided to deprecate the Classic and Casa theme, as well as to make the theme switch a per user setting.

97371e0 is a cleanup in which the customizability of the Bootstrap and Creative Start CSS files gets removed. I think we can assume that no one is using this feature -- otherwise we would have heard of this, because `creative.css` is based on Bootstrap 4 and the recent migration would have broken stuff for users in that case.
Making these files a part of our bundle simplifies the maintenance and the customizability options are fairly limited anyways, as one would have to generate a custom Bootstrap version. We should discourage that and rather guide people towards the customizability via theme variable files, which will make maintenance easier for us and the users.

c04eb5b removes the Casa and Classic theme files. They are [still available via the design system](https://design.btcpayserver.org/resources/css/#old-themes) and should be added as custom theme CSS file, like any other customized version.

The theme switch becomes a user facing setting, like we do on the website, docs and blog. This way users can choose if they prefer light or dark. If the admin sets a custom theme, the light/dark theme switch could be hidden.

A bunch of ideas, open for discussion :) /cc @dstrukt


pavlenex:
Concept ACK, I think removing those two was long-overdue and also ACK on user-preference theme selection.

How would this affect users using the casa theme and updating? I assume they'd be switched to a default one?

dennisreimann:
> How would this affect users using the casa theme and updating? I assume they'd be switched to a default one?

Right now, as it is implemented in this PR, they'd fall back to the default light theme.

Here's a demo of the current state -- the basic implementation is finished now and we can discuss the technical implications (mainly the refactoring of `ThemesSettings.cs` and everything that follows from that) now.

Kukks (approved):
looks good code wise

pavlenex:
Really good flow there @dennisreimann! Great work.

Zaxounette (approved):

dennisreimann:
@pavlenex This'd be a good one for user testing. I'll post it to the testers channel.

dstrukt:
Great work!

I think the next modifications to our menu are going to tie in nicely to this!

I updated the copy, as there are no pre-selectable themes (anymore), just adding a custom CSS file. I took a stab at it. 

"Please provide a CSS theme file."

Feel free to modify or improve as needed.

BTCPayServer/Views/Server/Theme.cshtml
```diff
@@ -3,64 +3,38 @@
...
+}
+
 <h2 class="mb-4">@ViewData["PageTitle"]</h2>
 
 <div class="row">
     <div class="col-lg-6">
         <form method="post">
             <h4 class="mb-3">Custom theme</h4>
-            <p>Select one of our predefined themes. Optionally you can also provide a CSS file that customizes the chosen theme.</p>
+            <p>Please provide a CSS theme file.</p>
```
Zaxounette, line 16:
```suggestion
            <p>Use the default Light or Dark Themes, or provide a CSS theme file below.</p>
```

Zaxounette (changes requested):
@dstrukt 
Iterated upon your proposed COPY.

BTCPayServer/Views/Server/Theme.cshtml
dstrukt: There we go.

dennisreimann: Thanks to both of you! 🙏

dennisreimann:
Rebased and ready to rock!

PR #2815, opened 2021-08-30
Add ability to set default payment method on invoice basis

bolatovumar:
See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/2793

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceExtensions.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceExtensions.cs
@@ -48,19 +48,5 @@ namespace BTCPayServer.Services.Invoices
                 eventAggregator.Publish(new InvoiceNeedUpdateEvent(invoice.Id));
             }
         }
-
-        public static PaymentMethodId GetDefaultPaymentId(
-            this InvoiceRepository invoiceRepository,
-            PaymentMethodId[] paymentMethodIds,
-            InvoiceEntity invoice
-        )
-        {
-            PaymentMethodId.TryParse(invoice.DefaultPaymentMethod, out var defaultPaymentId);
-            var chosen = paymentMethodIds.FirstOrDefault(f => f == defaultPaymentId) ??
-                         paymentMethodIds.FirstOrDefault(f => f.CryptoCode == defaultPaymentId?.CryptoCode) ??
-                         paymentMethodIds.FirstOrDefault();
-            return chosen;
-        }
-
     }
 }
--- a/BTCPayServer/Services/Invoices/InvoiceExtensions.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceExtensions.cs
@@ -10,8 +10,8 @@ namespace BTCPayServer.Services.Invoices
 {
     public static class InvoiceExtensions
     {
-        
-        public static async Task ActivateInvoicePaymentMethod(this InvoiceRepository invoiceRepository, 
+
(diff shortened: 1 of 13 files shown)


Kukks (changes requested):
Code looks good but you need a test

Kukks (approved):

Kukks (changes requested):
Selenium test you added fails. Maybe you need to wait a bit before trying to grab the currency dropdown

bolatovumar:
@Kukks I fixed the test (it was using the wrong selector) and confirmed that it passes locally but now a bunch of tests are failing in CI because of timeouts. I don't think it's related to my change as I only change a selector in one test. Is there a way to re-run the tests? I can't click on that button in CI.

bolatovumar:
@Kukks alright, build is green now.

Kukks (approved):

BTCPayServer/Services/Invoices/InvoiceExtensions.cs
```diff
@@ -49,5 +49,18 @@ public static class InvoiceExtensions
             }
         }
 
+        public static PaymentMethodId GetDefaultPaymentId(
+            this InvoiceRepository invoiceRepository,
```
NicolasDorier, line 53: 1. `invoiceRepository` is not used
2. Do we really need this? I think this extension method is used only once.

bolatovumar: @NicolasDorier I added this here because the similar method for getting the default payment id for a store was in a store extensions class os I just did the same here. I can move it out of here though, it doesn't really matter.

bolatovumar: @NicolasDorier updated in this commit: 97f5fac2024fc58c7d3b3f37f826b3d5884ef8ae (there is a failing test in CI but it's unrelated)

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -1025,6 +1025,11 @@
                         },
                         "description": "A specific set of payment methods to use for this invoice (ie. BTC, BTC-LightningNetwork). By default, select all payment methods enabled in the store."
                     },
+                    "defaultPaymentMethod": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "Default payment type for the invoice (e.g., BTC, BTC_LightningLike). Default payment method set for the store is used if this parameter is not specified."
```
ndeet, line 1031: nitpick: on line 1026 above the examples use `-` to concatenate the payment method strings `BTC-LightningNetwork`, not sure if underscore also works too but for consistency we should stick to one of them. As the Greenfield API returns `BTC-LightningNetwork` I would stick with that

bolatovumar: @ndeet Actually, both of them work. For some reason internally we use the one with the underscore but the one with a dash also works. I updated the example in this commit: be3b1f3ec79f353f74ba0efe333e1053c1a544ff

Just to provide an example, this works:

```json
{
    "currency": "USD",
    "amount": 100,
    "checkout": {
        "defaultPaymentMethod": "BTC-LightningNetwork"
    }
}
```

and this works:

```json
{
    "currency": "USD",
    "amount": 100,
    "checkout": {
        "defaultPaymentMethod": "BTC_LightningLike"
    }
}
```

PR #2830, opened 2021-08-31
Taproot support for wallets

sageprogrammer:
This PR adds experimantal support for taproot in btcpayserver.
It enables creating a wallet with Taproot support.

### DIFF
--- a/BTCPayServer.Tests/docker-compose.altcoins.yml
+++ b/BTCPayServer.Tests/docker-compose.altcoins.yml
@@ -84,7 +84,7 @@ services:
       - customer_lnd
       - merchant_lnd
   nbxplorer:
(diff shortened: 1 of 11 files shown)


NicolasDorier:
Seems good. I'll deploy this on signet.
Can you remove the useless formatting change?

Also would be good to have a test and make sure greeenfield API support it.
You can inspire yourself on the test `WalletAPITests`. No need to test all of it, but at least generating a new hotwallet using taproot, and see if you can get a new address and receive money on it.

NicolasDorier:
I added your PR to our signet instance: https://signet.demo.btcpayserver.org/
Tested manually sending and receiving work, and it works! :))

PR #2837, opened 2021-09-03
Decouple taproot activation from support in generate wallet

NicolasDorier:
Built upon @sageprogrammer modifications. Rather than hiding taproot before the activation, we show it in the UI but disable the choice.

I also made sure that we don't show taproot and segwit options for shitcoin not supporting it.

image

### DIFF
--- a/BTCPayServer/Views/Stores/ImportWallet/Xpub.cshtml
+++ b/BTCPayServer/Views/Stores/ImportWallet/Xpub.cshtml
@@ -47,32 +47,35 @@
     </tr>
     </thead>
     <tbody>
-    <tr>
-        <td rowspan="4">P2WPKH</td>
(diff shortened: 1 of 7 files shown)


BTCPayServer/Views/Stores/ImportWallet/Xpub.cshtml
```diff
@@ -83,7 +86,7 @@
     <tr>
         <td class="font-monospace">pkh(xpub…/0/*)</td>
     </tr>
-    @if (Model.CanUseTaproot)
+    @if (Model.SupportTaproot)
     {
```
sageprogrammer, line 90: Would it not be better with `@if (Model.SupportTaproot && Model.IsTaprootActivated)`

sageprogrammer:
Looks good to me.

PR #2842, opened 2021-09-05
Add ability to hide a product from POS app

bolatovumar:
See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/2827

| Before|After|
|---|---|
|Screen Shot 2021-09-05 at 3 03 38 PM|Screen Shot 2021-09-05 at 3 02 46 PM|

### DIFF
--- a/BTCPayServer/Services/Apps/AppService.cs
+++ b/BTCPayServer/Services/Apps/AppService.cs
@@ -290,7 +290,7 @@ namespace BTCPayServer.Services.Apps
                     itemNode.Add("image", new YamlScalarNode(item.Image));
                 }
                 itemNode.Add("custom", new YamlScalarNode(item.Custom.ToStringLowerInvariant()));
-                itemNode.Add("unavailable", new YamlScalarNode(item.Unavailable.ToStringLowerInvariant()));
+                itemNode.Add("disabled", new YamlScalarNode(item.Disabled.ToStringLowerInvariant()));
                 if (item.Inventory.HasValue)
                 {
                     itemNode.Add("inventory", new YamlScalarNode(item.Inventory.ToString()));
@@ -311,7 +311,7 @@ namespace BTCPayServer.Services.Apps
             var serializer = new SerializerBuilder().Build();
             return serializer.Serialize(mappingNode);
         }
-        public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency, bool filterUnavailable = false)
+        public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency, bool filterDisabled = false)
         {
             if (string.IsNullOrWhiteSpace(template))
                 return Array.Empty<ViewPointOfSaleViewModel.Item>();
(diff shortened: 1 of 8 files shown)


bolatovumar:
Screen Shot 2021-09-05 at 5 23 44 PM

Failing test is unrelated

dstrukt:
Concept ACK!

@Zaxounette Any wording suggestions? 

"Mark as hidden"? "Hide Item"? To maybe be a little more specific?

dennisreimann (commented):
I think it would be better to exclude hidden items when fetching them instead of skipping them in the tempate.

Other than that it looks good to me and I think this is a good addition!

Kukks (changes requested):
The functionality is more "enabling/disabling" than hiding, so I would prefer it renamed

pavlenex:
> The functionality is more "enabling/disabling" than hiding, so I would prefer it renamed

Wanted to say the same.

bolatovumar:
> The functionality is more "enabling/disabling" than hiding, so I would prefer it renamed

@Kukks @pavlenex renamed "hidden" to "unavailable" here: 868814900b1e15f952e1001aa12b0ea7b3159187. Does this name make sense?

> I think it would be better to exclude hidden items when fetching them instead of skipping them in the tempate.
> 
> Other than that it looks good to me and I think this is a good addition!

@dennisreimann the problem here is if I exclude these items at when they are fetched it will also hide them in the template editor since it's using the same source of data as the template. I guess I could have the template editor and the templates using different data properties but it feels like a bit much (maybe I'm wrong?).

dennisreimann:
@bolatovumar The [`AppService.Parse` method](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Services/Apps/AppService.cs#L313) could take a parameter to configure the fetching behavior, so that it could vary depending on the use case.

Zaxounette (changes requested):
> @Kukks @pavlenex renamed "hidden" to "unavailable" here: 8688149. Does this name make sense?

I'd replace that with `Enabled <> Disabled` purely and simply. With both words visible in the toggle UI.

bolatovumar:
> @bolatovumar The [`AppService.Parse` method](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Services/Apps/AppService.cs#L313) could take a parameter to configure the fetching behavior, so that it could vary depending on the use case.

@dennisreimann Alright, I added a parameter to filter hidden items out just for the app FE and not the template editor in this commit: 4040cff8ee2db1ac75544f72f2475d0f1ffd97ac

Also renamed "Unavailable" to "Disabled" here: 0d9dce0025101120308012a50936766dc27f5625

dennisreimann (approved):
Code looks good. The simple parameter works for now and could easily be extended if that‘s needed in the future.

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -310,7 +311,7 @@ public string SerializeTemplate(ViewPointOfSaleViewModel.Item[] items)
             var serializer = new SerializerBuilder().Build();
             return serializer.Serialize(mappingNode);
         }
-        public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency)
+        public ViewPointOfSaleViewModel.Item[] Parse(string template, string currency, bool filterDisabled = false)
```
Kukks, line 314: Meh, the parser should be agnostic of business logic imo (like it was before) but it's a small thing we can change later if we want.

Kukks (changes requested):
Can you add a test so that if disabled, an invoice isnt created?

BTCPayServer/Services/Apps/AppService.cs
bolatovumar: @Kukks added test here: 48c22e8db48057b0467b59da03fb8ef1ce861565

Also updated the `Parse` method not to accept `filterDisabled` and do filtering elsewhere: 40964014cd927a88c34fed7c4f379b2de038d81c

bolatovumar:
Selenium tests failing for random reasons, it doesn't really matter.

Kukks (approved):
verynice.gif

PR #2843, opened 2021-09-06
Disallow cancelling payment request when "Allow payee to create invoices in their own denomination" is not enabled

### LINKED ISSUE #2802: Do not allow invoice `cancellation` for `payment requests` when "Allow payee to create invoices in their own denomination" is not enabled
**Describe the bug**

The payment request can be easily canceled, which is a bit out of ordinary behavior because such an invoice is marked as `invalid marked`. If a user is used to seeing `invalid` status in their invoices view, they may make a mistake of labeling it as `complete` thinking the payment request is paid but has not received confirmation on the blockchain on time.

This is indeed an edge-case, but due to this unexpected error, a mistake can happen. I've talked to @Kukks about this and we agreed this shouldn't happen.

The proposal is to disable `payment request cancelation` unless `Allow payee to create invoices in their own denomination` is enabled on the payment request. The reason behind this is to allow payee who may have created a wrong amount to `cancel the invoice`

**To Reproduce the bug**
1. Create a payment request
2. Click `pay`button
3. Click `cancel invoice`
4. Go to `invoices` list and see the payment request invoice marked as `invalid`

Screenshot 2021-08-23 at 10 43 23

### Logs

Interestingly enough this also throws an error in logs 

```
021-08-23 08:18:04.491 +00:00 [INF] BTC: New block
2021-08-23 08:18:09.140 +00:00 [INF] User with ID 2c117299-f3e3-424c-8709-587a14c27b95 logged in with 2fa.
2021-08-23 08:30:13.469 +00:00 [ERR] Failed to invoke hub method 'Pay'.
System.NullReferenceException: Object reference not set to an instance of an object.
   at BTCPayServer.PaymentRequest.PaymentRequestHub.Pay(Nullable`1 amount) in /source/BTCPayServer/PaymentRequest/PaymentRequestHub.cs:line 52
   at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.ExecuteHubMethod(ObjectMethodExecutor methodExecutor, THub hub, Object[] arguments)
   at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.<>c__DisplayClass13_0.<<Invoke>g__ExecuteInvocation|0>d.MoveNext()
2021-08-23 08:30:22.624 +00:00 [INF] Invoice 9GxoEfPEf8Awf6YSF52qy new event: invoice_created (1001)
2021-08-23 08:30:26.275 +00:00 [INF] Invoice 9GxoEfPEf8Awf6YSF52qy new event: invoice_markedInvalid (1008)
```

bolatovumar:
close #2802

### DIFF
--- a/BTCPayServer.Tests/PaymentRequestTests.cs
+++ b/BTCPayServer.Tests/PaymentRequestTests.cs
@@ -234,12 +234,12 @@ namespace BTCPayServer.Tests
 
                 invoice = user.BitPay.GetInvoice(invoiceId, Facade.Merchant);
 
-                //a hack to generate invoices for the payment request is to manually create an invocie with an order id that matches:
+                //a hack to generate invoices for the payment request is to manually create an invoice with an order id that matches:
                 user.BitPay.CreateInvoice(new Invoice(1, "USD")
                 {
                     OrderId = PaymentRequestRepository.GetOrderIdForPaymentRequest(paymentRequestId)
                 });
-                //shouldnt crash
+                //shouldn't crash
                 await paymentRequestController.ViewPaymentRequest(paymentRequestId);
                 await paymentRequestController.CancelUnpaidPendingInvoice(paymentRequestId);
             }
--- a/BTCPayServer.Tests/PaymentRequestTests.cs
+++ b/BTCPayServer.Tests/PaymentRequestTests.cs
@@ -189,10 +189,20 @@ namespace BTCPayServer.Tests
                 var response = Assert
                     .IsType<RedirectToActionResult>(paymentRequestController.EditPaymentRequest(null, request).Result)
                     .RouteValues.First();
+                var invoiceId = response.Value.ToString();
+                await paymentRequestController.PayPaymentRequest(invoiceId, false);
+                Assert.IsType<BadRequestObjectResult>(await
+                    paymentRequestController.CancelUnpaidPendingInvoice(invoiceId, false));
+
+                request.AllowCustomPaymentAmounts = true;
+
+                response = Assert
+                    .IsType<RedirectToActionResult>(paymentRequestController.EditPaymentRequest(null, request).Result)
+                    .RouteValues.First();
 
                 var paymentRequestId = response.Value.ToString();
 
-                var invoiceId = Assert
+                invoiceId = Assert
                     .IsType<OkObjectResult>(await paymentRequestController.PayPaymentRequest(paymentRequestId, false))
                     .Value
                     .ToString();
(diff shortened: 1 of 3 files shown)


Kukks (changes requested):
You need to add the same logic to the backend.

bolatovumar:
> You need to add the same logic to the backend.

@Kukks done here: fda2b00c84cdb4422c822d2f6db09f1981b33989

Also updated the non-JS view to follow the same logic here: https://github.com/btcpayserver/btcpayserver/pull/2843/commits/9a8216fd2f62e8903ed94d452ac1c9d6ea8f27ab

Kukks (approved):
Awesome lgtm

Kukks (changes requested):
Actually, you now need to adjust the test `CanCancelPaymentWhenPossible` as the logic is different.

BTCPayServer.Tests/PaymentRequestTests.cs
```diff
@@ -189,10 +189,20 @@ public async Task CanCancelPaymentWhenPossible()
                 var response = Assert
                     .IsType<RedirectToActionResult>(paymentRequestController.EditPaymentRequest(null, request).Result)
                     .RouteValues.First();
+                var invoiceId = response.Value.ToString();
+                await paymentRequestController.PayPaymentRequest(invoiceId, false);
+                Assert.IsType<BadRequestObjectResult>(await
```
bolatovumar, line 194: I will first try to cancel the payment request without `AllowCustomPaymentAmounts` set to `true` and assert that it fails and then set `AllowCustomPaymentAmounts` to `true`  and expect the rest of the test to pass as is was previously.

bolatovumar:
> Actually, you now need to adjust the test `CanCancelPaymentWhenPossible` as the logic is different.

@Kukks updated tests in this commit: 37222aacff37ef7eb8cdcaa13edd6eecc607fcbb

BTCPayServer/Controllers/PaymentRequestController.cs
```diff
@@ -299,6 +299,10 @@ public async Task<IActionResult> CancelUnpaidPendingInvoice(string id, bool redi
                 return NotFound();
             }
 
+            if (!result.AllowCustomPaymentAmounts) {
+                return BadRequest("Not allowed to cancel this invoice");
```
NicolasDorier, line 303: @rockstardev do we have a vin diesel error page for this one? :p

NicolasDorier:
Ping @pavlenex you should take a look as well

PR #2857, opened 2021-09-08
Replace pruning wallet by wipe transactions

NicolasDorier:
It was always possible to prune transactions from history so heavy wallet slowing down can retrieve better performance.
Problem is, it is hard to explain in UX sense that some transaction, but not all, will get forgotten in the history.
Instead it is clearer to ask users to wipe all the transactions, then do a rescan.

Maybe we should keep both feature alive. I am still wondering...


NicolasDorier:
A reason to keep the pruning feature is that rescanning only work for shitcoin supporting it and administrator.
If you sweep while not in this case, you may end up with a empty wallet which you can't restore on the server.

NicolasDorier:
I updated this PR to have both clean and prune.

Issue #2856, opened 2021-09-07
Vuejs and summernote vulnerability?


NicolasDorier:
We got two reports of XSS (in [Point of Sale](https://huntr.dev/bounties/0fcdee5f-1f07-47ce-b650-ea8b4a7d35d8/?token=7a607e7fd66eb51ca4284358e497fd770985ceefdbd0a578361c659f71b2472d7d0a17890662a4e885214e4a137cee3132f2f0ee95eb1e757345c9f47db3cd6baa8e5b68c1312b7cbf00697a6957e782c8cd96d020c58f551392f69bf5123161b85e775c214e9032f6ed5340e3291c8e62d1ff05c03262d84d86a9165fcddd) & in [Payment Requests](https://huntr.dev/bounties/ffabdac8-7280-4806-b70c-9b0d1aafbb6e/?token=7a607e7fd66eb51ca4284358e497fd770985ceefdbd0a578361c659f71b2472d7d0a17890662a4e885214e4a137cee3132f2f0ee95eb1e757345c9f47db3cd6baa8e5b68c1312b7cbf00697a6957e782c8cd96d020c58f551392f69bf5123161b85e775c214e9032f6ed5340e3291c8e62d1ff05c03262d84d86a9165fcddd))

It seems two components are failing us: summernote and vuejs.

Both reports are subject to `summernote` issue: It is possible to execute some JS code with some carefully crafted link added with the rich editor: `"><img src=x onerror=alert(225552) ~2F>`. The fact that the JS executes right away is not the scary part. The scary part is that when it is saved, it get executed every times you run in the settings page of the payment request/point of sale.

For the `vuejs` issue, I noticed that the public view of the point of sale doesn't execute the javascript, while the public view of the payment request execute the javascript.

I was at first blaming our server side html sanitizer, but it is doing a sane job:

```
Input: <p><a href="http://&quot;><img src=x onerror=alert(225552) ~2F>" target="_blank">"&gt;<img src="x" onerror="alert(225552)" ~2f=""></a><br></p>
Output:<p><a href="http://&quot;&gt;&lt;img src=x onerror=alert(225552) ~2F&gt;" target="_blank">"&gt;<img src="x" class="img-fluid"></a><br></p>
```

If you copy the output in a HTML, you can see that it is interpreted correctly by a browser:
For example, you can see no injection in the public PoS page:

[...]

PR #2859, opened 2021-09-08
Fix Summernote XSS possibility

dennisreimann:
Addresses #2856, which is related to summernote/summernote#3782.
As this is an open issue on the Summernote repo, I tried fixing the bug by patching it directly in our code. 

If this work, we might propose it as a fix for summernote/summernote#3934 too, which isn't directly part of the other mentioned issue but also still ongoing.


NicolasDorier:
strange, build is broken but I don't see errors.

NicolasDorier:
Ah yes, there are errors. It is the bundler doesn't understand your syntax.

dennisreimann:
Build fixed, integration test is failing because HitBTC isn't available right now.

dennisreimann:
Shall we close this now? imho it might still make sense to include the changes, @NicolasDorier

PR #2863, opened 2021-09-09
Add CSP at the website level

NicolasDorier:
Long time ago, I attempted to add [CSP](https://github.com/btcpayserver/btcpayserver/pull/2391) to BTCPayServer and failed.

Given the multiple XSS vulnerabilities that appeared in the project in the past and [currently](https://github.com/btcpayserver/btcpayserver/issues/2856), I decided to give CSP another try.

The PR automatically set csp with `nonce-` or `sha256-` to handlers and inline scripts generated server side, thus protecting against client side injection attacks while not bothering the development server side.

The remove the need of https://github.com/btcpayserver/btcpayserver/pull/2859 because it protects use from XSS in dependencies as well.

I am testing manually as there are risks of things breaking.


NicolasDorier:
The break the payment button wizard, as it indeed insert javascript in the preview...

NicolasDorier:
Not simple... this is breaking lot's of thing. For example we can't do any inline events...

NicolasDorier:
So I create two tag helper automatically injecting csp attributes in the HTTP headers/html attributes. And it works!

PR #2872, opened 2021-09-11
Fix pay button CSP issue when using modal

### LINKED ISSUE #2864: When Use Modal is selected in paybutton wizard, we have a CSP issue error
Despite this, the rest seems to work properly, it only crashes the preview.
Bug introduced by the new CSP rule: https://github.com/btcpayserver/btcpayserver/pull/2863

image

dennisreimann:
Fixes #2864.

### DIFF
--- a/BTCPayServer/wwwroot/paybutton/paybutton.js
+++ b/BTCPayServer/wwwroot/paybutton/paybutton.js
@@ -96,7 +96,7 @@ function inputChanges(event, buttonSize) {
         // Styles
         getStyles('template-paybutton-styles') + (isSlider ? getStyles('template-slider-styles') : '') +
         // Form
-        '<form method="POST" ' + (srvModel.useModal ? ' ' : '') +' onsubmit="onBTCPayFormSubmit(event);return false" action="' + esc(srvModel.urlRoot) + actionUrl + '" class="btcpay-form btcpay-form--' + (srvModel.fitButtonInline ? 'inline' : 'block') +'">\n' +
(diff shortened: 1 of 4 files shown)


BTCPayServer/wwwroot/paybutton/paybutton.js
```diff
@@ -115,12 +93,10 @@ function inputChanges(event, buttonSize) {
     }
     
     var html =
-        //Scripts
-        (srvModel.useModal? getScripts(srvModel) :"") +
         // Styles
         getStyles('template-paybutton-styles') + (isSlider ? getStyles('template-slider-styles') : '') +
         // Form
-        '<form method="POST" '+ ( srvModel.useModal? ' onsubmit="onBTCPayFormSubmit(event);return false" ' : '' )+' action="' + esc(srvModel.urlRoot) + actionUrl + '" class="btcpay-form btcpay-form--' + (srvModel.fitButtonInline ? 'inline' : 'block') +'">\n' +
+        '<form method="POST" '+ ( srvModel.useModal? ' ' : '' )+' id="btcpay-paybutton-form" action="' + esc(srvModel.urlRoot) + actionUrl + '" class="btcpay-form btcpay-form--' + (srvModel.fitButtonInline ? 'inline' : 'block') +'">\n' +
```
NicolasDorier, line 99: It is a problem for people who have several button on a page.
We can just allow this particular event handler in the page instead with `unsafe-hashes` and `sha256-`.

I can refactor a bit myself to do this. (will edit your PR)

NicolasDorier:
Take a look at my commit. I also improved a bit the taghelpers.

This we we can add the handler explicitely in the cshtml.

BTCPayServer/wwwroot/paybutton/paybutton.js
dennisreimann: Good point! Had to make a slight adjustment in 1c7ce60, because otherwise it got called on submit regardless of the modal being used or not.

PR #2876, opened 2021-09-13
Improve public LN node info 

dennisreimann:
Allows to view clearnet and Tor connection info side by side.

lninfo

Also displays the store name in case the LN node is unavailable.

### DIFF
--- a/BTCPayServer/Controllers/StoresController.LightningLike.cs
+++ b/BTCPayServer/Controllers/StoresController.LightningLike.cs
@@ -109,7 +109,7 @@ namespace BTCPayServer.Controllers
                     var handler = _ServiceProvider.GetRequiredService<LightningLikePaymentHandler>();
                     try
                     {
-                        var info = await handler.GetNodeInfo(paymentMethod, network);
+                        var info = await handler.GetNodeInfo(paymentMethod, network, Request.IsOnion());
                         if (!vm.SkipPortTest)
                         {
                             using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(20));
--- a/BTCPayServer/Controllers/StoresController.LightningLike.cs
+++ b/BTCPayServer/Controllers/StoresController.LightningLike.cs
@@ -109,13 +109,13 @@ namespace BTCPayServer.Controllers
                     var handler = _ServiceProvider.GetRequiredService<LightningLikePaymentHandler>();
                     try
                     {
-                        var info = await handler.GetNodeInfo(Request.IsOnion(), paymentMethod, network);
+                        var info = await handler.GetNodeInfo(paymentMethod, network);
                         if (!vm.SkipPortTest)
                         {
                             using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(20));
-                            await handler.TestConnection(info, cts.Token);
+                            await handler.TestConnection(info.First(), cts.Token);
                         }
-                        TempData[WellKnownTempData.SuccessMessage] = $"Connection to the Lightning node successful. Your node address: {info}";
+                        TempData[WellKnownTempData.SuccessMessage] = $"Connection to the Lightning node successful. Your node address: {info.First()}";
                     }
                     catch (Exception ex)
                     {
--- a/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
@@ -135,7 +135,8 @@ namespace BTCPayServer.Payments.Lightning
                 }
                 catch (Exception ex)
                 {
-                    throw new PaymentMethodUnavailableException($"Error while connecting to the API ({ex.Message})");
+                    throw new PaymentMethodUnavailableException($"Error while connecting to the API: {ex.Message}" + 
+                                                                (!string.IsNullOrEmpty(ex.InnerException?.Message) ? $" ({ex.InnerException.Message})" : "")));
                 }
 
                 var nodeInfo = preferOnion != null && info.NodeInfoList.Any(i => i.IsTor == preferOnion)
--- a/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
@@ -68,8 +68,6 @@ namespace BTCPayServer.Payments.Lightning
             }
             //direct casting to (BTCPayNetwork) is fixed in other pull requests with better generic interfacing for handlers
             var storeBlob = store.GetStoreBlob();
-            var test = GetNodeInfo(supportedPaymentMethod, network);
-            
             var invoice = paymentMethod.ParentEntity;
             decimal due = Extensions.RoundUp(invoice.Price / paymentMethod.Rate, network.Divisibility);
             try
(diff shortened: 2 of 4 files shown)


BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -109,13 +109,13 @@ public async Task<IActionResult> SetupLightningNode(string storeId, LightningNod
                     var handler = _ServiceProvider.GetRequiredService<LightningLikePaymentHandler>();
                     try
                     {
-                        var info = await handler.GetNodeInfo(Request.IsOnion(), paymentMethod, network);
+                        var info = await handler.GetNodeInfo(paymentMethod, network);
                         if (!vm.SkipPortTest)
                         {
                             using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(20));
-                            await handler.TestConnection(info, cts.Token);
```
NicolasDorier, line 116: what if `info` is empty?

BTCPayServer/Controllers/StoresController.LightningLike.cs
```diff
@@ -109,13 +109,13 @@ public async Task<IActionResult> SetupLightningNode(string storeId, LightningNod
...
+                            await handler.TestConnection(info.First(), cts.Token);
```
NicolasDorier, line 116: what if `info` is empty?

dennisreimann: `GetNodeInfo` will [throw](https://github.com/btcpayserver/btcpayserver/pull/2876/files#diff-e4f820f06ea4508b3e67ede9be0bf524b091925a386355a87bd5a7973abbe438R146) an exception.

But I guess you mean this should be handled differently?

NicolasDorier: I was thinking about it, this test should prefer clearnet test, or we should improve the "test" to test on Tor as well. (probably in another PR)

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -118,7 +118,7 @@ public class LightningLikePaymentHandler : PaymentMethodHandlerBase<LightningSup
             };
         }
 
-        public async Task<NodeInfo> GetNodeInfo(bool preferOnion, LightningSupportedPaymentMethod supportedPaymentMethod, BTCPayNetwork network)
```
NicolasDorier, line 121: So I think you should keep `GetNodeInfo` as `preferOnion` is useful for `test connection` for example.
However, you can make it `boolean?`, and change the `FirstOrDefault(x)` to `Where(x).ToArray()`

NicolasDorier: We probably can get ride of the `preferOnion` later.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -139,8 +139,9 @@ public async Task<NodeInfo> GetNodeInfo(bool preferOnion, LightningSupportedPaym
                 {
                     throw new PaymentMethodUnavailableException($"Error while connecting to the API ({ex.Message})");
                 }
-                var nodeInfo = info.NodeInfoList.FirstOrDefault(i => i.IsTor == preferOnion) ?? info.NodeInfoList.FirstOrDefault();
```
NicolasDorier, line 142: Here it would get a list of tor nodes if tor is preferred. If the array is empty, taking the non tor list.

dennisreimann:
@NicolasDorier Thanks, I addressed your feedback. Let me know if there are still things I missed or that should get handled differently.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -137,10 +135,15 @@ public async Task<NodeInfo> GetNodeInfo(bool preferOnion, LightningSupportedPaym
...
                 {
-                    throw new PaymentMethodUnavailableException($"Error while connecting to the API ({ex.Message})");
+                    throw new PaymentMethodUnavailableException($"Error while connecting to the API: {ex.Message}" + 
+                                                                (!string.IsNullOrEmpty(ex.InnerException?.Message) ? $" ({ex.InnerException.Message})" : ""));
                 }
-                var nodeInfo = info.NodeInfoList.FirstOrDefault(i => i.IsTor == preferOnion) ?? info.NodeInfoList.FirstOrDefault();
-                if (nodeInfo == null)
+
+                var nodeInfo = preferOnion != null && info.NodeInfoList.Any(i => i.IsTor == preferOnion)
+                    ? info.NodeInfoList.Where(i => i.IsTor == preferOnion).ToArray()
```
NicolasDorier, line 143: `preferOnion.Value`

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -108,17 +106,17 @@ public class LightningLikePaymentHandler : PaymentMethodHandlerBase<LightningSup
                     throw new PaymentMethodUnavailableException($"Impossible to create lightning invoice ({ex.Message})", ex);
                 }
             }
-            var nodeInfo = await test;
+            var nodeInfo = await GetNodeInfo(supportedPaymentMethod, network, paymentMethod.PreferOnion);
```
NicolasDorier, line 109: You should keep `test` like I did before. The reason I was doing this, is because we are making 2 API calls to lightning server, so we want them to perform in parallel. This is why the `GetNodeInfo` is fired, but I only await later the result.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -108,17 +106,17 @@ public class LightningLikePaymentHandler : PaymentMethodHandlerBase<LightningSup
...
             return new LightningLikePaymentMethodDetails
             {
                 Activated = true,
                 BOLT11 = lightningInvoice.BOLT11,
                 InvoiceId = lightningInvoice.Id,
-                NodeInfo = nodeInfo.ToString()
+                NodeInfo = nodeInfo.FirstOrDefault()?.ToString()
```
NicolasDorier, line 115: I think you can use `First` here, because `GetNodeInfo` throw if there is no node.

BTCPayServer/Views/PublicLightningNodeInfo/ShowLightningNodeInfo.cshtml
```diff
@@ -2,10 +2,16 @@
 @inject ISettingsRepository _settingsRepository
 
 @using BTCPayServer.Abstractions.Contracts
+@using BTCPayServer.Lightning
 @model BTCPayServer.Controllers.ShowLightningNodeInfoViewModel
 @{
     Layout = null;
     var theme = await _settingsRepository.GetTheme();
+    var nodeInfoList = Model.NodeInfo.Select((value, i) => (value, i)).ToArray();
```
NicolasDorier, line 10: Just change `NodeInfo` to be an object of a class with all the details you want, and fill that in the controller rather than hacking around in the view.

dennisreimann:
@NicolasDorier Updated, thanks for the feedback!

dennisreimann:
Checked and rebased -- hit that button, @NicolasDorier!

PR #2894, opened 2021-09-19
Fix issues with Authorization Request page

### LINKED ISSUE #2858: Authorize API key page is broken when trying to select specific stores
**Describe the bug**
The screenshot explains all.

Technically the issue is caused by form POSTs on every press and they all use the same field name, overwriting the previous value. This cannot work this way and needs both design + technical overhaul.

Currently you can only create an API key for "all stores", not limit to 1 or more specific stores. This part is broken.

Example URL (just replace the hostname) = https://your-server.com/api-keys/authorize?permissions=btcpay.store.canviewinvoices&permissions=btcpay.store.cancreateinvoice&permissions=btcpay.store.webhooks.canmodifywebhooks&permissions=btcpay.store.canviewstoresettings&applicationName=My+Integration&strict=true&selectiveStores=true&redirect=http%3A%2F%2my-server-com%2Fapikey%2Fsave&applicationIdentifier=myintegration

**Screenshots**
image (1)

- BTCPay Server Version:   1.2.1.0
- Deployment Method:       Docker
- Browser:         All

**Proposed solution**
My suggestion would be to make this very easy. Put a store selector at the top and the permissions underneath. So you only need to select your store(s), consider the permissions and click "Authenticate app".

This solution, while it is easy, does prevent you from selecting different stores per permission, but this would be weird anyway. If you need different permissions per store, you'll likely create different API keys and not cram all into 1 API key.

bolatovumar:
1. Fixed a bunch of issues described in #2858 
2. Minor visual adjustment so that the row of stuff for each permission looks nicer:

|Before|After|
|---|---|
|before|after|

closes #2858

### DIFF
--- a/BTCPayServer/Controllers/ManageController.APIKeys.cs
+++ b/BTCPayServer/Controllers/ManageController.APIKeys.cs
@@ -384,10 +384,11 @@ namespace BTCPayServer.Controllers
                     permissionValueItem.StoreMode = permissionValueItem.StoreMode == AddApiKeyViewModel.ApiKeyStoreMode.Specific
                         ? AddApiKeyViewModel.ApiKeyStoreMode.AllStores
                         : AddApiKeyViewModel.ApiKeyStoreMode.Specific;
-                    // Make sure we don't keep specific store IDs if we switched back to "all stores" from "specific stores"
+                    // Reset values for "all stores" option to their original values
                     if (permissionValueItem.StoreMode == AddApiKeyViewModel.ApiKeyStoreMode.AllStores)
                     {
                         permissionValueItem.SpecificStores = new List<string>();
+                        permissionValueItem.Value = true;
                     }
 
                     if (permissionValueItem.StoreMode == AddApiKeyViewModel.ApiKeyStoreMode.Specific &&
--- a/BTCPayServer/Controllers/ManageController.APIKeys.cs
+++ b/BTCPayServer/Controllers/ManageController.APIKeys.cs
@@ -195,7 +195,25 @@ namespace BTCPayServer.Controllers
 
         private void AdjustVMForAuthorization(AuthorizeApiKeysViewModel vm)
         {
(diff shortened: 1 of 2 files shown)


pavlenex:
@woutersamaey can you check if this addresses your concerns from #2858?

woutersamaey:
I've asked @JudithWebdev to check this out as she's working on this atm.

bolatovumar:
> I've asked @JudithWebdev to check this out as she's working on this atm.

@woutersamaey the functionality now should be identical to the regular "Add API Key" page under `/Manage/AddApiKey` as I believe it's intended to be.

NicolasDorier:
@bolatovumar can you write a small test so we catch this kind of bug?

woutersamaey:
The code is good, but we have issues with the UI. It should be refactored so that you would first select the stores, and then the permissions. This is essential for an easy eCommerce integration as you'd never create an API key with permission A for store A and a different permission for store B. All stores should always get the same permissions and the UI should enforce that.

This is prolly out of scope for this PR though.

I discussed this with @Kukks and we decided not to use specific store permissions for the integrations for now. It's just not user-friendly enough to go this route.

Ideally you'd have an eCommerce site that has a "Connect" button in the backoffice. You click that and land on the Authorize API key page. You select the stores you want to connect and the required permissions are prefilled and disabled (strict mode).
You confirm and return to the eCommerce backoffice.
If the eCommerce site is multi-site capable (like Magento), you select which BTCPay Store should go with which website. Because you now have the API key, you can list the store IDs and build the dropdowns.

This flow is currently not possible as the user may make a mistake and fail to select the same stores for the permissions. So strict mode only applies to permissions, but not to the stores you're choosing under these permissions. We could make it so that strict mode also forces the same permissions on each store, but I feel it makes more sense to change the UI so stores come first and permissions second.

woutersamaey:
ACK for this PR as it does fix the bugs.

NicolasDorier:
@bolatovumar if we can do a refactor of the UI for this release, then we don't need the tests here and I can just merge this.

Kukks:
Let's do it!

PR #2896, opened 2021-09-20
Support preview/codegen for custom currencies in pay button

bolatovumar:
Previously only one of the default currencies could be specified for the pay button so that you get a preview and codegen. With this PR user can now input any currency (e.g., `SATS` or `JPY`) and get preview in place and appropriate codegen.

If you use a nonsense currency you will get an appropriate error when you generate an invoice:

Capture

|Before|After|
|---|---|
|before|after|
|before2|after2|

See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/2885


dennisreimann (approved):
Tested, works -- good job!

Issue #2884, opened 2021-09-15
Add social icons to footer and remove logo credits


pavlenex:
As discussed a while ago on Mattermost, we should remove the logo credits and add social icons there instead.

Add links to: 
- GitHub
- Mattermost
- Twitter
![Screenshot 2021-09-15 at 18 26 27](https://user-images.githubusercontent.com/36959754/133472401-73be1f22-bad2-4cfa-b44c-49e573c7fb62.png)

1nF0rmed:
Hey - I'd like to pick this up

pavlenex:
Feel free to @1nF0rmed 🚀

PR #2898, opened 2021-09-20
Adds social links to Layout

1nF0rmed:
**Fixes**: #2884 

**Description:**
This removes the logo credits in the base Layout and adds the
links to Github, Mattermost and Twitter.

image


dennisreimann (changes requested):
Hey, thanks for the contribution!

Can you remove the "Let's Get In Touch!" block from the homepage too? I think we can get rid of that with these changes.

We should also ensure this works good on mobile and I'd suggest stacking the links and the copyright in that case, instead of floating them.

The colors can also be optimized, let me know if I should assist with anything.

footer

NicolasDorier:
Awesome! Way better to have it in the footer. I agree that the section can be removed.

1nF0rmed:
> Can you remove the "Let's Get In Touch!" block from the homepage too? I think we can get rid of that with these changes.

Sure, I'll add a commit for the same

> We should also ensure this works good on mobile and I'd suggest stacking the links and the copyright in that case, instead of floating them.

So, we would have 2 layers. One layer with the social links horizontally aligned and the second layer with the copyright information.
Would that be correct?

> The colors can also be optimized, let me know if I should assist with anything.

The text color seems to have decent contrast to be visible. The icons for Github and Mattermost right now can't be seen that well. Can I change it so it looks like how it would in dark mode?

dennisreimann:
> So, we would have 2 layers. One layer with the social links horizontally aligned and the second layer with the copyright information.
> Would that be correct?

Yes, each link as a separate line and the copyright underneath the links on mobile, side by side (as currently) when it starts to fit. (might be the medium/MD breakpoint)

> The text color seems to have decent contrast to be visible. The icons for Github and Mattermost right now can't be seen that well. Can I change it so it looks like how it would in dark mode?

I can take care of that afterwards: I'd do some modifications in the design system to make the colors work and also add the icons to the icon sprite, so that we have them as proper SVGs.

dennisreimann:
@1nF0rmed Can you give me access to your fork, so that I can push my updates? Finetuned it a bit and introduced the icons to the sprite.

1nF0rmed:
> @1nF0rmed Can you give me access to your fork, so that I can push my updates? Finetuned it a bit and introduced the icons to the sprite.

Done

dennisreimann:
Thanks, pushed my changes -- ready to go now!

dennisreimann (approved):

pavlenex:
@dennisreimann doesn't look visible enough in the dark mode (especially important info such as version?)

Screenshot 2021-09-23 at 13 17 50

dennisreimann:
@pavlenex Fair enough, good catch! Good like that?

footer

pavlenex (approved):

PR #2901, opened 2021-09-20
JetBrains Rider Run Configuration for "docker-compose up dev"

woutersamaey:


dennisreimann:
I get the following error when trying to run this:

```
Error running 'docker-compose up dev':
Server Docker not found
```

When I adapt it in the config and select my "Docker" server it brings me one step further, but starting still fails:

```
docker.errors.TLSParameterError: Path to a certificate and key files must be provided through the client_config param. TLS configurations should map the Docker CLI client configurations. See https://docs.docker.com/engine/articles/https/ for API details.
```

woutersamaey:
Hmm yeah it may be better to register this run config as a shell script. That way it does not depend on your environment I guess…
Not sure how to respond to your issues. It’s very basic stuff really.
The TLS issues, could they be on your end? There’s no mention of this in the XML

NicolasDorier:
Is it possible to also add the docker-compose for altcoin? (-f docker-compose.altcoins.yml) so it shows as option as well?

dennisreimann:
> The TLS issues, could they be on your end? There’s no mention of this in the XML

I don't even know what this is about exactly, where it is missing a certificate or what the cert is for. I expected this to work just as fine as when I am running `docker-compose up dev` via CLI.

pavlenex:
Discussed with @woutersamaey and we decided to close this one.

PR #2905, opened 2021-09-22
Fixed wrong API permission for GreenField/StorePaymentMethodsController.cs

woutersamaey:
Viewing store payment methods needs CanViewStoreSettings instead of CanModifyStoreSettings as per the docs

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -2001,14 +2001,14 @@ namespace BTCPayServer.Tests
             
             Assert.True(methods.TryGetValue(new PaymentMethodId("BTC", PaymentTypes.LightningLike).ToStringNormalized(), out var item));
             var lightningNetworkPaymentMethodBaseData =Assert.IsType<JObject>(item.Data).ToObject<LightningNetworkPaymentMethodBaseData>();
-            Assert.Equal("###", lightningNetworkPaymentMethodBaseData.ConnectionString);
+            Assert.Equal("*NEED CanModifyStoreSettings PERMISSION TO VIEW*", lightningNetworkPaymentMethodBaseData.ConnectionString);
             
             
             methods = await adminClient.GetStorePaymentMethods(store.Id);
             
             Assert.True(methods.TryGetValue(new PaymentMethodId("BTC", PaymentTypes.LightningLike).ToStringNormalized(), out item));
             lightningNetworkPaymentMethodBaseData =Assert.IsType<JObject>(item.Data).ToObject<LightningNetworkPaymentMethodBaseData>();
-            Assert.NotEqual("###", lightningNetworkPaymentMethodBaseData.ConnectionString);
+            Assert.NotEqual("*NEED CanModifyStoreSettings PERMISSION TO VIEW*", lightningNetworkPaymentMethodBaseData.ConnectionString);
             
 
         }
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -1953,6 +1953,7 @@ namespace BTCPayServer.Tests
             var admin = tester.NewAccount();
             await admin.GrantAccessAsync(true);
             var adminClient = await admin.CreateClient(Policies.Unrestricted);
+            var viewerOnlyClient = await admin.CreateClient(Policies.CanViewStoreSettings);
             var store = await adminClient.GetStore(admin.StoreId);
 
             Assert.Empty(await adminClient.GetStorePaymentMethods(store.Id));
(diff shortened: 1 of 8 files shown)


woutersamaey:
@Kukks and @ndeet this may interest you. The permissions are wrong in code, but correct in the docs.

Kukks (approved):
i think it is an actual bug but maybe @NicolasDorier has some more insight?

Kukks (changes requested):
So, potentially this is on purpose because for `GetLightningNetworkPaymentMethod`, it is also `CanModifyStoreSettings`.

Most likely, the reasoning is that the payment method contains sensitive information that would compromise the security of your payment method, even if it is only with viewing rights. For example, the lightning connection string could contain the access token/macaroon to control and spend from your lightning node.

woutersamaey:
@Kukks this cuts both ways, doesn't it?

Either you compromise on security by weakening your LN config security
Or, you weaken your installation by creating and storing API keys in an eCommerce system that can be hacked and more control can be gained than was needed.

The best solution would be to censor the potentially dangerous fields if the lesser permissions is used, and include the sensitive info when an API key with more permissions is used.

That way you get security for both cases, but I guess some refactoring would be involved? Or can you simply filter out the values as a filter before data is output from the controller? (i.e. set the dangerous values to NULL or something like `*HIDDEN*` or `***`

Kukks:
> @Kukks this cuts both ways, doesn't it?
> 
> Either you compromise on security by weakening your LN config security
> Or, you weaken your installation by creating and storing API keys in an eCommerce system that can be hacked and more control can be gained than was needed because the API key was created with otherwise less permissions.
> 
> The best solution would be to censor the potentially dangerous fields if the lesser permissions is used, and include the sensitive info when an API key with more permissions is used.
> 
> That way you get security for both cases, but I guess some refactoring would be involved? Or can you simply filter the values before data is output from the controller? (i.e. set the dangerous values to something like `*HIDDEN*` or `***`
> 
> Just trying to find an easy solution here...

Yeah, this is probably the best way.

Kukks:
Fixed up the PR with the discussed changes.

woutersamaey:
Probably nitpicking here, but I believe `*HIDDEN*` would've been less confusing to newbies over that `###` you used.
I fear this will lead to more support questions.

Maybe better yet would be something like `*NEED CanModifyStoreSettings PERMISSION TO VIEW*`

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -1986,6 +1987,29 @@ void VerifyOnChain(Dictionary<string, GenericPaymentMethodData> dictionary)
...
+            Assert.True(methods.TryGetValue(new PaymentMethodId("BTC", PaymentTypes.LightningLike).ToStringNormalized(), out var item));
+            var lightningNetworkPaymentMethodBaseData =Assert.IsType<JObject>(item.Data).ToObject<LightningNetworkPaymentMethodBaseData>();
+            Assert.Equal("###", lightningNetworkPaymentMethodBaseData.ConnectionString);
+            
+            
+            methods = await adminClient.GetStorePaymentMethods(store.Id);
+            
+            Assert.True(methods.TryGetValue(new PaymentMethodId("BTC", PaymentTypes.LightningLike).ToStringNormalized(), out item));
+            lightningNetworkPaymentMethodBaseData =Assert.IsType<JObject>(item.Data).ToObject<LightningNetworkPaymentMethodBaseData>();
+            Assert.NotEqual("###", lightningNetworkPaymentMethodBaseData.ConnectionString);
```
Kukks, line 2011:
```suggestion
            Assert.NotEqual("*NEED CanModifyStoreSettings PERMISSION TO VIEW*", lightningNetworkPaymentMethodBaseData.ConnectionString);
```

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -1986,6 +1987,29 @@ void VerifyOnChain(Dictionary<string, GenericPaymentMethodData> dictionary)
...
+            
+            
+           await  adminClient.UpdateStoreLightningNetworkPaymentMethod(store.Id, "BTC",
+                new UpdateLightningNetworkPaymentMethodRequest(
+                    tester.GetLightningConnectionString(LightningConnectionType.CLightning, true), true));
+            methods = await viewerOnlyClient.GetStorePaymentMethods(store.Id);
+            
+            Assert.True(methods.TryGetValue(new PaymentMethodId("BTC", PaymentTypes.LightningLike).ToStringNormalized(), out var item));
+            var lightningNetworkPaymentMethodBaseData =Assert.IsType<JObject>(item.Data).ToObject<LightningNetworkPaymentMethodBaseData>();
+            Assert.Equal("###", lightningNetworkPaymentMethodBaseData.ConnectionString);
```
Kukks, line 2004:
```suggestion
            Assert.Equal("*NEED CanModifyStoreSettings PERMISSION TO VIEW*", lightningNetworkPaymentMethodBaseData.ConnectionString);
```

BTCPayServer/Payments/PaymentTypes.Lightning.cs
```diff
@@ -65,12 +65,20 @@ public override string GetTransactionLink(BTCPayNetworkBase network, string txId
...
                 {
-                    ConnectionString = lightningSupportedPaymentMethod.GetDisplayableConnectionString()
+                    ConnectionString = lightningSupportedPaymentMethod.IsInternalNode
+                        ?
+                        lightningSupportedPaymentMethod.GetDisplayableConnectionString()
+                        :
+                        canModifyStore
+                            ? lightningSupportedPaymentMethod.GetDisplayableConnectionString()
+                            :
+                            "###"
```
Kukks, line 81:
```suggestion
                            "*NEED CanModifyStoreSettings PERMISSION TO VIEW*"
```

Kukks (changes requested):

woutersamaey:
Awesome!

PR #2910, opened 2021-09-23
Experimental: Cache all the things, all the time

Kukks:


NicolasDorier:
I prefer we don't do this. Nobody complained about performance issues (if they did, it was a bug), and caching lead to difficult to diagnose bugs.

PR #2915, opened 2021-09-23
Improve app update buttons

dennisreimann:
Better responsive behaviour and clearer distinction between primary and secondary actions. Also improves the wording ("View Crowdfund/Point of Sale" vs. "View app") and gets rid of the superfluous "Back to apps list" button.

## Before

before

## After

after-1

after-2


dstrukt:
Agree it's redundant, looks good to me.

Send it!

dstrukt (approved):

PR #2927, opened 2021-09-27
Crowdfund themeing

dennisreimann:
Makes the theme switch a view component and brings it to the public crowdfunding UI.

General approach resembles what we have in the app: 
- No custom theme at the server level: We offer the theme switch, so that the end user gets its own preferred way of seeing the UI (light/dark). 
- Custom theme at the server level: Theme switch not available

With that, now the question is how we want to approach CSS customization at the app level:
- Crowdfunding CSS customization is done based on the above, taking into account that there is a light and a dark version
- When there is custom CSS at the app level, use the light version as a basis to customize things. 

The latter wouldn't break any existing setups, because that's basically what it was like before for those who are using custom CSS. The downside is that it would make adjustments based on the dark theme harder. The reason I would not want to offer a switch for choosing the base in the crowdfunding admin UI is that'd complicate things: In theory then there'd be potentially three layers of CSS then: Server Setting > App Setting > Custom App CSS.

I think it's worth discussion this and finding the most suitable approach, because we could extend it to all the end user facing views, like payment requests, pull payments and invoices as well.


pavlenex:
The theme switch is completely hidden for me now, probably https://github.com/btcpayserver/btcpayserver/pull/2927/commits/c7a874cfe09b79770f3c6a8da95ea867dfca8e4e

dennisreimann:
Hmm, it is there for me, in the main app as well as on the public crowdfund view.

app


cf

Checked Safari and Firefox.

dennisreimann:
Sorry @pavlenex, I had to force push once more to integrate the current master 😬

Can you check whether or not the missing theme switch is a caching issue? For me it works in Safari and Firefox. 
If not, can you use the web inspector to see what the markup/HTML looks like between the notifications and the logout link?

pavlenex (approved):
All good and working now, tACK.

pavlenex:
@dennisreimann  the only nit-pick is this
Screenshot 2021-10-06 at 11 37 31

PR #2897, opened 2021-09-20
LNURL Payment Method Support

Kukks:

PR #2945, opened 2021-10-04
Refactoring: Move payment config

dennisreimann:
This is a preliminary step for changes in #2897: Move some of the payment config options out of the checkout experience into the payments section of the store.

payment

ce

Also displays only those options that are available in the context of how the store is configured: If there is no on-chain wallet set up, those options won't be displayed -- same goes for Lightning.

As this will tend to become crowded over time, we will need to figure out where to move the payment method options.


Kukks (commented):
ack

Kukks (approved):
ack

dstrukt:
ack as well, will take a deeper look in the morning

PR #2946, opened 2021-10-04
Fix CSP for inline handlers in Safari

### LINKED ISSUE #2930: Inline JS doesn't work in Safari due to CSP
**Describe the bug**

Inline JS doesn't work in Safari and it makes @pavlenex  sad. When you try to interact with an element that has inline JS you see this error in the console:

Screen Shot 2021-09-29 at 9 26 10 PM

To solve this issue we can either relax CSP settings or update the inline JS to be not inline.

We have quite a bit of inline JS throughout:

Screen Shot 2021-09-29 at 9 27 05 PM
Screen Shot 2021-09-29 at 9 27 24 PM

**To Reproduce the bug**
1. Go to invoices page in Safari (I'm using Safari 14 but newer version should be fine as well)
2. Click on the checkmark on the top left which is supposed to select all invoices on the page
3. Notice it does nothing
4. Notice there is a CSP error in the console

https://user-images.githubusercontent.com/1934678/135387716-1116f77f-6435-4d5c-adaa-d3ecebafe7b3.mov

**Expected behavior**

Expect behaviors which rely on JS to work.

- BTCPay Server Version: 1.2.4.0
- Deployment Method: Doesn't matter
- Browser: Safari 14.1.2

dennisreimann:
Fixes #2930.

### DIFF
--- a/BTCPayServer/Views/Wallets/SignWithSeed.cshtml
+++ b/BTCPayServer/Views/Wallets/SignWithSeed.cshtml
@@ -5,12 +5,19 @@
 }
 
 @section Navbar {
-    <a asp-action="WalletPSBT" asp-route-walletId="@Context.GetRouteValue("walletId")" onclick="history.back();return false;">
+    <a asp-action="WalletPSBT" asp-route-walletId="@Context.GetRouteValue("walletId")" id="GoBack">
         <vc:icon symbol="back" />
     </a>
     <a asp-action="WalletSend" asp-route-walletId="@Context.GetRouteValue("walletId")" class="cancel">
         <vc:icon symbol="close" />
     </a>
+}
+
+@section PageFootContent
+{
+    <script>
+        delegate('click', '#GoBack', () => { history.back(); return false; })
+    </script>
 }
 
 <header class="text-center">
--- a/BTCPayServer/wwwroot/checkout/js/checkout.js
+++ b/BTCPayServer/wwwroot/checkout/js/checkout.js
@@ -1,0 +1,4 @@
+delegate('click', '.payment-method', function(e) {
+    closePaymentMethodDialog(e.target.dataset.paymentMethod);
+    return false;
+})
--- a/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
+++ b/BTCPayServer/Views/Invoice/Checkout-Body.cshtml
@@ -60,7 +60,7 @@
                         @foreach (var crypto in Model.AvailableCryptos)
                         {
                             <li class="vexmenuitem">
-                                <a href="@crypto.Link" onclick="closePaymentMethodDialog('@crypto.PaymentMethodId');return false;" rel="noreferrer noopener">
+                                <a href="@crypto.Link" class="payment-method" data-payment-method="@crypto.PaymentMethodId" rel="noreferrer noopener">
                                     <img alt="@crypto.PaymentMethodName" src="@crypto.CryptoImage" asp-append-version="true" />
                                     @crypto.PaymentMethodName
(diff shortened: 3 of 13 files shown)


pavlenex (approved):
tACK on Safari 15.00

bolatovumar (commented):
I didn't test everything but what I tested works as expected in Safari 14.

Looks like we might have a few more instances of inline JS left though:

Screen Shot 2021-10-04 at 6 36 18 PM
Screen Shot 2021-10-04 at 6 36 43 PM

NicolasDorier:
Inline is not the problem. There was an issue that I fixed on master already. @pavlenex should test if on master the problem persists, I think it si already fixed.

dennisreimann:
@NicolasDorier This isn't just about inline event handlers. Somehow some scripts like the toggling on the invoices list (checkboxes as well as time format) didn't work at all for me in Safari.

NicolasDorier:
@dennisreimann this is precisely what 6a28497 should have solved, which is only on master, not 1.2.4

dennisreimann:
@NicolasDorier I just pulled the current master and for me it doesn't work in Safari 15 (it doesn't work in Firtefox either though). On page load you get this message in Safari 15:

> The source list for Content Security Policy directive 'script-src' contains an invalid source: ''unsafe-hashes''. It will be ignored.

When trying to switch the time format the JS error console says:

> Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy.

> But frankly, I don't like not using inline at all, this is too much of a burden.

I thinks there are cases where inline handlers are the most concise way. There are also cases where we need the script block to defined the functions that get called anyways – here I think having everything defined per script is the better option. Using the `delegate` function this becomes concise too imho, because we can just have the callback instead of defining a separate named function that gets called via inline handler.

NicolasDorier:
that suck :(

BTCPayServer/wwwroot/checkout/js/checkout.js
```diff
@@ -0,0 +1,4 @@
+delegate('click', '.payment-method', function(e) {
```
NicolasDorier, line 1: this script isn't referenced anywhere?

BTCPayServer/Views/Wallets/SignWithSeed.cshtml
```diff
@@ -5,14 +5,21 @@
...
     </a>
     <a asp-action="WalletSend" asp-route-walletId="@Context.GetRouteValue("walletId")" class="cancel">
         <vc:icon symbol="close" />
     </a>
 }
 
+@section PageFootContent
+{
+    <script>
+        delegate('click', '#GoBack', () => { history.back(); return false; })
```
NicolasDorier, line 19: does this syntax well supported by brother? `() =>`

BTCPayServer/wwwroot/checkout/js/checkout.js
dennisreimann: It's defined in [utils.js](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/main/utils.js), which is included in the main and checkout [bundles](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/bundleconfig.json).

BTCPayServer/Views/Wallets/SignWithSeed.cshtml
dennisreimann: The ES6 standard is well supported by now. I use occasions like this to try to modernize the JS we have and potentially to get rid of jQuery.

PR #2949, opened 2021-10-05
Allow User to delete own account

Kukks:


satwo:
Tested ACK.

For extra security, maybe it would make sense to require a password before the account is deleted.

NicolasDorier:
@Kukks I think there is not a new way to make delete confirmation via JS @dennisreimann

PR #2936, opened 2021-10-01
Greenfield: Provide negative undue when overpaid.

Kukks:
closes #2935

PR #2951, opened 2021-10-06
Invoice's due can be negative, fix Amount field of invoice

NicolasDorier:
Fix https://github.com/btcpayserver/btcpayserver/discussions/2900


Kukks:
Isnt this the same as https://github.com/btcpayserver/btcpayserver/pull/2936

Issue #2110, opened 2020-12-07
Uppercasing BECH32 address and schema in QR code


rockstardev:
[Uppercase QR codes are easier to scan](https://github.com/btcpayserver/btcpayserver/pull/2060#issuecomment-723828348), so we should make the effort to uppercase whatever characters we can in the URI.

Since [Bech32 addresses are already uppercased by default in Bitcoin Core](https://github.com/bitcoin/bitcoin/pull/15371) I expect that shouldn't be a problem. Upper-casing the schema could be a different beast... but so far in my testing it seems that only CashApp implemented this in a wrong fashion. Also, before I changed it last year, BtcPayServer was uppercasing schema component of QR Code URI.

To make sure we're not creating problems for merchants, this time I would prefer we first test which wallets support different format and then decide whether:

1. We will uppercase BECH32
2. We will uppercase schema
3. We will do both

## List of tested wallets

Here is the list of wallets tested so far, feel free to contribute by testing with wallets you have.

Wallet Name | BITCOIN:BECH32 | BITCOIN:bech32 | bitcoin:BECH32 | bitcoin:bech32 | Validated By
-- | -- | -- | -- | -- | --
Atomic Wallet | ✔️ | ✔️ | ✔️ | ✔️ | @brijananand
BitBoxApp | ✔️ | ✔️ | ✔️ | ✔️ | @Stadicus
BitBoxApp for Android | ✔️ | ✔️ | ✔️ | ✔️ | @Stadicus
Bitcoin Wallet Android (Schildbach) | ✔️ | ✔️ | ✔️ | ✔️ | @freerko
Blockchain Info | ❌ | ❌ | ❌ | ❌ | @pavlenex
Blockstream Green | ✔️ | ✔️ | ✔️ | ✔️ | @britttttk
BlueWallet | ✔️ | ✔️ | ✔️ | ✔️ | @britttttk
BLW Android | ✔️ | ✔️ | ✔️ | ✔️ | @anitaposch
BRD| ❌| ❌ | ✔️ | ✔️ | @pavlenex 
Breze | ✔️ | ✔️ | ✔️ | ✔️ | @Kixunil @brijananand
BTCPay Server | ✔️ | ✔️ | ✔️ | ✔️ | @pavlenex
Casa | ✔️ | ✔️ | ✔️ | ✔️ | @jlopp 
CashApp | ✔️ | ✔️ | ✔️ | ✔️ | @britttttk
[...]

PR #2952, opened 2021-10-06
Chivo mode for the checkout

NicolasDorier:


NicolasDorier:
Chivo support status:

BITCOIN:BECH32  :  N
BITCOIN:bech32  :  N
bitcoin:BECH32  :  N
bitcoin:bech32  :  Y

For lightning, it only support lowercase, without prefix.

pavlenex:
Concept nack, we shouldn't da na fix, Chivo should follow the BIP entirely. "Chivo mode" also sounds quite bad.

jquiros2:
Yes, Chivo should and name on tick could be changed.  At least for now we have the workaround but makes updates more difficult for inexperienced in git/docker.  A shame how sometimes in history govs fool with disguises as democracy.
Thanks for the impulse @NicolasDorier.

pavlenex:
@jquiros2 Athena Bitcoin CEO [mentioned](https://twitter.com/EGravengaard/status/1445608970022916102?s=20) they'll fix this by the way, hence why we're closing it.

jquiros2:
@pavlenex.  So koibanx.com are the ones doing the chivo wallet (algorand).  Athena is payment gateways and ATM's.  Really messed up implementation.  When you try to send BTC out of chivo, it mostly gets stuck, and discounted, then you need to call or email support and a few days later they reimburse with no explanation.  I've been trying to get some funds out of chivo.  I had sent to chivo as a test and for funding an lnd channel.  So far I 've tried three times and it keeps getting stuck).  Of course chivo is what people want to use b/c the gov gave $30 worth to ppl who use it.  Chivo-chivo if fine.  Tough for merchants who do not want to use chivo to accept payments.  Just giving the scoop as I think it is important for BTC community to know how it's going.  Hopefully will work soon!  Still nothing on chivo -https://github.com/btcpayserver/btcpayserver/issues/2110 .   By the way, absolutely FANTASTIC that we as a world have BTCPAY.  Thank you!  Now just tryong to sort through  the current lnd issues but that's something else.

NicolasDorier:
@jquiros2 mentioned that Chivo 2.0 seems to be better with QR Codes and made this mode useless. :)

PR #2958, opened 2021-10-08
Support Topup Invoices in Apps

Kukks:
Backwards compatible support for topup invoices in apps.

Slight difference in the option naming (yes => 'Minimum', no => 'Fixed', 'Topup')
image

image

image

image

### DIFF
--- a/BTCPayServer/Views/AppsPublic/ViewCrowdfund.cshtml
+++ b/BTCPayServer/Views/AppsPublic/ViewCrowdfund.cshtml
@@ -329,9 +329,11 @@
                             <template v-if="perk.price && perk.price.value">
                                 {{perk.price.value.noExponents()}}
                                 {{targetCurrency}}
-                                <template v-if="perk.custom === 'true'">or more</template>
+                                <template v-if="perk.price.type == 1">or more</template>
                             </template>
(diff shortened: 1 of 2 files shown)


pavlenex:
@Kukks when I click on the `view pos` I get this, I tried dotnet restore and clean multiple times, not sure if it's just me.

Screenshot 2021-10-08 at 22 51 16

NicolasDorier:
you should just use property in the model rather than  fit logic in the cshtml

Kukks:
> The wording related to this issue is being discussed [here](https://chat.btcpayserver.org/btcpayserver/pl/8t9r7qqcxigudm6etzw3mzq44w).
> 
> My opinions:
> 
> * My main issue is that `Topup` means nothing in all languages except English.
>   I'd simply rename it `Custom`, because that's what it does, enables the payee to pay a custom amount, without the merchant enforcing a minimum.

done

> * Additionally, in your Tea Shop screenshot, would it be possible to have the amount field **not** displayed, and just have the CTA displayed?
>   I feel like in the case of a topup item, that field is irrelevant and unnecessary.

For a topup item, only cta is displayed. If you use the "Custom payments" option (https://i.imgur.com/MxIPQSm.png), the textbox will still show but is optional now. This is to mitigate a breaking change where every pos using it will now effectively lose lightning support (lnurl payment method slightly fixes this but is not as widely supported)

Issue #2963, opened 2021-10-10
v1.2.4.0 Default payment ignored


BTCBellyButton:
**Describe the bug**
Below $5 presents only Lightning network payment method  ->   Correct.
Above $150 presents only On chain payment method ->  Correct
Between $5 and $150 presents the on chain payment method  ->  BUG

**To Reproduce the bug**
Set the checkout experience like this:

![Screenshot from 2021-10-09 19-55-21](https://user-images.githubusercontent.com/68573637/136681815-cdecb473-02c3-47f8-977c-366d21414db5.png)

**Expected behavior**

When the amount is between $5 and  $150 the default payment option should be Lightning.
Instead on chain payment is presented.
![Screenshot from 2021-10-09 23-25-55](https://user-images.githubusercontent.com/68573637/136682033-db8c64ed-84b9-41dc-a6f6-60aedc2e3137.png)

It worked fine until after the update to versionv1.2.4.0.

**Your BTCPay Environment (please complete the following information):**
- BTCPay Server Version:    1.2.4.0
- Deployment Method:          Manual
- Browser:                              Brave, Chrome,

satwo:
I reproduced this using the same settings you had. Interestingly, my settings for "Enable payment methods only..." are flipped: "BTC (On-Chain)" is above "BTC (Off-Chain).

@pavlenex , I'd be happy to try to solve this one.

pavlenex:
Feel free to @satwo 🚀

NicolasDorier:
We need to fix that before next release.

BTCBellyButton:
How can I test it?

PR #2975, opened 2021-10-16
Fix default payment bug

ghost:
This attempts to fix the default payment bug described by @BTCBellyButton in #2963. 

The problem seems to be that the "Default Payment Method on checkout" in CreateInvoice is not currently connected to the Store DPM. So, if a user has both `BTC (On-Chain)` and `BTC (Off-Chain)` enabled in CheckoutExperience, the top option on the Invoice DPM dropdown will always be `BTC (On-Chain)`, regardless of what the Store DPM is set to. 

The obvious solution is to assign the Store DPM to the `selectedValue` of the Invoice DPM SelectList, which is populated by `AvailablePaymentMethods`. The problem with that is that the property`SupportedTransactionCurrencies` _also_ uses `AvailablePaymentMethods` to populate its SelectList items and so this would cause a new set of problems.
 
Wanting to avoid any major refactoring or hacky client-side JS, I just made a new property `AvailablePaymentMethodsWithDefault`, and everything seems to work as it should. But it feels like there should be a slicker solution that doesn't require making a new property. Any feedback would be highly appreciated.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -21,6 +21,7 @@ using OpenQA.Selenium;
 using OpenQA.Selenium.Chrome;
 using OpenQA.Selenium.Support.Extensions;
 using Xunit;
+using OpenQA.Selenium.Support.UI;
 
 namespace BTCPayServer.Tests
 {
@@ -352,7 +353,7 @@ namespace BTCPayServer.Tests
             decimal? amount = 100, 
             string currency = "USD", 
             string refundEmail = "",
-            string defaultPaymentMethod = "BTC"
+            string defaultPaymentMethod = null
         )
         {
             GoToInvoices();
@@ -364,7 +365,8 @@ namespace BTCPayServer.Tests
             currencyEl.SendKeys(currency);
             Driver.FindElement(By.Id("BuyerEmail")).SendKeys(refundEmail);
             Driver.FindElement(By.Name("StoreId")).SendKeys(storeName);
-            Driver.FindElement(By.Name("DefaultPaymentMethod")).SendKeys(defaultPaymentMethod);
+            if (defaultPaymentMethod is string)
+                new SelectElement(Driver.FindElement(By.Name("DefaultPaymentMethod"))).SelectByValue(defaultPaymentMethod);
             Driver.FindElement(By.Id("Create")).Click();
 
             var statusElement = FindAlertMessage();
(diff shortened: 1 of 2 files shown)


BTCBellyButton:
Just a note, guys.
It worked perfectly before updating to the latest release. Unfortunately I can't recall which version I was running before updating. I skipped a few earlier updates...

Now it works as expected above and below the threshold I set. It correctly shows only one payment method available.
In between I'd expect it to default to off chain and to allow to switch to on chain. 

Also I tried to delete the two limits and I expected to have only 1 payment methods, the default OFF chain...

Screenshot from 2021-10-16 21-22-12

 but I get both.

Screenshot from 2021-10-16 21-24-53

also when saving this settings...

Screenshot from 2021-10-16 21-34-15

Screenshot from 2021-10-16 21-35-55

BTCBellyButton:
This is also weird.
With these illogical settings everything works as expected except that between 5 and 150 no payment is shown. The keypad turns to 0 and that's it... The settings don't make sense but shouldn't the default payment be shown in any case?

Screenshot from 2021-10-16 21-51-49

satwo:
@BTCBellyButton  Regarding the red error messages, those are because you didn't specify a currency (i.e., 'USD'). That error message could probably be improved for clarity. 

> This is also weird.
With these illogical settings everything works as expected except that between 5 and 150 no payment is shown. The keypad turns to 0 and that's it... The settings don't make sense but shouldn't the default payment be shown in any case?

I couldn't reproduce this - when I set those same illogical settings and try to create an invoice in between 5 and 150, I get an error message on the create invoice page.

BTCBellyButton:
@satwo I think I am running master.
I've been updating since the very beginning with this:

git pull origin master
sudo ./build.sh

But I never had a problem until this one...

Also I trusted the footer maybe a bit too much...   :-)

Screenshot from 2021-10-17 17-55-24

If you want me to try something else, let me know!

BTCBellyButton:
>  That error message could probably be improved for clarity.

That'd be useful. Maybe also a note reminding to add the currency. "Do not forget the currency (i.e. USD, EUR...)" or something.
I picked the values with USD from the browser drop down autofill list and I didn't even realize they currency was needed.

BTCBellyButton:
> when I set those same illogical settings and try to create an invoice in between 5 and 150, I get an error message on the create invoice page.

This is what I get in that case:

https://user-images.githubusercontent.com/68573637/137650272-cffcc5cc-8187-4368-8fd7-0ab3979e9ecd.mp4

NicolasDorier:
@satwo I think a better solution would be to have a new option called `Same as store` (or just an empty choice), which is selected by default.

I checked the logic for getting the default payment method in the checkout, it is broken. It should try to take the invoice's payment id, and if unavailable, use the store's. This is not what it does now, preparing a commit to add to yoru PR.

PR #2983, opened 2021-10-18
Add ability to accept tips in POS terminal

bolatovumar:
This PR adds ability to accept tips in POS terminal view. Only fixed percentage options can be selected (no custom tip amount input unlike the cart view). The rationale is that most likely this view will run on some sort of touch screen device where it's easier to just tap the fixed percentage amount rather than type the amount in.

See video:

https://user-images.githubusercontent.com/1934678/138578580-7f35bc34-5e28-425a-be30-98ef72334653.mov

Also I'm hiding/showing relevant sections for each type of app view in the "Update Point of Sale" page. Since not all customization options apply to all app views it makes sense to only show those which apply to the currently selected view type.

See video:

https://user-images.githubusercontent.com/1934678/138578636-dcd63542-3a15-4c2d-afb1-a606e67359fa.mov

### DIFF
--- a/BTCPayServer/wwwroot/light-pos/app.js
+++ b/BTCPayServer/wwwroot/light-pos/app.js
@@ -90,10 +90,10 @@ document.addEventListener("DOMContentLoaded",function (ev) {
             tipClicked: function(percentage) {
                 this.payTotalNumeric -= this.tipTotalNumeric;
                 this.tipTotalNumeric = parseFloat((this.payTotalNumeric * (percentage / 100)).toFixed(this.srvModel.currencyInfo.divisibility));
-                this.payTotalNumeric += this.tipTotalNumeric;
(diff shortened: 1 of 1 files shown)


bolatovumar:
> And it seem it's not available for he list view which I am not sure if it's intentional or not.

Yeah, that's intentional, this is only for POS terminal.

> Can't enter the custom amount for tips on Safari 15 (I assume that's why the draft PR 😄 )

I left that out on purpose because if you running this app on some sort of POS terminal thingie you probably don't have a keyboard or anything to enter a custom amount, you probably only have a touch screen so you will probably only want to tap a button. I left the input just so that it shows the total tip amount. We we think you should be able to enter a custom amount with a keyboard I can change that.


The PR is in draft mode for now because I still want to make changes to the admin view to show/hide relevant options when a certain app type is selected.

OriginalTexasToast:
Restaurants in the US need specific invoices for tips because they are taxed differently.  Is there a way to see/archive the tips separately from the BTCpay dashboard or in the invoices?

PR #2994, opened 2021-10-19
Improve supporters display

dennisreimann:
This brings some minor visual updates to the supporter logos. I needed to update the SVGs for btcpayserver/btcpayserver-doc#972, which is about improving the display in dark mode.

While I was at it also made it so that the supporter name does not wrap and the logos all have the same visual weight.

## Before

before

## After

after-dark

after-light


pavlenex:
Unfortunately we cannot alter the BG logo. :( It has to remain as is.

PR #3027, opened 2021-10-28
Display and update App Name in settings

ghost:
Currently the "Name" property is not exposed in an app's settings/update page, which in addition to leaving the user in the dark about which app is being updated, may also result in a general confusion between the `Title` and `Name` properties.

This PR: 

- Displays the app `Name` in the header of the update page
- Allows updating of the `Name` property
- Changes the display label for app `Title` to "Display Title" and `Name` to "App Name" in order to make them more distinct and specific for the end user


dstrukt:
Display Title is good too, just another idea that crossed my mind, no strong opinions.

Regardless, great work!

PR #3033, opened 2021-10-29
Add admin badge to user list

satwo:
Currently, one must click "Edit" on each individual user to see whether that user has admin privs.

This PR provides list-view visibility of admin privs by adding an `ADMIN` badge beside an admin user's email.

image

### DIFF
--- a/BTCPayServer/Controllers/ServerController.Users.cs
+++ b/BTCPayServer/Controllers/ServerController.Users.cs
@@ -63,6 +63,12 @@ namespace BTCPayServer.Controllers
                 })
                 .ToListAsync();
             model.Total = await usersQuery.CountAsync();
+
+            foreach (UsersViewModel.UserViewModel uvm in model.Users)
+            {
+                var userId = await _UserManager.FindByIdAsync(uvm.Id);
+                uvm.IsAdmin = await _userService.IsAdminUser(userId);;
+            }
              
             return View(model);
         }
--- a/BTCPayServer/Views/Server/ListUsers.cshtml
+++ b/BTCPayServer/Views/Server/ListUsers.cshtml
@@ -66,7 +66,7 @@
                 @user.Email
                 @if (user.IsAdmin)
                 {
-                    <span class="badge">ADMIN</span>
+                    <span class="badge bg-info">Admin</span>
                 }
             </td>
             <td>@user.Created?.ToBrowserDate()</td>
--- a/BTCPayServer/Views/Server/ListUsers.cshtml
+++ b/BTCPayServer/Views/Server/ListUsers.cshtml
@@ -62,8 +62,8 @@
     @foreach (var user in Model.Users)
     {
         <tr>
-            <td>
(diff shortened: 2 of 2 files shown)


BTCPayServer/Controllers/ServerController.Users.cs
```diff
@@ -63,6 +63,12 @@ public partial class ServerController
                 })
                 .ToListAsync();
             model.Total = await usersQuery.CountAsync();
+
+            foreach (UsersViewModel.UserViewModel uvm in model.Users)
+            {
+                var userId = await _UserManager.FindByIdAsync(uvm.Id);
+                uvm.IsAdmin = await _userService.IsAdminUser(userId);;
+            }
```
satwo, line 71: @NicolasDorier @rockstardev @Kukks  Instead of this foreach loop, any ideas on a more elegant way to assign the `IsAdmin` directly in the LINQ Select() query above? 

I tried for a bit but couldn't make anything work with all the async stuff going on.

NicolasDorier: In the Select clause, you should be able to reach the Roles of the user `u.`.

satwo: Tried for a while and coming up short...

Zaxounette (changes requested):
I'll just ping @dennisreimann and @dstrukt to see if they can make it a bit prettier UI wise.

Other than that, tACK the idea.

BTCPayServer/Views/Server/ListUsers.cshtml
```diff
@@ -62,7 +62,13 @@
     @foreach (var user in Model.Users)
     {
         <tr>
-            <td>@user.Email</td>
+            <td>
+                @user.Email
+                @if (user.IsAdmin)
+                {
+                    <span class="badge">ADMIN</span>
```
dennisreimann, line 69:
```suggestion
                    <span class="badge bg-info">Admin</span>
```

dstrukt: ACK.

Great addition @satwo!

satwo: Thanks @dstrukt !

pavlenex (approved):
tACK https://github.com/btcpayserver/btcpayserver/pull/3033/commits/4af6299ec443b9468ce45490b6e4289241f30721

dennisreimann (approved):

Kukks (approved):

NicolasDorier:
this shouldn't have been merged. We shouldn't have one query per line.

dennisreimann:
> We shouldn't have one query per line.

Agreed, can you fix this and show @satwo and me how it should be done? We couldn't make it work.

satwo:
This is an issue I've already come across twice actually. @dennisreimann I will hit the books to try to understand `async` better and try to find a better solution for this situation.

Issue #3000, opened 2021-10-20
UI: can't see to which URLs webhooks got sent


ndeet:
This is only a minor thing but maybe would be helpful for others too. On the invoice details page you can see the webhook events that got fired which is very nice. You can't see from the UI where that webhook was sent to (URL registered on creating the webhook). So you would need to check the webhook config to know. It get's even more confusing if you have registered multiple webhooks, you can check the exact payload but not where it was sent to.

![image](https://user-images.githubusercontent.com/1136761/138074043-1d83fc96-f585-4bde-be2a-393d52931697.png)

satwo:
Have a fix coming for this shortly.

satwo:
@ndeet Solved this with #3034 but there's a  [small edge case](https://github.com/btcpayserver/btcpayserver/pull/3034#issuecomment-955241265), maybe you could weigh in on it?

ndeet:
Fixed by #3034

PR #3034, opened 2021-10-30
List webhook URLs

satwo:
Addresses #3000.

I limited the width of the URL field and made the full URL visible with a tooltip. The UI could definitely use some sprucing up though. 

image

### DIFF
--- a/BTCPayServer/Models/StoreViewModels/EditWebhookViewModel.cs
+++ b/BTCPayServer/Models/StoreViewModels/EditWebhookViewModel.cs
@@ -24,6 +24,7 @@ namespace BTCPayServer.Models.StoreViewModels
             Time = s.Timestamp;
             Type = blob.ReadRequestAs<WebhookEvent>().Type;
             WebhookId = s.Id;
+            PayloadUrl = s.Webhook?.GetBlob().Url;
         }
         public string Id { get; set; }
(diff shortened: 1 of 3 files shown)


NicolasDorier:
@dstrukt @dennisreimann if you feel like to brush it up later! :) Merging meanwhile.

satwo:
@NicolasDorier I just noticed one small edge case, which is that if the user modifies the `payloadUrl` after already having used the webhook, any history for that webhook will incorrectly reflect the new URL change.

Original correct history:
Screen Shot 2021-10-30 at 10 05 37 AM

False history after user modifies payloadUrl:
Screen Shot 2021-10-30 at 10 05 07 AM

To make this list view a true historical view, we'd have to change around the data structures somewhat. For example, when updating a webhook that has already been used, create a new webhook record in the db. Or we can disable updating a webhook url altogether if is has been used. What do you think?

ndeet:
Hey @satwo thank you very much 🤗  Looks great to me and exactly what I wanted to have 👍  

I see the case you described as it will only pick up the currently configured url for the webhook. Imo this is fine and only a minor issue. Not sure if it is worth it changing data structures to have historical data. Locking the field and force users to recreate the webhook would also be fine for me but not sure which use cases there might be and maybe it would be a breaking change for the current api version.

PR #3045, opened 2021-11-01
Improves maintenance domain name form element

dstrukt:
More small fixes.

Updated domain form section to flex-box (as opposed to the "row/g-2" & "col-sm" pattern cc @dennisreimann curious to hear your thoughts on this, as we might have some consistencies throughout the app regardless).

The display error below was caused by a stray "mb-2" class, and we could simply adjust and remove that class, if flex-box isn't the path forward.

Screen Shot 2021-10-31 at 5 09 05 PM

Otherwise, just some spacing and consistency edits.

After:
Screen Shot 2021-10-31 at 5 50 01 PM
Screen Shot 2021-10-31 at 5 51 40 PM

### DIFF
--- a/BTCPayServer/Views/Server/Maintenance.cshtml
+++ b/BTCPayServer/Views/Server/Maintenance.cshtml
@@ -11,12 +11,16 @@
             <h4 class="mb-2">Domain name</h4>
             <p class="text-secondary">You can change the domain name of your server by following <a href="https://docs.btcpayserver.org/Deployment/ChangeDomain/" target="_blank" rel="noreferrer noopener">this guide</a>.</pclass="text-secondary">
 
-            <div class="d-flex">
-                <input asp-for="DNSDomain" class="form-control flex-fill" disabled="@(Model.CanUseSSH ? null : "disabled")" />
-                <span asp-validation-for="DNSDomain" class="text-danger"></span>
-                <button name="command" type="submit" class="btn btn-primary ms-3" value="changedomain" title="Change domain" disabled="@(Model.CanUseSSH ? null : "disabled")">
-                    Confirm
-                </button>
+            <div class="row g-2">
+                <div class="col-sm">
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/Server/Maintenance.cshtml
```diff
@@ -8,38 +8,34 @@
 <form method="post">
     <div class="row mb-5">
         <div class="col-lg-6">
-            <h4 class="mb-3">Domain name</h4>
-            <p>You can change the domain name of your server by following <a href="https://docs.btcpayserver.org/Deployment/ChangeDomain/" target="_blank" rel="noreferrer noopener">this guide</a>.</p>
+            <h4 class="mb-2">Domain name</h4>
+            <p class="text-secondary">You can change the domain name of your server by following <a href="https://docs.btcpayserver.org/Deployment/ChangeDomain/" target="_blank" rel="noreferrer noopener">this guide</a>.</pclass="text-secondary">
```
dennisreimann, line 12: I'd rather remove the `text-secondary` class for the other paragraphs on that page, because we don't use them like that elsewhere. Ok?

dstrukt: Sorry, maybe it's just late and I'm misunderstanding, are you saying we should remove the text-secondary class specifically for this snippet of text?

If so, that's fine with me, I'll make the change in the morning!

dennisreimann: Yes, I'd remove it here and from the other paragraphs on that page as well.

dstrukt:
@dennisreimann Latest commit addresses your feedback!

Screenshot:
Screen Shot 2021-11-01 at 12 40 34 PM

dennisreimann (approved):
Nice, that's what I meant! Thank you 👍

PR #3046, opened 2021-11-01
Add setup links for LN URL integration

bolatovumar:
Currently if user doesn't have LN set up or if they do but don't have the LN URL option enabled they have to manually navigate to payment methods and enable LN if it's disabled or go to LN settings and enable LN URL.

This PR adds two setup links:

- If LN is not set up at all user will get a link to set up LN first (see screenshot below)
- If LN is set up but LN URL option is not turned on yet user will get a link that takes them exactly to the LN settings page and even to the exact heading (in case use is on a mobile phone so they don't have to swipe up)

Capture

### DIFF
--- a/BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
+++ b/BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
@@ -1,6 +1,6 @@
 @using BTCPayServer.Payments.Lightning
 @using BTCPayServer.Payments
-@using NBitcoin;
+@using NBitcoin
 @inject BTCPayNetworkProvider BTCPayNetworkProvider
 @{
     var store = Context.GetStoreData();
--- a/BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
+++ b/BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
@@ -1,35 +1,9 @@
 @using BTCPayServer.Payments.Lightning
-@using BTCPayServer.Payments
-@using NBitcoin
 @inject BTCPayNetworkProvider BTCPayNetworkProvider
 @{
     var store = Context.GetStoreData();
-    var lightningByCryptoCode = store
-            .GetSupportedPaymentMethods(BTCPayNetworkProvider)
-            .OfType<LightningSupportedPaymentMethod>()
-            .Where(method => method.PaymentId.PaymentType == LightningPaymentType.Instance)
-            .ToDictionary(c => c.CryptoCode.ToUpperInvariant());
-    var excludeFilters = store.GetStoreBlob().GetExcludedPaymentMethods();
-
-    var isLightningEnabled = false;
-    var isLNUrlEnabled = false;
-    
-    var paymentMethods = store.GetSupportedPaymentMethods(BTCPayNetworkProvider);
-    foreach (var paymentMethod in paymentMethods)
-    {
-        var paymentMethodId = paymentMethod.PaymentId;
-        switch (paymentMethodId.PaymentType)
-        {
-            case LNURLPayPaymentType lnurlPayPaymentType:
-                isLNUrlEnabled = !excludeFilters.Match(paymentMethodId);
-                break;
-            case LightningPaymentType _:
-                var lightning = lightningByCryptoCode.TryGet(paymentMethodId.CryptoCode);
-                isLightningEnabled = !excludeFilters.Match(paymentMethodId) && lightning != null;
-                break;
-        }
-    }
-
+    var isLightningEnabled = store.IsLightningEnabled(BTCPayNetworkProvider);
+    var isLNUrlEnabled = store.IsLNUrlEnabled(BTCPayNetworkProvider);
     var possible =
         isLightningEnabled &&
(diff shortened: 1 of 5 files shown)


BTCPayServer/Views/Stores/LightningSettings.cshtml
```diff
@@ -54,7 +54,7 @@
                         </p>
                     </div>
                     
-                    <h4 class="mt-5 mb-3">LNURL</h4>
+                    <h4 class="mt-5 mb-3" id="ln-url">LNURL</h4>
```
bolatovumar, line 57: This is so that we can link directly to the heading instead of just the whole page.

dstrukt:
ACK, great improvement for the existing UX.

I think you're going to like the new store centric UI changes that will improve the entire Wallet / Lightning configuration process.

satwo:
image

tACK, nice idea. Been meaning to do something similar in my `print` view improvement PR so I might just piggy back on this.

Had a couple of small nits about the wording/style, but with the changes coming that @dstrukt mentioned might as well leave it be.

dstrukt:
@satwo Feel free to roll with the nits, even though I think the UI will the next major release focus, it still could be a while.. so I'd go for it :)

BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
```diff
@@ -1,8 +1,39 @@
...
+                var lightning = lightningByCryptoCode.TryGet(paymentMethodId.CryptoCode);
+                isLightningEnabled = !excludeFilters.Match(paymentMethodId) && lightning != null;
+                break;
+        }
+    }
+
+    var possible =
+        isLightningEnabled &&
+        isLNUrlEnabled &&
+        store.GetSupportedPaymentMethods(BTCPayNetworkProvider).OfType<LNURLPaySupportedPaymentMethod>().Any(type => type.CryptoCode == "BTC");
```
dennisreimann, line 36: imho that's too much logic in the view and it should be moved to the controller/view model.

Other than that this PR is a good improvement!

satwo:
@dstrukt Sure:

- The wording could be more consistent with the relevant sections in `Payment methods`, i.e. "Enable" as the operative verb instead of "configure" or "setup". This could be especially helpful for those who don't read English very well.
- Consistent styling for these "You must enable/configure/setup [x]" links would be nice, in particular to make it obvious that they're clickable links rather than just info/warning messages. 

Just my two cents, nothing mission-critical.

BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
bolatovumar: @dennisreimann I agree! I will work on moving this code elsewhere.

dstrukt:
> @dstrukt Sure:
> 
> * The wording could be more consistent with the relevant sections in `Payment methods`, i.e. "Enable" as the operative verb instead of "configure" or "setup". This could be especially helpful for those who don't read English very well.
> * Consistent styling for these "You must enable/configure/setup [x]" links would be nice, in particular to make it obvious that they're clickable links rather than just info/warning messages.
> 
> Just my two cents, nothing mission-critical.

Great points.

- I think the "Setup" wording was copied from the Wallet CTA "Setup", which makes more sense in that context, but I'd agree "Enable" or "Configure" makes more sense in this context, however I don't have a strong opinion on which would be better..  Enable might be my preference?

- I think this is another area where this component styling was copied from the Wallet/Lightning setup .. which we've outgrown, and at the time, didn't realize this would be utilized so broadly. However, I agree, the styling could use a bit of love and consistency .. but unlike the wording, I'd opt to make another PR where we address this type of component / other similar components.

satwo:
Agree with all of that - I hadn't noticed that "Setup" is used for the initial LN activation, while "Enable" to toggle it on and off. "Enable" is also used for initializing and toggling LNURL.

In any case, we can open a separate PR for this - I'm re-using some of this code in #3050 and I imagine we'll have more of these "Setup/Configure/Enable [x]" redirects in the future, so it would make sense to have consistency throughout the app.

BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
```diff
@@ -10191,6 +10191,7 @@ input[type=number].hide-number-spin {
 }
 .btcpay-status {
   display: inline-flex;
+  flex-shrink: 0;
```
bolatovumar, line 10194: Added this to make sure the badge doesn't shrink in tight spaces (I saw it happen).

BTCPayServer/Views/Shared/LNURL/LightningAddressOption.cshtml
bolatovumar: @dennisreimann alright, done here: c6d09a8e0a0ee6e1f24c46c2cd8e01a6bc1afdd6

dennisreimann (approved):
LGTM

pavlenex (approved):
tACK https://github.com/btcpayserver/btcpayserver/pull/3046/commits/95cb821145dc10f2627879704637ba3d21991bf4

Kukks (approved):

PR #3050, opened 2021-11-02
Print view improvements

satwo:
WIP. EDIT: Ready for review.
 
- Adds a print button
- Modifies the warning message on the `Print` PoS view to include a link to the proper settings page (h/t @bolatovumar)
- Adds the css class `d-print-none` to hide some potentially undesired elements for the printed page (Store title, store description, and item inventory probably aren't appropriate for a printed page displayed in a storefront*). This also simplifies the formatting of the printed page.

Despite the last bullet point, I'm still running into trouble with the formatting of the bootstrap `card` elements on the page-to-print:

image

I also haven't figured out a way to prevent the items from going over the page break. I'm using the default 8.5x11 inch ANSI letter, but I assume we'd have similar issues with A4.

Paging @dstrukt @dennisreimann, any ideas?

*It would be nice to hear some feedback on this from users who plan to use the `print` PoS view, i.e.: what item elements do they need on the printed page? Presumably for formatting purposes, the fewer, the better (but again I'll defer to the CSS experts on this one).

### DIFF
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -536,3 +536,15 @@ header.masthead .header-content .header-content-inner p {
     height: 1rem;
     padding: 0 0.5rem 0 1rem;
 }
+
+
+@media print {
+    @page {
+        margin-top: 0;
+        margin-bottom: 0;
+    }
+    body {
+        padding-top: 72px;
+        padding-bottom: 72px ;
+    }
+}
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -537,7 +537,9 @@ header.masthead .header-content .header-content-inner p {
     padding: 0 0.5rem 0 1rem;
 }
 
-
+/* This hides unwanted metadata such as url, date, etc from appearing on a printed page. 
+The purpose is to clean the Print view of a PoS app, but it may cut off the headers and footers
+if someone tries to print a different page in BTCPay Server. */
 @media print {
     @page {
         margin-top: 0;
--- a/BTCPayServer/wwwroot/main/site.css
+++ b/BTCPayServer/wwwroot/main/site.css
@@ -536,17 +536,3 @@ header.masthead .header-content .header-content-inner p {
     height: 1rem;
     padding: 0 0.5rem 0 1rem;
 }
-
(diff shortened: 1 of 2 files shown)


BTCPayServer/wwwroot/main/site.css
```diff
@@ -536,3 +536,15 @@ header.masthead .header-content .header-content-inner p {
...
+@media print {
+    @page {
+        margin-top: 0;
+        margin-bottom: 0;
+    }
+    body {
+        padding-top: 72px;
+        padding-bottom: 72px ;
+    }
+}
```
satwo, line 550: Proof-of-concept for hiding URL, datetime, and other metadata which some browsers add to a printed page. It probably shouldn't be here in site.css (or should be done differently to not affect all of BTCPay Server) - maybe the front-end folks can help find a better solution for this, too?

dstrukt:
Will take a look and offer some suggestions, if I have any, when I get a moment later today.

dennisreimann:
> I also haven't figured out a way to prevent the items from going over the page break.

Have you tried something like this?

```css
@media (print) {
  .card {
    page-break-inside: avoid;
  }
}
```

satwo:
@dennisreimann Tested this with a simplied UI, no luck:

image

dennisreimann:
Ok, I tried various things -- first conclusion: Chrome sucks in that regard and Firefox does a better job.

As we are using a separate template for the print view anyways, we can strip it of everything fancy for displaying the cards and solely focus on a good print outcome. Chrome has mainly issues with grid and flex display in that regard, so we need to unset this and align the cards the old school way. This worked well enough for me and I'd throw all of these one-off solutions directly into this particular view:

```html
<style>
    .card-deck { display: unset; }
    .card { page-break-inside: avoid; width: 300px; margin: 0 1em 1em 0; float: left; }
</style>
```

satwo:
@dennisreimann Nice! That seemed to do the trick:

image

Still could use some tweaks - the non-print view of the `print` page is now slightly off-center, but I think that's a fine compromise to ensure the actual print functionality works properly. We can always improve it later, but at least now it should be usable for the intended purpose.

dennisreimann (approved):

dennisreimann:
tACK. Streamlined it a bit more with some minor tweaks.

pos-print

PR #3052, opened 2021-11-02
Bugfix #3049: Price type of edit item in PoS shows only "fixed"

### LINKED ISSUE #3049: Edit item in PoS only shows "Fixed" price type selected
**Describe the bug**
After loading the Update PoS Settings page and selecting an item to edit, it will always show the price type selected as `Fixed` regardless of what the actual price type is.

**To Reproduce the bug**
1. Go to Update Point of Sale
2. Edit an item -> change the price type to `Minimum` or `Custom`
3. Save the item, then save settings
4. Go back to edit the same item. It will again show `Fixed` as the price type (although it is the correct price type in the raw editor)

**Expected behavior**

The selected price type should be the same one you selected previously.

- BTCPay Server Version:    1.3.1.0
- Deployment Method:   Docker/development
- Browser:  Brave

**Additional context**

This seems to be only a front-end issue, as the database reflects the change and everything else works correctly.

fabu21:
The loop through all properties caused the problem. The "else" case was alway the last and so was the variable "custom" always "fixed". 
Removed the else case and added a default value.
Fixes #3049

### DIFF
--- a/BTCPayServer/Views/Apps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Apps/TemplateEditor.cshtml
@@ -210,8 +210,6 @@ document.addEventListener("DOMContentLoaded", function () {
                         }
                         if (productProperty.indexOf('price_type:') !== -1) {
                             custom = productProperty.replace('price_type:', '').trim();
-                        }else{
-                            custom = "fixed";
                         }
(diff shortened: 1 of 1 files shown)


satwo:
tACK, nice fix!

Kukks (approved):

PR #3055, opened 2021-11-03
Add View Style to Apps list

satwo:
When using/debugging PoS apps and needing to test a particular view style, I can never remember which app is set to which view style. 

This gives sortable, ListApps-level visibility to the view style of PoS apps. If Crowdfund (or a future app type) eventually has multiple view styles, this can be extended to incorporate those, too.

image

### DIFF
--- a/BTCPayServer/Controllers/AppsController.cs
+++ b/BTCPayServer/Controllers/AppsController.cs
@@ -73,8 +73,6 @@ namespace BTCPayServer.Controllers
                                 return app.StoreName;
                             case nameof(app.AppType):
                                 return app.AppType;
-                            case nameof(app.ViewStyle):
-                                return app.ViewStyle;
                             default:
                                 return app.Id;
                         }
--- a/BTCPayServer/Controllers/AppsController.cs
+++ b/BTCPayServer/Controllers/AppsController.cs
@@ -73,6 +73,8 @@ namespace BTCPayServer.Controllers
                                 return app.StoreName;
                             case nameof(app.AppType):
                                 return app.AppType;
+                            case nameof(app.ViewStyle):
+                                return app.ViewStyle;
                             default:
                                 return app.Id;
                         }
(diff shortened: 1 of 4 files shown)


dstrukt:
Ooo, interesting... I've been considering what more relevant information we could potentially surface on this screen as we restructure the front-end, one of my questions for this week's call if we manage to get that far down the list..

Even though it may be a while, and not aware of any new apps in the pipeline... my concerns are for the other types of apps, as a view style is specific to the POS, and leaves a blank space.

However I think being able to surface this information is helpful, because we could ..

1. Alternatively group & segment based on the types of app.

POS:
a
b
c
CF
a
b
etc..

2. Display icons based on the view type .. e.g.
Screen Shot 2021-11-02 at 11 34 54 PM

3. I suppose we could just fill that space with "Default" for the other types of apps, in the View Style column, but it feels like we can be more creative than this...

Worth more discussion imo, but I like where your heads at.

Thoughts?

dennisreimann:
I like what's proposed in your mockup @dstrukt and would vote for adding that info in the subline of the app. It makes more sense that way, because as you said it only applies to the POS app and making it a separate (sortable) column doesn't seem like the best solution imho.

satwo:
I agree @dstrukt @dennisreimann  that a column for view style isn't a pretty solution due to the blank space left for non-PoS apps. I had been going back and forth about it while developing it but decided to throw it out there to see what people would think. Putting it in the subline is much better, and I love the display icons idea too!

For the current UI, what about something like this instead?
image

I don't know whether there are any users managing dozens or hundreds of apps yet, but if/when that happens, I imagine being able to sort/filter on the view style would be helpful to them.

dennisreimann:
@satwo yes, that works as a solution until we rework that list.

satwo:
Ok, updated. The only slight issue is that sorting on App Type will not take into account the appended View Style. But I don't think that's really an issue right now; more important is the visibility of this information on the list.

pavlenex (approved):
tACK 

https://github.com/btcpayserver/btcpayserver/pull/3055/commits/642725694374b2952ac40df8bad4a96e6c2ede19

Kukks (approved):

PR #3060, opened 2021-11-03
Crowdfund: Fix perk value display

dennisreimann:
Previously this didn't take into account the target currency. This leads to wrong values if the target currency is SATS and the invoice currency is BTC. This updates the value to be calculated with the rate and also takes into account the network fee. (Basically adjusting it to how the contribution values are calculated)


pavlenex:
I wonder if this can fix issues I was having with https://github.com/btcpayserver/btcpayserver/pull/2777

dennisreimann:
@pavlenex Unfortunately not, but having a solution for that issue would improve this case as well. As it is right now, contributions that do not match the target currency will not be displayed in the perk contribution values as well. (Can be tested by changing the CF target currency after payments have been made)

PR #3068, opened 2021-11-04
Fix maintenance view

dennisreimann:
Something went wrong there :)


dstrukt:
Ooh, nice catch.

LGTM.

Issue #3002, opened 2021-10-21
Store-centric UI Navigation


pavlenex:
**Describe the feature**
For a long time, BTCPay Server navigation has been a top menu which distinguished some of the store-centric concepts, misleading users on the hierarch in BTCPay Server.

For the past couple of weeks, the concept of store-centric navigation has been proposed by @dstrukt presented on several design meetings.

This issue servers as a progress tracker for implementation of the **store-centric navigation**.

![General](https://user-images.githubusercontent.com/36959754/138273369-748803e8-f9e7-4233-afec-5381d5d4aa2e.png)

The new navigation abstracts payment methods (btc, ln, etc) and provides good connection to concepts.

This will be mainly a front-end change, but would require re-structure of the documentation, videos, etc, but hopefully will lead towards easier navigation for users by categorizing concepts properly.

This feature has received a from core-contributors and contributors, however your comments and feedback are welcome. Further enhacement to navigation will be done in #1711 

Figma progress can be seen here https://www.figma.com/file/lE9dYRMqoABkJrn0VdXTE4/?node-id=89%3A29205
Figma towards concept here https://www.figma.com/file/lE9dYRMqoABkJrn0VdXTE4/App?node-id=89%3A29206

pavlenex:
For transparency and easier, here's where we're atm with store-centric UI.

![Store Settings - General](https://user-images.githubusercontent.com/36959754/141506962-59c694a3-c9df-4618-bb0b-ec74fcb020ce.png)
![Store Settings - Rates (1)](https://user-images.githubusercontent.com/36959754/141507005-252f8dfa-ae5e-4b83-b620-6d4e1aff84e4.png)

Issue #2509, opened 2021-05-06
Enable selection, RBF, CPFP and other actions in transaction view


pavlenex:
Currently, we don't have a UI for users to select certain transactions and perform an action. 
In a scenario when the user may have a stuck transaction (unconfirmed) they may want to select it and do a CPFP or RBF.

We do something similar for the invoices, so we should try to enable it for wallet > transactions view.

![Screenshot 2021-05-06 at 11 28 49](https://user-images.githubusercontent.com/36959754/117276510-43598b00-ae5f-11eb-9cf6-ee53e7ca5416.png)

On top, we probably should distinguish better between confirmed and unconfirmed.

NicolasDorier:
I love it, that's perfect UI for CPFP

PR #3072, opened 2021-11-04
Wallet settings merge

### LINKED ISSUE #2206: Manage wallet > Settings > Account key > Scan QR is switching and it's unclear what it does
**Describe the bug**
In `Manage wallet > Settings > Account key` > `Scan QR ` there's a little icon and once clicked it shows 2 QR codes switching between them quickly.
1. The quick switching makes it hard to scan
2. It is not really clear if you need to scan 1 or 2
3. It is not clear what this option actually does

**To Reproduce the bug**

Make sure you have a wallet created.

1. Go to `Manage wallet > Settings > Account key`
2. Click on QR icon
3. See error

**Expected behaviour**
First of all it should be clear what this does, it's unclear. Second it probably shouldn't be next to the header.
Second, I am not sure how this is supposed to work but auto-switching does not seem like a desired behaviour.

**Screenshots**

https://user-images.githubusercontent.com/36959754/104845306-f57f3e00-58d4-11eb-83ac-7a57e1938a7b.mp4

- BTCPay Server Version:   1.0.6.6 and 1.0.6.7 
- Deployment Method:  Docker      
- Browser:  Firefox

### LINKED ISSUE #2626: Modify Wallet page and Wallet Setting page unification
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/2276

<sup>Originally posted by **pavlenex** February 11, 2021</sup>
With #2164 merged, we've introduced a new Modify Wallet page which in essence is very similar to the Wallet Settings page. However, they do have different information.

I'm opening this discussion to discuss on how we can either distinguish them or unify those pages. Because, right now the flow is quite confusing.
- What is the difference between the two?
- What's the goal of having a modify wallet page?
- What's the goal of having wallet settings page?
- What kind of information are we trying to show with them to the user?

Screenshot from 2021-02-11 13-20-31

Screenshot from 2021-02-11 13-20-49

Ping @dennisreimann @dstrukt @NicolasDorier </div>

dennisreimann:
Merges both wallet settings screen from the wallets and the store section. 

Still need to fix the tests, posting this including screenshot so that we can already discuss the current state.

Closes #2626. Closes #2206.

## Wallet Transactions

The Settings link here now links directly to the wallet settings in the store section (see below).

Moved the "Other actions" here as a temporary solution, which will be cleaned up when we rework the wallet screens. 
These now only include the actions related to the wallet transactions -- they didn't really fit into the previous settings screen anyways. 

wallet-transactions

## Wallet Onchain Settings

Brought the wallet related actions ("View seed" and "Register for payment links") here and merged the previous wallet settings (mostly derivation path and account keys) into this view.

wallet-settings


dstrukt:
As I mentioned on MM, would like to revisit this area of things .. have some additional thoughts WRT to organization, but are contingent on the left side bar implementation. 

Regardless, great work, the consolidation of these sections is beautiful to see!

ACK.

pavlenex:
To me, this looks like a good direction, really good to see the unification. There may be nits here and there on how to clean things up as Dstrukt mentioned, but so far, this seems to be going into proper direction, concept ACK, as usual great work @dennisreimann 🚀

pavlenex:
A side-note, in #2509 the intended UI direction that @NicolasDorier proposed is to add `actions` tab on the top (similar to the invoices page) where we can have multiple transaction selection (same as we do with invoices), that's why I think the `other actions` fits in well into the workflow. We should open a separate issue for improving the UI once #2509 gets implemented.

The only tricky thing once that is implemented , is that a user may be confused with `other actions`l being able to do multiple selections on the transactions, where in fact these are `global` settings, not per transactions.

dennisreimann:
@pavlenex As I said the Other Actions dropdown for now is just a temporary solution until I get to redo the wallet screens in #3002 - or better yet a separate issue.

pavlenex (approved):
tACK. Long term we probably should remove the `settings` link from the wallet page.

I think this resolves both issues. As dennis said there is a bit of cleanup we'll have to do, but this is good first step.

@NicolasDorier let us know what you think about unifing this flow.

NicolasDorier:
Cool! Should the `Other Actions` from the transaction list instead be on the top?
Because if there is lot's of transaction, then the user would need to scroll to bottom to see the actions.

satwo:
tACk, nice work!

dennisreimann:
@NicolasDorier tweaked it, so that the Actions now appear besides the headline.

image

dennisreimann:
Rebased and ready to rock.

PR #3083, opened 2021-11-11
Support LNURL Auth

### LINKED ISSUE #3061: Add LNURL-Auth support
**Describe the feature**
In #2504 we discussed the ability to implement the LNURL-Auth which would allow people to authenticate with their BTCPay passwordless, just by scanning from their wallet (mobile/btcpay hosted, etc). 

An example of such UX can be found here http://lnmarkets.com
And this is explanation on how LUNRUL-Auth works https://⚡.bigsun.xyz/lnurl-auth.html

Kukks:
closes #3061


pavlenex:
Concept ack but wasn't able to test it on regtest due to localhost. My wallet recognized it though.

I like how it's nested under devices since fido2 and lnurl are quite similar.

Screenshot 2021-11-15 at 21 32 54
Screenshot 2021-11-15 at 21 33 01

@Kukks weird thing is that it works without me having a store/LN/LNURL enabled as a payment method. I assume that's normal?

Kukks:
> Concept ack but wasn't able to test it on regtest due to localhost. My wallet recognized it though.
> 
> I like how it's nested under devices since fido2 and lnurl are quite similar.
> 
> Screenshot 2021-11-15 at 21 32 54 Screenshot 2021-11-15 at 21 33 01
> 
> @Kukks weird thing is that it works without me having a store/LN/LNURL enabled as a payment method. I assume that's normal?

Yup correct. LNURLAuth does not require you to set it up inside a store and works independently. Ironically, a non lightning wallet can implement this spec very easily (I've done so in the tests)

BTCPayServer/Views/Account/LoginWithLNURLAuth.cshtml
```diff
@@ -0,0 +1,72 @@
...
+    function check(){
+        const request = new XMLHttpRequest();    
+        request.onload  = function() {
+          if (request.readyState === 4 && request.status === 200) {
+            setTimeout(check, 1000);
+          } else if (request.readyState === 4 ){
+              document.getElementById("authform").submit();
+          }
+        }
+        request.open("GET", "@Url.Action("LoginCheck", "LNURLAuth", new {userId = Model.UserId})", true);
```
NicolasDorier, line 68:
```suggestion
        request.open("GET", @Safe.Json(Url.Action("LoginCheck", "LNURLAuth", new {userId = Model.UserId})), true);                    
```

BTCPayServer/Views/LNURLAuth/Create.cshtml
```diff
@@ -0,0 +1,64 @@
...
+    </div>
+</div>
+<script>
+    function check(){
+        const request = new XMLHttpRequest();    
+        request.onload  = function() {
+          if (request.readyState === 4 && request.status === 200) {
+            setTimeout(check, 1000);
+          } else if (request.readyState === 4 ){
+              window.location.href = "@Url.Action("RedirectToList", new {successMessage = "The lightning node will now act as a security device for your account"})";
```
NicolasDorier, line 56:
```suggestion
              window.location.href = @Safe.Json(@Url.Action("RedirectToList", new {successMessage = "The lightning node will now act as a security device for your account"}));
```

BTCPayServer/Controllers/AccountController.cs
```diff
@@ -208,6 +220,84 @@ private async Task<LoginWithFido2ViewModel> BuildFido2ViewModel(bool rememberMe,
...
+            if (user == null)
+            {
+                return NotFound();
+            }
+
+            var errorMessage = string.Empty;
+            try
+            {
+                var k1 = Encoders.Hex.DecodeData(viewModel.LNURLEndpoint.ParseQueryString().Get("k1"));
+                if (_lnurlAuthService.FinalLoginStore.TryGetValue(viewModel.UserId, out var storedk1) &&
```
NicolasDorier, line 269: this isn't thread safe, our `TryRemove` directly, don't call `TryGetValue`

NicolasDorier:
doesn't build

PR #3090, opened 2021-11-12
Remove Confirmed state in UI

### LINKED ISSUE #1789: Remove Confirmed state in UI
In #1760 , we were discussing removing the Completed state for the new API. If we do this, we will need to retain it for the BitPay API.

So my questions are:
* Should we remove Confirmed or Completed?  We currently have some features such as marking an invoice as complete
* How do we communicate to existing users who use the Bitpay API support and see a different state on the BTCPay UI to the one received through the API?
* Should we make the BitPay API just report that a payment is completed once it is confirmed by the store settings and remove the "IsCompleted" added logic?

satwo:
This replaces `paid` with `processing`, and both `completed` and `confirmed` with `settled` in the Invoice UI.

To avoid the risk of introducing a bug in the back-end Invoice logic, where most of the state-related logic is based around `InvoiceStatusLegacy` rather than the modern `InvoiceStatus`, I put most of the logic for this in the view. I'm not sure it's the cleanest/maintainable long-term solution, and so would love some feedback on this.

Still need to add a dismissible infobox to inform users of the change.

Closes #1789.

image

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -104,7 +104,7 @@ namespace BTCPayServer.Controllers
                 StoreLink = Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id }),
                 PaymentRequestLink = Url.Action(nameof(PaymentRequestController.ViewPaymentRequest), "PaymentRequest", new { id = invoice.Metadata.PaymentRequestId }),
                 Id = invoice.Id,
-                State = invoiceState.Status.ToString(),
+                State = invoiceState.Status.ToModernStatus().ToString(),
                 TransactionSpeed = invoice.SpeedPolicy == SpeedPolicy.HighSpeed ? "high" :
                                    invoice.SpeedPolicy == SpeedPolicy.MediumSpeed ? "medium" :
                                    invoice.SpeedPolicy == SpeedPolicy.LowMediumSpeed ? "low-medium" :
@@ -999,6 +999,5 @@ namespace BTCPayServer.Controllers
                 return result;
             }
         }
-
     }
 }
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -104,7 +104,7 @@ namespace BTCPayServer.Controllers
                 StoreLink = Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id }),
                 PaymentRequestLink = Url.Action(nameof(PaymentRequestController.ViewPaymentRequest), "PaymentRequest", new { id = invoice.Metadata.PaymentRequestId }),
                 Id = invoice.Id,
-                State = invoiceState.ToString(),
+                State = invoiceState.Status.ToModernStatus().ToString(),
                 TransactionSpeed = invoice.SpeedPolicy == SpeedPolicy.HighSpeed ? "high" :
                                    invoice.SpeedPolicy == SpeedPolicy.MediumSpeed ? "medium" :
                                    invoice.SpeedPolicy == SpeedPolicy.LowMediumSpeed ? "low-medium" :
@@ -128,7 +128,7 @@ namespace BTCPayServer.Controllers
                                     .Select(c => new Models.StoreViewModels.DeliveryViewModel(c))
                                     .ToList(),
                 CanMarkInvalid = invoiceState.CanMarkInvalid(),
-                CanMarkComplete = invoiceState.CanMarkComplete(),
+                CanMarkSettled = invoiceState.CanMarkComplete(),
             };
             model.Addresses = invoice.HistoricalAddresses.Select(h =>
                 new InvoiceDetailsModel.AddressModel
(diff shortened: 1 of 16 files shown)


BTCPayServer/Views/Invoice/Invoice.cshtml
```diff
@@ -65,7 +66,7 @@
                     }
                     else
                     {
-                        <button href="#" class="btn btn-secondary text-nowrap ms-2" data-bs-toggle="tooltip" title="You can only issue refunds on invoices with confirmed payments" disabled><span class="fa fa-undo"></span> Issue refund</button>
+                        <button href="#" class="btn btn-secondary text-nowrap ms-2" data-bs-toggle="tooltip" title="You can only issue refunds on invoices with settled payments" disabled><span class="fa fa-undo"></span> Issue refund</button>
```
satwo, line 69: Not sure about the wording change on this one.

pavlenex: Some random ideas, for example, `Only a settled invoice can be refunded.` Or if we want to be more contextual `Invoice needs to have a settled status to be refunded. Please wait for a transaction to confirm on a blockchain before attempting to refund it`

cc @Zaxounette

Zaxounette: > You can only refund invoices that have been settled

My wording proposal ^

satwo: Thanks for the feedback! Went with something like a combination of the two:

`You can only refund an invoice that has been settled. Please wait for the transaction to confirm on the blockchain before attempting to refund it.`

satwo:
Added the infobox:

image

Maybe someone more talented than I could pretty it up a bit. Also probably worth adding a "Don't show again" option to this, but I don't have the time at the moment.

BTCPayServer/Views/Invoice/Invoice.cshtml
satwo: Thanks for the feedback!

I went with something like a combination of the two:

`You can only refund an invoice that has been settled. Please wait for the transaction to confirm on the blockchain before attempting to refund it.`

BTCPayServer/Views/Invoice/ListInvoices.cshtml
```diff
@@ -197,6 +197,36 @@
             </a>
         </div>
 
+    <div class="alert alert-light alert-dismissible fade show" role="alert">
+        <h6><strong>Update v1.4.0:</strong> Invoice states have been updated to match the Greenfield API.</h6>
+        <ul>
```
dennisreimann, line 202:
```suggestion
        <ul class="list-unstyled">
```

dennisreimann:
> Also probably worth adding a "Don't show again" option to this (perhaps store/check a cookie with javascript?).

This would be the easiest option. If we need it as a persistent flag in our db, we could have it similar to the `storeBlob.Hints`

BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -110,7 +110,7 @@ public async Task<IActionResult> Invoice(string invoiceId)
                 StoreLink = Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id }),
                 PaymentRequestLink = Url.Action(nameof(PaymentRequestController.ViewPaymentRequest), "PaymentRequest", new { id = invoice.Metadata.PaymentRequestId }),
                 Id = invoice.Id,
-                State = invoiceState.ToString(),
```
NicolasDorier, line 113: This change will hide the additional status such as PaidLate, or Paid Partial

satwo: This was done order to avoid messing with the back-end logic (for reasons stated in the PR) and the additional status such as PaidPartial is accounted for in the modified View logic. 

This way of doing it also results in the casing being changed from how it was before. I'll take some more time to think of whether it can be done better.

satwo:
> If we need it as a persistent flag in our db, we could have it similar to the `storeBlob.Hints`

I like this idea more than introducing a new js library, but not sure where in the db this flag would belong. It ought to be at the user level, presumably. Possibly `Settings`?

dennisreimann:
@satwo If we want to solve it on the client-side a simple cookie would suffice. @kukks might advise on where to put this option if we want to solve it on the server-side, which I think would be best.

satwo:
Added a "Don't Show Again" button the infobox, which sets a flag in `PoliciesSettings` (I couldn't think of a better place to put this without modifying the db).

image

Should we put this same warning on the Invoice Details page, or will it suffice to stay on Invoice List?

Kukks:
> Added a "Don't Show Again" button the infobox, which sets a flag in `PoliciesSettings` (I couldn't think of a better place to put this without modifying the db).
> 
> image
> 
> Should we put this same warning on the Invoice Details page, or will it suffice to stay on Invoice List?

Shouldnt this be on a user level though?
Existing  users => flag  = true. Click on don't show = set to false.
New => flag = false

satwo:
> Shouldnt this be on a user level though? Existing users => flag = true. Click on don't show = set to false. New => flag = false

Good point, I guess right now it's at the Server level. User level makes more sense but I'm not sure where to put that - maybe add a new field to `AspNetUsers`?

Kukks:
Let's add a Blob column to the `ApplicationUser`  so that we can store this kind of metadata. We'll need to create a migration task so that existing users have this flag.

NicolasDorier:
@Kukks merging this for now, but I think we should have something more generic for presenting new features to existing users.

Issue #2619, opened 2021-06-24
[UI/UX] Empty State Improvements


pavlenex:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/2360

<div type='discussions-op-text'>

<sup>Originally posted by **dstrukt** March 14, 2021</sup>
I've gone through an audited the initial empty states UI and UX, and there are a few relatively simple things I think we can do to improve this experience. There are also a few inconsistencies I've notices as well.

So for almost all of our major tabs, you're meet with this for an empty state:

## [Store] - Basic 
<img width="1842" alt="Screen Shot 2021-03-14 at 10 46 43 AM" src="https://user-images.githubusercontent.com/6250771/111078437-998afd00-84b2-11eb-8995-84e48052032d.png">

For all screens, I'm proposing that we enable a slightly more cleaned up "empty/default state" view.

<img width="1045" alt="Screen Shot 2021-03-14 at 2 10 01 PM" src="https://user-images.githubusercontent.com/6250771/111084326-044a3180-84cf-11eb-9627-aea20c9feb1c.png">

I would like to carry out that visual pattern, and the above section is pretty straightforward ...but the following have a slightly more involved interaction pattern, that feels a bit frustrating.

You need a store to utilize a Wallet, Payment Request, Invoice, App .. pretty much everything.

I've provided an example of this (which is pretty consistent across all of these).

## [Apps] - More Complex

Before
<img width="1839" alt="Screen Shot 2021-03-14 at 11 03 32 AM" src="https://user-images.githubusercontent.com/6250771/111078910-f7204900-84b4-11eb-8766-118aa8dce0e8.png">
<img width="1830" alt="Screen Shot 2021-03-14 at 11 20 48 AM" src="https://user-images.githubusercontent.com/6250771/111079469-60a15700-84b7-11eb-80c7-6dc8dd7bf759.png">

This is pretty much the pattern for the other sections.

While this will probably only happen to a user once, and hopefully they'll create a store after being prompted, this UX pattern could be improved. The introduction of our dashboard will help with this as well, but in the meantime...

Proposal:

We implement a check if a store exists, if not, preemptively display a screen that provides a "Create Store" CTA and some explanation, eliminating the additional clicks / alert modals, etc... to determine this.

After
###### (with store)
[...]

PR #3092, opened 2021-11-13
Invoice View Improvements

dstrukt:
A lot has been discussed around this view, but there are some low hanging fruit that we've all agreed, in one form or another, that needs to be improved, so I thought I'd go ahead and make those simple changes.

Few reorders, sizing changes, label updates, etc.. to make things consistent with other views.

Before:
Screen Shot 2021-11-12 at 4 49 59 PM

After:
Screen Shot 2021-11-12 at 5 01 36 PM

-------------

1. I wonder what others think..should you be able to select LNURL from this list if it's not accessible? If not, I'm not 100% sure how to fix, might need some help from @dennisreimann on that one.
Screen Shot 2021-11-12 at 4 49 09 PM
Screen Shot 2021-11-12 at 4 14 58 PM

2. I'm not sure if these section labels make as much as sense as they could, but I think they're a step in the right direction, cc @pavlenex @Zaxounette.

------------

h/t @woutersamaey Imagine you'll have some thoughts, and will feed into your larger PR.

### DIFF
--- a/BTCPayServer/Models/InvoicingModels/CreateInvoiceModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/CreateInvoiceModel.cs
@@ -38,7 +38,7 @@ namespace BTCPayServer.Models.InvoicingModels
             get; set;
         }
 
-        [Display(Name = "Default payment method on checkout")]
+        [DisplayName("Default payment method on checkout")]
         public string DefaultPaymentMethod
         {
             get; set;
@@ -58,7 +58,7 @@ namespace BTCPayServer.Models.InvoicingModels
         }
 
         [Uri]
-        [DisplayName("Notification Url")]
+        [DisplayName("Notification URL")]
         public string NotificationUrl
         {
             get; set;
@@ -89,7 +89,7 @@ namespace BTCPayServer.Models.InvoicingModels
             get; set;
         }
 
-        [Display(Name = "Require Refund Email")]
+        [DisplayName("Require Refund Email")]
         public RequiresRefundEmail RequiresRefundEmail
         {
             get; set;
--- a/BTCPayServer/Models/WalletViewModels/PullPaymentsModel.cs
+++ b/BTCPayServer/Models/WalletViewModels/PullPaymentsModel.cs
@@ -49,6 +49,7 @@ namespace BTCPayServer.Models.WalletViewModels
         [Display(Name = "Custom CSS Code")]
         public string EmbeddedCSS { get; set; }
 
+        [Display(Name = "Payment Methods")]
         public IEnumerable<string> PaymentMethods { get; set; }
         public IEnumerable<SelectListItem> PaymentMethodItems { get; set; }
     }
--- a/BTCPayServer/Models/WalletViewModels/PullPaymentsModel.cs
+++ b/BTCPayServer/Models/WalletViewModels/PullPaymentsModel.cs
@@ -49,7 +49,6 @@ namespace BTCPayServer.Models.WalletViewModels
         [Display(Name = "Custom CSS Code")]
         public string EmbeddedCSS { get; set; }
 
-        [Display(Name = "Payment Methods")]
         public IEnumerable<string> PaymentMethods { get; set; }
         public IEnumerable<SelectListItem> PaymentMethodItems { get; set; }
     }
--- a/BTCPayServer/Views/Invoice/CreateInvoice.cshtml
+++ b/BTCPayServer/Views/Invoice/CreateInvoice.cshtml
@@ -82,7 +82,7 @@
                     <div class="form-group">
                         <label asp-for="DefaultPaymentMethod" class="form-label"></label>
                         <select asp-for="DefaultPaymentMethod" asp-items="Model.AvailablePaymentMethods" class="form-select">
-                            <option value="" selected>Store's default settings</option>
+                            <option value="" selected>Use the store’s default</option>
                         </select>
                         <span asp-validation-for="DefaultPaymentMethod" class="text-danger"></span>
                     </div>
--- a/BTCPayServer/Views/Invoice/CreateInvoice.cshtml
+++ b/BTCPayServer/Views/Invoice/CreateInvoice.cshtml
@@ -61,11 +61,6 @@
                         <input asp-for="ItemDesc" class="form-control" />
                         <span asp-validation-for="ItemDesc" class="text-danger"></span>
                     </div>
-                    <div class="form-group">
-                        <label asp-for="PosData" class="form-label"></label>
-                        <input asp-for="PosData" class="form-control" />
-                        <span asp-validation-for="PosData" class="text-danger"></span>
-                    </div>
                     <div class="form-group mb-4">
                         <label asp-for="SupportedTransactionCurrencies" class="form-label"></label>
                         @foreach (var item in Model.AvailablePaymentMethods)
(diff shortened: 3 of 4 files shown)


satwo:
Looks much better already! 

What about checkboxes instead of a selectlist for `Supported Transaction Currencies`? Selecting multiple options with cmd+click isn't such a great UX.

dstrukt:
Absolutely agree the multi-select component we use is pretty terrible, and been something on my list as well .. but not 100% sure I have the dev prowess to tackle that one, know how to iterate through the array on the View, but getting the Model to return the proper Information .. would be a good simple challenge though.

satwo:
Oh yeah, that can definitely have its own PR, I was just brainstorming general ideas for this page :)

Really glad to see the ball rolling on improving this view, it definitely needs it IMO.

tACK.

dennisreimann (approved):
Good job!

dennisreimann:
Made the payment methods list checkboxes, which fits better -- thanks @satwo for the suggestion!

image

> I wonder what others think..should you be able to select LNURL from this list if it's not accessible?

The invoice creation form is kinda tricky in that regard: The payment methods are tied to a store -- and hence the store is selectable in the form, we would have to reload the page whenever the store changes. I think we can enhance it and build an improved version once the store-centric changes are done or even meanwhile -- didn't do it here, because it'd involve a bit more changes.

woutersamaey:
I have a copy change suggestion.
Instead of “Store’s default settings” can’t we put “Use the store’s default payment method” or “Use the store’s default” (shorter version)?

It reads kinda weird as it is currently.

woutersamaey:
And i hate that POS Data field. What is that? If it’s an advanced option it should be hidden, or if not, there should be a tooltip / help icon.

dstrukt:
@dennisreimann Nice job! I was so close in my attempt .. thought more was necessary than actually was needed .. but was good to learn from your code!

> The invoice creation form is kinda tricky in that regard: The payment methods are tied to a store -- and hence the store is > selectable in the form, we would have to reload the page whenever the store changes. I think we can enhance it and build > an improved version once the store-centric changes are done or even meanwhile -- didn't do it here, because it'd involve a > bit more changes.

Sounds good.

@woutersamaey Agreed some of the copy could definitely be improved, and I think that's within the scope of this .. will take a stab at some of the copy per your suggestions, etc.. here in a bit to wrap up this PR.

Zaxounette:
> I have a copy change suggestion.
Instead of “Store’s default settings” can’t we put “Use the store’s default payment method” or “Use the store’s default” (shorter version)?
> It reads kinda weird as it is currently.

Agreed it could be imroved.

We could reuse the `Inherit from store settings` or we could improve on that as well while we're at it with something in the likes of `Inherit from store` (1) or just `Store default` (2).

dstrukt:
Outside of wording, which I like your suggestions @Zaxounette, I think the first thing it calls into question, for me, is .. when did I set these defaults .. and subsequently, how do I change them?

I like the concept, and I think it's the type of UX improvements that really push things to the next level, but can be equally frustrating if not well explained how the result was achieved.

satwo:
> And i hate that POS Data field. What is that? If it’s an advanced option it should be hidden, or if not, there should be a tooltip / help icon.
> 
> The 2 notification fields should perhaps also be hidden under an “advanced” section… this is very technical and more something you’d use with am API than through the GUI

Had the exact same thoughts.

dennisreimann:
Created the additional options and moved the POS data (with added explanation) and notification fields to that.

image

Also applied @woutersamaey's wording change to the default payment method field.

woutersamaey:
@dennisreimann thanks man!
To be 100% honenest, it still sounds a bit wierd because POS data is really just metadata? There is already a field for `orderId` and `itemDescription` so the extra explanation is suggesting a superfuous use case.

Maybe slightly use a different explanation?

To be 100% content, I'd prefer to see this field renamed to `MetaData`, but not sure if we could get consensus on that?

woutersamaey:
Nitpick, maybe uppercase the word `Url` to `URL` ?

dstrukt:
So damn quick @dennisreimann ... great work!

If we roll with MetaData, can it be Metadata?

+1 to "URL"

woutersamaey:
Actually in the invoice visible from the API there is already a field called `Metadata`, so if we were to apply it here, that would collide.
Maybe the POS Data is here for historic reasons? If that is the case, maybe it can be deleted and we use the new Metadata field for the same purpose?

A more experienced dev than me is going to have to weigh in I believe.

dennisreimann:
> Maybe the POS Data is here for historic reasons? 

AFAIK the `posData` property exists because of BitPay API compatibility. /cc @kukks

dstrukt:
Mistake. 

Wrong branch, getting tired, going to submit that PR and go to sleep.

pavlenex:
I like the cleanup, good job. One minor note is that it's weird that invoice form  is left-aligned, shouldn't it be centered? Or is this something we'll tackle in the future versions once we improve the #2619  empty states for consistency? Just want to make sure centering is planned in the future, that's all.


Screenshot 2021-11-15 at 21 39 03

pavlenex (changes requested):
@dstrukt `amount` is no longer a required field. Neiher is currency if the preferred currency is selected, we should reflect that in the design.

satwo:
A few more ideas:

-The checkbox list should default to all boxes checked. Right now, if you create an invoice and leave the payment methods as-is (assuming they are all enabled for the store), all three methods will be enabled in checkout, even though none are checked in Create Invoice.

-Instead of `Use the store's default` or `Inherit from store settings`, why not just pre-select the actual default, and append "_(Store default)_" or even just "_(default)_" to the selection? Therefore, if BTC On-Chain is the store default, the pre-selected option would be something like "BTC On-Chain _(default)_". I think this might be easier to understand and provide better/clearer information - the user can actually see what the store default is right there in the invoice, without having to go back to settings to check.

-`Supported Transaction Currencies` should only show store-enabled payment methods, but right now it shows all 3 possible methods regardless of whether any of them are disabled at the store level.

-When you change stores, the store-specific settings (defaults, available payment methods, etc) should change as well on the Create Invoice page.

I understand this is probably beyond the scope of this PR, and would be happy to open a new issue for it if others are in agreement.

dstrukt:
- Agree with all checked be default
- Agree on the second point, and it addresses my concern with knowing what is the current default .. I still wonder how that was set / how to adjust it, but we can address that at a later point. May need some help with that one though.
- Yeah, I've brought this one up too, but I think there's some technical context behind this that @dennisreimann can or already has provided somewhere. I think it would need to be followed up in an additional PR.
- Agreed, and will remove the need for the Store selection on this view, etc.. we'll have to do some proper QA for this if/when we switch over to the Store-Centric architecture.

Great feedback @satwo!!

satwo:
@dstrukt Made a slight change that mostly takes care of the first point (still possible to submit the form with all payment methods unselected; if this happens, all methods are enabled). 

Reading back over the thread again, I just realized that @dennisreimann already addressed some of my points, sorry Dennis! Indeed the new Store-Centric UI will go a long way to solving some of these issues.

For the other stuff not related to changing the selected store, it probably makes sense to make a new PR. One additional thing I've been thinking about is whether it's possible to unify `Supported Transaction Currencies` and `Default payment method on checkout` into a single, easy-to-understand control.

dstrukt:
Beautiful, thank you - I think this PR is good to go.

But @satwo YES!

Also could include typing your currency vs. having a selectable list as part of improving that experience. I know there's a few technical considerations involved in all of that, but being able to make those distinctions more simple and clear would be a huge improvement .. could be something we explore in mocks for proposal.

PR #3094, opened 2021-11-13
Improves Api Key View

dstrukt:
Been going through some of the subpages and doing some quick clean up while I work through these views in Figma to prep for the Store Centric UI updates ... so apologies for future potential PR spam.

List Before:
Screen Shot 2021-11-12 at 8 23 50 PM

List After:
Screen Shot 2021-11-12 at 8 24 04 PM

Empty State Before:
Screen Shot 2021-11-12 at 8 24 26 PM

Empty State After:
Screen Shot 2021-11-12 at 8 27 02 PM

Note: Might be worth checking the main conditional, I think it's good though.

Otherwise this was just aesthetic clean up, labels, titles, padding, etc..

### DIFF
--- a/BTCPayServer/Views/Manage/APIKeys.cshtml
+++ b/BTCPayServer/Views/Manage/APIKeys.cshtml
@@ -1,92 +1,87 @@
 @namespace BTCPayServer.Client
 @model BTCPayServer.Controllers.ManageController.ApiKeysViewModel
 @{
-    ViewData.SetActivePageAndTitle(ManageNavPages.APIKeys, "Manage your API Keys");
+    ViewData.SetActivePageAndTitle(ManageNavPages.APIKeys, "API Keys");
 }
+<h2 class="mb-4">@ViewData["Title"]</h2>
 
 <p>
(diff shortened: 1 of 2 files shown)


woutersamaey:
Doesn't it look weird you can have an api key "without permissions" as shown in the screenshots?

dstrukt:
That's a good point .. should at least >1 permission be required to create a key? I could see arguments for both side.

dennisreimann (approved):
Yeehaw!

woutersamaey:
> I could see arguments for both side.

Since permissions cannot be changed after the api key was created, it should always have at least 1 permission.

dstrukt:
@woutersamaey Will try to solve in a separate PR, might be another good small task for me to try to tackle, if I cannot, will reach out!

Otherwise, I think this PR is good to go!

Kukks:
> That's a good point .. should at least >1 permission be required to create a key? I could see arguments for both side.

> Doesn't it look weird you can have an api key "without permissions" as shown in the screenshots?

No there are use-cases for a key with no permissions. Such as using btcpay just for proof-of-account

Kukks (approved):

PR #3101, opened 2021-11-14
Show correct note when crowdfund is publicly visible

bolatovumar:
See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/2968

|Before|After|
|---|---|
|Capture2|Capture|


MaxHillebrand:
Thanks @bolatovumar, exactly what I imagined, lovely!!

Issue #2625, opened 2021-06-26
[UI/UX] Add descriptions to pull payments


pavlenex:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/2586

<div type='discussions-op-text'>

<sup>Originally posted by **dstrukt** June 20, 2021</sup>
Similar to payment requests (which currently has a description field), context is needed for the pull payment.

I'd like to propose a simple description field be added to editor (can be a simple textarea or our wysiwyg) and the end-user view.

Mockups for reference (one with existing editor ui, and potential new create ui) and the end-user view.

Is there anything I'm missing?

Feedback & thoughts welcome!

<img width="623" alt="Screen Shot 2021-06-20 at 6 58 35 AM" src="https://user-images.githubusercontent.com/6250771/122676952-0807e700-d195-11eb-9588-7b7ce069f926.png">
 
<img width="929" alt="Screen Shot 2021-06-20 at 6 41 13 AM" src="https://user-images.githubusercontent.com/6250771/122676947-050cf680-d195-11eb-95a9-bf9c0ae7ef9a.png"></div>

PR #3107, opened 2021-11-15
Updates Pull Payment Payment Method

dstrukt:
Similar to the recent invoice PR, updates the terrible multi-selector to a checkbox. There might be an even better component for this, but for the time being, a quick fix until we have something better.

Removes "back to list" and updates the label.

Before:
Screen Shot 2021-11-15 at 1 48 38 AM

After:
Screen Shot 2021-11-15 at 1 48 20 AM

cc @dennisreimann

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -1391,11 +1391,8 @@ namespace BTCPayServer.Tests
             await s.Server.CustomerLightningD.Pay(lnurlResponse.Pr);
             Assert.Equal(new LightMoney(0.0000001m, LightMoneyUnit.BTC),
                 lnurlResponse2.GetPaymentRequest(network).MinimumAmount);
-
-
-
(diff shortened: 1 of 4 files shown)


dennisreimann:
Moved the custom CSS fields into an Additional Options block, so that it works like everywhere else.

Also updated the user facing side for the case in which there is only one payment method.

 ## Before

image

## After

image

dennisreimann (approved):

dstrukt:
*chef's kiss*

LGTM!

pavlenex:
@dstrukt I remember you wanting to add a description? #2625 but i assume it's beyond the scope of this PR, just a reminder.

dstrukt:
@pavlenex I did indeed (https://github.com/btcpayserver/btcpayserver/issues/2625), but agreed, I think that's a separate PR.

I may try to attempt, but someone might beat me to it at the rate @dennisreimann and @satwo are moving.

PR #3111, opened 2021-11-15
Improve invoice warning for no payment method configured

### LINKED ISSUE #3085: Improve the warning when generating an invoice without a payment method configured.
**Describe the improvement**
Since recently, we began introducing the concept of a payment method into the UI. 

Screenshot 2021-11-12 at 11 56 07

When user generates an invoice ,we do say `set up a wallet` which would mean `an on-chain` payment method is required for invoice creation, which in fact, it's not, user can just set lightning network. To avoid this confusion, we should probably refer to `payment method`

Screenshot 2021-11-12 at 11 56 46

I suggest: `To create an invoice, you need to set up a payment method first`. This error could be used in #3084 as well.

Alternative solution would be to add `Lightning Wallet` in the UI, but I still think we should help users think of `payment methods` instead of wallets.

dennisreimann:
Closes #3085.

### DIFF
--- a/BTCPayServer/Controllers/InvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/InvoiceController.UI.cs
@@ -872,10 +872,10 @@ namespace BTCPayServer.Controllers
 
             if (!store.GetSupportedPaymentMethods(_NetworkProvider).Any())
             {
-                TempData.SetStatusMessageModel(new StatusMessageModel()
+                TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id })}' class='alert-link'>set up your wallet</a> first",
+                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id })}' class='alert-link'>set up a payment method</a> first",
                     AllowDismiss = false
                 });
                 return View(model);


BTCPayServer/Controllers/InvoiceController.UI.cs
```diff
@@ -872,10 +872,10 @@ public async Task<IActionResult> CreateInvoice(CreateInvoiceModel model, Cancell
 
             if (!store.GetSupportedPaymentMethods(_NetworkProvider).Any())
             {
-                TempData.SetStatusMessageModel(new StatusMessageModel()
+                TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id })}' class='alert-link'>set up your wallet</a> first",
+                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id })}' class='alert-link'>set up a payment method</a> first",
```
Zaxounette, line 878:
```suggestion
                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(StoresController.PaymentMethods), "Stores", new { storeId = store.Id })}' class='alert-link'>configure a payment method</a> first",
```

Zaxounette (commented):
I don't really like the words "set up" since it implies some potentially complexe manipulations from the users, whereas "configure" has an implicit meaning of it being a routine non-complexe task.

Just a suggestion.

BTCPayServer/Controllers/InvoiceController.UI.cs
dstrukt: +1 to "Configure"

pavlenex: Yes but we have `Set up` in the wallet view, which would mean we'd have to change those as well.

dstrukt: Good point .. considering this particular UI will be deprecated in the coming weeks, I'd opt for keeping it Set Up for the sake of consistency, and we can readdress the copy at a later point, as much as I also like Configure > Set up, can be something we adjust in the future.

My 2 sats FWIW.

pavlenex (approved):
Alright, I think for consistency, we should merge this PR and open a new issue for talking about `payment method configuration` rather than `setup` in the 1.5.0 or later.

Kukks (approved):

PR #3112, opened 2021-11-16
Improve Pay Button View

dstrukt:
This should hopefully help clean up and consolidate this view a bit more, dropping it somewhat incomplete as I think there will be some opinions, and gives you guys some time to think/work through some of the proposed ideas.

This can, and will continue to be improved, but this would be a nice step in the right direction, as we hope to surface this on the left sidebar in 1.4.0 .. and it never hurts to make this as easy to use and understand as possible, cause it's an awesome feature.

### CURRENT STATE AS OF THIS PR

Before:
Screen Shot 2021-11-15 at 9 39 06 PM
Screen Shot 2021-11-15 at 9 39 10 PM

After:
Screen Shot 2021-11-15 at 9 38 28 PM
Screen Shot 2021-11-15 at 9 38 38 PM

------

### Proposals
Things not done, but potentially could be added:
- Tucking the advanced settings behind some of our accordion components
- Creating a "Copy" component for the generated link on the right, and a clickable link to see the preview below the form element.
- Changing the X x Y px buttons to Small/Medium/Large ... Thoughts?
- Add a bit more context to the "Modal" checkbox, it's not very clear..
- I think if we have the accordions we can still keep the generated code at the very bottom, but that could arguably be moved above the "Advanced Options", but I'd have to see this in action first to make this call.
- Still working through copy, but we definitely need a succinct and clear explanation of what Pay Button is (for the initial "Enable" screen -- top little section on the mocks, simple 1-2 sentence would probably be sufficient) cc @pavlenex @Zaxounette, however anything and everything else is subject to improvement as well, tried my best to reduce and simplify.
- Anything else?

### Questions
- Does the current UX of "Enable" make sense? Why can't it be disabled after enabled?

Screen Shot 2021-11-15 at 9 36 45 PM
Screen Shot 2021-11-15 at 9 36 53 PM

### DIFF
--- a/BTCPayServer/Views/Stores/PayButton.cshtml
+++ b/BTCPayServer/Views/Stores/PayButton.cshtml
@@ -35,9 +35,8 @@
         }
     </template>
     <script>
-        var srvModel = @Safe.Json(Model);
-
-        var payButtonCtrl = new Vue({
+        const srvModel = @Safe.Json(Model);
+        const payButtonCtrl = new Vue({
             el: '#payButtonCtrl',
             data: {
                 srvModel: srvModel,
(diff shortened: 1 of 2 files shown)


dennisreimann (approved):
Niiice. :cake:

PR #3122, opened 2021-11-18
Add all buyer fields to create invoice page and adjust layout

bolatovumar:
This PR is a re-worked version of [this PR](https://github.com/btcpayserver/btcpayserver/pull/2603) based on feedback in this comment: https://github.com/btcpayserver/btcpayserver/pull/2603#pullrequestreview-697984211.

|Before|After|
|---|---|
|Capture|Capture|
|   |Capture2|

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -393,8 +393,9 @@ namespace BTCPayServer.Tests
             var currencyEl = Driver.FindElement(By.Id("Currency"));
             currencyEl.Clear();
             currencyEl.SendKeys(currency);
-            Driver.FindElement(By.Id("BuyerEmail")).SendKeys(refundEmail);
-            Driver.FindElement(By.Name("StoreId")).SendKeys(storeName);
+            Driver.FindElement(By.CssSelector("button[value=buyer-details]")).Click();
+            Driver.WaitForElement(By.Id("BuyerEmail")).SendKeys(refundEmail);
+            Driver.WaitForElement(By.Name("StoreId")).SendKeys(storeName);
             if (defaultPaymentMethod is string)
                 new SelectElement(Driver.FindElement(By.Name("DefaultPaymentMethod"))).SelectByValue(defaultPaymentMethod);
             if (requiresRefundEmail is bool)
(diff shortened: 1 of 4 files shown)


dstrukt:
Added this as a list item for the design call, but wasn't able to get to everything, how should we combine the great work being done on this PR with (https://github.com/btcpayserver/btcpayserver/pull/3092)? Will there any potential conflicts?

cc @dennisreimann 

----------

Really liking the introduction of the Buyer fields! However, I wonder .. should the language be "Customer"? 

cc @pavlenex @Zaxounette

bolatovumar:
@dstrukt we can merge #3092 and then I can rebase this PR against latest code and have the buyer fields added on top

satwo:
tACK, but I agree with @dstrukt that `Customer` is probably the better choice. 

Might be worth putting some validation on the `Phone` field too?

dstrukt:
Beautiful @bolatovumar, great work!

+1 to phone validation, unless there's something I'm missing.

bolatovumar:
@satwo @dstrukt I can add phone validation but one drawback I noticed with it is that the validation doesn't happen in real time, it only happens on form submit (unlike email). I'm guessing this is probably there are different phone formats and .NET needs to do this on the server to match the request locale or something. I left it out because of this.

Zaxounette:
> Really liking the introduction of the Buyer fields! However, I wonder .. should the language be "Customer"?

I agree, `buyer` can be changed to `customer` in the Buyer Detail title.

Buyer in each field can be delete altogether. With the title, we know those fields will relate to customer information and the specific detail needed in that field is sufficient (ie. Phone instead of Customer Phone)

dstrukt:
Agreed @Zaxounette nice catch!

satwo:
@bolatovumar I was thinking the same thing - not quite so simple to do international phone validation compared to domestic...

I'm sure it's fine for now but maybe something to think about long-term? (if/when businesses start to hook their phone systems into BTCPay Server.)

NicolasDorier:
Concept NACK.
Those fields are meant to be deprecated, as they have no meaning and impact.
If I remember @dennisreimann worked on a better solution somewhere, where it was dynamic (a table which you can add rows to)

bolatovumar:
@NicolasDorier ok, got it, closing this PR.

PR #3127, opened 2021-11-18
Simple Payment Request and Pull Payment Views Improvements

dstrukt:
Entirely aesthetic changes, from padding to font weights, border-radius, etc..

I have a few more changes I want to make to these views, but I'll follow up on that in another PR to keep this one simple.

Updates:
Screen Shot 2021-11-18 at 12 04 13 PM
Screen Shot 2021-11-18 at 12 04 19 PM

-------

The only issue I had was trying to resolve but couldn't was the conditional switch for adding this to the "Invoice Summary" on the Payment Request as shown:

Screen Shot 2021-11-18 at 12 08 06 PM

Really simple conditional, but for whatever reason, I wasn't able to get the empty state } else { condition to display "No details provided." as shown in the first screenshot.

Snippet of the existing conditional sans my attempts:
`@if (!string.IsNullOrEmpty(Model.Description) && Model.Description != "<br>")
{
@Safe.Raw(Model.Description)
}`

cc @satwo @dennisreimann

### DIFF
--- a/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
+++ b/BTCPayServer/Views/PaymentRequest/ViewPaymentRequest.cshtml
@@ -219,18 +219,18 @@
                             <dl class="mb-0 mt-md-4">
                                 <div class="d-flex d-print-inline-block flex-column mb-4 me-5">
                                     <dt class="h4 fw-semibold text-nowrap text-primary text-print-default order-2 order-sm-1 mb-1" v-text="srvModel.amountDueFormatted">@Model.AmountDueFormatted</dt>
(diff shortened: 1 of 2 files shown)


satwo:
> Really simple conditional, but for whatever reason, I wasn't able to get the empty state } else { condition to display "No details provided." as shown in the first screenshot.

I messed around for a bit and it seems to be working now. I guess the problem was vue-related: by putting the whole conditional inside the `<div v-html="srvModel.description">` element, it prevented the plain HTML `else` condition (i.e. without `Model.Description`) from rendering properly.

dennisreimann (approved):
Sweeeet.

dstrukt:
Beautiful, thank you @satwo!

LGTM

PR #3091, opened 2021-11-12
Store-centric UI

dennisreimann:
First rough draft with a basic new layout structure for mobile and desktop.

Putting it out there, because everyone is excited and curious.

Closes #3002.

PR #3128, opened 2021-11-19
Allow pull payments for store guests

Kukks:


satwo:
Great proof of concept @Kukks!

I guess the next question is how extensible we want this to be. We also need a way to create custom user roles and assign various permissions to them. I imagine we'd need to add/change some data structures for that.

dennisreimann (approved):
tACK. From a users perspective this works as it should.

I really like the permission tag helper concept — we can surely use this in other places as well.

Kukks:
> tACK. From a users perspective this works as it should.
> 
> I really like the permission tag helper concept — we can surely use this in other places as well.

Should we merge this now or should we wait for #3091 @dennisreimann?

dennisreimann:
I'm waiting for this to be merged so that I can continue in #3091 :)

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -47,20 +47,20 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
+                    break;
+                case Policies.CanViewStoreSettings:
+                    if (store != null || isAdmin)
                         success = true;
                     break;
                 case Policies.CanCreateInvoice:
-                    if (store.Role == StoreRoles.Owner ||
-                        store.Role == StoreRoles.Guest ||
-                        isAdmin ||
-                        store.GetStoreBlob().AnyoneCanInvoice)
```
NicolasDorier, line 63: `store.GetStoreBlob().AnyoneCanInvoice` gone?

Kukks: It never worked because we always fetched the store in relation to the current user. If the user did not have access to the store as a guest or owner, store == null and it would never reach here.

PR #3039, opened 2021-10-31
PR for Issue #3038 adds params for getInvoices pagination

TheHazeEffect:
This pull request is related to [Issue 3038](https://github.com/btcpayserver/btcpayserver/issues/3038) 
It adds take and skip query parameters to the getInvoices endpoint of the Greenfield API to allow pagination when fetching invoices

PR #3145, opened 2021-11-22
Greenfield: Add pagination for API GetNotifications

satwo:
This adds `skip` and `take` pagination for the GetNotifications API, in keeping with the legacy BitPay api. Similar to #3039.


satwo:
I don't think the failed test is related.

bolatovumar (approved):

PR #3157, opened 2021-11-25
Update Bitbank API

junderw:
Please fix my (probably broken) code when you have the time and merge it.

I wasn't sure how you want to assert success === 1.


junderw:
We will leave the prices up for a while after merge/release so that people on older versions won't break.

Kukks (approved):

junderw:
Thanks!

PR #3162, opened 2021-11-27
Fix possible null reference exception when creating a pull payment

bolatovumar:
I noticed we can run into null reference exception when creating a pull payment. This PR fixes these issues.

error
error2

### DIFF
--- a/BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
+++ b/BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
@@ -95,9 +95,13 @@ namespace BTCPayServer.Controllers
             model.PaymentMethodItems =
                 paymentMethodOptions.Select(id => new SelectListItem(id.ToPrettyString(), id.ToString(), true));
             model.Name ??= string.Empty;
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
```diff
@@ -95,9 +95,13 @@ public async Task<IActionResult> NewPullPayment(string storeId, NewPullPaymentMo
             model.PaymentMethodItems =
                 paymentMethodOptions.Select(id => new SelectListItem(id.ToPrettyString(), id.ToString(), true));
             model.Name ??= string.Empty;
-            model.Currency = model.Currency.ToUpperInvariant().Trim();
+            model.Currency = model.Currency != null ? model.Currency.ToUpperInvariant().Trim() : String.Empty;
```
Kukks, line 98:
```suggestion
            model.Currency = model.Currency?.ToUpperInvariant()?.Trim() ?? String.Empty;
```

Kukks (changes requested):

PR #3163, opened 2021-11-28
Cheat mode to mine blocks

woutersamaey:
Expands cheat mode so we can now mine 1 or more blocks.
Also improves JavaScript code a bit.


woutersamaey:
@NicolasDorier can we get this merged quickly please? I'd like to use this for my automated testing with Zapier.

We can now mine blocks, and this is needed to fake the processing and settlement webhooks.

PR #3164, opened 2021-11-29
Send block reward to a random address instead of the invoice address. Default mine 1 block.

woutersamaey:
In previous PR (#3163) I made a dumb mistake of sending the block reward to the invoice address. This is fixed in this PR.


NicolasDorier:
Do not hardcode, just generate to itself. You can use `cashcow.GetNewAddress()`

woutersamaey:
@NicolasDorier did as you suggested. Tnx!

Issue #3167, opened 2021-11-30
Invoice page - Dropdown magically disappears 


trigger67:
**Describe the bug**
In the invoice page, the "pay with" dropdown disappears !

**To Reproduce the bug**
Steps to reproduce the reported bug:
1. Generate an invoice, on a shop accepting more than one method of payment, like Bitcoin on-chain and lightning (try NicolasDorier donation page)
2. On the invoice payment page, click on the dropdown to select either BTC or Lightning
3. Then click on one of the two, but click on the right, on the "BTC" text (click on the span tag)

Tada, the drowdown disappears, and your choice has not been taken into account

- BTCPay Server Version: 1.3.6
- Browser: chromium

![image](https://user-images.githubusercontent.com/10185784/144065386-18833117-2a5b-4377-a54c-bb62508cfebf.png)

trigger67:
I will do a pull request to fix it.

dennisreimann:
@trigger67 Thanks for the contribution!

I think the proper way to fix it is to change the JS, so that we don't have to exclude single elements within the payment method via CSS. Here's what works for me in the `checkout.js`, can you double-check and if it works adapt the change here?

```js
delegate('click', '.payment-method', e => {
    const el = e.target.closest('.payment-method')
    closePaymentMethodDialog(el.dataset.paymentMethod);
    return false;
})
```

trigger67:
@dennisreimann I agree your code is cleaner, but the UX is a bit worse.
With my solution, when selecting an element, the dropdown disappears with a nice transition.
With your solution, it "freezes" a bit, and then the screen is refreshed (you can see it scrolls to the top of the screen).

Maybe there is a way to do it without affecting the UX in javascript...

dennisreimann:
Oh sorry, I haven't checked – need to take another look, there's a proper way to solve this with the Js approach, I must have made a mistake here.

dennisreimann:
Hmm, I just checked and as far as I can see this is working properly and as expected. I don't get a page refresh, the dropdown fades out and the chosen payment method is set.

NicolasDorier:
@trigger67 the bug of the page refreshed happened before, but not too long ago (should have been fixed in 1.3.6 already)
I can't reproduce either. Are you sure the server is updated?

trigger67:
@NicolasDorier @dennisreimann my bad I did a mistake, I checked again and it is working fine with the js code.
I will update the pull request later today!
Thank you

[...]

PR #3169, opened 2021-12-01
Fix issue in invoice when changing the payment method

trigger67:
Fixes the issue #3167 

Issue is that when clicking on the span, the PaymentMethodId sent to the server is "undefined".

### DIFF
--- a/BTCPayServer/wwwroot/checkout/js/checkout.js
+++ b/BTCPayServer/wwwroot/checkout/js/checkout.js
@@ -1,4 +1,5 @@
-delegate('click', '.payment-method', function(e) {
-    closePaymentMethodDialog(e.target.dataset.paymentMethod);
+delegate('click', '.payment-method', e => {
+    const el = e.target.closest('.payment-method')
+    closePaymentMethodDialog(el.dataset.paymentMethod);
     return false;
 })


NicolasDorier:
ping @dennisreimann should we merge this?
The bug he reported doesn't seem to happen to me, but it has been fixed on 1.3.1 (https://github.com/btcpayserver/btcpayserver/blob/master/Changelog.md#131) by this commit https://github.com/btcpayserver/btcpayserver/commit/933dc6be312a8d1265dda12c628779ccca0af9a7

May it be a cache issue? @trigger67 can you create an invoice with this problem?

trigger67:
@NicolasDorier I have this problem in production with the 1.3.6 version. I also have it locally on the master branch. And I see the bug on your website :)
Use chromium or firefox, generate an invoice to do yourself a donation , and click on the "BTC"  label.

dennisreimann:
Yes, it is caused by clicking on nested elements within the whole row. If you click on the white space in the center it doesn't happen, but when the label on the right is clicked.

trigger67:
Fixed, thank you @dennisreimann

dennisreimann (approved):

NicolasDorier:
I think I'll do another bug fix release, there is 2 or 3 commits I want in, this one included.

siwatanejo:
Beware, the diff for this PR ended up being just basically empty (only changing an EOL).

NicolasDorier:
@siwatanejo I noticed today it was still broken in 1.3.7!
But not on master

trigger67:
That's because @dennisreimann commited exactly the same update just before me : https://github.com/btcpayserver/btcpayserver/commit/f8e6b51e9d1b1391a64ce95d73868faf0280bec2#diff-9b845087c7c4d059626237a82135c874835bafc9738d7d81ea49fb31c7bb3e29

dennisreimann:
Yikes, that's a bummer. Thanks for researching this, @trigger67 !

PR #3172, opened 2021-12-03
Fix code display

dennisreimann:
Fixes a visual bug that is cause by the text-wrap class being applied to the code element instead of the parent pre element.

## Before

image

## After

image


woutersamaey:
Thanks for the quick fix.

PR #3177, opened 2021-12-06
Add pull payment grouping options

### LINKED ISSUE #3070: Add grouping capabilities for pull payments
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/3066

<sup>Originally posted by **NicolasDorier** November  4, 2021</sup>
image
</div>

bolatovumar:
Close #3070

https://user-images.githubusercontent.com/1934678/147776111-b81165c0-60c6-4bd4-b446-e0ad9e50d2a9.mov

### DIFF
--- a/BTCPayServer.Client/Models/PullPaymentBaseData.cs
+++ b/BTCPayServer.Client/Models/PullPaymentBaseData.cs
@@ -9,7 +9,8 @@ namespace BTCPayServer.Client.Models
     {
         Active,
         Expired,
-        Archived
+        Archived,
+        Future
     }
     public class PullPaymentData
     {
--- a/BTCPayServer.Client/Models/PullPaymentBaseData.cs
+++ b/BTCPayServer.Client/Models/PullPaymentBaseData.cs
@@ -5,6 +5,12 @@ using Newtonsoft.Json;
 
 namespace BTCPayServer.Client.Models
 {
+    public enum PullPaymentState
+    {
+        Active,
+        Expired,
+        Archived
+    }
     public class PullPaymentData
     {
         [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
--- a/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
+++ b/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
@@ -253,8 +253,9 @@ namespace BTCPayServer.Controllers
                         Completed = totalCompleted.RoundToSignificant(ni.Divisibility).ToString("C", nfi),
                         Limit = _currencyNameTable.DisplayFormatCurrency(ppBlob.Limit, ppBlob.Currency),
                         ResetIn = period?.End is DateTimeOffset nr ? ZeroIfNegative(nr - now).TimeString() : null,
-                        EndIn = pp.EndDate is DateTimeOffset end ? ZeroIfNegative(end - now).TimeString() : null
-                    }
+                        EndIn = pp.EndDate is DateTimeOffset end ? ZeroIfNegative(end - now).TimeString() : null,
+                    },
+                    Archived = pp.Archived
                 });
             }
             return View(vm);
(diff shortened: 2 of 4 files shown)


BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
```diff
@@ -141,45 +141,92 @@ await _pullPaymentService.CreatePullPayment(new HostedServices.CreatePullPayment
...
+            PullPaymentState pullPaymentState,
+            int skip = 0,
+            int count = 50,
+            string sortOrder = "desc"
+        )
         {
             await using var ctx = _dbContextFactory.CreateContext();
             var now = DateTimeOffset.UtcNow;
-            var ppsQuery = ctx.PullPayments
+            var ppsQuery = await ctx.PullPayments
```
bolatovumar, line 157: Because I'm filtering by payment method below by running `GetBlob().SupportedPaymentMethods.Contains(PaymentMethodId.Parse(vm.PaymentMethodId))` I had to convert this query to run the filtering on the client.

Also, I'm filtering by computed fields such as `IsRunning()` and `IsExpired()` below which will only work on the client.

Idk if this is acceptable as now we have to retrieve all pull payments before filtering. Let me know if you think there is an alternative approach I could take here.

BTCPayServer/Views/StorePullPayments/PullPayments.cshtml
```diff
@@ -117,15 +151,20 @@
...
-                            </a> -
-                            <a asp-action="ArchivePullPayment"
-                               asp-route-storeId="@Context.GetRouteValue("storeId")"
-                               asp-route-pullPaymentId="@pp.Id"
-                               data-bs-toggle="modal"
-                               data-bs-target="#ConfirmModal"
-                               data-description="Do you really want to archive the pull payment <strong>@pp.Name</strong>?">
-                                Archive
                             </a>
+                            @if (!pp.Archived)
```
bolatovumar, line 155: Removing the "Archive" button from the already archived items.

BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
NicolasDorier: We have a user who has 100 000 Pull payments/Payouts. This code here would completely explode his server, this is why paging has to happen BEFORE making the query to database.

Sadly, I don't know if there is a workaround this.

Also, if the payment state is active/running you could filter server side with archived status.

bolatovumar: @NicolasDorier I was thinking for payment methods maybe we could save the initially selected payment methods as strings in the DB. That _should_ work, no?

As for the other things I'm not sure.

bolatovumar: Even if we were to save the payment ids as strings for each pull payment we would have to convert all existing pull payments to also have these payment method id arrays.

BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
```diff
@@ -225,7 +265,7 @@ public TimeSpan ZeroIfNegative(TimeSpan time)
         {
             return View("Confirm", new ConfirmModel("Archive pull payment", "Do you really want to archive the pull payment?", "Archive"));
         }
-
```
bolatovumar, line 228: All of these changes are just trailing whitespaces removed automatically.

BTCPayServer/Controllers/StorePullPaymentsController.PullPayments.cs
bolatovumar: @NicolasDorier alright, I updated it to NOT filter on the client side. I had to remove filtering by payment method because I couldn't find a way to do this filtering not on the client.

bolatovumar:
ping @NicolasDorier

NicolasDorier:
sorry coming back to this one

lucidddreams:
This is a good feature, but would be better if the following grouping is also added

Awaiting - Unclaimed / Partially claimed pull payout
Completed - Fully claimed pull payout

PR #3181, opened 2021-12-06
New API endpoint: Send email using store SMTP

woutersamaey:
- Send a plain text email using the Store's SMTP
- No HTML mail support yet as we may want to develop our own email letterhead etc at a later time
- No attachment support yet, this could be an incremental improvement later on
- I'm using the `CanModifyStoreSettings` permission, but think we can make a new permission `CanSendStoreEmail` or similar as we are adding a new feature. Changing this later would mean breaking changes at that time, so would be best to consider this now.

**Usage**
```
POST /api/v1/stores/xxxxx/email
{"toName": "Wouter Samaey", "toEmail": "wouter.samaey@storefront.be", "subject": "Hey hey!", "body": "Testing testing"}
```

**Responses**
200 OK
404 if store not found
400 SMTP not configured yet
400 some other email problem

### DIFF
--- a/BTCPayServer.Client/Models/SendEmailRequest.cs
+++ b/BTCPayServer.Client/Models/SendEmailRequest.cs
@@ -1,0 +1,17 @@
+using System.Net.Mail;
+
+namespace BTCPayServer.Client.Models
+{
+    public class SendEmailRequest
+    {
+        public string toName;
+        public string toEmail;
+        public string subject;
+        public string body;
+
+        public MailAddress toMailAddress()
+        {
+            return new MailAddress(toEmail, toName);
+        }
+    }
+}
--- a/BTCPayServer.Client/Models/SendEmailRequest.cs
+++ b/BTCPayServer.Client/Models/SendEmailRequest.cs
@@ -1,18 +1,10 @@
-using System.Net.Mail;
-using MimeKit;
 
 namespace BTCPayServer.Client.Models
 {
     public class SendEmailRequest
     {
-        public string toName;
-        public string toEmail;
-        public string subject;
-        public string body;
-
-        public MailboxAddress toMailAddress()
-        {
-            return new MailboxAddress(toEmail, toName);
-        }
+        public string Email;
+        public string Subject;
+        public string Body;
     }
 }
--- a/BTCPayServer.Client/Models/SendEmailRequest.cs
+++ b/BTCPayServer.Client/Models/SendEmailRequest.cs
@@ -1,4 +1,5 @@
 using System.Net.Mail;
+using MimeKit;
 
 namespace BTCPayServer.Client.Models
 {
@@ -9,9 +10,9 @@ namespace BTCPayServer.Client.Models
         public string subject;
         public string body;
 
-        public MailAddress toMailAddress()
+        public MailboxAddress toMailAddress()
         {
-            return new MailAddress(toEmail, toName);
+            return new MailboxAddress(toEmail, toName);
         }
     }
 }
--- a/BTCPayServer/Controllers/GreenField/StoreSendEmail.cs
+++ b/BTCPayServer/Controllers/GreenField/StoreSendEmail.cs
@@ -31,19 +31,21 @@ namespace BTCPayServer.Controllers.GreenField
             var emailSettings = store.GetStoreBlob().EmailSettings ?? null;
             if (emailSettings != null || !emailSettings.IsComplete())
             {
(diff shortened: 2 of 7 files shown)


woutersamaey:
Swagger coming later

woutersamaey:
I should probably change the response code for "some other email problem" to something in the 500 range. Will investigate a bit more and update.

NicolasDorier:
@woutersamaey can you review/test then rebase on https://github.com/btcpayserver/btcpayserver/pull/3202 once merged?

BTCPayServer.Client/BTCPayServer.Client.csproj
```diff
@@ -27,6 +27,7 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
     <ItemGroup>
+      <PackageReference Include="MimeKit" Version="3.0.0" />
```
Kukks, line 30: we should not add this dependency to the client

BTCPayServer.Client/Models/SendEmailRequest.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace BTCPayServer.Client.Models
+{
+    public class SendEmailRequest
+    {
+        public string toName;
+        public string toEmail;
+        public string subject;
+        public string body;
+
+        public MailboxAddress toMailAddress()
```
Kukks, line 13: Can just be an extension method inside BTCPay  project

BTCPayServer/Controllers/GreenField/StoreSendEmail.cs
```diff
@@ -0,0 +1,52 @@
...
+                return this.CreateAPIError(404, "store-not-found", "The store was not found");
+            }
+
+            var emailSettings = store.GetStoreBlob().EmailSettings ?? null;
+            if (emailSettings != null || !emailSettings.IsComplete())
+            {
+                return this.CreateAPIError(404,"smtp-not-configured", "Store does not have an SMTP server configured.");
+            }
+
+            var client = await emailSettings.CreateSmtpClient();
```
Kukks, line 37: We should ideally use the `EmailSenderFactory` service here

Kukks (changes requested):

BTCPayServer.Client/Models/SendEmailRequest.cs
Kukks: Also: pascal case

woutersamaey: You mean putting this in a plugin? If so I don't want to do this. The reason I'm making this is so that sending emails in the name of the store becomes easier for Zapier users.

Kukks: No, c# has a concept called extension methods for example:

```cs
public static class EmailExtensions{
  public static  MailboxAddress ToMailAddress(this SendEmailRequest request){
	return new MailboxAddress(request.ToEmail, request.ToName);
  }
}
```


This could live inside Btcpayserver proj, while the Btcpayserver.client can remove the mimekit dependency

woutersamaey:
@Kukks do we have a markdown to HTML tool in BTCPay? It may be handy to allow markdown for the email body. This would allow you some basic email formatting without going full HTML which can later conflict with email templates.

I'd add a field `body_type` and set that to `plain` or `markdown`.

The markdown body can be put inside a full email template without breaking the HTML around it.

I'm thinking about this as with eCommerce emails, you'll likely want to have some rich content such as an order table or a few links, bullets, etc. Allowing only plain text is very very restrictive.

Kukks:
> @Kukks do we have a markdown to HTML tool in BTCPay? It may be handy to allow markdown for the email body. This would allow you some basic email formatting without going full HTML which can later conflict with email templates.
> 
> I'd add a field `body_type` and set that to `plain` or `markdown`.
> 
> The markdown body can be put inside a full email template without breaking the HTML around it.
> 
> I'm thinking about this as with eCommerce emails, you'll likely want to have some rich content such as an order table or a few links, bullets, etc. Allowing only plain text is very very restrictive.

This is out of scope for this PR. Let's keep it lean.

PS. I adapted this PR to sue the internal services we use for sending emails. Email bodies are html by default. The endpoint is now also /v1/stores/storeid/email/send so that we have /v1/stores/storeid/emaill for configuring the email settings

Issue #3156, opened 2021-11-24
Dash/Doge Won't Sync


MathijsCraft:
So, I keep getting the issue where my data gets corrupted.
Its for the coins Dash and Dogecoin.
https://docs.btcpayserver.org/FAQ/Synchronization/#cause-4-your-bitcoin-data-directory-is-corrupted 
I followed this guide like 20x (Yes I edited the commands to the coin) I also tried to manually delete the folders.
I asked on the chat platform but they didnt have an idea either.

pavlenex:
Hey! This issue is a duplicate of #2824

Unfortunately altcoins are maintained by their communities so you'd have to reach out to them somehow. This issue has been happening for months now, but there's no effort on communities end to address this. So there's no solution but to wait, bug someone from dash/doge team or remove the coin from your deployment.

PR #3182, opened 2021-12-08
Remove Ethereum support

NicolasDorier:
Ethereum support was added by @xpayserver .

I decided we remove it because it is unusable for several reason:

1. It use one account per invoice, but no wallet in ethereum support HD accounts, so you need to rely on a sweeper program, that isn't really available and documented.
2. Using one account per invoice means that the merchant need to do one transaction per invoice to sweep back to an exchange. At current ETH fee (>50 USD per transactions for months), it isn't economically viable. Need to use instead some smart stuff, so called "dynamic proxy " if I remember how it is called. But if we do that, better start from scratch.

If one day somebody add eth again, he can probably make his life easier by looking the code I removed in this PR.


pavlenex:
Concept ACK. Looking at things this is most likely to happen with Doge and Dash in the future due to #3156. Hopefully in the long run these all become plugins that we can nuke from the main codebase. LFG

NicolasDorier:
@pavlenex DOGE/Dash aren't that much an issue. We rely on NBitcoin for support, so removing support for it will remove at most 200 lines. But ETH code modifications are spread all around.

woutersamaey:
The changes you are making here will likely help identify which parts would need to be refactored should we want to switch to plugins for payment methods.

dennisreimann (approved):
tACK

PR #3195, opened 2021-12-12
Mention the missing API permission in the response of a Greenfield request

woutersamaey:
This PR makes the Greenfield API tell us which API permissions we are lacking for the request to work (avoiding `403 Forbidden`).

This is needed because of:
- An integration can now know what the user needs to do to make the integration work. He needs to use a better API key, but for normies to figure out which permission can be a pain. The integration can help the user give the correct advise.
- A developer will waste less time on silly API permission mistakes

This PR offers 2 solutions. Obviously **we should only keep 1 of the 2 solutions**.

**Solution 1**
Add an extra response header `X-Missing-API-Permission`. This is clean and does not affect the body. I.e. if some other output were to happen afterwards, it's not an issue.

**Solution 2**
Write the response by hand. This is less clean, but it is what you would expect from an API. Previously there was no body, so this becomes the only output and we CAN do it manually.

I have kept both so we can discuss and make a choice.

Note: This is not a security issue as we are publicly documenting this stuff anyway. It just helps API users figure out what is wrong and what is expected of them to fix.

Would love your inputs @NicolasDorier  and @Kukks. You may have better practise solutions, though mine works and gets the job done.

@Zaxounette and @britttttk This is going to be used in combination with Zapier so we can tell the user what permission he needs to make their automation/Zap work. We talked about this on the Zapier demo call.

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -77,9 +77,7 @@ namespace BTCPayServer.Tests
                 {
                     Assert.Equal("insufficient-api-permissions", e.APIError.Code);
                     Assert.NotNull(e.APIError.Message);
-                    Assert.IsType<GreenfieldPermissionAPIError>(e.APIError);
-
-                    GreenfieldPermissionAPIError permissionError = (GreenfieldPermissionAPIError)e.APIError;
+                    GreenfieldPermissionAPIError permissionError = Assert.IsType<GreenfieldPermissionAPIError>(e.APIError);
                     Assert.Equal(permissionError.MissingPermission, Policies.CanModifyStoreSettings);
                     Assert.NotNull(permissionError.MissingPermissionDescription);
                 }
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -66,21 +66,12 @@ namespace BTCPayServer.Tests
                 var user = tester.NewAccount();
                 user.GrantAccess();
                 var clientWithWrongPermissions = await user.CreateClient(Policies.CanViewProfile);
-                
-                try
-                {
-                    // Try to create a store with a client that has the wrong permissions.
-                    var data = await clientWithWrongPermissions.CreateStore(new CreateStoreRequest() { Name = "mystore" });
-                    throw new XunitException("This point should never be reached.");
-
-                }catch (GreenFieldAPIException e)
-                {
-                    Assert.Equal("insufficient-api-permissions", e.APIError.Code);
-                    Assert.NotNull(e.APIError.Message);
-                    GreenfieldPermissionAPIError permissionError = Assert.IsType<GreenfieldPermissionAPIError>(e.APIError);
-                    Assert.Equal(permissionError.MissingPermission, Policies.CanModifyStoreSettings);
-                    Assert.NotNull(permissionError.MissingPermissionDescription);
-                }
+                var e = await AssertAPIError("insufficient-api-permissions", () => clientWithWrongPermissions.CreateStore(new CreateStoreRequest() { Name = "mystore" }));
+                Assert.Equal("insufficient-api-permissions", e.APIError.Code);
+                Assert.NotNull(e.APIError.Message);
+                GreenfieldPermissionAPIError permissionError = Assert.IsType<GreenfieldPermissionAPIError>(e.APIError);
+                Assert.Equal(permissionError.MissingPermission, Policies.CanModifyStoreSettings);
+                Assert.NotNull(permissionError.MissingPermissionDescription);
             }
         }
 
@@ -611,10 +602,11 @@ namespace BTCPayServer.Tests
             return new DateTimeOffset(dateTimeOffset.Year, dateTimeOffset.Month, dateTimeOffset.Day, dateTimeOffset.Hour, dateTimeOffset.Minute, dateTimeOffset.Second, dateTimeOffset.Offset);
         }
 
-        private async Task AssertAPIError(string expectedError, Func<Task> act)
+        private async Task<GreenFieldAPIException> AssertAPIError(string expectedError, Func<Task> act)
         {
             var err = await Assert.ThrowsAsync<GreenFieldAPIException>(async () => await act());
             Assert.Equal(expectedError, err.APIError.Code);
+            return err;
         }
 
         [Fact(Timeout = TestTimeout)]
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -66,12 +66,11 @@ namespace BTCPayServer.Tests
                 var user = tester.NewAccount();
                 user.GrantAccess();
                 var clientWithWrongPermissions = await user.CreateClient(Policies.CanViewProfile);
-                var e = await AssertAPIError("insufficient-api-permissions", () => clientWithWrongPermissions.CreateStore(new CreateStoreRequest() { Name = "mystore" }));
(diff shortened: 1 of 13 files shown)


woutersamaey:
Writing the response manually works fine, because there is nothing else doing any writing. Currently the response would be `403 Forbidden` and no body. So if we write a body, it's the only output and there are no conflicts.

NicolasDorier:
I think that's very useful response. I prefer the json object, as people debugging it will more likely see it.

NicolasDorier:
I am looking how to do it more cleanly (with the json result as you did). I may have solution.

NicolasDorier:
Shit, we need something from .net5.0... I feel like I will migrate to .net6

NicolasDorier:
Ok we can't update because EF https://github.com/dotnet/efcore/issues/26629 
I am sick of EF, every update it is just pain in the ass.

NicolasDorier:
https://github.com/btcpayserver/btcpayserver/pull/3198

@woutersamaey can you add a test for this one?

BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -116,6 +119,22 @@ public class GreenFieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
...
                 context.Succeed(requirement);
+            }else{
+                // Solution 1: Just add an extra header.
+                _HttpContext.Response.Headers.Add("X-Missing-API-Permission", policy);
+                
+                // Solution 2: Write the entire response manually.
+                var outputObj = new
+                {
+                    ErrorMessage = "Missing API Permission \""+policy+"\"",
+                    MissingPermission = policy
```
NicolasDorier, line 130: it need to be same format as other greenfield errors:

```json
{
 "code": "insufficient-permission",
 "message": "Missing API Permission blahblah",
 "missingPermission": "blahblah"
}
```

BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -116,6 +119,22 @@ public class GreenFieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
...
             {
                 context.Succeed(requirement);
+            }else{
+                // Solution 1: Just add an extra header.
+                _HttpContext.Response.Headers.Add("X-Missing-API-Permission", policy);
+                
+                // Solution 2: Write the entire response manually.
+                var outputObj = new
+                {
+                    ErrorMessage = "Missing API Permission \""+policy+"\"",
```
NicolasDorier, line 129:
```suggestion
                    ErrorMessage = $"Missing API Permission \"{policy}\"",
```

BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -116,6 +119,22 @@ public class GreenFieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
...
+                    MissingPermission = policy
+                };
+                string output = JsonConvert.SerializeObject(outputObj);
+                var outputBytes = Encoding.UTF8.GetBytes(output);
+
+                _HttpContext.Response.StatusCode = 403;
+                _HttpContext.Response.Headers.Add("Content-Length", ""+outputBytes.Length);
```
NicolasDorier, line 136:
```suggestion
                _HttpContext.Response.Headers.Add("Content-Length", outputBytes.ToString());
```

NicolasDorier: need to add `Content-Type` to `application/json`

BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -116,6 +119,22 @@ public class GreenFieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
...
+                _HttpContext.Response.Headers.Add("X-Missing-API-Permission", policy);
+                
+                // Solution 2: Write the entire response manually.
+                var outputObj = new
+                {
+                    ErrorMessage = "Missing API Permission \""+policy+"\"",
+                    MissingPermission = policy
+                };
+                string output = JsonConvert.SerializeObject(outputObj);
+                var outputBytes = Encoding.UTF8.GetBytes(output);
```
NicolasDorier, line 133:
```suggestion
                var outputBytes = new UTF8Encoding(false).GetBytes(output);
```

woutersamaey:
@NicolasDorier I have 1 last TODO in code. Can you check pls? I want to access the Permission's human readable label, but that is in a controller dictionary.

I was able to access this before, but not from the new model I needed to make. I don't understand what the problem is, but you will :P

NicolasDorier:
`AddApiKeyViewModel.PermissionValueItem.PermissionDescriptions["blahblah"].Item2`

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -55,6 +56,35 @@ public async Task LocalClientTests()
...
+                    var data = await clientWithWrongPermissions.CreateStore(new CreateStoreRequest() { Name = "mystore" });
+                    throw new XunitException("This point should never be reached.");
+
+                }catch (GreenFieldAPIException e)
+                {
+                    Assert.Equal("insufficient-api-permissions", e.APIError.Code);
+                    Assert.NotNull(e.APIError.Message);
+                    Assert.IsType<GreenfieldPermissionAPIError>(e.APIError);
+
+                    GreenfieldPermissionAPIError permissionError = (GreenfieldPermissionAPIError)e.APIError;
```
NicolasDorier, line 82:
```suggestion
                    GreenfieldPermissionAPIError permissionError = Assert.IsType<GreenfieldPermissionAPIError>(e.APIError).APIError;
```

NicolasDorier:
seems like you broke tests

woutersamaey:
@NicolasDorier I found out the reason the tests break is because of `AssertHttpError()` and inside there is 
```
var ex = await Assert.ThrowsAsync<HttpRequestException>(act);
Assert.Contains(code.ToString(), ex.Message);
```
So the tests expect an exception with the HTTP error status as the message (`ex.Message` should be `401`).

Due to my changes, this is now different.
I played with it a bit in my last commit, so you can see which way I'm going, but it's a problem because I cannot access that `401` from the Exception. When I run, `ex.Message` is `Response status code does not indicate success: 401 (Unauthorized).` and not `401` so the Assert fails.

My guess is this is the same problem for all tests...

To test quickly, I simply ran 1 existing unit test, namely `CanCreateUsersViaAPI`. This one is fast, but you can pick any test.

woutersamaey:
@NicolasDorier we have our first Zapier user running into API permission issues. I can implement in Zapier JavaScript so we are ready, but the broken tests I could use some help with. Feedback in previous comment.

NicolasDorier:
Will take over this PR

NicolasDorier:
@woutersamaey the change about HttpRequest has been merged here https://github.com/btcpayserver/btcpayserver/pull/3204 I will rebase this PR.

woutersamaey:
Thanks. I will make the required changes in Zapier JavaScript. 

I was able to help the user in question. He wanted to use a trigger, but did not give the "Manage Webhooks" permission.

This stuff is not obvious to figure out.

When you turn a Zap (=automation) on, the integration auto-registers a new webhook for this specific trigger/event. When you turn the Zap off, the webhook is deleted again.

When a Zap fails, Zapier turns it off. This is normal error handling on their end. 

I will now throw a different type of error based on the missing permission response (if available, meaning you need to have updated your BTCPay instance) so that Zapier realizes it's a permissions error. They wil then not turn of the Zap, but instead send an email to the user and we can define the message.

This way we can tell the user "hey, your zap failed because you did not give permission X."

Everyone will start to party.

PR #3202, opened 2021-12-15
Change implementation of the Smtp server

NicolasDorier:
Many people in support are complaining about the smtp client not working really well.
Turns out SmtpClient of dotnet is [obsolete](https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient?view=net-6.0) and don't support new TLS version (see https://developpaper.com/system-net-mail-smtpclient-failed-to-send-email-through-ssl-tls-protocol/)

>At present, the latest version is TLS 1.3, and other available versions are TLS 1.2 and TLS 1.1. TLS 1.1 is planned to be abandoned in 2020

I chose mailkit instead, a general purpose email library, still maintained nowadays, MIT license, 32.2 million downloads.

A nice positive point: It automatically detect the protocol (SSL, STARTTLS) so we can even get rid of one of the settings! (No more enable SSL checkbox)

### DIFF
--- a/BTCPayServer/Services/Mails/EmailSettings.cs
+++ b/BTCPayServer/Services/Mails/EmailSettings.cs
@@ -77,7 +77,7 @@ namespace BTCPayServer.Services.Mails
             using var connectCancel = new CancellationTokenSource(10000);
             try
             {
-                await client.ConnectAsync(Server, Port.Value, MailKit.Security.SecureSocketOptions.StartTlsWhenAvailable, connectCancel.Token);
+                await client.ConnectAsync(Server, Port.Value, MailKit.Security.SecureSocketOptions.Auto, connectCancel.Token);
                 await client.AuthenticateAsync(Login, Password, connectCancel.Token);
             }
             catch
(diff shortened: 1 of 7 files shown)


NicolasDorier:
I tested with gmail, would be nice if some of you can give a try on other smtp services.

BTCPayServer/Services/Mails/EmailSettings.cs
```diff
@@ -41,42 +44,47 @@ public string From
...
-            client.EnableSsl = EnableSSL;
-            client.UseDefaultCredentials = false;
-            client.Credentials = new NetworkCredential(Login, Password);
-            client.DeliveryMethod = SmtpDeliveryMethod.Network;
-            client.Timeout = 10000;
+            SmtpClient client = new SmtpClient();
+            using var connectCancel = new CancellationTokenSource(10000);
+            try
+            {
+                await client.ConnectAsync(Server, Port.Value, MailKit.Security.SecureSocketOptions.StartTlsWhenAvailable, connectCancel.Token);
```
Kukks, line 80:
```suggestion
                await client.ConnectAsync(Server, Port.Value, MailKit.Security.SecureSocketOptions.Auto, connectCancel.Token);
```

Kukks (changes requested):
LGTM except for the one change I proposed.

I tried using privateemail.com settings and it kept returning "A task was cancelled." As soon as I did the suggestion, it worked

PR #3206, opened 2021-12-17
Fix: DateTime passed to bitpay API weren't assumed UTC, remove DateTi…

NicolasDorier:
Update to .net6.0 bumped to issues with non UTC dates. Entity framework now only support UTC, but it's not a problem since we should only deal in UTC in our code.

This also fix an issue in the GetInvoices call of bitpay's API, where result depended on the timezone of the server.


NicolasDorier:
Update to .net6.0 bumped to issues with non UTC dates. Entity framework now only support UTC, but it's not a problem since we should only deal in UTC in our code.

PR #3217, opened 2021-12-19
Changed API policy for calculating feeRate

woutersamaey:
I don't really see the need for "CanModifyStoreSettings" here. IMHO "CanViewStoreSettings" would be better.
It's just about getting the feeRate, not that you're modifying the store or anything.

This is a sub-permission, so it should not be a breaking change.


Kukks (changes requested):
concept ACK, please adjust the docs (swagger) first though

woutersamaey:
Swagger was updated

Kukks (approved):

PR #3221, opened 2021-12-21
PaymentReq "Created" field is not really a unix timestamp in JSON. Also renamed to "CreatedTime" like with invoices

woutersamaey:
Payment Requests have a field `Created` which should be a unix timestamp in seconds, but it is not. Current result is a string like `2021-12-21T09:21:06.8924122+00:00`, which is wrong.

This change makes it a unix timetsamp + renames `Created` to `CreatedTime` like it is with invoices.

This is a breaking change, but it does make things consistent.

### DIFF
--- a/BTCPayServer.Client/Models/PaymentRequestData.cs
+++ b/BTCPayServer.Client/Models/PaymentRequestData.cs
@@ -8,7 +8,8 @@ namespace BTCPayServer.Client.Models
     {
         [JsonConverter(typeof(StringEnumConverter))]
         public PaymentRequestData.PaymentRequestStatus Status { get; set; }
-        public DateTimeOffset Created { get; set; }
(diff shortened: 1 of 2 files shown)


Kukks (changes requested):
nack on the renaming of the field, ack on the correct time format

woutersamaey:
I left the name as it was.

In either case it is still a breaking change as the output is now a different format.

Kukks:
@NicolasDorier Thoughts?

NicolasDorier:
even if breaking change, it should be unix timestamp. This is not widely used API so that's ok

woutersamaey:
@NicolasDorier I would've still liked the field to be named `CreatedTime` instead of `Created` to be consistent. Since we are breaking BC, is this OK?

NicolasDorier:
Seems ok to me @Kukks  what do you think?

PR #3223, opened 2021-12-21
Added field "StoreId" to a Payment Request in Greenfield

woutersamaey:
Payment Requests do not have a field `StoreId`, so I added it, incl test + Swagger


Kukks:
> Payment Requests do not have a field `StoreId`, so I added it, incl test + Swagger

Is it needed? Considering that the endpoint already has the store id in it?

woutersamaey:
Strictly speaking, no, but we did the same thing for invoices, so I am doing it for consistency.
If we ever get a search endpoint or global list API it would also help.

dennisreimann (approved):

Kukks (approved):

PR #3224, opened 2021-12-21
Store centric UI: Part 3

dennisreimann:
* Store context is kept for Account and Server Settings 
* Store switch links to wallet transactions if wallet is set up - otherwise it links to the payment methods, where the wallet can be set up.
* Integrations now appear in the nav under Plugins
* Pay Button got moved to the sidebar nav under Apps
* Correct nav item highlighting


BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -108,17 +105,27 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
                     if (store != null || isAdmin)
                         success = true;
                     break;
+                case Policies.CanViewProfile:
+                    // FIXME: Is there a better way to check this?
+                    var endpoint = context.Resource as RouteEndpoint;
+                    if (endpoint?.RoutePattern?.RawText.IndexOf("Manage/", StringComparison.InvariantCultureIgnoreCase) == 0 || isAdmin)
```
dennisreimann, line 111: This check is weird and potentially unnecessary, but I wanted to ensure we are dealing with the account settings and not the settings of another user.

NicolasDorier: I don't understand why you need to add this policy?
Why would admin have access to view of all profiles?

dennisreimann: Ok, I'll simply remove it then.

Kukks (changes requested):

BTCPayServer/Views/Wallets/_Nav.cshtml
```diff
@@ -1,20 +1,24 @@
...
-    var network = BtcPayNetworkProvider.GetNetwork<BTCPayNetwork>(wallet.CryptoCode);
+    var walletId = Context.GetRouteValue("walletId")?.ToString();
+    var storeId = Context.GetRouteValue("storeId")?.ToString();
+    var cryptoCode = Context.GetRouteValue("cryptoCode")?.ToString();
+    var wallet = walletId != null ? WalletId.Parse(walletId) : new WalletId(storeId, cryptoCode);
+    var network = _btcPayNetworkProvider.GetNetwork<BTCPayNetwork>(wallet.CryptoCode);
 }
 <nav id="SectionNav" class="nav">
-    <a class="nav-link @ViewData.IsActivePage(WalletsNavPages.Transactions)" asp-action="WalletTransactions" asp-route-walletId="@Context.GetRouteValue("walletId")" id="SectionNav-Transactions">Transactions</a>
+    <a class="nav-link @ViewData.IsActivePage(WalletsNavPages.Transactions)" asp-controller="Wallets" asp-action="WalletTransactions" asp-route-walletId="@wallet" id="SectionNav-Transactions">Transactions</a>
```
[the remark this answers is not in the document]
dennisreimann: Done in c9f0b81.

dennisreimann:
Now all user-facing actions should have policies and hence the store context. 

This PR is ready for review now. Will send a separate last one once this is merged.

BTCPayServer/Controllers/VaultController.cs
```diff
@@ -18,6 +19,7 @@
 namespace BTCPayServer.Controllers
 {
     [Route("vault")]
+    [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanModifyStoreSettings)]
```
NicolasDorier, line 22: why? not needed afaik

BTCPayServer/Controllers/UserStoresController.cs
```diff
@@ -14,40 +15,36 @@
 namespace BTCPayServer.Controllers
 {
     [Route("stores")]
-    [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie)]
```
NicolasDorier, line 17: why did you remove it from the class level down to action level?

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -81,52 +85,107 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
                     if (store != null && (store.Role == StoreRoles.Owner || isAdmin))
                         success = true;
                     break;
                 case Policies.CanViewStoreSettings:
+                    store = await storeT.Value;
                     if (store != null || isAdmin)
                         success = true;
                     break;
+                case Policies.CanViewInvoices:
+                    success = true;
```
NicolasDorier, line 143: I don't understand this, it means that everybody has CanViewInvoices permissions over every store.

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -81,52 +85,107 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
+                    if (usedCookieFallback)
```
NicolasDorier, line 144: I don't understand this.

dennisreimann: Simplified it in 2e7d830.

BTCPayServer/Controllers/UserStoresController.cs
dennisreimann: Fixed in 97e8b83.

BTCPayServer/Controllers/VaultController.cs
dennisreimann: Fixed in 97e8b83.

dennisreimann:
@kukks Please also take a look at 2e7d830: I could simplify it by distinguishing the global and store case on the  [controller level](https://github.com/btcpayserver/btcpayserver/pull/3224/files#diff-06fcf681839e3a4010482db3e64aba26d0e6cab7d207514f7150d8976b8355e4R750). 

So it is either:
- `/invoices/` (unfiltered list)
- `/invoices?SearchTerm=storeId%3AID` (filters that store on the global list without settting the store id cookie value; there can also be multiple `storeId` filters)
- `/stores/ID/invoices` just invoices for that particular store, sets the store id cookie value

I think it is best to approach it like this, because it gives us easy options to handle the different cases of which we can potentially eliminate the global invoices one in the future.

PR #3227, opened 2021-12-22
Fix app permissions

dennisreimann:
As pointed out by @nicolasdorier [here](https://github.com/btcpayserver/btcpayserver/pull/3205#issuecomment-999437555).

### DIFF
--- a/BTCPayServer/Security/CookieAuthorizationHandler.cs
+++ b/BTCPayServer/Security/CookieAuthorizationHandler.cs
@@ -66,7 +66,7 @@ namespace BTCPayServer.Security
                 if (routeData.Values.TryGetValue("appId", out var vAppId))
                 {
                     string appId = vAppId as string;
-                    app = await _appService.GetApp(appId, null);
+                    app = await _appService.GetAppDataIfOwner(userId, appId);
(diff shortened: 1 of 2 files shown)


Kukks (changes requested):
I don't think this is where it should be changed. but rather in `CookieAuthorizationHandler`.

For example:
* If `appId` is present, use `GetAppDataIfOwner` not `GetApp`
* if storeid is pre-set, che check that `app?.StoreDataId== storeId`

This would solve it globally instead of making sure we match it in every usage. Plus it allows us to reduce a call to the DB to re fetch the app data

NicolasDorier:
Agree with @Kukks

dennisreimann:
Ok, but I think there are cases in which GetOwnedApp wouldn't be the right fit if appId is present. How to distinguish those?

dennisreimann:
@Kukks @NicolasDorier you were right, there are no occurrences where those permissions shouldn't be checked, so it is this simple one line change.

I've also added the fix for the login codes view here, which is needed to make the build work.

Kukks (changes requested):
Please also address my second point in the example

PR #3229, opened 2021-12-22
Mentioning Zapier in readme

woutersamaey:


pavlenex (commented):
Let's not try to pick favourites when it comes to integrations. Instead, let's add Zapier to docs and Btcpayserver.org. Not many people pay attention to readme, if we add Zapier we'd have to add a bunch more.

woutersamaey:
Understood, except this is generally useful to everyone as opposed to an integration like Woocommerce that's only for people who run Woocommerce. That was my reasoning.

woutersamaey:
If you feel this is picking favorites, we can add a section "Integrations" and link to the other repos. Sure, why not?

Kukks (changes requested):
This belongs on the website and docs not the readme.

woutersamaey:
Ok, I will aim to add to the docs and on the website instead.

PR #3230, opened 2021-12-23
Renamed Payment Request "Created" field to "CreatedTime" for consistancy since we already have a breaking change

woutersamaey:
New PR to finish what we discussed in #3221


woutersamaey:
@NicolasDorier we have @Kukks blessing :+1:

woutersamaey:
I was a bit surprised only 2 files needed changing, sure this is OK?

Kukks (changes requested):
It's not ok, the build is failing. If you use Rider/VS, you can "refactor rename" which makes sure to change all usages of it

woutersamaey:
Thanks for the advice @Kukks . Turns out it was only 1 occurrence. Let's see if the build succeeds this time.

Kukks (approved):

woutersamaey:
:tada:

PR #3232, opened 2021-12-26
New APIs for dealing with custodians/exchanges, Kraken implemented (deposit + trade + withdraw + check balances + get price + get previous trade info + get previous withdrawal info)

woutersamaey:
@NicolasDorier and @Kukks This is my initial draft for exchange APIs.

Instead of making this purely about exchanges, I've decided to make this about custodians. Custodians hold assets on behalf of the store and can (or not) have an asset exchange feature. This can be a trade vs fiat or other assets.

Given the fact that this space is rapidly evolving, this seamed like a good idea. We could have any party to be a custodian. Typically these would be exchanges, but we could see banks, fintech, CBDCs and even companies like PayPal come into this space and integrate their services. By taking this route, we keep ourselves open to any type of custodial fund storage/exchange.

Which exchanges to support etc can depend on installed plugins, or we could include some ourselves. My idea would be to start out simple and keep code in the core until mature. We can then later move to plugins. If we try to put everything in plugins from day 1, I feel we could run into refactoring hell. My suggestion would be to move to plugins only once things are stable and proven. But feel free to share your ideas.

The API is a WIP and could serve as a foundation for further brainstorm.

Some background thoughts are:
- "Custodians" are companies. We have a list GET call to see all that are supported (could be the result of multiple plugins installed).
- "Custodian accounts" represent a single account with a custodian and are registered on a store. This has an API key and can contain multiple types of assets (different types of fiat + crypto). In the UI this could be presented as a table / ledger with a column per asset.
- A store would probably have a "default custodian/exchange" that would be preferred for quick convert actions.
- Trades can be in amount or percentage
- Trades and transfers can be run in a dry run mode to calculate costs. This could (with time) be used to find the cheapest service to use in an "auto" mode.
- Custodians have a list of booleans that define what they support and what they can do. I've made quite a list, but we can add for sure. This list would mainly be used to support UI and show/hide features to the user. At the very minimum we could even have a "view only" custodian.
- I've only focussed on market orders, so no limit or stop-loss orders, etc. The aim is fiat => crypto and crypto => fiat, not trading. if we want, those can still be added without breaking backwards compatibility.
- There are 2 calls of interest: moving funds and trading funds. A "quick trade" endpoint does both in 1 go. This could be considered an advanced use case and something we could push til later, but the main idea here is that the user just wants to convert asap and does not want to track or do anything manually. BTCPay would in this case watch the transfer and trade when the funds have arrived.
- I have not considered the reverse (shopkeeper has fiat at the exchange and just needs some BTC in his store, for example to make a refund he cannot afford atm). This could also be useful and is the reverse of previous item.
- Some webhooks would also be good to have. They are all in the store scope, so no need to refactor or build global webhooks.
- My POV has mostly been about Zapier and how we can let shopkeepers easily convert from and to fiat. Similarly, every call could just as well be 1 button in the UI or in an app. This is why I included the "Quick Trade" call, even though you could build this by combininig the transfer + webhook + trade calls. It's just more steps, more moving parts and more things that could go wrong. I see no reason why BTCPay should not watch transfers to and from exchanges and do follow-up steps like trade/convert asap.

I made this Swagger file standalone, so you can open it without the other Swagger files. If you want to view as part of the whole API, just remove the first parts which act as a header / general definition. I did find it easier to just view the new stuff and not have it mixed with the existing calls.

Any other ideas are absolutely welcome!


NicolasDorier:
I think that the API must be as simple as possible, so we don't end up in a rabbit hole of trying to catch up with API of custodian changing, nor become a UI to manage custodian accounts. (out of scope)

As such:
1. there should be no `assetBalances`.
2. the specific api to add account need to be different depending on the provider. (Kraken has API key, but other accounts have other APIs)
3. `tradeType` we can leave it out for first version completely.
4. the `asset` properties, depends on the provider, we should not even try to map their currency code. As such, I think we should add a route to show what are the `asset` list for the exchange.
5. `isCostEstimate` isn't possible on most exchange, I think we should leave it out.
6. Forget everything I said about up to now, when we deal with exchange, assets aren't the important things: Currency pair are. Exchanges don't support assets per say, they support trading of pair of assets.
7. Move asset from/to custodian: Leave it out, this is too complicated.

Kukks:
> My idea would be to start out simple and keep code in the core until mature. We can then later move to plugins.

This is a bad idea. Removing code from core to plugin is a huge undertaking and will result in technical debt and/or breaking changes. 

My previous ideas for this style of work lives here: https://github.com/btcpayserver/btcTransmuter/tree/master/BtcTransmuter.Extension.Exchange

woutersamaey:
Okay. Maybe we should organize a call so we can better align?

My point of doing this exercise was so we can bring fiat closer. This means converting from/to and making deposits/withdrawals. 

I'm free most days.

What do you guys think?

woutersamaey:
I've updated the swagger file:
- Lots of simplifications
- Deposits to the custodian are possible by generating an address for a crypto code, identical to how you generate the LN wallet deposit address. In a 2nd step you would create a transaction to this address.
- Removed "quick trade" as it is too advanced.
- Removed "cost estimate" as it is too advanced.
- We have nothing for withdrawals from the custodian to BTCPay at this time

woutersamaey:
More refinements + `index.html` file for easy viewing in the browser.
This `index.html` file will be removed ultimately.

woutersamaey:
I've added some more final touches and made the API spec as high quality as I can (incl sample requests and responses).

Hope you all like it and we can use this as a solid base for discussion and iterate on this.

ndeet:
@woutersamaey nice! 

From the description this is a more general approach but it is based on Zapier needs. I wonder if there is no trading support in Zapier itself, so they do not have plugins that handle Kraken etc? (the answer seems to be no) anyway, if that would be the case you would only need to have wallet transaction (to address x) and withdrawal (get new address) and the heavy lifting is done by separate zapier plugins. On the other hand this would be a unique selling proposition for BTCPay and as you said this is a more general approach for different use cases. (just thinking loud here)

Not yet into C# and no experience here, but making this a separate plugin seems to be much better than putting it into core to keep the core clean and small. (like Kukks suggested)

Thoughts I had when reading the spec:

- What if the Merchant has mostly LN payments.
  - If custodian does not support LN you need a way to move funds to on-chain and then to custodian
  - LN and on-chain BTC both have cryptoCode `BTC` so you might want to use `PaymentMethod` instead?
- On trading you have `from` and `to` assets
  - as Nicolas suggested maybe just have `pair` which is then `BTCUSD` etc? (but you can assemble that under the hood then anyway I guess
- When you send assets to custodian
  - which fees per vbyte will be used, anything built in or should this be configurable too?

NicolasDorier:
I think for LN payment we can wait in later version.
For pair, it is up to the exchange really. Some exchange use BTC_USD, other use BTCUSD some XBTUSD, I think we shouldn't try to process anything here.

Will review today again.

NicolasDorier:
I think to simplify the dev, we can just dev the plugin on the main repo for now.
Then before merging, we'll move it to a separate plugin. Unsure about the right step yet.

woutersamaey:
> I think for LN payment we can wait in later version.

This was my idea as well. Per custodian there is a list of 'supportsXxx'. We will need to compile this list, but it will drive UI (what you can/cannot see/do). LN features should be there as well. 

> For pair, it is up to the exchange really. Some exchange use BTC_USD, other use BTCUSD some XBTUSD, I think we shouldn't try to process anything here.

Correct. The API spec is broad and some computation will need to be handled per custodian/exchange.

Same for the ledger entry results. I've applied some rules to that like item 0 must be the amount sold and item 1 the amount bought in a trade. Costs start at index 2 and up. This too will need some array organization, but it helps the API dev a lot in terms of practicality.

woutersamaey:
> From the description this is a more general approach but it is based on Zapier needs.

Zapier has some limitations, yes, but I was more thinking of the greater good. We have 0 exchange support in BTCPay and I believe this is the first disappointment a new merchant has. I've been asked this. Why can't BTCPay server auto-sell or auto-buy or ... with an exchange? This is the first step in that direction. 

> On the other hand this would be a unique selling proposition for BTCPay and as you said this is a more general approach for different use cases.

Yep! This is the goal. 

> - What if the Merchant has mostly LN payments.

Can be added in stage 2. Unless you see a breaking change?

>   - LN and on-chain BTC both have cryptoCode `BTC` so you might want to use `PaymentMethod` instead?

Yes, may be better!

> - When you send assets to custodian which fees per vbyte will be used, anything built in or should this be configurable too? 

Yes, i should include that along the lines of the existing calls. Good eye!

woutersamaey:
> I think to simplify the dev, we can just dev the plugin on the main repo for now.
> 
> Then before merging, we'll move it to a separate plugin. Unsure about the right step yet.

One very important point here (and this is Zapier specific) is that I'd need the API calls to be in the main project, not dependant on which plugins you have installed or not. 

This way the calls are always available, but may return empty results or always an error if you have no specific custodian/exchanges installed. So basic foundation must be there always + specific exchanges can be added on via plugins. 

This would be best for user experience in Zapier.

Kukks:
> > I think to simplify the dev, we can just dev the plugin on the main repo for now.
> > Then before merging, we'll move it to a separate plugin. Unsure about the right step yet.
> 
> One very important point here (and this is Zapier specific) is that I'd need the API calls to be in the main project, not dependant on which plugins you have installed or not.
> 
> This way the calls are always available, but may return empty results or always an error if you have no specific custodian/exchanges installed. So basic foundation must be there always + specific exchanges can be added on via plugins.
> 
> This would be best for user experience in Zapier.

might be a bit extra, but would an endpoint that specifies what plugins are installed be good enough?

woutersamaey:
> might be a bit extra, but would an endpoint that specifies what plugins are installed be good enough?

Unfortunately no. The Zapier app is 1 app and it cannot show/hide certain actions or triggers based on the BTCPay instance you have. Since everyone's instance is different, this would get out of hand too as everyone with a different set of plugins would result in a different feature set.

The only option would be to create a 2nd Zapier app that is tied to the plugin, but we would need to go through the entire process again of creating the app, filing for publication, testing, etc. This would also be confusing to the user and features would be distributed across Zapier apps. Also the connection/configuration would be double as it would be viewed as a separate app entirely. 

Best would be to have empty/basic API endpoints and then implement different exchanges through the use of plugins. If someone only want Kraken for example, they can just install that 1 plugin.

woutersamaey:
Can I maybe get an ACK on some of the API calls @NicolasDorier or @Kukks ? This way I can (slowly) get started.

woutersamaey:
@ndeet I applied your suggestion to use PaymentMethod instead of CryptoCode in the "Generate deposit address" call

NicolasDorier:
The swagger isn't valid now, test fail.

woutersamaey:
@NicolasDorier the tests fail because the document is work in progress and not supposed to be merged as it is right now.

I moved the draft API docs to a new location `http://127.0.0.1:14142/docs-draft` for easier viewing.

Hope the tests won't break anymore.

woutersamaey:
After talking with @NicolasDorier we concluded that we will:
- Move the list of assets to a separate PR as this is useful for other things as well (i.e. a dropdown for currency/crypto selection)
- Focus on tradable asset pairs instead of `from` and `to` assets and not normalize the assets in BTCPay. The custodian/exchange will need to translate from and to `cryptoCode` on it's own. We will encourage the plugin developer to support basic cryptoCodes like `BTC` and a few others.
- Postpone the implementation of withdrawals and think further on the spec. For now the spec looks okay, but we could be missing some stuff. this is the lowest priority.
- No UI stuff for now, those would be put in separate PRs. At the very minimum we need a form to enter and save your API key + register a custodian account with a store.
- Will implement 1 exchange (Kraken) directly in the controllers and refactor into interfaces when things are stable and we start to see structure. Iterate instead of over-analyzing.
- Will not use any multi-exchange clients. Will just write the JSON by hand for simplicity / maintainability.
- Tests will be using a Dummy custodian, no way to do actual integration tests unfortunately.
- Trade endpoint will lose the `type` field and it will always be a `market` trade. We can implement other trade types later and that should be done on different endpoints as the parameters will be different per trade type. The endpoint can be changed to `/api/v1/trade/market`. Others would be `/api/v1/trade/limit` etc, but those are out of scope.

If I forgot to mention some stuff, I will update as I recall.

pavlenex:
> Move the list of assets to a separate PR as this is useful for other things as well (i.e. a dropdown for currency/crypto selection)

Let's open up an issue for this one, plenty of people requested this as they found the manual entry unintuitive.

woutersamaey:
> > Move the list of assets to a separate PR as this is useful for other things as well (i.e. a dropdown for currency/crypto selection)
> 
> Let's open up an issue for this one, plenty of people requested this as they found the manual entry unintuitive.

Just wanted to point out that @NicolasDorier created a currency dropdown in #3347 . This would be the same data for this API call. I'll propose a new PR for this once this one is done.

woutersamaey:
This PR is now ready for review. It's been a lot of work, but everything is functional.

woutersamaey:
After talking this PR through with @dennisreimann I noticed some flaws in the docs and am fixing these very carefully now. Will ping when ready.

woutersamaey:
The docs are now complete. Please review guys...

woutersamaey:
I've now added complete tests for these endpoints:
- /api/v1/custodians
- /api/v1/custodian-accounts

Next steps would involve faking requests to Kraken's API

woutersamaey:
The new API endpoints are now 100% covered by tests by using a MockExchange with hard coded results. Many issues were uncovered and fixed in the meantime.

The only part that is not covered by tests are the actual Kraken API calls. I'll try to deal with this too.

woutersamaey:
I added tests for Kraken's public endpoints. These did not need to be mocked.

woutersamaey:
@Kukks What's your take on the 2 failed tests? I don't think they're related to the PR, or are they?

Kukks:
> @Kukks What's your take on the 2 failed tests? I don't think they're related to the PR, or are they?

It si fixed on master, rebase

woutersamaey:
Rebase complete. Let's see :D

woutersamaey:
@Kukks this time a test has timed out...

woutersamaey:
Now that Kraken supports LN, I'll try it out and post my findings. 
I don't expect any issues as there is plenty of abstraction, but let's see if things work out for LN as well.

woutersamaey:
Withdrawing from Kraken using LN doesn't seam possible at this time using their API, because they want you to register an LN invoice first and then request the withdrawal. 
Registering new destinations is not possible because they don't have an API endpoint for this. 
They are planning new withdrawal APIs so my guess is they will be adding this soon. 
Once it's released, we can update as needed. 
I do believe some small changes will be required at that time, but nothing we can't manage.

woutersamaey:
I rebased onto master + moved the "draft swagger" in with the main swagger after someone refactored the swagger JSON loading and my solution no longer worked + we needed to merge it anyways...

The new API is documented under the 2 topics `Custodians` and `Store Custodian Account` in the left navbar.

woutersamaey:
I rebased this PR on master and it is working. Will do more checks and look into packaging the plugin

woutersamaey:
Closing in favor of a new, clean PR.

PR #3234, opened 2021-12-27
Consistent and simplified button CTA text

dstrukt:
Just doing some more auditing will working through remaining store-centric tasks. Purely button text, and 1 whitespace cleanup.

I went with the "Create Payment Request" uppercasing vs. "Create a payment request" as this was what was most consistent throughout the app with the exception of a few places in the commits ... but if we prefer the latter, I can make that change as well.

cc @pavlenex @Zaxounette


NicolasDorier:
you don't have to make one commit at every change you know :p

dstrukt:
lol i was wondering if anyone was going to say anything about that .. apologies, that wasn't intentional this time, but was done over the last week or so slowly

NicolasDorier:
I don't mind either case, I can just squash all commits into one when merging.

PR #3240, opened 2021-12-28
Remove internal socks http proxy and use .net60 socks support

NicolasDorier:
Since we are on .net 6.0, we don't need our hacked socks http proxy anymore because .net support socks directly.

Supersede https://github.com/btcpayserver/btcpayserver/pull/1394


Sanjida990 (commented):
Can i pay with it

PR #3244, opened 2021-12-31
Run dotnet format

NicolasDorier:
Just a reformat end of the year cleaning. We don't have that much PR open now.


NicolasDorier:
Just a reformat end of the year cleaning. We don't have that much PR open now.

PR #3252, opened 2022-01-04
Fixed LND admin macaroon hexdump command

woutersamaey:
Argument `-plain` didn't work for me in the LND container, but `-p` did. 
The error I got was `xxd: invalid number 'ain'`

Added the `-c 256` for better readability. This param maxes out at 256.


dennisreimann (approved):
Tested, for me the old and new command yield the same results

woutersamaey:
@d11n I ran this command on the docker container directly using a bash shell. The container image is `btcpayserver/lnd:v0.14.1-beta`. I don't understand why it would work for you, except maybe I had some bad luck with different or a rare combination of data in the macaroon?

dennisreimann:
Ah, I don't log into the docker container, but use the volume contents directly … `/var/lib/docker/volumes/generated_lnd_bitcoin_datadir/_data/admin.macaroon`

dennisreimann:
Why not, it's the same file ;) #reckless

PR #3255, opened 2022-01-05
Setup guide on homepage

dennisreimann:
The basic two steps are displayed, guiding the user to setup a store and afterwards a wallet.

empty

I would not add more steps, because it gets tricky real fast: The current logic checks and sees if there is at least one store with an onchain wallet or Lightning node configured — if that's the case it is assumed the user knows how things are done and we can hide the guide.

TODO: Figure out what to display once the setup steps are done. Ideas for a MVP-dashboard welcome! /cc @dstrukt


dstrukt:
Agreed w.r.t complexity. 

Was trying to map out a more complex setup guide in Figma with the thinking off to the side, and I can almost imagine hearing you sigh when you saw that exploration .. maybe eventually we can help guide users more in the future for specific applications ... but for this release, love the simplicity!

However, w.r.t "Create Store" for no stores. I'm still not completely sure if I'm sold on the in-between state vs. a full page view post sign-up.

Screen Shot 2022-01-05 at 5 10 08 PM

Despite which direction we go, I do love the change to "Create Store" if no store, was going to mention we should address that (as the previous dropdown was odd) if possible, and you're way ahead of me.

If we do go with this direction, small nit, might be worth making the "Create Store" a primary button to draw more attention to the action.

Regardless, great work!

dennisreimann:
Updated the button.

image

Added a recent notifications "widget" to the homepage, for the case that everything is set up and the guide does not get displayed.

image

dstrukt:
Discussed in design call.

Plan to move forward with the latest proposed changes.

ACK!

dennisreimann:
Progressed as discussed in yesterdays design meeting. /cc @dstrukt @Zaxounette 

There is now also a new store home page, where one lands after creating a store. It contains a setup guide similar to the general homepage, but focused on the store. This is supposed to guide the user towards setting up at least one kind of wallet - onchain or Lightning.

store-home-1

Once the setup is done, the "What's Next" part appears. It contains the basic information about the core features and how they can be used. 

store-home-2

Feedback welcome!

dstrukt:
Yes!

1. A question that didn't, and maybe should have, occurred to me in the call .. should the Lightning Node (internal) be enabled by default? To potentially skip the empty state page when clicking on the sidebar tab, and remove the extra step here? Maybe there are details here I'm missing.

Wasn't 100% sure what we decided on that one, but something we could consider as well...?

2. The additional copy is interesting, I like selections you chose, and I think with a little bit of finessing this could be a really solid introduction to the power of btcpay within the app, as we discussed ... will follow up on that section with some simple formatting / think on what you're proposing a bit more.

dennisreimann:
To point 1: We can only do this under certain circumstances – by default the Internal Node is only available to the instance admin. 

I guess this is also the reason why we haven't set it up automatically yet.

pavlenex:
This is excellent work. I'm sure in the future we may consider ironing it out based on feedback, but as minimum viable product, this is imo doing what it's supposed to do.

I am a bit unsure about the `what's next section` mostly because it's text heavy and we still haven't really cleaned things up when it comes to apps, plugins, etc, so introducing all these concept in this way, to me, is a bit too much. First of all, text heavy means more things for translators to translate, and even then these will most likely change in the future.

So I propose that what's next, at this stage of us coming up with an MVP is a bit out of scope. I believe the proper time to add this would be after the cleanup (apps in plugins, etc)

dennisreimann:
@pavlenex I agree that it seems too text heavy, but also thought it might be better than nothing -- however, we can remove it for the first iteration and wait for what the dashboard brings.

As for the setup guides Home vs. Store Home: It's a bit tricky, because on the general homepage we have to deal with the possibility of multiple stores. So the check there is more like "is there a store? If so, is there at least one store with a wallet or lightning node?" (assuming the user know the setup process in case he has configured one at least once).
On the Store Home we can focus on that particular store, hence also guide the user towards setting up both, onchain and Lightning. That's my reasoning behind that difference.

pavlenex:
I think the problematic thing there is possibly naming `Home`, I expect to go whatever home is when I hit the logo, and was very annoyed that `home` in store kept leading me to something that didn't feel like home. So possibly `dashboard` vs `launchpad` naming would have to be done there to differentiate what we mean, otherwise the flow might be a bit odd to people. I'm not sure if this has been discussed on the  design meeting and decided on, so I'm just raising my two cents that it may not be what people are expecting to get.

When I hit home, I'd expect to land on a dashboard, but perhaps store check should be `launchpad` or similar, as I once[ pointed out on the UX chat](https://chat.btcpayserver.org/btcpayserver/pl/qn3eiskmg7fiumiubn4ee9pk9r) , but I am not even entirely sold this is clear, so def requires wider discussion.

dennisreimann:
Agree and think we can call it dashboard once we have something meaningful for the state when the store is set up.

dstrukt:
The original intention was to provide a "Home" for each store, where you will eventually see all relevant metrics (for that day, week, etc..) and other actions that will evolve over time, to start orienting the user to this tab.

Clicking on the logo will take you to the "Home" tab of the store instance you're currently in, as well as clicking on "Home" on the sidebar, intended functionality.

However, I could also see "Dashboard" being used as well, and prefer that to "Launchpad" personally.

pavlenex:
Just noticed that I get 403 when trying to set up a Lightning Node before the actual on-chain wallet.

NicolasDorier:
Can you also rebase, the integration tests error is on CanSendIPN which was flaky because of a bug I fixed.

BTCPayServer/Views/Home/Home.cshtml
```diff
@@ -1,84 +1,104 @@
...
-                <p class="text-muted mb-0">The payment server does not need to know your private keys, so your money can't be stolen.</p>
-            </div>
-        </div>
-        <div class="col-lg-4 col-md-6 text-center">
-            <div class="py-4 service-box">
-                <img src="~/img/qr-logo.png" class="mb-2" alt="" asp-append-version="true" />
-                <h3>Easy</h3>
-                <p class="text-muted mb-0">A user-friendly Bitcoin checkout page for your customers.</p>
+        <div class="col-lg-9 col-xl-6">
+            <p class="lead text-secondary">To start accepting transactions, you need to set up a store and wallet.</p>
```
pavlenex, line 47:
```suggestion
            <p class="lead text-secondary">To start accepting payments, set up a store and a wallet.</p>
```

BTCPayServer/Views/Stores/Dashboard.cshtml
```diff
@@ -0,0 +1,118 @@
...
+
+<h2>@ViewData["Title"]</h2>
+
+@if (isReady)
+{
+    <p class="lead text-secondary">This store is ready to accept transactions, good job!</p>
+}
+else
+{
+    <p class="lead text-secondary">To start accepting transactions, you need to set up a wallet or Lightning node.</p>
```
pavlenex, line 40:
```suggestion
    <p class="lead text-secondary">To start accepting payments, set up a wallet or a Lightning node.</p>
```

pavlenex:
Just quickly tested things out again, good improvements 🎉 

This flow to me still feels very weird. I don't really understand why the setups are different between `dashboard` and `home button`. Perhaps I am still misunderstanding the big idea behind it.

dennisreimann:
@pavlenex I described it [here](https://github.com/btcpayserver/btcpayserver/pull/3255#issuecomment-1008125529). What do you recommend?

pavlenex:
@dennisreimann Haven't thought to much about it, but it seems that setup flows should at least try to be consistent.

As you can see in a gif [here](https://github.com/btcpayserver/btcpayserver/pull/3255#issuecomment-1011026274) when I have no stores, and I try to set up my first store, the setup flows seem different. On dashboard I have the ability to set up wallet directly leading me to on-chain screen, or lightning, directly leading me to the LN node. Comparing it to the `store home` view I only see wallet setup which leads me to the payment method page. 

I think the first flow is a bit more intuitive. Is there a reason why these two setup flows are different?

One more thing I noticed, is when I have a store already and want to set up a new one, the `dashboard` setup flow no longer appears and I only see notifications even though i haven't set up the store yet.

NicolasDorier:
Is it normal: 

image

NicolasDorier:
image

NicolasDorier:
You think improving in next PR? it is a bit strange, Dashboard doesn't have an icon and done steps don't have a check mark.

image

Also, I think in maybe later PR, when a step is setup by clicking on the dashboard, when the step is done, it should return back to the dashboard, not to the store's settings.

NicolasDorier:
It seems to me the ListStores page should be removed at this point?

BTCPayServer/Controllers/UserStoresController.cs
```diff
@@ -82,7 +82,7 @@ public async Task<IActionResult> DeleteStorePost(string storeId)
         }
 
         [HttpGet]
-        [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanViewStoreSettings)]
+        [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanModifyStoreSettingsUnscoped)]
```
NicolasDorier, line 85: There shouldn't have any Policy here I think.

dennisreimann: If we omit it, the cookie auth handler does not populate CurrentStore. Shall I remove the policy for the list and create actions and resolve the CurrentStore inside the actions?

dennisreimann:
The missing icons are a caching problem, can you do a hard reload or clear the cache?

The disabled Internal Node isn't visible, because we switched the colors of the sidebar and main content – the latter having a plain white background now. Will it stay this way, @dstrukt?

NicolasDorier:
@dennisreimann the problem was indeed caching. That's weird though, why would have a caching issue? I think all resources have a `v=` query string to prevent this.

NicolasDorier:
Ah ok, not in the svg. I will check if I can't add it automatically

NicolasDorier:
Shouldn't have anymore cache issue with https://github.com/btcpayserver/btcpayserver/pull/3298

PR #3260, opened 2022-01-07
Improve Payout View

dstrukt:
There's still quite a bit I want to do with this view, but just made a few small aesthetic changes for the time being as it's becoming a more prominent view in the latest 1.4.0 update.

- Improve the empty state view to hide options if there are no entries.
- Change the tabbed view
- Moved the "actions" selector above the tab view, but would like to incorporate the action with the tab view, but might need help with this (will provide a basic mock)
- Removed the backgrounds of the tab lists

Screen Shot 2022-01-06 at 4 54 34 PM
Screen Shot 2022-01-06 at 5 00 09 PM

Edit: @dennisreimann Open to whatever else you want to improve, but I'd be comfortable merging this.

Edit: Fixed the alignment issue.


pavlenex:
Concept ACK.

dstrukt:
Fixed a row/column issue, should be good to go barring a code review.

dennisreimann (approved):
Updtaed it, so that the filters are present, regardless of the list of payouts being populated. As the `Model.Payouts` are only representative of the current filter, we'd otherwise use the ability to change the filter in case the current list is empty.

dstrukt:
Ooo, nice catch, much cleaner!

Updated a spacer above below the tabbed nav, we should be good to go.

dstrukt (commented):
Can we keep the `mb-4`? It looks better imo, unless there was a particular reason you wanted to use `mb-3`.

dennisreimann:
The item below had a negative margin of 1, which pulled it up. I removed it, so that the math fits again :)

dennisreimann:
`CanUseTorClient` failure is unrelated. Good to go, imho!

dstrukt:
Ahhh I see, beautiful.

Sned it!

PR #3261, opened 2022-01-07
Rename controllers

NicolasDorier:
Trying to do some cleaning of our code that may break other people PR while there isn't a lot.

This PR rename controllers to follow the following convention:
* Greenfield controller are prefixed by Greenfield (GreenfieldInvoiceController.cs)
* UI controllers are prefixed by UI (UIInvoiceController.cs)
 * Bitpay controllers are prefixed by Bitpay (BitpayInvoiceController.cs)

It will make browsing the code more easy.

This PR is hugely disruptive. I don't like doing those kind of whole codebase change, but I figured out that we don't have too much on our plate these days. No much PRs around, so this may be ok.

It is probable that all plugins will break after that and will need to be recompiled. We can't realistically ensure all or changes don't break plugins, so I think doing a sudden clean cut like this is better than breaking at every update caused by incremental changes.

image

Before this change, all controllers have same name, making it hell to browse the right one.


woutersamaey:
So you prefixed the UI related controllers, is that it?

NicolasDorier:
and greenfield and bitpay controllers

woutersamaey:
Awesome improvement! LGTM!

dstrukt:
Agreed, this is a nice improvement!

pavlenex:
Let's try to do this after #3255 so we don't cause a chaos on the important PR we need in for the next release.

NicolasDorier:
Yes, I will keep rebasing this PR

NicolasDorier:
boom!

PR #3264, opened 2022-01-07
Update Invoice Views

dstrukt:
Few minor changes:

- Updated the Amount and Currency to match the other instances in which we do this:
Screen Shot 2022-01-06 at 11 37 42 PM

- Moved the search to the left, actions to the right to be consistent with all of the other views.
- Cleaned up some of the padding on the collapsable element
- @dennisreimann I broke the "MassAction" <form>, the way the HTML was structured allowed the <form> to wrap both the action buttons and the table, but in order to achieve this updated view, I had to rearrange a few things, but I'm not 100% sure how to reconcile this so the actions no longer trigger the form. This PR is a low priority compared to everything else, but I bet you'd see a solution immediately.

Screen Shot 2022-01-07 at 1 17 09 AM


NicolasDorier:
fixed conflict

pavlenex:
Concept ACK.

dstrukt:
Thank you!

And sweet, just need to address the broken "MassAction" functionality, have pinged @dennisreimann, and once that's resolved, we should be good to merge.

dennisreimann:
Fixed the view and also added tests and batch unarchiving.

The help text/section now gets opened above the search input.

@dstrukt The markup is a bit tricky here, because the table needs to be part of the mass action form. That's why I pulled it up using CSS for breakpoints LG and above. Also improved the display for smaller breakpoints, please take one more look before we merge this, to ensure I haven't messed up anything you wanted to fix :)

image

dstrukt:
Nice, we may want to revisit the filter section in the future as we continue to refine this view, have a few ideas, low priority, but for now this works!

I like the subtle (?) move to the left of the action buttons.

LGTM!

PR #3267, opened 2022-01-07
Account menu improvement

dennisreimann:
As discussed in yesterdays design meeting, some changes to the account menu UI: This condenses the bottom part of the sidebar and aggregates the account related functions in a dropdown.

account

In Action:

account


dstrukt (approved):
Will continue to play around with it, but from an implementation standpoint this LGTM.

I love it!

What are your thoughts, do you like it?

dennisreimann:
Yes, way better now. 😁

PR #3268, opened 2022-01-07
Add Services to sidebar

dennisreimann:
WIP state with Services menu in sidebar.

### TODO

- [ ] Define which services (currently only the Lightning related appear)
- [ ] Add Icons for the services

### Screenshot

image


dstrukt:
Definitely makes sense to extract these, but was originally thinking these could go on the Lightning Node view as opposed to the sidebar?

Unpolished original mock:

Screen Shot 2022-01-07 at 1 06 10 PM

Thoughts?

dennisreimann:
Yes, they'll go there for sure. Wanted to discuss the menu first, because @NicolasDorier brought it up and I found it reasonable, as these are often used and we don't offer a Lightning UI ourselves.

pavlenex:
Concept ACK.

I think showcasing `services` `plugins` and `apps` in a single sidebar makes sense until we clean things up, even though it may seem overwhelming, bear in mind, that these will show only to the server admin, so the user who is tech savvy most likely and is familiar with BTCPay's in's and out's. Regular user of an instance would not see these.

dstrukt:
Ahh I see.

It does change my opinion a little bit if it's solely for the admin, and might be a reasonable middle ground for the 1.4.0 MVP.

I still think the most reasonable place for them to ultimately end up would be something like the mock above for context and orientation, however, I suppose this is also a bit dependent on what we hope to do long term w.r.t Lightning.

Just debating if it was worth adding this page now, as opposed to later, if we have intentions to change in the near future, but definitely something worth discussing on Thursday!

dennisreimann:
> I still think the most reasonable place for them to ultimately end up would be something like the mock above for context and orientation

Will add that page next ;)

pavlenex:
There are services that aren't lightning though, what will we do with them, they cannot be on a different page? Pretty much many of things from [here](https://docs.btcpayserver.org/Docker/#generated-docker-compose) are services in a way?

I think if our goal is for users to quickly use LN, then it makes sense to plug in RTL and similar services under `Lightning`. However I don't see this as a long-term solution, at some point we'll have to come up with our own simple UI that, hopefully in the long run would play nicely with the wallet's UI. The unification of on and off chain should probably by the long-term vision, users shouldn't have to thinker if they're using bitcoin or bitcoin ln.

dstrukt:
That's a good point, I originally only anticipated extracting the Lightning specific service to it's respective view for 1.4.0, not necessarily every Service ... but if we did continue to extract every Service, it would be within the context for which it is applicable, essentially doing away with the notion "Services" as it isn't the most intuitive UX. 

e.g. clicking on RTL or LT is just a shortcut to manage your funds as opposed to some service, even if that's how we treat it under the hood.

Thus, we'd extract one section at a time, with the the Lightning specific ones getting the first extraction, seems the most requested functionality, but could be wrong on this ... to see how well it works.

Agree with the long-term vision of L/BTC, and I think as we start to extract and isolate the most necessary tools and components to their respective places, merging and encapsulating them as we move forward will become easier .. as we've seen, it's been a bit difficult to consolidate everything, but the paths forward are becoming clearer.

PR #3273, opened 2022-01-09
Simple Lightning Node Selector Improvement

dstrukt:
Never liked how it was just simple text only for the unselected tab, and we didn't create a custom component for this, so I thought the rounded tabs could be an effective method for the time being.

Will likely carry this out to the payouts view as well, but we might want to separate the CSS.

cc @dennisreimann This might be a component we'll want to create for the future when we shift our focus to the design system work.

The secondary vs. primary text color bothers me a bit on this screen as well, but I left it alone, feels like the current secondary text (top) should be white as well...?

Sidenote: Going to try to document which type of component should be used in each situation, e.g. (button, rounded tab, line tabbed, etc..). In the meantime, trying to make sure it's at least consistent.

Screen Shot 2022-01-09 at 3 04 40 AM

### DIFF
--- a/BTCPayServer/Views/Stores/SetupLightningNode.cshtml
+++ b/BTCPayServer/Views/Stores/SetupLightningNode.cshtml
@@ -53,6 +53,11 @@
             color: var(--btcpay-secondary);
             opacity: .5;
         }
+        
+        #LightningNodeTypeTablist label + input + label {
+            margin-left: var(--btcpay-space-m);
+        }
+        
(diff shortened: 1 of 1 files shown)


dennisreimann:
Nice! I changed the spacing to use CSS instead of the Bootstrap class, because there's an extension point which e.g. LNbank uses. This way the plugins don't have to know or care about the spacing and we can easier make this a component later on.

image

dennisreimann (approved):

PR #3284, opened 2022-01-11
[Greenfield] Fix create/update payment requests expiry field and allow null currency field (Fix #3222)

### LINKED ISSUE #3222: Greenfield API: Creating a Payment Request does not work
**Describe the bug**

Cannot create a payment request using Greenfield API.
```
curl -X POST -v --header "Authorization: token YOUR-API-KEY" --header "Content-Type: application/json" --data '{  "amount": 5,  "title": "Zapier Test Payment Request",  "email": "8lUZyjrf@example.com",  "description": "My Description",  "expiryDate": 1642669393,  "embeddedCSS": "body: { background: red; }",  "allowCustomPaymentAmounts": false}' https://yourserver.com/api/v1/stores/xxx/payment-requests/
```

Error
```
[{"path":"expiryDate","message":"Unexpected character encountered while parsing value: 1. Path 'expiryDate', line 1, position 141."}]
```

If I set the `expiryDate` parameter to `NULL`, it works.

According to the docs, the  `expiryDate` is expected in unix timestamp seconds, so it should work.

**To Reproduce the bug**
1. Run the cURL command above, but use your API key + server URL + store ID

**Expected behavior**
Should create a Payment Request

- BTCPay Server Version:    master branch
- Deployment Method:        Docker

NicolasDorier:
The expiry field wasn't using the proper jsonconverter.
On top of it, we required the `currency` property to be set, but an empty currency (store's default) should be allowed.

I also updated the edit payment request page to not require the currency.

Fix #3222

### DIFF
--- a/BTCPayServer.Client/Models/PaymentRequestBaseData.cs
+++ b/BTCPayServer.Client/Models/PaymentRequestBaseData.cs
@@ -12,7 +12,8 @@ namespace BTCPayServer.Client.Models
         [JsonProperty(ItemConverterType = typeof(NumericStringJsonConverter))]
         public decimal Amount { get; set; }
         public string Currency { get; set; }
-        public DateTime? ExpiryDate { get; set; }
+        [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
+        public DateTimeOffset? ExpiryDate { get; set; }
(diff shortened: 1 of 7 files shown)


woutersamaey:
You forgot to update the docs, so I did this for you in #3285

PR #3285, opened 2022-01-11
Swagger: Added explanation for empty currency field in Create Payment Request

woutersamaey:


NicolasDorier:
so the reason I didn't declared currency as null is that the `currency` field is only optional for create/update. But never null when you do a GET.

NicolasDorier:
I actually changed the doc, but forgot to push before merge :/

woutersamaey:
> so the reason I didn't declared currency as null is that the `currency` field is only optional for create/update. But never null when you do a GET.

I think we can make the distinction if we define 2 models in Swagger. 1 for create/update and 1 for get. Both are the same except the nullable field. Or is that not what you meant?

NicolasDorier:
yeah we should define 2 model in swagger normally.

PR #3290, opened 2022-01-11
Improve footer

### LINKED ISSUE #2692: HTTP Onion Header ignored for local install
I have BTCPay Server installed on a rockpro64.

When opening it via Brave, with local ip or local domain, I don't see the `Open in Tor` option in brave.
This might come from the fact the page is not https.

But this is problematic, because there is no easy way to discover the onion address of your install outside going to server settings / services.

Maybe we should add again the Tor link in the login page...
That said, it seems to happen only if it is opened on HTTP. To make sure we don't get black listed again by google safe browsing, maybe we should show the Tor link on login page only if http.

dennisreimann:
* Extract footer into partial
* Add footer to simple layout (unauthenticated pages like login/register and error pages)
* Add "Copy Tor URL" link to footer 

Closes #2692.

### Login page

image

### Signed in app footer

image


dstrukt:
LGTM!

I debated if we should keep the icons, a little less noisy without, but I feel with the introduction of Tor .. it feels more, necessary?

PR #3296, opened 2022-01-12
Fix clipboard copy edge case

### LINKED ISSUE #3291: [Bug] Copy to clipboard doesn't work on receive address in Brave if HTTP (with domain other than localhost)
I host BTCPAY on rockpro64, 1.3.7.

If I click on the Copy button in the Wallet/Receive, it doesn't work.
However, the copy work on the checkout invoice.

dennisreimann:
Closes #3291.

### DIFF
--- a/BTCPayServer/wwwroot/js/copy-to-clipboard.js
+++ b/BTCPayServer/wwwroot/js/copy-to-clipboard.js
@@ -1,19 +1,35 @@
+const confirmCopy = (el, message) => {
+    el.innerText = message;
+    setTimeout(function () {
+        el.innerText = el.dataset.clipboardInitialText;
+    }, 2500);
+}
+
 window.copyToClipboard = function (e, data) {
-    if (navigator.clipboard) {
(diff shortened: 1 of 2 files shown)


BTCPayServer/wwwroot/js/copy-to-clipboard.js
```diff
@@ -1,19 +1,35 @@
...
+        copyEl.style.position = 'absolute';
+        copyEl.style.opacity = '0';
+        copyEl.value = data;
+        document.body.appendChild(copyEl);
+        copyEl.select();
+        document.execCommand('copy');
+        copyEl.remove();
+        confirmCopy(confirm, message);
     }
+    item.blur();
```
NicolasDorier, line 32: should blur really be here??

dennisreimann: Yes, otherwise the button or link stays focused and keeps the highlighted styles.

PR #3298, opened 2022-01-13
Add version to svg's href files to prevent cache issues

NicolasDorier:
We make extensive use of `<svg><use href>` pattern to reference various icons in the app.
However, the `href` doesn't have a `v=<hash>` appended to the reference.
So if the svg file change, the browser cache will serve an old version resulting in apparent bug such as https://github.com/btcpayserver/btcpayserver/pull/3255#issuecomment-1011799700

`asp-append-version="true"` isn't usable on this tag, so I decided to change our svg tag helper to append version automatically.

### DIFF
--- a/BTCPayServer/TagHelpers.cs
+++ b/BTCPayServer/TagHelpers.cs
@@ -3,10 +3,15 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):
Nice, didn't know this alternative existed! Thanks for fixing it :)

PR #3308, opened 2022-01-14
Add Lightning page with services

dennisreimann:
For now the Lightning landing page is plain and simple, containing only a list of services to manage the connected Lightning node.

We'll extend this after v1.4.0, but for now this is already better than throwing the user directly into the Lightning settings once the node is configured.

lightning

The second commit adapts the existing services pages.


dstrukt:
I'm very happy we were able to get this into the 1.4.0 release, the amount we've covered from the original mocks is wild.

We'll follow up with a polish pass for this and Wallet view in later releases.

Great work, LGTM!

BTCPayServer/Controllers/UIStoresController.LightningLike.cs
```diff
@@ -15,7 +18,61 @@ namespace BTCPayServer.Controllers
...
         [HttpGet("{storeId}/lightning/{cryptoCode}")]
+        public async Task<IActionResult> Lightning(string storeId, string cryptoCode)
+        {
+            var store = HttpContext.GetStoreData();
+            if (store == null)
+                return NotFound();
+            
+            var services = _externalServiceOptions.Value.ExternalServices.ToList()
+                .Where(service => _externalServiceTypes.Contains(service.Type))
+                .Select(service => new AdditionalServiceViewModel
```
NicolasDorier, line 41: Love nature and save trees, by removing this view model. You can just use ExternalServices directly.

dennisreimann: As it's a mixture of `ExternalService` and `OtherExternalService` (see below) I used the extra `AdditionalServiceViewModel` to treat them uniformly. If I'd want to reuse the existing models, I'd have to distinguish them in the `LightningViewModel` and also the view.

BTCPayServer/Views/UIStores/Lightning.cshtml
```diff
@@ -0,0 +1,65 @@
...
+           asp-controller="UIPublicLightningNodeInfo"
+           asp-action="ShowLightningNodeInfo"
+           asp-route-cryptoCode="@Model.CryptoCode"
+           asp-route-storeId="@Model.StoreId"
+           target="_blank">
+            Public Node Info
+        </a>
+    </div>
+    
+    @if (Model.Services.Any())
```
NicolasDorier, line 40: if external node type, it shouldn't show either

dennisreimann: Updated in a49bebf,

BTCPayServer/Views/UIStores/Lightning.cshtml
```diff
@@ -37,14 +37,14 @@
         </a>
     </div>
     
-    @if (Model.Services.Any())
+    @if (Model.Services != null && Model.Services.Any())
```
NicolasDorier, line 40: In the future, you can use `Model.Services?.Any() is true` instead

PR #3313, opened 2022-01-14
Home sweet home

dennisreimann:
Updates to the home behaviour and logo link, as discussed in yesterdays design meeting:

* Display global homepage only if there is no store: It then contains the setup guide to create a store.
* Logo links to global home only if there is no store.
* Once there is a store, the logo links to the dashboard of the currently selected store 
* Home redirects to the currently selected store, if there is no selected store but at least one store, redirect to the first of the users stores.

We might change this when we create a global instance dashboard.
Until then, we discussed this to be the desired behaviour of the homepage and logo link.

### DIFF
--- a/BTCPayServer.Tests/Extensions.cs
+++ b/BTCPayServer.Tests/Extensions.cs
@@ -23,7 +23,6 @@ namespace BTCPayServer.Tests
             s.Driver.FindElement(By.Id("Email")).SendKeys(email);
             s.Driver.FindElement(By.Id("Password")).SendKeys("123456");
             s.Driver.FindElement(By.Id("LoginButton")).Click();
-            s.Driver.AssertNoError();
         }
 
         public static void AssertNoError(this IWebDriver driver)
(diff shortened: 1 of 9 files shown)


NicolasDorier:
it seems breaking tests

dennisreimann:
@NicolasDorier Working now :)

BTCPayServer.Tests/Extensions.cs
```diff
@@ -23,7 +23,6 @@ public static void LogIn(this SeleniumTester s, string email)
             s.Driver.FindElement(By.Id("Email")).SendKeys(email);
             s.Driver.FindElement(By.Id("Password")).SendKeys("123456");
             s.Driver.FindElement(By.Id("LoginButton")).Click();
-            s.Driver.AssertNoError();
```
NicolasDorier, line 26: why remove this?

Fais5012: bitcoin:bc1q9xu0ztcwh2fmrxesmf2vx7phwy7hvcdepa95s9?amount=0.00262595

dennisreimann: I checked again and it was me not properly handling the store guest case. Resolved with c1a878f.

PR #3314, opened 2022-01-15
Theme colors

dennisreimann:
CSS updates and added styling for scrollbars.


dstrukt:
LGTM.

I think this is a good improvement until we can address a few tweaks to Light Theme colors .. now that we're introducing a few more colors to the view, it's clear it might be best to get a little bit more contrast in there, need to do some tests on my end.

Will catch a few of the other items on the list (Dark Mode inputs, etc..) with that update as well.

dstrukt (approved):

Issue #3180, opened 2021-12-06
1.4.0 Store-Centric RC UI/UX Issues


dstrukt:
Wanted to maintain for visibility and prioritization in parallel for issues out of scope of (#3091) Will update as issues are addressed, this is still a bit of a mess as we work through the remaining bits of a large PR.

P1: essential
P2: ideal, make decision on
P3: nice to have, but could release without

### Sidebar:

- [x] P1: Remove global context switch
- [x] P1: Apps don't switch based on store selection
- [x] P1: Wallet Empty state before configuration (See Figma).
- [x] P1: The "payments > wallets" tab item view should replace the top "wallet" view after creation instead of a new tab (per Figma mocks), will lead to dedicated Wallet page (with settings, etc..)
- [x] P1: Lightning Node Landing View Update (See Figma .. might be only P1 issue we don't resolve)
- [x] P1: "Wallet" sidebar item should still be selected on all tab views (only works on "Settings" tab currently)
- [x] P1: "Invoices" sidebar has no active state
- [x] P1: "Account" sidebar has no active states, for any of the subpage items
-
- [x] P2: limit the number of characters in the Store Selector
- [x] Fixed sidebar width? (240 or 280px)
- [x] Add ellipses on long display App titles (see example below) (@dennisreimann)
- [x] P2: Combining Account/ThemeSwitch/Logout into a functional component to reduce # of items on left sidebar
- [x] Add store icon to the StoreSelector dropdown (@dennisreimann)
-
- [x] P3: make left/right padding changes on sidebar (as mentioned on PR by @dstrukt)
- [x] P3: Remove plugin "status" element
- [x] P3: Left align store copy (name) in StoreSelector dropdown element (@dennisreimann)
- [x] P3: Change margin-top above StoreSelector from -s to -m (@dstrukt)

### Content:
- [x] P1: Hide/Remove the deprecated Payment Methods > Wallet component (@dennisreimann)
- [x] P1: Hide/Remove the deprecated Payment Methods > Lightning component (@dennisreimann)
- [x]  P1: remove tabbed navigation for payouts, pay button, and pull payments
-
- [x] P2: "Create app" is a modal .. or a full page? Modal might be more fitting.
- [x] P2: Tabbed nav mobile & tablet improvement (@dennisreimann)
- [x] P2: clean up the payouts view (@dstrukt - in progress)
- [x] P2: Noticeable time delay in creating apps, feel like this has been an issue for a while if I remember
[...]

PR #3316, opened 2022-01-16
Content Consistency Updates (1.4.0)

dstrukt:
Might make sense to merge this after @dennisreimann's latest changes.

These are just aesthetic changes, header sizes, paddings, etc... to make things consistent, been a lot easier to spot with the latest updates.

A few minor tweaks I'd like to improve with this PR, but I need some help with:

- [x]  Fix active state when "Account" is selected + sub-navigation @dennisreimann 
- [x]  "Profile" -> "Account" on Account tabbed navigation
- [x]  Change "General Settings" to "General" in Store Settings & make it the first tab
- [x]  "Email Server" -> "Email" in Server Settings in tabbed menu, change the content title from "Email" -> "Email Server"

The first item will close out one of the issues (https://github.com/btcpayserver/btcpayserver/issues/3180).

Added "Language" section:
Screen Shot 2022-01-15 at 4 28 34 PM

Moves API Key CTA to top right:
Screen Shot 2022-01-15 at 5 45 45 PM


pavlenex:
To answer questions, I agree with all except these two:

> Move "Language" section in "Checkout Appearance" to "General" in Store Settings

No, because this isn't a store language it's the language of the checkout page, so it's appropriate where it is.

> "Checkout Appearance" -> "Checkout" in Store Settings tabbed menu

I don't think we should change it, it communicates what it really is, the appearance of your checkout page. If we do `checkout` it may confuse people thinking it's payment method setting or something else.

dstrukt:
Ahh good call on the Language, thought this was App wide, noted.

That's a good point, I think I just took issue with it because there were multiple types of modifications, not just Appearance, but maybe it's just because that word doesn't feels like it captures everything, but I have no alternative suggestions, so will drop the issue.

NicolasDorier:
This probably break something in the CanSetupMail test ping @dennisreimann

dstrukt:
Oh, wow the account text switch was more simple than I anticipated, thought more would be involved.

dennisreimann (approved):
All TODOs done now :)

The failing link check is unrelated - this one should be merged next, I'll then finish #3323 based on this.

PR #3319, opened 2022-01-17
Make sure the SubscribeAsync handlers are executed in order

NicolasDorier:
When using `eventAggregator.Subscribe`, it is expected that the handler treat events sequentially.

However, `eventAggregator.SubscribeAsync` wasn't treating the events sequentially, and we would occasionally get events sent out of order, making `CanSendIPN` fail time to time.

This PR fix `eventAggregator.SubscribeAsync` so that the handler always treat one event at a time. Also, it await the `Notify` of the `BitpayIPNSender` to make sure that events are sent sequentially.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -569,7 +569,7 @@ namespace BTCPayServer.Tests
             bool paid = false;
             bool confirmed = false;
             bool completed = false;
-            while (!completed || !confirmed)
+            while (!completed || !confirmed || !receivedPayment)
             {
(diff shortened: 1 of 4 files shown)


NicolasDorier:
image
image

NicolasDorier:
this bug is haunting me, it doesn't bug when I put debug statements...

NicolasDorier:
So I found the root cause: The `receivePayment` event is sent by the nbxplorer listener. While the other events are sent by the `InvoiceWatcher`. So we can't guarantee ordering of receivePayment.

I think this is something we need to eventually fix.

PR #3320, opened 2022-01-17
Guest of a store should be able to browse invoices

NicolasDorier:
Before this PR, the store's guest couldn't do anything, because he would be redirected to a 403 page on selecting a store.


dennisreimann:
I just fixed this in #3313.

PR #3323, opened 2022-01-17
Remove redundant payment methods from store settings

dennisreimann:
Now that we have the payment methods and access to their settings in the navigation and sub navigation, their display on the payment methods page is obsolete. We can remove them there and call the page Payment, as only the payment configuration is left.


pavlenex:
I think this is a good idea, two things:
1. It's weird that store menu appears in Lightning Settings, where you cannot navigate to lightning settings from a store menu. Should we remove that menu? I am not sure what to do but if feels a bit odd. I assume that the long-term vision for this one is to have LN UI with settings page, just like we do with an on-chain wallet? Screenshot 2022-01-17 at 11 39 43

2. With this cleanup, I think we should just put `payments` into general as a `payments` section it would feel more natural to have general as the first option and reduce the amount of pages/nav options.

Screenshot 2022-01-17 at 11 39 56

dstrukt:
Awesome, it's finally happening!

And yes, we should remove the tabbed navigation on the Lightning Settings view, and yes @pavlenex that's the vision!

Definitely agree General should be first / default page .. I guess it makes sense to add Payments on that view as well .. but there's still a part of me debating if it should have it's own tab .. hmmm

dennisreimann:
@pavlenex 1.) is already included in the Lightning page PR. 

And yes, 2.) can be cleaned up as well, once dstrukt's other PR has been merged.

dennisreimann:
Marking this as ready for review as tests are passing now. We can still consolidate the General and Payment page -- but that can also happen in a separate PR.

pavlenex (approved):
LGTM, perhaps there's no need to consolidate at the moment, but if we go into direction of having a store setup with default currency, etc as suggested [here](https://chat.btcpayserver.org/btcpayserver/pl/6k5y8ckad78yfygqfikzc93bhw), it may, but for now, ready to go imo.

dennisreimann:
Ok, then let's merge this as it is and I'll try to fit it in as another PR for v1.4.0

PR #3331, opened 2022-01-18
Mobile navigation

### LINKED ISSUE #3277: Tabbed Navigation Functionality for Tablet & Mobile
Will update this, but creating a ticket for now for tracking, discussion, and the eventual PR.

Screen Shot 2022-01-06 at 7 23 13 PM

dennisreimann:
In #3277 there are three proposals for the mobile navigation:

As this is something to be tested and experienced, here's as first version the horizontal scrolling approach (first example).

image

Personally I think this one has some downsides:

- Horizontal scrolling is not a widely used UX pattern
- Items might be overlooked, because the user might not know that there are more nav links available (out of visible area)
- After navigation the scroll position resets and the currently active item might not be visible. (This could be doable using JS though)

To prevent the second problem, I've added scroll indicators on the ends (see screenshot).

I'd prefer the second approach, though I see the downside of an added click to open the dropdown before selecting the nav item.

This version was easier to implement though, so I thought let's give it a go and try it. 
Let me know your thoughts, @pavlenex @dstrukt.

Closes #3277.

### DIFF
--- a/BTCPayServer/Views/Shared/_NavLayout.cshtml
+++ b/BTCPayServer/Views/Shared/_NavLayout.cshtml
@@ -1,3 +1,4 @@
+@using NBitcoin
 @{
     Layout = "/Views/Shared/_Layout.cshtml";
     ViewBag.ShowMenu = ViewBag.ShowMenu ?? true;
@@ -18,6 +19,12 @@
 @if (ViewBag.ShowMenu)
 {
     <partial name="@ViewData["NavPartialName"].ToString()" />
+    <script>
+        (function () {
+            const activeEl = document.querySelector('#SectionNav .nav .active')
+            if (activeEl) activeEl.scrollIntoView({ inline: 'center' })
+        })()
+    </script>
 }
 
 <partial name="_StatusMessage" />
--- a/BTCPayServer/Views/Shared/_NavLayout.cshtml
+++ b/BTCPayServer/Views/Shared/_NavLayout.cshtml
@@ -17,9 +17,7 @@
 
 @if (ViewBag.ShowMenu)
 {
-    <nav class="nav">
-        <partial name="@ViewData["NavPartialName"].ToString()" />
-    </nav>
+    <partial name="@ViewData["NavPartialName"].ToString()" />
 }
 
 <partial name="_StatusMessage" />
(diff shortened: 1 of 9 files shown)


dstrukt:
Ooo I was curious myself, even if it's not feasible.

Will take a look later today!

pavlenex:
I mean many apps like Instagram use horizontal scrolling, I don't think it's an uncommon pattern.

dstrukt:
Took a look on iOS Safari. Looks great!

I like the horizontal menu, would definitely like to see a variety of device sizes test this though .. the concern that the gap between tabs could obscure the fact there are more nav items is a legitimate concern.

Another thing to consider is the viewport size, do we lock it to 1, and prevent zooming?

1. Would like the content overlay on nav open to be black, not white, in Light Mode, but that's a subtle note.

2. Some visual display bugs, imagine this is probably consistent across the app:

IMG_7336

IMG_7337

dennisreimann:
Updated the overlay/backdrop color and reduced the space between the nav items on mobile.

light

dark

pavlenex:
LGTM would require some of us playing a bit on mobile, but on the surface I don't see any problems except one:
This should be fixed 😄 
Not sure if this PR introduced it, but noticed it here.
Screenshot 2022-01-19 at 09 42 06

dennisreimann:
Oh yeah, thanks @pavlenex – we seem to be using the modal class wrongly there.

pavlenex:
@dennisreimann Seems like also in modals, delete a store and here 

https://user-images.githubusercontent.com/36959754/150100220-d064b19f-6b61-4209-a651-858b82c2acbb.mp4

dennisreimann:
Yup, sorry it was me rushing it. Will send a fix.

dstrukt:
Was thinking about this last night, I'm also wondering if it might be helpful to leave the bottom border for mobile as well..? It's a little less minimal I suppose, but does make the content quite clear.

dennisreimann:
> I'm also wondering if it might be helpful to leave the bottom border for mobile as well..? It's a little less minimal I suppose, but does make the length, etc.. quite clear.

Yes, tried and applied it, good suggestion!

I'd say this one is good to go. As I said it is already better than what we currently have and we should ship with this one applied.

dstrukt:
Will take a last look, but was happy with where it was, if the only additional is the bottom border .. approval incoming!

dstrukt (approved):

PR #3337, opened 2022-01-20
Store: Combine General and Payment settings

dennisreimann:
Merges the payment settings from the Payment page (removed now) into the General page.

Also introduces the default currency and rate provider fields to the Create Store view:

image

Last but not least, the now obsolete store hints could be removed from the store blob.


dennisreimann:
@NicolasDorier We discussed it in the design call and @pavlenex brought up the concern that it might be confusing for the user having to choose the price source. Is there a special reason to bring the attention to this setting or do you agree that it is an advanced feature we should keep out of the setup view?

dstrukt:
Want to second the concern w.r.t setting "Rates" specifically .. as a new user a while back, I'm not sure I would have understood what I was configuring. 

A potentially prudent move could be adding in the long-term, but removing for 1.4.0 until we can ensure the UX is more clear.

If nothing else, we might want to make it explicitly clear that it's fine to use the default.

NicolasDorier:
@dstrukt the default, is already filed in.
It is default, but it highly recommended to set it up. This is one of the most important setting and people rarely set it up.

IMHO this is not an "advanced feature" every store really need to set that up or they get screwed on exchange rate.

dstrukt:
All fair points, happy to roll with it and adjust based on any feedback we get!

PR #3340, opened 2022-01-21
Add warning about the tradeoff the paybutton

NicolasDorier:
Since it may not be obvious for some users:

Added warning when enabling:
image

Added warning on the page itself.
image


dennisreimann:
Good to communicate this tradeoff clearly, so +1 on that.

However: Why is it possible to create an invoice with the id of an existing one? Why isn't there a unique index?

dstrukt:
Agree to communicate that.

We might want to add the ability to disable it independently of the alert at the very bottom, if that isn't already implemented, just in case you dismiss the alert, but want to disable later.

NicolasDorier:
@dennisreimann not same Invoice Id, but same Order Id. (Which is what a ecommerce backend check)
@dstrukt this warning is always there, even if you dismiss this, next time you go on this page you see it again.

dstrukt:
Ahh, sweet, works for me!

BTCPayServer/Views/UIStores/PayButtonEnable.cshtml
```diff
@@ -1,9 +1,16 @@
...
 
 <h2 class="mt-1 mb-4">@ViewData["Title"]</h2>
 
+<div class="alert alert-warning alert-dismissible mb-5" role="alert">
+        <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close">
+            <vc:icon symbol="close" />
+        </button>
+        <b>Warning: This feature should not be activated on a BTCPay Server store processing commercial transactions.</b><br /><br />
+		By activating this feature, a malicious user can trick your server into thinking an order has been processed by creating a new invoice, reusing the same Order Id of another valid order but different amount or currency.
+</div>
```
dennisreimann, line 13: Can you move it inside of the `col-md-10` beneath? The text being so wide on desktop makes it harder to read.

BTCPayServer/Views/UIStores/PayButtonEnable.cshtml
```diff
@@ -1,9 +1,16 @@
...
 }
 
 <h2 class="mt-1 mb-4">@ViewData["Title"]</h2>
 
+<div class="alert alert-warning alert-dismissible mb-5" role="alert">
+        <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close">
+            <vc:icon symbol="close" />
+        </button>
+        <b>Warning: This feature should not be activated on a BTCPay Server store processing commercial transactions.</b><br /><br />
+		By activating this feature, a malicious user can trick your server into thinking an order has been processed by creating a new invoice, reusing the same Order Id of another valid order but different amount or currency.
```
dennisreimann, line 12:
```suggestion
        <p><strong>Warning:</strong> This feature should not be activated on a BTCPay Server store processing commercial transactions.</p>
        <p>By activating this feature, a malicious user can trick you into thinking an order has been processed by creating a new invoice, reusing the same Order Id of another valid order but different amount or currency.</p>
```

dennisreimann (approved):
Other than these, it's good!

PR #3341, opened 2022-01-21
Add configurable BOLT11Expiration for refunds (Fix #3281)

### LINKED ISSUE #3281: Pull payments expiry date configuration/setting
When attempting a pull payment, as the use case of a refund, using the [ACINQ Phoenix wallet ](https://github.com/ACINQ/phoenix) it will warn/error with a message "The BOLT 11 invoice must have an expiry date of at least 30 days from submission (Provided was only 6)."

Is there a way to configure the expiry time for pull payments? If not I think this would make a good feature.

NicolasDorier:
Fix #3281

When a customer asks a refund via lightning network, he can provide a BOLT11.
The problem is that we are requiring a BOLT11 that won't expire in the next 30 days. It is done to ensure that when the merchant actually pay, the BOLT11 will be valid. (assuming merchants pay after 30 days)

This settings of 30 days may be overly conservative, so I added a way to configure it via the UI/API.

* Add a new `BOLT11Expiration` field to Pull Payments, in the API and UI
* Add a new option in Store's payment settings to configure the default BOLT11Expiration to use when creating a refund.

I am not satisfied to put it in Payment settings. But I think we will refactor this in next release by adding a new `Refund` category, since we need as well to streamline the UX there.

This is how it looks like in the payment settings:
image

This is how it looks like in the create pull payment UI:
image

### DIFF
--- a/BTCPayServer.Client/JsonConverters/TimeSpanJsonConverter.cs
+++ b/BTCPayServer.Client/JsonConverters/TimeSpanJsonConverter.cs
@@ -29,6 +29,17 @@ namespace BTCPayServer.Client.JsonConverters
                 return TimeSpan.FromMinutes(value);
             }
         }
+        public class Days : TimeSpanJsonConverter
+        {
+            protected override long ToLong(TimeSpan value)
+            {
+                return (long)value.TotalDays;
+            }
+            protected override TimeSpan ToTimespan(long value)
+            {
+                return TimeSpan.FromDays(value);
+            }
+        }
         public override bool CanConvert(Type objectType)
         {
             return objectType == typeof(TimeSpan) || objectType == typeof(TimeSpan?);
--- a/BTCPayServer.Client/Models/CreatePullPaymentRequest.cs
+++ b/BTCPayServer.Client/Models/CreatePullPaymentRequest.cs
@@ -13,6 +13,9 @@ namespace BTCPayServer.Client.Models
         public string Currency { get; set; }
         [JsonConverter(typeof(TimeSpanJsonConverter.Seconds))]
         public TimeSpan? Period { get; set; }
+        [JsonConverter(typeof(TimeSpanJsonConverter.Days))]
+        [JsonProperty("BOLT11Expiration")]
+        public TimeSpan? BOLT11Expiration { get; set; }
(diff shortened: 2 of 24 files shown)


pavlenex:
Lightning setting should go into lightning page?
Screenshot 2022-01-21 at 09 56 41

I think just say `Refund expiration time for Bolt 11 invoices`

I agree this has to be  configurable option. I see a tendency for wallets to have 24hours default for expiration lately, but for refunding use-case it's not ideal time-frame.

dstrukt:
Few notes:

+1 to simplified wording:
`Refund expiration time for Bolt 11 invoices`

Also agree it should be configurable, but would 1 or 2 weeks be a suitable compromise of the two (24 hour does seem a bit short, and 1 month seems a bit long)?

On the pull payments page, the title should be `Lightning Network Settings` the same styling (h4 I believe) as `Additional Options`, and it should ideally be above as well.

W.R.T Payment/Lightning>Settings, I could see a case made for both, could also see adding a title `Refund` in the `payment` tab if we didn't want to break it out exclusively, or in the "Payment" section in the Lightning Settings view, depending which route we decide to go. Just my 2sats, but regardless, we will definitely continue to improve the consolidation of these settings.

NicolasDorier:
@pavlenex @dstrukt I will just keep it in payment for now, because this is a temporary thing, I think next release we will change this.

The reason is that refunds right now are cumbersome: For each refunds, you need to specify what to refund exactly, at which rate. This should go away soon, with it configurable directly in a new tab, so issuing a refund is just a matter of clicking the refund button.

NicolasDorier:
@dstrukt  I don't understand your point on additional options: this is an additioal options so it should be under it

NicolasDorier:
About the wording `Refund expiration time for Bolt 11 invoices` I am not sure it explain what it does properly.

The point is that if somebody for a refund give a BOLT 11 less than this expiration time, it will be rejected.

dstrukt:
> @dstrukt I don't understand your point on additional options: this is an additioal options so it should be under it

Ahhh this was my mistake, I missed the carrot, I thought it was just dropped under the accordion section.

pavlenex:
> What about Minimum acceptable expiration time for BOLT11 for refunds

ACK.

PR #3342, opened 2022-01-21
Add loading indicator for "Pay" button in POS terminal app

### LINKED ISSUE #3336: No loading indicator in POS app.
**Describe the bug**
If BTCPay instance is slow then POS user could be confused by "doing nothing", because there is no visual feedback after Pay click and waiting for invoice load. Only browser load could indicate it is loading, but when POS is installed in PWA mode no loading indicator is there. This could result in clicking Pay button multiple times and losing trust in "Bitcoin network" by staff.

**To Reproduce the bug**
1. Go to POS app frontend in keyboard mode on slow server or connectivity
2. Add it to home screen (install PWA)
3. Open app from home screen
4. Set amount and click Pay
5. Get confused by nothing happening
4. After a while see invoice

**Expected behavior**

Every action waiting for a response should show some loader. Here most important is to indicate loading after Pay button click.

bolatovumar:
Show loading indicator and disable the "Pay" button in POS terminal app when "Pay" button is clicked.

close #3336

### DIFF
--- a/BTCPayServer/Views/UIAppsPublic/PointOfSale/VueLight.cshtml
+++ b/BTCPayServer/Views/UIAppsPublic/PointOfSale/VueLight.cshtml
@@ -1,6 +1,6 @@
 @model BTCPayServer.Models.AppViewModels.ViewPointOfSaleViewModel
 <div id="app" class="l-pos-wrapper" v-cloak>
-    <form method="post" asp-controller="UIAppsPublic" asp-action="ViewPointOfSale" asp-route-appId="@Model.AppId" asp-antiforgery="false" data-buy>
+    <form method="post" asp-controller="UIAppsPublic" asp-action="ViewPointOfSale" asp-route-appId="@Model.AppId" asp-antiforgery="false" data-buy ref="form">
         <div class="l-pos-header bg-primary py-3 px-3">
             @if (!string.IsNullOrEmpty(Model.CustomLogoLink))
             {
@@ -26,8 +26,8 @@
 
         <div class="l-pos-controls mt-2">
             <div class="btn btn-outline-secondary btn-lg mb-0" v-on:click="clearTotal">Clear</div>
-            <button class="btn btn-primary btn-lg mb-0" id="pay-button" type="submit">
-                <div class="spinner-border spinner-border-sm" id="pay-button-spinner" role="status" hidden>
+            <button class="btn btn-primary btn-lg mb-0" id="pay-button" type="submit" v-on:click="payButtonClicked" v-bind:disabled="payButtonLoading">
+                <div v-if="payButtonLoading" class="spinner-border spinner-border-sm" id="pay-button-spinner" role="status">
                     <span class="sr-only">Loading...</span>
                 </div>
                 <b>Pay</b>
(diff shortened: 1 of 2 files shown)


dennisreimann (changes requested):
As this is a [Vue app](/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/light-pos/app.js), the handling of the loading state should also be done inside the Vue app. Otherwise rerendering could interfere with state that has been changed natively in the DOM, as Vue wouldn't know about these changes.

Kukks (changes requested):
the feature needs to be handled in Vue

bolatovumar:
@dennisreimann @Kukks moved over to Vue here: 833e4d1a8fab96ce3576216132e17c77cb31570e

BTCPayServer/wwwroot/light-pos/app.js
```diff
@@ -57,6 +65,13 @@ document.addEventListener("DOMContentLoaded",function (ev) {
                 this.tipTotal = null;
                 this.tipTotalNumeric = 0;
             },
+            payButtonClicked: function() {
+                this.payButtonLoading = true;
+                this.$refs.form.submit();
```
bolatovumar, line 70: Binding an event to the pay button in Vue prevents the default form submit for whatever reason so I have to trigger it manually.

BTCPayServer/wwwroot/light-pos/app.js
```diff
@@ -15,9 +15,17 @@ document.addEventListener("DOMContentLoaded",function (ev) {
...
+                payButtonLoading: false,
             }
         },
+        created: function() {
+            /** We need to unset state in case user clicks the browser back button */
+            window.addEventListener('pagehide', this.unsetPayButtonLoading);
+        },
+        destroyed: function() {
+            window.removeEventListener('pagehide', this.unsetPayButtonLoading);
+        },
```
dennisreimann, line 28: I don't think that these are necessary, because the whole app resets when navigating back.

BTCPayServer/wwwroot/light-pos/app.js
dennisreimann: For me it works fine without the ref and the programmatic form submit.

BTCPayServer/Views/UIAppsPublic/PointOfSale/VueLight.cshtml
```diff
@@ -25,7 +25,13 @@
         </div>
 
         <div class="l-pos-controls mt-2">
-            <div class="btn btn-outline-secondary btn-lg mb-0" v-on:click="clearTotal">Clear</div><button class="btn btn-primary btn-lg mb-0" type="submit"><b>Pay</b></button>
+            <div class="btn btn-outline-secondary btn-lg mb-0" v-on:click="clearTotal">Clear</div>
+            <button class="btn btn-primary btn-lg mb-0" id="pay-button" type="submit" v-on:click="payButtonClicked" v-bind:disabled="payButtonLoading">
```
dennisreimann, line 29: The `v-on:click="payButtonClicked"` here should rather be a `v-on:click="handleFormSubmit"` on the form.

BTCPayServer/wwwroot/light-pos/app.js
bolatovumar: Yeah, it works without it if I attach a submit event listener on the form instead of the click listener on the button.

BTCPayServer/wwwroot/light-pos/app.js
bolatovumar: @dennisreimann They are necessary if your browser cached the page completely in memory (which seems to happen for me pretty much always when I click the "Pay" button and then click the browser back button).

https://web.dev/bfcache/

BTCPayServer/Views/UIAppsPublic/PointOfSale/VueLight.cshtml
bolatovumar: @dennisreimann I assume you meant `v-on:submit`. Updated in this commit: https://github.com/btcpayserver/btcpayserver/pull/3342/commits/813138842f8083a996aa86f9c7a9746d3b671970

BTCPayServer/wwwroot/light-pos/app.js
dennisreimann: You are correct, I can reproduce it. Well, that's unfortunate, but let's handle it like this then.

dennisreimann (approved):

PR #3347, opened 2022-01-24
Add suggestion list for currency inputs

NicolasDorier:
This is how it looks, list is sorted alphabetically, except for the top currencies, which cover the majority of our users.

image

This is a suggestion list, which mean it auto complete, so if you already filled it, the suggestion doesn't appear. (I would prefer it appear, but I can't change this behavior easily)

image

Another upside: It shows SATS as a real currency. Not only it makes it easy to discover, but it solves an UX issue reported by @britttttk where people would use SAT rather than SATS.

image

~Note: Safari users won't see this feature. It seems it isn't supported by Safari.~ (edit: it works, https://github.com/btcpayserver/btcpayserver/pull/3347#pullrequestreview-860490961)

I also used this occasion to make a dedicated TagHelpers directory rather than stuffing everything in a single file.


NicolasDorier:
Ping @pavlenex @dstrukt

NicolasDorier:
Ping @pavlenex @dstrukt

bolatovumar (approved):
Tested on Safari 14 (as well as Chrome 97) and works fine on both.

woutersamaey:
Some of us have been waiting a long time for this 🙌

dstrukt:
FUCK YEAH! 

Looking good to me based on those screenshots - will do some testing when I get a moment today or tomorrow.. we chatted about a fuzzy search dropdown last time this issue was raised in the design call, unsure how feasible it would be, so this is pretty incredible to see implemented!

woutersamaey:
Yep. This could be more refined, but I'm also very happy about this. 

One thing I am missing is a label, so it could say "Bitcoin" instead of "BTC", but that'll be for a future improvement. This is not needed here I am thinking ahead for an API call where it would be more important as a key-value list.

NicolasDorier:
@bolatovumar good to know, many site say it was not supported, so I guess it is recent (I think Safari 13 or above)

@dstrukt well, it's kind of fuzzy, if you write `B`, then it will filter currencies by those starting by `B`.

@woutersamaey yeah, right now I can't put a label in this list. I use a browser feature, so what you see is what get entered in the input text. Doing smarter things means we need some JS magic, I wanted to keep that simple for now.

woutersamaey:
If we don't have any JavaScript on hand, I can greatly recommend the `select2` plugin. What we need is shown here: https://select2.org/appearance#labels

Screenshot of the part I'm talking about:
image

- It has an search field at the top
- Supports option groups
- Can be styled however we need using CSS.
- The label and value can be different, but obviously don't need to be.

If we have a JS plugin for this, forget I said anything.

pavlenex:
Let's get this one in for 1.4.0 and @woutersamaey you can open a new issue so we don't forget to tackle it in the 1.5.0?

woutersamaey:
> Let's get this one in for 1.4.0 and @woutersamaey you can open a new issue so we don't forget to tackle it in the 1.5.0?

Issue created as #3349 for whoever wants to follow.

PR #3348, opened 2022-01-24
Make users table responsive

bolatovumar:
Made the users list table responsive just like our other tables so that it doesn't overspill out of the container on mobile.

|Before (table overspills the container)|After (can swipe table horizontally)|
|---|---|
|users-mobile-before|users-mobile|

### DIFF
--- a/BTCPayServer/Views/UIServer/ListUsers.cshtml
+++ b/BTCPayServer/Views/UIServer/ListUsers.cshtml
@@ -40,54 +40,55 @@
         Add User
     </a>
 </div>
-
-<table class="table table-hover">
-    <thead>
-    <tr>
-        <th>
-            <a
-                asp-action="ListUsers"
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/UIServer/ListUsers.cshtml
```diff
@@ -40,54 +40,55 @@
...
-        </th>
-        <th>Created</th>
-        <th>Verified</th>
-        <th class="text-end">Actions</th>
-    </tr>
-    </thead>
-    <tbody>
-    @foreach (var user in Model.Users)
-    {
+<div class="table-responsive">
```
bolatovumar, line 43: The only change is this wrapping div as outlined here: https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables.

Everything else is the same.

PR #3352, opened 2022-01-24
Rename LNUrlAuthController -> UILNAuthController

NicolasDorier:


BTCPayServer/Controllers/UILNURLAuthController.cs
```diff
@@ -17,13 +17,13 @@ namespace BTCPayServer
 {
     [Route("lnurlauth")]
     [Authorize]
```
dennisreimann, line 19: While we are at it: Let's also add the correct policy here. I just came across this, because without policy it lacks the store context.

NicolasDorier: it is impossible to have store context here, as those actions are independant of stores.

NicolasDorier: ok I get it, we lose the context nevermind adding.

PR #3353, opened 2022-01-24
Make Lazy Payment on by default

Kukks:
This PR makes the lazy payment method on by default for new stores.

This PR also introduces a change where any invoice created via the Bitpay API will have lazy payment methods off no matter what the store setting is set to. This is done because in the Bitpay api, there is no way to activate it and in scenarios where an integration uses a custom checkout UI, there is no way to show payment details.
Additionally, it makes all internal invoice creation use the greenfield invoice creator instead( we previously used Bitpay style). This has one main drawback: the ipn settings in the apps + invoicecreate + pay button options will no longer be available.


pavlenex:
Concept ACK. All the bugs with the lazy payment we earlier had were resolved and I think it's an important step ahead.

NicolasDorier:
Let's discuss this after the 1.4

BTCPayServer/Components/StoreSelector/Default.cshtml
```diff
@@ -4,8 +4,7 @@
 @addTagHelper *, BundlerMinifier.TagHelpers
 
 @functions {
-    @* ReSharper disable once CSharpWarnings::CS1998 *@
-    private async Task LogoContent()
+    private void  LogoContent()
```
Kukks, line 7: fixes compiler warnings

pavlenex:
@Kukks you said this may not work with unified QR code well?

Kukks:
> @Kukks you said this may not work with unified QR code well?

Correct. BIP21 w/ BOLT11 enabled + Lazy payments are not compatible because the BOLT11 invoice will not be generated unless you go on the lightning payment method.

pavlenex:
Do we handle this somehow though, should we provide a warning if people try to enable both of them, especially since the lazy payment will be on by default?

Kukks:
> Do we handle this somehow though, should we provide a warning if people try to enable both of them, especially since the lazy payment will be on by default?

That's something to discuss and decide on. We would either hide or disable options and show warning message under options.

pavlenex:
I think that is a good plan for sure.

pavlenex:
Good question, it's been a bit controversial, I'll add it to dev meeting agenda today.

PR #3354, opened 2022-01-25
Adjust Pay Button

dstrukt:
Just adjusted a few minor paddings after poking around.

However in testing it, I realize when you "Disable", it redirects to the "Payment" tab in Store Settings .. shouldn't it redirect back to the default "Enable" Pay Button screen instead?

@dennisreimann Might need some help with that change though.


NicolasDorier:
Good point, doing that

NicolasDorier:
done, let me know

dennisreimann (approved):

dstrukt:
LGTM, thanks, send it!

PR #3358, opened 2022-01-26
Adapt desktop breakpoints in views

dennisreimann:
Desktop tweaks to account for the sidebar beneath the content on LG and too wide forms on XXL.


dstrukt:
Think `max-width: 640px;` or `max-width: 800px` would work for the `col-xxl-constrain`.

Posting for documentation.

dstrukt (approved):

PR #3390, opened 2022-01-30
Ensure Swagger docs are semantically and structurally correct

bolatovumar:
See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/3384


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -193,7 +193,7 @@ public async Task EnsureSwaggerPermissionsDocumented()
 
             JObject json = JObject.Parse(sresp);
 
-            Assert.Equal(description, json["components"]["securitySchemes"]["API Key"]["description"].Value<string>());
```
bolatovumar, line 196: Renamed `"API Key"` security scheme to `"API_Key"` as security scheme names are not allowed to have spaces in the name.

Capture

BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
```diff
@@ -415,69 +415,12 @@
...
                     {
-                        "API Key": [
+                        "API_Key": [
                             "btcpay.store.canviewinvoices"
                         ],
                         "Basic": []
                     }
                 ]
-            },
-            "delete": {
```
bolatovumar, line 425: `"/api/v1/stores/{storeId}/invoices/{invoiceId}/payment-methods"` doesn't have a `DELETE` operation associated with it so this is completely incorrect, it's just a copy of the `DELETE` for archiving of invoics.

Capture

BTCPayServer/wwwroot/swagger/v1/swagger.template.json
```diff
@@ -74,25 +74,22 @@
...
                 "name": "Authorization",
-                "in": "header",
-                "scheme": "token"
```
bolatovumar, line 82: `scheme` doesn't apply to `apiKey` type so needs to be removed.

https://swagger.io/specification/#security-scheme-object

Capture

BTCPayServer/wwwroot/swagger/v1/swagger.template.json
```diff
@@ -74,25 +74,22 @@
...
+                "in": "header"
             },
             "Basic": {
                 "type": "http",
                 "description": "BTCPay Server supports authenticating and authorizing users through the Basic HTTP authentication scheme. Send the user and password encoded in base64 with the format `Basic {base64(username:password)}`. Using this authentication method implicitly provides you with the `unrestricted` permission",
-                "name": "Authorization",
```
bolatovumar, line 87: `name` and `in` don't apply to `http` type.

https://swagger.io/specification/#security-scheme-object

BTCPayServer/wwwroot/swagger/v1/swagger.template.webhooks.json
```diff
@@ -196,17 +196,6 @@
                 ],
                 "summary": "Delete a webhook",
                 "description": "Delete a webhook",
-                "requestBody": {
```
bolatovumar, line 199: `delete` can't have `requestBody`

Capture

NicolasDorier:
Nice, do you know if we can write a test to detect those semantic issue?

NicolasDorier:
Actually we could just use the selenium tester to use https://editor.swagger.io/ to validate it. Hackish, but it works :D

bolatovumar:
> Nice, do you know if we can write a test to detect those semantic issue? I did a test that validate the schema, but there might be some library in CSharp to validate the swagger itself.

@NicolasDorier yeah, I was thinking of this as well. Will have to take a look into it.



> Actually we could just use the selenium tester to use https://editor.swagger.io/ to validate it. Hackish, but it works :D

lol, if all fails I guess we can do this

NicolasDorier:
@bolatovumar I think the solution of just using selenium and editor.swagger is actually the best. No more dependency, and no need of researching.

NicolasDorier:
merging that for now

PR #3393, opened 2022-01-31
Bump z-index on header

### LINKED ISSUE #3377: [Bug] Adjust z-index for Notification Modal
As originally reported by @Kukks on Mattermost.

image

bolatovumar:
Any element that's a child of the sibling of the BTCPay header which has a z-index defined will take precedence of the notifications dropdown element. To fix this issue we either need to bump the z-index on the header element itself OR move the notifications dropdown into its own element that is a sibling of the header rather than its child (more work).

Please let me know if there might be a reason why we would not want to bump the z-index on the header.

close #3377

### DIFF
--- a/BTCPayServer/wwwroot/main/layout.css
+++ b/BTCPayServer/wwwroot/main/layout.css
@@ -19,7 +19,7 @@
     --icon-size: 1.5rem;
     
     height: var(--header-height);
-    z-index: 1;
+    z-index: 9999999;
 }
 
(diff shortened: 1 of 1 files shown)


NicolasDorier:
merging, pinging @dennisreimann in case he has a good reason to fix it in other way :p

dennisreimann:
For now this is good, `MAX_INT` vibes are strong here. 😂

PR #3394, opened 2022-01-31
Remove payment methods not currently configured when creating invoice

### LINKED ISSUE #3380: Remove payment methods not currently configured when creating invoice, etc..
"Since the create invoice page is now linked to a store, we can remove payment method options which have not yet been configured." - Kukks

bolatovumar:
This PR updates the "Create invoice" page to only include the payment methods available which are currently configured for the store.

Also added an error message with a link to create a wallet which shows up at the top of the page if there are no payment methods configured at all.

close #3380

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -883,9 +883,12 @@ namespace BTCPayServer.Controllers
 
         private SelectList GetPaymentMethodsSelectList()
         {
-            return new SelectList(_paymentMethodHandlerDictionary.Distinct().SelectMany(handler =>
-                    handler.GetSupportedPaymentMethods()
-                        .Select(id => new SelectListItem(id.ToPrettyString(), id.ToString()))),
+            var store = GetCurrentStore();
+            var excludeFilter = store.GetStoreBlob().GetExcludedPaymentMethods();
+
+            return new SelectList(store.GetSupportedPaymentMethods(_NetworkProvider)
+                        .Where(s => !excludeFilter.Match(s.PaymentId))
+                        .Select(method => new SelectListItem(method.PaymentId.ToPrettyString(), method.PaymentId.ToString())),
                 nameof(SelectListItem.Value),
                 nameof(SelectListItem.Text));
         }
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -930,7 +930,7 @@ namespace BTCPayServer.Controllers
                     TempData.SetStatusMessageModel(new StatusMessageModel
                     {
                         Severity = StatusMessageModel.StatusSeverity.Error,
-                        Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.SetupWallet), "UIStores", new { cryptoCode = "BTC", storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
+                        Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.SetupWallet), "UIStores", new { cryptoCode = _NetworkProvider.DefaultNetwork.CryptoCode, storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
                         AllowDismiss = false
                     });
                 }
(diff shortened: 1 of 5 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -866,6 +869,17 @@ public async Task<IActionResult> CreateInvoice(InvoicesModel? model = null)
                 var store = await _StoreRepository.FindStore(model.StoreId, GetUserId());
                 if (store == null)
                     return NotFound();
+
+                if (!store.GetSupportedPaymentMethods(_NetworkProvider).Any())
```
bolatovumar, line 873: Added an error message at the top of the page asking user to create a wallet first before trying to create an invoice.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -866,6 +869,17 @@ public async Task<IActionResult> CreateInvoice(InvoicesModel? model = null)
...
+                {
+                    TempData.SetStatusMessageModel(new StatusMessageModel
+                    {
+                        Severity = StatusMessageModel.StatusSeverity.Error,
+                        Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.SetupWallet), "UIStores", new { cryptoCode = "BTC", storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
```
bolatovumar, line 878: Link to create a wallet has `BTC` hardcoded. Not sure if this is OK but I figured you would probably want to have a BTC wallet first. We also do this for LN wallets:

Capture


Alternatively we would have to iterate over all derivation schemes and generate a link for each one which is kind of weird to potentially have multiple links in the error message.

Or alternative we could have a whole new wallet page (which we had previously) that lists all wallets which you can create.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -899,7 +913,7 @@ public async Task<IActionResult> CreateInvoice(CreateInvoiceModel model, Cancell
                 TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.GeneralSettings), "UIStores", new { storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
+                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.SetupWallet), "UIStores", new { cryptoCode = "BTC", storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
```
bolatovumar, line 916: Updated this link to be the same as the above as it's the same kind of error.

BTCPayServer/Views/UIInvoice/CreateInvoice.cshtml
```diff
@@ -27,6 +27,8 @@
 
 <partial name="_StatusMessage" />
 
+@if (!TempData.HasErrorMessage())
```
bolatovumar, line 30: I figured we probably don't want to show the form to the user if they don't have a wallet created first. Otherwise they could end up trying to create an invoice without any payment methods available.

Instead they will see an error message at the top with a link to the page which allows them to create a new wallet.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
Kukks, line 873: You might also need to check if the available payment methods have been disabled (there are helper methods in the store blob)

Kukks (changes requested):

BTCPayServer/Extensions.cs
```diff
@@ -115,6 +115,15 @@ public static bool HasStatusMessage(this ITempDataDictionary tempData)
                    tempData.Peek(WellKnownTempData.ErrorMessage) ??
                    tempData.Peek("StatusMessageModel")) != null;
         }
+
+        public static bool HasErrorMessage(this ITempDataDictionary tempData)
+        {
+            return (
+                tempData.Peek("StatusMessageModel") != null &&
```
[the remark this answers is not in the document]
franciscode12: Mándame monedas

BTCPayServer/Controllers/UIInvoiceController.UI.cs
franciscode12: Envíame monedas

BTCPayServer/Extensions.cs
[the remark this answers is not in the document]
bolatovumar: Done here: 39309d02cf2c2d2bce80c7a385b40a9bf381d2ed

BTCPayServer/Controllers/UIInvoiceController.UI.cs
bolatovumar: @Kukks like this? 85574797d277c2cb4f15301d0ef5aa0007f7be1a

BTCPayServer/Views/UIInvoice/CreateInvoice.cshtml
Kukks: This is too generic. What if we were displaying an unrelated error to your description? I would remove it

Kukks (changes requested):

BTCPayServer/Controllers/UIInvoiceController.UI.cs
NicolasDorier: Can you add the DefaultNetwork I added on https://github.com/btcpayserver/btcpayserver/commit/0afc2cd2cb42d13bd6c00ab0b4d5790a244dc6de ? rather than hardcode BTC?

bolatovumar: @NicolasDorier looks like @Kukks did this already,

PR #3397, opened 2022-01-31
Pay Button Alert: Add missing alert-link classes

dennisreimann:
## Before

before

## After

after


dstrukt (approved):

NicolasDorier:
doesn't seems like this is clickable though

dennisreimann:
Agreed. We should adapt the styles of the `alert-link` class, because this is used in multiple places. @dstrukt

dstrukt:
Agreed, I've wanted to address the stylings for the alerts .. will add it to the design system list.

dennisreimann:
Update the alert link styles:

image

PR #3411, opened 2022-02-04
Prevent initial scroll to section nav

dennisreimann:
On pages taht contain a section nav (sub navigation) the previous `scrollIntoView` led to an initial scroll position start at the section nav. See the store or server settings for examples.

This fixes it so that the scroll vertical position always starts at 0.

## Before

image

## After

image

### DIFF
--- a/BTCPayServer/Views/Shared/_NavLayout.cshtml
+++ b/BTCPayServer/Views/Shared/_NavLayout.cshtml
@@ -22,7 +22,7 @@
     <script>
         (function () {
             const activeEl = document.querySelector('#SectionNav .nav .active')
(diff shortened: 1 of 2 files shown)


bolatovumar (commented):
I tried testing this but I don't notice any difference (it could be that I'm just looking in the wrong place or testing it wrong).

dennisreimann:
@bolatovumar You can go to the store or server settings on master - it'll scroll the content area right to the start of the subnav. This PR fixes it and the scroll position will start at the top of the page and you'll also see the headline.

dstrukt:
LGTM!

dstrukt (approved):

Issue #3408, opened 2022-02-03
[UI/UX Proposal] Wallet View V2


dstrukt:
Continuing to flatten the UI navigation pattern, the latest proposal to improve the Wallet view, as discussed in the design meeting:

I've got a few things I wanted to polish up, as there are parts of this that are still quite rough, but posting for transparency, feedback, and iterative progress as we start to build.

TODO:
- The only navigation issue at the time of posting is how the user navigates back to the "transaction" view, but this is an easy issue and will be resolved.
- Mobile View mocks
- Fixed header section (balance, send/receive, settings) on scroll .. to match other patterns

GIF Clickthrough (had to compress the hell out of it to post, might record a smoother one on my other machine):
![TW_LOOK_02](https://user-images.githubusercontent.com/6250771/152406342-9e113d2b-8a4e-428b-b4d1-ac5f5793310a.gif)

Static assets, not to be heavily scrutinized yet, but posting for details and progress:
<img width="1440" alt="WalletV2Transactions" src="https://user-images.githubusercontent.com/6250771/152406431-0a775932-fa4c-4c84-bb45-f6f13cb89a67.png">
<img width="1440" alt="WalletV2Settings" src="https://user-images.githubusercontent.com/6250771/152406429-96c87b44-82ab-41d3-8eaf-c9c04dd99967.png">
<img width="1440" alt="WalletV2Receive" src="https://user-images.githubusercontent.com/6250771/152406423-e4146cf2-b2ec-4698-a8a0-2212d5a7bf5e.png">
<img width="1440" alt="WalletV2Send" src="https://user-images.githubusercontent.com/6250771/152407088-c5025b52-6c36-499d-a41c-0aff0d81924b.png">
<img width="1440" alt="WalletV2Rescan" src="https://user-images.githubusercontent.com/6250771/152407084-8672ef6e-93f2-4e17-8644-09c14b08c209.png">

PR #3416, opened 2022-02-04
Sticky headers

### LINKED ISSUE #3344: [UI/UX Proposal] Fixed Header on Apps, Server Settings, Account Settings & Store Settings
Will update as I think of more.

Apps Animated Example (doesn't seem to animate in Safari):
FixedHeaderFeature

Advantages:
- Puts CTAs up top as opposed to the bottom, helps with long content scroll
- Makes CTAs a little more clear
- Keeps Delete button down at the bottom
- Could be a design pattern we carry over to all views similar to this...

Drawbacks:
- Potential inconsistency in some other views w.r.t CTAs
- On mobile, would this be fixed at the top, or would we resort to dropping it to the bottom?

Considerations:
The content starts at a top offset of 80px with the addition of the heading / spacings, but we may not want the element to be that tall, so we could shorten the height of the header element on scroll, after a certain scroll position, while still keeping the `<div>` at `top: 0px;`. I would have illustrated that in the prototype, but it's not an easy (or even possible) thing to do with Figma. Can discuss this more in-depth on a design call.

Server Settings Animated Example (doesn't seem to animate in Safari):
FixedHeaderFeatureSS

Considerations:
- Would we fix this on mobile? I'd imagine so.
- It does potentially remove the need for the header on mobile, but we could simply hide this...

dennisreimann:
Building on #3411 this adds sticky headers for the store, server and account settings pages, as well as for the app config pages.

Closes #3344.

## Mobile

sticky-mobile

## Desktop

sticky-desktop

### DIFF
--- a/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
+++ b/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
@@ -7,7 +7,7 @@
 
 <form method="post">
     <div class="sticky-header-setup"></div>
-    <header class="sticky-header d-sm-flex align-items-center justify-content-between mb-2">
+    <header class="sticky-header d-sm-flex align-items-center justify-content-between">
         <h2 class="mb-0">@ViewData["Title"]</h2>
         <div class="d-flex gap-3 mt-3 mt-sm-0">
             <button type="submit" class="btn btn-primary order-sm-1" id="SaveSettings">Save</button>
--- a/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
+++ b/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
@@ -10,7 +10,7 @@
     <h2 class="mb-0">@ViewData["Title"]</h2>
     <div class="d-flex gap-3 mt-3 mt-sm-0">
         <button type="submit" class="btn btn-primary order-sm-1" id="SaveSettings">Save</button>
-        <a class="btn btn-secondary" asp-action="ViewPointOfSale" asp-controller="UIAppsPublic" asp-route-appId="@Model.Id" id="ViewApp">View</a>
+        <a target="_blank" class="btn btn-secondary" asp-action="ViewPointOfSale" asp-controller="UIAppsPublic" asp-route-appId="@Model.Id" id="ViewApp">View</a>
         <a class="btn btn-secondary" asp-action="ListInvoices" asp-controller="UIInvoice" asp-route-searchterm="@Model.SearchTerm">Invoices</a>
     </div>
 </header>
--- a/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
+++ b/BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
@@ -5,19 +5,19 @@
     ViewData.SetActivePage(AppsNavPages.Update, "Update Point of Sale", Model.Id);
 }
 
-<div class="sticky-header-setup"></div>
-<header class="sticky-header d-sm-flex align-items-center justify-content-between mb-2">
-    <h2 class="mb-0">@ViewData["Title"]</h2>
-    <div class="d-flex gap-3 mt-3 mt-sm-0">
-        <button type="submit" class="btn btn-primary order-sm-1" id="SaveSettings">Save</button>
-        <a target="_blank" class="btn btn-secondary" asp-action="ViewPointOfSale" asp-controller="UIAppsPublic" asp-route-appId="@Model.Id" id="ViewApp">View</a>
-        <a class="btn btn-secondary" asp-action="ListInvoices" asp-controller="UIInvoice" asp-route-searchterm="@Model.SearchTerm">Invoices</a>
-    </div>
-</header>
-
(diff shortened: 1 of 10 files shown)


bolatovumar (approved):
Tested in Chrome and Safari 14. Seems to work fine in both.

Not all pages have a sticky header but I guess that's expected?

dennisreimann:
Yes, we'll apply this pattern where it makes sense. Once this is merged I'll get to #3408 where it's also going to be used in the wallet views.

BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
```diff
@@ -5,9 +5,17 @@
     ViewData.SetActivePage(AppsNavPages.Update, "Update Point of Sale", Model.Id);
 }
 
-<partial name="_StatusMessage" />
+<div class="sticky-header-setup"></div>
+<header class="sticky-header d-sm-flex align-items-center justify-content-between mb-2">
+    <h2 class="mb-0">@ViewData["Title"]</h2>
+    <div class="d-flex gap-2 mt-3 mt-sm-0">
```
dennisreimann, line 11: We also need to increase the gap here then.

dennisreimann, line 11:
```suggestion
    <div class="d-flex gap-3 mt-3 mt-sm-0">
```

dstrukt:
Few notes, probably worth tackling in this PR.

Crowdfund:
- Turn "Sounds" off by default
- Tuck "Callback Notification URL" into the Additional Options

dennisreimann:
Good points, let's do it in another PR after this one.

dstrukt:
I'm thinking "View" and "Invoices" should open another window...thoughts before I make that change?

dstrukt (approved):

dennisreimann:
I'd open another window/tab for View, invoices I'm not sure.

dstrukt:
Hmm, also not sure .. added to "View" for now.

BTCPayServer/wwwroot/main/layout.css
```diff
@@ -354,7 +354,7 @@
 }
 
 .sticky-header #SectionNav {
-    margin-bottom: calc(var(--btcpay-space-m) * -1);
+    margin-bottom: calc(var(--btcpay-space-l) * -1);
```
dennisreimann, line 357: @dstrukt I've adapted this to work with e72f04d4f0770c861a7ef914dbbde5e5bdf10db2. So now the sticky header ends right below the subnav border bottom. Ok?

BTCPayServer/Views/UIApps/UpdatePointOfSale.cshtml
```diff
@@ -7,7 +7,7 @@
 
 <form method="post">
     <div class="sticky-header-setup"></div>
-    <header class="sticky-header d-sm-flex align-items-center justify-content-between mb-2">
```
dennisreimann: @dstrukt Removing those bottom margins, as with the added bottom padding in e72f04d4f0770c861a7ef914dbbde5e5bdf10db2 this seemed to much of a spacing.

dstrukt (approved):

dstrukt:
LGTM!

NicolasDorier:
weird, will check 1 test is timing out and it seems reproducible in the CI. Will take a look later.

dennisreimann:
@NicolasDorier It must have something to do with the sticky positioning of the header, which I removed for the Account section in [ffd628e](https://github.com/btcpayserver/btcpayserver/pull/3416/commits/ffd628ea9a9096abc1ccc71b2ecb52fc4e0cd30f) - for this commit the `CanCreateApiKeys` test passes.

NicolasDorier:
doesn't seems to be the case. Checking

NicolasDorier:
So the problem is really just that the test is slow... increasing timeout

NicolasDorier:
@dennisreimann so the issue was that SetCheckbox was taking way too long (I think 10 or 20 sec each)

It uses `WaitForAndClick`, which is a hack when selenium don't want to click properly. This method wait for JS to be ready, and then click. Somehow, it seems the JS takes long to be ready.

I fixed it by making `WaitForAndClick` attempt to just click without waiting before falling back on the wait.

But maybe there is another problem related as to why JS is taking so long to be ready?

NicolasDorier:
There is something weird related to this PR @dennisreimann 

It seems the WaitForClick needed to retry several time, so I turned off the retry to see what was prevent it from working the first time.

It turns out that by scolling to the checkbox, the sticky header prevent to click on it.

```
 element click intercepted: Element <input id="btcpay.server.canmodifyserversettings" type="checkbox" class="form-check-input ms-n4" data-val="true" data-val-required="The Value field is required." name="PermissionValues[4].Value" value="true"> is not clickable at point (365, 7). Other element would receive the click: <header class="sticky-header mb-l">...</header>
```

dennisreimann:
Applied the scroll padding using JS because we cannot solve it with plain CSS: The sticky header heights vary depending on the content and breakpoint, hence we calculate and set it dynamically.

PR #3425, opened 2022-02-07
Greenfield: Store Users

### LINKED ISSUE #3426: Update Store API Call for adding users
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/3419

<sup>Originally posted by **niko-underwood** February  4, 2022</sup>
I'm looking to update the API call for "Update Store" to have the ability to add an existing BTCPayServer user to an existing store. Currently there is only an API call to create a new user. Here is a link to the update store API call for reference: https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Stores_UpdateStore

Use case:

A 3rd party host has a frontend form asking the user to for their name, email, password, and new store they would like to create. Ideally this form data can all be passed via API to create the user, store, and add the user to that store.

Thanks! 
</div>

Kukks:
closes #3426

Note: this changes the behaviour of deleting a store when removing the last admin and instead returns an error.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.StoreUsers.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.StoreUsers.cs
@@ -12,13 +12,13 @@ namespace BTCPayServer.Client
         public virtual async Task<IEnumerable<StoreUserData>> GetStoreUsers(string storeId,
             CancellationToken token = default)
         {
-            var response = await _httpClient.SendAsync(CreateHttpRequest($"api/v1/stores/{storeId}/users"), token);
+            using var response = await _httpClient.SendAsync(CreateHttpRequest($"api/v1/stores/{storeId}/users"), token);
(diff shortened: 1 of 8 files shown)


BTCPayServer.Client/BTCPayServerClient.StoreUsers.cs
```diff
@@ -0,0 +1,37 @@
...
+                CreateHttpRequest($"api/v1/stores/{storeId}/users/{userId}", method: HttpMethod.Delete), token);
+            await HandleResponse(response);
+        }
+
+        public virtual async Task<StoreData> AddStoreUser(string storeId, StoreUserData request,
+            CancellationToken token = default)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+            var response = await _httpClient.SendAsync(
```
NicolasDorier, line 31: you need to always use `using` on the response, or you leak stuff.

PR #3428, opened 2022-02-07
Make payout labels unique

### LINKED ISSUE #3421: "Multiple" pull payout transaction check will only show 1 completed transaction
**Describe the bug**
When processing multiple pull payout and checking on the transaction using wallet > transaction detail  (beside label) not all transaction included in that pull payout is showing.

**To Reproduce the bug**
1. Create multiple pull payout and fill the pull payment with bitcoin address
2.  Go to payments payouts > select all the pending pull payment > send and approve multiple "awaiting for approval" payouts
Screen Shot 2022-02-06 at 3 00 43 PM

3. Go to wallet check the transaction using the transaction details icon beside the label
Screen Shot 2022-02-06 at 2 56 55 PM

**Expected behavior**
Only 1 pull payout will show in the completed even if you processed multiple payouts
Screen Shot 2022-02-06 at 2 57 51 PM

- BTCPay Server Version:   1.4.3
- Deployment Method:       docker
- Browser:                  chrome
-

Kukks:
fixes #3421

### DIFF
--- a/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
+++ b/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.Events;
 using BTCPayServer.Logging;
@@ -23,7 +25,8 @@ namespace BTCPayServer.HostedServices
         private readonly EventAggregator _eventAggregator;
         private readonly WalletRepository _walletRepository;
 
-        public TransactionLabelMarkerHostedService(EventAggregator eventAggregator, WalletRepository walletRepository, Logs logs) :
+        public TransactionLabelMarkerHostedService(EventAggregator eventAggregator, WalletRepository walletRepository,
+            Logs logs) :
             base(eventAggregator, logs)
         {
             _eventAggregator = eventAggregator;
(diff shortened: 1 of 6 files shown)


NicolasDorier:
I don't think this fix his problem. This PR is just including the ID of the pull payment in the payout label. That will be quite ugly as the label are meant to show short text, not a 70 characters blob.

Kukks:
> I don't think this fix his problem. This PR is just including the ID of the pull payment in the payout label. That will be quite ugly as the label are meant to show short text, not a 70 characters blob.
> 
> If anything we should change the link to the info to show a list of payout rather than only one.

it is a way to fix it as the text needs to be unique to not override each other. But yes agree, it is ugly

NicolasDorier:
My point is that there should be a unique Payout label, but the link to it, when clicked, should send to a page showing the list of payouts.

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -70,37 +72,58 @@ protected override async Task ProcessEvent(object evt, CancellationToken cancell
...
+                        {
+                            await _walletRepository.SetWalletTransactionInfo(updateTransactionLabel.WalletId,
+                                pair.Key.ToString(), walletTransactionInfo);
+                        }
+                    }));
+
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine(e);
```
NicolasDorier, line 124: remove debug code

BTCPayServer/Services/Labels/LabelFactory.cs
```diff
@@ -33,32 +35,37 @@ WalletRepository walletRepository
...
             }
         }
 
         public IEnumerable<ColoredLabel> GetWalletColoredLabels(WalletBlobInfo walletBlobInfo, HttpRequest request)
         {
-            foreach (var kv in walletBlobInfo.LabelColors)
-            {
-                yield return CreateLabel(new RawLabel() { Text = kv.Key }, kv.Value, request);
-            }
+            return walletBlobInfo.LabelColors.SelectMany(kv => CreateLabel(new RawLabel() { Text = kv.Key }, kv.Value, request));
```
NicolasDorier, line 50: please don't do this sort of rewrite, it complicates the review and doesn't improve readability

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
Kukks: done

PR #3433, opened 2022-02-07
Always use compressed pubkey for SIN (fix #3432)

### LINKED ISSUE #3432: When using server-side pairing, the SIN is different than expected when uncompressed public key is used
The [docs](https://docs.btcpayserver.org/CustomIntegration/#server-side-pairing) say:

> The second way, is to generate your private key via some bitcoin library then:
> 
> 1. Go to the store's settings
> 2. Click on Access tokens
> 3. Click on Create new Token
> 4. Select merchant's facade and enter your public key,
> 5. Click request pairing
> 6. Click on Approve

Note that they don't specify whether the provided pubkey should be compressed or uncompressed. If we take the [BitPay setup script](https://github.com/bitpay/nodejs-bitpay-client/blob/master/src/setup/BitPaySetup.ts) for example, it will output the public key with `keyPair.getPublic('hex')` which results in an uncompressed key. Of course the same works with the `btcpay` SDK, so what I did was:

```javascript
const btcpay = require('btcpay')
const keyPair = btcpay.crypto.generate_key_pair()
console.log('Public key', keyPair.getPublic('hex'))
console.log('SIN', btcpay.crypto.get_sin_from_key(keyPair))
```

...which resulted in:

```
0459cb199455a288c6e6b032da21d902330fc8d38eccfab25495905dfb0a61e82e7eddb3897af2ad3ebfa651608b9f264d76fc562c99f85a01c63f1922f01247ca
TfGgz8qVP39nrZLQs7VKJ3AJLq32QKV3bHa
```

However, to my surprise, putting this into the BTCPay "Add token" page, I got a different SIN displayed:

image
image

I wasn't sure where this other SIN `TfGyEvDV848jjiycGgfMy3ipAG7jv1D3vwU` came from.

Turns out the issue is that I put in an **uncompressed** public key. Putting in the **compressed** version of the public key (using `keyPair.getPublic().encodeCompressed('hex')` instead of `keyPair.getPublic('hex')`), `0259cb199455a288c6e6b032da21d902330fc8d38eccfab25495905dfb0a61e82e`, results in the correct SIN to be displayed.

Now, all of the client libraries I looked at will ensure that the compressed private key is used in the SIN algorithm. For example the `btcpay` node.js libary does it [here](https://github.com/btcpayserver/node-btcpay/blob/92cfe4342c018e4329b4653e426a98a6a102697c/src/core/cryptography.ts#L19-L26
) with `encodeCompressed`.

I believe that it shouldn't matter whether the compressed or the uncompressed public key is entered into the UI, the SIN should be the same (based on the compressed key like the client libraries also calculate it) either way.

CherryDT:
This change makes sure that even when somebody inputs an uncompressed public key into the "add token" page, the resulting SIN will be calculated from the _compressed_ key, since client libraries will also always use the compressed key for SIN generation.

(Note: There is no need to check `IsCompressed` first because `.Compress()` is just a no-op if the key is already compressed and doesn't throw an exception.)

This fixes #3432

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.cs
+++ b/BTCPayServer/Controllers/UIStoresController.cs
@@ -777,7 +777,7 @@ namespace BTCPayServer.Controllers
             var tokenRequest = new TokenRequest()
             {
                 Label = model.Label,
-                Id = model.PublicKey == null ? null : NBitpayClient.Extensions.BitIdExtensions.GetBitIDSIN(new PubKey(model.PublicKey))
+                Id = model.PublicKey == null ? null : NBitpayClient.Extensions.BitIdExtensions.GetBitIDSIN(new PubKey(model.PublicKey).Compress())
             };
 
             string pairingCode = null;


Kukks (approved):
Lgtm!

NicolasDorier:
I am not sure if I should revert this... I checked NBitpayClient, it isn't compressing.

CherryDT:
You are right, but I don't understand why and how it behaves with BitPay in that case. Look at JS, PHP and Go libraries (including `btcpay`), they are all compressing. If you are saying that there is a chance someone using _this_ client will want to use an uncompressed key, then I don't know what's the right solution, maybe a warning?

PR #3437, opened 2022-02-08
App updates

### LINKED ISSUE #3422: [UI/UXCrowdfund Updates

dennisreimann:
Closes #3422.

One item left from the original issue is
- [ ] Improve "Reset goal" .. not clear

Maybe @dstrukt has an idea how to improve it.

To make the "Make Crowdfund Public" toggle more prominent I moved it below the description field and introduced an additional section "Goal", to separate out the additional fields below, because the top part felt crowded.

image

### DIFF
--- a/BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
+++ b/BTCPayServer/Models/AppViewModels/UpdateCrowdfundViewModel.cs
@@ -36,7 +36,7 @@ namespace BTCPayServer.Models.AppViewModels
 
         [Required]
         [Display(Name = "Make Crowdfund Public")]
-        public bool Enabled { get; set; } = false;
+        public bool Enabled { get; set; }
 
         [Required]
         [Display(Name = "Enable background animations on new payments")]
@@ -44,11 +44,11 @@ namespace BTCPayServer.Models.AppViewModels
 
         [Required]
         [Display(Name = "Enable sounds on new payments")]
-        public bool SoundsEnabled { get; set; } = true;
+        public bool SoundsEnabled { get; set; }
 
         [Required]
         [Display(Name = "Enable Disqus Comments")]
-        public bool DisqusEnabled { get; set; } = true;
+        public bool DisqusEnabled { get; set; }
 
         [Display(Name = "Disqus Shortname")]
         public string DisqusShortname { get; set; }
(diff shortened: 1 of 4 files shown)


Zaxounette:
Do we know the context of why that wording was used in the past?

If nothing specific, we can change it altogether and take example on blog post wording.
My proposal:
* Have the button change between `Draft` and `Public` on toggle
* Move up the toggle item along side the save button up top. It'll be a prominent placement, can't miss it. I also can't make sense of why we'd leave the toggle in the middle of the page, between two configuration sections.

dstrukt:
1. I'm not aware of any context, and feel we should scrap it, as it could be simplified as per your suggestion I like the idea of simplifying the wording to Draft/Public. 

I also see it either as the top item above all input forms, or in the header, but potentially as:

[WILL PROVIDE MOCK]

2. I've improved the Target/Currency to reduce space and correlate the two better:

Screen Shot 2022-02-09 at 5 54 55 AM

3.
- I've removed the "Other Actions" header on Crowdfund/POS, it's just unnecessary at this stage.
- Updated alert message bottom margin.


Beyond Original Items:

4. Would be nice to set a minimum 3-4 line height for the Animation/Sounds textareas, for the time being, thoughts?

5.. I'm still not 100% sure what to do with the Reset Goal .. I don't know that I really understand the purpose of the feature. 

6. "Count all invoices created on the store as part of the goal" 

Also feels like a strange functional element that I don't really understand the use case for, but should stay off by default.. and possibly could be tucked under additional/advanced settings?

7. Share / Crowdfund URL .. I think this might be out of scope for this PR, but making the sharable link / visibility of the end URL more clear is always a good thing.

[WILL PROVIDE MOCK WITH DRAFT/PUBLIC]

dstrukt:
w.r.t point #2

The old copy for the currency field was 
`Primary currency used for targets and stats. (e.g. BTC, LTC, USD, etc.)`

Which the way it was implemented made me think it was solely correlated to the target amount display -- but according to the label, it has greater implications for the display of currency across the crowdfund app view (perks, etc..). 

This isn't consistent with how we handle this across btcpay ..

e.g. simlarily with `Create Invoice` it's
Amount & Currency, but it singularly affects that input / display, but in this case, the "Currency" seems to affect the entirety of the app display as mentioned above

My question is, should that "currency" input field be given more special treatment (e.g. a new section) beyond what I've done, to highlight the impact it has on the various parts of the end-user view? 

tl;dr This is the app's currency display, not just the target amount.

It's early, so hope this makes sense.

NicolasDorier:
Merging that for now, I want to make a small release.

PR #3451, opened 2022-02-12
Redirect to transactions list after wallet creation

### LINKED ISSUE #3450: [UX] "Create Wallet" Redirection Improvement
Currently after setting up your on-chain wallet, you're directed to the Store Settings > "Payment" tab, but we should be redirecting to the on-chain wallet view.

Currently works this way with setting up your Lightning Node.

Any objections to this?

dennisreimann:
Closes #3450.


dstrukt:
Ahhh, damn .. I've been doing some more .NET learnin' and could have almost done this one!

Getting there.

Thanks for picking it up!

LGTM!

PR #3469, opened 2022-02-17
Adds border to sidebar

dstrukt:
Dead simple, but looks nicer, imo.

Screen Shot 2022-02-16 at 11 02 21 PM

### DIFF
--- a/BTCPayServer/wwwroot/main/layout.css
+++ b/BTCPayServer/wwwroot/main/layout.css
@@ -17,7 +17,6 @@
     --button-height: 40px;
     --button-padding: 7px;
     --icon-size: 1.5rem;
-    
     height: var(--header-height);
     z-index: 1031; /* needs a value between fixed and the offcanvas backdrop, see https://getbootstrap.com/docs/5.1/layout/z-index/ */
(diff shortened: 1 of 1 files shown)


BTCPayServer/wwwroot/main/layout.css
```diff
@@ -17,9 +17,9 @@
     --button-height: 40px;
     --button-padding: 7px;
     --icon-size: 1.5rem;
-    
     height: var(--header-height);
     z-index: 9999999;
+    border-right: 1px solid var(--btcpay-body-border-light);
```
dennisreimann, line 22: If done here this applies to the mobile menu as well. It should be down below in the "L and up" media query.

dennisreimann (changes requested):

BTCPayServer/wwwroot/main/layout.css
dennisreimann: Rebased and updated.

dennisreimann (approved):

dstrukt:
Oo great catch!

LGTM.

PR #3471, opened 2022-02-17
Sticky header updates

dennisreimann:
Makes the sticky header a tag helper and adds it to the remaining list and edit/detail views.

The only thing that needs some cleanup is the button area for editing payment requests: As @dstrukt also noted, this is too many buttons and we should move at least two of them.

image

Leaving it up for discussion in todays design meeting.

### DIFF
--- a/BTCPayServer/TagHelpers/StickyHeader.cs
+++ b/BTCPayServer/TagHelpers/StickyHeader.cs
@@ -1,0 +1,30 @@
+using System.Text;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace BTCPayServer.TagHelpers;
+
+[HtmlTargetElement("sticky-header")]
+public class StickyHeader : TagHelper
+{
+    private readonly HtmlEncoder _htmlEncoder;
+    
+    [HtmlAttributeName("class")]
+    public string CssClass { get; set; }
+    
+    public StickyHeader(HtmlEncoder htmlEncoder)
+    {
+        _htmlEncoder = htmlEncoder;
+    }
+    
+    public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+    {
+        var content = await output.GetChildContentAsync(NullHtmlEncoder.Default);
+        var additionalClasses = string.IsNullOrEmpty(CssClass) ? "" : $" {_htmlEncoder.Encode(CssClass)}";
+        output.TagName = null;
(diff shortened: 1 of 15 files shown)


dstrukt (approved):
LGTM

BTCPayServer/TagHelpers/StickyHeader.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        _htmlEncoder = htmlEncoder;
+    }
+    
+    public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+    {
+        var content = await output.GetChildContentAsync(NullHtmlEncoder.Default);
+        var additionalClasses = string.IsNullOrEmpty(CssClass) ? "" : $" {_htmlEncoder.Encode(CssClass)}";
+        output.TagName = null;
+        output.Content.SetHtmlContent("<div class=\"sticky-header-setup\"></div>" +
```
Kukks, line 26: this is some hardcore inline html/js stuff

Kukks (changes requested):
While I appreciate the cleaner html syntax, that tag helper is very hardcore with executing js dynamically through it. Maybe a razer component is a better fit?

BTCPayServer/TagHelpers/StickyHeader.cs
dennisreimann: It's the sticky-header boilerplate code we had spread across several pages before. Unfortunately view components allow for specifying the inner content inside the hosting view, which is what we need here.

dennisreimann: For the JS: We could move it to the site.js file, but then it'd execute on page load, whereas having it inline like this makes it run immediately after the sticky header is rendered. This would prevent jumping to the wrong scroll position when using links with anchors that target specific IDs.

NicolasDorier: It doesn't work because of CSP

NicolasDorier: Better is just to call a function in the `site.js` in the onload of the header. But even this, you need to register it to the CSP.

dennisreimann: Cleaned it up and we can get around the inline JS and tag helper, so I removed that.

VitaTerry (commented):
Those Moves Are The Stupidest Moves Hacking Is Not Allowed Now I Have Yours Thanx Let The Good Times Roll Friends

PR #3474, opened 2022-02-18
Add ability to clear tx label filter

bolatovumar:
I noticed there is no way to clear tx label filter after you set it without manually removing it from the query string which is not great. This PR adds a button to clear the filter.

Capture
Capture2

### DIFF
--- a/BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
+++ b/BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
@@ -89,6 +89,10 @@
                         <a asp-route-labelFilter="@label.Text" class="badge position-relative text-white flex-grow-0" style="background-color:@label.Color;color:@label.TextColor !important;">@label.Text</a>
                     }
                 </div>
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -89,6 +89,10 @@
                         <a asp-route-labelFilter="@label.Text" class="badge position-relative text-white flex-grow-0" style="background-color:@label.Color;color:@label.TextColor !important;">@label.Text</a>
                     }
                 </div>
+                @if (Context.Request.Query.ContainsKey("labelFilter"))
+                {
+                    <a asp-route-labelFilter="" class="btn btn-primary d-flex align-items-center">Clear filter</a>
```
dennisreimann, line 94:
```suggestion
                    <a asp-route-labelFilter="" class="btn btn-secondary d-flex align-items-center">Clear filter</a>
```

dennisreimann (commented):
Good point, let's make it a secondary button.

Kukks (approved):

PR #3476, opened 2022-02-18
Transfer Processors

Kukks:
This PR introduces a few things:
* Payouts can now be directly nested under a store instead of through a pull payment.
* The Wallet Send screen now has an option to "schedule" instead of simply creating a transaction. When you click on schedule, all transaction destinations are converted into approved payouts. Any options relating to fees or coin selection are discarded.
* There is a new concept introduced, called "Transfer Processors".  Transfer Processors are services for stores that process payouts that are awaiting payment. Each processor specifies which payment methods it can handle.  BTCPay Server will have some forms of transfer processors baked in but it has been designed to allow the Plugin System to provide additional processors.
* The initial transfer processors provided are "automated processors", for on chain and lightning payment methods. They can be configured to process payouts every X amount of minutes. For  on-chain, this means payments are batched into one transaction, resulting in more efficient and cheaper fees for processing.


pavlenex:
Great work so far @Kukks 🚀 

Would you need #1851 to proceed with this one or not? Do we want #1851 in at all?

Kukks:
Ideally both #1841 and #3428 make it in alongside this

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -257,14 +257,16 @@ public async Task<IActionResult> InitiatePayment(PaymentMethodId paymentMethodId
...
             {
                 case UriClaimDestination uriClaimDestination:
                     uriClaimDestination.BitcoinUrl.Amount = new Money(blob.CryptoAmount.Value, MoneyUnit.BTC);
-                    bip21.Add(uriClaimDestination.ToString());
+                    
+                    bip21.Add(uriClaimDestination+suffix);
                     break;
                 case AddressClaimDestination addressClaimDestination:
-                    bip21.Add(network.GenerateBIP21(addressClaimDestination.Address.ToString(), new Money(blob.CryptoAmount.Value, MoneyUnit.BTC)).ToString());
+                    bip21.Add(network.GenerateBIP21(addressClaimDestination.Address.ToString(), new Money(blob.CryptoAmount.Value, MoneyUnit.BTC))+suffix);
```
NicolasDorier, line 269: You are trolling me here! `GenerateBIP21` returns a `PaymentUrlBuilder`, not a string, it was a refactoring I did long ago to prevent bugs we had many time cause by forgetting to happen `?` to the URI before the query params, and with this you reintroduce bugs!

Just add the key value to `PaymentUrlBuilder.QueryParams` rather than adding suffice.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -257,14 +257,16 @@ public async Task<IActionResult> InitiatePayment(PaymentMethodId paymentMethodId
...
                 continue;
             var claim = await ParseClaimDestination(paymentMethodId, blob.Destination);
+            var suffix = $"&payout={payout.Id}";
             switch (claim.destination)
             {
                 case UriClaimDestination uriClaimDestination:
                     uriClaimDestination.BitcoinUrl.Amount = new Money(blob.CryptoAmount.Value, MoneyUnit.BTC);
-                    bip21.Add(uriClaimDestination.ToString());
+                    
+                    bip21.Add(uriClaimDestination+suffix);
```
NicolasDorier, line 266: same comment as below here, you can get the `uriClaimDestination.BitcoinUrl` to get the `PaymentUrlBuilder`.

BTCPayServer/Controllers/GreenField/GreenfieldPullPaymentController.cs
```diff
@@ -275,13 +270,96 @@ public async Task<IActionResult> CreatePayout(string pullPaymentId, CreatePayout
...
+                return this.CreateValidationError(ModelState);
+            }
+            var result = await _pullPaymentService.Claim(new ClaimRequest()
+            {
+                Destination = destination.destination,
+                PullPaymentId = request.PullPaymentId,
+                PreApprove = request.Approved,
+                Value = request.Amount,
+                PaymentMethodId = paymentMethodId,
+                StoreId = storeId
```
NicolasDorier, line 362: why did you duplicate the siwtch below?

BTCPayServer/Controllers/GreenField/GreenfieldStoreLightningAutomatedTransferProcessorsController.cs
```diff
@@ -0,0 +1,93 @@
...
+using BTCPayServer.TransferProcessors.Settings;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using TransferProcessorData = BTCPayServer.TransferProcessors.TransferProcessorData;
+
+namespace BTCPayServer.Controllers.Greenfield
+{
+    [ApiController]
+    [Authorize(AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+    public class GreenfieldStoreAutomatedLightningTransferProcessorsController : ControllerBase
```
NicolasDorier, line 19: What about just `GreenfieldTransferProcessor`, seems like Java name here :p

BTCPayServer/Controllers/GreenField/GreenfieldPullPaymentController.cs
Kukks: fixed it now using extracted method

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: Yes, but `UnknownParameters` is a read-only Dictionary :(

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: fixed

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
Kukks: fixed it using 
```
uriClaimDestination.BitcoinUrl.Amount = new Money(blob.CryptoAmount.Value, MoneyUnit.BTC);
                    var newUri = new UriBuilder(uriClaimDestination.BitcoinUrl.Uri);
                    BTCPayServerClient.AppendPayloadToQuery(newUri, new KeyValuePair<string, object>("payout", payout.Id));
                    bip21.Add(newUri.Uri.ToString());
```

BTCPayServer/Controllers/GreenField/GreenfieldStoreLightningAutomatedTransferProcessorsController.cs
Kukks: Yeah, unfortunately this is an api for a specific processor that does lightning payments in an automated way :P

NicolasDorier:
@Kukks , I scheduled for later, then send the selected payout, then Schedule for later, here is the result:
image

NicolasDorier:
Actually the problem happen when we send two times to the same destination.

Kukks:
> @Kukks , I scheduled for later, then send the selected payout, then Schedule for later, here is the result: image

fixed, was meant to show like this:
image

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -244,6 +249,15 @@ private async Task MigrateLighingAddressSettingRename()
            }
         }
 
+        private async Task MigrateAddStoreToPayout()
+        {
+            await using var ctx = _DBContextFactory.CreateContext();
+
+            await ctx.Payouts.Include(data => data.PullPaymentData)
```
NicolasDorier, line 256: this will explode on big instances, you should do a loop with batch. (.NET6.0 have a new extension method for that ! [Enumerable.Chunk()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.chunk?view=net-6.0))

BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
```diff
@@ -20,6 +20,7 @@ public class FeeRateOption
         }
         public List<TransactionOutput> Outputs { get; set; } = new List<TransactionOutput>();
 
+        public List<string> ScheduledTransfersToInclude = new List<string>();
```
NicolasDorier, line 23: I think isn't used.

Kukks: correct, will remove

BTCPayServer/Controllers/GreenField/GreenfieldStoreLightningAutomatedTransferProcessorsController.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public GreenfieldStoreAutomatedLightningTransferProcessorsController(TransferProcessorService transferProcessorService,
+            EventAggregator eventAggregator)
+        {
+            _transferProcessorService = transferProcessorService;
+            _eventAggregator = eventAggregator;
+        }
+
+        [Authorize(Policy = Policies.CanViewStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/transfer-processors/" + nameof(LightningAutomatedTransferSenderFactory))]
```
NicolasDorier, line 32: why you don't use PaymentMethodId here?

BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
```diff
@@ -49,6 +52,13 @@
 
 <partial name="_StatusMessage" />
 
+    @if (Context.GetStoreData()?.TransferProcessors?.Any(data => data.PaymentMethod.Equals(Model.PaymentMethodId, StringComparison.InvariantCultureIgnoreCase)) is not true && TransferProcessorFactories.Any(factory => factory.GetSupportedPaymentMethods().Contains(paymentMethodId)))
+    {
```
NicolasDorier, line 56: We shouldn't push users to use a transfer processor, this is still experimental stuff now.
On top of it, this warning can't be removed, it should be removable like what we did to announce the new label's status in invoices page.

Kukks: Only if Lord @pavlenex agrees as he really wanted more helper alerts around this feature.

BTCPayServer/Controllers/GreenField/GreenfieldStoreLightningAutomatedTransferProcessorsController.cs
Kukks: It is optional (there are 2 routes)

BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
pavlenex: I'm fine with either removing a warning, in which case we'll probably loose on valuable feedback as the feature matures or by making it dismissible and make it go away once payment method is configured with processor.

Kukks:
Renamed it to Payout processors (if you deployed this pr, you need to nuke db)

BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
NicolasDorier: the thing is that a warning is warranted when this feature will be used by all users, but that's not the case.

What I fear with this warning is that noobs don't know what it does and just blindly follow the warning just to make it go away. Ending up in a situation where money is being sent without them understanding why.

Issue #3458, opened 2022-02-15
Greenfield API: Apps 


Kukks:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/3439

<div type='discussions-op-text'>

<sup>Originally posted by **AndrewIammatteo** February  8, 2022</sup>
Hi,

I have not come across the feature of being able to copy/export/import apps. This would be a very useful features. This has mostly been a headache for testing, but I am also working on a workaround for tips going to a separate Store wallet than the original purchase and this would be very helpful. (I am redirecting the base store app to the employee tipping apps and there needs to be one per employee)

Thanks</div>

bolatovumar:
I can't believe we haven't added support for apps in Greenfield API yet. I created a draft PR which implements basic functionality for adding an app here: https://github.com/btcpayserver/btcpayserver/pull/3511.

Feedback welcome!

bolatovumar:
@dennisreimann basic functionality is done but I will probably be adding more as all you can do at the moment is just create a new app, you can't do anything else.

pavlenex:
@bolatovumar would you like to open a separate issue for that as to-do or update this one so we're aware what's left to implement? Would a separate tracker issue be useful?

bolatovumar:
@pavlenex I can update this one with a list to track progress.

bolatovumar:
List of things to implement below

**Basic stuff**

- [x] Add endpoint for creating a PoS app: https://github.com/btcpayserver/btcpayserver/pull/3511
- [ ] Add endpoint for updating a PoS app
- [ ] Add endpoint for creating a Crowdfund app
- [ ] Add endpoint for updating a Crowdfund app
- [x] Add endpoint for deleting apps: https://github.com/btcpayserver/btcpayserver/pull/3894
- [x] Add endpoint for retrieving an app: https://github.com/btcpayserver/btcpayserver/pull/3894
- [ ] Add endpoint for listing all apps

**Improvements**

- [ ] Return app details for PoS app (not just the basics)
- [ ] Return app details for Crowdfund app (not just the basics)

bolatovumar:
@PatMulligan yes, it will! The PR is ready: https://github.com/btcpayserver/btcpayserver/pull/4102

dennisreimann:
Seems like this is done, @Kukks @bolatovumar ?

pavlenex:
Seems so to me, I think @bolatovumar would know for sure.

[...]

PR #3511, opened 2022-02-27
Add support for creating POS apps through Greenfield API

bolatovumar:
This PR adds support for creating POS apps through Greenfield API. Will be adding support for Crowdfund app in a follow up PR.

Address #3458


BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -0,0 +1,102 @@
...
+            var defaultCurrency = (await _storeRepository.FindStore(request.StoreId)).GetStoreBlob().DefaultCurrency;
+            Enum.TryParse(request.AppType, out AppType appType);
+            switch (appType)
+            {
+                case AppType.Crowdfund:
+                    var emptyCrowdfund = new CrowdfundSettings { TargetCurrency = defaultCurrency };
+                    appData.SetSettings(emptyCrowdfund);
+                    break;
+                case AppType.PointOfSale:
+                    var empty = new PointOfSaleSettings { Currency = defaultCurrency };
```
bolatovumar, line 62: Empty POS app has the template with example products set by default (exactly like the app created through the UI). Do we want to keep doing that for the API or should it be empty by default?

BTCPayServer.Client/Models/CreateAppRequest.cs
```diff
@@ -0,0 +1,9 @@
+namespace BTCPayServer.Client.Models
+{
+    public class CreateAppRequest
+    {
+        public string StoreId { get; set; }
```
bolatovumar, line 5: I only added support for the basic required fields for now. Should I make more fields configurable for the initial app creation request? It probably won't work well to have more fields because the fields for POS and Crowdfund apps differ.

I guess we could also add app-specific endpoints such as `/api/v1/app/pos` and `/api/v1/app/crowdfund` which could accept the full set of parameters for each app type?

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -0,0 +1,102 @@
...
+            StoreRepository storeRepository,
+            UserManager<ApplicationUser> userManager,
+            BTCPayNetworkProvider btcPayNetworkProvider
+        )
+        {
+            _appService = appService;
+            _storeRepository = storeRepository;
+        }
+
+        [HttpPost("~/api/v1/apps")]
```
bolatovumar, line 36: Currently this is the only functionality supported. You can create an app and that's it. Once the implementation for this functionality is ACKed I can add more functionality in follow-up PRs.

bolatovumar (commented):
Once the code is looking good and is ACKed I will add tests and docs.

bolatovumar:
@NicolasDorier @Kukks thoughts on this PR? I fixed the broken build.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -0,0 +1,103 @@
...
+        [Authorize(Policy = Policies.CanModifyStoreSettingsUnscoped, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        public async Task<IActionResult> CreateStore(CreateAppRequest request)
```
NicolasDorier, line 39: Should be `CreateApp`

NicolasDorier:
@bolatovumar I think it should be one route per app type, as parameters are largely different.

BTCPayServer.Client/Models/CreateAppRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public class CreatePointOfSaleAppRequest : CreateAppRequest
+    {
+        public string Currency { get; set; } = null;
+        public string Title { get; set; } = null;
+        public string Description { get; set; } = null;
+        public string Template { get; set; } = null;
+        public string DefaultView { get; set; } = null;
+        public bool ShowCustomAmount { get; set; } = true;
+        public bool ShowDiscount { get; set; } = true;
```
bolatovumar, line 17: I mirrored the defaults that you would get if you were to create a new app from the UI. However, I'm not sure if it's a good idea. For example, `ShowDiscount` is set to `true` by default which means anyone can just input any discount amount for the cart view.

BTCPayServer.Client/Models/PointOfSaleAppData.cs
```diff
@@ -0,0 +1,20 @@
...
+        public string AppType { get; set; }
+        public string Name { get; set; }
+        public string StoreId { get; set; }
+        [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
+        public DateTimeOffset Created { get; set; }
+    }
+    
+    public class PointOfSaleAppData : AppDataBase
+    {
+        // We can add POS specific things here later
```
bolatovumar, line 18: I didn't include any POS specific things in the response. We can add them later if needed. Should be easy to do.

bolatovumar (commented):
@NicolasDorier added support to creating POS apps through the API. If this PR looks good I can add support for the Crowdfund app in a follow up PR. I can then add support for modifying app settings in separate PRs.

BTCPayServer.Client/Models/CreateAppRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+        public string AppType { get; set; }
+    }
+
+    public class CreatePointOfSaleAppRequest : CreateAppRequest
+    {
+        public string Currency { get; set; } = null;
+        public string Title { get; set; } = null;
+        public string Description { get; set; } = null;
+        public string Template { get; set; } = null;
+        public string DefaultView { get; set; } = null;
```
NicolasDorier, line 15: Just accept directly the enum and add `[JsonConverter(typeof(StringEnumConverter))]` on top of the property, so you don't have to do an `Enum.Parse`, and if the user put a bad value, it will autmatically gives him back the reason.

NicolasDorier:
@bolatovumar I think only thing missing is the small comment I made, and a small test.

NicolasDorier:
Meh test is good enough

BTCPayServer.Client/Models/CreateAppRequest.cs
bolatovumar: @NicolasDorier to do this I would have to move the enum somewhere that's accessible from here (I can't access its namespace from here now) and then update the reference to it in all places where it's used which is quite a few places. Is it worth it?

bolatovumar: Done in https://github.com/btcpayserver/btcpayserver/pull/3511/commits/c917aec401bf858b956f97d76b637a717e7b40d9

PR #3519, opened 2022-03-03
Use simple layout for 2FA views

dennisreimann:
Uses the reduced layout for the intermediary login steps.

2fa

### DIFF
--- a/BTCPayServer/Views/Shared/_LayoutSignedOut.cshtml
+++ b/BTCPayServer/Views/Shared/_LayoutSignedOut.cshtml
@@ -1,5 +1,6 @@
 ﻿@{
     Layout = "_LayoutSimple";
+    ViewBag.ShowTitle ??= true;
     ViewBag.ShowLeadText ??= false;
 }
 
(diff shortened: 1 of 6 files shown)


pavlenex:
Thanks for tackling this, LGTM.

Are there any best practices when it comes to authentication method presentation? For example, here we showcase two methods right away, which for some may be confusing? Wouldn't it be better to prompt user how they want to authenticate or perhaps provide a default U2F with an option to switch to OTP with a click?

dennisreimann:
@pavlenex I've thoight about this, but found this to be the best solution imho, because it is the fastest to get though:
- The 2FA input is autofocused, meaning you can instantly past the OTP (if you use a password manager that copies it)
- The FIDO2 option is also available, meaning you can tap your device and be done

These allow for quick interaction, whereas otherwise one would need to first choose the method and have an additional click. For sites that let you choose I find this to be a bit cumbersome — and if they have a default it is oftentimes not the one I want, meaning we'd then also have to make this configurable.

pavlenex:
Fair point, but at the same time it may be slightly overwhelming on a single screen.

Another problem with this is that we should communicate it better, with FIDO 2 you don't need to just use a security key, you can use a fingerprint if you have it on laptop or PC. Just some ideas for the future.

Kukks (approved):
ACK from my end.  I use both options and having an extra screen in between would be very annoying for me.

dennisreimann:
One more take: Let's merge and ship as it is already better than the current version. We can then iterate once we get feedback.

PR #3530, opened 2022-03-04
Dashboard

### LINKED ISSUE #1711: Dashboard (replace homepage)

dennisreimann:
Started to mock up the UI, the data is fake right now. For the charts I'm using [charts.js](https://www.chartjs.org/docs/), which is very flexible and well maintained. They support many types of charts and I think it's a good fit.

## Current State

image

image

## What's New Modal

image

Once done, this closes #1711.


dstrukt:
🔥

It's all coming together as planned!

dstrukt:
I think one of the things I'm quickly realizing is building in the ability to show/hide/configure what cards are visible .. additionally how much we make customizable, e.g. are some cards always visible by default.

Secondarily, how this factors into building plug-ins and where we store those settings / configuration. Wordpress, etc.. have always been a bit of a mess, and I'd like for us to think deeply about it as we start to introduce more to improve that experience.

Just noting for the next set of discussions.

dennisreimann:
@dstrukt I agree with the points above, but would note that we should ship Dashboard v1 and then use the feedback to iterate. I think providing the basic functionality in v1.5.0 is good and we can then see how to further improve and customize it.

dennisreimann:
Progressing …

grafik

dstrukt:
> @dstrukt I agree with the points above, but would note that we should ship Dashboard v1 and then use the feedback to iterate. I think providing the basic functionality in v1.5.0 is good and we can then see how to further improve and customize it.

SGTM! 

I'm mostly making notes, sometimes even just for myself, for a later reference .. as well as and throwing out ideas to be discussed, but always agree, ship and iterate, especially for such a non-critical feature.

And as mentioned on MM .. looking fantastic!

-- 

One nit-pick .. the :last-child for the Recent TXs & Invoices shouldn't have a bottom border, it's cleaner :)

dstrukt (approved):

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -0,0 +1,53 @@
...
+            <input type="radio" class="btn-check" name="filter" id="filter-1m">
+            <label class="btn btn-link" for="filter-1m">1M</label>
+            <input type="radio" class="btn-check" name="filter" id="filter-1y">
+            <label class="btn btn-link" for="filter-1y">1Y</label>
+            <input type="radio" class="btn-check" name="filter" id="filter-all">
+            <label class="btn btn-link" for="filter-all">ALL</label>
+        </div>
+    </header>
+    <div class="chart-container">
+        <canvas id="StoreWalletBalance-@Model.Store.Id"></canvas>
```
NicolasDorier, line 24: I think canvas aren't supported by Tor browser

dennisreimann: Yes, there are alternatives like [Chartist](https://gionkunz.github.io/chartist-js/examples.html), which render the charts as SVG.

dennisreimann: Done in [ee6e34c](https://github.com/btcpayserver/btcpayserver/pull/3530/commits/ee6e34cbe6d0b383c8d4d00756fe25fb287bb508).

pavlenex:
This button gives 404.
Screenshot 2022-04-06 at 12 35 42

dstrukt:
Few Notes & Nits:

Wallet Balance:
- Wondering if we need 9 Y-axis increment values on the Wallet Balance, might look nicer with say 5? 
- Circle point radius down to 7px
- Line width to 2px

Top Right Section (needs a conceptual name lol):
- Manage button should align right (like View All, etc.. in Recent Invoices)

Recent Transactions:
- If empty, the <p> tag shouldn't have any additional margin underneath, something like this... but this exact impl. might not be the right way to handle it.
Screen Shot 2022-04-06 at 3 34 53 PM

Recent Invoices:
- I'm thinking we should probably bump the "Create Invoice" primary button to a text-only button, or remove entirely ..? It currently feels a bit too prominent, esp. considering we remove the option after the list begins populating .. hmmm..
Screen Shot 2022-04-06 at 3 46 42 PM
- The :last-child list item bottom border should be hidden, applies to the Recent Transactions section as well, or any applicable section.
Screen Shot 2022-04-06 at 3 49 46 PM

dennisreimann:
@dstrukt Addressed your points in [21aa1b9](https://github.com/btcpayserver/btcpayserver/pull/3530/commits/21aa1b96a08b2f5a4f9fd0f3ec65fc440e7d18b1).

> Wondering if we need 9 Y-axis increment values on the Wallet Balance, might look nicer with say 5?

This depends on the data and is rendered by the chart library - I haven't found an option to customize that setting yet.

> - The :last-child list item bottom border should be hidden, applies to the Recent Transactions section as well, or any applicable section. 

I've added this in btcpayserver/btcpayserver#3565 to prevent conflicts. The design update PR should also get merged for v1.5.0, so this should be fine :)

dennisreimann:
Alright, all data tiles are implemented now 🎉

grafik

BTCPayServer/Views/UIStores/Dashboard.cshtml
```diff
@@ -2,67 +2,150 @@
...
+            <div class="modal-header">
+                <h4 class="modal-title" id="WhatsNewTitle">What's New</h4>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close">
+                    <vc:icon symbol="close"/>
+                </button>
+            </div>
+            <div class="modal-body">
+                <h5 class="alert-heading">Updated in v1.5.0</h5>
+                <p class="mb-0">Stores now have a neat dashboard like the one you see here! 🗠🎉</p>
+                <hr style="height:1px;background-color:var(--btcpay-body-text-muted);margin:var(--btcpay-space-m) 0;" />
```
dennisreimann, line 29: We need to add proper text here.

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -181,12 +182,101 @@ private async Task<ViewCrowdfundViewModel> GetInfo(AppData appData)
...
+                var date = (DateTimeOffset.UtcNow - TimeSpan.FromDays(i)).Date;
+                if (!series.Any(e => e.Date == date))
+                {
+                    series = series.Append(new SalesStatsItem
+                    {
+                        Date = date,
+                        Label = date.ToString("MMM dd", CultureInfo.InvariantCulture)
+                    });
+                }
+            }
```
dennisreimann, line 259: I kinda rushed this part - there might be a more efficient way to aggregate the data and fill the gaps in one go.

BTCPayServer/Views/UIStores/Dashboard.cshtml
dstrukt: @pavlenex Have any suggestions?

Bangalisch:
Yo, 

Looks super cool and it feels satisfying to see the dash tab filled up. 
Tho one point I think might be something to consider. 

When create a new store the dash is filled with the Setup instructions (store, wallet and LN) 
If you connect a Bitcoin wallet, the dash immediately changes to the full fledge view with containers. 
You now kinda lost the flow of setting up the store, as you don't get the LN option no more. 
Where if you do it the otherway around, so hook up a LN first it does give you the IMO right view. See screenshots for what I mean. The difference in TEST (Bitcoin wallet first) and TEST2 ( LN wallet first) :) 
screenshot 21

screenshot 23

screenshot 22

Maybe do the same for the Bitcoin wallet, until user decides either to active or not. Then change afterwards to full view.

dennisreimann:
@Bangalisch it is implemented lime this, because the balance graph and recent transactions are only available for on chain wallets, whereas we don't have any data that's special for Lightning.

BTCPayServer/Views/UIStores/Dashboard.cshtml
pavlenex: I think it's okay as it is but perhaps something like:

`Brand-new dashboard for your store showcasing most important data`?

pavlenex:
@dennisreimann @NicolasDorier anything else you'd like to do with this PR? To me it looks like it's ready for some final user-testing and feedback, which I did and looks quite solid, good job 🎉

pavlenex:
It seems balance doesn't account for sent out transacitons.
To replicate:
1. Have 100 BTC Balance
2. Send 50 BTC to a different store.
3. Mine some blocks (I mined 10)
3. Check dashboard graph it doens't change, the graph still shows old balance.
Screenshot 2022-04-11 at 12 00 41
Screenshot 2022-04-11 at 12 02 10

NicolasDorier:
@pavlenex it is expected, the graph is basically refreshing every 10 min

NicolasDorier:
mmh maybe I can just rewrite the last datapoint to be equal to `available_balance`.

NicolasDorier:
@pavlenex @dennisreimann I pushed a commit which make sure the last point is equal to the latest balance, as well as increasing the number of point for weekly histogram. This is needed to prevent the last line to spread over a full day when it's only the transaction 10 min ago that moved the value.

NicolasDorier:
Let's fucking go

PR #3575, opened 2022-03-26
added inputmode attr to relevant form input fields

dafunction:
When making changes to numerical amounts in various input fields across the BTCPay Server app on iOS devices, I noticed the decimal pad is not shown when selecting the form input.  It seems even though the form input `type` is number, the system still defaults to the normal keyboard, but on the number page, which is technically fine.  However, adding the `inputmode` attribute set to `decimal` or `numeric` will prompt the system to display the decimal pad or number pad instead when the input field is selected.

This should only impact iOS devices, and Android devices should remain the same since they already show the decimal pad for the input fields in question.

old
Pasted image 20220325213053
new
Pasted image 20220325213235

references:  [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode), [Caniuse](https://caniuse.com/input-inputmode)

### DIFF
--- a/BTCPayServer/Views/UIApps/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/UIApps/TemplateEditor.cshtml
@@ -89,7 +89,7 @@
                             </div>
                             <div class="form-group">
                                 <label class="form-label">Inventory (leave blank to not use inventory feature)</label>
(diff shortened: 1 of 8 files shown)


dennisreimann (approved):
Thanks for the contribution!

woutersamaey (approved):

PR #3577, opened 2022-03-27
Adjust user search input width

bolatovumar:
User search input is really small, especially on larger screens. This PR adjusts it and also makes sure it looks good on screens of different widths.

**Before:**
Capture

**After:**
Capture2


dennisreimann (approved):
Let's use this as a chance to unify it with the other pages, like the invoices list: I've adapated the header to have the Add button besides the headline. The max-width for the search field is the maximum width it can take on the M breakpoint.

Kukks (approved):

Issue #3583, opened 2022-03-28
Minor Visual bug in Claiming a Pull payment, weird bullet point by providing invalid address 


Bangalisch:
Yo, 

When you provide an invalid BTC address by claiming Funds from a Pull Payment, it gives a weird bullet point before the tekst. 

![PullPayment-Bulletpoint](https://user-images.githubusercontent.com/102448109/160456201-2e963d8b-2783-439b-bc78-e0daf4f118e7.JPG)

Bangalisch:
To provide to this a little more @pavlenex it's not only in invalid address. 
Seems its something todo with the Alert it sets off :) 

Came across this one just now : 
![PullPayment-Bulletpoint2](https://user-images.githubusercontent.com/102448109/160510637-6bb8a994-baa9-4e1f-a396-45962cdd51d4.JPG)

bolatovumar:
For the bullet point issue it seems to be on purpose because there _could_ be more than one error at once so it makes sense that they are shown with a bullet point. I guess we could hide the bullet point if there is only one error?

PR #3584, opened 2022-03-29
Adjust pull payment badge color

bolatovumar:
Adjusting badge color from black which turns to green on hover which doesn't work well when:

- Badge is green so text is the same color as background on hover
- Pretty much on all other badges that are not green on hover because green text looks weird on all of these badge colors

White color which this PR updates the badge text to works on all badge background colors.

address #3583


dennisreimann (approved):
tACK.

Kukks (approved):

PR #3585, opened 2022-03-29
Use the store's default currency when creating entities

### LINKED ISSUE #3581: Set default store currency in new invoices
Hey guys, 

When creating a new invoice the Currency field is empty, while we do set a default store currency. 
This feels very weird and as a extra since you can leave it empty and it does fill in the default store currency. 

CurrencyDefault

It would very nice if this would read (consistantly) that i set a default currency in the store. 

Greatings.

### LINKED ISSUE #3582: Pull payments has BTC as default currency, not store wide set default currency.
Hey guys, 

When creating New pull payment it allways displays BTC as standard return currency. 
WHere i set my default store currency to be USD, this feels very confusing. 

I (we, spoke with pavlenex) feel like this should be filled up with Store set Default Currency in this example USD. 

CurrencySetPullPayment

Greatings

dennisreimann:
Prefills the invoice/paymentrequest/pull payment creation forms currency inputs with the store's default currency.

Also cleans up an old code path we had for the invoice creation, in which the store was selectable. We don't offer that anymore and invoices are created explicitely for one store.

Closes #3581. Closes #3582.

### DIFF
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -189,6 +189,7 @@ namespace BTCPayServer.Tests
                 (await apiKeyRepo.GetKey(accessToken)).GetBlob().Permissions);
 
             //let's test the app identifier system
+            TestLogs.LogInformation("Checking app identifier system");
             authUrl = BTCPayServerClient.GenerateAuthorizeUri(s.ServerUri,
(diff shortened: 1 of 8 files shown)


bolatovumar (approved):

pavlenex:
Seems there are conflicts here we should resolve @dennisreimann because of the API test PR

dennisreimann:
Yes, I'll rebase my PRs.

Kukks (approved):

pavlenex (approved):
tACK

PR #3561, opened 2022-03-17
Add transaction info PATCH endpoint

bolatovumar:
Currently the only way to a comment and a label to a wallet transaction is through the UI. This PR adds support for updating wallet transaction comment and labels through Greenfield API.

PR #3588, opened 2022-03-30
Add label filter for onchain tx API endpoint

### LINKED ISSUE #3587: [GreenField API] Get store on-chain wallet transactions
Is there any way to find, via API, on-chain wallet transactions by the comment field or label?

The doc [https://docs.btcpayserver.org/API/Greenfield/v1/#operation/StoreOnChainWallets_ShowOnChainWalletTransactions](url) don't mention any way.

bolatovumar:
close #3587

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
@@ -173,6 +173,7 @@ namespace BTCPayServer.Controllers.Greenfield
             string storeId,
             string cryptoCode,
             [FromQuery] TransactionStatus[]? statusFilter = null,
+            [FromQuery] string? labelFilter = null,
             [FromQuery] int skip = 0,
             [FromQuery] int limit = int.MaxValue
         )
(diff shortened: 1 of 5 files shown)


BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -197,6 +198,23 @@ public async Task<IActionResult> UnReserveOnChainWalletReceiveAddress(string sto
                 filteredFlatList.AddRange(txs.ReplacedTransactions.Transactions);
             }
 
+            if (labelFilter != null)
```
bolatovumar, line 201: I'm not sure how to test this as there is currently no way to set the label through the API, only through the UI. Maybe if https://github.com/btcpayserver/btcpayserver/pull/3561 is merged in then I can use that to patch the transaction with a label and then test the filtering.

bolatovumar: Added test in https://github.com/btcpayserver/btcpayserver/pull/3588/commits/3b99b91a48c092bd55434cf16db57bb2c3f9ac9c

PR #3603, opened 2022-04-02
Delete LN addresses on store delete

### LINKED ISSUE #3596: Lightning Address doesn't get deleted when deleting store

bolatovumar:
fixes #3596 by deleting LN URL addresses associated with a store when it's deleted. Also added a startup task to delete LN URL addresses which do not belong to any store.

### DIFF
--- a/BTCPayServer/Hosting/MigrationStartupTask.cs
+++ b/BTCPayServer/Hosting/MigrationStartupTask.cs
@@ -175,6 +175,12 @@ namespace BTCPayServer.Hosting
                     settings.LighingAddressSettingRename = true;
                     await _Settings.UpdateSetting(settings);
                 }
+                if (!settings.UnreachableLightningAddressRemove)
+                {
+                    await RemoveLightningAddressesOfDeletedStores();
+                    settings.UnreachableLightningAddressRemove = true;
+                    await _Settings.UpdateSetting(settings);
+                }
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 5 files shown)


BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -171,10 +171,16 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
                 }
                 if (!settings.LighingAddressSettingRename)
                 {
-                    await MigrateLighingAddressSettingRename();
+                    await MigrateLightningAddressSettingRename();
```
bolatovumar, line 174: There was a typo here (`Lighing` instead of `Lightning`).

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -342,6 +353,9 @@ public async Task<bool> DeleteStore(string storeId)
                 .ToArrayAsync();
             foreach (var w in webhooks)
                 ctx.Webhooks.Remove(w);
+
+            await DeleteLightningAddresses(storeId);
```
bolatovumar, line 357: Delete lightning address for the store that is being deleted so we don't end up with orphaned ones.

NicolasDorier:
Suggested on https://github.com/btcpayserver/btcpayserver/pull/3613 to instead having a separate table for LN Address with delete on cascade.

Issue #3604, opened 2022-04-02
Add option to select preferred payment method


nosovk:
There is an option to create PAT Button. It helps a lot, because it's easy to integrate it in any site. We just have to modify some get params:
https://btcpayserver.com/api/v1/invoices?storeId=storeid&price=5000&currency=USD

Where *storeid* is our store identifier.
*price* is amount to pay
*currency* is base currency, to which all other currencies will be converted.

It seems pretty solid, till I stared creating different buttons on my site.
Like "Pay LTC", "Pay DASH" etc. I found that there is no way to setup default payment method in invoice generation link. As a result we can't create "Pay DASH" button, because actual payment scheme will be set to default for store.

It would be nice to have additional get param - *defaultPaymentMethod*, in which we could declare which of payment methods should be shown from start.
Like that one:
https://btcpayserver.com/api/v1/invoices?storeId=storeid&price=5000&currency=USD&defaultPaymentMethod=DASH
That link should open payment widget with DASH coin selected and equivalent of 5000usd

PR #3606, opened 2022-04-03
Add ability to set default payment method for pay button

bolatovumar:
Added an option for setting a default payment method for pay button (e.g., you can set it to LN, or Doge or whatever by default instead of using whatever is the default option for the store).

The code for the default payment method in the generated HTML only shows up if user changes the selection from the default one. There is no need to include the default one in the HTML because that's what will be used anyway.

Capture

close https://github.com/btcpayserver/btcpayserver/issues/3604

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.cs
+++ b/BTCPayServer/Controllers/UIStoresController.cs
@@ -422,6 +422,29 @@ namespace BTCPayServer.Controllers
 
         void SetCryptoCurrencies(CheckoutAppearanceViewModel vm, Data.StoreData storeData)
         {
+            var choices = GetEnabledPaymentMethodChoices(storeData);
+            var chosen = GetDefaultPaymentMethodChoice(storeData);
+
+            vm.PaymentMethods = new SelectList(choices, nameof(chosen.Value), nameof(chosen.Name), chosen?.Value);
+            vm.DefaultPaymentMethod = chosen?.Value;
+        }
+
+        CheckoutAppearanceViewModel.Format[] GetEnabledPaymentMethodChoices(Data.StoreData storeData)
+        {
+            var enabled = storeData.GetEnabledPaymentIds(_NetworkProvider);
+            
+            return enabled
+                .Select(o =>
+                    new CheckoutAppearanceViewModel.Format()
+                    {
+                        Name = o.ToPrettyString(),
+                        Value = o.ToString(),
+                        PaymentId = o
+                    }).ToArray();
+        }
+
+        CheckoutAppearanceViewModel.Format GetDefaultPaymentMethodChoice(Data.StoreData storeData)
+        {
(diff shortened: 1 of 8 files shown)


BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
bolatovumar, line 1: Added this while I was at it which led to a bunch of `?` added to explicitly declare things as nullable.

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -421,6 +422,29 @@ public IActionResult CheckoutAppearance()
         }
 
         void SetCryptoCurrencies(CheckoutAppearanceViewModel vm, Data.StoreData storeData)
+        {
+            var choices = GetEnabledPaymentMethodChoices(storeData);
```
bolatovumar, line 426: Created these two new methods (`GetEnabledPaymentMethodChoices` and `GetDefaultPaymentMethodChoice`) so that we can reuse the default payment options code that's used for the store checkout settings.

BTCPayServer/Controllers/UIStoresController.cs
NicolasDorier: when you do this, make sure to fix all the warnings

bolatovumar: @NicolasDorier yeah, I did. All warnings are fixed now.

NicolasDorier:
test failure unrelated

NicolasDorier:
@bolatovumar the default payment method should have an option `Use store's settings default`, selected by default.

bolatovumar:
> @bolatovumar the default payment method should have an option `Use store's settings default`, selected by default.

@NicolasDorier the default selection for the pay button is exactly the same as the one used for the store (so, if you select LN as the default method for the store the pay button selection will change with it). It's using the same code underneath to get the default selected value.

NicolasDorier:
@bolatovumar I saw this, but what I ask is different, I think you are selecting the current default in the listbox. But then imagine the button is copied, then the user change the store's default currency. The button will still use the previous one.

bolatovumar:
> @bolatovumar I saw this, but what I ask is different, I think you are selecting the current default in the listbox. But then imagine the button is copied, then the user change the store's default currency. The button will still use the previous one.

@NicolasDorier alright, updated in this commit: 4af669f40fdc3b68e898c81a7b0d0ffd7e31b5b5

NicolasDorier:
Code ACK, @pavlenex you can give it a try?

pavlenex:
Yeah LGTM @NicolasDorier ✅

dennisreimann (approved):
utACK

PR #3609, opened 2022-04-04
CI test fixes

dennisreimann:
Trying to fix that damn test.

### DIFF
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -136,7 +136,7 @@ namespace BTCPayServer.Tests
             var authUrl = BTCPayServerClient.GenerateAuthorizeUri(s.ServerUri,
                 new[] { Policies.CanModifyStoreSettings, Policies.CanModifyServerSettings }, applicationDetails: (appidentifier, new Uri(callbackUrl))).ToString();
             TestLogs.LogInformation($"Going to auth URL {authUrl}");
-            s.Driver.Navigate().GoToUrl(authUrl);
+            s.GoToUrl(authUrl);
             Assert.Contains(appidentifier, s.Driver.PageSource);
             Assert.Equal("hidden", s.Driver.FindElement(By.Id("btcpay.store.canmodifystoresettings")).GetAttribute("type").ToLowerInvariant());
             Assert.Equal("true", s.Driver.FindElement(By.Id("btcpay.store.canmodifystoresettings")).GetAttribute("value").ToLowerInvariant());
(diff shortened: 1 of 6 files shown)


bolatovumar:
lol it failed anyway

woutersamaey:
Omg 😱

dennisreimann:
@NicolasDorier This one downgrades NBXplorer again, see my [comments in the chat](https://chat.btcpayserver.org/btcpayserver/pl/3bdjfy76mfrmbqn531w8rau66w): There seems to be a regression exposed by the `CanPayWithTwoCurrencies` test. We can use this PR to also fix that one.

BTCPayServer.Tests/docker-compose.altcoins.yml
```diff
@@ -89,11 +89,11 @@ services:
     expose:
       - "4444"
   nbxplorer:
-    image: nicolasdorier/nbxplorer:2.3.6
+    image: nicolasdorier/nbxplorer:2.2.18
```
dennisreimann, line 92: We should find a fix for the new NBXplorer version, downgrading won't help.

BTCPayServer.Tests/ApiKeysTests.cs
```diff
@@ -19,7 +20,7 @@ namespace BTCPayServer.Tests
 {
     public class ApiKeysTests : UnitTestBase
     {
-        public const int TestTimeout = TestUtils.TestTimeout;
+        public const int TestTimeout = 120_000;
```
dennisreimann, line 23: As simple as it is, this is the actual fix. 

The default timeout of 60 seconds is too short. The test simply takes very long and given 60 seconds it erratically times out in various places. The test does A LOT of requests back and forth to test various API keys. Sometimes the latency seems to be heavy and in sum it takes rather long.

woutersamaey: Are we sure there isn't a problem with the API keys page which makes it take unusually long? I too have had the impression that the API page is slow, but it's just a feeling.

dennisreimann: @woutersamaey I've [tested this at length](https://app.circleci.com/pipelines/github/btcpayserver/btcpayserver?branch=pull%2F3609&filter=all) and there's not a single or clear reason why it fails (sometimes it just works). I think this is just a matter of the latency between requests on CI — you can run the test locally and it perfoms regularly like any other of our Selenium tests.

woutersamaey: Thanks for your hard work @dennisreimann !

Issue #1256, opened 2020-01-07
Generation of a digital receipt


bjadel:
**Is your feature request related to a problem? Please describe.**
If you run a point of sale it might be legally necessary to offer a digital receipt to the customer.

**Describe the solution you'd like**
Generation of a digital receipt in the form of a (signed) PDF, which can be downloaded via QRCode. This could be implemented as btcTransmuter service. When a new invoice is created, then generate a digital receipt. In the GUI a QRCode could be displayed for a successful transaction, which the user can use to download the receipt if she/he likes. 

In Germany the receipt has to fulfill the following requirements:
https://www.gesetze-im-internet.de/kassensichv/__6.html

The question is what is needed in other countries.

bolatovumar:
Same feature request was discussed on the Lightning dev mailing list. One problem with providing digital receipts is that every country has their own requirements for what receipts must contain.

bjadel:
> Same feature request was discussed on the Lightning dev mailing list. One problem with providing digital receipts is that every country has their own requirements for what receipts must contain.

Yeah, that's right. That's why a very configurable variant would also be important.

pavlenex:
What we can do maybe is show a print view in `Invoice` and perhaps allow field selection (we do something similar for `Payment Request` so for each invoice you'd have a custom field to enter data you need and then select the fields you want included, it's very hard to make this part customizable, so if you have any real receipts it would be helpful to see what kind of information you need.

woutersamaey:
What if we create a text field supporting HTML and placeholders? Similar to an email template system? That way people can customize the receipt as per their countries requirement. Some default or presets could be provided for common countries?

pavlenex:
@woutersamaey That's good idea. Also when invoice is paid, if user uses print-mode of some sort, we can show print icon at the successfully paid invoice at the checkout so they can click right away and print it out if needed.

Kukks:
Use transmuter for this instead of cramming all these features in there.. Currently you can send an email receipt but we can extend it to show a page if you want. it also already handles html + templating

nvk:
Hello coming to comment from twitter thread. Most customers buying from a business account will need to print the invoice, some BTCPayServer are not running emails correctly for receipt.

I propose something as simple as a HTML print call on the page. And if you want to get fancy, a printer friendly css style format with proper invoice headings would be nice.

[...]

PR #3611, opened 2022-04-05
Store Emails

Kukks:
This PR starts adding email sending capabilities based on invoice events.

PR #3612, opened 2022-04-05
Public Invoice receipt

Kukks:
This PR introduces an option for stores to have a public invoice receipt page for customers. Can be a store-wide setting or on an invoice level. The merchant can optionally also add in additional data, by providing it under the invoice's `metadata` property in `receiptData` JSON property.


dstrukt:
@pavlenex @Kukks Happy to provide design work, but might need a bit of context / scoping here, lot going on in this PR.

pavlenex:
@dstrukt In the past, we allowed people to view invoices from the btcpay server. This was by default and anyone could preview a hashed url of the invoice which served as a receipt. However, some members were concerned about the privacy implications of this, and we then disabled people to view an invoice.

The use case here is that a merchant would want to have their invoice previewed publicly, like a receipt. I think this was requested [here](https://github.com/btcpayserver/btcpayserver/discussions/2214) 

I actually think that we should call it `receipt` and also add it to the invoice checkout itself, so that a merchant can troubleshoot an invoice and verify if it's been paid without having to access and perhaps having the need to even have access to the backoffice.

@Kukks I tested the PR, and I like that this is optional. I saw you've added it in the `Checkout appearance` and I think if we have a button on the invoice saying `Preview receipt` then it makes sense for it to be there.

Otherwise, I was not able to test this. I tried visiting the invoice URL when I'm logged out but it didn't work, can you provide some guidiance on how we can actually preview the invoice as a non admin?

Kukks:
> @dstrukt In the past, we allowed people to view invoices from the btcpay server. This was by default and anyone could preview a hashed url of the invoice which served as a receipt. However, some members were concerned about the privacy implications of this, and we then disabled people to view an invoice.
> 
> The use case here is that a merchant would want to have their invoice previewed publicly, like a receipt. I think this was requested [here](https://github.com/btcpayserver/btcpayserver/discussions/2214)
> 
> I actually think that we should call it `receipt` and also add it to the invoice checkout itself, so that a merchant can troubleshoot an invoice and verify if it's been paid without having to access and perhaps having the need to even have access to the backoffice.
> 
> @Kukks I tested the PR, and I like that this is optional. I saw you've added it in the `Checkout appearance` and I think if we have a button on the invoice saying `Preview receipt` then it makes sense for it to be there.
> 
> Otherwise, I was not able to test this. I tried visiting the invoice URL when I'm logged out but it didn't work, can you provide some guidiance on how we can actually preview the invoice as a non admin?

Some more insight into why we disabled it:
> We have an option when creating an invoice via API to add data `posData`. When this is added, we display it "prettily" on the invoice details. BUT, some users ended up using it to store highly sensitive and compromising data in it and so we hurried up and made the invoice detail page viewable by store admins only.

Since then, `posData` is still available, but we've restructured it under a `metadata` object. Metadata can store all this sensitive information without it being displayed, unless you store data under metadata => posData.  
So, we can actually restore the original functionality (but people would need to ensure they are still not storing compromising data under `posData`).

BUT, we can do better. An admin overview of an invoice should not be what the customer sees and has as a receipt or proof of payment to a merchant. 

The goals of this PR is to:
* Have a new page tailored to consumers to have a proof of payment/ receipt.
* Potentially show a QR code on this page so that merchants can quickly verify the receipt is real.
* Potentially have a link on the invoice checkout page on the paid invoice screen.
* Have a link to receipt page from invoice details page.
* Make public receipt opt-in for each store.
* Allow a merchant to specify additional data to render on this UI `metadata=> receiptData`

Kukks:
> > @dstrukt In the past, we allowed people to view invoices from the btcpay server. This was by default and anyone could preview a hashed url of the invoice which served as a receipt. However, some members were concerned about the privacy implications of this, and we then disabled people to view an invoice.
> > The use case here is that a merchant would want to have their invoice previewed publicly, like a receipt. I think this was requested [here](https://github.com/btcpayserver/btcpayserver/discussions/2214)
> > I actually think that we should call it `receipt` and also add it to the invoice checkout itself, so that a merchant can troubleshoot an invoice and verify if it's been paid without having to access and perhaps having the need to even have access to the backoffice.
> > @Kukks I tested the PR, and I like that this is optional. I saw you've added it in the `Checkout appearance` and I think if we have a button on the invoice saying `Preview receipt` then it makes sense for it to be there.
> > Otherwise, I was not able to test this. I tried visiting the invoice URL when I'm logged out but it didn't work, can you provide some guidiance on how we can actually preview the invoice as a non admin?
> 
> Some more insight into why we disabled it:
> 
> > We have an option when creating an invoice via API to add data `posData`. When this is added, we display it "prettily" on the invoice details. BUT, some users ended up using it to store highly sensitive and compromising data in it and so we hurried up and made the invoice detail page viewable by store admins only.
> 
> Since then, `posData` is still available, but we've restructured it under a `metadata` object. Metadata can store all this sensitive information without it being displayed, unless you store data under metadata => posData. So, we can actually restore the original functionality (but people would need to ensure they are still not storing compromising data under `posData`).
> 
> BUT, we can do better. An admin overview of an invoice should not be what the customer sees and has as a receipt or proof of payment to a merchant.
> 
> The goals of this PR is to:
> 
> * Have a new page tailored to consumers to have a proof of payment/ receipt.
> * Potentially show a QR code on this page so that merchants can quickly verify the receipt is real.
> * Potentially have a link on the invoice checkout page on the paid invoice screen.
> * Have a link to receipt page from invoice details page.
> * Make public receipt opt-in for each store.
> * Allow a merchant to specify additional data to render on this UI `metadata=> receiptData`

Additionally, this feature can be tied in with #3611 to send receipts to customers!

pavlenex:
>The goals of this PR is to:
Have a new page tailored to consumers to have a proof of payment/ receipt.
Potentially show a QR code on this page so that merchants can quickly verify the receipt is real.
Potentially have a link on the invoice checkout page on the paid invoice screen.
Have a link to receipt page from invoice details page.
Make public receipt opt-in for each store.
Allow a merchant to specify additional data to render on this UI metadata=> receiptData

@Kukks This sounds like a good plan. 
- Where would a QR point to, how could a merchant quickly verify the transaction without having to go to the backoffice? 

Since this is still a draft I assume not all functions are there, but concept ACK, I think this will turn into a very practical feature and connect a few missing dots we currently have.

Kukks:
> > The goals of this PR is to:
> > Have a new page tailored to consumers to have a proof of payment/ receipt.
> > Potentially show a QR code on this page so that merchants can quickly verify the receipt is real.
> > Potentially have a link on the invoice checkout page on the paid invoice screen.
> > Have a link to receipt page from invoice details page.
> > Make public receipt opt-in for each store.
> > Allow a merchant to specify additional data to render on this UI metadata=> receiptData
> 
> @Kukks This sounds like a good plan.
> 
> * Where would a QR point to, how could a merchant quickly verify the transaction without having to go to the backoffice?
> 
> Since this is still a draft I assume not all functions are there, but concept ACK, I think this will turn into a very practical feature and connect a few missing dots we currently have.

> * Where would a QR point to, how could a merchant quickly verify the transaction without having to go to the backoffice?
It would still require the merchant to log into the backend. 

The Qr code could simply lead to the actual invoice details page. OR it could just be a URL of itself, where if the store admin is logged in, it would show a "verified checkmark" underneath

pavlenex:
Closes #1256

Kukks:
Here are some screens of current design and capabilities.

msedge_E8PCdxQ1qz
rider64_UaPwPRRkAj
Postman_TQWKHDwHUz
Postman_wn1U5D8CKD
msedge_t7lLjDJpYB
msedge_CLRogtKgMF
msedge_KT429yhPlC
msedge_qDV7nMkPgG
msedge_E8PCdxQ1qz

dstrukt:
Pretty much everything, with some A/B decisions we would make that would carry throughout. Still might tinker around with this a bit more, but overall what @Kukks already did was great, so this just explores some of it a little further!

Screen Shot 2022-06-09 at 2 12 22 AM
Screen Shot 2022-06-09 at 2 12 37 AM
Screen Shot 2022-06-09 at 2 12 46 AM
Screen Shot 2022-06-09 at 2 13 09 AM
Screen Shot 2022-06-09 at 2 13 20 AM
Screen Shot 2022-06-09 at 2 13 26 AM

dennisreimann:
Some metadata for testing:

```json
{
  "metadata": {
    "receiptShowQR": true,
    "receiptShowPayments": true,
    "receiptData": {
        "Product": "Ice Cream",
        "Location": "https://www.gozo.com",
        "List": [1, 2, 3],
        "Nested": {
            "Hello": "World",
            "Boolean": true
        }
    }
  }
}
```

I'm not sure if it should be configurable like that though. The `showQR` and `showPayments` should be default or configurable by the user. They feel superfluous imho.

How would a merchant add the `receiptData`?

dennisreimann:
Updated the view.

## Light

grafik




## Dark 

grafik

dstrukt:
Looking good! 

- Tend to agree the QR is opt-in, as it has cool potential use-cases, but isn't for everyone.
- Show payments, could also see that being opt-in for a client who wants to see the details of the payment, but this is a bit trickier, and might lean for it to be on by default ... will think on it a bit more.

Excited to see the POS data formatted in simple columns, that should definitely be displayed by default :) 

Have wanted to minify the payment details sections where we can, e.g.
- `Payments` instead of `Payment Details`
- Transaction ID becomes a column header, and the body is still a link, but represented by an ellipsed `8ad9if...0ieo39u` to get it all in one line

- It doesn't always make sense to stack rows in the table, but maybe we explore doubling up, where it makes sense, in the future.

Don't have to get to everything in this PR, just pointing out some areas I see potential for improvements, as this will apply to Payment Request stuff as well, and can be done when we address those changes.

dennisreimann:
For QR and Payments: We could make them query parameters to provide an easier and more flexible way to show that info. 

The trasnactions and POS data tables are wrapped in a `responsive-table` div, so they'll be scrollable on mobile. Not great, but the view works when printed and I think we should rework it when we get to redo the other tables in a better/more responsive way.

dstrukt:
Sounds good @dennisreimann!

dstrukt (approved):

Kukks:
The data under additional data is not the product information stuff that it populated in the invoice, but rather specific data the invoice creator passed on to specifically show on the receipt page. 

We could do an additional pr that let's you specify showReceiptProductInfo which would show a new section with the that info. Can potentially also do showReceiptCustomerInfo

dstrukt:
Definitely think that'd be be helpful to have access to the product information from `showReceiptProductInfo`.

For customer information, who, outside the admin or store user, would need access to that information?

Kukks:
> Definitely think that'd be be helpful to be able to output the product information from `showReceiptProductInfo`.
> 
> That's interesting, for customer info., who, outside the admin or store user, would need access to that information? I ask, cause I could see that being a potential privacy issue.

Maybe for the customer to confirm they have the right address?

dstrukt:
Ahh I see, there's still the privacy issue involved, but maybe that's just being overly cautious.

Don't have a strong opinion either way, but good to know it's an option.

NicolasDorier:
1. The metadata shouldn't be exposed in the receipts. We don't know if people have put confidential info in there
2. Metadata shouldn't be used to put things with logic inside. Only exception has been OrderId and TaxIncluded and I think refundEmail for backward compat reason.

I think there should have no receipts data associated to an Invoice, just showing the history of payments is fine.

Kukks:
Only a well known Metadata key is exposed, not all the Metadata.
@nicolasdorier

Kukks:
I disagree, there needs to be customizability options on the receipt to fit
various usecases.

dennisreimann:
Ok, how do we wanna solve the customizability? I think it is now very convenient to do all that via the API, which ist the only way right now, correct?

> For QR and Payments: We could make them query parameters to provide an easier and more flexible way to show that info.

Though this is flexible, it also doesn't seem very intuitive and my impression is we should just make that default and display it all the time.

Will we do the `showReceiptProductInfo`/meta data in a separate PR? That way we can get the basics done and discuss the rest and finetune afterwards.

Kukks:
Query string settings is a client side solution, which the customer can control. Invoice based setting provides the merchant the full control over visibility. Personally l find that payment data should be controlled by the merchant as they might not want to have a shareable and linkable page to their wallet.

Kukks:
Query string settings is a client side solution, which the customer can control. Invoice based setting provides the merchant the full control over visibility. Personally l find that payment data should be controlled by the merchant as they might not want to have a shareable and linkable page to their wallet.

dstrukt:
Will address more when I get a moment, but w.r.t to No.6 @pavlenex yes you're correct, either far-right or left with the primary CTA stacked next to secondaries, etc.. not nested between. I believe we've been doing far left while flattening and consolidating all of our CTAs.

Kukks:
> I have several questions, so the current flow is:
> 
> 1. User wants to have a public receipt for their invoices.
> 2. User enables this in the `Checkout appearance`
> 3. Receipt is available from admin dashboard, but how can a customer access this?
> 4. I do see a `QR on the mocks` and payment information ,but all I see locally is this:
>    Screenshot 2022-06-27 at 11 01 24
> 5. I see `additional details` how does one add those, where exactly is it expected to be added?
> 6. @dstrukt is it common that a primary CTA on the invoice page (refund button) is in between the two secondary CTA's? Shouldn't it be on left, right and the other two secondary CTA's next to it?

2) Can be enabled on Checkout appearance, or on a per-invoice basis via API.
3) API returns the link to receipt when you fetch invoice details via API. It is `invoice url/receipt` though
4) In this PR, QR codes are not on by default. They need to be enabled when creating the invoice (via API)
5) via API

for 4, we can maybe add a setting somewhere in apps around receipt config.
for 5, we can add an additional setting to show product info when present. Additionally, can make the create invoice screen have some way to add this data to it

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -147,6 +150,101 @@ public async Task<IActionResult> Invoice(string invoiceId)
...
+            public Dictionary<string, object>? AdditionalData { get; set; }
+            public bool ShowQR { get; set; }
+            public bool ShowPayments { get; set; }
+            public List<ViewPaymentRequestViewModel.PaymentRequestInvoicePayment>? Payments { get; set; }
+        }
+        
+        [HttpGet("i/{invoiceId}/receipt")]
+        public async Task<IActionResult> InvoiceReceipt(string invoiceId)
+        {
+            var i = await _InvoiceRepository.GetInvoice(invoiceId);
```
NicolasDorier, line 171: if `i is null` just fast return here. So you can remove a bunch of complicated logic later.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -147,6 +150,101 @@ public async Task<IActionResult> Invoice(string invoiceId)
...
+            var store = await _StoreRepository.GetStoreByInvoiceId(i.Id);
+            if (i is not null && i.InvoicePublicReceipt is null)
+            {
+                i.InvoicePublicReceipt = store?.GetStoreBlob().InvoicePublicReceipt;
```
NicolasDorier, line 175: can replace this condition in one like:
`i.InvoicePublicReceit ??= store.GetStoreBlob().InvoicePublicReceipt;`

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -147,6 +150,101 @@ public async Task<IActionResult> Invoice(string invoiceId)
...
+            }
+            if (i?.InvoicePublicReceipt is true)
```
NicolasDorier, line 177: if you checked for `i` to not be null before, you can just replace this with `if(i.InvoicePublicReceipt)`

NicolasDorier: Actually you should test for `if(!i.InvoicePublicReceipt)` and fast return.

BTCPayServer/Views/UIInvoice/PosData.cshtml
```diff
@@ -41,7 +41,7 @@
                 else
                 {
                     <td colspan="2">
-                        @Html.Raw($"<h{Model.Level + 3} class='mt-3'>{key}</h{Model.Level + 3}>")
+                        @Html.Raw($"<h{Model.Level + 3} class=\"mt-4 mb-3\">{key}</h{Model.Level + 3}>")
```
NicolasDorier, line 44: `{key}` isn't sanitized

NicolasDorier:
We should not use Metadata to put things that have logic behind. This is not done for this.
The metadata are for the user to put any shit relevant to him, not to anything else.

If you want the receipt to be configurable, you should just store more data in `InvoiceEntity` (which is serialized as a blob in the db).

You also need to document those customization in the API with different fields.

pavlenex:
Since the receipt is a customer-facing thing, I suggest we implement it within our new upcoming checkout pages, because without having an easy way to provide a way for customer to grab it, I don't think many people would use it, API is certainly great for customization, but people need a way to quickly access it the invoice, I think this can be done by having a button on a checkout page once it's settled.

NicolasDorier:
@Kukks 

Can you ACK/NACK my changes:
1. I renamed `HideQR` and `HidePayments` to `ShowQR` and `ShowPayments`
2. I made the `Enabled` , `ShowPayments`, `ShowQR` being `true` by default. (I don't see why a merchant doesn't want receipt ability)
3. I added in the UI a checkbox to configure the `ShowPayment` and `ShowQR` in checkout appearance
4. I exposed the `ReceiptOptions` directly in StoreData, and ViewModels. This will allow us to add more receipt options in the future without having to hunt all the places to get/set + additional props to create.
5. I changed the behavior if an invoice doesn't specify any `ReceiptOptions`. Rather than initializing it by default to whatever options are set in the store's data at the creation of the invoice, I keep it null. The difference is that old invoices, or invoices who doesn't care about ReceiptOptions will still be able to show/hide the receipts by changing the values directly in the store settings, rather than those values being set when this invoice is created.

Question:

Why does the checkout doesn't have a link to the public receipt? This is in checkout appearance, but doesn't appear in the checkout. I think @dstrukt should tell us how to present the link from the checkout to the receipt

Still needed:

Selenium tests.

Kukks:
Ack, I have some improvements around the checkout planned to making the receipt more visible:

* When no redirect url is specified, use the receipt link for the checkout success button 

* make apps not set the redirect url back to themselves but add the app url to the receipt data and have it rendered

* when redirect url is set and receipts enabled, show a receipt button on checkout success too

dstrukt:
Like and agree with @Kukks on all points.

For point no.3 if that means two buttons to choose from, then ACK from me!

For point no.2, just want to understand, you would be presenting the user with the receipt, and to navigate back to the app you would have to follow the link in the receipt?

Only other question was going be what qualifies as "success" for an on-chain payment specifically .. is the receipt shown after X confirmation, or simply after the payment has been sent?

Kukks:
Yes, if the payment is still processing, the receipt page is accessible but
shows an alert warning.

https://chat.btcpayserver.org/files/umnzds1cq3d75mm7irquobwr8a/public?h=PqD5V4MhqiOoFNNNRfRt8Koo8tA4oi4Y2-Z7tXGJzs4

NicolasDorier:
@Kukks can you add a some tests? as well to expose the receipt as you said, else nobody will ever know this feature exists.

dstrukt:
Have a lot to say about the checkout UI, but considering we're getting to that for 1.7.0, this is perfectly acceptable for now and there's no reason to continue to delay merging this for those nits :)

BTCPayServer/Views/UIInvoice/Checkout-Body.cshtml
```diff
@@ -204,8 +204,12 @@
...
                 </div>
                 <div class="success-message">{{$t("This invoice has been paid")}}</div>
-                <a class="action-button" :href="srvModel.merchantRefLink" v-show="!isModal">
-                    <span v-html="$t('Return to StoreName', srvModel)"></span>
+                <a class="action-button" :href="srvModel.merchantRefLink" v-if="!isModal && srvModel.merchantRefLink">
+                    <span v-if="srvModel.receiptLink != srvModel.merchantRefLink" v-html="$t('Return to StoreName', srvModel)"></span>
+                    <span v-else v-html="$t('View receipt', srvModel)"></span>
+                </a> 
+                <a class="action-button" :href="srvModel.receiptLink" :target="isModal?'_blank':'_top'" v-if="srvModel.receiptLink &&  (srvModel.merchantRefLink != srvModel.receiptLink) || isModal">
+                    <span v-html="$t('View receipt', srvModel)"></span>
```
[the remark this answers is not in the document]
Kukks: fixed

NicolasDorier:
I updated the translations, now we have "View receipt"

Kukks:
They are configurable, through the api, by the Marchant. As a merchant, you
may not want to expose the used address or destination publicly as it may
leak. However, we can always add a toggle for the user if it is enabled by
merchant

PR #3613, opened 2022-04-05
Fix LN address store orphan fix

### LINKED ISSUE #3596: Lightning Address doesn't get deleted when deleting store

Kukks:
Fixes #3596

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
+using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Abstractions.Extensions;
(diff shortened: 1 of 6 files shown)


pavlenex:
@Kukks We already have a PR for this? https://github.com/btcpayserver/btcpayserver/pull/3603

Kukks:
> @Kukks We already have a PR for this? #3603

That's what I get for coding over the Atlantic ocean :(

bolatovumar:
@Kukks feel free to merge whichever PR you think is more robust.

NicolasDorier:
I don't think using plugin infrastructure for that is good idea, it's not a plugin. I would rather use the event aggregator.

NicolasDorier:
Actually @Kukks can you remove all those hacks and instead just having a new table for LN Address with just a DELETE ON CASCADE? That way it will just work.

NicolasDorier:
Closing in favor of https://github.com/btcpayserver/btcpayserver/pull/3638

Issue #3567, opened 2022-03-22
Checkout page improvements


pavlenex:
This is an issue tracker for the checkout page. Here, we'll try to keep a list of things we'd like to improve when redesigning a page. Note that the checkout page is the only part of the BTCPay with which customer interacts, so it's a sensitive area and all the changes need to be carefully made. Improvements shouldn't just be for aesthetical reasons, we should try to understand how people interact with this page and areas that are problematic, then see how to solve those through good design.

- [ ] The **timer** should appear only after x% of the time is left. There's no need to cause anxiety, for example, we should show a timer if 25% of the time is left.
- [ ] We should explicitly mention that user needs to Scan to pay or copy in **text**. There have been numerous reports where a newbie users didn't know they had to scan a QR code or copy the address, etc. One good example of combining design with text on the checkout page is CashApp.
![Screenshot 2022-03-22 at 10 41 32](https://user-images.githubusercontent.com/36959754/159451873-4f86ba46-301b-4b42-8726-c78ba9c9fca5.png)
- [ ] The **currency selector** is one of the most problematic areas as people aren't aware that they can switch between different currencies. For LN and BTC, a long term solution would be  https://bitcoinqr.dev a unified QR code. For altcoins, meh who cares.
- [ ] It has been mentioned a few times that we may want to provide some additional **support/tooltip** for customers having trouble paying
- [ ] The way in which communicate the **rate** is not the best. Saying simply 1 BTC=39564 USD may not always be clear for users. Buyer is probably more interested in seeing the fiat amount together with the rate right away, however we hide this information behind a not very visible expander. 
![Screenshot 2022-03-22 at 10 45 37](https://user-images.githubusercontent.com/36959754/159452744-fd041ed3-c82c-4e8c-90d6-d56b13f7ae76.png)
- [ ] Once an **invoice expires**, there's no way to re-generate it, instead the user just sees "invoice expired" page which isn't really helpful. While this isn't related to this issue, it's good to have a discussion around how we handle expired invoices.

BenGWeeks:
Looking forward to this :-)

pavlenex:
https://github.com/btcpayserver/btcpayserver/issues/3567

BenGWeeks:
Please see comments here: https://github.com/btcpayserver/btcpayserver/discussions/3670

PR #3614, opened 2022-04-05
Specify PayJoin enabled in Payment Link heading

orangesurf:
I wish to make a suggestion to the way the "Payment link" heading is displayed in checkout.

Currently it shows "Payment link" even if PayJoin is enabled.

There is an icon and information on hover but I missed these, I think altering the heading would make it even more obvious.


dstrukt:
Do you mind posting screenshots?

I can follow along in the code, but I'm curious to see how it looks in context, thanks!

pavlenex:
@3609 
Screenshot 2022-04-08 at 12 08 57

Screenshot 2022-04-08 at 12 08 52

I think for now it's good to emphasize PayJoin. I however think once we do #3567 we really have to think how much of these complexities a user should know about and how much should be handled automatically by their wallet. We just have way too much information that a regular person paying in bitcoin doesn't know about and we should try to abstract those.

So for now I think this is okay, but in bigger perspective, we should really try to thinker about abstracting the complexities. I even think `payment link` itself is very confusing for a user.

dstrukt:
Agree @pavlenex, and thanks for the screenshots.

orangesurf:
Thanks for doing the screenshots @pavlenex. By "So for now I think this is okay," do you mean how it is currently or how the pull request is?

PR #3624, opened 2022-04-09
Fix visual bug when user clicks on "Create refund" without selecting an option

### LINKED ISSUE #3621: Minor Visual bug in "Issue refund"
Hey guys, 
Was going through 1.5 dashboard, LOOKS SEXY @dennisreimann !
Tho, after issuing a refund and making a mistake I came across this lil bug, not related to the 1.5 release. 

So whenever you go through the flow of the refund.
You get to pick what you want to pay back to the costumer, the "What to Refund" tab. 
Here its a MUST to pick a radial button before continue, well if you hit the "Create Refund" before selecting a radial, it breaks the tags of the Radials, the functionality of them stays in tact, its just visual.

screenshot 20

bolatovumar:
|Before|After|
|---|---|
|Capture2|Capture|

fix #3621

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -311,7 +311,7 @@ namespace BTCPayServer.Controllers
                             model.RefundStep = RefundSteps.SelectCustomAmount;
                             return View(model);
                         default:
-                            ModelState.AddModelError(nameof(model.SelectedRefundOption), "Invalid choice");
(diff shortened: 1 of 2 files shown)


dstrukt:
Nice!

Small nit, we should probably move it above the options under the title, or above the button at least.

It could lost under the button, esp. on mobile.

bolatovumar:
> Nice!
> 
> Small nit, we should probably move it above the options under the title, or above the button at least.
> 
> It could lost under the button, esp. on mobile.

@dstrukt moved it above the button and also updated the error message as "Invalid option" is pretty generic.

Capture

PR #3638, opened 2022-04-14
LN Address - db schema fix method

### LINKED ISSUE #3596: Lightning Address doesn't get deleted when deleting store

Kukks:
replaces #3613 and #3603

fixes #3596


BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -1713,11 +1713,6 @@ public async Task CanUseLNAddress()
             s.FindAlertMessage(StatusMessageModel.StatusSeverity.Success);
 
             s.Driver.ToggleCollapse("AddAddress");
-            s.Driver.FindElement(By.Id("Add_Username")).SendKeys(lnaddress1);
```
NicolasDorier, line 1716: why removing this?

BTCPayServer.Data/Migrations/20220414132313_AddLightningAddress.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "LightningAddresses",
+                columns: table => new
+                {
+                    Username = table.Column<string>(type: "TEXT", nullable: false),
+                    StoreDataId = table.Column<string>(type: "TEXT", nullable: false),
+                    Blob = table.Column<byte[]>( nullable: true)
```
NicolasDorier, line 26: wondering if we shouldn't start storing JSONB for postgres rather than bytes?

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -175,6 +178,12 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
                     settings.LighingAddressSettingRename = true;
                     await _Settings.UpdateSetting(settings);
                 }
+                if (!settings.LighingAddressOrphanFix)
```
NicolasDorier, line 181: it's not really a fix anymore, this is a migration to tables.

BTCPayServer/Hosting/MigrationStartupTask.cs
```diff
@@ -183,6 +192,49 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default)
...
+                await _Settings.GetSettingAsync<UILNURLController.LightningAddressSettings>(
+                    nameof(UILNURLController.LightningAddressSettings));
+
+            if (lightningAddressSettings is null)
+            {
+                return;
+            }
+
+            var storeids = lightningAddressSettings.StoreToItemMap.Keys.ToArray();
+            var existingStores = await ctx.Stores.Where(data => storeids.Contains(data.Id)).Select(data => data.Id ).ToArrayAsync();
```
NicolasDorier, line 209: `ToHashSetAsync()` if possible, else just `.ToHashSet` on the array.

BTCPayServer.Data/Migrations/20220414132313_AddLightningAddress.cs
Kukks: we should probably do this decision later and apply universally

BTCPayServer.Tests/SeleniumTests.cs
Kukks: It now updates and replaces instead

Issue #3630, opened 2022-04-12
Implement the new/latest Currency selector on the Pay Button


dstrukt:
This may have been intentional, but currently the new Currency selector is not implemented on the Pay Button view:

<img width="908" alt="Screen Shot 2022-04-12 at 4 00 55 PM" src="https://user-images.githubusercontent.com/6250771/163068571-89af90da-03ef-4175-95ca-b75bdd6a5c7d.png">

dafunction:
@dstrukt just to make sure i'm understanding, you're talking about the dropdown selector, yes?

something like this:
![2022-04-14 18-03-48 2022-04-14 18_04_33](https://user-images.githubusercontent.com/18316226/163491275-0c3c84fb-fbef-4a29-a02a-e2d35bbcd989.gif)

NicolasDorier:
His PR should have fixed it

dstrukt:
@dafunction This is just a visual inconsistency.

Notice how the carrot for the currency selector only appears on hover/select, unlike the Select input right below it, always present, same icon/glyph asset .. the two should, ideally, be the same.

Will check out the PR though.

PR #3642, opened 2022-04-14
add currency-selector input on Payment Request and Pay Button pages

dafunction:
fix for issue #3630


dennisreimann:
Is it correct to remove the Vue handlers for the Pay Button currency input here? Just saw the diff, haven't checked if it is working, but wanted to bring it up, because I think removing the Vue handlers might have unintended consequences.

dafunction:
@dennisreimann i can confirm i did test the change before i committed it and all seemed to be working fine.  i just followed the other implementations of `currency-selector` in the app.  i'm happy to add the vue handlers back to the input element in a separate commit if they are crucial.

dennisreimann:
@dafunction I just checked and fixed it: As mentioned in #3646, the pay button preview and code sample do not update without the Vue handlers. Can you give my PR a review?

PR #3653, opened 2022-04-19
Design updates

dennisreimann:
Contains:

- btcpayserver/btcpayserver-design#12
- btcpayserver/btcpayserver-design#33
- btcpayserver/btcpayserver-design#35
- btcpayserver/btcpayserver-design#37
- btcpayserver/btcpayserver-design#39

### DIFF
--- a/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
+++ b/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
@@ -11,7 +11,7 @@
   --btcpay-root-font-size: 16px;
   --btcpay-body-font-family: var(--btcpay-font-family-base);
   --btcpay-body-font-size: 0.875rem;
-  --btcpay-body-font-weight: 400;
+  --btcpay-body-font-weight: var(--btcpay-font-weight-normal);
   --btcpay-body-line-height: 1.6;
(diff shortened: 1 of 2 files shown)


dstrukt:
Looks great!

Notes:

1. Not part of the original list, but would like to improve the input focus:
- only 1 additional pixel inset, same opacity green, like so:
Screen Shot 2022-04-19 at 9 09 13 AM

2. I'm debating if only the `h1` and `h2` tags should be `bold` font-weight, with the subsequent headers h3, h4, etc.. being `semibold`. However, I do like how much more easily it all reads with everything bold ... hmmm. Thoughts?

dstrukt (approved):

dennisreimann:
1. is done in [d284ef0](https://github.com/btcpayserver/btcpayserver/pull/3653/commits/d284ef05d4c659a1406fcae1fee754ffc4695dcb)
2. actually I like the contrast the all bold headers bring. Though it is bolder, I also find it to be a bit more quiet, as we aren't dealing with three different font weights for headings and regular text.

Here's a quick comparison:

## All bold

bold

##  h1, h2 bold - rest semibold

semibold

dstrukt:
LGTM!

Let's roll with the bold headers!

PR #3654, opened 2022-04-20
Remove replace hardcoded BTC string to default network

NicolasDorier:


dennisreimann:
Yay, good point!

gruve-p:
Thanks for this PR.
Unfortunately it does not solve the issue that the balance is wrong (and the charts are not working) on the dashboard for GRS:

mainnet:
image

testnet:
image

gruve-p:
It seems the dashboard only populates BTC and not for altcoins

dennisreimann:
@gruve-p Thanks for reporting that, it's addressed in #3876.

PR #3658, opened 2022-04-21
Greenfield: Proper response for store not found

dennisreimann:
Previously the "network not configured" error message was also displayed when indeed the store could not be found. This is misleading, hence we should have a separate error for that case.

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.Store.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.Store.cs
@@ -7,11 +7,9 @@ using BTCPayServer.Client;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Configuration;
 using BTCPayServer.Data;
-using BTCPayServer.HostedServices;
 using BTCPayServer.Lightning;
 using BTCPayServer.Payments;
 using BTCPayServer.Payments.Lightning;
-using BTCPayServer.Security;
 using BTCPayServer.Services;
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.Store.cs
```diff
@@ -113,14 +114,18 @@ public override Task<IActionResult> CreateInvoice(string cryptoCode, CreateLight
...
-            var store = HttpContext.GetStoreData();
-            if (store == null || network == null)
+            if (network == null)
             {
                 throw ErrorCryptoCodeNotFound();
             }
-
+            
+            var store = HttpContext.GetStoreData();
+            if (store == null)
```
NicolasDorier, line 124: You should have a `return StoreNotFound();`

If it's not in the controller, use this.
```csharp
        private IActionResult StoreNotFound()
        {
            return this.CreateAPIError(404, "store-not-found", "The store was not found");
        }
```

Kukks (changes requested):
ditto what nico said

BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.Store.cs
dennisreimann: Done in [d4e5c00](https://github.com/btcpayserver/btcpayserver/pull/3658/commits/d4e5c00bb6e4488cfaf33dfe0f45e180003bb4d6).

Kukks (approved):

PR #3678, opened 2022-04-26
Fix 0 amount invoice UI

Kukks:


NicolasDorier:
Please do it without this massive copy/pasted blob that is untested

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -721,6 +780,10 @@ private string ExchangeRate(PaymentMethod paymentMethod)
         [HttpGet("invoice/status")]
         public async Task<IActionResult> GetStatus(string invoiceId, string? paymentMethodId = null, string? implicitPaymentMethodId = null, [FromQuery] string? lang = null)
         {
+            if (paymentMethodId == "null")
```
NicolasDorier, line 783: wtf?

fernandohre: Yep. I had the same feeling.
@Kukks instead of doing this, why don't you call _.HasValue_ method since it's nullable?

PR #3679, opened 2022-04-26
Exchange api no kraken

Kukks:
This PR is @woutersamaey's #3232  PR but without kraken support bundled in.


woutersamaey:
If this PR gets merged it will upset my Github track record, no? 😂

BTCPayServer.Abstractions/Custodians/Client/Exception/AssetBalancesUnavailableException.cs
```diff
@@ -0,0 +1,13 @@
+namespace BTCPayServer.Abstractions.Custodians.Client.Exception;
+
+public class AssetBalancesUnavailableException : CustodianApiException
+{
+    public System.Exception OriginalException { get; }
```
NicolasDorier, line 5: The Exception base class already has a `InnerException` for this.

BTCPayServer/Controllers/LnurlAuthService.cs
```diff
@@ -72,8 +75,9 @@ public async Task<bool> CompleteCreation(string name, string userId, ECDSASignat
                 CreationStore.Remove(userId, out _);
                 return true;
             }
-            catch (Exception)
+            catch (Exception e)
```
NicolasDorier, line 78: remove this debug line which has nothing to do with the PR

BTCPayServer/Controllers/GreenField/GreenfieldCustodianAccountController.cs
```diff
@@ -0,0 +1,485 @@
...
+        [HttpPut("~/api/v1/stores/{storeId}/custodian-accounts/{accountId}")]
+        [Authorize(Policy = Policies.CanManageCustodianAccounts,
+            AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        public async Task<IActionResult> UpdateCustodianAccount(string storeId, string accountId,
+            CreateCustodianAccountRequest request)
+        {
+            request ??= new CreateCustodianAccountRequest();
+
+            // TODO these couple of lines are used a lot. How do we DRY?
+            var custodianAccount = await _custodianAccountRepository.FindById(accountId);
```
NicolasDorier, line 186: `FindById` should include both accountId and storeId

BTCPayServer.Data/Migrations/20220115184620_AddCustodianAccountData.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    Id = table.Column<string>(maxLength: 50, nullable: false),
+                    StoreId = table.Column<string>(maxLength: 50, nullable: false),
+                    CustodianCode = table.Column<string>(maxLength: 50, nullable: false),
+                    Name = table.Column<string>(maxLength: 50, nullable: true),
+                    Blob = table.Column<byte[]>(nullable: true)
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_CustodianAccount", x => x.Id);
```
NicolasDorier, line 29: I think primary key should be `x.Id, x.StoreId`

BTCPayServer/Controllers/GreenField/GreenfieldCustodianAccountController.cs
woutersamaey: The account ID is always auto-generated server-side, so a user cannot do what you say. See `CreateCustodianAccountRequest`. There is no field for ID. The user supplies `Code` which would be `kraken` for example, a name and a config JObject.

woutersamaey: I think I'd prefer to add a check if the found account's store ID matches the storeId of the request. This way we can keep `FindById()` simple + still provide the security. Would that be OK?

BTCPayServer.Data/Migrations/20220115184620_AddCustodianAccountData.cs
woutersamaey: Why is this? Do you expect the same account ID to be in 2 stores? The account ID is always generated server-side. The user cannot set or change it, thus the combination of 2 fields as PK should not be needed.

BTCPayServer/Controllers/GreenField/GreenfieldCustodianAccountController.cs
woutersamaey: We should comment out the line `custodianAccount.StoreId = storeId;` so you cannot move the account to another store.
Or, if we were to allow this, we should check if the old + new store IDs are under the user's control. This is more complex. We could add this in a later iteration if we want.

woutersamaey:
@Kukks can I make changes on your branch or will you process @NicolasDorier feedback?

BTCPayServer.Data/Migrations/20220115184620_AddCustodianAccountData.cs
NicolasDorier: @woutersamaey the reason is that you are querying the custodian accounts of Stores in several places.
If your primary key is `Id,StoreId` then such request won't use the index of the primary key.
If your primary key is `StoreId,Id`, it will.

woutersamaey: Ok, so it is for performance reasons. Tnx.

BTCPayServer/Controllers/GreenField/GreenfieldCustodianAccountController.cs
```diff
@@ -0,0 +1,483 @@
...
+        public async Task<IActionResult> ViewCustodianAccount(string storeId, string accountId,
+            [FromQuery] bool assetBalances = false, CancellationToken cancellationToken = default)
+        {
+            var store = HttpContext.GetStoreData();
+            if (store == null)
+            {
+                return this.CreateAPIError(404, "store-not-found", "The store was not found");
+            }
+
+            var custodianAccountData = _custodianAccountRepository.FindById(accountId).Result;
```
NicolasDorier, line 99: At this point, there is no permission check about whether the accountId belongs to the store. As such, a malicious store can access the any store's account if he knows the accountId.

BTCPayServer.Data/Migrations/20220115184620_AddCustodianAccountData.cs
NicolasDorier: I kept primary key as `AccountId`. There is already an index on `StoreId`.

In the custodian repository I however require to specify both AccountId and StoreId, as there was a bug where a malicious user, just by knowing the AccountId could bypass permission checks.

woutersamaey:
@NicolasDorier @Kukks how do I proceed with my Kraken plugin?
Will it be broken after the recent changes?

PR #3682, opened 2022-04-28
Auto approve refunds

Kukks:
Since we can now auto-approve payouts as they are created, we can also make refunds almost fully automated beyond just creating the refund.

Since a refund can only be created by a store admin, there is no security dilemma. 

I am divided between making auto-approval for ALL refunds or only for those that payout payment method currency  ==  pull payment currency 

The main reason is that the user may be able to game it a bit and wait to submit the refund until the rates is in their favor.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -224,7 +224,7 @@ namespace BTCPayServer.Controllers
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
         public async Task<IActionResult> Refund(string invoiceId, RefundModel model, CancellationToken cancellationToken)
         {
-            using var ctx = _dbContextFactory.CreateContext();
+            await using var ctx = _dbContextFactory.CreateContext();
 
             var invoice = GetCurrentInvoice();
             if (invoice == null)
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -288,21 +288,25 @@ public async Task<IActionResult> Refund(string invoiceId, RefundModel model, Can
                         Name = $"Refund {invoice.Id}",
                         PaymentMethodIds = new[] { paymentMethodId },
                         StoreId = invoice.StoreId,
-                        BOLT11Expiration = store.GetStoreBlob().RefundBOLT11Expiration
+                        BOLT11Expiration = store.GetStoreBlob().RefundBOLT11Expiration,
+                        //AutoApproveClaims = true
```
Kukks, line 292: Alternatively, we could auto-approve all payouts

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -288,21 +288,25 @@ public async Task<IActionResult> Refund(string invoiceId, RefundModel model, Can
...
                     };
                     switch (model.SelectedRefundOption)
                     {
                         case "RateThen":
                             createPullPayment.Currency = paymentMethodId.CryptoCode;
                             createPullPayment.Amount = model.CryptoAmountThen;
+                            createPullPayment.AutoApproveClaims = true;
```
Kukks, line 299: Currently, we will auto-approve all payouts when there is no rate conversion invovled

dennisreimann (approved):
ACK for the concept of auto approving only for those payouts which payment method currency  ==  pull payment currency. As discussed on Mattermost.

PR #3689, opened 2022-04-29
Allow settings to be tied to a store

Kukks:
This PR allows Settings records to be directly tied to a store so that plugins and features can directly have data tied to a store (and that data is deleted on store deletion too).


dennisreimann:
For the dev this is basically a key/value store for (store-related) settings, right? 
If the storeId is omitted, this can be used for general settings too?

```
var value = await settingsRepo.GetSettingAsync<string>("key",storeId);
await settingsRepo.UpdateSetting("value","key", storeId);
```

I find it a bit weird that when saving, the key isn't the first argument. I see in the code that it is this way, because key and storeId are nullable — what would the use case for omitting the key be?

NicolasDorier:
@Kukks Can you use a different table `StoreSettings`?

Also don't attempt to cache it, cache was fine for server settings as the number of them is limited and they are often requested all over the place, but that's not the case for stores.

NicolasDorier:
Also, why Id and Name are different?

Kukks:
> Also, why Id and Name are different? What is the purpose of Name?

Id = Name+StoreId
Name = Name

Kukks:
> @Kukks Can you use a different table `StoreSettings`?
> 
> Also don't attempt to cache it, cache was fine for server settings as the number of them is limited and they are often requested all over the place, but that's not the case for stores.
> 
> The StoreRepository is more adapted to get the settings of a store.

But we don't have access to store repo in plugins, only ISettingsRepo

I can add a flag `cache = false`

NicolasDorier:
@Kukks I think it's fine to add a new interface and make StoreRepository implement it then.

> I can add a flag cache = false

I prefer you don't attempt to put a square peg in a round hole `SettingsRepository` was meant for server settings.
Additional table wouldn't need the Name hack either, just need a `StoreId`, `Name`, `Blob` with (StoreId, Name being primary keys)

NicolasDorier:
Closing in favor of #3851

PR #3696, opened 2022-04-30
Do not always provide counting in list views

NicolasDorier:
As I was analyzing our usage of the database on our mainnet instance, I discovered that the most costly db requests concern the `.CountAsync()` we are running against the database.

For example, for a store with 9050 invoices, it accounts for 700ms/800ms of the page processing time. (so ~90% of it)

It turns out that counting isn't a cheap operation for database, as it requires a full scan over all the rows. Indexes aren't helping the case.

So now, pagination's count is nullable. It can get populated when it is cheap, but won't be most of the time.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -836,37 +836,6 @@ namespace BTCPayServer.Tests
                 $"enddate:{time.AddSeconds(-1).ToString("yyyy-MM-dd HH:mm:ss")}");
         }
 
-        [Fact(Timeout = LongRunningTestTimeout)]
-        [Trait("Integration", "Integration")]
-        public async Task CanListNotifications()
-        {
-            using var tester = CreateServerTester();
-            await tester.StartAsync();
-            var acc = tester.NewAccount();
-            acc.GrantAccess(true);
-            acc.RegisterDerivationScheme("BTC");
-
-            const string newVersion = "1.0.4.4";
-            var ctrl = acc.GetController<UINotificationsController>();
-            var resp = await ctrl.Generate(newVersion);
-
-            var vm = Assert.IsType<Models.NotificationViewModels.IndexViewModel>(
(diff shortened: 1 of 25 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -836,37 +836,6 @@ public async Task CanListInvoices()
                 $"enddate:{time.AddSeconds(-1).ToString("yyyy-MM-dd HH:mm:ss")}");
         }
 
-        [Fact(Timeout = LongRunningTestTimeout)]
-        [Trait("Integration", "Integration")]
-        public async Task CanListNotifications()
```
NicolasDorier, line 841: This test is actually a copy of  `CanCheckForNewVersion`, but more superficial.

dennisreimann:
From a users perspective it might be valuable information how many items there are in total. Also if I read this correctly, now it might be possible to page to an empty page, right? 

Even though it makes it more complex, maybe we can support a more efficient count, like you did in #3694?

NicolasDorier:
@dennisreimann counting is very tricky thing in database. Notice that not even youtube or reddit count accurately.

This link show some hack one can do that work in a few circumstances:

https://www.cybertec-postgresql.com/en/postgresql-count-made-fast/

This link suggests some oteher solution, but require special postgres plugin and it only output an estimation of row, not perfect count.

https://www.citusdata.com/blog/2016/10/12/count-performance/

This is not simple thing.

The core of the problem is that if you want all the invoices of a user that has 9000 invoices. Fetching the page take 5ms, but counting 9000 invoices take 600 ms. If the user has 10x that amount, then it takes 6 sec to load the page.

ishristov:
Just an idea - would it be possible to count the invoices on invoice creation AND invoice filters change and then set the value somewhere (with keys totalInvoicesCount, filteredInvoicesCount) where it can be read more efficiently? That way there will never be an expensive count operation for a simple page view and at the same time there will be proper info for displays and pagination.

Issue #3697, opened 2022-05-01
Minor visual Bug in Payout Processor tab


Bangalisch:
Yo yo, 

Was going through the Payout processor and noticed a small visual bug or IMO it could use a touch up. 
It states : "IntervalMinutes" could use a space maybe or "Interval in minutes" ? 
Also, in our blog we state it's possible to  "pre-set time interval, for example, hourly, daily, weekly, and more."  This isn't really being made clear on the settings tab it self. Could be good to make this clear to the user in the interface :) 

![screenshot 38](https://user-images.githubusercontent.com/102448109/166131518-031eb679-8367-4dbc-b5be-abfedf567528.jpg)

dstrukt:
Agreed, was something I was planning to address in the next design call.

Also, was thinking we should combine all of the payment methods into one view.

Could add the minutes helper text to the input .. we have a similar component with the grey background, but this mock should suffice.

<img width="760" alt="Screen Shot 2022-05-01 at 12 47 06 AM" src="https://user-images.githubusercontent.com/6250771/166137056-d9bdaa4b-b2f8-45bb-9571-2923202a536e.png">

Bangalisch:
Looks better imo already.
The only thing in this view tho, you made both of the options mandatory * 
I might not want to set automated LN payments, maybe not make them * 
Do like the Minutes tekst inside the field.

dstrukt:
Oh good call, I overlooked that, just re-used a component, but will remove the mandatory label.

dstrukt:
Leaving this ticket open til the next design call, wanted to get some more insights into the direction of where this is headed.

PR #3698, opened 2022-05-01
Updated Payout processor Label for setting interval

Bangalisch:
Yo,

As i described in #3697  went ahead and looked it up. I think it should be something like this? 
Also changed the description a little, make it more formal instead of reading like a blog one liner twice ( Onchain and lightning stated the same text) 

Cheers.


NicolasDorier:
ping @Kukks

Kukks (approved):

PR #3703, opened 2022-05-02
Improve payout processors UI

### LINKED ISSUE #3697: Minor visual Bug in Payout Processor tab

dennisreimann:
Some quick fixes, closes #3697.

@dstrukt This doesn't integrate both options on one page and I think this might not even be desired: iirc @Kukks separated the configuration options to make the payout processors extendable by plugins. Let's revisit the concept once we gain more insights here.


dstrukt:
Ahhh I was going to ask in this week's design call why they were separated.

Sounds good, if it's done with the intention of future architectural groundwork, no qualms from me.

PR #3708, opened 2022-05-04
Update validation of crowdfund app settings

### LINKED ISSUE #3625: Small workflow/ visual bug in creating a Crowdfund app
Yo,

Just created a couple of Crowdfund apps. 
Was playing around with the reset at some point and found a small "bug". 

In the Topic Goal 
Set all fields except dates cuz it doesn't tell you to. 
Select reset goal to Day, still date doesn't tell you it's mandatory. 

Fill up the rest of the crowdfunding page. 
Save and now it tells you you forgot to put a start date.
Again where it earlier did not tell you it's mandatory. 
It might need a * just like the Description fields or display field once you set it to reset differently then Never.

See screen shots for example ; 
screenshot 7

screenshot 8

Cheers

bolatovumar:
1. Updated FE validation for crowdfund app settings
2. Fixed a few null reference exceptions which could be triggered by editing certain crowdfund app settings

Close #3625

### DIFF
--- a/BTCPayServer/wwwroot/main/site.js
+++ b/BTCPayServer/wwwroot/main/site.js
@@ -1,3 +1,5 @@
+const flatpickrInstances = [];
+
 document.addEventListener("DOMContentLoaded", function () {
     // sticky header
     const stickyHeader = document.querySelector('.sticky-header-setup + .sticky-header');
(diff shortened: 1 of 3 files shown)


BTCPayServer/wwwroot/main/site.js
```diff
@@ -1,3 +1,5 @@
+const flatpickrInstances = [];
```
bolatovumar, line 1: In order for us to act on flatpickr value changes we need access to the underlying instances so we store them in this array and access the values in the array like in the above file.

BTCPayServer/wwwroot/main/site.js
```diff
@@ -72,7 +74,10 @@ document.addEventListener("DOMContentLoaded", function () {
     }
 
     $(".input-group-clear").on("click", function () {
-        $(this).parents(".input-group").find("input").val(null);
+        const input = $(this).parents(".input-group").find("input");
+        const event = new CustomEvent('input-group-clear-input-value-cleared', { detail: input });
+        input.val(null);
+        document.dispatchEvent(event);
```
bolatovumar, line 80: This will trigger an event when user clears an input through the input clearing element so that we can react to it like I did in the above file.

Kukks (approved):

PR #3709, opened 2022-05-04
Support Lnurl Withdraw in pull payments

Kukks:
LNURL Withdraw is possible in pull payments with this PR, but only if the configurations are aligned with the requirements needed:
* The pull payment must be in BTC currency
* The pull payment must have lightning enabled as a method
* ~~Auto approve must be enabled~~ - Not anymore! If a payout needs to be approved, the flow will continue as usual.


Kukks:
Screenshots:
image
image

Issue #3266, opened 2022-01-07
Add concise 1-sentence description for features


pavlenex:
As discussed on UX channel on Mattermost, we have plenty of features that sometimes may not be entirely clear to people who use BTCPay. Adding a concise description of features may help with this.

![Screenshot 2022-01-07 at 12 21 05](https://user-images.githubusercontent.com/36959754/148537372-faff0e4c-aaf0-439b-8ee8-f04630ee420b.png)

_For example, explaining a payment request as a invoice which doesn't have expiration time, always provides current rate when a payer and is good for sharing a link with a clients._

This requires wider discussion, but I think, for starters we can improve the definitions of:
- Payment Request
- Invoice
- Pull Payments
- Payouts

dennisreimann:
Also relevant for #3255.

dstrukt:
Happy to take care of the PR when we have the wording, might take a shot this week while I browse the docs.

cc @Zaxounette

bolatovumar:
This is a great idea!

pavlenex:
I think we can expand the scope of it, for not just the pages mentioned, but also some sub-pages. For example I think `Payments` page in `Store Settings` deserve explanation. For example `Configure your store payment settings that global for all payment methods`. We'll probably need copywriter to help with these.

dstrukt:
@pavlenex Tend to agree it would be helpful to provide more context for the some of the sub-pages that are integral to the functionality and use of btcpay.

I've been thinking about this more as I step through all of these views, and the big question in my mind is not that we should explain the feature, definitely clear we should, but where?

1. For example, even having a 1-2 sentence explanation on the Payment Request / Payout / Pull Payment, etc.. list view would be incredibly obtrusive and become quite a nuisance after a short while if it's always "present".

We could alleviate this by treating the short explanation like an alert (could be styled slightly differently) that can be dismissed once the user has read or feels they understand.

2. Exploring the concept from no.1 further ... we have the (?) asset, which has historically linked directly out to the external documentation. We could leverage this to drop an accordion explaining the feature in a simple way, and providing a link to read more in the docs. More permanent and less obtrusive than no.1. Even further, we could even "expand" it on first-time users, and when they close the section, it's then always accessible.

For example. I used @BenGWeeks' copy, but obviously we could refine this, add more simple line-by-line (second item), etc.. bu this should give an idea of what's possible:
[...]

PR #3712, opened 2022-05-04
Rewording: Integrations become (store) plugins

### LINKED ISSUE #3594: `Integrations` should be `Store Plugins` and `Plugins` should be `Server Plugins`
Right now we're using quite a few different terms. `Integrations` within the store is basically a store plugin page.

`Store plugins` should show plugins server admin activated  and allow people to access the plugin (click on them), but shouldn't allow adding plugins or disabling them.

`Server plugins` is basically what we now call `plugins` which are available to admins only.

Both pages should allow people to access plugins and open them. Right now we have plugins like whitepaper or hairstylist which once activated nothing happens, because a user isn't aware where they can access them, so allowing easy way to access activated plugin should be had in mind when re-making these pages.

I've discussed this with @Kukks and it seems we'd want to do it for 1.5.0 if we can, otherwise for 1.6.0

dennisreimann:
We're keeping the `store-integrations-list` extension point name for backwards compatibility, otherwise I renamed all "integration" occurences in the views and also the routes.

Note: We still refer to external software like WooCommerce, Drupal, Magento and such as "integrations", I think that distinction makes sense.

Closes #3594.


pavlenex:
Would it make sense to somehow emphasize these are `Store plugins` vs `Plugins` in the server's side? 
At least maybe in the heading ,but not in the menu? Or add a sentence explaining it through  #3266  later on?

Screenshot 2022-05-10 at 14 08 05

pavlenex:
You're on point @dennisreimann I was actually thinking about the same as soon as I hit that green comment button. Let's keep it unified and see if it causes confusions to admins, though obviously those people aren't just regular users and are a bit tech savvier.

BTCPayServer/Views/UIStores/Plugins.cshtml
```diff
@@ -1,7 +1,7 @@
-@model IntegrationsViewModel
+@model PluginsViewModel
 @{
     Layout = "../Shared/_NavLayout.cshtml";
-    ViewData.SetActivePage(StoreNavPages.Integrations, "Integrations", Context.GetStoreData().Id);
+    ViewData.SetActivePage(StoreNavPages.Plugins, "Plugins", Context.GetStoreData().Id);
```
Kukks, line 4: I'm worried this change may break plugins.

dennisreimann: Can we add a deprecation warning and handle it backwards-compatible?

dennisreimann:
@Kukks see [eea52fd](https://github.com/btcpayserver/btcpayserver/pull/3712/commits/eea52fdf68cb44ba3fad1acf6573cd17614ff157).

Kukks (approved):

PR #3713, opened 2022-05-05
A few tables in the Cart checkout are having bad widths and margins

ishristov:
There's a loose css that matches a lot of tables unnecessary and was causing some issues in the Cart checkout.

The css is this:
```
*:not([class^="table-responsive"]) > .table {
  width: calc(100% + 1rem);
  margin: 1.5rem -0.5rem;
}
```

I couldn't remove it cause it's probably there for a reason so that's why I explicitly fixed only the affected tables by adding some additional predefined classes.

Before:
No left/right margin.
cart-before

After:
Default margin.
cart-after

Before:
confirmation-before

After:
confirmation-after

### DIFF
--- a/BTCPayServer/Views/UIAppsPublic/PointOfSale/Cart.cshtml
+++ b/BTCPayServer/Views/UIAppsPublic/PointOfSale/Cart.cshtml
@@ -119,7 +119,7 @@
                 </button>
             </div>
             <div class="modal-body p-0">
-                <table id="js-cart-summary" class="table w-100 m-0">
+                <table id="js-cart-summary" class="table m-0">
                     <tbody class="my-3">
                         <tr>
                             <td colspan="2" class="border-top-0 h5">Summary</td>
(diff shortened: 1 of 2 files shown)


Kukks (approved):
ack from my end

dstrukt:
Improving the UI for this view is currently something I'm working on, but definitely ACK a stop-gap patch of the current view until we can properly address those improvements.

dennisreimann:
My bad, recent changes I made to the table styles didn't account for that - thanks for bringing it up!

Can you change this part in bootstrap.css so that the tables styles get applied only for tables within `main`?

```css
/* Pull tables out of the grid by their horizontal column padding */
main *:not([class^="table-responsive"]) > .table {
  width: calc(100% + 1rem);
  margin: 1.5rem -0.5rem;
}

main [class^="table-responsive"] {
  width: calc(100% + 1rem);
  margin: 1.5rem -0.5rem;
}
```

The other changes can be reverted then. I've already applied [these changes in the design repo](https://github.com/btcpayserver/btcpayserver-design/commit/374db511582dcca80b1d45d0fca784cdd7aadd67).

Thanks again!

ishristov:
@dennisreimann, I've removed my changes and fixed the css in the `bootstrap.css` per your suggestion.

dennisreimann (approved):

PR #3717, opened 2022-05-05
Removes `Invoice Summary` Section on Invoice List View

dstrukt:
As discussed in Design Call 32, but adding for discussion on Design Call 33.

- Removes `<tr>` section
- Removes accordion carrot CTA

Updated View:
Screen Shot 2022-05-05 at 6 45 37 AM


Kukks:
While I appreciate that this is attempting to make everything more minimal and clean, it does make it more annoying to quickly glance at key info of invoices

dstrukt:
Yeah I can see both sides of the argument, wanted to get the general sentiment, as I imagined this would be a more controversial PR .. however, if the Summary is kept in this view, definitely think an improved down carot icon with a larger hit area would be a huge improvement.

dstrukt:
Going to revisit this and re-open in the future.

PR #3718, opened 2022-05-05
Fix shopify id / order no

### LINKED ISSUE #3636: Display Shopify Order ID in the Invoices page
**Description**
To make invoices reconcilable with Shopify orders, it looks like you're currently using an internal ID (that is also used in the Url of the order in Shopify).

image

**Steps to reproduce the bug**

1. Make an order in Shopify that has the integration with BTCPay Server
2. Open the Invoices screen in BTCPay Server
3. Notice that the Order numbers are of the form "shopify-4688949215318"
4. It would be preferential if the order was "shopify-4688949215318 (# 1030)" (where # 1030 is the Order number in Shopify and # 1030 links to the Order in Shopify, such as https://someshop.myshopify.com/admin/orders/4688949215318)

**Expected behavior**

The Order number from Shopify is reconcilable to the Order in Shopify.

**BTCPay Environment:**
- BTCPay Server Version: v1.4.9.0
- Deployment Method: Umbrel
- Browser: Edge

Kukks:
fixes #3636

### DIFF
--- a/BTCPayServer/Plugins/Shopify/UIShopifyController.cs
+++ b/BTCPayServer/Plugins/Shopify/UIShopifyController.cs
@@ -112,7 +112,7 @@ namespace BTCPayServer.Plugins.Shopify
             var invoiceOrderId = $"{ShopifyOrderMarkerHostedService.SHOPIFY_ORDER_ID_PREFIX}{orderId}";
             var matchedExistingInvoices = await _invoiceRepository.GetInvoices(new InvoiceQuery()
             {
-                OrderId = new[] { invoiceOrderId },
+                TextSearch = invoiceOrderId,
                 StoreId = new[] { storeId }
             });
             matchedExistingInvoices = matchedExistingInvoices.Where(entity =>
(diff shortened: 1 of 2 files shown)


BTCPayServer/Plugins/Shopify/UIShopifyController.cs
```diff
@@ -112,7 +112,7 @@ public async Task<IActionResult> ShopifyJavascript(string storeId)
             var invoiceOrderId = $"{ShopifyOrderMarkerHostedService.SHOPIFY_ORDER_ID_PREFIX}{orderId}";
             var matchedExistingInvoices = await _invoiceRepository.GetInvoices(new InvoiceQuery()
             {
-                OrderId = new[] { invoiceOrderId },
+                TextSearch = invoiceOrderId,
```
NicolasDorier, line 115: `TextSearch` should be used only for users, not for logic, as it is possible you match the wrong invoices here.

Kukks: `invoiceOrderId` should be fairly unique though.

PR #3719, opened 2022-05-05
Minor update to Lightning Address view

dstrukt:
Was debating if it made sense to break the three "settings" types into their own columns with the same conditional to show in the row if there is a value submitted, but opted not to for this first iteration, but if there's a strong vote to do so, will revisit to implement.

Current:
Screen Shot 2022-05-05 at 9 44 25 AM

Update:
Screen Shot 2022-05-05 at 9 43 54 AM

### DIFF
--- a/BTCPayServer/Views/UILNURL/EditLightningAddress.cshtml
+++ b/BTCPayServer/Views/UILNURL/EditLightningAddress.cshtml
@@ -114,8 +114,8 @@
                         <tr>
                             <td>
                                 <div class="input-group" data-clipboard="@address">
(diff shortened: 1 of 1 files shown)


pavlenex:
The current option seems more visually appealing to me. What are we trying to solve with this one?

dstrukt:
oo no, not switched, just simplifying the view. 

i prefer the more simplified one, but if there's no love for it, happy to close :)

pavlenex:
Personally not a fan of copy buttons not being aligned properly.

dstrukt:
Will revisit in the future.

PR #3721, opened 2022-05-05
Redirect to payment method setup

dennisreimann:
As discussed in btcpayserver/organization#87: Solves the progressive disclosure approach (brought up by @dstrukt in #3651) by simply redirecting to the right view in case no payment method is set up, yet.

This lands you on the dashboard, where in this case the setup guide is presented.

grafik

Closes #3651.


dstrukt:
Nice, this is much better / clearer .. glad it was a simple enough solution as well. 

Will continue to use and see if it warrants any additional showing/hiding in the future, but afaict, this seems perfectly suitable!

dstrukt (approved):

PR #3725, opened 2022-05-06
Lightning Address: Optional alternative hostname

### LINKED ISSUE #3722: Lightning address with http 301 magic
Here's what I'd like to do:

Run a BTC Pay server on `pay.example.com`, but use `me@example.com` as the lightning address. Why? Because it looks nicer than `me@pay.example.com`.

I was thinking this should be a matter of adding an http 301 redirect for `example.com/.wellknown/whatever` to `pay.example.com`. However in the Lightning Address plugin, I would need the option to override the domain name.

(I could also set up the BTC Pay server at `example.com/pay`, but a subdomain seems a bit more practical for every use case except a vanity lightning address)

dennisreimann:
Closes #3722.

### DIFF
--- a/BTCPayServer.Data/Data/LightingAddressData.cs
+++ b/BTCPayServer.Data/Data/LightingAddressData.cs
@@ -26,6 +26,7 @@ public class LightningAddressData
 public class LightningAddressDataBlob
 {
     public string CurrencyCode { get; set; }
+    public string AlternativeHostname { get; set; }
     public decimal? Min { get; set; }
     public decimal? Max { get; set; }
 }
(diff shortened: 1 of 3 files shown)


Kukks:
> We could restrict it to the set of alternative domains, that are configured for BTCPay. But I think for this use case this isn't really intuitive: one's not really using the main domain with BTCPay, just redirecting the lnurlp requests to it.
> 
> tbh I'm not sure about what this entails in terms of security, I just wanted to bring this up for discussion and am open to extend it – you know better about the LNURL specs than I do.

I think it will be too hard for 99% of people to do this. This PR does not add the functionality required for it: it is already supported. The issue is that users must do reverse proxy magic

dennisreimann:
@Sjors mentioned in #3722 that it didn't work for him with a simple rewrite and that the Lightning Address in the meta data needs to reflect the host as well. I couldn't test it yet, but will try to dig deeper now that you say it is supposed to work already.

dennisreimann:
I can confirm it works with a simple redirect, the Lightning Address code doesn't need a change.

Issue #3645, opened 2022-04-16
Send a `verification email` when a user changes their email to avoid lock-out.


Eskyee:
a already User is Locked out of their BTCPay Account. 
They Have Changed their email address, and said they never received email confirmation, 
systems locked me out with: "Invalid login attempt." and "You must have a confirmed email to log in.". 

A verify email was sent from our server, which has now timed out.

So now the Admin has to verify them again, but in the users section, web view  their is only edit or remove.
their is nothing to say, send verify email again, to user, or no toggle button to verify them again.

So it looks like its via the command terminal is the only way to verify a user again, as their is only,   is admin - function.
Would be nice to have this in the front browser,  with a verify a user toggle or send verify email again.
also not had this problem in years, so I have forgotten the commands to do this.

Make the email address associated for user verified. terminal commands ??
I have search the BTCPay docs, but also couldn't find no information here. About verify user. for admins.

Any ideas how to do this manually, via terminal.

cheers

pavlenex:
From what I understood, previously registered user, changed their email address, but never got email confirmation when changing it, hence they cannot sign in. This would mean that anyone changing an email on their instance would be locked out. Could anyone try to replicate the issue? @bolatovumar ?

ishristov:
@pavlenex, I tried to replicate:
- I changed my email and clicked "save" (didn't click the "send verification email"). I didn't receive any email, but email was changed and I could login with the new email;
- I repeated the same and changed to my old email, it worked again.

If I login with the server admin, I can see that even though the email is changed, the user is still marked as "verified". So the email change doesn't make them "unverified".

pavlenex:
Thanks for veryfing @ishristov then this is something we should address.

Eskyee:
@pavlenex  any idea of the commands to do this via ssh " to manually change their email and resend password "

cheers!

pavlenex:
I am not aware of command to change users email and it seems it's not possible to change their email or re-send the verification email. Perhaps this is a functionality we should add. Maybe @bolatovumar has an idea on the approach and perhaps @Kukks know a command that can be executed to change user's email in the database?

bolatovumar:
@pavlenex is the thought here to allow server admins to update existing user's email address to something else or allow them to resend the verification email? (or both?) I can look into this (although I can't promise anything as I don't really know anything about the email verification functionality).

pavlenex:
@bolatovumar I'd say both would be fine, how do you feel about it?

[...]

PR #3726, opened 2022-05-08
Allow resending verification email for users

bolatovumar:
This PR adds ability to resend a verification email to those users who may have lost the original email or it has timed out as described in #3645.

This partially addresses the issue described in #3645. In a follow-up PR I will add an ability for admins to change users email address in case they lost access to the original one they signed up with.

### DIFF
--- a/BTCPayServer/Views/UIServer/ListUsers.cshtml
+++ b/BTCPayServer/Views/UIServer/ListUsers.cshtml
@@ -95,6 +95,9 @@
                     }
                 </td>
                 <td class="text-end">
+                    @if (user.RequiresEmailConfirmation && !user.Disabled) {
+                        <a asp-action="SendVerificationEmail" asp-route-userId="@user.Id">Resend verification email</a> <span>-</span>
+                    }
                     <a asp-action="User" asp-route-userId="@user.Id">Edit</a> <span> - </span> <a asp-action="DeleteUser" asp-route-userId="@user.Id">Remove</a>
                     - <a asp-action="ToggleUser"
                          asp-route-enable="@user.Disabled"
--- a/BTCPayServer/Views/UIServer/ListUsers.cshtml
+++ b/BTCPayServer/Views/UIServer/ListUsers.cshtml
@@ -96,7 +96,8 @@
                 </td>
                 <td class="text-end">
                     @if (!user.Verified && !user.Disabled) {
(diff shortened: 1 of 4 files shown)


pavlenex (commented):
Thanks for addresing this one. I've see that re-send should be triggered on the user's page, but I think I may be testing this incorrectly (bear in mind I've set up both store and server smtp locally), but I still don't see a way to re-send a confirmation to the user, I assume re-send is allowed only if they initially authenticated through the link?

Meaning, if I as an admin, had this disabled, then enabled it after a while, only user registring afterwards and getting email link will be able to have their verification email re-sent?

pavlenex:
Okay, this works as expected and actually it will probably solve the problem that I've seen many first-time users face. They set up SMTP incorrectly and then require email verification.

This PR, fixes this, by allowing admins to re-send emails for users that registered when email verification was required.

Screenshot 2022-05-10 at 13 32 24


concept ACK, and tACK needs someone to make sure there aren't edge cases in the code.

bolatovumar:
> to make sure there aren't edge cases in the code.

It's using exactly the same code as we use to send the verification email when user first signs up so I think it's pretty safe. Also, this link will only be available for users who signed up when the server policy required email verification.

BTCPayServer/Views/UIServer/ListUsers.cshtml
```diff
@@ -95,6 +95,9 @@
                     }
                 </td>
                 <td class="text-end">
+                    @if (user.RequiresEmailConfirmation && !user.Disabled) {
```
Kukks, line 98: You should use `user.Verified` not `RequiresEmailConfirmation ` as the `RequiresEmailConfirmation ` is only a flag to show that this requirement is present for the user and not that he is waiting for verification.

(IE. You require verification for users. A user registers and `RequiresEmailConfirmation` is set to true for User. You turn off verification for users. The user still needs to verify email as he was created before you changed the setting)

Kukks (changes requested):

BTCPayServer/Views/UIServer/ListUsers.cshtml
bolatovumar: @Kukks sure, I can do that. I didn't use this flag initially because I wanted to allow the server admin to resend the email anyway, even if the user is verified already as long as the user was created when the policy required email confirmation.

bolatovumar: @Kukks done here: afed3577794023d59d60425425fc4f5518b83088

PR #3731, opened 2022-05-09
Improve Refund Flow

### LINKED ISSUE #3715: Improve Refund Flow
Starting this issue to track the improvements for the Refund flow, not 100% ready to be worked on just yet, but should be after this week's design call.

Feel free to add things that could also be addressed in the comments or link any other issues pertaining to this.

[TODO Design Mocks for all points]

- [ ] Options should be in modals that overlay the Invoice Detail view
Current:
Screen Shot 2022-05-05 at 4 19 49 AM
Screen Shot 2022-05-05 at 4 20 00 AM
Screen Shot 2022-05-05 at 4 21 27 AM
Expected:
Screen Shot 2022-05-05 at 5 09 07 AM

- [ ] All options (Payment Method, Amount, etc...) should be on a single modal view (TBD)

- [ ] Selecting Custom Amount should expand an accordion on the same view

- [ ] Should alert the user if they're going over pay the refund amount when selecting the custom amount(as reported by @Bangalisch)

- [ ] An option to also offer a refund LESS the costs involved in issuing the refund (for example onchain fee to send it out), as suggested by @Kukks 

- [ ] Should the Currency be the new updated currency selector to prevent having to type the currency (TBD)
Screen Shot 2022-05-05 at 4 21 27 AM

dennisreimann:
- [x] Options should be in modals that overlay the Invoice Detail view
- [x]  All options (Payment Method, Amount, etc...) should be on a single modal view
- [x] Selecting Custom Amount should expand an accordion on the same view
- [x]  Should the Currency be the new updated currency selector to prevent having to type the currency (TBD)
- [ ]  Should alert the user if they're going over pay the refund amount when selecting the custom amount
- [ ]  An option to also offer a refund LESS the costs involved in issuing the refund (for example onchain fee to send it out)

Closes #3715.


dstrukt:
Let's go!

Should have an update for the crowdfund PR as well, as usual, killing it!

dennisreimann:
As the current state is already an improvement and the remaining two points require more conceptual work and code, we could do that in a separate PR. Opening it up for review.

dstrukt:
Nice, will review asap.

pavlenex (approved):
tACK.

There's quite a lot of cleanup we can do wrt to the refund flow. For example our options still could be way simpler, for example it can be full vs partial refund then we'd guide users towards currency, etc, but @dennisreimann suggested we do the remaining two points in a separate PR.

dstrukt (approved):

dstrukt:
Agree @pavlenex, for sure, we'll tighten things up in this next pass .. lot of easy wins with this PR already, great work @dennisreimann!

Kukks (approved):

PR #3734, opened 2022-05-11
Support LUD18 for LN Address

Kukks:


pavlenex:
I don't know what this does, any explanation?

Kukks:
> I don't know what this does, any explanation?

Initially, I thought this was one of the holy grail precursor to one of the most requested feature: Requesting data from the customer during checkout.

However @fiatjaf pointed out that lud18 (https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md) is only meant to identify the sender of the payment to the receiver and not provide custom information. The sending app actually fills in this data in the background automatically and there is no manual entry of data.

For example, it can provide a name, email, a pubkey (could be any public key, not the lightning node), a lightning address, or even proof that a specific user sent it via LNURL-AUTH.

Kukks:
image
image

fiatjaf:
Don't blame me. I do want services to be able to request arbitrary information from wallets at checkout time, but I wasn't capable of convincing wallet makers yet.

This is a good start, though.

pavlenex:
> Requesting data from the customer during checkout.

Yeah, we need to implement this particular thing on a much larger scale, for all payment methods not just LNURL.

Kukks:
> > Requesting data from the customer during checkout.
> 
> Yeah, we need to implement this particular thing on a much larger scale, for all payment methods not just LNURL.

Yes, I talked with @woutersamaey  on a potential option for it. Either way, this has become a small opt-in addition the ln address functionality

Tastry (changes requested):
Ok

Kukks:
No interest, closing

PR #3742, opened 2022-05-13
Improve invoice list performance

NicolasDorier:
So after my last improve removing the paging count, I was eager to see if the speed of my invoice list with 9000 invoices was faster.

To my surprise, this was still slow.

Here is the query plan:

image

503ms for getting the 50 first invoice is really bad.
I found out that the query plan was performing poorly because we are ordering the invoices by date.
Because we don't have any index on the creation date of invoices, this mean there is a full table scan. (nested loop over 9000 rows)

I added an index on the created column and compared the same request:

image

This turn the 500ms query into 10ms!

Now my invoice list show up instantly :)


NicolasDorier:
Maybe merging this after https://github.com/btcpayserver/btcpayserver/pull/3679 to prevent conflicts on db changes

PR #3743, opened 2022-05-13
Re-add S3 and Google storage options and bump libs

Kukks:


dennisreimann:
lgtm. why were they removed?

NicolasDorier:
Nobody is missing it, and this isn't well tested. I think we shouldn't integrate to providers.

fernandohre (approved):

Kukks:
@NicolasDorier How about we only bump the libs then?

NicolasDorier:
I think we should go the plugin route indeed @bolatovumar

PR #3744, opened 2022-05-13
Wallet transactions export

dennisreimann:
Adds an export (CSV and JSON) to the wallet transactions, similar to the invoices export.

The exported data needs some more work, but in general this is mostly done. Feedback welcome!

Wallet transactions export

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.cs
@@ -1289,16 +1289,14 @@ namespace BTCPayServer.Controllers
                 return NotFound();
             
             var wallet = _walletProvider.GetWallet(paymentMethod.Network);
-            var walletBlobAsync = WalletRepository.GetWalletInfo(walletId);
             var walletTransactionsInfoAsync = WalletRepository.GetWalletTransactionsInfo(walletId);
             var transactions = await wallet.FetchTransactions(paymentMethod.AccountDerivation);
-            var walletBlob = await walletBlobAsync;
             var walletTransactionsInfo = await walletTransactionsInfoAsync;
             var input = transactions.UnconfirmedTransactions.Transactions
                 .Concat(transactions.ConfirmedTransactions.Transactions)
                 .OrderByDescending(t => t.Timestamp)
                 .ToList();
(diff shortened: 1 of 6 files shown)


dstrukt:
Nice, great first steps toward feature and functionality pairing!

Enables us to address the next set of updates across all views equally.

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -1278,6 +1279,39 @@ await ExplorerClientProvider.GetExplorerClient(walletId.CryptoCode)
             }
         }
 
+        [HttpGet("{walletId}/export")]
+        [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanViewInvoices)]
```
NicolasDorier, line 1283: I don't think CanViewInvoices should have this right, we should keep it to `Policies.CanModifyStoreSettings` (already required at controller level)

NicolasDorier, line 1283: I don't think CanViewInvoices should have this right, we should keep it to `Policies.CanModifyStoreSettings` (already required at controller level)

NicolasDorier:
This need a small selenium test that click on json/csv. Can modify an existing test.

NicolasDorier:
This need a small selenium test that click on json/csv. Can modify an existing test.

BTCPayServer/Controllers/UIWalletsController.cs
dennisreimann: Fixed in [32d6677](https://github.com/btcpayserver/btcpayserver/pull/3744/commits/32d66771c09cd219309654cd2965911610d9e3b2).

dennisreimann:
Test cases added and the export format is finalized now. Ready for review :)

NicolasDorier:
@dennisreimann I understand selenium discourage download. However, you still need to exercise the code paths.
You can user the tester and instantiate the controller directly and call it if your prefer.

dennisreimann:
@NicolasDorier Updated it, not sure if this is what you wanted though. Feel free to change it further :)

dennisreimann:
Ok, it looks like the CSV download check doesn't work in CI anyways, removed it.

Issue #3661, opened 2022-04-21
Add `store's default currency` to the dashboard wallet balance


pavlenex:
User should be able to toggle between a store default currency and a BTC balance on the dashboard.

Not sure how we could make `toggle` action obvious, or should it be side by side. I think most wallets have toggle, so should be fine as long as it's clear a user can click. The conversion logic should already exist in the `Wallet` > `Send` tab where a BTC amount is converted to store's default currency.

![Screenshot 2022-04-21 at 12 06 53](https://user-images.githubusercontent.com/36959754/164432228-56f231bc-19a0-4cca-9e1d-ac5a29881a6e.png)

This should address most concerns in https://github.com/btcpayserver/btcpayserver/discussions/3637

![Screenshot 2022-04-21 at 12 10 34](https://user-images.githubusercontent.com/36959754/164432892-af1f17a7-b487-4feb-a601-c2b4c9890226.png)

ishristov:
Is the "wallet balance" on the dashboard a combination of the on-chain balance AND the lightning wallet that's configured for the store?

IMO if you are managing a store, the more important info that you care about is the received amounts for a particular period (in btc and fiat) and not the wallet balance. The wallet balance graph (balance for different days) is not very useful and shouldn't be the center of the dashboard. The received payments (settled invoices) should be.

With that being said, issues in #3637 are something different and not really addressed by this.

pavlenex:
You can filter periods in dashboard. 

> Is the "wallet balance" on the dashboard a combination of the on-chain balance AND the lightning wallet that's configured for the store?

No, on-chain only.

ishristov:
I understand that LN is different. But I'm talking from the user's perspective. They (I'm also running a store for selling stuff so I'm also on that side) don't care about a "wallet balance" if they also receive LN payments. What they MOSTLY care about when they login in btcpay is to see what PAYMENTS have been received, not the balance 1 week ago (the current graph in the new dashboard). The important thing is how much value in FIAT have been received, because usually there's a legit business running the store and they want to tax and account these payments. Which is done by checking the total FIAT representation of the paid invoices.

I know I (and they) can make almost everything in excel/google sheets but is that what you really want for the btcpay users? To export files on their mobile phone, then importing it in google sheets and then making a pivot table just to see a sum of the paid invoices in USD? :)

[...]

PR #3752, opened 2022-05-16
Added toggle button to switch to store default currency

SakshamSolanki:
Added a toggle switch on the dashboard to show store's wallet balance in its default currency #3661


SakshamSolanki:
I have added you as a collaborator on my fork

dennisreimann:
Thanks, I pushed my changes.

I think we should use our own store exchange rates instead of querying CoinGecko. For that to work we might need to add an API or edit the existing rates endpoint, but we can do so in a separate PR.

gruve-p:
Does this PR intend to fix: https://github.com/btcpayserver/btcpayserver/pull/3654#issuecomment-1103801393 ?

dennisreimann:
Updated the script and display:

grafik

This is now using our own rates API endpoint.

I think we can move this out of the draft status now.

dennisreimann:
@SakshamSolanki Can you open up this PR for review?

dennisreimann:
Updated according to the comments in yesterdays design meeting. 

Also displaying the labels on the Y-axis in the currency now when switched to the default currency. Found and fixed a bug while implementing this too :)

dashboard-currency

dennisreimann:
@pavlenex @dstrukt I think this is ready now, do you agree?

dstrukt:
I was happy with what I saw in the design call, so I think so as well .. was planning to take one more look when I do my next set of reviews later today, but feel free to proceed before that as well.

dennisreimann:
@pavlenex I've now fixed it, so that it uses the right decimals for USD regardless of the locale. We could do the same for EUR and other currencies.

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
             <label class="btn btn-link" for="filter-year">1Y</label>
         </div>
     </header>
-    <div class="ct-chart ct-major-eleventh"></div>
+    <div class="ct-chart"></div>
     <script>
         (function () {
+            const defaultCurrency = @Safe.Json(storeBlob.DefaultCurrency);
+            const cryptoBalance = document.getElementById("balance-@Model.CryptoCode");
```
NicolasDorier, line 55: @Safe.Json

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
...
+            const defaultBalance = document.getElementById("balance-@storeBlob.DefaultCurrency");
```
NicolasDorier, line 56: @Safe.Json

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
...
+            let rate = null;
+            let data = { series: @Safe.Json(Model.Series), labels: @Safe.Json(Model.Labels), balance: @Safe.Json(Model.Balance) };
+            
+            document.getElementById("balance-@storeBlob.DefaultCurrency").style.display = 'none';
```
NicolasDorier, line 60: @Safe.Json

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
...
                 delegate('change', `#${id} [name="filter"]`, async e => {
                     const type = e.target.value;
                     await update(type);
                 })
-            })
+                delegate('change', '#DefaultCurrencyToggle input', async e => {
+                    const { target } = e;
+                    if (target.value === defaultCurrency) {
+                        const currencyPair = @Safe.Json($"{Model.CryptoCode}_{storeBlob.DefaultCurrency}");
+                        const response = await fetch(`@Context.Request.PathBase/api/rates?storeId=@Model.Store.Id&currencyPairs=${currencyPair}`);
```
NicolasDorier, line 125: @Safe.Json

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
...
+                    render(data);
                 }
             };
-            render({ series: @Safe.Json(Model.Series), labels: @Safe.Json(Model.Labels), balance: @Safe.Json(Model.Balance) });
+            const toDefaultCurrency = (value, rate) => {
+                return Math.round((value * rate) * 100) / 100;
+            }
+            const displayDefaultCurrency = (value, currency) => {
+                const locale = currency === "USD" ? 'en-US' : navigator.language;
+                const opts = { currency, style: 'decimal', minimumFractionDigits: 2 };
```
NicolasDorier, line 112: do not hardcode the fraction of digit, in the Model just add a property called `Divisibility`.
You can get it by checking how [ViewCrowdfundViewModel](https://github.com/btcpayserver/btcpayserver/blob/5ac4135a13784fe8cc4aa535378d9038fb7e6940/BTCPayServer/Models/AppViewModels/ViewCrowdfundViewModel.cs#L42) did it.

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -20,50 +48,98 @@
...
-                    showArea: true
-                });
+                }, renderOpts);
                 // prevent y-axis labels from getting cut off
                 window.setTimeout(() => {
                     const yLabels = [...document.querySelectorAll('.ct-label.ct-vertical.ct-start')];
                     if (yLabels) {
-                        const width = Math.max(...(yLabels.map(l => l.innerText.length * 8)))
-                        chart.update(null, { axisY: { offset: width } })
+                        const factor = rate ? 6 : 8;
```
NicolasDorier, line 90: how is `6` and `8` calculated? is it dependent on the number of digit? if so check https://github.com/btcpayserver/btcpayserver/pull/3752/files#r897476851

dstrukt (approved):
Realizing I never properly signed off.

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
dennisreimann: No, these are kind of magic numbers for the rendering. tbh I haven't figured out yet why we need them in the first place, but for now they work and I might be able to fix this properly when I get back to it.

PR #3757, opened 2022-05-18
Upgrade Lightning lib; add and pass cancellation tokens in Greenfield

dennisreimann:
Based on btcpayserver/BTCPayServer.Lightning#75


NicolasDorier:
every test fails

dennisreimann:
Whatever went wrong, now we're good and green!

Kukks (approved):

PR #3759, opened 2022-05-19
Fix a couple of mobile display issues

### LINKED ISSUE #3699: Improve `store settings` view alignment on mobile [iPhone 8]
Just noticed that the Store Settings view isn't well aligned on iPhone 8 for some reason, would be nice if we could fix this.

https://user-images.githubusercontent.com/36959754/166134812-b7b106da-5614-48c9-bfc0-b7c3f64a3a3b.MOV

bolatovumar:
|Before|After|
|---|---|
|Capture|Capture2|

Close #3699

### DIFF
--- a/BTCPayServer/Views/UIServer/Services.cshtml
+++ b/BTCPayServer/Views/UIServer/Services.cshtml
@@ -117,7 +117,7 @@
             {
                 <tr>
                     <td>@s.Name</td>
-                    <td style="word-wrap: anywhere;">
+                    <td style="word-break: break-all;">
                         <code>@s.Link</code>
                     </td>
                 </tr>
--- a/BTCPayServer/Views/UIStores/GeneralSettings.cshtml
+++ b/BTCPayServer/Views/UIStores/GeneralSettings.cshtml
@@ -45,11 +45,14 @@
                 <a href="https://docs.btcpayserver.org/FAQ/Stores/#add-network-fee-to-invoice-vary-with-mining-fees" target="_blank" rel="noreferrer noopener">
                     <span class="fa fa-question-circle-o text-secondary" title="More information..."></span>
                 </a>
-                <select asp-for="NetworkFeeMode" class="form-select w-auto">
-                    <option value="MultiplePaymentsOnly">... only if the customer makes more than one payment for the invoice</option>
-                    <option value="Always">... on every payment</option>
-                    <option value="Never">Never add network fee</option>
-                </select>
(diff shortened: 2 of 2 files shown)


BTCPayServer/Views/UIStores/GeneralSettings.cshtml
```diff
@@ -45,11 +45,14 @@
                 <a href="https://docs.btcpayserver.org/FAQ/Stores/#add-network-fee-to-invoice-vary-with-mining-fees" target="_blank" rel="noreferrer noopener">
                     <span class="fa fa-question-circle-o text-secondary" title="More information..."></span>
                 </a>
-                <select asp-for="NetworkFeeMode" class="form-select w-auto">
-                    <option value="MultiplePaymentsOnly">... only if the customer makes more than one payment for the invoice</option>
-                    <option value="Always">... on every payment</option>
-                    <option value="Never">Never add network fee</option>
-                </select>
+                @* For whatever reason wrapping the select with this div fixes this Safari bug: https://github.com/btcpayserver/btcpayserver/issues/3699 *@
```
bolatovumar, line 48: I don't understand why this fixes the issue but it does. I was able to repro the bug in Safari only, this issue doesn't reproduce in Chrome. Also, it's intermittent in Safari but once it happens I can reproduce it consistently but sometimes it stops happening all of a sudden. I think it's related to changing zoom levels.

BTCPayServer/Views/UIServer/Services.cshtml
```diff
@@ -117,7 +117,7 @@
             {
                 <tr>
                     <td>@s.Name</td>
-                    <td style="word-wrap: anywhere;">
```
bolatovumar, line 120: `word-wrap: anywhere;` doesn't actually do anything. `anywhere` is not a valid value: https://www.w3schools.com/cssref/css3_pr_word-wrap.asp

NicolasDorier: ping @dennisreimann

dennisreimann: The `text-break` class should work here ([details(https://getbootstrap.com/docs/5.2/utilities/text/#word-break)).

BTCPayServer/Views/UIStores/GeneralSettings.cshtml
dennisreimann: I bet it is caused by the text in `MultiplePaymentsOnly` being longer than the width of the content allows for.

Issue #3190, opened 2021-12-09
Bug allowing Spam/Denial of Service, due to gap limits in other wallets (& related feature requests to fix)


ChristopherA:
We are using the docker version of BTCPay v1.3.6.0 at https://btcpay.blockchaincommons.com.

Original report: https://github.com/btcpayserver/btcpayserver-docker/issues/561

**Describe the bug**

BTCPay does to make it easy to stop spam attacks, resulting in a possible large address gap, breaking spending wallets (in particular, Ledger Live).

Details:

We had a spammer hit our BTCpay server, creating just under 500 invoices in a bit less than an hour. All with the same email address (sample@email.tst), all for either 0 BTC or 1 BTC. I presume they were all from the same IP address, though that's not recorded with the invoices and if it's somewhere else, I don't know where.

It'd be nice to automatically block a spammer like this, after they've generated several invoices within a very short period of time.

Alternatively, it'd be nice to be able to manually block them, perhaps with a "block IP" or "block email" button on an invoice.

It turns out that this many requests for addresses breaks Ledger Live's gap limit, so it does not see a valid UTXO transaction even though I've increased the Gap Limit in their Settings/Experimental there to the max 999.

So this is now not just spam, but a denial-of-service attack.

It this point I know what my xpub path is, but not the specific index that this particular payment's address is on that path (I'm presuming 1000+?) such that I can manually generate the private key spend this UTXO. (I'll also have to start BTCPay up with a new xpub & path.)

Thus a secondary feature request is that BTCPay's "Invoice Details" add the path and index for each of the addresses it created, and maybe support creation of single UTXO PSBT QR/UR to spend it based on the xpub and path (for details on QR/UR, see https://github.com/BlockchainCommons/crypto-commons/blob/master/README.md#urs).

**Your BTCPay Environment (please complete the following information):**
- BTCPay Server Version:   BTCPayServer v1.3.6.0
- Deployment Method:        Docker
- Browser:                 Safari

cc/ @ShannonA

pavlenex:
@NicolasDorier any thoughts on this one?

NicolasDorier:
Invoice creation is typically done via a backend, so we can't filter by IP. (Except in the specific case of the pay button, crowdfund)

As for the request of seeing the path and index of each address created, you can find it in the invoice details already.

![image](https://user-images.githubusercontent.com/3020646/148946904-2b7ca35f-dbb1-4c93-b8d8-92b455c5c633.png)

[...]

PR #3774, opened 2022-05-23
Improve performance when lot's of pending invoices

NicolasDorier:


pavlenex:
@NicolasDorier A few points about preventing spam were raised in #3190 is there anything we can do to prevent these kind of attacks besides optimizing for them?

PR #3776, opened 2022-05-25
Change default BOLT11 expiry from 30 days to 1 day

bolatovumar:
See https://github.com/btcpayserver/btcpayserver/discussions/3686

Update the default BOLT11 invoice expiration time from 30 days to 1 day but I'm not sure if that's enough. Looks like the default as specified in BOLT11 is 3600 seconds i.e. 1 hour. I wonder if we should change it to be the same.

We could also allow the user to specify the default value not just in days but in days/hours/minutes/seconds as needed.

Capture

https://github.com/lightning/bolts/blob/master/11-payment-encoding.md

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -496,9 +496,9 @@ namespace BTCPayServer.Tests
                 Amount = 12.3m,
                 Currency = "BTC",
                 PaymentMethods = new[] { "BTC" },
-                BOLT11Expiration = TimeSpan.FromDays(31.0)
+                BOLT11Expiration = TimeSpan.FromDays(2.0)
             });
-            Assert.Equal(TimeSpan.FromDays(31.0), test2.BOLT11Expiration);
+            Assert.Equal(TimeSpan.FromDays(2.0), test2.BOLT11Expiration);
 
             TestLogs.LogInformation("Can't archive without knowing the walletId");
             var ex = await AssertAPIError("missing-permission", async () => await client.ArchivePullPayment("lol", result.Id));
@@ -507,7 +507,7 @@ namespace BTCPayServer.Tests
             await AssertAPIError("unauthenticated", async () => await unauthenticated.ArchivePullPayment(storeId, result.Id));
             await client.ArchivePullPayment(storeId, result.Id);
             result = await unauthenticated.GetPullPayment(result.Id);
-            Assert.Equal(TimeSpan.FromDays(30.0), result.BOLT11Expiration);
+            Assert.Equal(TimeSpan.FromDays(1.0), result.BOLT11Expiration);
             Assert.True(result.Archived);
             var pps = await client.GetPullPayments(storeId);
             result = Assert.Single(pps);
--- a/BTCPayServer/Data/StoreBlob.cs
+++ b/BTCPayServer/Data/StoreBlob.cs
@@ -23,7 +23,7 @@ namespace BTCPayServer.Data
         public StoreBlob()
         {
             InvoiceExpiration = TimeSpan.FromMinutes(15);
-            RefundBOLT11Expiration = TimeSpan.FromDays(30);
+            RefundBOLT11Expiration = TimeSpan.FromDays(1);
             MonitoringExpiration = TimeSpan.FromDays(1);
             PaymentTolerance = 0;
             ShowRecommendedFee = true;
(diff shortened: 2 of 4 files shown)


pavlenex:
I honestly don't know what's best, perhaps this is one of those things we tweet out and see what people thing about it? I think you're bringing a fair point around 1 hour, perhaps a wider testing of ln wallets is needed with this. @petzsch this seems like something we can user test?
1. Create a spreadsheet with 10-15 lightning wallets.
2. Generate an invoice with each and note on the expiration date for every. Add it to the column. 
3. Create a conclusion on % of wallets that use x experiation time for invoices and improve our refund flow accordingly. 
Personally I have a feeling lots of wallets implemented 24hours, but if I'm wrong then this will be a big issue, unless something other than bolt11 is used, if so we may want to advertise that instead of bolt11 (lightning address, lnurl?)

petzsch:
Maybe https://www.bitrefill.com/blog/top-11-lightning-network-wallets-bitrefill/ is a good ressource for popular Lightning wallets, personaly I would also add lntxbot (Telegram lightning bot).

I can create the spreadsheet and slowly do some testing, but setting up liquidity to all of the wallets might be a bit nerv wrecking. I can ask on the Testers channel if someone wants to help.

bolatovumar:
@pavlenex I just tested a bunch of wallets and seems like 1 day is pretty safe to use?

Muun: 24 hours
Phoenix: 7 days
Bluewallet: 24 hours
Wallet of Satoshi: 24 hours
Simple Bitcoin Wallet (SBW): 14 days
LightningTipBot: 24 hours

pavlenex:
Thanks for testing that @bolatovumar. I think 1 hour will definitely not be enough time for merchants to react, but one day potentially can be. So I think we should stick to 24hours as you originally proposed.

petzsch:
loosely related question: Can we have such a config option for regular pull payments as well (might be out of scope for this PR)? It confused the heck out of me that this setting only applies to refunds but not regular pull payments, see:

https://github.com/btcpayserver/btcpayserver/discussions/3686#discussioncomment-2816828

NicolasDorier:
There is a fundamental UX issue related to BOLT11 and refunds. @Kukks needs to weight in.

The issue is that when requesting a refund (payout), the merchant will likely settle it once a week or once every few day as it isn't automated and need review.

If users starts submitting 1 days expiration BOLT11, then it will be impossible to refund them most of the time, meaning the merchant need to contact them again...

bolatovumar:
> Lightning wallet should allow customization of expiration, if they don't then it shouldn't be used for refunds until they allow it.

@NicolasDorier of all the LN wallets I tested none actually allow you to customize expiration time. Best they do is just show you what the expiration time is and how is left.

NicolasDorier:
Then we should remove the ability to refund via BOLT11 completely. Need to see with @Kukks

pavlenex:
> If users starts submitting BOLT11 with 1 day expiration, then it will be impossible to refund them most of the time, as by the time the merchant see it, it will already be expired, so the merchant need to contact them again...

At the stage where lightning is used only by smaller, tech savvier merchant's I'd say I disagree, because smaller merchants are way more proactive in handling this vs big systems. It's a chicken and egg problem. Either customer provides bolt11 with bad expiration time, then emails merchants that they cannot submit a refund, or they submit a bolt11, it expires meanwhile so merchant would need to contact them to re-submit, which is I think better scenario. Most wallets go with 1 day by default it seems, so for now I'd leave it at 1 day and let's see what happens. For now we had plenty reports where users just don't know how to do 30-day invoice, which is the reason why we wanted to change it.

NicolasDorier:
@pavlenex I think a better alternative is just to not support this feature. BOLT11 isn't adapted.

petzsch:
Would be a pitty if we had drop bolt11 support for payouts. At least the Breez and SBW don't allow me to create an LNURL or LN-Address for recieving over lightning. (hopefully that changes in the future)

Kukks:
Removing the feature is not a solution.

bolatovumar:
> I think the best thing to do then it is to notify lightning wallets to add the ability to setup expiration

Should I close this PR then?

frnandu:
Even when set to 1 day, when I get invoices for refunds, most are less than 24h, so they get refused with "...Invoice must have an expiry date of at least 1 day...". 
Shouldn't there be an option for specifying hours instead of days???

jacobburrell:
Running into the same issues with pull payments. 

Seems impossible to submit an invoice to claim funds from a pull payment with many wallets.

30-day default that doesn't seem to respond to the store's setting makes it fail.

With `auto-approved` pull payments, there probably shouldn't be a required minimum at all.

Can we at least put something in docs explaining a workaround, or the reason it fails?

TowyTowy:
@jacobburrell @frnandu Any solutions? Having this issue too, would love to be able to have my client be able to withdraw his funds himself.

PR #3784, opened 2022-05-26
Simple Notification Modal Improvement

dstrukt:
Simple changes:
- Adjusts border radius
- Changes background-color to take advantage of the hover (currently background is the same as the hover state color .. might have originally been in the mocks, but correcting)
- Adds border to help differentiate from background

Screen Shot 2022-05-26 at 7 18 20 AM

### DIFF
--- a/BTCPayServer/wwwroot/main/layout.css
+++ b/BTCPayServer/wwwroot/main/layout.css
@@ -296,7 +296,7 @@
 
 #NotificationsDropdown {
     border: 0;
-    border-radius: 8px;
+    border-radius: var(--btcpay-border-radius-l);
     background-color: var(--btcpay-body-bg);
     box-shadow: 0 8px 24px rgba(0, 0, 0, 16%);
     padding: 0;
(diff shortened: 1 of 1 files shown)


NicolasDorier:
it fails because you forgot to separate the numbers with comma

dstrukt:
Ahhh good catch, not sure how I missed those / what I was thinking, will update and rebase.

dstrukt:
Should be fixed now.

BTCPayServer/wwwroot/main/layout.css
```diff
@@ -296,12 +296,13 @@
 
 #NotificationsDropdown {
     border: 0;
-    border-radius: 4px;
-    background: var(--btcpay-bg-tile);
-    box-shadow: 0 2px 16px rgba(0, 0, 0, 0.08);
+    border-radius: 8px;
```
dennisreimann, line 299:
```suggestion
    border-radius: var(--btcpay-border-radius-l);
```

dennisreimann (approved):

dstrukt:
Good call on that last variable change, need to remember to do that myself!

PR #3790, opened 2022-05-30
Greenfield: Fix GetDepositAddress return type

dennisreimann:
The local clients `GetFromActionResult` cannot handle the `JValue` return type, because it gets invoked with `GetFromActionResult<string>`.


Kukks (approved):

dennisreimann:
Will fix the test

PR #3792, opened 2022-05-30
Improve payouts UI

dennisreimann:
Some UI improvements for the payouts section. /cc @dstrukt 

## List

Made the numbers badges so that they are more noticeable.

grafik

## Confirm

Reworked the list to make it look nicer and the button text to be clearer.

payouts-confirm

## Result

Made the status a headline and added a message with details to that.

payout-results

### DIFF
--- a/BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
+++ b/BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
@@ -103,7 +103,7 @@ namespace BTCPayServer.Data.Payouts.LightningLike
             {
                 var blob = payoutData.GetBlob(_btcPayNetworkJsonSerializerSettings);
 
-                return new ConfirmVM()
+                return new ConfirmVM
                 {
                     Amount = blob.CryptoAmount.Value,
                     Destination = blob.Destination,
@@ -133,7 +133,7 @@ namespace BTCPayServer.Data.Payouts.LightningLike
                 var boltAmount = bolt11PaymentRequest.MinimumAmount.ToDecimal(LightMoneyUnit.BTC);
                 if (boltAmount != payoutBlob.CryptoAmount)
                 {
-                    results.Add(new ResultVM()
+                    results.Add(new ResultVM
                     {
                         PayoutId = payoutData.Id,
                         Result = PayResult.Error,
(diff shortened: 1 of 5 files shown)


dennisreimann:
Will hide the badges for the cases where there are no entries.

dennisreimann:
Done, good to go unless @dstrukt wants to do further changes.

dstrukt:
Nice changes @dennisreimann, all looks great! 

1. w.r.t to the pills, want to poke around on 1 more thing in Figma real quick, but should have an update for you soon.

2. I mentioned it to you @dennisreimann, but for my own curiosity's sake I wanted to bring it up again...

I was wondering if it was possible or made sense to try to consolidate the off-chain/on-chain into icons and include it as a "type" in the table like so (even if it's not done in this PR):

image

So that we could eventually push the filtering structure to something closer to how we will eventually handle Invoices, etc..  

e.g. this would simply become a dropdown to select the payment type, thus keeping the Title/CTA/tabbed status hierarchy and pushing the search/filter a bit lower down ... but arguably that might be worth doing all at once.

It also might require me showing all of these views side-by-side with the consistent components, but hopefully for the sake of discussion this all makes sense.

dstrukt:
Minor, and unnecessary depending on the amount of time it would take, but ... I think the only change I would make to the pills would be to fill in the pill, with a 1-2 shade lighter (on light), darker (on dark) grey. 

I do like the minimalism of the outline, but it starts to a look a bit aliased, and a 1-2px thicker border looks too heavy, thus, filled.

NicolasDorier:
btw in the future: let's avoid showing the count of stuff that are potentially large. It's really a tricky performance issue.
Payouts for heavy users, I expect that the count can delay the display of the page of 10-30s.

dennisreimann:
@NicolasDorier While this is technically reasonable, the UX downside shouldn't be discarded though: These counts offer orientation that's otherwise lost and one wouldn't know if there are items behind the tabs without navigating every single tab.

Maybe there's a middle ground and we can cache those counts or something like that, though I know how much of a PITA this can be too.

dennisreimann:
>  think the only change I would make to the pills would be to fill in the pill, with a 1-2 shade lighter (on light), darker (on dark) grey.

Done! We should be good now :)

Kukks (approved):

PR #3805, opened 2022-06-01
Removes `Plugins` tab from navigation on Server Settings

dstrukt:
Maybe this was kept for the sake of redundancy, or some other reason I'm not privy to...? If so, apologies for this PR.

Otherwise, it's redundant, and the "manage plugins" link is easy to find in the left sidebar, so thought we could go ahead and remove it from the tab menu!


dennisreimann (approved):

Kukks (changes requested):
Managing plugins still happens in the Server section of the site. If you remove this, there would be no active navigation in the top menu when on the Plugins page

dennisreimann:
I agree with @dstrukt here, it simply removes the Server Settings subav link in favor of the one in the sidebar.

Bild eingefügt am 2022-6-1 09-41

That link takes you to the Plugins page which doesn‘t have the Server Settings top nav and hence there is no real connection to the Server Settings anyways. 

Broader context: The Plugins page will become more of a separate thing on its own in the future – kinda like a Plugin Store. SO also in that sense I think we shpould move it out of the Server Settings.

dennisreimann:
@pavlenex What do you think? I agree with @dstrukt here and it should be removed.

PR #3488, opened 2022-02-22
Crowdfund finetuning

dennisreimann:
WIP to address #3461. Requires input from @dstrukt to continue.

PR #3806, opened 2022-06-01
Updates Crowdfund & POS Modal

dstrukt:
Originally posted in https://github.com/btcpayserver/btcpayserver/pull/3488, but felt it was probably best to extract to it's own PR to be merged after the changes from #3488 have been made.

## Modal 
- [x] Updates modal title (went with Add/Create Item instead as it's used for both POS/CF)
- [x] Breaks price/currency into new line
- [x] Adds configured currency (USD, etc..) to the input field (@dennisreimann might need help with this one)
- [x] Adds Image Url to Image label
- [x] Moves additional label for `Inventory` & `ID` to the helper text line

This one is a bit trickier.
Remove the "Buy Button Text" field .. We should just default to "Pay" for the Point of Sale, not sure this even applies to the Crowdfund ... but this might be a bit controversial or cause conflicts with people who utilize this feature. I can see why it's utilized, but feels a bit unnecessary imo.

Edit: opted not to remove.

Screen Shot 2022-05-29 at 7 06 25 PM

Additionally, we could add a preview as well, but not sure how feasible the whole feature is technically, outside of simple Input field -> Preview JS mapping. cc @dennisreimann 

With Updates:
Screen Shot 2022-05-31 at 6 13 23 PM

### DIFF
--- a/BTCPayServer/Controllers/UIAppsController.Crowdfund.cs
+++ b/BTCPayServer/Controllers/UIAppsController.Crowdfund.cs
@@ -25,7 +25,7 @@ namespace BTCPayServer.Controllers
 
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
         [HttpGet("{appId}/settings/crowdfund")]
-        public IActionResult UpdateCrowdfund(string appId)
+        public async Task<IActionResult> UpdateCrowdfund(string appId)
         {
             var app = GetCurrentApp();
             if (app == null)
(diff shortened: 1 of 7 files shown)


dennisreimann:
Rebased and added the currency.

grafik

Also improve dthe currency input on the edit view.

grafik

dstrukt:
Beautiful, and yes, I'm glad you were able to get the currency input as well .. will be nice to update that component and have it's functionality cover all view!

Thank you - this one is ready to go!

cc @NicolasDorier @Kukks

PR #3815, opened 2022-06-03
Add Refunds list to Invoice details page

Kukks:
(this is missing the refunded badge and progress as I have another PR that makes this easier to compute to show)
image


dstrukt:
love it!

Kukks:
Another change I did in this PR is that I separated the Invoice info, product info, buyer info and pos data into separate columns. Also, when there is no info for that column, it is not shown. This has made the scrolling heigh of the page shorter. Thoughts @dstrukt @dennisreimann?

dennisreimann (approved):
tACK. I like, this is also a good preparation for the bigger redesign of that page, which @dstrukt is planning.

Issue #3785, opened 2022-05-26
Deleting an admin gives a 500 error, and cannot disable the same user


sethforprivacy:
**Describe the bug**
When attempting to delete a user, BTCPay Server throws a 500 error and does not delete it.

When I attempt to disable that same user, the UI says that the user is disabled via the toast message but the user is not disabled and the logs complain as well.

I also tried disabling/deleting after removing admin privs from the user, also without success.

**To Reproduce the bug**
Steps to reproduce the reported bug:
1. Create a new user and make them admin
2. Login as the user
3. Logout
4. Login as another admin
5. Attempt to delete or disable the user

**Expected behavior**

Would expect disabling/deleting to disable or delete.

**Screenshots**

500 when attempting to delete:

![image](https://user-images.githubusercontent.com/40500387/170551164-73c5d8ac-831e-49ed-bdbb-08283caad8a5.png)

Disabling failing but saying it succeeded:

![image](https://user-images.githubusercontent.com/40500387/170551220-0062db19-0e35-40f6-b17c-717c48196564.png)
![image](https://user-images.githubusercontent.com/40500387/170551245-09042c4a-1222-43c3-a05b-2ddca38c328b.png)

**Your BTCPay Environment (please complete the following information):**
- BTCPay Server Version:   v1.5.3.0
- Deployment Method:        Docker
- Browser:  Brave

**Logs**

```
fail: Microsoft.AspNetCore.Server.Kestrel: Connection id "0HMHV6EOA4NEV", Request id "0HMHV6EOA4NEV:0000013C": An unhandled exception was thrown by the application.
System.InvalidOperationException: The instance of entity type 'ApplicationUser' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.ThrowIdentityConflict(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry, Boolean updateDuplicate)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
[...]

PR #3818, opened 2022-06-04
Fix remove and enable issues with non-last-admin user

rustywave:
- "Remove" issue likely due to multiple context instances. Occurs when this line is called before removal attempt:
`await _userManager.GetUsersInRoleAsync(Roles.ServerAdmin);` 
Not easy to deal with UserManager ChangeTracker so went with solution as described here:
https://stackoverflow.com/questions/60150988/aspnetusermanager-tracking-error-when-adding-and-deleting-users

- "Enable" issue due to conditional check. Added additional check for user id

- Discovered existing bug with modifying users when first disabling, then marking as admin, which remains an issue.

#3785

### DIFF
--- a/BTCPayServer/Services/UserService.cs
+++ b/BTCPayServer/Services/UserService.cs
@@ -109,6 +109,7 @@ namespace BTCPayServer.Services
 
             await Task.WhenAll(files.Select(file => _fileService.RemoveFile(file.Id, userId)));
 
+            user = await _userManager.FindByIdAsync(userId);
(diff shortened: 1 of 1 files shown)


Kukks (approved):
Awesome, thank you. ACK

pavlenex:
Thanks for your contributions @rustywave. Looking forward to more, let us know if we can help somehow. Feel free to join us on Mattermost if you already haven't chat.btcpayserver.org, we have a #developers channel there. 🎉

Issue #3803, opened 2022-05-31
Creating a `Payment Request` should redirect you to the list view upon creation


dstrukt:
Similar to `Pull Payments`, `Invoices`, etc.. to be consistent.

However, currently when you first create one it keeps you back on the "Update/Edit" view.

An additional consideration is should it also take you back to the List view when you're editing an existing Payment Request...? I would think so.

rustywave:
Is anyone working on this yet? If not I can try

rustywave:
#3821 

Both "Create" and "Edit" actions redirect to list view upon creation now. Also updated the success notification message to reflect which action was taken.

rustywave:
Not complete, need to review tests first

dstrukt:
I don't believe anyone is working on it, thanks for picking it up.. will review when the PR is ready!

rustywave:
> I don't believe anyone is working on it, thanks for picking it up.. will review when the PR is ready!

Done, thanks! I left some notes for clarity on the unit tests.

pavlenex:
Done in https://github.com/btcpayserver/btcpayserver/pull/3825

PR #3821, opened 2022-06-04
Creating and editing a payment request now redirects to list view upon completion #3803

rustywave:
Both actions use the same method. Also updated notification message to reflect create or edit action.

image


rustywave:
Closing first, need to check tests

PR #3823, opened 2022-06-05
Minor tweak to Payout UI

dstrukt:
Had to keep tweaking it :) 

Currently, I've removed the `.border` class, but still want to change the `.bg-tile` to the equivalent of `var(--btcpay-neutral-800);` but wasn't sure which bootstrap class would be best to use cc @dennisreimann?

Will make that change based on your suggestion, rebase, and submit for approval.

Will ideally look like this:

Screen Shot 2022-06-05 at 4 43 06 AM

### DIFF
--- a/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
+++ b/BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
@@ -10922,6 +10922,15 @@ input[type="number"].hide-number-spin {
   -moz-appearance: textfield;
 }
 
+/* custom bg and border classes */
+.bg-medium {
+  background-color: var(--btcpay-body-bg-medium);
+}
+
+.border-medium {
+  border-color: var(--btcpay-body-border-medium);
+}
+
 /* set color for text on background */
 .bg-primary {
   color: var(--btcpay-primary-text);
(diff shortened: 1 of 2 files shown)


dennisreimann (approved):
Done :)

dstrukt:
Nice, LGTM.

Thanks @dennisreimann!

BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
```diff
@@ -10922,6 +10922,15 @@ input[type="number"].hide-number-spin {
   -moz-appearance: textfield;
 }
 
+/* custom bg and border classes */
```
Kukks, line 10925: Are you sure about adding things to this file? Isn't this file going to be overwritten once we update bootstrap?

Kukks (changes requested):

BTCPayServer/wwwroot/main/bootstrap/bootstrap.css
dennisreimann: These are additions to the Bootstrap file we append as part of the design system build process. So all is good here.

PR #3825, opened 2022-06-05
Creating and editing a payment request now redirects to list view upon completion

rustywave:
Create and edit now redirect to list view. Both actions use the same method. Also updated notification message.

- xUnit: Kept extraneous redirect value so xUnit tests pass w/o modification. If there is a better alternative please let me know and I will fix. The tests won't have access to the new payment req id otherwise, but could maybe expose via ViewData but that seems messier.  I noticed Invoices and Pull Payments don't have these same tests.

> return RedirectToAction(nameof(GetPaymentRequests), new { storeId = store.Id, **payReqId = data.Id** });

- New notification message:
image

#3803

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -758,6 +758,7 @@ namespace BTCPayServer.Tests
             currencyInput.SendKeys("BTC");
             
             s.Driver.FindElement(By.Id("SaveButton")).Click();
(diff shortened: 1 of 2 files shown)


dstrukt:
tACK on my end.

Nice work @rustywave!

dstrukt (approved):

Kukks (approved):
ACK, thanks for your first contribution 🎉

PR #3834, opened 2022-06-07
Mobile header improvements

### LINKED ISSUE #3826: Minor updates to mobile header navigation
Been doing some heavy testing on the mobile side of things while working through a few designs, few improvements I'd like to suggest:

- Moving the hamburger menu to the left .. was an oversight on my original mocks, and it's a bit annoying to click to the right for the menu to open on the left (which I like that it does) ... so it makes sense to keep them both on the same side. Open to feedback here though.

- Making the StoreSelector fill the remainder of the top bar on mobile / tablet .. arguably we could do like 60% on tablet, but regardless, feels a bit more polished to me, but open to feedback here as well:

Screen Shot 2022-06-03 at 4 38 13 AM
Screen Shot 2022-06-03 at 4 38 29 AM

If there's support this is probably a ticket I could take care of, but wanted to see if there's love before any work is started.

Current (for reference):
Screen Shot 2022-06-05 at 5 01 25 PM
Screen Shot 2022-06-05 at 5 01 31 PM

cc @dennisreimann

dennisreimann:
- Moves the menu button to the left, besides the logo
- Improves width and text display of the store selector
 
Closes  #3826.

### DIFF
--- a/BTCPayServer/Views/Shared/_Layout.cshtml
+++ b/BTCPayServer/Views/Shared/_Layout.cshtml
@@ -28,14 +28,14 @@
 <body class="d-flex flex-column flex-lg-row min-vh-100">
 <header id="mainMenu" class="btcpay-header d-flex flex-column">
     <div id="mainMenuHead">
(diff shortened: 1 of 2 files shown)


dstrukt (approved):
LGTM!

Send it.

Kukks (approved):

PR #3836, opened 2022-06-08
Allow uncompiled plugins

NicolasDorier:
I want to make plugins more "dynamic" my having the ability to not have to compile your plugins to deploy them, maybe even allow live modification of plugins.

This PR isn't here yet, it's just a PoC right now, which use the current .btcpay packages, and automatically compile .cs files inside it if .dll isn't present.

The plugins can reference btcpayserver directly.
However, there isn't really point to merge it yet, as requesting to build a plugin isn't that hard, and this still doesn't allow editable plugins.


dennisreimann:
@NicolasDorier Is this still relevant?

NicolasDorier:
yes, leaving it for later

PR #3838, opened 2022-06-08
Dashboard: Add Lightning balances and services

### LINKED ISSUE #3663: Add lightning to the dashboard
Since we have a wallet balance in the dashboard, we should consider at least adding lightning balance. 

There are many tricky things for this to be implemented, so opening this one for discussion and as a tracker for 1.6.0 or a release in between.

dennisreimann:
Closes #3663.

Things to be discussed and edited:
- Size and responsive behaviour of the tile
- Is there more info to display?

ln-tile

### DIFF
--- a/BTCPayServer.Client/BTCPayServer.Client.csproj
+++ b/BTCPayServer.Client/BTCPayServer.Client.csproj
@@ -28,7 +28,7 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
     <ItemGroup>
-      <PackageReference Include="BTCPayServer.Lightning.Common" Version="1.3.4" />
+      <PackageReference Include="BTCPayServer.Lightning.Common" Version="1.3.6" />
       <PackageReference Include="NBitcoin" Version="7.0.1" />
       <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     </ItemGroup>
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -48,7 +48,7 @@
   <ItemGroup>
     <PackageReference Include="BIP78.Sender" Version="0.2.2" />
     <PackageReference Include="BTCPayServer.Hwi" Version="2.0.2" />
-    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.3.8" />
+    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.3.10" />
     <PackageReference Include="BuildBundlerMinifier" Version="3.2.449" />
     <PackageReference Include="BundlerMinifier.Core" Version="3.2.435" />
     <PackageReference Include="BundlerMinifier.TagHelpers" Version="3.2.435" />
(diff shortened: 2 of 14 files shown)


dstrukt:
Nice!

Still exploring mocks, have quite a few suggestions here that will change this view :) 

But from a data POV, only other immediate things:

Lightning:
- Node Alias
- Number of Peers
- Number of (Active) Channels .. could eventually show other channel states

On-Chain:
- Unconfirmed Balance?
- Reserved Balance?

dennisreimann:
> On-Chain: Unconfirmed Balance? Reserved Balance?

If it is more than zero it will be displayed, otherwise we are omitting it. (less noise)

dstrukt:
ACK!

dennisreimann:
As discussed in yesterdays design meeting: The balance and services tiles are now two separate things.

## Mobile

ln-mobile

## Tablet

ln-tablet

## Desktop

ln-desktop

dstrukt:
Hell yeah, will review this in a few hours!

dstrukt:
Conceptual:

- [ ] I think the `Wallet Balance` may need to be updated to `On-Chain Balance` to make the distinction clearer with the `Lightning Balance` underneath .. thoughts?

Functional nit:

- [ ] If no balance on either, "show details" should be hidden, doesn't do anything

Suggestions:

`Manage` for the Lightning Services links to the Lightning tab or a place to configure .. or the way to configure, as if you don't have them configured, the services link to a 404 .. then I don't know where to go to resolve.

`Show Details` does the same, and `Node Info` is just moved over to the other card.

Screen Shot 2022-06-10 at 11 49 21 PM

Otherwise, looks great, excellent work!

pavlenex:
> I think the Wallet Balance may need to be updated to On-Chain Balance to make the distinction clearer with the Lightning Balance underneath .. thoughts? cc @pavlenex

I don't think it's needed, people using Lightning at this stage probably have good understanding. Plus it's structured so it say's "lightning balance" then on-chain below. I wouldn't change on-chain wallet wording for now.

dennisreimann:
> it currently links you to a 404

That's because these services are fake in the dev env.

>  If no balance on either, "show details" should be hidden, doesn't do anything, regardless of which direction we go (see below)

Updated. 

For the rest I'd say the current version is good as it is and the headlines don't seem ideal, as you noted. Right now we have "BTC in channels" and "BTC on-chain", which is clearer imho. We can ship and see if people get confused here, maybe this isn't a big problem and they'll notice that theis refers to the Lightning balance, because it is part of the Lightning balance tile ;)

NicolasDorier:
so nice.
I think we'll need eventually to refactor the Dashboard so it get refreshed in the background, or maybe asynchronously via some JS. We are requesting lot's of things sequentially, and this add up quick.

dennisreimann:
@NicolasDorier agreed, there's already an endpoint for the data fetching of the wallet balance tile – this can be done for the other tiles as well. 

We'd need some kind of loading state then, ideally so that the tiles already have the correct size and the dashboard doesn't look shaky once the data starts coming in. 

Separate PR imho.

dstrukt (approved):

gruve-p:
Works fine for GRS:
image

PR #3845, opened 2022-06-09
support onion ln conn strings

### LINKED ISSUE #1648: Adapt Lightning Client to use The Tor HttpClient
`LightningClientFactoryService` should switch to use an HttpClient with Tor when the conn string is an onion service

Kukks:
I'm not sure if onion based ln conenction strings ever worked in btcpay? closes #1648

### DIFF
--- a/BTCPayServer/Hosting/BTCPayServerServices.cs
+++ b/BTCPayServer/Hosting/BTCPayServerServices.cs
@@ -105,6 +105,8 @@ namespace BTCPayServer.Hosting
             services.AddSingleton<ISwaggerProvider, DefaultSwaggerProvider>();
             services.TryAddSingleton<SocketFactory>();
             services.TryAddSingleton<LightningClientFactoryService>();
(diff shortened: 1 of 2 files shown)


Kukks:
> was not expected

This sounds like an nginx issue

Kukks:
hmm, this is how lnbits solved it: https://github.com/PrinzMonty/lnbits-legend/commit/f0bb470b74343159c76da26ab7cce3e620942c52

NicolasDorier:
I think this code should work, but you should try on a real server hosted on onion to be sure

we don't use the proxy anymore

since .net6.0, onion is supported directly by .net

I think I removed the proxy a while ago

NicolasDorier:
The error of @dennisreimann seems to show that it worked though, because he got a http response back.

NicolasDorier:
I merge this, even if we can't really test, we can't break something that never worked in the first place.

dennisreimann:
Alright, I'll try this in production once the next version comes out.

PR #3846, opened 2022-06-09
Add new issue forms for filing Bug Reports

### LINKED ISSUE #3788: Create issue forms for the main repo
[Issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) are great way to have a form which users could fill in to provide relevant info when filing an issue, it's certainly way better to what we have now.

sample-issue-form

https://youtu.be/qQE1BUkf2-s

Live demo here https://github.com/mishmanners/GitHub-Like-A-Boss/issues/new?assignees=octocat&labels=bug%2Ctriage&template=test.yml&title=%5BBug%5D%3A+

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

pavlenex:
Closes #3788 
This now gives us better way to get information from people when filing a bug report. It was painful to get it working as GH is very picky about YML and has their own custom way of validating things, so it took quite a few attempts to make it work.

Live demo https://github.com/pavlenex/btcpayserver/issues/new/choose 

ezgif com-gif-maker-2

If you guys have any suggestions on what kind of information should we require prior to people submitting a bug report, let me know and I'll add it.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,72 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report! Please provide as much information as you can. It helps us better understand the problem and fix it faster.
+  - type: textarea
+    id: version
+    attributes:
+      label: What is your BTCPay version?
+      description: You can version in footer's bottom right corner
```
Kukks, line 14:
```suggestion
      description: You can see the version in the footer's bottom right corner
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,72 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report! Please provide as much information as you can. It helps us better understand the problem and fix it faster.
+  - type: textarea
+    id: version
+    attributes:
+      label: What is your BTCPay version?
+      description: You can see the version in the footer's bottom right corner
+      placeholder: I'm running BTCPay v 1.xxxx
```
dennisreimann, line 15:
```suggestion
      placeholder: I'm running BTCPay v1.X.X.X
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,72 @@
...
+    attributes:
+      label: What browsers are you seeing the problem on?
+      multiple: true
+      options:
+        - Firefox
+        - Chrome
+        - Safari
+        - Microsoft Edge
+        - Brave
+        - Other (not applicable)
```
dennisreimann, line 59: Not sure if this should be a dropdown, because there are so many options and we should at least add Tor and popular mobile ones too.

pavlenex: Good point , addressed in https://github.com/btcpayserver/btcpayserver/pull/3846/commits/7ccbda21751701c515df7930cc6e75880166f464

pavlenex:
The only "problem" with this approach is that a template cannot be altered. So if contributors would like to file a ticket, they'd need to go through this, then edit. I mean I can probably add the ability to submit a blank issue, but then there's a possibility people will ignore it. Thoughts?

dstrukt:
Ahh that's kinda tricky.

Here's how lightning labs' is configured, and i like theirs. 

Can obv. still can just clear out everything (for the users who do this) to submit a bug without regard, but I guess it's alternative given the trade-offs of above, which are tricky .. both leave a bit to be desired on both sides...

Apologies if you already knew of doing it this way and ruled it out:

Screen Shot 2022-06-10 at 12 07 46 AM

Screen Shot 2022-06-10 at 12 01 37 AM
Screen Shot 2022-06-10 at 12 01 45 AM

pavlenex:
@dstrukt We already have that and it sucks. Forms are way better.

pavlenex:
Enabled blank issues. Will merge it and see how people use it.
Screenshot 2022-06-10 at 13 10 49

PR #3851, opened 2022-06-10
Store Settings feature with own table

Kukks:


pavlenex:
What does this PR address?

Kukks:
It is the alternative of https://github.com/btcpayserver/btcpayserver/pull/3689 but with nicolas' preference

NicolasDorier:
The tests were failing, I refactored the code to include directly inside StoreRepository and Using IStoreRepository to expose the store repo to plugins. It doesn't make sense to have a separate repository just for settings.

I removed code related to cache. It's better we don't cache this, unlike server settings which are retrieved at every request, and all the same for all users, this is not the case for store settings.

I removed the `WaitSettingsChanged` as well, because we shouldn't have unbounded calls to WaitSettingsChanged.
I also removed publish to EventAggregator. Because the SettingsChanged event is for server settings. This event also doesn't support a null settings, because we can't delete server settings, but we can delete store settings.

PR #3857, opened 2022-06-11
adds payouts settings button

dstrukt:
It's not the most attractive thing right now with the "Actions" drop-down button right below it, but I think regardless of whether "Settings" is a link to the `Store Settings > Payouts` or we end up rolling the settings into the Payouts view like the Wallet views ... it's a necessary addition. 

Mirrors the functionality we have on Notifications as well.

Also, the `Actions` button will be remove/improved along with similar buttons in the near future, so this is a temporary problem.

Screen Shot 2022-06-10 at 8 03 12 PM

### DIFF
--- a/BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
+++ b/BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
@@ -72,7 +72,12 @@
 <div id="descriptor" class="collapse">
     <div class="d-flex px-4 py-4 mb-4 bg-tile rounded">
         <div class="flex-fill">
-            <p class="mb-3">Payouts allow you to process pull payments, in the form of refunds, salary payouts, or withdrawals.</p>
+            <p class="mb-3">
+                Payouts allow you to process pull payments, in the form of refunds, salary payouts, or withdrawals.
+                You can also
+                <a id="PayoutProcessors" asp-action="ConfigureStorePayoutProcessors" asp-controller="UIPayoutProcessors" asp-route-storeId="@storeId">configure payout processors</a>
+                to automate payouts.
+            </p>
             <a href="https://docs.btcpayserver.org/Payouts/" target="_blank" rel="noreferrer noopener">Learn More</a>
         </div>
         <button type="button" class="btn-close ms-auto" data-bs-toggle="collapse" data-bs-target="#descriptor" aria-expanded="false" aria-label="Close">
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
```diff
@@ -73,21 +73,27 @@
         <a class="alert-link p-0" asp-action="ConfigureStorePayoutProcessors" asp-controller="UIPayoutProcessors" asp-route-storeId="@storeId">Configure now</a>
     </div>
 }
-<h2 class="mt-1 mb-4">@ViewData["Title"]</h2>
+
+<div class="d-flex flex-wrap align-items-center justify-content-between gap-3 mb-4">
+    <h2 class="mb-0">@ViewData["Title"]</h2>
+    <a id="PayoutProcessors" asp-action="ConfigureStorePayoutProcessors" asp-controller="UIPayoutProcessors" asp-route-storeId="@storeId" class="btn btn-secondary">
```
dennisreimann, line 79:
```suggestion
    <a id="PayoutProcessors" asp-action="ConfigureStorePayoutProcessors" asp-controller="UIPayoutProcessors" asp-route-storeId="@storeId" class="btn btn-secondary d-flex align-items-center">
```

dennisreimann:
@pavlenex I talked to @NicolasDorier about it and he raised the concern that it might be a bit too early to feature the processors more prominently. What do you think?

pavlenex:
@dennisreimann I agree, but I'm not sure if that's related with this PR, where would you like to feature processors? Are you referring to a pop up message we show to people?

dennisreimann:
It applies for both: We have the "protip" alert and not this settings button, which guides people towards using the automation. Nicolas' point is, that it might be too early to advertise the payout processors so prominently.

pavlenex:
I'd agree with that.

dennisreimann:
Ok, let's make this a draft then, so we can merge it with v1.7 or other upcoming releases.

Lola806:
How do I collect i don't know how

Specialk209:
Help I'm lost I'm trying to pay biance for my transfer of crypto to USD cash I don't know nothing about this have auto miner app from phone and have 38,000 doll I need to pay 64 dol to the exchange to recieve in Bitcoin they say I can pay with Bitcoin but I don't know how do it

jlena1507:
I need to exchange BTC to USD BITCOIN AMOUNT1349226

dstrukt:
Bump @dennisreimann @pavlenex

dennisreimann:
Not sure we still need this, as we have worked on the view in the meantime, this is what it looks like right now:

grafik

Cleaning it up this intermediate step wuith the big button seemed very noisy:

grafik

So I think the best is to keep the Pro Tip alert for the case where there is no payout processor yet and also have the link to configure them in the expandable info section like this:

grafik

The latter is the state I just pushed in [99f1556](https://github.com/btcpayserver/btcpayserver/pull/3857/commits/99f1556481cc1bff5f568f8f9feb338df913375a).

dennisreimann (approved):

dstrukt:
Fair point, and agree, the Pro Tip is a better solution that solves the issue as originally intended -- looks good to me!

PR #3863, opened 2022-06-13
Server Settings: Consolidate Storage and Files

### LINKED ISSUE #3861: Consolidate `Server Settings > Services > External storage services` with `Server Settings > Files`
Currently you tab back and forth between Server Settings > `Services` > External storage tab, and the `Files` tab to manage files and the configuration of your BTCPay Server file server.

I propose we combine the two so that all of your files / settings are managed in the same tab.

### Current `Server Settings > Services > External storage`:
Screen Shot 2022-06-12 at 6 41 48 PM
Screen Shot 2022-06-12 at 6 42 13 PM
Screen Shot 2022-06-12 at 6 42 32 PM

### Current `Files`:
[ ]

### Proposed: 

[TODO MOCKS]

dennisreimann:
Closes #3861.

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.Storage.cs
+++ b/BTCPayServer/Controllers/UIServerController.Storage.cs
@@ -236,10 +236,10 @@ namespace BTCPayServer.Controllers
             if (forceChoice || savedSettings == null)
             {
                 var providersList = _StorageProviderServices.Select(a =>
-                    new SelectListItem(a.StorageProvider().ToString(), a.StorageProvider().ToString())
+                    new SelectListItem(typeof(StorageProvider).DisplayName(a.StorageProvider().ToString()), a.StorageProvider().ToString())
                 );
 
-                return View(new ChooseStorageViewModel()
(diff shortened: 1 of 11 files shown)


Rulas12 (approved):

dstrukt:
That was fast, didn't expect it to get worked on for a little while ... regardless, looks great!

Only thing I noticed, we're still keeping the `External storage services` in the Services view .. is that for the sake of redundancy / documentation, or am I missing something?

dstrukt (approved):

dstrukt:
I don't think it's necessary anymore with the current changes, cc @pavlenex we missing any context / reason to keep?

pavlenex:
I don't think keeping them is necessary anymore either, so we can probably remove it, unless we have a good reason to keep them, which I cannot think of.

dennisreimann:
Ok, removing it now :)

NicolasDorier:
I think you can cleanup more code since `ExternalStorageServices` of the viewmodel isn't used anymore.

Issue #3853, opened 2022-06-10
Improve Shopify error when incorrect store url is entered instead of giving 500


gkrizek:
### What is your BTCPay version?

I'm running v1.5.4.0

### How did you deploy BTCPay Server?

btcpayserver-docker repo, so the standard Docker deployment

### What happened?

When trying to connect a Shopify store via the Plugin, BTCPay Server throws a 500. It is not able to save the Shopify integration information

### How did you encounter this bug?

1) Go to your store's settings page
2) Click on Plugins
3) Besides Shopify, click Setup
4) Enter in your store details and keys
5) Click Save
6) Receive a 500 error

### Relevant log output

```shell
System.NullReferenceException: Object reference not set to an instance of an object.
   at BTCPayServer.Plugins.Shopify.ShopifyApiClient.CheckScopes() in /source/BTCPayServer/Plugins/Shopify/ShopifyApiClient.cs:line 76
   at BTCPayServer.Plugins.Shopify.UIShopifyController.EditShopify(String storeId, ShopifySettings vm, String command) in /source/BTCPayServer/Plugins/Shopify/UIShopifyController.cs:line 252
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfIActionResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeNextActionFilterAsync>g__Awaited|10_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
[...]

PR #3865, opened 2022-06-14
Update Fully Noded link

dennisreimann:

PR #3867, opened 2022-06-14
add exception check to shopify UI Controller

BaderSZ:
Fix for issue #3853 

The Shopify plugin crashes and returns a 500 error if an incorrect URL is entered during setup. Added a try-catch block to check, works locally with regtest.

image

### DIFF
--- a/BTCPayServer/Plugins/Shopify/UIShopifyController.cs
+++ b/BTCPayServer/Plugins/Shopify/UIShopifyController.cs
@@ -164,11 +164,11 @@ namespace BTCPayServer.Plugins.Shopify
                     order = await client.GetOrder(orderId);
                 }
 
-                 return Ok(new
-                    {
-                        invoiceId = firstInvoiceSettled.Id,
-                        status = firstInvoiceSettled.Status.ToString().ToLowerInvariant()
-                    });
+                return Ok(new
+                {
+                    invoiceId = firstInvoiceSettled.Id,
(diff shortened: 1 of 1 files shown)


BaderSZ:
These pipeline errors seem to be project wide. Or is this something I need to fix?

dennisreimann:
@BaderSZ Yes, this is something I fixed in #3865.

dennisreimann (approved):

NicolasDorier:
you shouldn't catch NullReferenceException, but check what is null and set the error message.
Catching exception as logic handling is ugly.

NicolasDorier:
Issue is `vm.ShopName` being null.

NicolasDorier:
Fixing it properly, there are several code smell in the shopify client that make it impossible for you to fix properly without this hack.

NicolasDorier:
@BaderSZ I fixed it on https://github.com/btcpayserver/btcpayserver/commit/0da97c5da31a15778fc1dc5fd00000f71157537b

The problem with your PR is that you are catching the `NullReferenceException`, this is putting bandaid on the issue, the real issue is that there shouldn't have a `NullReferenceException` and that the previous call to `OrdersCount()` should have failed with the error message from shopify reported to the UI.

Looking at the implementation of `ShopifyAPIClient`, it was only handling a single type of error, while hiding all the rest.
For example, `OrderCount` had an error, but it was totally ignored, so it crashed along the road later on. (in CheckScopes)

BaderSZ:
@NicolasDorier sure thing, thanks for fixing it!

PR #3873, opened 2022-06-15
Move `View` action to the `Name` column in Payouts & Payment Requests

dstrukt:
More minor tweaks from me:
- Trying to minimize the number of actions on the right, in prep for future consolidation, by moving the "View" action to "Name" column, and adds a `target="_blank"` to open the Preview in another tab. If we don't like the `target` functionality, I think we should eventually add a "Back to store" or some similar CTA on the "Live View/Preview" to make it easier to get back to the store.
- Changed `Refunded` to `Claimed` table head in Payouts as it's more general, cc @pavlenex

The first point may be a bit controversial, so if there's no love, will remove and resubmit with the second item patch.

Before:
Screen Shot 2022-06-15 at 1 46 25 AM
Screen Shot 2022-06-15 at 1 46 32 AM

After:
Screen Shot 2022-06-15 at 1 47 49 AM
Screen Shot 2022-06-15 at 1 47 58 AM

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -759,35 +759,38 @@ namespace BTCPayServer.Tests
             currencyInput.SendKeys("BTC");
             
             s.Driver.FindElement(By.Id("SaveButton")).Click();
-            s.Driver.FindElement(By.XPath($"//a[starts-with(@id, 'Edit-')]")).Click();
+            s.Driver.FindElement(By.XPath("//a[starts-with(@id, 'Edit-')]")).Click();
+            var editUrl = s.Driver.Url;
+            
             s.Driver.FindElement(By.Id("ViewPaymentRequest")).Click();
-            s.Driver.SwitchTo().Window(s.Driver.WindowHandles.Last());
+            var viewUrl = s.Driver.Url;
+            
             Assert.Equal("Amount due", s.Driver.FindElement(By.CssSelector("[data-test='amount-due-title']")).Text);
             Assert.Equal("Pay Invoice",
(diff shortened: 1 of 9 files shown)


dennisreimann (approved):

dennisreimann:
Now that I started fixing the tests I wonder whether or not we should rather link the edit page with the name and keep the view in the Actions list. At least my expectation is to get to the editing view when clicking the name, not to the public version of the item.

dstrukt:
Hmm that's a good point, I think you might be right .. happy to roll with the `edit` for the name, and view as the "primary" action, achieves the original intention.

dennisreimann:
I have removed the `target="_blank"` from the links, because that made the tests go crazy. We can put it back and work on the tests, but I think we shouldn't interfere with the standard behaviour here: If a user wants to open a separate tab or window, they can do so via other means. (personally I don't like that behaviour, but that's just subjective I think).

I like the back button idea though and we can place it in the top bar of those public pages in case the user is signed in to the backend.

dstrukt:
It's better than the current experience imo, but yeah I don't love `target="_blank"` either, but figured it would work as a stopgap until we can properly address the latter. 

Happy to remove until we can do so!

dennisreimann:
Alright, as discussed via MM I've now added the back links above the footer. This will be cleaned up in one of the next releases, where we consolidate the public pull payment and payment request views with what we work towards on the public receipt page.

dennisreimann (approved):
Good to roll!

NicolasDorier:
For next time: If there is doubt about whether or not to use `_blank` you can do like we do in the invoice list by offering both `Checkout [^]`.

PR #3875, opened 2022-06-15
Fix warning from CLI parser

gruve-p:
`-p` is deprecated as an abbreviation for `--project`, and using `-p` generates a warning.

Source: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/deprecate-p-option-dotnet-run


dennisreimann (changes requested):
Thanks! As the docs states, this now needs a double dash.

dennisreimann (approved):

PR #3877, opened 2022-06-15
Adding description to speedPolicy parameter.

### LINKED ISSUE #3808: Greenfield API: Explain speedPolicy setting
The speedPolicy (checkout options) api docs need to be explained further as we do not have an equivalent setting in the wallet settings and this is confusing new developers. 

In stores wallet settings you can set the number of confirmations 0, 1, etc. when an invoice is considered "settled" but on the `speedPolicy` setting we only have enum values which do not make clear how many confirmations this will mean:

Enum: "HighSpeed" "MediumSpeed" "LowSpeed" "LowMediumSpeed" 

We should add additional info e.g. "HighSpeed" = 0 confirmations etc. 

Also add a note that this will override the wallet settings.

ndeet:
Fixes #3808 

Proper description with explanation on how many confirmations the settings relate to.

image

Note: for some reason the enum values are shown twice, but this is also true for the current 1.5.4.0 release so this "bug" was not introduced with my change here. I checked but I do not see why they should be duplicated. Also on the docs.btcpayserver.org page of the api docs this does not happen.

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
@@ -1063,7 +1063,7 @@
             },
             "SpeedPolicy": {
                 "type": "string",
-                "description": "",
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):

dennisreimann:
Nice! You can rebase to make the tests work, but it's just the Fully Noded linkcheck that is failing here.

ndeet:
Thank you! Rebased and force pushed. At some point when the docs grow 10x we won't be able to merge anything because of linkchecker failing all the time on temporary connectivity issues of various sites 😅

PR #3880, opened 2022-06-17
Refactor Authorize UI

### LINKED ISSUE #2902: Refactor Authorize UI
The code is good, but we have issues with the UI. It should be refactored so that you would first select the stores, and then the permissions. This is essential for an easy eCommerce integration as you'd never create an API key with permission A for store A and a different permission for store B. All stores should always get the same permissions and the UI should enforce that.

This is prolly out of scope for this PR though.

I discussed this with @Kukks and we decided not to use specific store permissions for the integrations for now. It's just not user-friendly enough to go this route.

Ideally you'd have an eCommerce site that has a "Connect" button in the backoffice. You click that and land on the Authorize API key page. You select the stores you want to connect and the required permissions are prefilled and disabled (strict mode).
You confirm and return to the eCommerce backoffice.
If the eCommerce site is multi-site capable (like Magento), you select which BTCPay Store should go with which website. Because you now have the API key, you can list the store IDs and build the dropdowns.

This flow is currently not possible as the user may make a mistake and fail to select the same stores for the permissions. So strict mode only applies to permissions, but not to the stores you're choosing under these permissions. We could make it so that strict mode also forces the same permissions on each store, but I feel it makes more sense to change the UI so stores come first and permissions second.

_Originally posted by @woutersamaey in https://github.com/btcpayserver/btcpayserver/issues/2894#issuecomment-923818336_

dennisreimann:
Closes #2902.

## Approach

- If there is at least one store policy, which has been requested, we show the store selection upfront? Same goes for non-strict mode, because the user might select a store permission.
- Seen the other way around: The only time we then will not show the store selection upfront is in strict mode and no store policies requested.

## To be discussed

- For the store selection: Do we need the ability to select multiple stores (like it is right now at per policy level) or is it going to be just one and we assume one API key == one store.
- Once we have this done, should it be like that with the general API key configuration too? Right now we have the most flexible solution, with permissions and potentially multiple stores. Should it stay that way or shall the other API config also have store selection first and then hav the permissions be applied to that store(s) only?

### DIFF
--- a/BTCPayServer.Tests/ApiKeysTests.cs
+++ b/BTCPayServer.Tests/ApiKeysTests.cs
@@ -234,6 +234,14 @@ namespace BTCPayServer.Tests
             TestLogs.LogInformation($"Checking API key permissions: {allAPIKey}");
             var apikeydata = await TestApiAgainstAccessToken<ApiKeyData>(allAPIKey, "api/v1/api-keys/current", tester.PayTester.HttpClient);
             Assert.Equal(checkedPermissionCount, apikeydata.Permissions.Length);
+            
+            TestLogs.LogInformation("Checking empty permissions");
+            authUrl = BTCPayServerClient.GenerateAuthorizeUri(s.ServerUri, Array.Empty<string>(), false, true).ToString();
+            s.GoToUrl(authUrl);
+            select = new SelectElement(s.Driver.FindElement(By.Id("StoreId")));
+            select.SelectByIndex(0);
+            s.Driver.FindElement(By.Id("continue")).Click();
+            Assert.Contains("There are no associated permissions to the API key being requested", s.Driver.PageSource);
         }
 
         async Task TestApiAgainstAccessToken(string accessToken, ServerTester tester, TestAccount testAccount,
--- a/BTCPayServer/Controllers/UIManageController.APIKeys.cs
+++ b/BTCPayServer/Controllers/UIManageController.APIKeys.cs
@@ -203,12 +203,16 @@ namespace BTCPayServer.Controllers
                     return RedirectToAction("APIKeys", new { key = key.Id });
 
                 default:
-                    var requestPermissions = Permission.ToPermissions(viewModel.Permissions?.Split(';').ToArray()).ToList();
-                    var existingApiKey = await CheckForMatchingApiKey(requestPermissions, viewModel);
-                    if (existingApiKey != null)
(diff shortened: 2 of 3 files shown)


woutersamaey:
This sounds good to me.

1 API key per store also works. I have no requirement to reuse the same API key for multiple stores, because in Magento you configure 1 store at a time. If you have 3 stores, you'll be making 3 configurations. I expect other eCommerce software to be similar.

As for the 2nd question, I'm not following 100%. The API permissions page in can be improved in many ways, like visualize the hierarchy of permissions (i.e. unrestricted rules all others), so I'd like the page alone unless we can totally redo it.

woutersamaey:
I have not reviewed the code or ran it. Would love to see some screenshots here though.

dstrukt:
Hype. 

Will review this evening.

ndeet:
1 Agree that would work too I think without introducing a separate flag.

2 Not sure if we should remove the most flexible config we have at the moment although I think you will seldomly create an API key that allows permission X for store A and permission Y for store B. But what you might want to be able to do from the admin UI is to have a "everything" API key that has all permissions for all stores. I think this would be useful for testing and development? But no hard feelings either way.

woutersamaey:
> a "everything" API key that has all permissions for all stores.

This is called the “Unrestricted” permission. It exists but is shown in a list like any other permission even though it is like a “master switch” when looking at the code. 

The ideal UI would be a tree-like structure where “Unrestricted” is a checkbox at the top and if hou check that, everything underneath becomes also checked (and disabled so the values don’t submit). 

But there is more. Some permissions are also parents of others. For this you need to look into the code as it’s not clear visually.

dennisreimann:
Updated it to improve the confirm case: It's handling existing API keys better now and the confirm view is improved too:

auth-existing

dstrukt:
Haven't had a moment to properly review this PR (should have time this evening), but the screens are looking great, awesome to see the design system holding up so well on generalized components, and good to see areas where we can continue to improve as well!

dennisreimann:
@Kukks Thanks for spotting those! Added tests for the mentioned cases and fixed them.

dstrukt (approved):
tACK + discussion and alignment in design calls

NicolasDorier:
@Kukks if that feel good for you now let's merge this

Kukks (approved):

PR #3881, opened 2022-06-18
Remove `Store ID` from Store Settings & various minor UI tweaks

dstrukt:
More tweaks while working on Figma mocks.

- Removes `Store ID` from Store Settings, as it's no longer necessary, and has been mentioned a few times we should remove
Screen Shot 2022-06-17 at 5 15 45 PM

- Adjusts main container left & right padding to `80px;` for the largest breakpoint

- Changes `danger` styling on the `Archive` button on Invoice Detail view to our secondary button style .. the `danger` styling suggests an error when simply Archiving an invoice is not an error or reason to be concerned, like Deleting a store, etc..

Screen Shot 2022-06-17 at 5 11 04 PM

- Updates the `max-width` & margin of the user search input, looks a bit nicer


dennisreimann (approved):
tACK :)

NicolasDorier:
I believe Store Id should be exposed, as all our greenfield API relies on it. If we don't expose it in the UI, noobs trying to use our API will give up.

dennisreimann:
Ok, I saw that we also refer to it in the docs that way. Dropped that commit (plus the test fix) and rebased.

PR #3886, opened 2022-06-20
Lightning balance: Handle null values

dennisreimann:
This way we can display only the values the underlying node supports, instead of pretending to know they are zero. See also btcpayserver/BTCPayServer.Lightning#81.


NicolasDorier:
you can bump lib's version

PR #3888, opened 2022-06-21
Fix nav height issue on mobile devices

### LINKED ISSUE #3807: Side nav bottom section cut off on mobile
**Describe the bug**
When using BTCPay Server on a mobile device's browser, the bottom elements of the navigation are not visible.

**Screenshots**
image

- BTCPay Server Version:    1.5.4
- Deployment Method:      Docker
- Browser: Edge for Android

bolatovumar:
close #3807

The reason this bug happen is that when we use "100vh" to set height (which is the default for the sidebar nav) the browser bar is not counted towards the height of the screen as long as it's in view so 100vh actually ends up giving us a height which is bigger than the visible height of the screen. This is a well-known issue and I've run into it before. See here for details:

https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser
https://ilxanlar.medium.com/you-shouldnt-rely-on-css-100vh-and-here-s-why-1b4721e74487

|Before|After|
|---|---|
|before|after|

### DIFF
--- a/BTCPayServer/wwwroot/main/layout.css
+++ b/BTCPayServer/wwwroot/main/layout.css
@@ -419,6 +419,14 @@
         outline: 0;
         transform: translateX(-100%);
         transition: transform var(--btcpay-transition-duration-fast) ease-in-out;
+        /* Fixes https://github.com/btcpayserver/btcpayserver/issues/3807 */
+        height: calc(100vh - var(--mobile-header-height)); /* This line is a fallback for browsers which don't support "fill-available" */
+        height: -moz-available;
+        height: -webkit-fill-available;
+        height: fill-available;
+        /* Since we can't do "calc(fill-available - var(--mobile-header-height));" I'm using "padding-bottom" instead */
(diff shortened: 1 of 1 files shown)


BTCPayServer/wwwroot/main/layout.css
```diff
@@ -419,6 +419,14 @@
         outline: 0;
         transform: translateX(-100%);
         transition: transform var(--btcpay-transition-duration-fast) ease-in-out;
+        /* Fixes https://github.com/btcpayserver/btcpayserver/issues/3807 */
+        height: calc(100vh - var(--mobile-header-height)); /* This line is a fallback for browsers which don't support "fill-available" */
```
bolatovumar, line 423: The only other way to fix this issue is by using JS but I wanted to avoid having to use JS for this.

dstrukt:
+1 to not using JS, this seems like a reasonable, well supported (browser) approach!

NicolasDorier:
f*ck css

PR #3891, opened 2022-06-22
Improve email settings validation and UX

NicolasDorier:
When we clicked on `Send a test mail`, a cryptic `Fill out the required field` was shown. Now the specific missing fields are shown:

image

I also made the `From` field mandatory, since I don't believe it's possible to send a mail without it.

I removed `Display From` field which couldn't work properly due to a bug before.
Instead, the `From` field can accept a RFC822 style of address, this make the email clients properly show the name of the sender/recipient rather than only the email address.

image

Improvement of error message on mailbox parsing.
image

Before we were happily accepting buggy inputs.

I tried to update all places I could with this new MailboxAddress parsing.
Notable changes are NotificationEmail or RefundEmail or BuyerEmail which are now accepting RFC822 style.

Note that concerning email, we have two types of validation: `[EmailAddresss]` which is just an email address (for example, it is used in the Login page) and `[MailboxAddress]` which is the RFC822 style (either a normal address, or Full Name <addr@example.com>)

Lastly, we were sending mail to `test@example.com <test@example.com>` when we sent email to users.
This has been changed, so if we don't know the fullname we just send the mail to `test@example.com` without repeating twice the email. (this is the `ApplicationUser.GetMailboxAddress` extension method I created)


BTCPayServer/Plugins/PayButton/Models/PayButtonViewModel.cs
```diff
@@ -34,7 +35,7 @@ public class PayButtonViewModel
         public string ServerIpn { get; set; }
         [Url]
         public string BrowserRedirect { get; set; }
-        [EmailAddress]
+        [MailboxAddress]
```
dennisreimann, line 38: I think for this one it'd be ok to have a normal email address — but if I get your comment in the description correctly, that'd also work as valid input, so I guess that's correct here.

dstrukt:
tACK.

Great improvement!

dstrukt (approved):

BTCPayServer/Plugins/PayButton/Models/PayButtonViewModel.cs
NicolasDorier: Yes, you can either enter just the email, or fullname + email.

dennisreimann: Perfect, I like 👍

PR #3892, opened 2022-06-22
Remove option of picking LNURL method from custom amount perk in crowdfund app

### LINKED ISSUE #3746: Crowdfund App Perks with custom price are not taken into account of goal (LNURL method at least)
**Describe the bug**

Crowdfund App Perks with custom price are not taken into account of goal (LNURL method at least)

**To Reproduce the bug**
1. Create Crowdfund App
2. Create some Perks with sums
3. Create one Perk with "Price Custom"
4. To go Crowdfund page
5. Click by Custom Perk
6. To pay a Custom Perk by Lightning for example (I didn't test Bitcoin on-chain) by LNURL method
7. When a Custom Perk is paid the sum is not taken into account - a goal volume is not increased and the amount is visible only in the general invoices, but is not visible in the invoices of Crowdfunding itself!

**Expected behavior**

The custom price contribution should be taken into account both in the amount of the goal and visible in invoices of crowdfunding admin area, for example.

- BTCPay Server Version:    1.5.2.0
- Deployment Method:        LunaNode's template
- Browser:                  Firefox 100.*

**Logs (if applicable)**

**Setup Parameters**

**Additional context**

bolatovumar:
Having two ways to get to the LNURL payment method doesn't really make sense, better to leave it to be handled at the invoice level.

close #3746

### DIFF
--- a/BTCPayServer/wwwroot/crowdfund/app.js
+++ b/BTCPayServer/wwwroot/crowdfund/app.js
@@ -25,7 +25,6 @@ document.addEventListener("DOMContentLoaded",function (ev) {
                 amount: null,
                 expanded: false,
                 lnurl: window.lnurl,
-                currentMode: 'standard'
             }
         },
         computed: {
@@ -46,6 +45,7 @@ document.addEventListener("DOMContentLoaded",function (ev) {
             },
             expand: function(){
                 if(this.canExpand){
+                    console.log('lol');
                     this.expanded = true;
                 }
             },
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -1732,35 +1732,8 @@ namespace BTCPayServer.Tests
                 var payoutsData = await ctx.Payouts.Where(p => p.PullPaymentDataId == pullPaymentId).ToListAsync();
                 Assert.True(payoutsData.All(p => p.State == PayoutState.Completed));
             });
-
-            //lnurl crowdfund support
-            s.GoToStore();
-            s.Driver.FindElement(By.Id("StoreNav-CreateApp")).Click();
-            s.Driver.FindElement(By.Name("AppName")).SendKeys("CF" + Guid.NewGuid());
-            s.Driver.FindElement(By.Id("SelectedAppType")).SendKeys("Crowdfund");
-            s.Driver.FindElement(By.Id("Create")).Click();
-            Assert.Contains("App successfully created", s.FindAlertMessage().Text);
-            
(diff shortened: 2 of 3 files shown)


pavlenex:
tACK. 

While testing I noticed something weird. 

Something weird is going on.
1. Create a perk with a fixed price `0`
2. Click contribute
3. Try to click on a perk, nothing happens.
Not sure if this was introduced in your PR or generally a bug we can perhaps address in this PR.

https://user-images.githubusercontent.com/36959754/174971766-ba0283d4-1d64-4bdd-927a-ca4d8964f9a2.mov

In theory, when invoice is 0 it will automatically mark invoice as `settled`. I am not sure what's going on and if we even need to allow 0 in invoices. @Kukks what is this wizardry?

Kukks:
Still nack from my end on just nuking features. Please wait till I am back.

Kukks (changes requested):
Nack on just nuking the whole thing.

bolatovumar:
> Can you explain what feature does this bring, LNURL is already available on the invoice? This just removes an extra step.

Yeah, I figured since it's already available through the invoice having it on the perk itself doesn't provide much extra benefits. I can look into fixing the one on the perk itself if we want to keep it though.

bolatovumar:
> tACK. Custom price is now working as expected.
> 
> While testing I noticed something weird with `fixed price`which I don't think is related with this PR, but perhaps we can tackle it or open a follow up issue from this comment.
> 
> To replicate
> 
> 1. Create a perk with a fixed price `0`
> 2. Click contribute
> 3. Try to click on a perk, nothing happens.
>    Not sure if this was introduced in your PR or generally a bug we can perhaps address in this PR.
> 
>  Screen.Recording.2022-06-22.at.09.35.27.mov 
> In theory, when invoice is 0 it will automatically mark invoice as `settled`. I am not sure what's going on and if we even need to allow 0 in invoices.
> 
> @Kukks what is this wizardry?

@pavlenex seems like this is an existing issue, not introduced by this PR. Tested both on latest `master` and on my instance of BTCPay and can repro it in both. I don't even know what the expected behavior here would be.

NicolasDorier:
Ping @Kukks

Kukks:
Still abroad

BTCPayServer/wwwroot/crowdfund/app.js
```diff
@@ -46,6 +45,7 @@ document.addEventListener("DOMContentLoaded",function (ev) {
             },
             expand: function(){
                 if(this.canExpand){
+                    console.log('lol');
```
NicolasDorier, line 48: need to remove this in any case

NicolasDorier:
I need to understand more about this. We should talk about it in a meeting with @Kukks .

PR #3907, opened 2022-06-28
Dashboard: Load Lightning balance async, display default currency

dennisreimann:
Loading the data asynchronously and rednering it via JavaScript increases the complexity a bit, but I wanted to open this up for discussion and see if this is a potential scaling solution for the dachboard and other tiles.

Also added the default currency switch, now that we are doing more stuff on the frontend in that tile as well.

grafik

### DIFF
--- a/BTCPayServer/Components/StoreLightningBalance/StoreLightningBalanceViewModel.cs
+++ b/BTCPayServer/Components/StoreLightningBalance/StoreLightningBalanceViewModel.cs
@@ -1,14 +1,22 @@
+using BTCPayServer.Client.JsonConverters;
 using BTCPayServer.Data;
 using BTCPayServer.Lightning;
+using BTCPayServer.Services.Rates;
+using Newtonsoft.Json;
 
 namespace BTCPayServer.Components.StoreLightningBalance;
 
 public class StoreLightningBalanceViewModel
 {
     public string CryptoCode { get; set; }
+    public string DefaultCurrency { get; set; }
+    public CurrencyData CurrencyData { get; set; }
     public StoreData Store { get; set; }
-    public WalletId WalletId { get; set; }
+    
+    [JsonConverter(typeof (LightMoneyJsonConverter))]
     public LightMoney TotalOnchain { get; set; }
+    
+    [JsonConverter(typeof (LightMoneyJsonConverter))]
     public LightMoney TotalOffchain { get; set; }
     public LightningNodeBalance Balance { get; set; }
     public string ProblemDescription { get; set; }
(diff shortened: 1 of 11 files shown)


Kukks:
> 

I think the main issue is the database hammering we end up doing. Some installs have so much data, that it can cause the dashboard to take over a minute to load

dennisreimann:
That might be the case for the app/transactions/invoice tiles, but for the LN balance that shouldn't matter. 

For this one I just wanted to reduce the potential impact, which the LN node response latency might have.

NicolasDorier:
why not using the same approach as I think the notifications are?

If I remember for notifications, we fetch the whole html directly rendered by the component in JS and display that.

dennisreimann:
Good point, that would also work. 

What about the idea of making it async in general? It'll increase the complexity a bit nevertheless, but I think it might be worth it.

NicolasDorier:
I like it being async. It would solve the problem of people who just get stuck waiting for a screen to appear when they don't actually care about the screen.

dstrukt:
+1 to async if possible for the reasons @NicolasDorier listed

dennisreimann:
I like this and it's also a bit more modular. We can take it a bit further and do the same for the app/invoices/transactions tiles.

Kukks:
offtopic: would be cool if we could switch to using blazor

dennisreimann:
@Kukks This is only for the LN tile, but I'm working on the other ones right now. Let's have different PRs to keep the scope smaller.

BTCPayServer/Components/StoreLightningBalance/StoreLightningBalanceViewModel.cs
```diff
@@ -1,15 +1,24 @@
...
 
 public class StoreLightningBalanceViewModel
 {
     public string CryptoCode { get; set; }
+    public string DefaultCurrency { get; set; }
+    public CurrencyData CurrencyData { get; set; }
     public StoreData Store { get; set; }
-    public WalletId WalletId { get; set; }
+    
+    [JsonConverter(typeof (LightMoneyJsonConverter))]
```
NicolasDorier, line 16: do you still need that?

dennisreimann: Nope, thanks for spotting!

dennisreimann: Removed in [0f0b2f8](https://github.com/btcpayserver/btcpayserver/pull/3907/commits/0f0b2f839d22e1333de0b136310658fdd4c4b1e6)

dennisreimann:
If we don't have the capacity to review it, we can ship it later.

dennisreimann:
#3916 has the final version plus a fix for the keypad pos stats.

PR #3908, opened 2022-06-29
Can disable cert check for email server

NicolasDorier:
This is meant to address https://github.com/btcpayserver/btcpayserver/discussions/3895 reported by @michnovka

We were before silently removing cert check when the hostname was considered local.
If we have an option for it, then it doesn't make sense to have such silent behavior anymore.

It is migrating old stores with local hostnames to skip cert check so it doesn't blow up unexpectedly for them.

image

### DIFF
--- a/BTCPayServer.Client/Models/EmailSettingsData.cs
+++ b/BTCPayServer.Client/Models/EmailSettingsData.cs
@@ -25,4 +25,5 @@ public class EmailSettingsData
     {
         get; set;
     }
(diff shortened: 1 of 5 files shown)


Kukks (changes requested):
ACK, but we should do a migration for all existing settings which have settings configured and `Extensions.IsLocalNetwork(Server)` = true to mitigate the breaking change

NicolasDorier:
@Kukks updated with migration

Kukks (approved):

PR #3910, opened 2022-06-29
Create interface for providing store id to plugins

### LINKED ISSUE #3824: Some plugins are hidden on left sidebar when navigating to other views
This is a strange one I'm not 100% sure why it happens or if it only happens to SideShift & FixedFloat plugins, those seem to be the ones where this occurs most commonly, but recorded a quick video just clicking between views to show the behavior:

https://user-images.githubusercontent.com/6250771/172049376-86f2c96f-1a48-449a-a243-38814a676b4e.mov

If not a bug, I'll close.

Kukks:
in prep to properly fix #3824

### DIFF
--- a/BTCPayServer.Abstractions/Contracts/IScopeProvider.cs
+++ b/BTCPayServer.Abstractions/Contracts/IScopeProvider.cs
@@ -1,0 +1,7 @@
+﻿#nullable enable
+namespace BTCPayServer.Abstractions.Contracts;
+
+public interface IScopeProvider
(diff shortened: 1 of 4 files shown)


dennisreimann (approved):
LGTM!

NicolasDorier:
Yes, please use the HttpContext in the implementation of the interface instead. It will prevent foot shooting.

Kukks:
@NicolasDorier   just changed it

PR #3911, opened 2022-06-29
Enhance Store email capability

Kukks:
Currenty the new email rules can send an email when an invoice event occurs. However, there is currently no way to customize the email based on the invoice, making the feature a bit useless.

This PR:
* ~adds the rich text editor to the body input~
* allows you to use some of the properties from the Invoice (based on greenfield api properties. I've taken a imple approach for now using just  a string.replace mechanism, but we can update this to a dynamic linq approach so that users can customize further (e.g. `{Invoice.Metadata["something"].ToString().ToUpper()}`)

~NOT READY:
Since this all takes place as a background service, there is an issue around how to handle items such as the "checkout link", as we are not aware of the btcpay url at that moment. Thoughts? @nicolasdorier~

### DIFF
--- a/BTCPayServer/HostedServices/StoreEmailRuleProcessorSender.cs
+++ b/BTCPayServer/HostedServices/StoreEmailRuleProcessorSender.cs
@@ -92,6 +92,6 @@ public class StoreEmailRuleProcessorSender : EventHostedServiceBase
             .Replace("{Invoice.Price}", i.Amount.ToString(CultureInfo.InvariantCulture))
             .Replace("{Invoice.Currency}", i.Currency)
             .Replace("{Invoice.Status}", i.Status.ToString())
-            .Replace("{Invoice.AdditionalStatus}", i.AdditionalStatus.ToString())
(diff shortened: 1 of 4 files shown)


BTCPayServer/HostedServices/StoreEmailRuleProcessorSender.cs
```diff
@@ -54,17 +63,37 @@ protected override async Task ProcessEvent(object evt, CancellationToken cancell
...
     }
+
+    private string Interpolator(string str, InvoiceData i)
+    {
+        //TODO: we should switch to https://dotnetfiddle.net/MoqJFk later
+        return str.Replace("{Invoice.Id}", i.Id)
+            .Replace("{Invoice.StoreId}", i.StoreId)
+            .Replace("{Invoice.Price}", i.Amount.ToString(CultureInfo.InvariantCulture))
+            .Replace("{Invoice.Currency}", i.Currency)
+            .Replace("{Invoice.Currency}", i.Status.ToString())
```
NicolasDorier, line 95: typo?

PR #3912, opened 2022-06-29
remove lnurl from crowdfund as per requests

Kukks:


pavlenex:
How is this different compared to #3892?

Kukks:
> How is this different compared to #3892?

It actually removes all the related code and also the testing code @bolatovumar  left in

PR #3914, opened 2022-06-29
Do not show the Send button if in the send wallet view

NicolasDorier:
Ping @dstrukt 

Myself and another user I know got tricked into clicking on "Send" in the page when we wanted to click on "Send transaction".
This PR is removing the button when we are on the send view so we can't fuck up and lose everything.

Before:
image

After:
image

### DIFF
--- a/BTCPayServer/Components/WalletNav/Default.cshtml
+++ b/BTCPayServer/Components/WalletNav/Default.cshtml
@@ -15,7 +15,7 @@
         </div>
     </a>
     <div class="d-flex gap-3 mt-3 mt-sm-0" permission="@Policies.CanModifyStoreSettings">
-        @if (!Model.Network.ReadonlyWallet)
+        @if (!Model.Network.ReadonlyWallet && ViewData.IsActivePage(WalletsNavPages.Send) is null)
         {
             <a class="btn btn-primary" asp-controller="UIWallets" asp-action="WalletSend" asp-route-walletId="@Model.WalletId" id="WalletNav-Send">Send</a>
         }


dennisreimann:
Please let's make send/receive secondary buttons, but not remove them.

pavlenex:
Problem is that it's sticky, so at some point, on smaller devices you won't even see CTA's bellow, leading you into thinking those are primary ones, no matter how prominent or not they are.

dstrukt:
Apologies, had a lot going on today. 

Typing up a response right now, will edit.

dstrukt:
Hmmm knew/was worried this day was coming, but was hoping it wouldn't be a problem until we could address in full ..  root problem is that there's confusion among navigation and sub-page CTAs.

The problem with the proposed Active/Inactive button is it's not a behavior our users are used to anywhere else in the app, it would be a one-off solution for this view. As an additional consideration while thinking about this ... whatever we do here will define the patterns we use on other pages, if we continue with this navigation pattern.

### Thoughts and historical choices thinking through this...

1. I attempted to address with full-page views (like the Wallet config) for Send/Receive, but this was deemed a little too much, so we scrapped that idea .. and also the Wallet Settings was introduced, so it wasn't just as simple as:
output

Additionally, a problems we also face with the current design is the `Transactions` view isn't as obvious as it could or should be when navigating to one of the subpages.

### TL;DR The feasible options here from my POV:

1. Utilize the sub-page tabbed navigation (like Lightning, Settings, etc..) for the time being, bit of work to revert, but fundamentally it's the most stable option atm. It's not as exciting, but I still think this is the way to go until we define a more usable pattern .. we might have rushed getting this bit in 1.4.0.

2. Disable the button, as suggested above, which does solve the issue functionality, but as mentioned, this navigation structure is not used anywhere else and doesn't match functional patterns we've established. Additionally, we'd simple be recreating our existing tabbed navigation utilizing the other button styles:

Screen Shot 2022-06-29 at 10 22 34 PM

Apologies for the book, but this deserved it.

pavlenex:
Thanks for context @dstrukt I think long-term we should def try to iron out what you proposed in 1.
I think pop up gives a nice focus on important actions. I would leave this out for 1.8.0 or minor release before it, as we already have a lot of things for 1.7.0

PR #3918, opened 2022-06-30
Add refund badge to invoice lists

Kukks:
image
image

### DIFF
--- a/BTCPayServer/Services/Invoices/InvoiceEntity.cs
+++ b/BTCPayServer/Services/Invoices/InvoiceEntity.cs
@@ -392,7 +392,7 @@ namespace BTCPayServer.Services.Invoices
             return GetPayments(network.CryptoCode, accountedOnly);
         }
 #pragma warning restore CS0618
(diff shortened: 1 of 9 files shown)


dstrukt:
Small nit on the badge copy .. shouldn't it be `Refunded` to match the tense of the other status?

dstrukt:
ahhh I see, wasn't aware we didn't know the state.

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -392,7 +392,7 @@ public List<PaymentEntity> GetPayments(BTCPayNetworkBase network, bool accounted
             return GetPayments(network.CryptoCode, accountedOnly);
         }
 #pragma warning restore CS0618
-        public bool Refundable { get; set; }
+        // public bool Refundable { get; set; }
```
Kukks, line 395: I found these props in the codebase, they didn't seem used. Should we remove @NicolasDorier

PR #3921, opened 2022-06-30
Improve wallet nav

dennisreimann:
Use subnav structure as everywhere else. Alternative for #3914.

grafik

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -544,6 +544,7 @@ namespace BTCPayServer.Tests
                 }
             }
             Driver.Navigate().Refresh();
+            Driver.FindElement(By.Id("GoBack")).Click();
             return addressStr;
         }
 
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -544,7 +544,7 @@ namespace BTCPayServer.Tests
                 }
             }
             Driver.Navigate().Refresh();
-            Driver.FindElement(By.Id("GoBack")).Click();
+            Driver.FindElement(By.Id("CancelWizard")).Click();
             return addressStr;
         }
 
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -11,7 +11,6 @@ using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.Lightning;
-using BTCPayServer.Lightning.CLightning;
 using BTCPayServer.Payments;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Invoices;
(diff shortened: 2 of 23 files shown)


pavlenex:
This feels like a step back imo. I would more like to see a better approach, as Dstrukt outlined here. 
176595304-470e1b10-5397-44dc-9345-b668beebde5f


Don't think this is a burning issue and we can patiently wait to properly address it rather than pushing quick patches. For example by introducing this, we now have to change all docs images, and potentially some videos.

dstrukt:
Hmm.

It might be a step back, but also might be the most reasonable approach at this moment.

dennisreimann:
Another quick win alternative to also go into the direction of the proposed would be to use the wizard layout for the Send view.

grafik

dstrukt:
It is possible to remember where the user opened this view from? If from `Settings` back to `Settings`, Transactions, etc..? Could see this coming in handy for Payouts or other instances where you jump to this context.

dennisreimann:
> It is possible to remember where the user opened this view from?

Yes, that's possible, I'm using this for the closing X currently.

dennisreimann:
Updated it, so that you can test it.

dstrukt:
Excited to see this in action!

Testing here in a few.

dstrukt:
output

Few thoughts:

- We could push the Send/Receive buttons back to the top right as the primary CTAs of the page and leave Transactions / Settings in the sub-nav as it is for the time being.

If that is done:
- `Receive` also becomes a full-page like `Send`

Otherwise, this is much more usable, I like Send on a wizard full-page, for numerous reasons, some discussed already. 

Wish the transition between views was a little smoother though .. but we may just be limited here for the time being, curious if this bothers others...

Great work!

dstrukt (changes requested):
Forgot to add, see above.

NicolasDorier:
awesome. I think a test need to be fixed @dennisreimann

dennisreimann:
I'm not really sure about putting the receive view into the wizard as well, because it does not work exactly like in your mock: We do not automatically generate a new address right away, so the only thing the user would get to see after clicking Receive would be the button to generate a new address. 

What do you think?

dstrukt:
Understandable, but yet another slightly deviant functional behavior if we don't.

It's strange to throw a full-page view from the sub-nav for only one, and similarly it'd be strange to pop Send / Receive up to the top right for 1 to open a full-page, and the other to not... hmm.

dennisreimann:
Yeah, I get that and it doesn't seem quite right either way. 

One more point why I feel this might not work: other than the navigation, there is no natural way to finish that view: you copy the address, but there's no CTA that leads you to the next view. So that's different from the other wizard actions and the way to leave would be the back or cancel button, which seems weird imho. 

I'll give the wizard version of Receive a try and will "dress it up".

dstrukt:
Hmm, glad you also see the conundrum of the situation, and why the "reverted" tabbed navigation is the prudent but stable solution atm .. would also love to see the buttons or a better primary CTA work .. just might require some more exploration at this time.

dstrukt:
For those just following the PR, but CBA to build, current state:
output

- I like the Send/Receive being full-page views, it's certainly focused and better than a modal, works well on mobile, and given the context above .. the only thing that's a bit clunky is the transition between views, but there's probably not much we can do about this at this time.

- We could remove the "Back" icon on the Receive view, bit redundant

- The back button on the "Send" view states is interesting, it's a nice functional element .. was trying to figure out all the ways it's triggered, as it's not visible by default.

- Without the sub-nav menu the `Transactions` view is not as easily accessible/discoverable ... which has been a concern raised .. but I'm guessing you explored that and opted for removing? 

Overall, it does address the original concern @NicolasDorier brought up, which is great.

As usual, great work!

dennisreimann:
@dstrukt I've reworked the back button behaviour so that it consistently only shows if there was a previous step inside the wizard.

Let's address the clearer link to the Transactions page in the future.

dstrukt:
Sounds good w.r.t the latter, and reviewing now.

NicolasDorier:
Clicked around, seemed good to me

dstrukt:
Likewise, tACK!

dstrukt (approved):

PR #3928, opened 2022-07-04
Fix build warnings

dennisreimann:
Fix build warnings that I didn't notice before …


pavlenex:
I think you're referring to these two:

Screenshot 2022-07-05 at 18 07 36

PR #3930, opened 2022-07-05
Make LNURL enabled when only method

Kukks:
This fixes the scenario where LNURL for standard invoices are disabled, but the POS Print view only shows LNURL, so the QR code would always error out. The fix is to bypass the setting when lnurl is the only enabled payment method on the invoice


dennisreimann:
tACK, but tests need fixing.

PR #3934, opened 2022-07-06
Added Alby, as Joule has been delisted on Chrome

BitcoinMitchell:
As pointed out in https://github.com/joule-labs/joule-extension/issues/311#issuecomment-1156823457, Joule has been delisted from the Chrome Web Store and most likely won't be back soon, as it has to be refactored to Manifest V3 and the owner is otherwise occupied.

They have recommended [Alby](https://getalby.com/) as an alternative that people could use.


bolatovumar (approved):
I wonder if we should also remove Joule? When going to the Joule website and following the install link I just get a 404 because it's been delisted from the Chrome store. Once Joule is re-listed we can add it back in.

NicolasDorier:
Yes, can you remove Joule as well? though, it's strange we have a 404 and tests still passing.

pavlenex:
Can you connect BTCPay to Alby though?

dennisreimann:
There's no BTCPay connector in Alby, but you can connect the underlying LND node directly:

REST API host: https://mybtcpay.com/lnd-rest/btc
Macaroon: admin.macaroon in HEX format

We could offer that info in the page as we do for the other services.

NicolasDorier:
Actually Joule is still maintained hooray!

https://github.com/joule-labs/joule-extension/issues/311

bumi:
We have Alby users connected to their BTCPay Server. 
It would be great to add a BTCPay Server Connector to make it easier for users to pair. I am happy to add that. 
Can you point me to that config JSON or where users can get the connection details? 

But as @dennisreimann said people simply need the REST API host and the macaroon.

wbobeirne:
👍 I'll give y'all a heads up if I get it back into stores. Sorry for leaving you with a dead link!

NicolasDorier:
Alby implemented an easy way to pair with btcpay https://github.com/getAlby/lightning-browser-extension/pull/1176

PR #3936, opened 2022-07-07
Redirect to invoice details instead of list upon creation

### LINKED ISSUE #3909: [Bug]: Invoice filter persists even through other actions

1.5.4 / master	

n/a local

### What happened?

When using filters on invoices, either through clicking on links to invoices from apps, or simply using the filter system, the filters are persisted.  However, I created an invoice via UI, and was redirected to the invoice list, with the filters still there, meaning my new invoice was not showing.

* Go to an app
*  Click on invoices button
* click on create invoice
* invoice is created and you are redirected to the invoice list with a success message
* invoice is not shown as it is filtered out 

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

Fixes:
* Make invoice creation action take you to invoice details page
OR 
* Clear the filter when creating an invoice.

bolatovumar:
We keep the invoice filters sticky in a cookie so that the user doesn't have to set them again every time they visit the invoices page. However, this results in an issue where the invoice which you just created will not be shown to you if you have filters apply which exclude this newly created invoice.

To get around this issue we can either go to the invoice details page on invoice creation OR clear the cookie or override it somehow for that one time. I took the first approach here since it's kind of weird to clear the cookie or not respect it just in one special case.

close #3909

### DIFF
--- a/BTCPayServer.Tests/CheckoutUITests.cs
+++ b/BTCPayServer.Tests/CheckoutUITests.cs
@@ -37,9 +37,6 @@ namespace BTCPayServer.Tests
             s.Driver.AssertElementNotFound(By.Id("emailAddressFormInput"));
             s.GoToHome();
             s.CreateInvoice();
-            s.Driver.FindElement(By.ClassName("invoice-details-link")).Click();
-            s.Driver.AssertNoError();
-            s.Driver.Navigate().Back();
             s.Driver.FindElement(By.ClassName("invoice-checkout-link")).Click();
             Assert.NotEmpty(s.Driver.FindElements(By.Id("checkoutCtrl")));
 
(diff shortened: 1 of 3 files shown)


NicolasDorier:
yes it makes more sense, I think there is also a checkout link in the invoice detail page anyway

bolatovumar:
> yes it makes more sense, I think there is also a checkout link in the invoice detail page anyway.
> 
> I expect tests to break as selenium probably depends on the redirection.

@NicolasDorier, yes, I will fix the tests.

bolatovumar:
@NicolasDorier fixed the tests in https://github.com/btcpayserver/btcpayserver/pull/3936/commits/0dd2b21dd8a7f43e9da72a5c48b17c7fa12ed6e7

PR #3937, opened 2022-07-07
Blazor - Lightning Balance POC

Kukks:
This PR is a proof of concept of integrating Blazor (server not wasm) into BTCPay and using it for the dashboard. I converted the Store Lightning balance component to use it. Note that we can probably optimize a bit more


NicolasDorier:
hey @dennisreimann what do you think about it?
I suggested to Kukks to try using blazor, because the pattern of querying some route, get HTML back and showing that is exactly what Blazor specialize at.

dennisreimann:
I haven't tested it yet, but in general I like that approach and maybe that‘s a good way to kick things off and get started for using Blazor elsewhere as well. 👍

NicolasDorier:
So I checked Blazor server architecture.
It seems that the state of controls is saved server side. Meaning we can get RAM DDoS if a user open multiple tab.
However, I don't think that's too bad if we don't use it on public facing pages. Also it can probably be throttled by digging a bit.

There is however one UX issue: If the user let the page open, while disconnecting for a while, the server might have lost his state, and thus anything depending on blazor will stop working until the page is refreshed.

PR #3938, opened 2022-07-07
Blazor - Lightning Balance POC

Kukks:
This is the same as #3937, except that it offloads the fetching of rates and store specific settings to the page.  Personally, I prefer this approach  as we can fetch data once and feed them into controls on the page. 

Eventually, we could even do the dashboard page as an actual blazor component, and have the currency selector change as part of a page state.


NicolasDorier:
@Kukks I agree on a global approach for rates. But I think this should be kept separate matter than Blazor.

Kukks:
> @Kukks I agree on a global approach for rates. But I think this should be kept separate matter than Blazor. Having a global approach to get the rate, would allow us to easily allow default currency everywhere we are currently showing bitcoins.

This keeps it a separate manner. The dashboard would fetch the rate and then feed that rate to blazor

dennisreimann:
It fetches the rate once [here](https://github.com/btcpayserver/btcpayserver/pull/3938/files#diff-4d86ed2d71f4f5e86b7a48ef3b0dd0f297f7786c567a800b97b8ae0ce1f168fdR26) and applies it on the dashboard level. In general I like the direction of this, but I cannot judge what technicalities it involves — I haven't used Blazor before, but I like the more unified way of being able to write everything in C#.

dennisreimann (approved):
Rebased and tested.

tACK, though I must say the Blazor overlay when switching away from the dashboard is kinda weird. Can we get rid of that?

PR #3941, opened 2022-07-07
Redirect to dashboard instead of apps list on app delete

### LINKED ISSUE #3898: App Deletion Redirect Destination?
Currently when you delete an `App`, it brings you back to a list view of your `Apps`:

Screen Shot 2022-06-25 at 1 03 01 PM

I'm wondering if this should be the intended functionality, as this is not a view the user sees otherwise and this *could* potentially be a bit confusing...? In addition to being a view we don't actively support.

Not 100% sure how to handle atm, we could just shelve the issue, and wait until we transition these apps to `Plugins`.. thoughts?

Noting this issue for the future.

bolatovumar:
Redirect to dashboard instead of apps list on app delete since we don't really actively maintain this view and this redirect is the only way to get to it currently (if you ignore the fact that user can just type out the URL).

close #3898

### DIFF
--- a/BTCPayServer.Tests/CrowdfundTests.cs
+++ b/BTCPayServer.Tests/CrowdfundTests.cs
@@ -32,6 +32,7 @@ namespace BTCPayServer.Tests
             await user.GrantAccessAsync();
             var user2 = tester.NewAccount();
             await user2.GrantAccessAsync();
+            var stores = user.GetController<UIStoresController>();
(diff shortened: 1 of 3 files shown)


dennisreimann (approved):
Let's also remove the apps list view then :)

dstrukt:
tACK.

And agree with @dennisreimann.

dstrukt (approved):

PR #3947, opened 2022-07-08
Add Bug Fix in Changelog.md

petzsch:
Fixed Ubuntu 22.04 update issue with new SSH client key algo.


petzsch:
As this is only a change for btcpayserver-docker, it's not relevant for the btcpayserver Changelog.

Closed.

PR #3956, opened 2022-07-11
Invoice summary: Fix indentation and heading levels

dennisreimann:
UI fixes for the payment summary partials, which are used on the invoices list and details page.

## Before

grafik

## After

grafik


Kukks (approved):

dstrukt:
Nice catch - noticed that yesterday as well and was going to patch, but you beat me to it, and more cleanly than I would have!

tACK.

dstrukt (approved):

PR #3960, opened 2022-07-12
Fix LN Address payouts

Kukks:
LN Address was validated when creating the claim but the paying sdection did not support it.

### DIFF
--- a/BTCPayServer/Extensions.cs
+++ b/BTCPayServer/Extensions.cs
@@ -10,9 +10,6 @@ using System.Security.Claims;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


BTCPayServer/Extensions.cs
```diff
@@ -41,6 +41,27 @@ namespace BTCPayServer
 {
     public static class Extensions
     {
+        //from https://stackoverflow.com/a/1374644/275504
+        public static bool IsValidEmail(this string email)
```
NicolasDorier, line 45: Replace with `return MailboxAddressValidator.TryParse(str, out var ma) && ma.ToString() == ma.Address;` I think it should work. System.Net.Mail is deprecated.

Issue #3597, opened 2022-03-31
Improve handling of `refunded` invoice in invoices, invoice details and invoice export.


pavlenex:
Currently, if an invoice is `refunded`:

- It's not visible in `invoice export` as `refunded` it's `settled` the amount is in +, which can lead to accounting problems.
- In `invoices view` you cannot distinguish refunded invoices, they say `settled` so it's hard for a merchant to pick a particular invoice that may need to be refunded or view a refunded one as they're all `settled`
- In the `invoice details` there's no field indicating an invoice is refunded in any way
- In invoice details there's no indicator of the refunded amount , and you can even click `issue a refund` on an invoice already refunded.

This is an example of an invoice detail of an already refunded invoice.
![Screenshot 2022-03-31 at 16 11 47](https://user-images.githubusercontent.com/36959754/161121562-7dc30dc1-d3aa-460a-8c4f-768479807c23.png)
This invoices view shows 3 invoices as settled, all of them `refunded`
![Screenshot 2022-03-31 at 20 08 43](https://user-images.githubusercontent.com/36959754/161121685-1332d7f9-4c8d-41fa-b568-f23fe1a8d62a.png)
And most problematic of all, this is an invoice export containing reporting for refunded invoices

![Screenshot 2022-03-31 at 20 10 30](https://user-images.githubusercontent.com/36959754/161122013-0cffc780-986e-4f7d-9dc0-8959ee1a32a2.png)

This issue can potentially be broken down into several issues, but for now we'll use it as a tracker

- [ ] Add refunded invoices in invoice export and show refunded amount
- [x] Add refunded `label` in invoices page view to distinguish easier refunded from settled.
- [x] Add indicator that invoice is refunded in invoice details and don't show `issue a refund` button, instead show `preview a refund`

Luckily we log in refunded in database, but we don't show it in the UI, show a fix for all of these shouldn't be that hard as @Kukks pointed out.
I had an idea of introducing `refunded` and `refunded-partial` status, but Kukks was against it.

Thanks to Bas for pointing majority of these problems, I'm just logging an issue.

rustywave:
Hi @Kukks , was advised that updating the export was the only thing remaining on this ticket. Found a couple items for clarification if you have time:

1. Refunds are list of Pull Payments. However, through the invoice detail page you can only perform a full or partial refund once. Should we anticipate multiple refunds in the future? Seems like we should.
[...]

PR #3968, opened 2022-07-13
Added refund fields to invoice export

rustywave:
Draft-only, will wait for feedback before proceeding.

Notes:
- Only first refund is included, if available. There can be multiple refunds, however. See ticket for clarification.
- .IncludeRefunds not added into InvoiceQuery call because does not filter the list
- Added as strings; there are a mix of number and string for currency fields. Can change if needed
- Added to end of list in order to not break client-side integrations
- If no refunds, field will be empty instead of 0 to indicate no refund present

image

#3597

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -971,7 +971,7 @@ namespace BTCPayServer.Controllers
         [BitpayAPIConstraint(false)]
         public async Task<IActionResult> Export(string format, string? searchTerm = null, int timezoneOffset = 0)
         {
-            var model = new InvoiceExport(_CurrencyNameTable);
+            var model = new InvoiceExport(_CurrencyNameTable, _paymentHostedService, _jsonSerializerSettings);
 
             InvoiceQuery invoiceQuery = GetInvoiceQuery(searchTerm, timezoneOffset);
             invoiceQuery.StoreId = new[] { GetCurrentStore().Id };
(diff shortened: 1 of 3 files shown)


Kukks (changes requested):
Good start, but:
* You need to also check the payouts inside the pull payment and see if the refund has been done. You can use 
```
var progress = pullPaymentHostedService.CalculatePullPaymentProgress(pp, DateTimeOffset.UtcNow);
 ```

 to get the amount refunded so far:  `var remaining = progress.Limit - progress.Completed - progress.Awaiting;`

rustywave:
Good call -- I actually did it that way because it looks like the invoice is tagged with `Refund` on the list view as soon as a refund is issued. Should I add extra fields in the export to account for any of this? If not, I will proceed with what you have above. Also, left some comments on the ticket that may or may not need clarification.

rustywave:
Okay I think I need to add the refunds as separate line items -- you can have multiple payments but currently only one refund per invoice. It kind of doesn't make sense to add the refund as a column because refunds are not tied to payments.  Will see what I can do here and reply back.

Kukks:
This is indeed problematic. I think we need to rethink this.
The main problem is that the export is not of invoices, but of payments, so
refunds may not entirely make sense on that context. Perhaps a better
solution is a new export specifically around invoices

rustywave:
I started down the road of adding refunds as a separate line item but then realized refunds have multiple payouts (from your suggestion)...I think it makes sense to list payouts out if we are listing payments out. But I agree with your point -- this isn't exactly an invoice export now and it's probably best to clarify what is desired for this.

PR #3975, opened 2022-07-15
Move TagHelpers to Abstractions

dennisreimann:
Makes them available for use in plugins. Also cleans up the tag helper references in the view code: As we have it in the root view imports, the individual directives in the views are superfluous.

PR #3977, opened 2022-07-16
Plugins: Add authorization hook

dennisreimann:
Makes the `PolicyRequirement` available to plugins.
Adds a filter hook to the authorization handlers, so that plugins can extend and leverage the existing authorization policies and permissions. Together with #3975 it also enables plugins to use the `permission` tag helper with custom policies.

Let's think a bit about how we want to name hooks and their argument classes. I opted for a verb-based approach for naming the filter (i.e. `handle-authorization-requirement`), because I think it describes the purpose of the hooks better (vs. something like `authorization-requirement-handler`). Wordpress seems to be inconsistent in that regard and they have a [mix of names](https://codex.wordpress.org/Plugin_API/Filter_Reference) for their filters.

For the argument class (which is also the type of the return value) I think it makes sense to have the type of hook in the name (i.e. `AuthorizationFilterHandle`).

Sidenote: In the long run I think we also need a way of documenting the hooks, their arguments and return values. Same goes for the UI extension points.


BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -144,14 +147,20 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
                     if (context.User != null)
                         success = true;
                     break;
+                default:
+                    if (Policies.IsPluginPolicy(requirement.Policy))
+                    {
+                        var args = new AuthorizationFilterArguments(context, requirement, _httpContext);
+                        success = (bool)await _pluginHookService.ApplyFilter("handle-authorization-requirement", args);
```
Kukks, line 154: Just checked the filter execution code and this will not work:
```
foreach (IPluginHookFilter pluginHookFilter in filters)
            {
                try
                {
                    args = await pluginHookFilter.Execute(args);
                }
                catch (Exception e)
                {
                    _logger.LogError(e, $"Filter on hook {hook} failed");
                }
            }
```
the filter will need to return the same type or else if another plugin tries to use it  it will fail

Kukks (changes requested):

BTCPayServer/Security/CookieAuthorizationHandler.cs
dennisreimann: Updated in [235ed1f](https://github.com/btcpayserver/btcpayserver/pull/3977/commits/235ed1f5396c1b6f28e7dc850c85da3171574ee9). I think it also makes for a good convention to pass the same type back and forth.

Kukks (approved):

PR #3978, opened 2022-07-17
Added custodian account trade support

woutersamaey:
- Custodian Account UI completely redone using Vue JS
- Trade support using a modal
- Dropdowns to change the "from" and "to" asset. Comes prefilled when you clicked the "trade" link for the asset.
- Price auto-updates every few seconds
- Easy access buttons to enter 10 ,25, 50, 75, 90 or 100% of assets to convert
- Easy button to swap assets from / to
- Indicator showing the estimated qty you will receive
- Indicator showing you the qty to remain after the trade
- During execution, the modal cannot be closed or canceled
- On trade success you get full ledger details of all the changes to your custodian account + your account holdings auto-refresh

Minor TODOs:
- Hiding small holdings / dust amounts would be nice
- Prevent the selection of an asset to trade, if you only have dust amount

**Account view**
Screenshot 2022-07-18 at 00 05 08

**Trade modal**
Screenshot 2022-07-18 at 00 05 20

**Asset selection, both from and to assets can be changed, incl a "swap left/right" button**
Screenshot 2022-07-18 at 00 11 46

**Trade is executing, loading spinner**
Screenshot 2022-07-18 at 00 05 44

**Trade completed**
Screenshot 2022-07-18 at 00 09 41

### DIFF
--- a/BTCPayServer/Controllers/UICustodianAccountsController.cs
+++ b/BTCPayServer/Controllers/UICustodianAccountsController.cs
@@ -471,8 +471,8 @@ namespace BTCPayServer.Controllers
             }
             catch (GreenfieldAPIException e)
             {
-                // TODO how to we return the same response as the original?
-                return BadRequest(e);
+                var result = new ObjectResult(e.APIError) { StatusCode = e.HttpCode };
+                return result;
             }
         }
 
--- a/BTCPayServer/Controllers/UICustodianAccountsController.cs
+++ b/BTCPayServer/Controllers/UICustodianAccountsController.cs
@@ -437,17 +437,12 @@ namespace BTCPayServer.Controllers
                             {
                                 var quote = await tradingCustodian.GetQuoteForAssetAsync(assetToTrade, assetToTradeInto,
                                     config, default);
-
-                                // Should we use "bid" or "ask"? Depends on the pair... are we buying or selling?
+                                
                                 // TODO Ask is normally a higher number than Bid!! Let's check this!! Maybe a Unit Test?
-                                vm.Price = quote.Ask;
-
-                                // vm.FormattedPrice =
-                                //     _currencyNameTable.DisplayFormatCurrency(vm.Price, assetToTrade);
-                                // vm.FormattedFiatValue =
-                                //     _currencyNameTable.DisplayFormatCurrency(pair.Value.Qty * quote.Bid,
-                                //         pair.Value.FiatAsset);
-                                // vm.FiatValue = vm.Qty * vm.Price;
+                                vm.Ask = quote.Ask;
+                                vm.Bid = quote.Bid;
+                                vm.FromAsset = quote.FromAsset;
+                                vm.ToAsset = quote.ToAsset;
                             }
                             catch (WrongTradingPairException)
                             {
--- a/BTCPayServer/Controllers/UICustodianAccountsController.cs
+++ b/BTCPayServer/Controllers/UICustodianAccountsController.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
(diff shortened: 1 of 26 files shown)


BTCPayServer/wwwroot/js/custodian-account.js
```diff
@@ -0,0 +1,270 @@
...
+            this.trade.isExecuting = true;
+
+            // Prevent the modal from closing by clicking outside or via the keyboard
+            this.modals.trade._config.backdrop = 'static';
+            this.modals.trade._config.keyboard = false;
+
+            let _this = this;
+
+            let token = $("input[name='__RequestVerificationToken']").val();
+            window.jQuery.ajax({
```
dennisreimann, line 126: Let's use the native [fetch](https://javascript.info/fetch) function instead of jQuery — at least for new features we're building. That'll allow us to hopefully one day migrate away from jQuery :)

woutersamaey: Funny part is that I actually used Fetch API in another file but the jQuery habbit must’ve taken over here 😂

Will fix this.

woutersamaey:
To test this PR, you will need my Kraken plugin (not included in this PR).
The mock custodian that we have does not have trade support and this would be very hard to mock properly.

I have a branch which is the same as this PR, but with the Kraken plugin added here: https://github.com/woutersamaey/btcpayserver/tree/kraken-plugin

dstrukt:
Love to see how this coming along!

Excited to test and share more thoughts when I get a moment.

dennisreimann:
The validation also seems to require round integer amounts, which isn't correct. And please don't ask why I have the Songbird token, they gave it to me … 😂

grafik

woutersamaey:
You're the frontend guru @dennisreimann. Your feedback would've taken me a long time to get right.

dennisreimann:
There should also be a confirmation for the delete action. I just recently did #3964, which will allow you to do it via modal as we do elsewhere.

woutersamaey:
@dennisreimann I just pushed 3 improvements:
- Input fields are now bigger. Hope you like my CSS, but unsure. It may be too primitive for your style.
- Input field validation is fixed, you can now trade amounts with decimals
- The Kraken plugin got an update and a critical fix (please note the submodule needs to be pulled + the project was moved from inside the `Plugins` folder to the root folder. Should not be an issue, but if it is, you know what is going on.)

dstrukt:
> Please accept my apology that these screens do not match your design work. Still, I'm very much open to your suggestions. I guess I was caught up so much that I just winged it :P

No worries at all. As I believe @dennisreimann already mentioned, we can always iterate and improve upon this great foundation.

BTCPayServer/Controllers/UICustodianAccountsController.cs
```diff
@@ -345,7 +344,7 @@ public async Task<IActionResult> DeleteCustodianAccount(string storeId, string a
             if (isDeleted)
             {
                 TempData[WellKnownTempData.SuccessMessage] = "Custodian account deleted";
-                return RedirectToAction("Dashboard", "UIStores", new { storeId });
+                return RedirectToAction(nameof(UIStoresController.Dashboard), "UIStores", new { storeId });
```
woutersamaey, line 347: Should we not use `nameof` for `UIStores` as well? I get why it's used, but why only 50% then?

dennisreimann: From what I've tried this won't work, because [`nameof` for controllers doesn't work without custom extensions](https://stackoverflow.com/questions/48558148/redirecttoaction-with-nameofnameofcontroller-fails-to-locate-the-action).

woutersamaey:
Rebased to master + Can now hide "small amounts" + Validation in trade modal.

TODO remaining:
- Try to hide scientific qty notation caused by Javascript when small numbers are encountered.
- Replace jQuery with Fetch API

woutersamaey:
@dennisreimann I think we've reached perfection now. Made some huge UX improvements. Too many small details to count especially when changing assets or swapping them. Sometimes the qty would remain or the price of combo wouldn't work out and you'd get a miniscule number in scientific notation. This has all been handled.

Can you review please, so we can get on with the next steps (withdrawing funds to store wallet)?

dennisreimann (approved):
Good work, @woutersamaey 👍

dstrukt:
Excited to test the latest this week!

woutersamaey:
@dstrukt if you want to test, please use my branch at https://github.com/woutersamaey/btcpayserver/tree/kraken-plugin

It’s the same code as this PR, but with the Kraken plugin included. You’ll need it. You’ll also need an API key for your Kraken account with full permissions.

AfolabiOlaoluwa:
I think this is incredible! An eye opener as to a better way to add exchange to non-custodial BTCPayServer. I love this pattern better than btctransmutter. Well done @woutersamaey.

PR #3979, opened 2022-07-18
Add support for updating POS app through Greenfield API

bolatovumar:
With this PR you will be able to update a POS app through Greenfield API.

Part of #3458

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -195,7 +195,7 @@ namespace BTCPayServer.Tests
 
         [Fact(Timeout = TestTimeout)]
         [Trait("Integration", "Integration")]
-        public async Task CanCreateReadAndDeletePointOfSaleApp()
+        public async Task CanCreateReadUpdateAndDeletePointOfSaleApp()
         {
             using var tester = CreateServerTester();
             await tester.StartAsync();
@@ -203,8 +203,58 @@ namespace BTCPayServer.Tests
             await user.RegisterDerivationSchemeAsync("BTC");
             var client = await user.CreateClient();
 
-            // Test creating a POS app
-            var app = await client.CreatePointOfSaleApp(user.StoreId, new CreatePointOfSaleAppRequest() { AppName = "test app from API"  });
+            // Test validation for creating the app
+            await AssertValidationError(new[] { "AppName" },
+                async () => await client.CreatePointOfSaleApp(user.StoreId, new CreatePointOfSaleAppRequest() {}));
+            await AssertValidationError(new[] { "AppName" },
+                async () => await client.CreatePointOfSaleApp(
+                    user.StoreId,
+                    new CreatePointOfSaleAppRequest()
+                    {
+                        AppName = "this is a really long app name this is a really long app name this is a really long app name",
+                    }
+                )
+            );
+            await AssertValidationError(new[] { "Currency" },
+                async () => await client.CreatePointOfSaleApp(
+                    user.StoreId,
+                    new CreatePointOfSaleAppRequest()
+                    {
+                        AppName = "good name",
+                        Currency = "fake currency"
+                    }
+                )
+            );
+            await AssertValidationError(new[] { "Template" },
+                async () => await client.CreatePointOfSaleApp(
+                    user.StoreId,
+                    new CreatePointOfSaleAppRequest()
+                    {
+                        AppName = "good name",
+                        Template = "lol invalid template"
+                    }
+                )
+            );
+            await AssertValidationError(new[] { "AppName", "Currency", "Template" },
+                async () => await client.CreatePointOfSaleApp(
+                    user.StoreId,
+                    new CreatePointOfSaleAppRequest()
+                    {
+                        Currency = "fake currency",
+                        Template = "lol invalid template"
+                    }
+                )
+            );
+
+            // Test creating a POS app successfully
+            var app = await client.CreatePointOfSaleApp(
+                user.StoreId,
+                new CreatePointOfSaleAppRequest()
+                {
+                    AppName = "test app from API",
+                    Currency = "JPY"
+                }
+            );
             Assert.Equal("test app from API", app.Name);
             Assert.Equal(user.StoreId, app.StoreId);
             Assert.Equal("PointOfSale", app.AppType);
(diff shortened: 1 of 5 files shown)


BTCPayServer.Client/Models/CreateAppRequest.cs
```diff
@@ -25,9 +25,9 @@ public class CreatePointOfSaleAppRequest : CreateAppRequest
         public string Template { get; set; } = null;
         [JsonConverter(typeof(StringEnumConverter))]
         public PosViewType DefaultView { get; set; }
-        public bool ShowCustomAmount { get; set; } = true;
-        public bool ShowDiscount { get; set; } = true;
-        public bool EnableTips { get; set; } = true;
+        public bool? ShowCustomAmount { get; set; } = null;
```
bolatovumar, line 28: Fallback in the code instead of here and mark these as optional as they are.

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -89,7 +89,7 @@ public async Task ApiKeysControllerTests()
             await user.MakeAdmin();
             var client = await user.CreateClient(Policies.CanViewProfile);
             var clientBasic = await user.CreateClient();
-            //Get current api key
```
bolatovumar, line 92: The whitespaces were auto-trimmed. Ignore all of these.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -22,33 +23,38 @@ public class GreenfieldAppsController : ControllerBase
...
             _storeRepository = storeRepository;
+            _currencies = currencies;
         }
 
         [HttpPost("~/api/v1/stores/{storeId}/apps/pos")]
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
         public async Task<IActionResult> CreatePointOfSaleApp(string storeId, CreatePointOfSaleAppRequest request)
         {
-            var validationResult = Validate(request);
+            var store = await _storeRepository.FindStore(storeId);
```
bolatovumar, line 45: Moved the store validation to the top since it doesn't make sense to first validate the request and only then validate that the store we are trying to create an app for exists, should be the other way around.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -60,11 +66,11 @@ public async Task<IActionResult> CreatePointOfSaleApp(string storeId, CreatePoin
...
-                ShowCustomAmount = request.ShowCustomAmount,
-                ShowDiscount = request.ShowDiscount,
-                EnableTips = request.EnableTips,
-                Currency = request.Currency ?? defaultCurrency,
-                Template = request.Template,
+                ShowCustomAmount = request.ShowCustomAmount ?? true,
+                ShowDiscount = request.ShowDiscount ?? true,
+                EnableTips = request.EnableTips ?? true,
+                Currency = request.Currency,
+                Template = request.Template != null ? _appService.SerializeTemplate(_appService.Parse(request.Template, request.Currency)) : null,
```
bolatovumar, line 73: Previously the template was saved raw which is wrong cuz you can end up saving an invalid template. I added validation for this as well now.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -74,18 +80,74 @@ public async Task<IActionResult> CreatePointOfSaleApp(string storeId, CreatePoin
...
+            var validationResult = ValidatePOSAppRequest(request);
+            if (validationResult != null)
+            {
+                return validationResult;
+            }
+
+            app.Name = request.AppName ?? app.Name;
+            app.SetSettings(new PointOfSaleSettings
+            {
+                Title = request.Title ?? settings.Title,
```
bolatovumar, line 115: Use the request properties or fallback to what exists on the app already if not provided.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -130,7 +192,35 @@ private PointOfSaleAppData ToModel(AppData appData)
             };
         }
 
-        private IActionResult? Validate(CreateAppRequest request)
+        private IActionResult? ValidatePOSAppRequest(CreatePointOfSaleAppRequest request)
+        {
+            var validationResult = ValidateCreateAppRequest(request);
+            if (request.Currency != null && _currencies.GetCurrencyData(request.Currency, false) == null)
```
bolatovumar, line 198: Added previously missing validation for currency and template validity.

Kukks (changes requested):
Please do a full UPDATE (Http PUT not PATCH) instead of selective property updating, as we have not defined this update mechanism for any other entity or endpoint.

NicolasDorier:
I think that it is fine. While it is strange to have our first PATCH, it doesn't mean that is bad.

NicolasDorier:
@bolatovumar I think this PATCH doesn't work as expected. Because you call `ValidateCreateAppRequest` which doesn't expect to stumble across null values.

NicolasDorier:
@bolatovumar I thought with it with @Kukks , I think that it should be PUT.

Not only because the PATCH has already some bug here, but because I believe PATCH for all routes with a GET and PUT can be created automatically the following way:

1. Do the json from the GET
2. Patch the json with the JObject of the PATCH
3. Do the PUT.

It can be generalized for every routes having both PUT and GET, without corner case to handle manually.

bolatovumar:
@Kukks @NicolasDorier alright, updated to be a `PUT` instead. Failing test is unrelated.

Capture

BTCPayServer/Controllers/GreenField/LocalBTCPayServerClient.cs
```diff
@@ -1103,6 +1103,14 @@ await GetController<GreenfieldPullPaymentController>()
                 await GetController<GreenfieldAppsController>().CreatePointOfSaleApp(storeId, request));
         }
 
+        public override async Task<PointOfSaleAppData> PutPointOfSaleApp(
```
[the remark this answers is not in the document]
bolatovumar: Updated in https://github.com/btcpayserver/btcpayserver/pull/3979/commits/0dfbc3725f47cae92ed9d1032ed30b0b60c6c1f3

BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
```diff
@@ -17,128 +17,83 @@
...
+                    "name": "appId",
+                    "in": "path",
+                    "required": true,
+                    "description": "App ID",
+                    "schema": {
+                        "type": "string"
+                    }
+                }
+            ],
+            "patch": {
```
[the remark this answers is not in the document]
bolatovumar: Updated in https://github.com/btcpayserver/btcpayserver/pull/3979/commits/a8d6f78955569f1e2f2352f0f24c06c60e0ece6a

Kukks (approved):

bolatovumar:
Failing test is still unrelated.

bolatovumar:
@NicolasDorier ping. I rebased the branch and now tests are green.

Kukks (approved):

PR #3982, opened 2022-07-19
Fix Local client

Kukks:
This cleans up the client factory for plugins so that it is less hectic looking. Additionally, it fixes a bug where if you reuse the factory after setting a store, the state might stick.


dennisreimann (approved):
tACK, tested it on top of my LNbank branch and it's working as intended :)

PR #3987, opened 2022-07-20
Make POS and Crowdfund plugins

dennisreimann:
Structural changes that turn the core apps into plugins, like we did with the Pay Button in #3862.


dstrukt:
*it'shappening.gif*

dennisreimann:
@Kukks I think we'd need to change the `AppType` to a string then, because the enum isn't extendible and there are some places that use the enum.

dennisreimann:
Back into draft mode, still need to convert the public views and parts of the app controllers.

Kukks (approved):
Code looks OK but it's quite a lot to go through.

I noticed that the apps rendered on the nav are with individual nav extensions per app. Maybe that is too much granularity? Maybe it should be one general nav extension and eachapp type registers an icon? N

Not sure, just trying to anticipate what we need to do to eventually allowing custom app types through non system plugins.

dennisreimann:
@Kukks the majority of changes is moved files, the rest of the changes  update the references to account for the movements. 

To allow for custom apps we'd have to change the app type from enum to string, so that plugins can extend that list. I think there are more changes that can be done, but moving the files was the first an major step that needed to happen, that's why I started with that.

PR #3993, opened 2022-07-21
Fix Crowdfund + additional issues

Kukks:
* Show free when invoice is fixed and 0 amount in apps
* animations on crowdfund not enabled broke JS
* Redirect to receipt after POS
* Show app link in receipt page
* Do not show 404 page in iframe when using crowdfund and a 0 amount invoice
* Perk expansion in crowdfund was broken

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.cs
@@ -87,14 +87,14 @@ namespace BTCPayServer.Controllers
 
         internal async Task<DataWrapper<InvoiceResponse>> CreateInvoiceCore(BitpayCreateInvoiceRequest invoice,
             StoreData store, string serverUrl, List<string>? additionalTags = null,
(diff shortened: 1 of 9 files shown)


BTCPayServer/Controllers/UIInvoiceController.cs
```diff
@@ -327,6 +327,10 @@ internal async Task<InvoiceEntity> CreateInvoiceCoreRaw(InvoiceEntity entity, St
                 entity.InternalTags.Add(AppService.GetAppInternalTag(app.Id));
             }
 
+            if (entityManipulator != null)
```
Kukks, line 330: this is a hack until we switch over to #3353

NicolasDorier:
please can you put a description to what you are actually fixing?

Kukks:
> please can you put a description to what you are actually fixing?

Ah sorry, the commit message didn't get added to the pull request

Kukks:
breaking into small prs

PR #3996, opened 2022-07-22
Fix 0 amount invoices with no payment methods crashing invoice ui

Kukks:


NicolasDorier:
I am not fan of how this corner case is forcing us to make sure `paymentMethodHandler` isn't null.

Can you write at least a test replicating this issue so I understand where it blows up?
I think it's not the first time where we have bug with this corner case.

dennisreimann:
Behaviour on current master (8806ba76eb25ee31cfab5ce284072b927c8c1c70): It doesn't crash, but immediately settles the invoice after creation:

```
info: Events:         Invoice JzugCLxCnnYPneKpSbSvkm new event: invoice_created (1001)
info: Events:         Invoice status is Settled
info: Events:         Invoice JzugCLxCnnYPneKpSbSvkm new event: invoice_paidInFull (1003)
info: Events:         Invoice JzugCLxCnnYPneKpSbSvkm new event: invoice_confirmed (1005)
info: Events:         Invoice JzugCLxCnnYPneKpSbSvkm new event: invoice_completed (1006)
info: Events:         Invoice JzugCLxCnnYPneKpSbSvkm is not monitored anymore.
```

Consequently, once the invoice details view is shown, there isn't even the option to checkout anymore, but to issue a refund — which doesn't make sense for a zero amount invoice that hasn't been paid.

grafik

It is the same across different payment methods being enabled as far as I see.

Kukks:
We should remove the refund option from the UI potentially.

The issue is when you forcefully go to the checkout url. (imagine cresting the invoice through api and using the checkout url to load it). This results in a 500 error or some vuejs errors

Issue #3573, opened 2022-03-24
Invoice filtering should match the `greenfield api` statuses.


pavlenex:
It's time that we address invoice filtering statuses.

Right now, when you click on invoice filters this is what you get:
![Screenshot 2022-03-24 at 21 38 02](https://user-images.githubusercontent.com/36959754/160009692-d7aa6753-2f12-4da5-834f-be9acbd9f3dc.png)

A few things:
- `Paid` status no longer exist, calling invoices `paid` is inaccurate.
-  `Paid`was replaced with `Settled`
- Drop down `paid` filters `settled` and `processing` 

Questions:
1. Should we just use terminology that 100% matches statuses and show that
2. Should we use collared labels in dropdown so people can easier distinguish statuses?
3. `Unusual` is weird but very useful, any other name for this one, its' not a status per se, but a collection of statuses that require attention?

This issue requires collaborative effort from back-end and design teams to come up with technically accurate but easy to use solutions that doesn't introduce complex terminology and innacurate naming. pinging @NicolasDorier @dennisreimann @dstrukt @Kukks

bolatovumar:
> Should we just use terminology that 100% matches statuses and show that

I think this makes sense. It's especially weird when you click on "Paid" and see "Settled" and "Processing" in the "Status" column while this banner shows up above. It's not too bad though and people would probably make the connection between these.

![Capture](https://user-images.githubusercontent.com/1934678/160264844-df4c6a80-5d08-410d-986f-fd13b1241791.PNG)

> Should we use collared labels in dropdown so people can easier distinguish statuses?

Some statuses map to labels of different colors. "Paid over" maps to green "Settled (paid over)" and red "Invalid (paid over)"

> Unusual is weird but very useful, any other name for this one, its' not a status per se, but a collection of statuses that require attention?

I kind of like "Unusual". Another name that could apply here is "Unexpected".

pavlenex:
@bolatovumar I think we should use the coloured statuses, it kinda communicates well if something is okay, wait or bad, so I'd match those same way we colorize them in the invoices page.

I guess you're right about the `unusual` I was just thinking if there's a better term like `Action needed` or something implying these invocies in most cases need to be addressed or require a reaction from a merchant.

phershbe:
I can work on this if you still need somebody to!

bolatovumar:
[...]

PR #4000, opened 2022-07-26
Update ListInvoices to say Settled instead of Paid

BitcoinABee:
Updated ListInvoices to say Settled instead of Paid

This is a PR to fix https://github.com/btcpayserver/btcpayserver/issues/3573 . @pavlenex @Kukks @NicolasDorier @bolatovumar, please review and let me know if you have any feedback.

As discussed in the thread on the issue and per our call few months back, we are not making any color changes. We are only changing Paid to Settled as Paid no longer exists as a status.


pavlenex:
Wooah 4000th PR! Congrats @BitcoinABee 🥳 here's to many more! 🥂

PR #4013, opened 2022-08-02
Add ability to mark auth handle as successful

dennisreimann:
Without this, there is no way to let the handle finish with a successful state. I somehow missed to add this in #3977.


Kukks (changes requested):
Tests fail

dennisreimann:
@Kukks Yes, but those test failures are unrelated (`CanUseTorClient` again, @NicolasDorier) and the link check.

Kukks (approved):

PR #4014, opened 2022-08-02
Support accepting 0 amount bolt 11 invoices for payouts

Kukks:
This PR adds support for specifying an explicit amount for a payout that is different from the bolt11 amount provided, as long as the explicit amount is higher than the bolt11 amount.


pavlenex:
Should we accept 0-amount for our zero-amount invoices as well in the future?

Kukks:
> Should we add bolt11 zero-amount for our zero-amount invoices as well in the future as well, seems like a reasonable use-case, though I haven't fully understood how a zero-amount would work for a payout?

Right now, a payout asks you for both a destination and an amount. The amount can be optional if the destination has the amount included. This PR makes the opposite true for lightning (the destination does not need an amount if the amount was provided, aka 0 amount bolt11 invoices).

Personally don't think it's very good ux right now, as most wallets handle 0 amount invoices very differently and it's not always obvious you can edit/change/should input the amount.

NicolasDorier:
If Bolt11 UX for 0 amount sucks, I guess we better stay on LNURL. I don't think that's a good idea to propose an option for this. This is so niche very few people will understand it, and nobody ever requested it.

dennisreimann:
We could so it similarly as with the BIP21 switch, evaluate the support and keep track of it in a little matrix. We can have a better informed decision and maybe delay it until it's good enough. I'll come up with the matrix and will post it here.

dennisreimann:
These wallets support it and let the user simply input a custom amount:

- Zeus
- BlueWallet
- Wallet of Satoshi
- Bitcoin Beach Wallet
- Phoenix
- Blixt

Not supported in:

- Alby (https://github.com/getAlby/lightning-browser-extension/issues/823)

Untested:

- Breez
- Muun
- Bitkit
- Simple Bitcoin Wallet

BTCPayServer/HostedServices/PullPaymentHostedService.cs
```diff
@@ -830,6 +830,31 @@ public static string GetErrorMessage(PayoutPaidResult result)
...
+                null when destination.Amount != null => (null,destination.Amount),
+                not null when destination.Amount is null => (null,amount),
+                not null when destination.Amount != null && amount != destination.Amount &&
+                              destination.IsExplicitAmountMinimum &&
+                              payoutCurrency == "BTC" && ppCurrency == "SATS" &&
+                              new Money(amount, MoneyUnit.Satoshi).ToUnit(MoneyUnit.BTC) < destination.Amount =>
+                    ($"Amount is implied in both destination ({destination.Amount}) and payout request ({amount}), but the payout request amount is less than the destination amount",null),
+                not null when destination.Amount != null && amount != destination.Amount &&
+                              destination.IsExplicitAmountMinimum &&
+                              payoutCurrency != "BTC" && ppCurrency != "SATS" &&
```
Kukks, line 848:
```suggestion
                              !(payoutCurrency == "BTC" && ppCurrency == "SATS") &&
```

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
```diff
@@ -264,7 +264,7 @@ public async Task<IActionResult> ProcessLightningPayout(string cryptoCode, strin
             PaymentMethodId pmi, CancellationToken cancellationToken)
         {
             var boltAmount = bolt11PaymentRequest.MinimumAmount.ToDecimal(LightMoneyUnit.BTC);
-            if (boltAmount != payoutBlob.CryptoAmount)
+            if (boltAmount != payoutBlob.CryptoAmount && boltAmount > payoutBlob.CryptoAmount)
```
Kukks, line 267:
```suggestion
            if (boltAmount > payoutBlob.CryptoAmount)
```

Issue #3878, opened 2022-06-16
Edit pull payment


dennisreimann:
Working on #3873 I noticed that we do not have an edit view for pull payments. Though the amount might be fixed after creation, we now have some more fields like title and description, which should be editable afterwards.

dstrukt:
ACK.

daviogg:
Can I try to implement this issue?

dennisreimann:
Sure, go ahead and let me know if you have any questions. You can get in touch via [Mattermost](https://chat.btcpayserver.org).

HamroRamro:
Someone please close this issue since it has been merged in https://github.com/btcpayserver/btcpayserver/pull/4016

PR #4016, opened 2022-08-03
Edit view for Pull Payments

daviogg:
https://github.com/btcpayserver/btcpayserver/issues/3878

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -30,7 +30,6 @@ using Newtonsoft.Json.Linq;
 using OpenQA.Selenium;
 using OpenQA.Selenium.Support.Extensions;
 using OpenQA.Selenium.Support.UI;
-using Renci.SshNet.Security.Cryptography;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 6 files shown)


dennisreimann:
Good work and thanks for the contribution!

I've added [three more commits on my fork](https://github.com/dennisreimann/btcpayserver/commits/enhancement/edit-pull-payment) for things I came across while reviewing. Normally I'd note them in the suggestions, but some of it was a bit more involved, so I figured it might be best to do the changes as separate commits:

- To keep the store context, we need to have the `storeId` in the edit URL — otherwise the navigation will be empty for the edit page.
- The links and redirects got edited to match the behaviour of the payment requests

You can either add me as a collaborator to your fork or integrate my changes by cherry-picking the commits. With the former I could just push them here — whatever way you prefer.

Thanks again!

dennisreimann (approved):
tACK.

@daviogg Can you test it with the changes I did? If you want to, you can also add a simple Selenium test that checks the editing of the description for example.

daviogg:
The test it is to be placed within the file `SeleniumTests.cs` right? Or we need to create a specific class for PullPayments tests?

dennisreimann:
You van put it into an existing case if that exists. Or add a new one in the Selenium Tests.

daviogg:
I did selenium tests but some tests are broken, seems external to this PR

dennisreimann:
Yes, the test failures are again unrelated. Thanks for providing the test case! Good to go now :)

Kukks (approved):

PR #4022, opened 2022-08-05
Greenfield: Invoices API without Store path

Kukks:


NicolasDorier:
Do not merge it in the short term, I think this has some none obvious security implication, I need to test it roughly.

PR #4024, opened 2022-08-05
Custodian Account Deposit UI

woutersamaey:
UI to make deposits to a custodian account.
- Deposits are made using a modal
- Select the asset, then the payment method
- A QR code appears just like in "Wallet Receive"
- QR code generation was a bit refactored (moved  tag generation to a separate service so it can be used in a controller without view.
- Only OnChain is supported for now, but that's normal as the Custodian API is also OnChain only for now.
- Button "Create Transaction" which takes you to the "Send" page with the address prefilled + cancel takes you back to the custodian account overview page as you'd expect

To test, please use my other branch which is the same thing as this one, except it has the Kraken plugin pre-installed: https://github.com/woutersamaey/btcpayserver/tree/custodian-deposit-support-with-kraken

Some screenshots:

*Existing overview page*
image

*Deposit modal*
image


woutersamaey:
@dennisreimann and @NicolasDorier please review so I can continue with withdrawal support.

@NicolasDorier just FYI I noticed 1 strange thing after refactoring the QR Code generation logic. The "Send" page from my Store Wallet did not show a QR code at first. Only after I clicked "Generate next address" the QR code appeared. After that, it did work and I could not reproduce. I did check the code extensively and looked at other files in `Components` but didn't see anything that could be wrong.

During a debugging session I did see that the constructor was not called the first time and only after clicking the button "Generate next address" was it called and running the Component, which is very odd.

It may be nothing, but I thought I'd mention it so the "Send" page doesn't break unexpectedly. If you do run into this issue, please advise. I did only very simple, basic refactoring, so if I made a mistake, you'll probably know better than me what to do.

woutersamaey:
Tests failed because of dead link (not my fault)

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
```diff
@@ -15,333 +20,390 @@
...
+                <h5 class="modal-title">Deposit</h5>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close">
+                    <vc:icon symbol="close"/>
+                </button>
+            </div>
+            <div class="modal-body">
+                <p v-if="deposit.errorMsg" class="alert alert-danger">{{ deposit.errorMsg }}</p>
+                <div class="form-group">
+                    <label class="form-label" for="DepositAsset">Asset to Deposit</label>
+                    <select class="form-control" v-model="deposit.asset" name="DepositAsset">
```
dennisreimann, line 164:
```suggestion
                    <select class="form-select" v-model="deposit.asset" name="DepositAsset">
```

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
```diff
@@ -15,333 +20,390 @@
...
+                        <option v-for="option in availableAssetsToDeposit" v-bind:value="option">
+                            {{ option }}
+                        </option>
+                    </select>
                 </div>
-                <div class="modal-body">
-                    <p>Deposits are coming soon, but if you need this today, you can use our <a rel="noopener noreferrer" href="https://docs.btcpayserver.org/API/Greenfield/v1/" target="_blank">Greenfield API "Get a deposit address for custodian" endpoint</a> to get a deposit address to send your assets to.</p>
+                <div class="form-group">
+                    <label class="form-label" for="DepositPaymentNetwork">Payment Method</label>
+                    <select class="form-control" v-model="deposit.paymentMethod" name="DepositPaymentNetwork">
```
dennisreimann, line 172:
```suggestion
                    <select class="form-select" v-model="deposit.paymentMethod" name="DepositPaymentNetwork">
```

dennisreimann (commented):
One more thing I noticed: When the API key is expired, the loading indicator keeps spinning and the error isn't reported. In my case I'm running into an `account.depositablePaymentMethods is null` in the dev console. The same also happens for the fake custodian.
But I think you should check whether or not the API key is still working in general.

For me the wallet receive page is still behaving normally, so I didn't notice anything breaking there.

woutersamaey:
> One more thing I noticed: When the API key is expired, the loading indicator keeps spinning and the error isn't reported. In my case I'm running into an `account.depositablePaymentMethods is null` in the dev console. The same also happens for the fake custodian. But I think you should check whether or not the API key is still working in general.

This issue has been fixed. Just needed to check for `NULL`.
Your expired API key may be unrelated to this issue, because the list of what assets are depositable is public and not the result of a private API call.

BTCPayServer/Services/QRCodeImgTagGenerator.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using QRCoder;
+
+namespace BTCPayServer.Services;
+
+public class QRCodeImgTagGenerator
+{
+    private static QRCodeGenerator qrGenerator = new();
+    
+    public string generateImgTag(string data)
```
Kukks, line 10:
```suggestion
    public string GenerateImgTag(string data)
```
Pascal case

BTCPayServer/Models/CustodianAccountViewModels/DepositPrepareViewModel.cs
```diff
@@ -0,0 +1,12 @@
+namespace BTCPayServer.Models.CustodianAccountViewModels;
+
+public class DepositPrepareViewModel
+{
+    public string PaymentMethod { get; set; }
+    public string Address { get; set; }
+    public string AddressQRHtml { get; set; }
+    public string Link { get; set; }
+    public string LinkQRHtml { get; set; }
```
Kukks, line 9: Let's not start passing HTML through API calls, this is a horrible practice from back in the jquery mayhem days.

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
```diff
@@ -15,333 +20,390 @@
...
+                    </select>
                 </div>
-                <div class="modal-footer">
-                    <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+                <div v-if="deposit.link || deposit.address">
+                    <div class="tab-content text-center">
+                        <div v-if="deposit.link" class="tab-pane" id="link-tab" role="tabpanel">
+                            <div class="qr-container mb-3">
+                                <img :src="deposit.cryptoImageUrl" class="qr-icon" :alt="deposit.asset"/>
+                                <div v-html="deposit.linkQRHtml"></div>
```
Kukks, line 183: we have a dedicated Vue component that can render a qr code from the link

BTCPayServer/Controllers/UICustodianAccountsController.cs
```diff
@@ -476,6 +477,90 @@ private JObject RemoveUnusedFieldsFromConfig(JObject storedData, JObject newData
             }
         }
 
+        [HttpGet("/stores/{storeId}/custodian-accounts/{accountId}/deposit/prepare")]
+        public async Task<IActionResult> GetDepositPrepareAjax(string storeId, string accountId,
```
Kukks, line 481: Don't name things with "Ajax"

BTCPayServer/Controllers/UICustodianAccountsController.cs
```diff
@@ -476,6 +477,90 @@ private JObject RemoveUnusedFieldsFromConfig(JObject storedData, JObject newData
...
+                    }
+
+                    try
+                    {
+                        var depositAddressResult =
+                            await depositableCustodian.GetDepositAddressAsync(paymentMethod, config, default);
+                        vm.Address = depositAddressResult.Address;
+                        vm.AddressQRHtml = _qrCodeImgTagGenerator.generateImgTag(vm.Address);
+                        
+                        if (paymentMethod.Equals("BTC-OnChain"))
```
Kukks, line 523: Please use the `PaymentMethodId` class instead of hardcoded strings to determine things like this.

Kukks (changes requested):
Please revert the QR code changes, and use the Vue QR component instead.

BTCPayServer/Controllers/UICustodianAccountsController.cs
woutersamaey: Done

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
woutersamaey: Good point. Done.

BTCPayServer/Models/CustodianAccountViewModels/DepositPrepareViewModel.cs
woutersamaey: Done.

BTCPayServer/Services/QRCodeImgTagGenerator.cs
woutersamaey: This file no longer exists since we're using the VueJS QR component

BTCPayServer/Controllers/UICustodianAccountsController.cs
woutersamaey: Good point. Done

woutersamaey:
@Kukks I have addressed all your feedback + Also added a loading spinner while the deposit address is being loaded.
Can I get your blessing now please?

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
```diff
@@ -15,333 +20,398 @@
...
-                                <button v-on:click="setTradeQtyPercent(25)" class="btn btn-secondary" type="button">25%</button>
-                                <button v-on:click="setTradeQtyPercent(50)" class="btn btn-secondary" type="button">50%</button>
-                                <button v-on:click="setTradeQtyPercent(75)" class="btn btn-secondary" type="button">75%</button>
-                                <button v-on:click="setTradeQtyPercent(90)" class="btn btn-secondary" type="button">90%</button>
-                                <button v-on:click="setTradeQtyPercent(100)" class="btn btn-secondary" type="button">100%</button>
-                            </div>
+                                    <input name="Qty" type="number" min="0" step="any" :max="trade.maxQtyToTrade" :min="getMinQtyToTrade()" class="form-control qty" v-bind:class="{ 'is-invalid': trade.qty < getMinQtyToTrade() || trade.qty > trade.maxQtyToTrade }" v-model="trade.qty"/>
+                                    <select name="FromAsset" v-model="trade.assetToTrade" class="form-control">
```
dennisreimann, line 263:
```suggestion
                                    <select name="FromAsset" v-model="trade.assetToTrade" class="form-select">
```

BTCPayServer/Views/UICustodianAccounts/ViewCustodianAccount.cshtml
```diff
@@ -15,333 +20,398 @@
...
+                            <button v-if="canSwapTradeAssets()" type="button" class="btn btn-secondary btn-square" v-on:click="swapTradeAssets()" aria-label="Swap assets">
+                                <i class="fa fa-arrows-h" aria-hidden="true"></i>
+                            </button>
+                        </div>
+                        <div class="col-side">
+                            <label class="form-label">
+                                Into
+                                <div class="input-group">
+                                    <input disabled="disabled" type="number" class="form-control qty" v-model="tradeQtyToReceive"/>
+                                    <select name="ToAsset" v-model="trade.assetToTradeInto" class="form-control">
```
dennisreimann, line 283:
```suggestion
                                    <select name="ToAsset" v-model="trade.assetToTradeInto" class="form-select">
```

BTCPayServer/wwwroot/js/custodian-account.js
```diff
@@ -62,6 +75,40 @@ new Vue({
...
+                        r.push(pm);
+                    }
+                }
+            }
+            return r.sort();
+        },
+        createDepositTransactionUrl: function () {
+            if (this.deposit.address && this.deposit.paymentMethod) {
+                // TODO support LN
+                return '/wallets/S-' + this.account.storeId + '-' + this.deposit.asset + '/send?defaultDestination=' + encodeURI(this.deposit.address)+"&returnUrl="+encodeURI(window.location.href);
```
NicolasDorier, line 108: Never build URLs like this. Instead, generate it server side using the `LinkGenerator`, and pass them to the front.

Search `GetUriByAction` to see how to use it.

woutersamaey: You’re right. Silly mistake. Will fix asap.

woutersamaey: Fixed it.

woutersamaey:
@dennisreimann I switched all selects to `form-select` + fixed the width since this class is wider than before. I made it so that tickers with 4 characters can fit.

woutersamaey:
@dennisreimann @Kukks @NicolasDorier all feedback has been addresses and fixed. Can I get your final blessing pls?

Kukks (approved):

PR #4029, opened 2022-08-09
Change PriceProvider(Mona)

wakiyamap:
I want to change the price rate supply from bittrex to bitbank.
Bitbank has more volume in monacoin.


wakiyamap:
Bitbank will close once to abolish BTC pairs.
https://blog.bitbank.cc/614293404/

PR #4031, opened 2022-08-10
Allow bind and port for http too

Kukks:


dennisreimann (commented):
tACK. Works for me with both HTTP and HTTPS.

PR #4032, opened 2022-08-11
Refactor Payouts

Kukks:


dennisreimann (commented):
For HOLD invoices, I think this still needs more work. I haven't tried it yet, but retrying timed out payouts via `TrypayBolt` (which uses `lightningClient.Pay`) would not work: Even though the invoice isn't settled/cancelled, it is regarded as paid already and our client returns with `TotalAmount == null` (see [here](https://github.com/dennisreimann/btcpayserver/blob/plugins/BTCPayServer.Plugins.LNbank/Services/Wallets/WalletService.cs#L202)).

So in case this does not work, the correct way would be to use `lightningClient.GetPayment` and check that result, as this has the correct and explicit `LightningPaymentStatus`.

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
```diff
@@ -286,14 +315,21 @@ public static async Task<ResultVM> TrypayBolt(ILightningClient lightningClient,
...
-            return new ResultVM
+            catch (Exception ex) when (ex is TaskCanceledException or OperationCanceledException)
             {
-                PayoutId = payoutData.Id,
-                Result = result.Result,
-                Destination = payoutBlob.Destination,
-                Message = result.ErrorDetail
-            };
+                // Timeout, potentially caused by hold invoices
+                // Payment will be saved as pending, the LightningInvoiceWatcher will handle settling/cancelling
```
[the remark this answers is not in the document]
Kukks: It is a separate thing at https://github.com/btcpayserver/btcpayserver/pull/4032/files#diff-2ccdb459526f43e79d32da64e7cf37de2ba176cd0abb0422c326202d5c36e808

Kukks:
> For HOLD invoices, I think this still needs more work. I haven't tried it yet, but retrying timed out payouts via `TrypayBolt` (which uses `lightningClient.Pay`) would not work: Even though the invoice isn't settled/cancelled, it is regarded as paid already and our client returns with `TotalAmount == null` (see [here](https://github.com/dennisreimann/btcpayserver/blob/plugins/BTCPayServer.Plugins.LNbank/Services/Wallets/WalletService.cs?rgh-link-date=2022-08-11T14%3A08%3A58Z#L202)).
> 
> So in case this does not work, the correct way would be to use `lightningClient.GetPayment` and check that result, as this has the correct and explicit `LightningPaymentStatus`.

It does not retry, it sets the payout in the "in progress" state. `LightningPendingPayoutListener` will routinely go through all LN in progress payouts, and attempt to call `GetPayment` on it to see if there is a response

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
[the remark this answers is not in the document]
dennisreimann: Alright, nice, I didn't catch that. That also answers my other comment.

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
```diff
@@ -286,14 +315,21 @@ public static async Task<ResultVM> TrypayBolt(ILightningClient lightningClient,
...
+                
+                payoutData.SetProofBlob(proofBlob, null);
+                return new ResultVM
+                {
+                    PayoutId = payoutData.Id,
+                    Result = PayResult.Ok,
+                    Destination = payoutBlob.Destination,
+                    Message = "The payment timed out. We will verify if it completed later."
+                };
+            }
```
dennisreimann, line 332: What about the other errors that might occur, like not being able to find a route?

dennisreimann, line 332: What about the other errors that might occur, like not being able to find a route?

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
dennisreimann, line 321:
```suggestion
                // Payment will be saved as pending, the LightningPendingPayoutListener will handle settling/cancelling
```

BTCPayServer/Payments/Lightning/LightningPendingPayoutListener.cs
```diff
@@ -0,0 +1,141 @@
...
+                                    payoutData.State = PayoutState.Completed;
+                                    payoutLightningBlob.Preimage = payment.Preimage;
+                                    payoutData.SetProofBlob(payoutLightningBlob, null);
+                                    break;
+                                case LightningPaymentStatus.Failed:
+                                    payoutData.State = PayoutState.Cancelled;
+                                    break;
+                                default:
+                                    payoutData.State = payoutData.State;
+                                    break;
```
dennisreimann, line 123: Why this?

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -2455,6 +2456,50 @@ public async Task DisabledEnabledUserTests()
...
+                });
+            Assert.Empty(await adminClient.GetStoreLightningAutomatedPayoutProcessors(admin.StoreId, "BTC_LightningNetwork"));
+            await adminClient.UpdateStoreLightningAutomatedPayoutProcessors(admin.StoreId, "BTC_LightningNetwork",
+                new LightningAutomatedPayoutSettings() {IntervalSeconds = TimeSpan.FromSeconds(2)});
+            Assert.Equal(2, Assert.Single( await adminClient.GetStoreLightningAutomatedPayoutProcessors(admin.StoreId, "BTC_LightningNetwork")).IntervalSeconds.TotalSeconds);
+            await TestUtils.EventuallyAsync(async () =>
+            {
+                var payoutC =
+                    (await adminClient.GetStorePayouts(admin.StoreId, false)).Single(data => data.Id == payout.Id);
+                Assert.Equal(payoutC.State, PayoutState.Completed );
```
dennisreimann, line 2499:
```suggestion
                Assert.Equal(PayoutState.Completed, payoutC.State);
```

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
Kukks: should be handled elsehwere

dennisreimann (approved):
tACK. Tested both regular and hold invoices. 

From my experience debugging those background tasks in LNbank I'd advise to add some logging. This will be very valuable when resolving issues.

One minor thing I noticed: The `SendTimeout` variable could be made `private`. The ten minute delay seems reasonable.

Note: We can later on surface the proof in the UI.

PR #4035, opened 2022-08-13
API docs: Fix duplicate speed policy schema

dennisreimann:
See the additional discussion in #3877. As we merged the Swagger JSON files for the docs, the duplicate store speed policy schema overwrote the updated invoice speed policy schema. This is now fixed by using a unified schema.

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.invoices.json
@@ -1089,22 +1089,6 @@
                     }
                 }
             },
-            "SpeedPolicy": {
-                "type": "string",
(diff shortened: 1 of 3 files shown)


ndeet (approved):
Wow, very good catch. Makes total sense to define it only once in a central place. Interesting that it only overwrote for the docs.btcpayserver.org page but not for the docs on the BTCPay Server instance - anyway great work thanks a lot!

dennisreimann:
The way it is handled in the app is more dynamic. For the docs we just merge all the JSON files into a single file, so the last one depending on the filename wins. 

However, this hints at another thing I noticed when I tried to validate the Swagger files a while ago: we can make more improvements and in general I think we'd benefit from having a validation process for those files.

I have a branch for this already, need to polish it though.

Kukks (approved):

Issue #4030, opened 2022-08-10
Invoice creation error re-ordering


Kukks:
When an invoice is being created using a specific payment method, and the rate fails, we are returning an incorrect error message. 
Currently, we return the wallet is not configured, when we should return that the rate is not available.

_Issue created from a [Mattermost message](https://chat.btcpayserver.org/_redirect/pl/5567eazpsi8h9dd48ooo58tejc) *by `@gauravp`*._

HamroRamro:
https://github.com/btcpayserver/btcpayserver/pull/4039 <- PR for your review, sir @Kukks . 

cc: @pavlenex

HamroRamro:
Good to close since its merged now. @Kukks @pavlenex @ndeet

PR #4039, opened 2022-08-15
Add no rate found error message in Invoices

HamroRamro:
Hello,

This fixes the issue raised by @Kukks in https://github.com/btcpayserver/btcpayserver/issues/4030. 

I noticed that there is a checkbox is the UI to show supported payment methods so else statement can only say "rate is not available" and point the end user to check the logs. 

The "if" above the changed else takes care of instance where there are no supported payment methods configured.


Kukks (changes requested):
Hmm, while your error message is correct, the original error message is also correct. Can you combine them?

HamroRamro:
> Hmm, while your error message is correct, the original error message is also correct. Can you combine them?

Done. Can you please review?

Kukks (approved):

PR #4041, opened 2022-08-15
Fix swagger validation errors

dennisreimann:
Combined the files with the same approach as in the docs:

`jq -rs 'reduce .[] as $item ({}; . * $item)' swagger.template.* > openapi.json`

Afterwards the bundled version can be validated using these commands:

`npx swagger-cli validate openapi.json && npx @redocly/cli lint openapi.json`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -300,3 +300,5 @@ BTCPayServer/testpwd
 .DS_Store
 Packed Plugins
 Plugins/packed
+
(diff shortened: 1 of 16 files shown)


dennisreimann:
@kukks I revised my previous approach, to for now keep the files as they are, bundle them using `jq` and solely focus on fixing the validation errors in this PR. We can extend this and integrate it into the build, but for this PR my goal was to clean up first.

Kukks (approved):

PR #4048, opened 2022-08-18
Add Yadio exchange rate provider

bolatovumar:
Yadio is a service which provides exchange rates for many currencies such as the Lebanese Pound which have the real exchange rate which differs from the official rate significantly (32,525 LBP for USD real rate vs 1,510 official rate).

See more details here: https://yadio.io/info.html

See discussions here:
https://github.com/btcpayserver/btcpayserver/discussions/4001
https://github.com/btcpayserver/btcpayserver/discussions/2489#discussioncomment-3102370

Notice that this screenshot shows 766,245,075 LBP for 1 BTC when the using the official rate it's 35,366,556. It's a 21x difference.

Capture

### DIFF
--- a/BTCPayServer.Rating/Providers/YadioRateProvider.cs
+++ b/BTCPayServer.Rating/Providers/YadioRateProvider.cs
@@ -1,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


Kukks (approved):
code looks good, just will wait for the tests to pass. 

Just in time for the next minor release :)

PR #4065, opened 2022-08-20
Allow specifing fee block target for onchain payout processor

Kukks:
resolves #4050

### DIFF
--- a/BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutBlob.cs
+++ b/BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutBlob.cs
@@ -4,5 +4,5 @@ namespace BTCPayServer.PayoutProcessors.OnChain;
 
 public class OnChainAutomatedPayoutBlob : AutomatedPayoutBlob
 {
-    public int FeeTargetBlock { get; set; }
+    public int FeeTargetBlock { get; set; } = 1;
 }
--- a/BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutBlob.cs
+++ b/BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutBlob.cs
@@ -1,0 +1,8 @@
+﻿using BTCPayServer.PayoutProcessors.Settings;
+
+namespace BTCPayServer.PayoutProcessors.OnChain;
+
+public class OnChainAutomatedPayoutBlob : AutomatedPayoutBlob
+{
+    public int FeeTargetBlock { get; set; }
+}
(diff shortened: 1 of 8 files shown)


BTCPayServer/Views/UIOnChainAutomatedPayoutProcessors/Configure.cshtml
```diff
@@ -22,6 +22,13 @@
                     <input asp-for="IntervalMinutes" class="form-control" inputmode="numeric" style="max-width:10ch;">
                     <span class="input-group-text">minutes</span>
                 </div>
+            </div> 
+            <div class="form-group">
+                <label asp-for="FeeTargetBlock" class="form-label" data-required>Fee block target</label>
+                <div class="input-group">
+                    <input asp-for="FeeTargetBlock" class="form-control" min="1" inputmode="numeric" style="max-width:10ch;">
```
bolatovumar, line 29: Should there also be a max specified here?

BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutProcessor.cs
```diff
@@ -89,8 +89,9 @@ protected override async Task Process(ISupportedPaymentMethod paymentMethod, Lis
             decimal? failedAmount = null;
             var changeAddress = await explorerClient.GetUnusedAsync(
                 storePaymentMethod.AccountDerivation, DerivationFeature.Change, 0, true);
-
-            var feeRate = await explorerClient.GetFeeRateAsync(1, new FeeRate(1m));
+            
+            var processorBlob = GetBlob(_PayoutProcesserSettings);
+            var feeRate = await explorerClient.GetFeeRateAsync(processorBlob.FeeTargetBlock,new FeeRate(1m));
```
bolatovumar, line 94: Should this fallback to 1 just in case?

bolatovumar (approved):
Failing test seems to be unrelated.

BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutProcessor.cs
Kukks: It does, the last param is a falllback

BTCPayServer/PayoutProcessors/OnChain/OnChainAutomatedPayoutBlob.cs
```diff
@@ -0,0 +1,8 @@
+﻿using BTCPayServer.PayoutProcessors.Settings;
+
+namespace BTCPayServer.PayoutProcessors.OnChain;
+
+public class OnChainAutomatedPayoutBlob : AutomatedPayoutBlob
+{
+    public int FeeTargetBlock { get; set; }
```
dennisreimann, line 7:
```suggestion
    public int FeeTargetBlock { get; set; } = 1;
```

dennisreimann: With this the default is applied when the setup form is opened. As it is the default value I think it makes sense to have it in there already, otherwise it's a zero, which prevents the form from being submitted, because of an invalid value (`min=1`).

dennisreimann (approved):
tACK

PR #4066, opened 2022-08-20
Re-do onchain label system

Kukks:
Currently, we can only place labels on transactions, resulting in an inferior coin selection system.  This PR aims to re-implement so that labels can be placed on both transactions AND on scripts (addresses).

This means that when you generate an address, you can label it with who you shared it with, and when you send out money, you can also label their address. 

DESCRIPTION NOT COMPLETE. STILL WIP

### DIFF
--- a/BTCPayServer.Data/Data/WalletTransactionData.cs
+++ b/BTCPayServer.Data/Data/WalletTransactionData.cs
@@ -14,6 +14,7 @@ namespace BTCPayServer.Data
         public byte[] Blob { get; set; }
 
         public List<WalletLabelData> WalletLabels { get; set; }
+        public List<WalletScriptData> WalletScripts { get; set; }
 
         internal static void OnModelCreating(ModelBuilder builder)
         {
--- a/BTCPayServer.Data/Data/WalletTransactionData.cs
+++ b/BTCPayServer.Data/Data/WalletTransactionData.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 
 namespace BTCPayServer.Data
@@ -7,9 +9,11 @@ namespace BTCPayServer.Data
         public string WalletDataId { get; set; }
         public WalletData WalletData { get; set; }
         public string TransactionId { get; set; }
+        [Obsolete]
         public string Labels { get; set; }
         public byte[] Blob { get; set; }
 
+        public List<WalletLabelData> WalletLabels { get; set; }
 
         internal static void OnModelCreating(ModelBuilder builder)
         {
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
@@ -126,6 +126,67 @@ namespace BTCPayServer.Controllers.Greenfield
             });
         }
 
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address/{address}")]
+        public async Task<IActionResult> GetOnChainWalletExistingAddress(string storeId, string cryptoCode, string address)
+        {
+            if (IsInvalidWalletRequest(cryptoCode, out var network,
+                    out var derivationScheme, out var actionResult))
+                return actionResult;
+            var bitcoinAddress =  BitcoinAddress.Create(address, network.NBitcoinNetwork);
+            var explorerClient = _explorerClientProvider.GetExplorerClient(network);
+           var kpi = await explorerClient.GetKeyInformationAsync(derivationScheme.AccountDerivation, bitcoinAddress.ScriptPubKey);
+           var bip21 = network.GenerateBIP21(kpi.Address?.ToString(), null);
+           var allowedPayjoin = derivationScheme.IsHotWallet && Store.GetStoreBlob().PayJoinEnabled;
+           if (allowedPayjoin)
+           {
+               bip21.QueryParams.Add(PayjoinClient.BIP21EndpointKey, Request.GetAbsoluteUri(Url.Action(nameof(PayJoinEndpointController.Submit), "PayJoinEndpoint", new { cryptoCode })));
+               
+           }
+
+           var walletId = new WalletId(storeId, cryptoCode);
+           var script = bitcoinAddress.ScriptPubKey.ToString();
+           if (!(await _walletRepository.GetLabelsForScripts(walletId, new[] {script})).ScriptLabels.TryGetValue(script,
+                   out var labels))
+           {
+               labels = new();
+           }
+           
+           return Ok(new OnChainWalletAddressData()
+           {
+               Labels = labels.Select(data => data.GetLabel()).ToList(),
+               Address = kpi.Address?.ToString(),
+               PaymentLink = bip21.ToString(),
+               KeyPath = kpi.KeyPath
+           });
+        }
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpPatch("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address/{address}/labels")]
+        public async Task<IActionResult> PatchOnChainWalletExistingAddress(string storeId, string cryptoCode,
+            string address, PatchLabelsRequest request)
+        {
+            if (IsInvalidWalletRequest(cryptoCode, out var network,
+                    out var derivationScheme, out var actionResult))
+                return actionResult;
+            var walletId = new WalletId(storeId, cryptoCode);
+            var script = BitcoinAddress.Create(address, network.NBitcoinNetwork).ScriptPubKey.ToString();
+            if (request.Labels != null)
+            {
+                await _walletRepository.AddLabels(walletId, request.Labels.Select(Label.Parse).ToArray(),
+                    new[] {script}, null);
+            }
+
+            if (request.RemoveLabels != null)
+            {
+                await _walletRepository.RemoveLabel(walletId, request.RemoveLabels.ToArray(), new[] {script}, null);
+            }
+
+            return await GetOnChainWalletExistingAddress(storeId, cryptoCode, address);
+        }
+
+
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
         [HttpGet("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address")]
         public async Task<IActionResult> GetOnChainWalletReceiveAddress(string storeId, string cryptoCode, bool forceGenerate = false)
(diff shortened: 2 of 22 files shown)


BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -845,6 +846,38 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
 
+            modelBuilder.Entity("BTCPayServer.Data.WalletLabelData", b =>
+                {
+                    b.Property<string>("WalletDataId")
+                        .HasColumnType("TEXT");
+
+                    b.Property<string>("Label")
+                        .HasColumnType("TEXT");
+
+                    b.Property<string>("Data")
```
NicolasDorier, line 857: We will change that to jsonb later for postgres. But don't do it now.

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
 
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpGet("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address/{address}")]
+        public async Task<IActionResult> GetOnChainWalletExistingAddress(string storeId, string cryptoCode, string address)
+        {
+            if (IsInvalidWalletRequest(cryptoCode, out var network,
+                    out var derivationScheme, out var actionResult))
+                return actionResult;
+            var bitcoinAddress =  BitcoinAddress.Create(address, network.NBitcoinNetwork);
```
NicolasDorier, line 137: need `address` validation

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
+            var explorerClient = _explorerClientProvider.GetExplorerClient(network);
+           var kpi = await explorerClient.GetKeyInformationAsync(derivationScheme.AccountDerivation, bitcoinAddress.ScriptPubKey);
```
NicolasDorier, line 139: what is `null`?

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
+           var bip21 = network.GenerateBIP21(kpi.Address?.ToString(), null);
```
NicolasDorier, line 140: `kpi.Address`->`address` so we don't have to test if null

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
             });
         }
 
+
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
```
NicolasDorier, line 130: Probably not right policy

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
+           var allowedPayjoin = derivationScheme.IsHotWallet && Store.GetStoreBlob().PayJoinEnabled;
+           if (allowedPayjoin)
+           {
+               bip21.QueryParams.Add(PayjoinClient.BIP21EndpointKey, Request.GetAbsoluteUri(Url.Action(nameof(PayJoinEndpointController.Submit), "PayJoinEndpoint", new { cryptoCode })));
+               
+           }
+
+           var walletId = new WalletId(storeId, cryptoCode);
+           var script = bitcoinAddress.ScriptPubKey.ToString();
+           if (!(await _walletRepository.GetLabelsForScripts(walletId, new[] {script})).ScriptLabels.TryGetValue(script,
```
NicolasDorier, line 150: Helper function `_walletRepository.GetLabelsForScript` would make that easier to read

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpPatch("~/api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/wallet/address/{address}/labels")]
+        public async Task<IActionResult> PatchOnChainWalletExistingAddress(string storeId, string cryptoCode,
+            string address, PatchLabelsRequest request)
+        {
+            if (IsInvalidWalletRequest(cryptoCode, out var network,
+                    out var derivationScheme, out var actionResult))
+                return actionResult;
+            var walletId = new WalletId(storeId, cryptoCode);
+            var script = BitcoinAddress.Create(address, network.NBitcoinNetwork).ScriptPubKey.ToString();
```
NicolasDorier, line 174: need validate `address`

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -126,6 +126,67 @@ public async Task<IActionResult> GetOnChainFeeRate(string storeId, string crypto
...
+            if (request.Labels != null)
```
NicolasDorier, line 175: I think `request` might be null

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -319,25 +337,30 @@ public async Task<IActionResult> GetOnChainWalletUTXOs(string storeId, string cr
...
+
+                    labels.TransactionComments!.TryGetValue(coin.OutPoint.Hash.ToString(), out var comment);
                     return new OnChainWalletUTXOData()
                     {
                         Outpoint = coin.OutPoint,
                         Amount = coin.Value.GetValue(network),
-                        Comment = info?.Comment,
-                        Labels = info?.Labels,
+                        Comment = comment,
+                        Labels = utxoLabels?.Select(data => data.GetLabel()).ToDictionary(data => data.Text),
```
NicolasDorier, line 356: note for later: Comment and Labels should be empty string and empty array rather than null.
It's a bug in current impl

BTCPayServer/Data/WalletTransactionDataExtensions.cs
```diff
@@ -9,11 +9,28 @@
 
 namespace BTCPayServer.Data
 {
+
+    public static class WalletLabelDataExtensions
+    {
+        public static LabelData GetLabel(this WalletLabelData walletLabelData)
+        {
+            return string.IsNullOrEmpty(walletLabelData.Data) ? new RawLabel(walletLabelData.Label) : Label.Parse(walletLabelData.Data);
```
NicolasDorier, line 17: we should remove this hack when we do a migration. Just using JSONB for all.

BTCPayServer/Data/WalletTransactionDataExtensions.cs
```diff
@@ -26,28 +43,28 @@ public static WalletTransactionInfo GetBlobInfo(this WalletTransactionData walle
...
+            //     {
+            //         // Legacy path
+            //         foreach (var token in walletTransactionData.Labels.Split(',',
+            //             StringSplitOptions.RemoveEmptyEntries))
+            //         {
+            //             var l = Label.Parse(token);
+            //             blobInfo.Labels.TryAdd(l.Text, l);
+            //         }
+            //     }
+            // }
```
NicolasDorier, line 67: to delete

BTCPayServer.Data/Data/WalletTransactionData.cs
```diff
@@ -7,9 +9,11 @@ public class WalletTransactionData
         public string WalletDataId { get; set; }
         public WalletData WalletData { get; set; }
         public string TransactionId { get; set; }
+        [Obsolete]
```
NicolasDorier, line 12: Should be removed, should add AddtionalData attribute, the migration should use this instead.

BTCPayServer/Data/WalletTransactionDataExtensions.cs
```diff
@@ -59,15 +76,15 @@ public static void SetBlobInfo(this WalletTransactionData walletTransactionData,
...
-            walletTransactionData.Labels = new JArray(
-                blobInfo.Labels.Select(l => JsonConvert.SerializeObject(l.Value, LabelSerializerSettings))
-                .Select(l => JObject.Parse(l))
-                .OfType<JToken>()
-                .ToArray()).ToString();
+            // walletTransactionData.Labels = new JArray(
+            //     blobInfo.Labels.Select(l => JsonConvert.SerializeObject(l.Value, LabelSerializerSettings))
+            //     .Select(l => JObject.Parse(l))
+            //     .OfType<JToken>()
+            //     .ToArray()).ToString();
```
NicolasDorier, line 87: to delete

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -35,55 +50,337 @@ public async Task SetWalletInfo(WalletId walletId, WalletBlobInfo blob)
...
-                            .Where(data => transactionIds == null || transactionIds.Contains(data.TransactionId))
-                            .Select(w => w)
-                            .ToArrayAsync())
-                            .ToDictionary(w => w.TransactionId, w => w.GetBlobInfo());
+            var walletIdStr = walletId.ToString();
+            scripts ??= Array.Empty<string>();
+            txs ??= Array.Empty<string>();
+            await using var context = _ContextFactory.CreateContext();
+           var exist =  await context.Wallets.FindAsync(walletId.ToString());
+           if (exist is null)
```
NicolasDorier, line 72: should create this at the same time as the store rather than lazily. So we don't have to deal with the fact it can be NULL afterward.

BTCPayServer.Data/Data/WalletTransactionData.cs
Kukks: this is a database table class so we cannot do this

BTCPayServer.Data/Migrations/ApplicationDbContextModelSnapshot.cs
Kukks: we already do this for some types in btcpay so may as well keep going

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
Kukks: we use the policy for other wallet related actions

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -35,55 +42,360 @@ public async Task SetWalletInfo(WalletId walletId, WalletBlobInfo blob)
...
+
+            return result;
+
+        }
+
+        public async Task AssociateTransactionToScripts(Transaction transactionDataTransaction)
+        {
+            var txHash = transactionDataTransaction.GetHash().ToString();
+            var scripts = transactionDataTransaction.Inputs
+                .Select(txIn => txIn.GetSigner().ScriptPubKey.ToString())
```
NicolasDorier, line 349: we can't use this, it doesn't work with taproot

woutersamaey:
@Kukks have you seen this BIP describing a CSV format for label import/export? It's super easy.

Maybe this fits your work or something to consider?

https://github.com/craigraw/bips/blob/master/bip-wallet-labels.mediawiki

NicolasDorier:
https://github.com/btcpayserver/btcpayserver/pull/4179

Issue #3954, opened 2022-07-11
Improve Wallet Transactions view & pagination functionality


dstrukt:
As originally requested by @NicolasDorier on MM, summarized a high-level list of tasks he requested + some tasks for consideration while we tackle the original request, if this should be broken down further for implementation, please let me know.

- [ ] Load the most recent 1 month's worth of transactions by default
- [ ] Add a "Load/View More" button at the bottom of the scroll view, which triggers the loading of the next month of transactions
- [ ] Either continue to show another button the user must click until the end of the list of transactions OR hitting the bottom of the view triggers the infinite scroll with no additional button click necessary .. DECISION NEEDED
- [ ] Add support for search, specifically for TX id
- [ ] Add support for date range filter (similar to what we'll do for the Invoice view) .. DECISION NEEDED

Accompanying mock to be posted soon.

HamroRamro:
PR https://github.com/btcpayserver/btcpayserver/pull/4074 up for review for 1 through 3 , @dstrukt

pavlenex:
@dstrukt Can you elaborate what's left for this one? Do you have any screens ready or is this perhaps entirely done?

dennisreimann:
We can revisit this once #4914 is done, because the remainders here sound like we can simply reuse the patterns we are about to establish in the invoices list.

dstrukt:
@dennisreimann Is correct, we can revisit this ticket with some of the patterns we're establishing, as well as https://github.com/btcpayserver/btcpayserver/issues/3872

AbsurdlySuspicious:
Would also be nice:
- "Only IN/OUT transactions" filter
- "No label" filter

pavlenex:
@dstrukt This issue is now 3 years old, can you re-check and ensure if the items listed are still relevant?

dstrukt:
@pavlenex The Reporting arguably addresses the custom time filter, but I still think it would be nice in the UI?

There is no `Search`, for TX ID, etc.. or filters as @AbsurdlySuspicious suggested, which I also think would be great additions.

So the 2 checklist items are still relevant imo, and a `good first issue` / contributor task?

pavlenex:
I'll break these small improvements into separate concrete issues and tag them as good first issues, then close this issue.

PR #4074, opened 2022-08-24
Update WalletTransactions pagination default settings

HamroRamro:
As of now 3/5 task of https://github.com/btcpayserver/btcpayserver/issues/3954 is resolved here. Will work on the remaining later this week/month.

- [X] Load 1 month's worth of transactions by default

- [X] Add a "Load/View More" button at the bottom of the scroll view, which triggers the loading of the next month of transactions

- [X] Display Show All after loading 90 days of data which triggers infinite scroll with no additional button click necessary

Please let me know if you have any feedback @dstrukt @NicolasDorier. 
cc: @pavlenex @Kukks


BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -280,7 +293,19 @@
...
+        <div align="center">
+            <button type="button" class="btn btn-primary" onclick="location.href='@Url.Action(nameof(UIWalletsController.WalletTransactions), new { walletId = walletId , interval = nextInterval})'">@loadMoreButtonText</button>
+        </div>
+    }
+    else
+    {
+         <div align="center">
+            <button type="button" class="btn btn-primary" onclick="window.scrollTo({ top: 0, behavior: 'smooth' })">Go to top</button>
+        </div>
+    }
```
dennisreimann, line 308:
```suggestion
    <div class="text-center mb-5">
        @if (displayLoadMoreButton)
        {
            <a role="button" class="btn btn-primary" asp-action="WalletTransactions" asp-route-walletId="@walletId" asp-route-interval="@nextInterval">@loadMoreButtonText</a>
        }
        else
        {
            <button type="button" class="btn btn-primary" onclick="window.scrollTo({ top: 0, behavior: 'smooth' })">Go to top</button>
        }
    </div>
```

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -259,20 +259,31 @@ public async Task<IActionResult> ListWallets()
             WalletId walletId,
             string? labelFilter = null,
             int skip = 0,
-            int count = 50
+            int count = 50,
+            TimeSpan? interval = null
```
dennisreimann, line 263: We can simplify it quite a bit in the other parts also when we use an `int` here and also rename it, to reveal the intention better: 
  
```suggestion
            int days = 30
```

dstrukt:
Awesome, will take a look this evening!

BTCPayServer/Extensions/StringExtensions.cs
```diff
@@ -15,5 +15,16 @@ public static class StringExtensions
...
+        public static bool HasValue(this string input)
+        {
+            return !string.IsNullOrEmpty(input);
+        }
+
+        public static bool IsNullOrEmpty(this string input)
+        {
+            return string.IsNullOrEmpty(input);
+        }
+
```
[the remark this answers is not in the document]
HamroRamro: Thank you for the feedback @dennisreimann. I have reverted the changes and have modified relevant code accordingly.

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
HamroRamro: Awesome tip! Made the change!

BTCPayServer/Controllers/UIWalletsController.cs
HamroRamro: Changed it to days. LOT CLEANER!! Thank you for the suggestion!

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -2,6 +2,14 @@
 @{
     var walletId = Context.GetRouteValue("walletId").ToString();
     var labelFilter = Context.Request.Query["labelFilter"].ToString();
+
+    var nextInterval = Model.Days + 30;
+    var scrollToBottom = nextInterval > 60;
+    var displayLoadMoreButton = nextInterval < 365000;
+
+    var loadMoreButtonText = nextInterval > 90 ? "Show All" : "Load More";
+    nextInterval = nextInterval > 90 ? 365000 : nextInterval;
```
dennisreimann, line 11: It can be simplified to this, when we apply the other checks that are used just once in-place (see other suggestions):

```suggestion
    var nextInterval = Model.Days > 60 ? 365000 : Model.Days + 30;
    var displayLoadMoreButton = nextInterval < 365000;
```

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -280,7 +288,17 @@
             </table>
         </div>
     </div>
-    <vc:pager view-model="Model"/>
+
+    <div class="text-center mb-5">
+        @if (displayLoadMoreButton)
+        {
+            <a role="button" class="btn btn-primary" asp-action="WalletTransactions" asp-route-walletId="@walletId" asp-route-days="@nextInterval">@loadMoreButtonText</a>
```
dennisreimann, line 295:
```suggestion
            <a role="button" class="btn btn-primary" asp-action="WalletTransactions" asp-route-walletId="@walletId" asp-route-days="@nextInterval">@(nextInterval > 90 ? "Show All" : "Load More")</a>
```

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -294,3 +312,10 @@ else
     <br/>
     If some transactions appear in BTCPay Server, but are missing in another wallet, <a href="https://docs.btcpayserver.org/FAQ/Wallet/#missing-payments-in-my-software-or-hardware-wallet" rel="noreferrer noopener">follow these instructions</a>.
 </p>
+
+@if (scrollToBottom)
```
dennisreimann, line 316:
```suggestion
@if (nextInterval > 60)
```

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -294,3 +312,10 @@ else
...
+{
+    <script type="text/javascript">
+        window.scrollTo({ top: document.body.scrollHeight, behavior: 'smooth' })
+    </script>
+}
```
dennisreimann, line 321: Let's also move this part into the `@section PageFootContent` above, because otherwise it might get overlooked when scripts are spread across multiple places.

dstrukt (changes requested):
First of all, excellent work!

Few nits:
- [ ] We should switch to our `btn-secondary` style

- [ ] Something worth considering too is making this a full-width button, e.g (@dennisreimann thoughts?):
Screen Shot 2022-08-25 at 2 48 56 AM
Screen Shot 2022-08-25 at 2 58 44 AM

- [ ] Functionally, should we start infinite scroll loading when the user hits the bottom of the view after 1 or 2 "Load More" clicks, as opposed to continuing to click until "View All"? @dennisreimann Thoughts?

Otherwise, looks & works great!

dennisreimann:
I'd rather left-align the button than having it full width.

Infinite scrolling: Ideally we'd do it all via AJAX, so that even initially there's no page reload. We could switch to that once the "Load more" button was pressed, which seems like an intuitive thing. RIght now it feeld kinda random to switch to infinite mode after two clicks.

dstrukt:
- hmm, i prefer center, even if it's not `full-wdith` (even though I do like this), I also realize it could be too much .. maybe there's a middle ground of the 2? Regardless, would be more than happy to proceed with the standard button size.

- w.r.t to the infinite scroll, agreed with the switch after the first press, that would be even better!

HamroRamro:
Hi @dstrukt and @dennisreimann ,

Thanks for all the feedback. I have made the code a little cleaner. 

I have made the button btn-secondary. I do prefer the button to be center rather than full width. Full width is a little too heavy for me. 
I have also made it so that it gives you option to Show All after 1 Load More. So 
1st page load = 30 days of data. 
2nd page load  = 60 days of data and button changes to "Show All"

Once Show All is hit, it loads everything and takes you to the bottom.

And technically..... you CAN SCROLL UP "INFINITELY" haha :)

I will work on the tx search and date filter next but please feel free to suggest more UI/UX enhancements!!

dennisreimann:
I think having a "Load All" button that simply fetches the rest will bring trouble for wallets with many transactions. We have users with lots of transactions and I don't think this would be a good way to handle it. 

Can we rather make it load in 30 day interval steps with AJAX and the infitite scroll? It doesn't have to be part of this PR, but I think that'd be a more user-friendly and scalable way.

HamroRamro:
I agree with you. Load All can be heavy with wallets with a lot of transactions.

I'll take a stab at your suggestions over the weekend. Thanks for the feedback ser

dstrukt:
ACK to @dennisreimann's suggestion, and looking forward to the next iteration @HamroRamro 💪

HamroRamro:
Hello gentlemen @dstrukt @dennisreimann ,

I have implemented infinite scroll. It will load 30 days worth of data every time you scroll to the bottom. 
Also got rid of the "Load More" completely because it will always add on 30 days more worth of data on scrolling down. 

It shows "Loading..." in the center while fetching data. 

Let me know your thoughts. Thank you!

BTCPayServer/BTCPayServer.csproj
```diff
@@ -95,6 +95,7 @@
   </ItemGroup>
 
 	<ItemGroup>
+	  <None Include="Views\UIWallets\_WalletTransactions.cshtml" />
```
dennisreimann, line 98: Can you remove the changes in this file? They are superfluous.

BTCPayServer/Views/UIWallets/_WalletTransactions.cshtml
```diff
@@ -0,0 +1,287 @@
+@model ListTransactionsViewModel
+@{
+    var walletId = Context.GetRouteValue("walletId").ToString();
+    var labelFilter = Context.Request.Query["labelFilter"].ToString();
+    ViewData.SetActivePage(WalletsNavPages.Transactions, $"{Model.CryptoCode} Transactions", walletId);
+}
+
+@section PageHeadContent {
```
dennisreimann, line 8: Please don't split it up like this, the sections do not belong into partial views, because they won't render anymore.

The `ViewData.SetActivePage` also belongs to the outer page.

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -1,310 +1,56 @@
...
+                        $("#walletTransactions").empty().append(data);
+                    }
+                    _inCallback = false;
+                    $('div#loadingText').empty();
+                });
+
+            }
+        }
+
+        $(window).scroll(function() {
```
dennisreimann, line 37: Let's not use jQuery for new code, can you convert it into the native equivalents like [fetch](https://javascript.info/fetch)?

dennisreimann (changes requested):
The changeset became pretty big with the latest change and in some parts the indentation got messed up.

I think the AJAX part can be solved in a better way. Can you make the annotated changes and grant me access to you fork, so that I can do some more modifications afterwards?

HamroRamro:
Hi @dennisreimann , I have added you as a contributor to my fork. I am out of town today so I will try to implement your suggestions tomorrow.

dstrukt:
@HamroRamro Great work!

I do like the infinite load UX much more .. debating if I like it being triggered by hitting the bottom of the page by default, or if we should have a singular "Load More" then the second, and onward content is loaded with the infinite scroll .. are there potential issues of someone abusing this (API calls, etc..)? Regardless, I think we could easily go with the current iteration, and we'll get some feedback, and adjust as needed.

dennisreimann:
I've pushed the cleanups taht resolve my previous comments. This is still in WIP state, as I found that tehre is something weird going on when using the `skip` and `count` parameters. I've simplified the general approach quite a bit by removing the previous changes based on days as an interval and reusing what we have with the pager already.

However, I couldn't find out why it behaves so weirdly and e.g. the second page is just empty when using `?skip=50&count=50` for instance. I've re-added the pager to verify this isn't just an issue with the AJAX calls.

@NicolasDorier Debugging it I found that the [`else` branch here](https://github.com/btcpayserver/btcpayserver/blob/8873c51f2be85f12ff73b4674e3c4346c89abc53/BTCPayServer/Services/Wallets/BTCPayWallet.cs#L230) doesn't yield the expected results - we'll need to take a closer look what's going on.

@HamroRamro Sorry for rewriting larger chunks, if you want to, we can go through the changes in detail if you have questions.

dennisreimann:
Ok, problems found and solved:

- [f7fc866](https://github.com/btcpayserver/btcpayserver/pull/4074/commits/f7fc866febab0caf8e272c559fb8ec41666262f2) fixes a bug on the Controller level, which lead to the weird behaviour I describe in my previous comment.
- There's still a bug in NBXplorer's `get_wallets_recent`, which @NicolasDorier is taking care of. [c760a05](https://github.com/btcpayserver/btcpayserver/pull/4074/commits/c760a0565cae4fc77797503c81a1027dc13f002c) circumvents it and should be removed once the NBXplorer fix is integrated.

I added the workaround, so that @HamroRamro and @dstrukt can take a look already. I think the current behaviour matches what we agreed upon and I think it's a good solution like that.

HamroRamro:
Hi @dennisreimann , this looks great! 👍  Thank you so much for doing this! And sorry for late response, I was out of town with family! 

The user first sees Load More and Go To Top and once they hit Load More once, it triggers the infinite scroll. I like it! 

Lets see what @dstrukt has to say :)

dstrukt:
Apologies for my delay in review, will take a look in the morning!

Appreciate all of the work that has gone into this.

NicolasDorier:
I think, in the request, the `skip` should always be true, then after the request we should manually skip the rows.

It's less efficient of course, but this will ensure the bug is fixed even if somebody didn't update NBXplorer.
I may consider later to detect the NBXPlorer version and just remove this workaround for what is higher, but can do that in another PR.

dstrukt:
I think the section being loading in is bugged...?

Screen Shot 2022-09-18 at 2 14 20 PM

dennisreimann (approved):
Good to go now :)

dstrukt:
Strange. I pulled the latest, but still getting similar results, but slightly more broken ...

Screen Shot 2022-09-19 at 10 53 20 PM

dennisreimann:
@dstrukt can you run `dotnet clean && dotnet restore` and also stop and restart the docker-compose?

@NicolasDorier I've upgraded NBXplorer in both docker-compose files, so that the bugfix is now included. Both cases work properly now.

dstrukt (approved):
Functionally and aesthetically, LGTM!

Only thing I'm not really a fan of is the "Back to top" button, think it should be removed, imo, but I could be persuaded.

NicolasDorier:
Need rebase, the two .cs files are just style issue, the site.css, I don't know how to merge properly.

pavlenex:
@NicolasDorier do you want this one in the v1.6.11?

dennisreimann:
@pavlenex We added #4150 with the fix as a subset of this one. So we'll ship the feature part of this with one of the next releases after 1.6.11

dennisreimann:
Now that the tests are green again and the release is done, this is ready to be reviewed and merged. @NicolasDorier

NicolasDorier:
This is awesome

PR #4075, opened 2022-08-24
Remove option to generate a watch-only wallet

dennisreimann:
This was sparked by a discussion with @pavlenex: Unfortunately users seem to not store their seed phrase, even though the wallet setup process asks them to do so. This is especially bad for watch-only wallets, as without having it imported into an external wallet, there will be no way to access the funds later on.

Our idea was to remove the option of generating watch-only wallets, which would require the user to first create it another way (e.g. external wallet software or hardware wallet). This leaves the user with the option to import an existing wallet as watch-only, so that they have at least the previous setup instead of relying on them storing the seed properly.

Publishing this as a draft to open the topic for discussion.


Kukks:
> This was sparked by a discussion with @pavlenex: Unfortunately users seem to not store their seed phrase, even though the wallet setup process asks them to do so. This is especially bad for watch-only wallets, as without having it imported into an external wallet, there will be no way to access the funds later on.
> 
> Our idea was to remove the option of generating watch-only wallets, which would require the user to first create it another way (e.g. external wallet software or hardware wallet). This leaves the user with the option to import an existing wallet as watch-only, so that they have at least the previous setup instead of relying on them storing the seed properly.
> 
> Publishing this as a draft to open the topic for discussion.

The main issue around directing users to create a wallet elsewhere would be the same reason we introduced wallet creation: all wallets have different formats and importing them in btcpay always breaks in some way or another. 

I am for hiding it away so that users have to actively looking for this feature.  

We should maybe also introduce a seed re-entry confirmation screen.

dstrukt:
Agree in general, and second @Kukks' points.

We had discussed the seed re-entry screen, and I believe I started to mock up some stuff, but it's been a minute since I've looked at those, so will revisit, post and we can address some of the other minor improvements in that flow in the process.

dennisreimann:
Ok, agreed, seed confirmation would be the proper way to solve it and it'd also help with the generate hot wallet option. We'll redo it like that!

PR #4076, opened 2022-08-24
PoS UI fix: scale-down item image

prusnak:
before:

Screenshot 2022-08-24 at 17 23 04

after:

Screenshot 2022-08-24 at 17 23 14

image source: https://cdn.rohlik.cz/images/grocery/products/762961/762961-1646145765627.jpg

### DIFF
--- a/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
+++ b/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
@@ -27,7 +27,7 @@
 </script>
 
 <script id="template-cart-item-image" type="text/template">
-    <img src="{image}" width="50" alt="">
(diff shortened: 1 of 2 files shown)


Kukks (approved):
ack from my end but what do you think @dennisreimann  and @dstrukt? Does this handle all cases gracefully?

prusnak:
Another option is using `contain` if you want to scale-up images that are smaller than the bounding box.

prusnak:
Added another commit 3853af712165ed9307af1725247af8fdcdb11963  which fixes the same issue in cart.

before:

Screenshot 2022-08-24 at 20 45 39

after:

Screenshot 2022-08-24 at 20 46 17

dennisreimann (approved):
Good point, `scale-down` works well here.

dstrukt (approved):
LGTM!

Nice change.

PR #4084, opened 2022-08-28
Make sure end date is after start date in Crowdfund app

bolatovumar:
I noticed we don't validate that end date is after start date in the crowdfund app. We should probably check for this. This PR adds this check.

Capture

### DIFF
--- a/BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
+++ b/BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
@@ -126,7 +126,7 @@ namespace BTCPayServer.Plugins.Crowdfund.Controllers
             if (!string.IsNullOrEmpty(request.ChoiceKey))
             {
                 var choices = _appService.GetPOSItems(settings.PerksTemplate, settings.TargetCurrency);
-                choice = choices.FirstOrDefault(c => c.Id == request.ChoiceKey);
+                choice = choices?.FirstOrDefault(c => c.Id == request.ChoiceKey);
                 if (choice == null)
                     return NotFound("Incorrect option provided");
                 title = choice.Title;
@@ -379,7 +379,13 @@ namespace BTCPayServer.Plugins.Crowdfund.Controllers
         {
             if (string.IsNullOrWhiteSpace(currency))
             {
-                currency = (await _storeRepository.FindStore(storeId)).GetStoreBlob().DefaultCurrency;
+                var store = await _storeRepository.FindStore(storeId);
+                if (store == null)
+                {
+                    throw new Exception($"Could not find store with id {storeId}");
+                }
+
+                currency = store.GetStoreBlob().DefaultCurrency;
             }
             return currency.Trim().ToUpperInvariant();
         }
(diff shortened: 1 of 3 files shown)


BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
```diff
@@ -374,7 +380,13 @@ private async Task<string> GetStoreDefaultCurrentIfEmpty(string storeId, string
         {
             if (string.IsNullOrWhiteSpace(currency))
             {
-                currency = (await _storeRepository.FindStore(storeId)).GetStoreBlob().DefaultCurrency;
+                var store = await _storeRepository.FindStore(storeId);
+                if (store == null)
```
bolatovumar, line 384: `store` can be `null` here technically so we need to check for that.

BTCPayServer/Views/Shared/Crowdfund/UpdateCrowdfund.cshtml
```diff
@@ -121,8 +120,9 @@
                             <vc:icon symbol="close"/>
                         </button>
                     </div>
-                    <span asp-validation-for="EndDate" class="text-danger"></span>
                 </div>
+                <span asp-validation-for="StartDate" class="text-danger"></span>
```
bolatovumar, line 124: Placed these below both of the elements so  the layout doesn't get messed up if there is an error.

BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
```diff
@@ -287,12 +288,17 @@ public async Task<IActionResult> UpdateCrowdfund(string appId, UpdateCrowdfundVi
 
             if (Enum.Parse<CrowdfundResetEvery>(vm.ResetEvery) != CrowdfundResetEvery.Never && !vm.StartDate.HasValue)
             {
-                ModelState.AddModelError(nameof(vm.StartDate), "A start date is needed when the goal resets every X amount of time.");
+                ModelState.AddModelError(nameof(vm.StartDate), "A start date is needed when the goal resets every X amount of time");
```
bolatovumar, line 291: Removed a period character at the end since other errors don't have it.

BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
```diff
@@ -287,12 +288,17 @@ public async Task<IActionResult> UpdateCrowdfund(string appId, UpdateCrowdfundVi
...
             }
 
             if (Enum.Parse<CrowdfundResetEvery>(vm.ResetEvery) != CrowdfundResetEvery.Never && vm.ResetEveryAmount <= 0)
             {
-                ModelState.AddModelError(nameof(vm.ResetEveryAmount), "You must reset the goal at a minimum of 1 ");
+                ModelState.AddModelError(nameof(vm.ResetEveryAmount), "You must reset the goal at a minimum of 1");
```
bolatovumar, line 296: There was an extra trailing space here which I removed.

bolatovumar:
Failing test unrelated

Capture

dennisreimann (approved):
tACK

dstrukt:
tACK, nice catch

dstrukt (approved):

PR #4093, opened 2022-09-02
 Upgrade Lightning library to 1.4.1

### LINKED ISSUE #2182: TLS thumbprint is not updated when LND changes it
**Describe the bug**
LND occasionally changes TLS certificate which causes BTCPayServer connection to break since TLS fingerprint changes as well. The user gets confusing error message and has to manually reset the link to fix it. (This is not apparent from the message.)

**Suggested solution**
Provide `certpath`  config option similar to `macaroonpath` that will accept the path to TLS certificate, re-read it when connection to LND fails and attempt to connect again (if changed).
If this is difficult or lot of work, suggest resetting the link in the error message as a temporary solution.

**To Reproduce the bug**
1. Setup BTCPayServer with *internal* LND ("Use internal node" link")
2. Cause LND to change certificate e.g. by changing domain name or some IP address.
3. Attempt to create LN invoice from BTCPayServer

**Expected behavior**

LN invoice is created without any error messages.

- BTCPay Server Version:    1.0.6.0, looking at the changelog it should behave the same using the latest version
- Deployment Method:       Debian package, should be irrelevant unless Docker deployment does some magic trick to fix it
- Browser:                 irrelevant

**Logs (if applicable)**

I only got small part from an affected user, this message should be relevant:

> Impossible to create lightning invoice (The SSL connection could not be established, see inner exception)

I can ask for more, however given the fact that TLS cert thumbprint is not (and can not be) dynamically updated in configuration and resetting the link helped I don't expect it's needed.

**Setup Parameters**

Config contains `btc.lightning=` option with `certthumbprint=XYZ`

Kixunil:
This enables new feature of specifying `certfilepath`

Closes #2182

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -48,7 +48,7 @@
   <ItemGroup>
     <PackageReference Include="BIP78.Sender" Version="0.2.2" />
     <PackageReference Include="BTCPayServer.Hwi" Version="2.0.2" />
-    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.3.13" />
+    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.4.1" />
     <PackageReference Include="BuildBundlerMinifier" Version="3.2.449" />
     <PackageReference Include="BundlerMinifier.Core" Version="3.2.435" />
     <PackageReference Include="BundlerMinifier.TagHelpers" Version="3.2.435" />


dennisreimann:
@NicolasDorier we still need to get to upgrading the Client as well, I think otherwise this doesn't fix the original issue. 

Reminder: we have this compatibility problem, which now requires the Client to also use .net6 instead of .netstandard2 as the platform.

NicolasDorier:
I did

PR #4100, opened 2022-09-05
Use mempool as default block explorer

junderw:
This will allow users to utilize mempool for the default block explorer.

Eventually, it would be nice if we could detect if the user is using tor and use mempool's hidden service onion url.
Also, since mempool can also be self-hosted, it would be nice to eventually allow the admin to set the block explorer URL for txids to their liking on a per coin basis.

(Also, side note, mempool will soon be able to add nonWitnessUtxo (full parent raw tx) to PSBT inputs. See: [here](https://github.com/mempool/mempool/pull/2514))


NicolasDorier:
ACK, will check later while test fail

bolatovumar (approved):
Failing tests seem unrelated

NicolasDorier:
@dennisreimann yes but we would need better refactoring:

`GetBlockExplorerLink(bool tor)` should be created. Then we should make sure that `BlockExplorerLink` isn't used directly anywhere.

PR #4106, opened 2022-09-08
Show iframe when showing invoice in Shopify plugin

### LINKED ISSUE #4105: [Bug]: Invoice display broken on Shopify

BTCPay Server: v1.6.3.0 (on Umbrel)
Store: https://robotechy.com/

Umbrel Raspberry Pi docker image.

### What happened?

We have our Umbrel node server the JS file over HTTPS as per https://jorijn.com/installing-nginx-reverse-proxy-with-ssl-certificate-umbrel-btcpayserver/. I noticed that the JS file was failing to load for two reasons, (1) our SSL had expired, and (2) the URL to the JS file has been changed. The JS file is now loading (after updating the link), and we see the "Pay with BTCPay" icon again. However, when you click on it, it just has "Displaying Invoice..." but nothing actually pops up.

1. Go to https://robotechy.com
2. Go through to the checkout for an item (feel free to enter "BTCPayServer" and will tell son to ignore order).
3. Choose "BTCPay Server" method.
4. Click on "Pay with BTCPay".

### Relevant log output

_No response_

### What browser do you use?

Browser(s): Edge and Chrome

### Additional information

_No response_

bolatovumar:
For whatever reason we do not call `btcpay.showFrame` after `btcpay.showInvoice` in the Shopify plugin even though we do it in other places. This results in the iframe being added to the document but not shown.

See here and here:

https://github.com/btcpayserver/btcpayserver/blob/5bbaa48b492556df32aaa8cd81b7e7ac9ea33794/BTCPayServer/wwwroot/payment-request/app.js#L131-L132

https://github.com/btcpayserver/btcpayserver/blob/4d0f76f9e8cf85c4104f248d5877d15dbc844a39/BTCPayServer/wwwroot/crowdfund/app.js#L222-L223

Close #4105

### DIFF
--- a/BTCPayServer/wwwroot/modal/btcpay.js
+++ b/BTCPayServer/wwwroot/modal/btcpay.js
@@ -105,7 +105,7 @@
         onModalReceiveMessageMethod(event);
     }
 
-    function showInvoice(invoiceId, params) {
+    function appendInvoiceFrame(invoiceId, params) {
         showingInvoice = true;
         window.document.body.appendChild(iframe);
         var invoiceUrl = origin + '/invoice?id=' + invoiceId + '&view=modal';
@@ -113,6 +113,11 @@
             invoiceUrl += '&animateEntrance=false';
         }
         iframe.src = invoiceUrl;
+    }
+
+    function appendAndShowInvoiceFrame(invoiceId, params) {
+        appendInvoiceFrame(invoiceId, params);
+        showFrame();
     }
 
     function setButtonListeners() {
@@ -133,7 +138,9 @@
     window.btcpay = {
         showFrame: showFrame,
         hideFrame: hideFrame,
-        showInvoice: showInvoice,
+        showInvoice: appendInvoiceFrame,
+        appendInvoiceFrame: appendInvoiceFrame,
+        appendAndShowInvoiceFrame: appendAndShowInvoiceFrame,
         onModalWillEnter: onModalWillEnter,
         onModalWillLeave: onModalWillLeave,
         setApiUrlPrefix: setApiUrlPrefix,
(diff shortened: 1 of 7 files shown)


BTCPayServer/wwwroot/shopify/btcpay-shopify.js
```diff
@@ -133,6 +133,7 @@ window.BTCPayShopifyIntegrationModule = function () {
                 });
             });
             window.btcpay.showInvoice(currentInvoiceData.invoiceId);
+            window.btcpay.showFrame();
```
bolatovumar, line 136: Idk if there is an actual reason we weren't calling `showFrame` here previously or if it's just an oversight.

bolatovumar:
Failing test unrelated

Capture

Kukks:
Do you think we need to add this directly to the modal js instead of specifically here for shopify?

bolatovumar:
> Do you think we need to add this directly to the modal js instead of specifically here for shopify?

@Kukks done in https://github.com/btcpayserver/btcpayserver/pull/4106/commits/66d7ad84c933ef9b13f57380ed1083835d341421

Looks like we can't just wholesale replace the old method with the new one because there is actually an alternative way to show the iframe by doing a `postMessage` which will end up showing/hiding the invoice as appropriate so I just created the method which can be used in other places.

BTCPayServer/wwwroot/modal/btcpay.js
```diff
@@ -133,7 +138,9 @@
     window.btcpay = {
         showFrame: showFrame,
         hideFrame: hideFrame,
-        showInvoice: showInvoice,
+        showInvoice: appendInvoiceFrame,
```
bolatovumar, line 141: I kept the old global method name around in case there is someone who relies on it? If we don't care about it then it can just be deleted.

NicolasDorier:
I let @Kukks decided whether to merge this PR, I don't know much about shopify

BTCPayServer/wwwroot/modal/btcpay.js
Kukks: This is good, we should strive to always be backwards compatible

PR #4113, opened 2022-09-12
Fix runtime error on ubuntu 22.04 in dev mode

NicolasDorier:
User reported:

> Hi there! I’m in the process of setting up BTCPayserver on Ubuntu 22.04, followed all of the directions pretty carefully, including a fresh install of .NET 6.0 from the MS repos… I’m getting a weird error when trying to view served up web pages:  System.InvalidOperationException: Cannot find compilation library location for package ‘System.Security.Cryptography.Pkcs’  Anyone seen such an issue?

It seems to be a common issue, reported on 

https://stackoverflow.com/questions/70236519/invalidoperationexception-cannot-find-compilation-library-location-for-package/70741481#70741481

https://github.com/dotnet/sdk/issues/16818

https://github.com/toddams/RazorLight/issues/460

The workaround should fix the issue.

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -94,6 +94,11 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="6.0.7" />
   </ItemGroup>
 
+   <PropertyGroup>
+     <!--https://stackoverflow.com/questions/70236519/invalidoperationexception-cannot-find-compilation-library-location-for-package/70741481#70741481-->
+     <GenerateRuntimeConfigDevFile>true</GenerateRuntimeConfigDevFile>
+   </PropertyGroup>
+
 	<ItemGroup>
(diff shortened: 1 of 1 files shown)


dennisreimann:
@woutersamaey you had a different problem that was related to the runtime or SDK version, right?

woutersamaey:
My issue was that the .NET SDK was suddenly not installed anymore. I only had the runtime, but no SDK. This was due to an "apt update" selecting the wrong packages.
As a result I could not build anymore.
The solution to my issue was telling `apt` to give preference to the Microsoft repo over the Ubuntu one so the correct packages were found instead.
So no, sounds unrelated.

NicolasDorier:
Keeping that open until somebody can replicate the issue and check this fix it

NicolasDorier:
I followed these instructions to install dotnet 6.0 on ubuntu 22.04.

https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu#2204

Can't replicate the issue

pox:
@NicolasDorier I can now replicate this issue on Ubuntu with v1.7.3 (and with `master`).

PR #4131, opened 2022-09-17
Store settings: Add branding options

dennisreimann:
Adds logo upload and brand color definition in the store settings.

grafik

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -644,15 +644,26 @@ namespace BTCPayServer.Tests
             s.Driver.FindElement(By.Id("enable-pay-button")).Click();
             s.Driver.FindElement(By.Id("disable-pay-button")).Click();
             s.FindAlertMessage();
-            s.GoToStore(StoreNavPages.General);
+            s.GoToStore();
             Assert.False(s.Driver.FindElement(By.Id("AnyoneCanCreateInvoice")).Selected);
             s.Driver.SetCheckbox(By.Id("AnyoneCanCreateInvoice"), true);
             s.Driver.FindElement(By.Id("Save")).Click();
             s.FindAlertMessage();
             Assert.True(s.Driver.FindElement(By.Id("AnyoneCanCreateInvoice")).Selected);
 
+            // Store settings: Set and unset brand color
+            s.GoToStore();
+            s.Driver.FindElement(By.Id("BrandColor")).SendKeys("#f7931a");
+            s.Driver.FindElement(By.Id("Save")).Click();
+            Assert.Contains("Store successfully updated", s.FindAlertMessage().Text);
+            Assert.Equal("#f7931a", s.Driver.FindElement(By.Id("BrandColor")).GetAttribute("value"));
+            s.Driver.FindElement(By.Id("BrandColor")).Clear();
+            s.Driver.FindElement(By.Id("Save")).Click();
+            Assert.Contains("Store successfully updated", s.FindAlertMessage().Text);
+            Assert.Equal(string.Empty, s.Driver.FindElement(By.Id("BrandColor")).GetAttribute("value"));
+
             // Alice should be able to delete the store
-            s.GoToStore(StoreNavPages.General);
+            s.GoToStore();
             s.Driver.FindElement(By.Id("DeleteStore")).Click();
             s.Driver.WaitForElement(By.Id("ConfirmInput")).SendKeys("DELETE");
             s.Driver.FindElement(By.Id("ConfirmContinue")).Click();
--- a/BTCPayServer/ColorPalette.cs
+++ b/BTCPayServer/ColorPalette.cs
@@ -1,12 +1,18 @@
 using System;
 using System.Drawing;
 using System.Text;
+using System.Text.RegularExpressions;
 using NBitcoin.Crypto;
 
 namespace BTCPayServer
 {
     public class ColorPalette
     {
+        public const string Pattern = "^#[0-9a-fA-F]{6}$";
+        public static bool IsValid(string color)
+        {
+            return Regex.Match(color, Pattern).Success;
+        }
         public string TextColor(string bgColor)
         {
             int nThreshold = 105;
--- a/BTCPayServer/Components/StoreSelector/Default.cshtml
+++ b/BTCPayServer/Components/StoreSelector/Default.cshtml
@@ -1,5 +1,10 @@
-@inject BTCPayServer.Services.BTCPayServerEnvironment _env
-@inject SignInManager<ApplicationUser> _signInManager
+@using Microsoft.AspNetCore.Mvc.TagHelpers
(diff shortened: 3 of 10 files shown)


dstrukt:
Oh shit, let's go!

Will take a look here shortly. 

Have you changed your mind about the `Brand Color` attribute, and the potential scope creep of it .. or would this just be relegated to the Checkout Invoice?

dstrukt:
Few notes:
- [ ] This was my default, we should set it to the BTCPay brand green
Screen Shot 2022-09-17 at 5 14 08 PM

- [ ] We should link them to the Settings view as opposed to just showing a path
Screen Shot 2022-09-17 at 5 15 40 PM

However, with the second point, should we be enabling the Local File Storage by default to prevent having to configure .. or are there considerations here I'm not aware of?

- [ ] We should keep the ratio 1:1 crop I'm thinking .. ? Tried a horizontal ratio, and this is the result .. I guess the user would figure it out with enough tries, but considering we don't have a Preview (unless this is possible sans validation), it might be a little annoying, but it's a minor consideration.
Screen Shot 2022-09-17 at 5 20 16 PM

Otherwise, looks great - awesome work!

dstrukt (changes requested):
Few notes:
- [ ] This was my default, we should set it to the BTCPay brand green
Screen Shot 2022-09-17 at 5 14 08 PM

- [x] We should link them to the Server Settings view as opposed to just showing the path, no?
Screen Shot 2022-09-17 at 5 15 40 PM

However, with the second point, should we be enabling the Local File Storage by default to prevent having to configure .. or are there considerations here I'm not aware of?

- [x] We should keep the ratio 1:1 crop I'm thinking .. ? Tried a horizontal ratio, and this is the result .. I guess the user would figure it out with enough tries, but considering we don't have a Preview (unless this is possible sans validation), it might be a little annoying, but it's a minor consideration.
Screen Shot 2022-09-17 at 5 20 16 PM

- [ ] This might be out of scope, but if we go with the 1:1 square (only works if it's the 1:1 thought), we could also replace the icon here with the uploaded Store Icon
Screen Shot 2022-09-17 at 5 25 33 PM


Otherwise, looks great - awesome work!

dennisreimann:
> This was my default, we should set it to the BTCPay brand green

I've updated it, so that it displays "None" in case the brand color is undefined. I wouldn't set a default here, because that will make it harder for us to deal with the unset case: If it is undefined, the BTCPay primary color will get used anyways and we also benefit from having the variation colors for e.g. the hover states. 

> Have you changed your mind about the Brand Color attribute, and the potential scope creep of it

Things like hover states etc. are my particular concern with regards to defining the colors via the UI: We'll have to see if a "one color fits all" approach works good enough — I think we can make it happen and offer a more finegrained approach via CSS variables.

The other points are addressed, thanks for your review and feedback!

dstrukt:
>  I wouldn't set a default here, because that will make it harder for us to deal with the unset case

Sounds good!

> Things like hover states etc. are my particular concern with regards to defining the colors via the UI: We'll have to see if a "one color fits all" approach works good enough — I think we can make it happen and offer a more finegrained approach via CSS variables.

An alternative to this would be .. Pick a color:
- Blue
- Red
- Green
- Purple
- etc..
And we just utilize our current palette for those different states (hover, etc..), so there's still choice, but it's a bit more within our control. However, I can see people who want very specific control of the exact HEX value for their brand feeling a bit limited by this.. hmm.

dennisreimann:
Let's go with that one particular color as the easy customization option and then we'll offer more finegrained control via CSS variables – I think this will work.

dstrukt:
Don't know what you had in mind, or what's easier with our existing components, but we could roll with something like this .. or if not, potential inspiration:

Screen Shot 2022-09-18 at 2 18 14 PM

dstrukt:
Another thing I've noticed ... should we remove this functional aspect in the `Checkout Appearance` tab?

Screen Shot 2022-09-18 at 6 50 40 PM

dennisreimann:
> but we could roll with something like this .. or if not, potential inspiration for you in the process.

As you mentioned before, I think poeple would rather expect to pick their exact brand color than choose from a list of predefined ones. Don't worry, we'll make the hover states etc. work too, without adding more choices here. People that want to customize more will get a chance to do so by using CSS variables.

> should we remove this in Checkout Appearance tab once this configuration is hooked up to all of the other views...?

Good point! Once the migration to the new checkout is fully done, we'll remove that option - but we cannot do so while people will be able to opt-in to the nex checkout version.

NicolasDorier:
You aren't validating the `BrandColor`. I think you can use `ColorTranslator.FromHtml` for this.

NicolasDorier:
btw the color picker sucks. You can't copy paste a #rbg color you had from somewhere else, and need to convert manually to decimal :(

Kukks:
I think it is browser dependent so it is of varying quality

dennisreimann:
Updated in [6cabed5](https://github.com/btcpayserver/btcpayserver/pull/4131/commits/6cabed56bf833c588cf4f0494eec9bd496445d3f), but as you can see that made it more complex. As the color input also doesn't allow for unsetting the color, I'd go with @NicolasDorier's proposed solution of making it a text input that accepts only colors in HEX format and we can validate that like in the commit I referenced.

@dstrukt For the logo format: Shall we just drop that format requirement and display whatever gets uploaded? Of course we can constrain the width/height, but as we are going to display it above the store name I think we can be more flexible here. I'm sure people will start complaining, because their logo doesn't support a square format.

Kukks:
> Updated in [6cabed5](https://github.com/btcpayserver/btcpayserver/pull/4131/commits/6cabed56bf833c588cf4f0494eec9bd496445d3f), but as you can see that made it more complex. As the color input also doesn't allow for unsetting the color, I'd go with @NicolasDorier's proposed solution of making it a text input that accepts only colors in HEX format and we can validate that like in the commit I referenced.
> 
> @dstrukt For the logo format: Shall we just drop that format requirement and display whatever gets uploaded? Of course we can constrain the width/height, but as we are going to display it above the store name I think we can be more flexible here. I'm sure people will start complaining, because their logo doesn't support a square format.

i would argue any CSS color should be acceptable. see https://stackoverflow.com/a/33184805/275504

dennisreimann:
> i would argue any CSS color should be acceptable.

That's fine too, we'd have to validate with `ColorTranslator.FromHtml` then.

NicolasDorier:
I don't believe we should allow any CSS color. This is a moving target. `color: var(--myvar)` for example is a valid css color.
We should restrict as much as we can to avoid injections. This is such a small feature it's not worth it to be comprehensive here.

dennisreimann:
@NicolasDorier Updated it. For now we'll require a HEX value then — one can either enter it directly into the input or use the color picker. Bot inputs are synced.

grafik

NicolasDorier:
I fixed the pattern, which wasn't checking with `^` and `$` also, there was some warning in the js console when editing the color wrongly. Which I fixed.

@dennisreimann review and you can squash and merge this one.

PR #4150, opened 2022-09-21
Fix pagination of wallet's transactions

NicolasDorier:
Only the first page of the wallet's transaction was returning the expected result.
There was two bugs, one at BTCPay Server level, and one at NBX.
Since we can't guarantee NBX version, I apply a workaround for old version.

This a bug fixed from #4074 by @dennisreimann and @HamroRamro


dennisreimann (commented):
Shouldn't we also include the docker-compose update with the latest NBXplorer version?

simpson420b:
Where is my money

dennisreimann (approved):
tACK.

PR #4151, opened 2022-09-22
Remove redundant exception status from invoice state label

bolatovumar:
We are showing exception status of the invoice twice on the invoice details page if the status is one of those that can't be changed by the user. This is redundant because the main status string will also always contain the exception status.

|Before|After|
|---|---|
|Capture|after|


bolatovumar:
Failing test unrelated

dennisreimann (approved):

PR #4155, opened 2022-09-23
Improve currency selection

### LINKED ISSUE #4154: Improve Currency Selector
Expected Experience:
- Dropdown should appear without having to delete text first

Current Experience:
improve-payment-method-selector

dennisreimann:
Removes the current value on focus, so that the user gets to see the available options. If no selection or change is made, the value is reset to the previous value on blur.

Closes #4154.

https://user-images.githubusercontent.com/886/191958053-524f85f5-da6b-4782-944e-dae455e57ced.mp4

### DIFF
--- a/BTCPayServer/TagHelpers/CurrenciesSuggestionsTagHelper.cs
+++ b/BTCPayServer/TagHelpers/CurrenciesSuggestionsTagHelper.cs
@@ -13,30 +13,37 @@ namespace BTCPayServer.TagHelpers
         public CurrenciesSuggestionsTagHelper(CurrencyNameTable currencies)
         {
             _currencies = currencies;
+            
         }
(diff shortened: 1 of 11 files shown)


NicolasDorier:
Nice and simple workaround Concept ACK

dstrukt (approved):
Works exactly as discussed - beautiful!

tACK

woutersamaey:
Just curious. Are all browsers behaving the same way? I'm not very familiar with this dropdown feature.

dennisreimann:
@woutersamaey Yes, the display varies, but in general it works the same across browsers. See [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) for reference.

PR #4137, opened 2022-09-19
Form Builder

Kukks:
Closes #505

PR #4157, opened 2022-09-24
Checkout v2

### LINKED ISSUE #3567: Checkout page improvements

dennisreimann:
Close #3567

Builds on #4131 and potentially #4137. Posting this early WIP state so that everyone can follow along.

For now it's just the additional settings we talked about in the recent design meetings: The new checkout will be an opt-in version until we refined it enough to make it the default. 

https://user-images.githubusercontent.com/886/192082672-80b33bfe-6ff9-4a56-b711-c71a64a53e98.mp4

If the user opts-in to the new checkout version, they get access to the settings ­— for now:

- Choose which data to request from the user (will require #4137 or a simplified version of the new forms concept)
- Enable BIP21 (the setting is mirrored from the LN settings page), because it will play a more prominent role in v2

Once we've shipped an finalized the new checkout, we can clean up and eliminate/consolidate settings (like the logo URL and requiring a refund email).


dstrukt:
oooo baby!

Not sure of the best wording atm, but `full address` is a bit confusing to me...

Since it's already in the label, we can get away with less in the dropdown...
`Email address only`
`Physical address + email address` this might be too much, but it's more descriptive and accurate

dennisreimann:
Yeah, we should find concise wordings for these, as we'll also offer them as templates in the form PR.

grafik

pavlenex:
> Physical address + email address might be too much, but it's more descriptive and clear (imo)

Term `Delivery address` is commonly used in e-commerce.

dstrukt:
> Terms Delivery address or Shipping address is commonly used in e-commerce.

Here we go .. I need to stop checking GH on my phone. 

ACK

dennisreimann:
Some more updates and progress. @dstrukt and I talked through a few things and there'll be refinements arount the payment methods, timer and other minor things, but in general this is coming along nicely :)

## Top up invoice

Bildschirmfoto vom 2022-10-13 19-20-19

## Timer, partial payment details

Bildschirmfoto vom 2022-10-13 21-17-47

## Custom brand color

Bildschirmfoto vom 2022-10-13 22-09-33

woutersamaey:
Have you considered the button to activate the NFC reader? This button is currently part of the LNURL NFC plugin and appears as a weirdly big button at the bottom (probably because that's where the extension point is for plugins?), but I feel we need more/different options.

dstrukt:
Great call out @woutersamaey and yes we do, and will consider this as well!

NicolasDorier:
I really don't like the choice: Lightning or LNUrl
It's confusing as fuck. even me I don't know the difference, what I should choose as a customer. 
LNUrl doesn't need to be used at all for invoice that aren't top-up invoice, which is something normally we do I thought.

woutersamaey:
One more thing I'd like to point out: We should try to make the checkout fit on the most common smartphone screen size without need for scrolling. We can find out which size this is and attempt a compact view (if not possible a "compact" setting or "style" would also work for me).

The main reason are these smartphone-like POS devices. They run Android and show the checkout so the customer can scan at the store. Having to scroll here kills the experience.

I understand there are many POS like devices etc, but aiming for a compact view that works on 90%+ of common size devices would be a success. 

For POS you'd add the site/checkout to the home screen and thus the browser's menu bars do not appear, giving you more height. Actually the full screen height is available. 

While this maybe controversial and we should not target specific devices, I still feel the use case is valid.

I was even considering making a Javascript that detects how much % the content is too big and maybe scale the entire page a few percent for the cases where the page almost fits, but is still a few pixels too high. This is what we have now. The QR is in 99% visible, but you need to scroll like 5 pixels down to see it entirely. This sucks.

As I said, if too complex, I can settle for a separate "compact" style that could be POS only.

dstrukt:
@woutersamaey Great feedback, and yes I share this sentiment as well, and am attempting to make this as compact as possible in each step, despite the variable complexity and needs.

dstrukt:
Few notes:

> ## One payment method
> checkout-3

- [ ] View Details should be `semi-bold`
- [ ] The BTC balance should be the next larger font-size
- [ ] The fiat conversion should also be semibold
- [ ] Scan copy should be 1-2 ticks darker, need to check if it's our white, etc.. though
 
> ## Details opened
> checkout-2

- [ ] Improve the fiat formatting `1 BTC = 19,403.03 USD`

> ## Invoice paid
> checkout-4

- [ ] We should add more spacing between `the invoice has been paid`
- [ ] `Invoice has been paid` to `Invoice paid` .. imo, shorter and more concise

> ## Invoice expired
> checkout-5

- [ ] We should refine and reduce the `Invoice Expired` copy
- [ ] We should add more spacing between `Order ID` and the paragraph

woutersamaey:
Should we say "This invoice has been paid"? Why not shorter the shorter "Paid"?

Overall I don't like the word "invoice". I know it has meaning in Lightning and BTCPay uses it, but as far as the customer is concerned it is not an invoice from the merchant! So, we should not use this word on the checkout. We can say "payment" but not "invoice".

Thoughts?

dstrukt:
I had `Payment Complete` in the mocks, but regardless if we go that route, tend to agree it could be better.

woutersamaey:
> I think it makes sense to keep `Invoice expired` the way it is though.

Well, we're still not taking about a proper invoice (like in accounting, which the customer cares about) so I'd just put "expired" or "time to pay has expired". 

No need to get confusing and use the "invoice" word. 

Chances are the merchant has a "My Invoices" in the "My Account" section, so there's always the chance of collision with that.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -603,23 +604,26 @@ private async Task<AddressInvoiceData[]> GetAddresses(string[] selectedItems)
...
+        [ReferrerPolicy("origin")]
         public async Task<IActionResult> Checkout(string? invoiceId, string? id = null, string? paymentMethodId = null,
             [FromQuery] string? view = null, [FromQuery] string? lang = null)
         {
-            //Keep compatibility with Bitpay
-            invoiceId = invoiceId ?? id;
-            //
+            // Keep compatibility with Bitpay
+            invoiceId ??= id;
+
```
dennisreimann, line 614: Minor optimizations that were proposed by the IDE.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -783,12 +794,19 @@ public async Task<IActionResult> CheckoutNoScript(string? invoiceId, string? id
...
                 StoreId = store.Id,
                 AvailableCryptos = invoice.GetPaymentMethods()
-                                          .Where(i => i.Network != null)
+                                          .Where(i => i.Network != null &&
+                                              // TODO: These cases and implementation need to be discussed
+                                              (!storeBlob.UseNewCheckout ||
+                                                // Exclude LNURL for non-topup invoices
+                                                (invoice.IsUnsetTopUp() || i.GetId().PaymentType is not LNURLPayPaymentType)) &&
+                                                // Exclude Lightning if OnChainWithLnInvoiceFallback is active
+                                                (!storeBlob.OnChainWithLnInvoiceFallback || i.GetId().PaymentType is not LightningPaymentType)
```
dennisreimann, line 803: As stated in the TODO: These cases and implementation need to be discussed. We talked about them briefly on Mattermost, but this needs more discussion — see details in the general review comment.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -907,6 +925,14 @@ public async Task<IActionResult> UpdateCustomer(string invoiceId, [FromBody] Upd
             return Ok("{}");
         }
 
+        [HttpPost("i/{invoiceId}/Form")]
+        [HttpPost("invoice/Form")]
+        public IActionResult UpdateForm(string invoiceId)
+        {
+            // TODO: Forms integration 
+            return Ok();
```
dennisreimann, line 933: Open TODO which requires an implementation similar to the existing `UpdateCustomer` action.

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -500,6 +493,14 @@ public async Task<IActionResult> CheckoutAppearance(CheckoutAppearanceViewModel
                     PaymentMethod = paymentMethodId
                 });
             }
+
+            blob.UseNewCheckout = model.UseNewCheckout;
+            if (blob.UseNewCheckout)
+            {
+                blob.CheckoutFormId = model.CheckoutFormId;
+                blob.OnChainWithLnInvoiceFallback = model.OnChainWithLnInvoiceFallback;
+            }
```
dennisreimann, line 502: Change those settings only if the `UseNewCheckout` option is active, because they otherwise might get overridden with values that weren't actively set.

BTCPayServer/Models/BitpayCreateInvoiceRequest.cs
```diff
@@ -81,6 +81,9 @@ public class BitpayCreateInvoiceRequest
         public bool? RedirectAutomatically { get; set; }
         [JsonProperty(PropertyName = "requiresRefundEmail", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public bool? RequiresRefundEmail { get; set; }
+        
+        [JsonProperty(PropertyName = "checkoutFormId", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string CheckoutFormId { get; set; }
```
dennisreimann, line 86: I'm not entirely sure if this needs to be here.

BTCPayServer/Models/StoreViewModels/CheckoutAppearanceViewModel.cs
```diff
@@ -20,11 +21,21 @@ public void SetLanguages(LanguageService langService, string defaultLang)
...
             DefaultLang = chosen.Value;
         }
+        
         public SelectList Languages { get; set; }
 
+        [Display(Name = "Request customer data on checkout")]
+        public string CheckoutFormId { get; set; }
+
+        [Display(Name = "Include Lightning invoice fallback to on-chain BIP21 payment URL")]
+        public bool OnChainWithLnInvoiceFallback { get; set; }
```
dennisreimann, line 31: As the BIP21 setting plays an important roile in the new checkout, we mirror it here in the checkout appearance settings in addition to the Lightning settings.

BTCPayServer/Services/Stores/CheckoutFormSelectList.cs
```diff
@@ -0,0 +1,55 @@
...
+   [Display(Name = "Request email address only")]
+   Email,
+
+   [Display(Name = "Request shipping address")]
+   Address
+}
+
+public static class CheckoutFormSelectList
+{
+    public static SelectList ForStore(StoreData store, string selectedFormId, bool isStoreEntity)
```
dennisreimann, line 27: Introduced this as those options are shown in multiple places (store settings, POS settings, invoice creation). We can later on extend this with the custom forms defined by the user, which are also tied to a store.

BTCPayServer/Views/Shared/PointOfSale/UpdatePointOfSale.cshtml
```diff
@@ -72,19 +76,26 @@
...
+                    <label asp-for="CheckoutFormId" class="form-label"></label>
+                    <select asp-for="CheckoutFormId" class="form-select w-auto" asp-items="@checkoutFormOptions"></select>
+                    <span asp-validation-for="CheckoutFormId" class="text-danger"></span>
+                }
+                else
+                {
+                    <label asp-for="RequiresRefundEmail" class="form-label"></label>
+                    <select asp-for="RequiresRefundEmail" asp-items="@Html.GetEnumSelectList<RequiresRefundEmail>()" class="form-select w-auto"></select>
+                    <span asp-validation-for="RequiresRefundEmail" class="text-danger"></span>
+                }
```
dennisreimann, line 98: Display either the settings for the old or new checkout version, while they exist in parallel.

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
```diff
@@ -0,0 +1,545 @@
...
+        ? await FileService.GetFileUrl(Context.Request.GetAbsoluteRootUri(), Model.LogoFileId)
+        : Model.CustomLogoLink;
+}
+@functions {
+    private string PaymentMethodName(PaymentModel.AvailableCrypto pm)
+    {
+        return Model.AltcoinsBuild
+            ? $"{pm.PaymentMethodName} {pm.CryptoCode}"
+            : pm.PaymentMethodName.Replace("Bitcoin (", "").Replace(")", "").Replace("Lightning ", "");
+    }
```
dennisreimann, line 27: Quickfix for adjusting the payemnt method names while th old and new checkout exist in parallel. I didn't want to have the changes here affect the old version, so I figured it might be easiest to do those adjustments in the view.

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
```diff
@@ -0,0 +1,545 @@
...
+                            <p class="my-5 text-center">TODO: Forms integration -> {{srvModel.checkoutFormId}}</p>
+                        </template>
+                        <template v-else-if="srvModel.requiresRefundEmail">
+                            <p>{{$t("Contact_Body")}}</p>
+                            <div class="form-group">
+                                <label class="form-label" for="Email">{{$t("Contact and Refund Email")}}</label>
+                                <input class="form-control" id="Email" name="Email">
+                                <span class="text-danger" hidden>{{$t("Please enter a valid email address")}}</span>
+                            </div>
+                        </template>
```
dennisreimann, line 164: Open TODO: Proper form integration.

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
```diff
@@ -0,0 +1,545 @@
...
+    <script src="~/vendor/i18next/i18nextXHRBackend.js" asp-append-version="true"></script>
+    <script src="~/vendor/i18next/vue-i18next.js" asp-append-version="true"></script>
+    <script src="~/js/copy-to-clipboard.js" asp-append-version="true"></script>
+    <script src="~/main/utils.js" asp-append-version="true"></script>
+    <script src="~/checkout-v2/checkout.js" asp-append-version="true"></script>
+    @if (Env.CheatMode)
+    {
+        <partial name="Checkout-Cheating" model="@Model" />
+    }
+    <script>
```
dennisreimann, line 300: Open TODO: Move the JS parts of the new checkout out of the template into a separate JS file.

BTCPayServer/Views/UIInvoice/CreateInvoice.cshtml
```diff
@@ -90,9 +94,18 @@
...
+                    <label asp-for="CheckoutFormId" class="form-label"></label>
+                    <select asp-for="CheckoutFormId" class="form-select w-auto" asp-items="@checkoutFormOptions"></select>
+                    <span asp-validation-for="CheckoutFormId" class="text-danger"></span>
+                }
+                else
+                {
+                    <label asp-for="RequiresRefundEmail" class="form-label"></label>
+                    <select asp-for="RequiresRefundEmail" asp-items="@Html.GetEnumSelectList<RequiresRefundEmail>()" class="form-select w-auto"></select>
+                    <span asp-validation-for="RequiresRefundEmail" class="text-danger"></span>
+                }
```
dennisreimann, line 108: Display either the settings for the old or new checkout version, while they exist in parallel.

BTCPayServer/wwwroot/locales/en.json
```diff
@@ -47,7 +47,17 @@
...
+  "View Details": "View Details",
+  "QR_Text": "Scan the QR code, or tap to copy the address.",
+  "Total Price": "Total Price",
+  "Exchange Rate": "Exchange Rate",
+  "Recommended Fee": "Recommended Fee",
+  "Feerate": "{{feeRate}} sat/byte",
+  "Amount Paid": "Amount Paid",
+  "Amount Due": "Amount Due",
+  "Pay in wallet": "Pay in wallet",
+  "Invoice paid": "Invoice paid"
```
dennisreimann, line 62: New translations were added, we will try to keep additions minimal so that we can still leverage the existing translations.

BTCPayServer/wwwroot/main/site.css
```diff
@@ -84,15 +84,15 @@ a.unobtrusive-link {
...
+[aria-expanded] > svg.icon-caret-down {
     flex-shrink: 0;
     width: 24px;
     height: 24px;
     margin-left: auto;
     transition: transform 0.2s ease-in-out;
 }
 
-[data-bs-toggle="collapse"][aria-expanded="true"] svg.icon {
+[aria-expanded="true"] > svg.icon-caret-down {
```
dennisreimann, line 95: General adjustment that works backwards compatible and allows us to reuse those styles without Bootstrap components.

dennisreimann (commented):
As discussed in [yesterdays design call](https://www.youtube.com/watch?v=N_Kj8rH8DNc): The basics are done, there are still some things to address and refine — but before this PR grows even bigger, I'd like to invite feedback and technical review, so that we can merge this as a first step and get to the refinements in a second and maybe even third PR.

## Outline of the most important changes made in this PR

- Introduce the `UseNewCheckout` option to opt-in to Checkout v2. We'll have this as a non-default in v1.7 so that we can gather feedback from the reckless and brave users out there, who want to test the new version in prod.
- When enabled, this option will also give access to configure another new feature, which is currently also being developed in #4137: The `CheckoutFormId` is the Checkout v2 equivalent of `RequiresRefundEmail` and extends that functionality to request custom data from the customer. That's why a lot of changes in this PR introduce that setting to almost all places that also have the refund email setting.
- For the initial version we intend to only ship generic versions of the custom forms, with `Email` and `Address` being the first options. Those generic options will act as "shortcuts" which can be used without configuring custom forms for yourself — similar to the `InternalNode` option when configuring the Lightning Node without the connection string details.
- In conjunction with #4137 the `CheckoutFormId` can later contain either those generic form names or the custom form IDs. In the implementation here I tried to anticipate how this might work, appreciate @kukks' feedback here.
- The ViewModel and Controller part are shared between the old and new checkout (with some minor additions to the ViewModel), the rest of the changes happen mostly in the View part.

## To be discussed in review here

- Payment method exclusion rules and implentation:
  - LNURL should not be shown for non-topup invoices.
  - Omit Lightning as separate payment method, if `OnChainWithLnInvoiceFallback` is active. In the design calls we made the point to go in that direction to increase simplicity and push BIP21 compatibility, but this also needs to be discussed in terms of [wallet support](https://bitcoinqr.dev/#support).

## Open TODOs for separate PRs

- Tests for Checkout v2
- Refined the UI/UX of the payment methods
- Integrate the form builder templates
- Improve the cheating capabilities (expiring invoices and Lightning payments)
- Move the JS parts of the new checkout out of the template into a separate JS file
- Adjust the design so that the QR will be shown above the fold on smaller devices

pavlenex:
Tested it a bit, still will do a bit more, but seems like great work! 🎉 
1. What happened with `powered by BTCPay Server` link?
2. I think when invoice is expiring, we need to not just tell people it's expiring, but also make sure to tell them what to do? `Pay your invoice before it expires`?
Screenshot 2022-10-21 at 18 15 54
3. I think when invoice is partially paid, we'll have to work on some text, and perhaps make it more obvious, the text just blends in into the design. I still think we don't make it obvious that they need to pay the amount exactly show.
Screenshot 2022-10-21 at 18 11 35
5. I am still not convinced it'll be obvious to people that they have to click on a QR in order to copy the address, also not convinced they won't enter the fiat amount instead of exactly BTC amount. I don't have a clear suggestion about this, but I think we really have to think about this as we work on improving the v2 checkout page.

Overall, there are bits that we have to iron out, so I don't think any of my comments are blockers for merging this one and then iterating and improving it for 1.7.0

PR #4166, opened 2022-09-27
bump ms aspnet packages

Kukks:


NicolasDorier:
Can you bump the dockerfile sdk and runtime as well?

pavlenex:
On a dev call yesterday we agreed @NicolasDorier will do the bumps and merge this PR.

NicolasDorier:
Some bump of docker file were missing

Dezzj:
> Can you bump the dockerfile sdk and runtime as well?

PR #4171, opened 2022-09-27
Allow bind and port for http too

Kukks:


Kukks:
@NicolasDorier  The idea behind this PR is to allow using the bind and port env vars even when not using https

PR #4172, opened 2022-09-28
Refactor labels

NicolasDorier:
This PR change the backend of `Get/SetWalletInfo` and `Get/SetWalletTransactionsInfo` to keep same functionalities but backed by tables rather than blob.

There are three new tables:
* `WalletLabels`: At present, we only use this one to store the color of a label. The label is identified by it's text value.
* `WalletTaints`: A taint has a `WalletLabel` and get attached to a `WalletObject` we store here the current `LabelData` blobs.
* `WalletObjects`: Identify a `tx`, a `script` or an `outpoint` in the wallet. A `WalletObject` can have multiple taints. For transactions, we store `comment`.

There is currently no concept of taint in the UI or in the API. As such:
* `Removing a label` will remove all taints with a specific label from the transaction.
* `Adding a label` will create a taint with a `RawLabel` and a random color. If the a taint with same label already exists, this operation does nothing.
* `Getting a label` will return the first taint of the object having this label.

To do In later PR:
* `Get/SetWalletInfo` and `Get/SetWalletTransactionsInfo` aren't adapted to the new backend, and the logic is quite convoluted as a consequence. In a later PR we should replace those calls by direct queries to database or adding `AddLabel/RemoveLabel/SetWalletObjectBlob/GetWalletObjectBlob` methods to the `WalletRepository`.
* Renaming `*LabelData` by `*TaintData`.
* Implement the notion of stickiness. So a taint can be on an address, and stick to any transaction sending money to this address and its descendants.

The following tables are now unused and deprecated: `Wallets`, `WalletTransactions`.


John45901 (approved):

NicolasDorier:
Probably scrapping this PR... I think the design too complicated, will do another one.

PR #4180, opened 2022-09-29
Greenfield: Get Lightning invoices

dennisreimann:
Matching the data added in btcpayserver/BTCPayServer.Lightning#98 and btcpayserver/BTCPayServer.Lightning#99.


NicolasDorier:
@dennisreimann I adjusted some things:
1. Do not make the client assume the default value of `pendingOnly`.
2. The customRecord shouldn't be null if empty. If I understand, it's used in KeySend? If so, if KeySend isn't used, it should be null, and otherwise it should be set, even if empty.

You can squash/merge if you are fine with it.

dennisreimann:
Thanks for the adjustments!

PR #4126, opened 2022-09-16
POS - CustomAmount disabled by default

daviogg:
https://github.com/btcpayserver/btcpayserver/issues/4118

PR #4185, opened 2022-09-30
POS: Fix keypad view without custom amount

### LINKED ISSUE #4183: [Bug]: Invoice page is not shown for new Apps where custom payments are not checked

v1.6.11.0

Lunanode deployment.

### What happened?

Keypad POS doesn't generate invoice page if you don't check the "User can input custom amount" option (which is hidden for that POS type).

1. I created a new POS;
2. Removed all "tea" example items;
3. Picked the "Keypad only" Point-of-sale style;
4. Saved pos, opened pos;
5. After I entered amount and clicked "pay", everything disappeared, except the POS title. No qr code, no invoice page, no keypad.

The only way I could make it work as expected (to show the invoice page) was to switch to the "Item list only" style and check the "User can input custom amount" (it is hidden otherwise). Then I selected the "Keypad only" style and saved the POS. And it started to work as expected.

This is 100% a regression as it was working previously.

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
The custom amount option was disabled by default in #4126. This requires some additional adaptations in the post action as otherwise the invoice won't be generated.

Fixes #4183.

### DIFF
--- a/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
+++ b/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
@@ -109,7 +109,7 @@ namespace BTCPayServer.Plugins.PointOfSale.Controllers
         [DomainMappingConstraint(AppType.PointOfSale)]
         [RateLimitsFilter(ZoneLimits.PublicInvoices, Scope = RateLimitsScope.RemoteAddress)]
         public async Task<IActionResult> ViewPointOfSale(string appId,
-                                                        PosViewType viewType,
+                                                        PosViewType? viewType,
                                                         [ModelBinder(typeof(InvariantDecimalModelBinder))] decimal? amount,
                                                         string email,
                                                         string orderId,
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Seems like this one require a new version.

dennisreimann:
@bolatovumar can you review this and see if #4199 is still necessary then?

BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
```diff
@@ -129,12 +129,14 @@ public async Task<IActionResult> ViewPointOfSale(string appId, PosViewType? view
                 return NotFound();
             var settings = app.GetSettings<PointOfSaleSettings>();
             settings.DefaultView = settings.EnableShoppingCart ? PosViewType.Cart : settings.DefaultView;
-            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && settings.DefaultView != PosViewType.Cart)
+            var currentView = viewType ?? settings.DefaultView;
+            if (string.IsNullOrEmpty(choiceKey) && !settings.ShowCustomAmount && 
+                currentView != PosViewType.Cart && currentView != PosViewType.Light)
```
bolatovumar, line 134: Why `currentView != PosViewType.Cart` needed here? I think the only app where `!settings.ShowCustomAmount` shouldn't have an impact is `PosViewType.Light` since this setting doesn't even make sense for this app and you can't even select it. For all other apps, however, NOT having a `choiceKey` AND having `ShowCustomAmount` be `false` is an error.

bolatovumar (approved):
Issue seems fixed now.

BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
dennisreimann: Because when you submit a cart, there also is no `choiceKey`: The list of items (or a single item) are part of the `posData` variable. That's why it's been like that before — this PR only added the `currentView != PosViewType.Light` part.

PR #4191, opened 2022-10-05
Date display improvements

dennisreimann:
Fixes styling issues introduced in #4074, because the `max-width` was to small for localized dates.

Also adds the ability to choose the prefered initial display format, which can be the localized or relative date.


NicolasDorier:
This issue was happening also before the refactoring there btw

PR #4197, opened 2022-10-06
Make sure string is valid URL before rendering it as such in invoice details POS data section

bolatovumar:
I noticed that we output anything that sort of looks like a URL as if it was actually a URL in the POS data section of invoice details. However, since the method we use doesn't actually make sure that something is a valid URL, random stuff ends up being rendered as a URL. This PR fixes it by making sure that something is a valid URL before outputting it on the page.

|Before|After|
|---|---|
|before|after|

### DIFF
--- a/BTCPayServer/Views/UIInvoice/PosData.cshtml
+++ b/BTCPayServer/Views/UIInvoice/PosData.cshtml
@@ -1,6 +1,15 @@
 @model (Dictionary<string, object> Items, int Level)
 
-<table class="table table-hover my-0">
+@functions{
+    public bool IsValidURL(string source)
+    {
+        Uri uriResult;
+        return Uri.TryCreate(source, UriKind.Absolute, out uriResult) && 
+            (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps);
+    }
+}
+
+<table class="table my-0">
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/UIInvoice/PosData.cshtml
```diff
@@ -1,6 +1,15 @@
 @model (Dictionary<string, object> Items, int Level)
 
-<table class="table table-hover my-0">
+@functions{
+    public bool IsValidURL(string source)
+    {
+        Uri uriResult;
+        return Uri.TryCreate(source, UriKind.Absolute, out uriResult) && 
+            (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps);
```
Kukks, line 8: This is interesting and makes sense from an initial glance but what if you wanted to add a deep link?

Kukks (approved):
Ack, my only comment was about the scheme whitelisting and it's not a serious one for now.

BTCPayServer/Views/UIInvoice/PosData.cshtml
dennisreimann: imho requiring a scheme here is a fair assumption for links, ACK from my side.

PR #4199, opened 2022-10-06
Fix custom amount disabled keypad app redirect

### LINKED ISSUE #4198: [Bug]: Keypad app doesn't redirect to checkout if "User can input custom amount" is disabled

1.6.11

Locally

### What happened?

Can't make a payment in the keypad app because the redirect to checkout never happens if  "User can input custom amount" option is disabled.

1. Create a PoS app
2. Select the "Keypad only" view
3. Open the app in keypad view
4. Try to make a payment
5. Notice you just get redirected to the static app instead of to the checkout
6. Now go back to app settings
7. Change the view to anything else
8. Enable the "User can input custom amount" toggle
9. Switch the view back to "Keypad only"
10. Save the app settings
11. Repeat steps 3 and 4
12. Notice you can now make a payment because you are properly redirected to checkout view

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

bolatovumar:
Fix #4198

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
@@ -296,7 +296,7 @@
                     "showCustomAmount": {
                         "type": "boolean",
                         "description": "Whether to include a special item in the store which allows user to input a custom payment amount",
-                        "default": true,
+                        "default": false,
                         "nullable": true
                     },
                     "showDiscount": {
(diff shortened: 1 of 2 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
```diff
@@ -296,7 +296,7 @@
                     "showCustomAmount": {
                         "type": "boolean",
                         "description": "Whether to include a special item in the store which allows user to input a custom payment amount",
-                        "default": true,
+                        "default": false,
```
bolatovumar, line 299: The behavior was updated recently to have this default to `false` but docs weren't updated accordingly.

PR #4205, opened 2022-10-10
Import xpub: Surface error details

dennisreimann:
Checks if the input is an output descriptor and explicitely handles that case instead of catching any errors. This allows us to display more detailed information about why an import might fail.

## Example

multisig-import-01

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.Onchain.cs
+++ b/BTCPayServer/Controllers/UIStoresController.Onchain.cs
@@ -119,17 +119,16 @@ namespace BTCPayServer.Controllers
                             accountSettings.AccountKeyPath =
                                 vm.KeyPath == null ? null : KeyPath.Parse(vm.KeyPath);
                             accountSettings.RootFingerprint = string.IsNullOrEmpty(vm.RootFingerprint)
-                                ? (HDFingerprint?)null
-                                : new HDFingerprint(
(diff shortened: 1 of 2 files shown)


samuelmonday6:
At TGS, we are a community of Traders that learn and earn together 1) 100
The trading journey becomes more fun when you have like minded people to experience it with!
We help each other grow and earn
The Best Traders in the country are living the trading life here. What are you waiting for?
We at TRADINGGAMESTRONG provide people with services that not only help them in learning a skill but also a way of making an impact in their lives by creating a path to financial freedom.
Trading is a skill you can learn but only with efficient practice complete knowledge you can make good returns to your finances.
Comment YES to trade in a COMMUNITY
Follow https://www.instagram.com/p/CgXj835qPP3/?igshid=NjZiMGI4OTY= for daily stock markets education.
#traders #traderlifestyle #daytrader #forexprofit #stocksportant #stockmarketinvesting #stockmarketindia #intradaytrading #banknifty #forextraders #forexgroup #forexanalysis #tradingstocks
#tradingsignals #tradingplan #trading #indianstockmarket #tradingstrategy #tradingprofit #tradingcourse #tradingcommunity #stockmarkettips #financialfreedom #stockmarketeducation #stockmarketcourse #stockmarketcourses #forex #nifty50 #forexlifestyle #gs

PR #4215, opened 2022-10-17
Fix: PayjoinController could throw HTTP 500 of a few corner cases

NicolasDorier:
I activated nullable on `PayJoinEndpointController`, finding it was hard to review correctness during another refactoring, and I found that a few corner cases were not properly handled, and might throw error 500.

I am unsure if those corner cases could be used though, but it was too difficult to analyze, so I just fixed the warnings with nullable just in case.

On top of this `PayjoinReceiverContext.Invoice` was never set, which mean we weren't saving some of the logs happening for the invoice during a payjoin.


Kukks (approved):
look ok

PR #4222, opened 2022-10-19
Remove the bundle minifier

NicolasDorier:
This remove dependency on the bundle minifier project. (https://github.com/madskristensen/BundlerMinifier)
* The project is barely maintained
* BTCPay docker install uses by default HTTP2 and gzip compression, minifying further complicate debugging for no meaningful user improvement.
* Increase the dev loop as there isn't an additional build step for bundling the javascript files.

In the future I am open to add nodejs tooling which is more mature for this kind of work, but any dependency should only be at dev-time for front end, not build time.
The artifacts of such tooling would need to be commited into git.

Script used to update the cshtml files:
```csharp
using BundlerMinifier;
using Newtonsoft.Json;
using System.Text.RegularExpressions;

var b = BundleHandler.GetBundles(@"C:\Sources\btcpayserver\BTCPayServer\bundleconfig.json");

var bundles = b.ToDictionary(c => c.OutputFileName, c => c.GetAbsoluteInputFiles(true));


foreach (var file in EnumerateFiles(@"C:\Sources\btcpayserver\BTCPayServer\Views"))
{
	var text = File.ReadAllText(file);
	var matches = Regex.Matches(text, "(\\s*?)<bundle name=\"(.*?)\".*");
	HashSet<string> includedScripts = new HashSet<string>();
	bool changed = false;
	foreach (var match in matches.OfType<Match>())
	{	
		if (match.Success)
		{
			var scripts =
				bundles[match.Groups[2].Value]
				.Select(i => new
				{
					Spaces = match.Groups[1].Value.Replace("\r\n","").Replace("\n", ""),
					Path = i.Replace(@"C:\Sources\btcpayserver\BTCPayServer\wwwroot\", "").Replace('\\', '/'),
					IsJs = i.EndsWith(".js")
				})
				.Where(i => includedScripts.Add(i.Path))
				.Select(i => 
				i.IsJs ? $"{i.Spaces}<script src=\"~/{i.Path}\" asp-append-version=\"true\"></script>" :
						 $"{i.Spaces}<link href=\"~/{i.Path}\" asp-append-version=\"true\"  rel=\"stylesheet\" />")
				.ToArray();

			var scriptsStr = "\r\n" + String.Join("\n", scripts) + "\r\n";
			text = text.Replace(match.Value, scriptsStr);
			changed = true;
		}
	}
	if (changed)
	{
		Console.WriteLine("Changed " + file);
		File.WriteAllText(file, text);
	}
}

IEnumerable<string> EnumerateFiles(string folder)
{
	foreach (var file in Directory.GetFiles(folder, "*.cshtml"))
	{
		yield return file;
	}
	foreach (var subfolder in Directory.GetDirectories(folder))
	{
		foreach (var file in EnumerateFiles(subfolder))
		{
			yield return file;
		}
	}
}
```


woutersamaey:
Concept ACK, like we discussed

dennisreimann:
Concept ACK. What I also really like about this is that it makes us more flexible when adding or removing single files.

PR #4224, opened 2022-10-20
Allow plugins to do something before and after automatic payouts

woutersamaey:
This PR adds 2 new hooks for plugins so they can do something:
- Right before automatic payouts occur
- After automatic payouts occurred

I need this for a plugin where I want to quickly auto-approve or reject a previous approval before the payout executes.

### DIFF
--- a/BTCPayServer/PayoutProcessors/AfterPayoutFilterData.cs
+++ b/BTCPayServer/PayoutProcessors/AfterPayoutFilterData.cs
@@ -1,0 +1,19 @@
+using System.Collections.Generic;
+using BTCPayServer.Data;
+using BTCPayServer.Payments;
+
+namespace BTCPayServer.PayoutProcessors;
+
+public class AfterPayoutFilterData
+{
+    private readonly StoreData _store;
+    private readonly ISupportedPaymentMethod _paymentMethod;
+    private readonly List<PayoutData> _payoutDatas;
+
+    public AfterPayoutFilterData(StoreData store, ISupportedPaymentMethod paymentMethod, List<PayoutData> payoutDatas)
+    {
+        _store = store;
+        _paymentMethod = paymentMethod;
+        _payoutDatas = payoutDatas;
+    }
+}
--- a/BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
+++ b/BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
 using BTCPayServer.HostedServices;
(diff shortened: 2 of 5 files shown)


Kukks (changes requested):
I agree with the concept but the issue is the filter data is using types which are not accessible to plugins as they live inside the btcpayserver project. I would focus on providing data in more generic, primitive types for now

woutersamaey:
@Kukks which type would you recommend to encapsulate these multiple values? Should we use a HashTable or Dictionary?

extraterestrial1:
Your shit sucks it doesn't payout and if y'all they to scam my phone one more fucking time I'll report your shit to msrc cyber security devision as well as trade commission for new York defense btw the eu usa and swiss usa crypto trx departments and ill sew for every 6 times y'all have offered me 22k in bit as well don't scam any one else if i hear anyone i know get another duck as s email from y'all i swear I'll shit this project down do you understand me you can get ahold of me at extraterestrial1@extrateredtrialsbanking.co

pavlenex:
> Your shit sucks it doesn't payout and if y'all they to scam my phone one more fucking time I'll report your shit to msrc cyber security devision as well as trade commission for new York defense btw the eu usa and swiss usa crypto trx departments and ill sew for every 6 times y'all have offered me 22k in bit as well don't scam any one else if i hear anyone i know get another duck as s email from y'all i swear I'll shit this project down do you understand me you can get ahold of me at [extraterestrial1@extrateredtrialsbanking.co](mailto:extraterestrial1@extrateredtrialsbanking.co)

Hey, The bitcoin bonus scam does use our software, but we don't have any control over their installation of it. BTCPay is a self-hosted, free and open source software anyone can use. You can report them to their domain registrant (we did this already) or simply local law enforcement. Please do not insult contributors here, as your rage (which I completely understand) is directed at the wrong side. https://docs.btcpayserver.org/FAQ/General/#what-if-i-have-a-problem-with-a-paid-invoice

woutersamaey:
> I agree with the concept but the issue is the filter data is using types which are not accessible to plugins as they live inside the btcpayserver project. I would focus on providing data in more generic, primitive types for now

Is this still needed @Kukks ? I believe plugins can now reference these objects so my PR is good to go?

Kukks:
> I should probably change the namespaces where it says `Abstractions` right? But besides that, we should be ok?
Yes

Just one thing, these will only apply to situations where the payouts are handled by these specific builtin payout processors. Anything using the API or doing it outside of these wont trigger these actions for you. For example, the coinjoin plugin automatically batches payouts using its own mechanism ( I could not get it to fit with a payout processor implementation unfortunately)

woutersamaey:
That's fine @Kukks 
These hooks are meant to be for the automatic payout processors only. Not for ALL outgoing payments. I understand that may not be feasible (you could connect to your node and do a transaction behind BTCPay's back) and it was never the intention.

A simple use case could be a plugin that logs the automated payouts once they have run...

woutersamaey:
I rebased this PR and waiting for tests to pass

NicolasDorier:
I don't think event aggregator work in your case, because as far as I understand, you want your hook to be processed synchronously to change the data. EventAggregator is only adapted when the subscriber are asynchronous.

woutersamaey:
> I don't think event aggregator work in your case, because as far as I understand, you want your hook to be processed synchronously to change the data. EventAggregator is only adapted when the subscriber are asynchronous.

Correct.

Kukks (approved):

Issue #4231, opened 2022-10-24
[Bug]: 


NicolasDorier:
### What is your BTCPay version?

1.6.12

### How did you deploy BTCPay Server?

docker

### What happened?

Errors on the swagger schema reported by user.
```
Errors: 
        -paths.'/api/v1/stores/{storeId}/custodian-accounts'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
        -attribute components.schemas.CustodianData.items is missing
        -paths.'/api/v1/stores/{storeId}/custodian-accounts/{accountId}'. Declared path parameter accountId needs to be defined as a path parameter in path or operation level
        -paths.'/api/v1/stores/{storeId}/users'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
        -paths.'/api/v1/stores/{storeId}/custodian-accounts/{accountId}'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
Warnings: 
        -paths.'/api/v1/stores/{storeId}/custodian-accounts'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
        -attribute components.schemas.CustodianData.items is missing
        -paths.'/api/v1/stores/{storeId}/custodian-accounts/{accountId}'. Declared path parameter accountId needs to be defined as a path parameter in path or operation level
        -paths.'/api/v1/stores/{storeId}/users'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
        -paths.'/api/v1/stores/{storeId}/custodian-accounts/{accountId}'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
```
this is using maven's plugin openapi-generator-maven-plugin

### How did you encounter this bug?

using https://docs.btcpayserver.org/API/Greenfield/v1/swagger.json

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

NicolasDorier:
It might be caused by https://docs.btcpayserver.org/API/Greenfield/v1/swagger.json not being in sync with the latest version from
https://mainnet.demo.btcpayserver.org/swagger/v1/swagger.json

Ping @dennisreimann

NicolasDorier:
on mainnet, user reported 
```
Errors: 
        -attribute components.schemas.CustodianData.items is missing
        -paths.'/api/v1/stores/{storeId}/users'. Declared path parameter storeId needs to be defined as a path parameter in path or operation level
Warnings: 
[...]

PR #4235, opened 2022-10-25
Fix OpenAPI 3.0 validation errors and warnings

dennisreimann:
This is not a complete fix of all warning, as 16 warnings remain left — nevertheless, together with btcpayserver/btcpayserver-doc#1177 this is a stab at solving problems like described in #4231.

This is the command I ran — it uses `swagger-cli` for validation (no problems) and `redocly/cli` for linting. The latter is used as it surfaces more problems than just the validation.

```sh
# go to Swagger directory
cd BTCPayServer/wwwroot/swagger/v1

# deep-merge all files using jq
jq -rs 'def deepmerge(a;b):
  reduce b[] as $item (a;
    reduce ($item | keys_unsorted[]) as $key (.;
      $item[$key] as $val | ($val | type) as $type | .[$key] = if ($type == "object") then
        deepmerge({}; [if .[$key] == null then {} else .[$key] end, $val])
      elif ($type == "array") then
        (.[$key] + $val | unique)
      else
        $val
      end)
    );
  deepmerge({}; .)' swagger.template.* > openapi.json
  
# validatee using the swagger-cli tool
npx swagger-cli validate openapi.json

# lint using the redocly-cli tool
npx @redocly/cli lint openapi.json
```

It's a bit cumbersome to do the merging with `jq` and the proper way to do this would be to have a root file, which references and includes the other files. That's also what the OpenAPI tools support, as this is a native feature of the spec. The downside is, that we'd have to split the files differently.

We might get to reshape this, once we [upgrade to OpenAPI 3.1](https://www.apimatic.io/blog/2021/09/migrating-to-and-from-openapi-3-1/), which also seems to have proper webhooks support.


NicolasDorier:
@dennisreimann I think a good idea for getting those issues fixed once for all, would be to include the node tooling in a docker container, and with docker exec, in a unit test, running those linters.

I think I'll do at one point.

PR #4237, opened 2022-10-25
UI: Improve access token pairing

### LINKED ISSUE #4133: Consolidate Access Token Pairing Permission view
Currently when creating an Access Token, this is the flow

1. Hit `Create Token` CTA
2. Token Details
Screen Shot 2022-09-18 at 7 17 42 PM
3. Choose Store Pairing
Screen Shot 2022-09-18 at 7 19 04 PM

I can see an instance of why you might want to create a token for another store from this view, but it seems unnecessary and redundant.

I propose removing that Pairing view, and just associating the token to the current store you're creating it from.
- Removes the additional Pairing view step
- Less bugs or permissions, etc.. to potential worry about, and/or test

Open to hearing opposing opinions as well, but wanted to toss this one out there.

dennisreimann:
Closes #4133 and makes the access token pairing UI work like the Authorize UI (#3880).

If this is accessed with the store as context, we show it withing the general UI and without store selection:

grafik

If the request is coming from an external app, we show it in the wizard UI and with store selection:

grafik

The controller part is mostly cleanups.

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.cs
+++ b/BTCPayServer/Controllers/UIStoresController.cs
@@ -109,6 +109,10 @@ namespace BTCPayServer.Controllers
         private readonly EventAggregator _EventAggregator;
         private readonly IOptions<ExternalServicesOptions> _externalServiceOptions;
 
+        public string? GeneratedPairingCode { get; set; }
+        public WebhookSender WebhookNotificationManager { get; }
+        public IDataProtector DataProtector { get; }
+
         [TempData]
         public bool StoreNotConfigured
         {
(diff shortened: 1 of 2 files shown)


dstrukt (approved):
LGTM!

Some form of bread-crumbing is something we'll definitely need to address after 1.7.0, but this is a great optimization and clean-up.

dennisreimann:
@ndeet and I talked about this briefly yesterday: It seems like that pairing view is just used for legacy BitPay style integrations, so we shouldn't give it more effort than necessary, as its usage is declining anyways.

PR #4238, opened 2022-10-26
Add Greenfield invoice refund endpoint

bolatovumar:
This PR adds an ability to refund an invoice through Greenfield API by POSTing to `/api/v1/stores/{storeId}/invoices/{invoiceId}/payment-methods/{paymentMethod}/refund`.

See discussion here: https://github.com/btcpayserver/btcpayserver/discussions/4181

### DIFF
--- a/BTCPayServer.Client/Models/RefundInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/RefundInvoiceRequest.cs
@@ -1,0 +1,25 @@
+#nullable enable
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace BTCPayServer.Client.Models
+{
+    public enum RefundVariant
+    {
+        RateThen,
+        CurrentRate,
+        Fiat,
+        Custom,
+        NotSet
+    }
+
+    public class RefundInvoiceRequest
+    {
+        public string? Name { get; set; } = null;
+        public string? Description { get; set; } = null;
+        [JsonConverter(typeof(StringEnumConverter))]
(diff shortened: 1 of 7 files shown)


BTCPayServer.Client/Models/RefundInvoiceRequest.cs
```diff
@@ -0,0 +1,25 @@
...
+        Custom,
+        NotSet
+    }
+
+    public class RefundInvoiceRequest
+    {
+        public string? Name { get; set; } = null;
+        public string? Description { get; set; } = null;
+        [JsonConverter(typeof(StringEnumConverter))]
+        public RefundVariant RefundVariant { get; set; } = RefundVariant.NotSet;
```
bolatovumar, line 21: I made this required instead of defaulting to `RefundVariant.RateThen` like what's done in the UI. I'm not sure about this, maybe we should make this optional and default to `RefundVariant.RateThen`?

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -240,16 +240,6 @@ public async Task<IActionResult> InvoiceReceipt(string invoiceId)
             var network = _NetworkProvider.GetNetwork(paymentMethodId.CryptoCode);
             return network == null ? null : paymentMethodId.PaymentType.GetTransactionLink(network, txId);
         }
-        bool CanRefund(InvoiceState invoiceState)
```
bolatovumar, line 243: Moved this to `BTCPayServer/Services/Invoices/InvoiceEntity.cs` so it can be re-used.

PR #4252, opened 2022-10-31
Greenfield: Graceful return for in-flight HTLCs

### LINKED ISSUE #3781: [Greenfield API]: Graceful return for in-flight HTLCs
When invoking the Pay Lightning Invoice from Internal Node endpoint (/api/v1/server/lightning/{cryptoCode}/invoices/pay) and the HTLC gets stuck mid-flight, the application times out and no invoice meta data is returned.

Without the invoice id being returned, it's not possible to check the status later to see if the payment settles or fails.

Likely also affects the Store Lightning Pay Invoice endpoint (/api/v1/stores/{storeId}/lightning/{cryptoCode}/invoices/pay) although I haven't tested it.

Edit: Happens on both CLN and LND.

Suggestions/Ideas:

- Always return a response with an invoice ID and a status when invoking this end-point
- Get invoice endpoint (/api/v1/server/lightning/{cryptoCode}/invoices/{id}) lookup by Bolt11 or invoice ID
- Parameter to set timeout
- New endpoint to Get All Invoices on the node to fetch the invoice id

dennisreimann:
Based on btcpayserver/BTCPayServer.Lightning#106 this closes #3781.


BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.cs
```diff
@@ -222,16 +221,26 @@ public virtual async Task<IActionResult> PayInvoice(string cryptoCode, PayLightn
...
+                    SendTimeout = lightningInvoice.SendTimeout
+                }
                 : null;
             var result = await lightningClient.Pay(lightningInvoice.BOLT11, param, cancellationToken);
             
-            if (result.Result == PayResult.Ok && bolt11?.PaymentHash is not null)
+            if (result.Result is PayResult.Ok or PayResult.Unknown && bolt11?.PaymentHash is not null)
             {
+                // get a new instance of the LN client, because the old one might have disposed its HTTPClient
+                lightningClient = await GetLightningClient(cryptoCode, true);
```
dennisreimann, line 239: Not sure why this happens, but I ran into cases in which the HTTP client of the LN client has been disposed. Without getting a new instance it then crashed.

BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.cs
```diff
@@ -241,19 +250,25 @@ public virtual async Task<IActionResult> PayInvoice(string cryptoCode, PayLightn
                     CreatedAt = payment.CreatedAt,
                     TotalAmount = payment.AmountSent,
                     FeeAmount = payment.Fee,
-                });
+                };
+                return result.Result is PayResult.Ok ? Ok(data) : Accepted(data);
```
dennisreimann, line 254: Decided to also go for another HTTP status here in case the payment was only initiated (Accepted) instead of successfully sent (Ok).

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
```diff
@@ -281,6 +281,8 @@ public async Task<IActionResult> ProcessLightningPayout(string cryptoCode, strin
             var proofBlob = new PayoutLightningBlob() {PaymentHash = bolt11PaymentRequest.PaymentHash.ToString()};
             try
             {
+                // TODO: Incorporate the changes from this PR here:
+                // https://github.com/btcpayserver/BTCPayServer.Lightning/pull/106
```
dennisreimann, line 285: @Kukks We can update it in this one or a later PR.

Kukks (approved):

NicolasDorier:
@dennisreimann there is conflict on the swagger template.
Also can you update the route's description to explain what happen in case the payment isn't immediately terminated?

PR #4274, opened 2022-11-07
Greenfield: Wallet Objects

Kukks:
This exposes the new wallet object/link system through Greenfield


BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -2713,24 +2713,102 @@ await adminClient.GenerateOnChainWallet(admin.StoreId, "BTC", new GenerateOnChai
...
+        [Fact(Timeout = 60 * 2 * 1000)]
         [Trait("Integration", "Integration")]
-        public async Task CustodiansControllerTests()
+        public async Task CanUseWalletObjectsAPI()
         {
             using var tester = CreateServerTester();
             await tester.StartAsync();
-            await tester.PayTester.EnableExperimental();
-            var unauthClient = new BTCPayServerClient(tester.PayTester.ServerUri);
-            await AssertHttpError(401, async () => await unauthClient.GetCustodians());
```
dennisreimann, line 2726: Does this remove the CustodiansControllerTests?

dennisreimann (commented):
What are wallet object? Docs seem to be missing.

BTCPayServer.Tests/GreenfieldAPITests.cs
Kukks: fixed

Kukks:
> What are wallet object? Docs seem to be missing.

Wallet objects is the new concept of how we can store data around an on-chain wallet (and replaces the underlying store for labels). See #4179

Kukks:
> > What are wallet object? Docs seem to be missing.
> 
> Wallet objects is the new concept of how we can store data around an on-chain wallet (and replaces the underlying store for labels). See #4179

Will add docs once we agree API looks ok

NicolasDorier:
Please in future PR stop doing all those code formatting change, it makes it exhausting to review the PR :(

NicolasDorier:
API seems ok for me

Kukks:
> Please in future PR stop doing all those code formatting change, it makes it exhausting to review the PR :(

Yeah, sorry, it comes as a habit, and then realize when I'm knee-deep in complex changes. Isn't there a way to specify in the solution how formatting should be done so that everything doesn't change between us?

dennisreimann:
>  Isn't there a way to specify in the solution how formatting should be done so that everything doesn't change between us?

I'm also guilty of this. Let's see how we can improve this after v1.7

ndeet:
For the php library we use .editorconfig and php-cs-fixer can be run locally to check and fix formatting. It also runs on [each PR](https://github.com/btcpayserver/btcpayserver-greenfield-php/blob/master/.github/workflows/code_style.yml) and formatting needs to be fixed before anything is merged. 

Not sure for C# and what you already tried but what I found quickly is this https://dev.to/felipetofoli/code-formatting-for-c-projects-dotnet-format-github-actions-54p8 which looks like similar what we have. You could introduce it after a big initial PR runs it on all the current codebase.

Issue #4261, opened 2022-11-02
Checkout v2 finetuning


dennisreimann:
Finish the open TODOs from #4157:

- [x] Show payment details on result pages, see #4240
- [x] Integrate the form builder templates from #4137 
    - We decided to separate those steps and have the form after the invoice has been paid.
- [x] Adjust the design so that the QR will be shown above the fold on smaller devices
- [x] Improve the cheating capabilities
    - [x] Mining block
    - [x] Lightning payments
    - [x] Expiring invoices
- [x] Move the JS parts of the new checkout out of the template into a separate JS file
- [x] Plugins integration
- [x] Refine the UI/UX of the payment methods
- [x] Tests for Checkout v2

Some more [feedback from Pav](https://github.com/btcpayserver/btcpayserver/pull/4157#issuecomment-1287185881):

- [x] What happened with powered by BTCPay Server link?
    - It was missing, fixed! 
- [x] If no logo is sent should we default to BTCPay Logo? Looks kinda empty and not fun without a logo?
    - In the design meeting we discussed this and decided to just omit the logo in that case. This is mainly to not confuse the customer, because BTCPay isn't selling anything.
- [x] Consider adding logo recommended size
- [x] Add hover icon over amount, just like we have for the QR code, to make it easier for people to know amount can be copied by clicking.
- [x] When invoice is expiring, we need to not just tell people it's expiring, but also make sure to tell them what to do? Pay your invoice before it expires?
- [x] When invoice is expiring, we need to not just tell people it's expiring, but also make sure to tell them what to do? Pay your invoice before it expires?
- [x] When invoice is partially paid, we'll have to work on some text, and perhaps make it more obvious, the text just blends in into the design. I still think we don't make it obvious that they need to pay the amount exactly show.
- [x] When invoice is partially paid, we'll have to work on some text, and perhaps make it more obvious, the text just blends in into the design. I still think we don't make it obvious that they need to pay the amount exactly show.
- [x] Change text fromInclude Lightning invoice fallback to on-chain BIP21 payment URL to Unify on-chain and lightning tab can potentially link to https://bitcoinqr.dev/ if you think extra context is needed
- [x]  not convinced people won't enter the fiat amount instead of exactly BTC amount. I don't have a clear suggestion about this, but I think we really have to think about this as we work on improving the v2 checkout page.
    - I'd consider this addressed with [these changes](https://github.com/btcpayserver/btcpayserver/pull/4276#issuecomment-1309521854). 
[...]

PR #4276, opened 2022-11-08
Checkout v2 finetuning

dennisreimann:
WIP for #4261.


dennisreimann:
Unified the cases "expiring" and "partially paid" and put them into a single info section:

grafik

dennisreimann:
@kukks brought up this:

> the amount header shows the due amount, but the sub header invoice amount always shows the full inv amount. One would not realize that it was updated and think the new due amount equals the invoice amount?

Before:

grafik

With the now current state I moved the fiat price (here $21.00) into the Details section, put it under Total Price and labeled it Total FIat. That might also circumvent the issue @pavlenex brought up, where people enter the fiat amount in their wallet and end up paying not the excat crypto amount.

Current: 

grafik

Feedback by @dstrukt:

- CON: sometimes people want to see the fiat converted price (counter to pav's point, which is also a very valid one)
- PRO: gain us a little bit more vertical space

raduButucelea23:
Tested the PR successfully. The "Total Fiat" field appears in the "Details" section. 

- Before:
image
- After:
image

raduButucelea23:
Perhaps I missed it in the thread, but was the "expiry time" field completely removed? @dennisreimann

dennisreimann:
It's not entirely gone, but in contrast to before we only show it  once there's only some portion of the time to pay left. 

I think it's a percentage (like for the last 25% of the expiry time), but maybe a fixed amount of time would be better – like 10 minutes left or so. 

We could also get fancy here and incorporate the payment method, confirmations required and so on, but for a first shot the current state will work imho.

dennisreimann:
Addressing the point " Adjust the design so that the QR will be shown above the fold on smaller devices": I decreased the padding here and for the header from 2rem to 1.5rem — as this goes for the content and header, we save a total of 1.5rem above the fold. This should work, I think.

grafik

dstrukt:
Will do proper review tonight .. but for posterity and others viewing the PR, @dennisreimann and I are working through the TODOs on the Issue and via MM, will try to surface the decisions as much as well.

dstrukt:
For those concerned, have been posting visual clean up, tweaks, suggestions, and discussion topics on:

https://github.com/btcpayserver/btcpayserver/issues/4261

dstrukt:
Minor nits, should have probably made these changes myself, but just messing around in the editor to tighten things up:

- [ ] Add `mt-2` to the description header element
- [ ] On description bump the bottom-padding class to `mt-4` from `mt-3`
- [ ] Bump the description font-size to `h5` from `h6` ... maybe

- [ ] Bump the expiry timer container top & bottom inner padding to `-m` from `-s`
- [ ] Bump the expiry timer container margin to `mb-2` from `mb-1`

- [ ] Bump the `Scan QR` copy padding-top to `mt-2` from `mt-1`
- [ ] Bump the `Scan QR` copy padding-bottom to `mt-4` from `mt-3`

- [ ] Bump padding-bottom on the view details list to `pb-4` from `pb-3`


- [ ] Bump left absolute positioning on the hover copy icon to `-1.5rem` from `-1.25rem`
- [ ] Add `200ms` fade-in delay on the icon

dstrukt (approved):

dstrukt:
Remaining TODOs:

- [ ] Update the carot asset (sent directly @dennisreimann via MM), and should we align the carot and the title on the same horizontal line? As opposed to absolute positioning on the carot?

Open questions for call:

- Should we remove the spinning indicator in the expire time?
- Round the QR code borders on the inner element? How much control do we have?
- Displaying addresses via a `Copy` button that would display everything relevant (e.g. BOLT 11, Bitcoin Address, Payjoin, etc..) as opposed to exposing everything by default
- Multiple payment methods...

dennisreimann:
Updated the plugins integration once more with new and separate extension points for Checkout v2. The necessary plugin updates are available in Kukks/btcpayserver#8.

sideshift

fixedfloat

pavlenex (approved):
tACK

dennisreimann:
I think we can merge it in this state and I could do the last cleanups of the Checkout Appearance view in a separate and smaller PR. Right now the Checkout Appearance form is a bit confusing, because it contains the settings alltogether, but some of them don't apply, depending on which version of the checkout is used.

BTCPayServer/wwwroot/js/copy-to-clipboard.js
```diff
@@ -11,25 +15,11 @@ window.copyToClipboard = function (e, data) {
...
-        confirm.dataset.clipboardInitial = confirm.innerHTML;
-        confirm.style.minWidth = confirm.getBoundingClientRect().width + 'px';
-    }
+    const message = confirm.getAttribute('data-clipboard-confirm') || 'Copied';
     if (navigator.clipboard) {
-        navigator.clipboard.writeText(data).then(function () {
+        navigator.clipboard.writeText(data).then(() => {
             confirmCopy(confirm, message);
         });
-    } else {
```
NicolasDorier, line 23: Why you removed this? This is necessary for the copy to work on non-https websites.

dennisreimann:
@pavlenex Adressed 2. and 3. with the recent commits. Will get to 1. now.

Also updated the Checkout Appearance view so that it is clearer, which options apply to which version of the checkout:

## v2
options-v2




## v1

options-v1

NicolasDorier:
@dennisreimann I see that you are modifying some libs. While you are at it, can you update the test `CheckJsContent`? it is basically making sure nobody is tempering the source of the external libs.

dennisreimann:
Done! ✅

PR #4285, opened 2022-11-14
Use the plugin builder website instead of docker to fetch plugins

NicolasDorier:
Until now, we were fetching information of plugins directly from https://github.com/btcpayserver/btcpayserver-plugins
This worked fine as the number of plugins were manageable, but it is now growing. Some users reported the page couldn't even open as github were flooded with too many request (I think @petzsch reported that somewhere)

We are now using our own plugin build server on https://plugin-builder.btcpayserver.org/ (see [source code](https://github.com/btcpayserver/btcpayserver-plugin-builder/)) which exposes two endpoints: One to list plugins and the other to download them.

I removed the `--plugin-remote` which allowed to use a different github repository to fetch the plugins.

Instead, I added two server policies which allow the server administrator to use a different plugin-builder server.

image

I didn't removed `--plugin-remote` completely, it is just ignored, as passing a command line option which doesn't exist anymore will crash btcpayserver on startup. I think we can completely remove it in a few month.

Codewise, note that the `PluginService` is now transient and not a singleton. The reason is that it resolve the `PluginBuilderClient`, whose internal `HttpClient.BaseAddress` is set to the URL set in the server policies. If it was a singleton, the `BaseAddress` wouldn't be updated when the server policies are changing.

Ping @Kukks @dennisreimann

### DIFF
--- a/BTCPayServer/Configuration/BTCPayServerOptions.cs
+++ b/BTCPayServer/Configuration/BTCPayServerOptions.cs
@@ -140,14 +140,15 @@ namespace BTCPayServer.Configuration
             }
 
             DisableRegistration = conf.GetOrDefault<bool>("disable-registration", true);
-            PluginRemote = conf.GetOrDefault("plugin-remote", "btcpayserver/btcpayserver-plugins");
+            var pluginRemote = conf.GetOrDefault<string>("plugin-remote", null);
+            if (pluginRemote != null)
+                Logs.Configuration.LogWarning("plugin-remote is an obsolete configuration setting, please remove it from configuration");
             RecommendedPlugins = conf.GetOrDefault("recommended-plugins", "").ToLowerInvariant().Split('\r', '\n', '\t', ' ').Where(s => !string.IsNullOrEmpty(s)).Distinct().ToArray();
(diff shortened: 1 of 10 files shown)


pavlenex:
Do we want this in 1.7.0 or should this need more testing, hence 1.8.0, my feeling tells me the later, but I'd thought I should  ask?

dennisreimann:
Concept ACK, but I agree on shipping this after v1.7 so that we can first migrate all the plugins there and properly test it.

NicolasDorier:
ok, let's do for after 1.7.0 then

NicolasDorier:
@Kukks uploaded his plugins. I think we can merge this one for 1.7.0 if we get the complete list of plugins.

dennisreimann:
@NicolasDorier done - good job with the plugins-builder! :)

dennisreimann (approved):
tACK.

The only thing I noticed is the order, which seems to be kinda random. Maybe for now, we should sort the plugins in alphabetic order.

NicolasDorier:
I will do this on plugin-builder side.

NicolasDorier:
I archived the old repo: https://github.com/btcpayserver/btcpayserver-plugins

PR #4293, opened 2022-11-15
Code analysis

jesterhodl:
Hi, this is my first PR to BTCPayServer. I decided to learn a bit about the code base by upgrading the old code analysis packages to the ones in .NET SDK, called NETAnalyzers.
This PR does two things:
- Removes obsolete analyzer packages
- Enables NETAnalyzers on the whole project (but keeps them silent)
   This allows the developers to effectively set <AnalysisMode> in the individual .csproj files to minimum, recomended or all to see CA findings. Or not, and keep it quiet
- Fixes:
  - [CA1834: Use StringBuilder.Append(char) for single character strings](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1834)
  - [CA1847: Use string.Contains(char) instead of string.Contains(string) with single characters](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1847)

If this is OK to do, We would have a good base for me and other devs to find good things to fix. Some of them performance related, some security, etc.

### DIFF
--- a/BTCPayServer.Common/Logging/ConsoleLogger.cs
+++ b/BTCPayServer.Common/Logging/ConsoleLogger.cs
@@ -180,7 +180,7 @@ namespace BTCPayServer.Logging
             logBuilder.Append(": ");
             var lenAfter = logBuilder.ToString().Length;
             while (lenAfter++ < 18)
-                logBuilder.Append(" ");
+                logBuilder.Append(' ');
             // scope information
             GetScopeInformation(logBuilder);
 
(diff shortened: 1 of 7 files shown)


jesterhodl:
Looks like changing interface names broke the build of the tests. Any advice on how to proceed with such a change?

NicolasDorier:
@jesterhodl do not change interface names at this point. You can put some `#pragma` to ignore the warning I believe.

jesterhodl:
Ok, I removed the interface name and put 2 other fixes from the performance category.
How's that?

dennisreimann (approved):
tACK

PR #4297, opened 2022-11-17
Refactor walletobj API, make wallet object graph directionless

NicolasDorier:
This PR:
1. Make the wallet object graph undirected. The database still contains the Parents/Child terminology, it should be renamed to A/B in another PR.

This is because I didn't find good answer as to whether a node should be a child or parent of another in any situation, so it would have been easy to shoot yourself in the foot by having the wrong direction.
If specific situation require some kind of interpretation of it, then this specific situation can still store data in the `linkData` about it.

2. Refactor the wallet object API done on https://github.com/btcpayserver/btcpayserver/pull/4274

The API was attempting to do too much at once, and not triggering API's 404 errors. It is unlikely users of this API will want to add thousands of tags at once, as such we shouldn't try to batch multiple add/remove in a single HTTP call. It makes it impossible to identify what entity is responsible if the add/remove fails. And if it would silently fail, it would be worse.

3. Added a route for getting a single wallet object


Collom75:
How do I pull the wallet in the community I am literally stuck and have no clue

PR #4301, opened 2022-11-18
Wallet object scripts

Kukks:
This PR starts introducing the concept of being able to label individual addresses and utxos with the new wallet objects system.

When a new tx is detected, we will look for objects where:
* the spent utxos 
* the tx output scripts/addresses
* the tx newly created utxos

Then, we will link these *existing* objects to the tx object (which will be created if it does not exist).

Currently, we do not transfer labels from these objects to the tx object. That will be the scope of a separate PR, "Wallet object link stickyness"

As an introduction, the wallet receive page will now tag the receiving address with a wallet object of type `receive`.

### DIFF
--- a/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
+++ b/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
@@ -40,6 +40,9 @@ namespace BTCPayServer.HostedServices
         {
             switch (evt)
             {
+                // For each new transaction that we detect, we check if we can find
+                // any utxo or script object matching it.
+                // If we find, then we create a link between them and the tx object.
                 case NewOnChainTransactionEvent transactionEvent:
                 {
                     var txHash = transactionEvent.NewTransactionEvent.TransactionData.TransactionHash.ToString();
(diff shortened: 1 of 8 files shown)


BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -22,42 +23,88 @@ namespace BTCPayServer.HostedServices
...
+                                Id = txOut.TxOut.ScriptPubKey.ToString(), Type = WalletObjectData.Types.Script
+                            },
+                            new OnChainWalletObjectId()
+                            {
+                                Id = txOut.ToCoin().Outpoint.ToString(), Type = WalletObjectData.Types.Utxo
+                            }
+                            
+                            } )).Distinct().ToArray();
+                    
+                    // we are intentionally excluding wallet id filter so that we reduce db trips
```
NicolasDorier, line 68: can't do this, this will provoke full scan

NicolasDorier: mmh, I don't think there is a way around that. I'll probably need to add an index. We can't even get the wallets impacted by the event, as we have no table keeping track of Wallet<->Derivation scheme

Bobbieann0709 (approved):

PR #4309, opened 2022-11-18
Update What's New

### LINKED ISSUE #4305: Remove `what's new` button from the store creation page
> Do you guys think it makes sense to showcase what's new button during the initial store setup? Do you think it potentially distracts user? Is this information relevant? Does a new user care about what's new, when they don't know what's already exist?

Screenshot 2022-11-17 at 8 57 24 AM

It seems that the consensus is to remove this button from the store creation page. D11 suggested we remove it altogether, which is up for a discussion.

dennisreimann:
Add v1.7 info and hide the button in case the store isn't set up, yet.

Added a link to this [discussion](https://github.com/btcpayserver/btcpayserver/discussions/4308) to collect user feedback. 

Closes #4305.

grafik


pavlenex (approved):

dstrukt:
tACK

But weren't we considering removing this feature entirely? ...or did I misunderstand the conclusion?

dennisreimann:
My point was that we don't give it much love and we don't know if anyone is using it at all.  

Then again, it's simple enough to fill it with a sentence or two prior to a release …

pavlenex:
I think for now we're removing it from the store creation page?

dennisreimann:
Exactly. Once you see the Dashboard, the button still occurs where it was.

pavlenex (approved):

dennisreimann:
Let's merge this shortly before the release, otherwise the failing link check test will be annoying us.

dstrukt (approved):
Ahh I see, sounds good .. can always reassess later.

tACK

Collom75 (approved):

PR #4311, opened 2022-11-18
Make store creation field same width

### LINKED ISSUE #4306: Make store creation field same width
We should ensure these fields are nice and tidy, same width.
Good first issue for newcomers to pick up.

Screenshot 2022-11-17 at 8 56 18 AM

dennisreimann:
Closes #4306.

grafik

### DIFF
--- a/BTCPayServer/Views/UIUserStores/CreateStore.cshtml
+++ b/BTCPayServer/Views/UIUserStores/CreateStore.cshtml
@@ -21,12 +21,12 @@
             </div>
             <div class="form-group">
                 <label asp-for="DefaultCurrency" class="form-label" data-required></label>
-                <input asp-for="DefaultCurrency" class="form-control w-auto" currency-selection />
+                <input asp-for="DefaultCurrency" class="form-control w-250px" currency-selection />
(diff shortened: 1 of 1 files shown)


pavlenex:
@dennisreimann Why the name field isn't the same width?

dennisreimann:
I think we shouldn't constrain it to be too small. And I also think the selects shouldn't be much wider.

What's your take @dstrukt?

pavlenex:
Could we achieve the same width without limiting the characters? How realistic it is that a store name is super-long? Wouldn't that break other things, such as store selection drop-down?

dstrukt:
> I think we shouldn't constrain it to be too small.

Generally agree with this.

However, typically I would opt to make all of the inputs the same width, so think I agree with @pav here, and in this case, if we're worried about the `Store Name` length being too short by matching the `selects`, personally I'd just bump the other 2 up to the next column width so they can all be consistent.

I don't think a super long store name is going to be a problem here, an edge case imo, so not really worried about that, and potentially bumping all column widths by 1 would completely alleviate this concern.

dennisreimann:
Ok, updated once more. Good to go now.

PR #4329, opened 2022-11-22
Add documentation link to plugins

NicolasDorier:
Now it is possible to set up document link in the plugin-builder site. Ideally, it should be in the manifest of the plugin, but it would require to modify the plugin interface which would break all plugins. This will be an exercise for another day.

@dennisreimann , I have set the documentation link manually for podserver and lnbank since they are already documented.

image
image


dennisreimann (approved):
tACK

PR #4330, opened 2022-11-22
Fix flaky test

NicolasDorier:
`CanUseLNPayoutProcessor` is flaky, and I noticed it was because the hosted service `LightningAutomatedPayoutProcessor` is blocked waiting for the `Process` to finish.

Tracked it down to a couple of call to LNURL. Passing down the cancellation token should propagate the `LightningAutomatedPayoutProcessor` closure, and thus stop the flakyness.

### DIFF
--- a/BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
+++ b/BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
@@ -75,7 +75,6 @@ public abstract class BaseAutomatedPayoutProcessor<T> : BaseAsyncService where T
                 await context.SaveChangesAsync();
             }
         }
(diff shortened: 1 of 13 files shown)


BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
```diff
@@ -75,6 +75,7 @@ private async Task Act()
                 await context.SaveChangesAsync();
             }
         }
+        Logs.PayServer.LogInformation($"WAITING {CancellationToken.IsCancellationRequested}");
```
dennisreimann, line 78:
```suggestion
```

This looks like a debugging leftover.

dennisreimann (approved):
tACK.

dennisreimann:
Now we just need to figure out how to also fix `CanUseTorClient` :D

PR #4337, opened 2022-11-23
Wallet object contagion

Kukks:
proof of concept for the spread of wallet object links

image

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.cs
@@ -578,17 +578,24 @@ namespace BTCPayServer.Controllers
                 var utxos = await _walletProvider.GetWallet(network)
                     .GetUnspentCoins(schemeSettings.AccountDerivation, false, cancellation);
 
-                var walletTransactionsInfoAsync = await this.WalletRepository.GetWalletTransactionsInfo(walletId, utxos.Select(u => u.OutPoint.Hash.ToString()).Distinct().ToArray());
+                var walletTransactionsInfoAsync = await this.WalletRepository.GetWalletTransactionsInfo(walletId, 
+                    utxos.SelectMany(u => GetWalletObjectsQuery.Get(u)).Distinct().ToArray());
                 vm.InputsAvailable = utxos.Select(coin =>
                 {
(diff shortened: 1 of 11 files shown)


BTCPayServer/Services/WalletRepository.cs
```diff
@@ -303,18 +323,59 @@ public async Task EnsureWalletObjectLink(WalletObjectId a, WalletObjectId b, JOb
...
+            if (contagion is null)
+            {
+                await EnsureWalletObjectLink(a, b, data, infectionRate);
+            }
+            if ( contagion.InfectionRate == 0)
+            {
+                return;
+            }
+
+            foreach (WalletObjectData parentNeighbour in contagion.Parent.GetNeighbours())
```
NicolasDorier, line 369: `contagion.GetNeighbours` already returns both parent and child, so you don't need to duplicate this foreach with the one below.

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -578,17 +578,30 @@ private async Task<bool> CanUseHotWallet()
                 var utxos = await _walletProvider.GetWallet(network)
                     .GetUnspentCoins(schemeSettings.AccountDerivation, false, cancellation);
 
-                var walletTransactionsInfoAsync = await this.WalletRepository.GetWalletTransactionsInfo(walletId, utxos.Select(u => u.OutPoint.Hash.ToString()).Distinct().ToArray());
+                var walletTransactionsInfoAsync = await this.WalletRepository.GetWalletTransactionsInfo(walletId, 
+                    utxos.SelectMany(u => new []
+                    {
+                        new ObjectTypeId(WalletObjectData.Types.Tx, u.OutPoint.Hash.ToString()),
+                        new ObjectTypeId(WalletObjectData.Types.Script,  u.ScriptPubKey.ToString()),
```
NicolasDorier, line 585: id's should be `.ToHex` not `.ToString`.

Jombie69:
How do I get my money?

Clarkv61:
If anyone has the knowledge to help me withdraw the funds I will pay very well when the funds arrive.

BTCPayServer/Services/WalletRepository.cs
ittybittybit: ok thank you for the heads up Nicolas

PR #4340, opened 2022-11-23
Integration: Form Builder + Checkout v2

dennisreimann:
For testing purposes: Here's a branch that integrates the current states of those two pending PRs. #4276 💚 #4137

We can close this one afterwards, only for testing while they aren't merged to master.


dennisreimann:
There are three conflicting files between these PRs, but we can easily resolve them.

pavlenex:
Not sure which of the PR messed things up, I assuem Checkout v2, but when I experimented a bit with min max value for the payment method I got the blank no QR invoice.
Screenshot 2022-11-23 at 20 13 40
Screenshot 2022-11-23 at 20 13 58

dennisreimann:
Obsolete, now that Checkout v2 is merged and Form Builder got updated. Use the Form Builder PR to test from here on.

PR #4342, opened 2022-11-24
Disable receipts for payment request invoices

Kukks:
Payment requests have a receipt-ish style by default. Receipts for each individual invoice of a payment request can be quite confusing as individually they do not prove the pay request was settled.


dennisreimann (approved):

dennisreimann:
@NicolasDorier I tested this and think we should integrate it. Give it a final review tomorrow.

PR #4349, opened 2022-11-25
Ensure redirecturl is local for form builder

Kukks:
After Tempdata was replaced with query strings for formbuilder, the endpoint opened up some attack vectors as anyone could control the links instead of being set internally by the form entrypoint.

Attack vector example:
https://localhost:14142/forms/Email?redirectUrl=http://bitcoinsv.com


dennisreimann:
Can you apply this to #4350? @NicolasDorier is also working on that part of the code currently.

PR #4350, opened 2022-11-25
Some adjustment for Forms

NicolasDorier:
* Properly validate the parameters before setting them.
* Remove some dead code and `requiresRefundEmail` being duplicated in some situations.
* Refactor the Input form providers, so they can implement validation.
* Removing subclasses of `Field`, as we can't properly serialize/deserialize them.
* Ensure redirecturl is local for form builder (#4349)


BTCPayServer/Controllers/UIPaymentRequestController.cs
```diff
@@ -195,32 +201,37 @@ public async Task<IActionResult> ViewPaymentRequestForm(string payReqId, [FromFo
...
                 case { } when string.IsNullOrEmpty(prFormId):
-                    break;
-                
+                case { } when Request.Method == "GET" && prBlob.FormResponse is not null:
+                    return RedirectToAction("ViewPaymentRequest", new { payReqId });
+                case { } when Request.Method == "GET" && prBlob.FormResponse is null:
+                        break;
                 default:
                     // POST case: Handle form submit
-                    if (!string.IsNullOrEmpty(formData) && formId == prFormId)
```
dennisreimann, line 202: I think the `formId == prFormId` part should get applied to the `if` below.

BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
```diff
@@ -230,9 +234,12 @@ public async Task<IActionResult> ViewPointOfSale(string appId, PosViewType? view
                 
                 default:
                     // POST case: Handle form submit
-                    if (!string.IsNullOrEmpty(formData) && formId == posFormId)
```
dennisreimann, line 233: Same for `formId == posFormId` here.

BTCPayServer/Controllers/UIPaymentRequestController.cs
NicolasDorier: `formId` doesn't exists, found out we could remove it as the only form id that mater is the one in the blob.

Kukks:
By making everything on field without the subclass, I don't know how we will be able extend later for more advanced scenarios.

NicolasDorier:
Let's extend it later, we couldn't use it for now.

PR #4354, opened 2022-11-26
Make checkout CSS and logo paths relative

bolatovumar:
See discussion: https://github.com/btcpayserver/btcpayserver/discussions/4195

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoresController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoresController.cs
@@ -12,6 +12,7 @@ using BTCPayServer.Security;
 using BTCPayServer.Services.Stores;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Cors;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using StoreData = BTCPayServer.Data.StoreData;
@@ -149,7 +150,7 @@ namespace BTCPayServer.Controllers.Greenfield
             };
         }
 
-        private static void ToModel(StoreBaseData restModel, Data.StoreData model, PaymentMethodId defaultPaymentMethod)
+        private void ToModel(StoreBaseData restModel, Data.StoreData model, PaymentMethodId defaultPaymentMethod)
         {
             var blob = model.GetStoreBlob();
             model.StoreName = restModel.Name;
@@ -185,6 +186,7 @@ namespace BTCPayServer.Controllers.Greenfield
             blob.LightningDescriptionTemplate = restModel.LightningDescriptionTemplate;
             blob.PaymentTolerance = restModel.PaymentTolerance;
             blob.PayJoinEnabled = restModel.PayJoinEnabled;
+            blob.NormalizeToRelativeLinks(Request);
             model.SetStoreBlob(blob);
         }
 
--- a/BTCPayServer/Controllers/UIStoresController.cs
+++ b/BTCPayServer/Controllers/UIStoresController.cs
@@ -506,19 +506,16 @@ namespace BTCPayServer.Controllers
                 blob.OnChainWithLnInvoiceFallback = model.OnChainWithLnInvoiceFallback;
             }
 
-            var schemeAndHost = $"{Request.Scheme}://{Request.Host.ToString()}";
-            
             blob.RequiresRefundEmail = model.RequiresRefundEmail;
             blob.LazyPaymentMethods = model.LazyPaymentMethods;
             blob.RedirectAutomatically = model.RedirectAutomatically;
             blob.ReceiptOptions = model.ReceiptOptions.ToDTO();
-            // Replace absolute URL with relative to avoid this issue: https://github.com/btcpayserver/btcpayserver/discussions/4195
-            blob.CustomLogo = model.CustomLogo?.IndexOf(schemeAndHost) == 0 ? model.CustomLogo?.Replace(schemeAndHost, String.Empty) : model.CustomLogo;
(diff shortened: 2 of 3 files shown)


BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -1,6 +1,4 @@
 @using BTCPayServer.Payments
-@using Microsoft.AspNetCore.Mvc.TagHelpers
```
bolatovumar, line 2: These are unused.

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -18,6 +16,50 @@
             document.getElementById('CustomCSS').value = ['dark', 'legacy'].includes(theme)
                 ? `/checkout/css/themes/${theme}.css`
                 : ''
+            document.dispatchEvent(new CustomEvent('custom-css-input-update', { detail: document.getElementById('CustomCSS') }));
```
bolatovumar, line 19: Since the value is updated directly in JS the event listener below won't pick up on it so I had to dispatch this custom event to trigger a re-validation.

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -161,3 +203,10 @@
         </form>
     </div>
 </div>
+
+<template id="absoluteURValidation">
```
bolatovumar, line 207: HTML templates are not visible and are only used to actually insert stuff into the document based on the template.

Kixunil:
Awesome! One thing I'm unsure about is if the slash at the beginning should stay or not. Especially when root path is used. If I had to guess I'd say it needs to be stripped too.

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
NicolasDorier: did this as well recently, this is the way to go for warnings.

NicolasDorier:
I think there is a bug with rootPath in 1.7.0 for now.

NicolasDorier:
@bolatovumar I think it's better if we silently rewrite the input of the user.
Reasons are:
* Simpler
* Fix the issue also at the API level as well
* Will work with rootPath (in your PR you don't have the rootpath)
* I don't believe there is a single reason to use the local domain. If there is, we can fix it anyway later easily.

bolatovumar:
@NicolasDorier ok, I can update it. @Kixunil any objections to changing it to what Nicolas is proposing above?

https://github.com/btcpayserver/btcpayserver/pull/4354#issuecomment-1328447666

Kixunil:
Not really. I was a bit worried about users thinking it's some kind of a bug seeing their input modified (and not realizing it does actually do the right thing) but maybe let's try and if there will be (many) reports add some popup explaining that the modification was intentional and will work correctly.

dennisreimann:
2 sats here: I think a better approach in general would be to solve it like with the recently added [logo upload](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UIStoresController.cs#L654) option, with which we store the file id instead of a URL. 

Upload vs. URL not just provides a better UX, but also gives us the ability to handle this in a central place, instead of rewriting the places where we accept URLs.

bolatovumar:
@dennisreimann maybe we should allow both? URL or upload file. If we just replace the URL option with a file upload option we can end up messing up the experience for those users who already have a custom URL set in the older version of BTCPay, right?

Kukks:
Maybe we can add urls as a file under the storage option. That way we have
a file ID consistently but the end result could be a a remote file of an
explicit url

Kalvin1972:
Who are you and why are you writing me

Kixunil:
Wouldn't removal of URL option break users who use *external* resources? Not sure if there are such and not implying it's a good idea to do so.

dennisreimann:
I like @kukks' point, because it'd still allow for using external URLs, but we kinda discourage using them (which is a good thing here imho).

To your point, @Kixunil: We could do this in a backwards-compatible manner or potentially even migrate the files (at least the ones that are hosted on the instance).

Kixunil:
@dennisreimann if by migrating you mean downloading from external location into BTCPay I wouldn't advise it. One of the uses of external resources is improved caching if the file is shared among multiple hosts. Such migration would destroy this use case.

NicolasDorier:
This fix doesn't work
`$"{Request.Scheme}://{Request.Host.ToString()}"` will not replace the non-onion link with a relative link. As `$"{Request.Scheme}://{Request.Host.ToString()}"` will be onion address, and it replace a url that has a clearnet address (so it does nothing)

bolatovumar:
> `$"{Request.Scheme}://{Request.Host.ToString()}"` will not replace the non-onion link with a relative link.

@NicolasDorier It will though. Or do you mean that we should always detect if the URL is an absolute URLs with an onion address no matter what the request host/scheme is and always replace it if it an onion one? (current solution will not do this if you enter a onion link when user is updating the logo/CSS path while on clearnet).

Kixunil:
If BTCPay has both clearnet and onion and you're accessing it over clearnet then onion is presumably not known so the link is not detected and vice-versa. I think this acceptable limitation since I'd expect users to enter the link with same hostname. But if BTCPay could optionally work better when provided the list of all domain names it's accessible at that'd be great.

NicolasDorier:
I think you are right actually should be ok

NicolasDorier:
@Kixunil that's not the case, the reverse proxy is passing the real onion host via HTTP, so it's fine.
I just misunderstood the code.

NicolasDorier:
I adapted the code so it works also for greenfield.

PR #4362, opened 2022-11-27
refactored(UIStoresController.Dashboard.cs): Little changes

pneb:
A little changes has been made


pavlenex:
Can you provide more context on what's been changed and why, it's good idea to give a bit of context on what and why you're making a change, usually helps prioritize actual reviews.

pneb:
> Can you provide more context on what's been changed and why, it's good idea to give a bit of context on what and why you're making a change, usually helps prioritize actual reviews.

- used the ActivatorUtilities.CreateInstance method to create a generic method for creating view models.
- StoreId and cryptoCode were moved to a generic method.

NicolasDorier:
Closing this, this is harder to read and it also doesn't build.
Don't do small refactoring like this, it's difficult to review, doesn't necessarily make the code easier to read, and risk of introducing subtle bugs.

PR #4367, opened 2022-11-28
Fix logos when rootPath is used

NicolasDorier:
When `rootPath` is used, many icon aren't working properly.

I think the problem came after an update of dotnet runtime. So I copied their class in our project, and fixed it to how it was working before.

Note some place I added `asp-append-version`.

```html
<link href="@Model.CustomCSSLink" rel="stylesheet" asp-append-version="true"/>
```

This isn't actually needed, but somehow, if I don't put it, and `CustomCSSLink` use the `~/` path, it isn't properly rewritten.

This PR is needed for https://github.com/btcpayserver/btcpayserver/pull/4354 to work properly.


dennisreimann (approved):
tACK

PR #4376, opened 2022-11-28
PayButton: Fix CSP problems in Firefox

### LINKED ISSUE #4325: [Bug]: The invoice hyperlink is enabled during "Pay Button" plugin configuration

 BTCPay Server v1.6.12.0

docker deploy

### What happened?

504 Gateway Time-out

Under "Plugins", select "Pay Button" to access the configuration window. 
Click "this link" button on the right-hand side of the window, just below the "Preview" of the pay button. 

Expected behaviour: The hyperlink should be disabled during the configuration of the "Pay Button". 

image

### Relevant log output

_No response_

### What browser do you use?

Brave 1.45.118

### Additional information

_No response_

dennisreimann:
Firefox does not support [`unsafe-hashes`](https://caniuse.com/?search=unsafe-hashes), so I figured it might be best to get rid of the inline event handlers in general.

Closes #4325.

### DIFF
--- a/BTCPayServer/Views/Shared/PayButton/PayButton.cshtml
+++ b/BTCPayServer/Views/Shared/PayButton/PayButton.cshtml
@@ -32,7 +32,10 @@
         xhttp.send(new FormData(event.target));
     }
     document.querySelectorAll(".btcpay-form").forEach(function(el) {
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Wouldn't that be a problem if we have two button on the same page?

PR #4377, opened 2022-11-28
Lock user: Improve return code and fix docs

### LINKED ISSUE #4310: [Bug]: Toggle User doesn't deactivate a user

I'm running BTCPay v1.6.12.0

I deployed the server with a script generated by the btcpayserver-Configurator 

### What happened?

I used the api to lock (toggle) a user but it doesn't seem to have any effect. It returns a 200 but the user still has access. It does work when I lock the user via the web interface.

1. I called the endpoint /api/v1/users/{idOfUser}/lock with the corresponding body
{
"locked": true
}
2. returns a 405 (despite having an unrestricted key)

### Relevant log output

_No response_

### What browser do you use?

Postman

### Additional information

_No response_

dennisreimann:
The docs state that the `DELETE` method should be used, though the controller wants `POST`. The latter seems appropriate here, as the action can be used for locking and unlocking.

Also adapted the action to return a status code based on the actual outcome of the user toggle call.

Closes #4310.


Kukks (changes requested):
Can you also update the btcpay client and the local client accordingly with the new response?

Kalvin1972 (approved):

dennisreimann:
@Kukks Updated, please re-review :)

Kukks (approved):

PR #4382, opened 2022-11-29
Adapt ln payouts to handle unknown status

Kukks:


BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -155,25 +160,28 @@ public async Task<IActionResult> GetLNURLForPullPayment(string cryptoCode, strin
...
-                    }
-
+                    var payResult = await UILightningLikePayoutController.TrypayBolt(client,
+                        claimResponse.PayoutData.GetBlob(_btcPayNetworkJsonSerializerSettings),
+                        claimResponse.PayoutData, result, pmi, CancellationToken.None);
+                   
                     switch (payResult.Result)
                     {
                         case PayResult.Ok:
+                        case PayResult.Unknown:
```
dennisreimann, line 170: Why is `Unknown` handled like a success case? The most likely cause for an unknown response is an in-flight payment (for which the request timed out) and the actual result needss to be checked until the payment reaches a determined state.

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -184,7 +192,7 @@ await _pullPaymentHostedService.MarkPaid(new MarkPayoutRequest()
                             return Ok(new LNUrlStatusResponse
```
dennisreimann, line 192: Below we use a `BadRequest` return to also indicate the error with the status code. Something other than `Ok` would be appropriate here too (e.g. `422 Unprocessable Entity`)

BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
```diff
@@ -309,6 +309,11 @@ public async Task<IActionResult> ProcessLightningPayout(string cryptoCode, strin
                         // ignored
                     }
                 }
+                else if(result.Result == PayResult.Unknown)
```
dennisreimann, line 312: I think this part isn't complete, because above it still uses a custom timeout, which interferes with the changes made in btcpayserver/BTCPayServer.Lightning#106. I marked that as a TODO (see the [comment above](https://github.com/btcpayserver/btcpayserver/pull/4382/files#diff-b69cedfa07ecb8474effb9d8be7de27948a33cc4ad46583b6b15e94f4ba58073R284)): The complete way would be to also remove the custom timeout, now that the `Unknown` case gets handled.

BTCPayServer/Controllers/UILNURLController.cs
Kukks: Because this is an lnurl endpoint and can only return a binary result. Since we cannot cancel an inflight payment, we must tell the other wallet to expect it soon.

BTCPayServer/Controllers/UILNURLController.cs
Kukks: It wouldn't be. If we return a failure, then the receiver might still receive the funds after the fact

BTCPayServer/Controllers/UILNURLController.cs
keidav85: I have no idea how to do this or what this is I get emails and text like crazy sure would be a awesome thing if all this is true my family could use it if I can figure it out not going good reason y I haven't been on my stuff in a yr or so lol  what exactly do I need to do to insure this transfer and stop all these messages or what is this

dennisreimann: For the HTTP response I get that — what is the `claimResponse.PayoutData.State` in that case? As long as that serves to distinguish success and unknown, it should be ok then.

Kukks: yes, `claimResponse.PayoutData.State` will be Pending if result was `unknown`, `Completed` if `Ok`

dennisreimann (approved):

dennisreimann:
utACK, code looks good, but `CanUsePullPaymentsViaUI` seems to fail related to the changes.

BTCPayServer/Controllers/UILNURLController.cs
Treelife658: Ok

PR #4386, opened 2022-11-30
Local file system storage as default

dennisreimann:
Checks whether or not a file storage has been set. If not, it sets the local file system storage as default.

### DIFF
--- a/BTCPayServer/Storage/StorageExtensions.cs
+++ b/BTCPayServer/Storage/StorageExtensions.cs
@@ -1,6 +1,5 @@
 using System;
 using System.IO;
-using System.Runtime.InteropServices.ComTypes;
 using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Configuration;
 using BTCPayServer.Storage.Services;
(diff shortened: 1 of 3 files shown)


BTCPayServer/Storage/StorageExtensions.cs
```diff
@@ -25,49 +24,35 @@ public static void AddProviderStorage(this IServiceCollection serviceCollection)
...
-                {
-                    tmpdirInfo = Directory.CreateDirectory(datadirs.Value.TempStorageDir);
-                }
-                else
-                {
-                    tmpdirInfo = new DirectoryInfo(datadirs.Value.TempStorageDir);
-                }
+                var tmpdirInfo = Directory.Exists(datadirs.Value.TempStorageDir)
+                    ? new DirectoryInfo(datadirs.Value.TempStorageDir)
+                    : Directory.CreateDirectory(datadirs.Value.TempStorageDir);
```
dennisreimann, line 46: These are just simplifications of the existing code.

pavlenex:
Concept ack.

NicolasDorier:
The PR is migrating current installs to use file storage provider.
However, it isn't putting the file storage provider for new instances. The migrations don't run for new installs.

dennisreimann:
@NicolasDorier Updated as discussed.

dstrukt (approved):
tACK!

PR #4391, opened 2022-12-02
App domain redirect

dennisreimann:
An experimental stab at #4384: If present, it uses the domain mapping of an app as canonical reference and redirects to it.

Marking this as experimental and inviting discussing, because the behaviour of the `DomainMappingConstraint` needed to change quite a bit: 

- Before it exited early if the `appId` is present (direct access to the POS path, not the domain mapping)
- To make the domain mapping canonical we now need to check if there's a mapping for the app
- If so, we redirect to the mapped domain if it starts with `http`

For the latter, the `DomainToAppMappingItem.Domain` also needed a change:

- Before it strictly validated this property to be a hostname (e.g. `mybtcpay.com`)
- Now it accepts a URI, so that it can also include the protocol (and port, which is nice for local testing, e.g. `https://localhost:14142`)
- Sidenote: imho this improves the UX a bit as I assume most users wouldn't discern between a URL and strict hostname when we ask them to enter a domain. (I was also wondering why the port isn't allowed — in production this wouldn't be used anyways, but it would help with local dev and testing)

The implementation works in a backwards compatible-manner and would require the "domain" part of the mapping to be defined as a URI to redirect and work as intended in #4384.

The fix in a73b893284d7c00bc640be8df05328f13644cff6 is something I found along the way of implementing this, because I saw that the app list in the ViewBag grew.

### DIFF
--- a/BTCPayServer/Filters/DomainMappingConstraintAttribute.cs
+++ b/BTCPayServer/Filters/DomainMappingConstraintAttribute.cs
@@ -35,29 +35,20 @@ namespace BTCPayServer.Filters
                 var matchedDomainMapping = mapping.FirstOrDefault(item => item.AppId == appId);
                 
                 // App is accessed via path, redirect to canonical domain
-                if (matchedDomainMapping != null && matchedDomainMapping.Domain.StartsWith("http"))
+                if (matchedDomainMapping != null)
                 {
-                    context.RouteContext.HttpContext.Response.Redirect(matchedDomainMapping.Domain);
+                    var req = context.RouteContext.HttpContext.Request;
+                    var url = new UriBuilder(req.Scheme, matchedDomainMapping.Domain).ToString();
+                    context.RouteContext.HttpContext.Response.Redirect(url);
                     return true;
                 }
             }
             
             if (hasDomainMapping)
             {
-                var matchedDomainMapping = mapping.FirstOrDefault(item =>
-                {
-                    if (Uri.CheckHostName(item.Domain) != UriHostNameType.Unknown)
-                    {
-                        return item.Domain.Equals(context.RouteContext.HttpContext.Request.Host.Value,
-                            StringComparison.InvariantCultureIgnoreCase);
-                    }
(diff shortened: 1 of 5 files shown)


dstrukt:
ooo, this is an interesting one, curious to see how this shapes up!

NicolasDorier:
I don't really like that domain mapping can be now either url or hostname.
I agree this can improve the UX, but if that's the only reason, you could just silently fix user input during the Save.
The problem with it being url is now we need to handle both case everywhere we use the domain mapping.

I don't understand why you need it to be an URL for this feature to work.

dennisreimann:
> The problem with it being url is now we need to handle both case everywhere we use the domain mapping.

I think the `DomainMappingConstraintAttribute` takes care of that. Other than that it should be just the input side — I'd have to check. 

> If this is because you need the scheme/port to redirect to, just take the same one as the current request.

Ok, I can look into that. The only advantage of using the URL then would be that the port can be included and it'd work in the dev env too — however, this is most likely unneccessary for production deployments.

dennisreimann:
@NicolasDorier I switched it back to use the hostname :)

BTCPayServer/Filters/DomainMappingConstraintAttribute.cs
```diff
@@ -15,45 +12,62 @@ public class DomainMappingConstraintAttribute : Attribute, IActionConstraint
...
+            if (hasAppId && hasDomainMapping)
+            {
+                var appId = (string)context.RouteContext.RouteData.Values["appId"];
+                var matchedDomainMapping = mapping.FirstOrDefault(item => item.AppId == appId);
+                
+                // App is accessed via path, redirect to canonical domain
+                if (matchedDomainMapping != null)
+                {
+                    var req = context.RouteContext.HttpContext.Request;
+                    var url = new UriBuilder(req.Scheme, matchedDomainMapping.Domain).ToString();
```
NicolasDorier, line 41: port?

PR #4392, opened 2022-12-02
Wallet import: Surface detailed error messages

dennisreimann:
Similar to #4205, this checks if the input is an output descriptor and display more detailed information about why an import might fail. Before, it only reported that the wallet file has an "incorrect format".

grafik

@NicolasDorier while we are at it: Is there a specific reason for this three pubkeys limit?


Kukks (approved):

NicolasDorier:
There shouldn't be a limit.

dennisreimann:
@NicolasDorier I fixed the limit issue in #4402.

PR #4400, opened 2022-12-04
Fix store selector transition

dennisreimann:
Fixes btcpayserver/btcpayserver-design#30, a transition bug in WebKit browsers.


dstrukt (approved):
Surprised it was such a simple fix, but LGTM!

PR #4402, opened 2022-12-05
Fix Output Descriptor parsing for WSH multisig case

dennisreimann:
Reuse existing function for extracting from a multisig descriptor, instead of recursively parsing the inner output descriptor. The latter would run into invalid cases, because it'd be interpreted as bare multisig, which supports only up to three public keys. 

For further details see MetacoSA/NBitcoin#1151.


NicolasDorier:
@dennisreimann it would be useful to add a test to `DerivationSchemeParser.ParseOutputDescriptor` about that.

PR #4380, opened 2022-11-29
Upgrade Bootstrap to v5.2.3; Design System improvements

dennisreimann:
This one is a bigger update, especially in terms of variables, because [they added a bunch of CSS custom properties](https://blog.getbootstrap.com/2022/05/16/using-bootstrap-css-vars/).

I could rewrite and simplify chunks of our customization script — nevertheless, practically everything should be the same as before.

Closes btcpayserver/btcpayserver-design#41.

grafik

## Design System Updates

- Closes btcpayserver/btcpayserver-design#23
- Closes btcpayserver/btcpayserver-design#44
- Closes btcpayserver/btcpayserver-design#45
- Closes btcpayserver/btcpayserver-design#53

PR #4409, opened 2022-12-06
Fix mobile nav

dennisreimann:
This seems to have broken with the recent Bootstrap upgrade (#4380) and I didn't notice. WIthout this fix the nav doesn't show up on mobile.


dstrukt (approved):
Oh wow .. yeah I overlooked this as well. 

LGTM on iPhone, and Chrome/Safari simulator as well!

PR #4413, opened 2022-12-07
Generate a wallet object for all scripts, save source in generatedBy rather than receive label

NicolasDorier:
1. `script` objects weren't created in the walletobjects if they were generated from invoices.
2. No `tx` object were created if a transaction were received by an address generated in the wallet (not in invoice)
3. I removed the receive label. We keep this information in the object's data instead. This would have been quite heavy in the UI.


BTCPayServer/Services/Wallets/BTCPayWallet.cs
```diff
@@ -82,6 +88,12 @@ public async Task<KeyPathInformation> ReserveAddressAsync(DerivationStrategyBase
                 await _Client.TrackAsync(derivationStrategy).ConfigureAwait(false);
                 pathInfo = await _Client.GetUnusedAsync(derivationStrategy, DerivationFeature.Deposit, 0, true).ConfigureAwait(false);
             }
+            if (storeId != null)
+            {
+                await WalletRepository.EnsureWalletObject(
+                    new WalletObjectId(new WalletId(storeId, Network.CryptoCode), WalletObjectData.Types.Script, pathInfo.ScriptPubKey.ToHex()),
+                    new JObject() { ["generatedBy"] = "wallet" });
```
Kukks, line 95:
```suggestion
                    new JObject() { ["generatedBy"] = generatedBy });
```

Kukks (changes requested):
I like the new way of ensuring it gets the wallet obj generated. 

But we still need to generate labels else they will not show on the UI. 

Should probably add this new option to greenfield too

BTCPayServer.Tests/GreenfieldAPITests.cs
```diff
@@ -1935,6 +1935,9 @@ await client.MarkInvoiceStatus(user.StoreId, inv.Id, new MarkInvoiceStatusReques
                         RedirectURL = "http://toto.com/lol"
                     }
                 });
+            var invoiceObject = await client.GetOnChainWalletObject(user.StoreId, "BTC", new OnChainWalletObjectId("invoice", newInvoice.Id), false);
```
Kukks, line 1938: the last param is `includeNeighbourData` which is set to false. This test should not pass

NicolasDorier: it should pass, because this request get the link information, which is just what I need.

PR #4416, opened 2022-12-08
Use better default than coingecko when creating a new store

NicolasDorier:
Many people do not pay attention to set the rate source when creating a store.
While coingecko is kind of the best guess if we had no information about the currency, it turns out we know the currency.

This PR attempt to select a better recommendation based on the default currency of the store at creation time.

```csharp
public string GetRecommendedExchange()
{
return DefaultCurrency switch
{
    "EUR" => "kraken",
    "USD" => "kraken",
    "GBP" => "kraken",
    "CHF" => "kraken",
    "GTQ" => "bitpay",
    "COP" => "yadio",
    "JPY" => "bitbank",
    _ => "coingecko"
};
}
```

image

Please give other recommendations on https://docs.google.com/forms/d/e/1FAIpQLSd_r7N9tKzmzKACh7vEqJzSjNIpQzrNK5he_wbUor5mnb1-lg/viewform?usp=sf_link or in response to this PR.

### DIFF
--- a/BTCPayServer/Data/StoreBlob.cs
+++ b/BTCPayServer/Data/StoreBlob.cs
@@ -178,20 +178,19 @@ namespace BTCPayServer.Data
             return rules;
         }
 
-        public string GetRecommendedExchange()
+        public static JObject RecommendedExchanges = new ()
         {
-            return DefaultCurrency switch
-            {
-                "EUR" => "kraken",
-                "USD" => "kraken",
-                "GBP" => "kraken",
-                "CHF" => "kraken",
-                "GTQ" => "bitpay",
-                "COP" => "yadio",
-                "JPY" => "bitbank",
-                _ => "coingecko"
-            };
-        }
+            { "EUR", "kraken" },
+            { "USD", "kraken" },
+            { "GBP", "kraken" },
+            { "CHF", "kraken" },
+            { "GTQ", "yadio" },
+            { "COP", "kraken" },
+            { "JPY", "bitbank" }
+        };
+
+        public string GetRecommendedExchange() =>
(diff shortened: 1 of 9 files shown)


pavlenex:
I'm not sure about the UX here, what does `recommended` mean to the user? Which exchange rate is this?

Wouldn't it be better to say `Kraken (Recommended)` or `Kraken (Default for your currency)` I think more context than `recommended` would be better.

dennisreimann:
@pavlenex agreed, what about something like this?

Peek 2022-12-13 15-21

dstrukt (approved):
tACK.

@dennisreimann Only suggestion for the latest UX adjustment would be to make it

A) `Recommended - Kraken` pushed to the select menu directly, potentially removing that last line of copy.
B) Making the exchange name text white so it's a little faster to parse.

BTCPayServer/Data/StoreBlob.cs
```diff
@@ -178,20 +178,19 @@ public RateRules GetDefaultRateRules(BTCPayNetworkProvider networkProvider)
...
+        public static JObject RecommendedExchanges = new ()
         {
-            return DefaultCurrency switch
-            {
-                "EUR" => "kraken",
-                "USD" => "kraken",
-                "GBP" => "kraken",
-                "CHF" => "kraken",
-                "GTQ" => "bitpay",
-                "COP" => "yadio",
```
NicolasDorier: you changed `COP` and `GTQ` recommendation in this refactoring. Is it on purpose?

dennisreimann: Sorry, I must have messed it up when copying the values – glad you caught it.

PR #4417, opened 2022-12-08
Have address wallet objects rather than script objects

NicolasDorier:
I was talking with @Kukks about it.

Currently we create `script` wallet objects, but the wallet objects is a tool aimed at being exposed to the user.
As such scripts are too low level, instead we should put addresses.

Sadly, NBXplorer doesn't return all the data we need for this right now.
So I updated NBXplorer to get the addresses, and if we don't have those, then we fallback to trying to guess the address from the scriptPubKey. This fallback would give unexpected result for liquid blinded addresses, but that's not too bad.

### DIFF
--- a/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
+++ b/BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
@@ -16,6 +16,7 @@ using BTCPayServer.Services.Apps;
 using BTCPayServer.Services.Labels;
 using BTCPayServer.Services.PaymentRequests;
 using NBitcoin;
+using NBXplorer.DerivationStrategy;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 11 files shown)


BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -98,5 +101,11 @@ protected override async Task ProcessEvent(object evt, CancellationToken cancell
                 }
             }
         }
+
+        private BitcoinAddress GetAddress(NBXplorer.Models.MatchedOutput txOut, string cryptoCode)
+        {
+            // Old version of NBX doesn't give address in the event, so we need to guess
+            return (txOut.Address ?? txOut.ScriptPubKey.GetDestinationAddress(NetworkProvider.GetNetwork<BTCPayNetwork>(cryptoCode).NBitcoinNetwork));
```
Kukks, line 108: can we  use ` nbxnetwork.CreateAddress(derivatonScheme, utxo.KeyPath, utxo.ScriptPubKey);`

Kukks (changes requested):

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
NicolasDorier: Done, it made me think we should also ignore the event if `DerivationStrategy` is null.

PR #4429, opened 2022-12-13
Remove system plugins from the plugin list

NicolasDorier:
I believe we should remove "System plugin section" there is nothing a user can do with it.
It just means that the plugin is in the core code.... so what? it's kind of useless information for users, they can do nothing with it.

Before:
image

After:
image


dennisreimann (approved):
utACK.

PR #4431, opened 2022-12-14
Add links to docs and API in the footer

NicolasDorier:
The API page isn't linked anywhere, so people can't find it.

We have a documentation page on our documentation website, however, since plugins may add their own API, it's better users go to their own page.

I also added link to the documentation website.

image


dennisreimann:
concept ACK.

We should add proper icons (@dstrukt) and I'll take a look how we might handle that it's getting crowded down there :)

pavlenex:
I always imagined a `Developers` section in the sidebar would make more sense, especially because our greenfield and legacy are scattered around. I mean, I don't mind this, but it indeed looks crowded out there and I'm not even sure if it will increase the visibility that much.

dstrukt:
Will get us some icons here this evening so we can close this out!

dstrukt:
Icons sent via DM to @dennisreimann

NicolasDorier:
@pavlenex the bitpay API shouldn't be documented. This is old stuff, we need to let it die.

The docs aren't only useful for developers. Just need a link to the API docs.
@dennisreimann  can we squeeze this one for the 1.7.2 release?

dennisreimann (approved):

PR #4453, opened 2022-12-17
Checkout: Fix modal iframe clipboard permissions

dennisreimann:
WebKit-based browser require a [permissions policy](https://web.dev/async-clipboard/#permissions-policy-integration) to be set on the iframe element. See the discussions [here](https://github.com/btcpayserver/btcpayserver/discussions/4308#discussioncomment-4399342) and [on Mattermost](https://chat.btcpayserver.org/btcpayserver/pl/z7kdgidcjtnd8f5zs5648t1dhe).

### DIFF
--- a/BTCPayServer/wwwroot/js/copy-to-clipboard.js
+++ b/BTCPayServer/wwwroot/js/copy-to-clipboard.js
@@ -18,15 +18,16 @@ window.copyToClipboard = async function (e, data) {
     const message = confirm.getAttribute('data-clipboard-confirm') || 'Copied';
     // Check compatibility and permissions:
     // https://web.dev/async-clipboard/#security-and-permissions
-    let hasPermission = false;
-    try {
-        const permissionStatus = await navigator.permissions.query({ name: 'clipboard-write', allowWithoutGesture: false });
-        hasPermission = permissionStatus.state === 'granted';
-    } catch (err) {}
+    let hasPermission = true;
+    if (navigator.clipboard && navigator.permissions) {
+        try {
+            const permissionStatus = await navigator.permissions.query({ name: 'clipboard-write', allowWithoutGesture: false });
+            hasPermission = permissionStatus.state === 'granted';
+        } catch (err) {}
+    }
     if (navigator.clipboard && hasPermission) {
-        navigator.clipboard.writeText(data).then(function () {
-            confirmCopy(confirm, message);
-        });
+        await navigator.clipboard.writeText(data);
+        confirmCopy(confirm, message);
     } else {
         const copyEl = document.createElement('textarea');
         copyEl.style.position = 'absolute';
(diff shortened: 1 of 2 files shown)


BTCPayServer/wwwroot/js/copy-to-clipboard.js
```diff
@@ -9,14 +9,21 @@ function confirmCopy(el, message) {
     }, 2500);
 }
 
-window.copyToClipboard = function (e, data) {
+window.copyToClipboard = async function (e, data) {
```
NicolasDorier, line 12: are you sure it isn't breaking change?

check that `$clipboardBtn.click(window.copyToClipboard);` in `APIKeys.cshml` still work.

BTCPayServer/wwwroot/js/copy-to-clipboard.js
```diff
@@ -9,14 +9,21 @@ function confirmCopy(el, message) {
...
-    if (navigator.clipboard) {
+    // Check compatibility and permissions:
+    // https://web.dev/async-clipboard/#security-and-permissions
+    let hasPermission = false;
+    try {
+        const permissionStatus = await navigator.permissions.query({ name: 'clipboard-write', allowWithoutGesture: false });
+        hasPermission = permissionStatus.state === 'granted';
+    } catch (err) {}
+    if (navigator.clipboard && hasPermission) {
         navigator.clipboard.writeText(data).then(function () {
```
NicolasDorier, line 27: note that the fact it is an async function may have impact. `navigator.clipboard.writeText(data)` have restriction on some browser as far as I remember.
Like it can't be called if the function calling isn't directly a event handler of click. Which may not be the case on an async function.

BTCPayServer/wwwroot/js/copy-to-clipboard.js
```diff
@@ -9,14 +9,21 @@ function confirmCopy(el, message) {
...
     e.preventDefault();
     const item = e.target.closest('[data-clipboard]') || e.target.closest('[data-clipboard-target]') || e.target;
     const confirm = item.dataset.clipboardConfirmElement
         ? document.getElementById(item.dataset.clipboardConfirmElement) || item
         : item.querySelector('[data-clipboard-confirm]') || item;
     const message = confirm.getAttribute('data-clipboard-confirm') || 'Copied';
-    if (navigator.clipboard) {
+    // Check compatibility and permissions:
+    // https://web.dev/async-clipboard/#security-and-permissions
+    let hasPermission = false;
```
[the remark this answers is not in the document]
NicolasDorier: Also move the check inside the `if()` as it is only relevant to navigator with `navigator.clipboard`

BTCPayServer/wwwroot/js/copy-to-clipboard.js
NicolasDorier: Since you are using `await` here, maybe we can cleanup the `.then`?

dennisreimann:
@NicolasDorier Updated and checked, it works with the async. Kept the permission check out of the if though, so that the else branch can work as a fallback.

PR #4456, opened 2022-12-18
Checkout v2: Reduce Altcoin name on payment method pill

### LINKED ISSUE #4455: [Checkout V2] Reduce Altcoin name on tab
Should be `Monero` ideally, not `Monero XMR`, as each of these shitcoins have their own name and tickers, e.g. `Bitcoin Cash` and don't need both.

I might be mistaken, but AFAIA.

Current (CoinCards instance):
Screen Shot 2022-12-17 at 8 07 07 PM

dennisreimann:
Closes #4455.


dstrukt (approved):
LGTM 🏆

PR #4462, opened 2022-12-21
Add GreenField endpoint for fetching all apps

bolatovumar:
Adds two new endpoints:

- `"api/v1/stores/{storeId}/apps"` -- returns basic app data for all app for a store
- `"api/v1/apps"` -- returns basic app data for all apps for all stores for a user

Capture
Capture2


dstrukt:
Apologies for my potential ignorance here, but will this still be relevant when we move the remaining  `Apps` to `Plugins` .. also made me curious if this is still something we're planning to do / what the timeline for that was.

bolatovumar:
> Apologies for my potential ignorance here, but will this still be relevant when we move the remaining `Apps` to `Plugins`?
> 
> Also made me curious if this is still something we're planning to do / what the timeline for that was.

Good question. I'm not sure actually. If the concept of apps will remain I think it will still be relevant? If it's a plugin that will come pre-installed with BTCPay then I assume people will want to create these apps and be able to fetch the data about them for whatever needs they have.

Issue #4452, opened 2022-12-16
[Bug]: [Checkout v1] Language dropdown cut off


ndeet:
### What is your BTCPay version?

BTCPay v1.7.2.0

### How did you deploy BTCPay Server?

Docker

### What happened?

The language dropdown is cut off due to some z-index or relative/absolute positioning

![image](https://user-images.githubusercontent.com/1136761/208196381-b3e5d3d0-7823-48be-91a2-6b1116f8a0e8.png)

### How did you encounter this bug?

Create an invoice and go to checkout page.

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

cdmoss:
The problem is due to the `absolute` positioning of the dropdown container, the page seems to not bother resizing to accommodate an absolutely positioned element when it changes size. 

Changing the dropdown positioning to `relative` allows the dropdown to expand properly, but pushes all the content below it down since its now in normal document flow. This doesn't look too bad imo since only the "Powered by BTCPayServer" element is pushed down - unless you are in a dev environment, since the fake payment stuff is also pushed down and it looks quite weird
![image](https://user-images.githubusercontent.com/40777533/209068368-e859f410-7e72-43d8-aa1d-60eba86678fd.png). But I'm not sure if this matters much since it only impacts a dev environment. I can open a PR if someone approves of this solution, otherwise I can look for alternatives.

cdmoss:
A better solution may be to make the height of one of the enclosing divs `100vh`. Adding it to either the `modal page` div or the `modal-dialog` div allows the dropdown to expand properly. I've opened a PR for this #4465

cdmoss:
A better solution may be to make the height of one of the enclosing divs `100vh`. Adding it to either the `modal page` div or the `modal-dialog` div allows the dropdown to expand properly. I've opened a PR for this #4465

ndeet:
@cdmoss cc @dennisreimann Thanks for the fix, unfortunately id did not fix the issue with 1.7.3. It seems that the `min-vh-100` class does not exist with the used Bootstrap version and thus has no effect: https://github.com/twbs/bootstrap/blob/v4.0.0/scss/utilities/_sizing.scss

it is there from 4.2.0 onwards though: https://github.com/twbs/bootstrap/blob/v4.2.0/scss/utilities/_sizing.scss, maybe we can upgrade or add the class as a custom class ourself?

dennisreimann:
[...]

PR #4465, opened 2022-12-22
fixes language dropdown cutoff on checkout page

cdmoss:
This is a simple fix that prevents the language dropdown on the Checkout page from being cut off as per #4452. The dropdown was being cutoff because the enclosing modal div was initially sizing to the collapsed dropdown and didn't resize when the dropdown expanded (due to its absolute positioning). This PR assigns a height of 100vh to the enclosing modal div, so that the dropdown has room to expand properly.

### DIFF
--- a/BTCPayServer/Views/UIInvoice/Checkout.cshtml
+++ b/BTCPayServer/Views/UIInvoice/Checkout.cshtml
@@ -86,7 +86,7 @@
 
     <invoice>
         <div class="no-bounce" id="checkoutCtrl" v-cloak>
-            <div class="modal page">
+            <div style="height: 100vh" class="modal page">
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):

BTCPayServer/Views/UIInvoice/Checkout.cshtml
```diff
@@ -86,7 +86,7 @@
 
     <invoice>
         <div class="no-bounce" id="checkoutCtrl" v-cloak>
-            <div class="modal page">
+            <div style="height: 100vh" class="modal page">
```
dennisreimann, line 89:
```suggestion
            <div class="modal page min-vh-100">
```

dennisreimann:
Thanks for the fix!

PR #4471, opened 2022-12-22
Checkout v2: Configure countdown timer

dennisreimann:
This addresses feedback by @astupidmoose left [here](https://github.com/btcpayserver/btcpayserver/discussions/4308#discussioncomment-4438926): Make the countdown timer configurable with a minutes setting. This way the merchant has full control over when to display the timer. They could even set it to equal the invoice expiry, so that it is shown right from the beginning.

grafik

### DIFF
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -22,7 +22,7 @@ namespace BTCPayServer.Client.Models
 
         [JsonConverter(typeof(TimeSpanJsonConverter.Seconds))]
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
-        public TimeSpan TimerExpiration { get; set; } = TimeSpan.FromMinutes(5);
+        public TimeSpan DisplayExpirationTimer { get; set; } = TimeSpan.FromMinutes(5);
 
         [JsonConverter(typeof(TimeSpanJsonConverter.Seconds))]
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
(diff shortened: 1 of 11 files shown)


bolatovumar (approved):
Code looks good.

dstrukt:
Make sense, the only concerns I have are the fact that the invoice `Expiry Time` is in a different location to this `Display Expiry Time after X Time` config. setting .. but this might be a temporary thing until while we iron out all of the details for Checkout V2?

dennisreimann:
I'd argue that the invoice expiry is a more general setting whereas the timer is indeed just a UI-related setting. 

For me it'd work this way, but I'd be open to evaluate this once we are ready to consolidate the settings in general.

dstrukt (approved):
SGTM!

Just wanted to make sure I surfaced that concern.

NicolasDorier:
@dennisreimann I think you should flip around to `Show a timer ... seconds before invoice expiration`.

Currently, if the merchant change the expiration time from 10 to 15 min, he would have to think about changing the timer as well, which he will probably forget. Changing to relative to the expiration time mean it won't need to adjust.

I think the property name in the API should be changed, as name looks like it's invoice expiration while it's not.
Maybe `displayExpirationTimer`. It shows it signals it is a UI concept.

NicolasDorier:
imho, you should keep it in seconds, prevent foot shooting and I think it makes sense to say "I want a timer at the last 90sec"

pavlenex:
> imho, you should keep it in seconds, prevent foot shooting and I think it makes sense to say "I want a timer at the last 90sec"

Can you explain how seconds prevent foot shooting, I don't understand it? To me, as a user, I'm thinking in minutes when it comes to invoice timer, certainly not seconds.

NicolasDorier:
@pavlenex because the API has seconds, but the backend as minutes. Meaning it's a matter of time before somebody step in a mine by thinking he passes seconds instead of minutes.

@dennisreimann I think you should follow the example of `expirationMinutes` . Maybe call the property `displayExpirationTimerMinutes` so it is just coherent with the rest of the properties, so people don't shoot themselves in the foot.

PR #4480, opened 2022-12-26
Update license

yahiheb:


dstrukt:
couldn't this be solved with a dynamic {current_year} output vs. a manual update to the date? maybe there's something i'm not aware of, disregard if so.

cc @dennisreimann

dennisreimann:
I haven't found a canocical resource on this, but from what I've read (e.g. [here](https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year) and [here](https://opensource.stackexchange.com/questions/1522/what-should-be-written-in-mit-license-year-full-name) it should be OK to have only the year of the original publication. 

If necessary, let's just [automate this](https://github.com/marketplace/actions/license-year-updater) tedious task.

dstrukt:
Agree @dennisreimann.

Sounds like we could get away with the singular year after a quick read, but if ya'll didn't interpret it the same and we're still not certain, let's just automate the range.

dennisreimann (approved):

Billiej88 (commented):
Changes

PR #4481, opened 2022-12-26
Ensure only valid non-negative numbers in tip input

### LINKED ISSUE #4478: [Bug]: Negative tipping amount

I'm running BTCPay 1.6.9.0

Manual

### What happened?

It is possible, on btcpay with the POS on cart view, when tips are activated, to put a negative number as tip amount.
i.e I need to pay 2€, I put -1 for tips and the invoice will only be 1 euro.
I think it should be impossible to input a negative amount in this field as the concept of a tip is a bonus and can't be a penalty.

Choose products and add them to the cart.
Click on "confirm" to validate the chart,
Add a negative tip amount (i.e "-1")
The invoice amount is the amount of the selected products minus the tip.

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

Screenshot from 2022-12-24 10-37-22

bolatovumar:
close #4478

### DIFF
--- a/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
+++ b/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
@@ -80,7 +80,15 @@
             <th colspan="5" class="border-0">
                 <div class="input-group mb-2">
                     <span class="input-group-text"><i class="fa fa-money fa-fw"></i></span>
-                    <input class="js-cart-tip form-control form-control-lg" type="number" min="0" step="@Model.Step" value="{tip}" name="tip" placeholder="Tip in @(Model.CurrencyInfo.CurrencySymbol != null ? Model.CurrencyInfo.CurrencySymbol : Model.CurrencyCode)">
+                    <input
+                        class="js-cart-tip form-control form-control-lg"
+                        type="number"
+                        min="0"
+                        step="@Model.Step"
+                        value="{tip}"
+                        name="tip"
+                        placeholder="Tip in @(Model.CurrencyInfo.CurrencySymbol != null ? Model.CurrencyInfo.CurrencySymbol : Model.CurrencyCode)"
+                    />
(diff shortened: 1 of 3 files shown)


BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
```diff
@@ -80,7 +80,15 @@
             <th colspan="5" class="border-0">
                 <div class="input-group mb-2">
                     <span class="input-group-text"><i class="fa fa-money fa-fw"></i></span>
-                    <input class="js-cart-tip form-control form-control-lg" type="number" min="0" step="@Model.Step" value="{tip}" name="tip" placeholder="Tip in @(Model.CurrencyInfo.CurrencySymbol != null ? Model.CurrencyInfo.CurrencySymbol : Model.CurrencyCode)">
+                    <input
```
bolatovumar, line 83: Just formatting change here to make it more readable.

BTCPayServer/wwwroot/cart/js/cart.js
```diff
@@ -340,7 +340,6 @@ Cart.prototype.updatePosData = function() {
       tip: this.tip? this.tip: 0,
       total: this.getTotal(true)
     };
-    console.warn(result);
```
bolatovumar, line 343: Removed this as it was logging to the console for no reason.

dennisreimann (approved):
Is the custom JS for validation necessary? I think having the number input would suffice.

NicolasDorier:
Can you also clip this number in the backend in the POST?

bolatovumar:
> Is the custom JS for validation necessary? I think having the number input would suffice.

@dennisreimann number input actually allows negative numbers if you type them out explicitly. It also allows weird characters such as `e` since these are technically numbers (I wonder if Pi is allowed lol)

 > Can you also clip this number in the backend in the POST?

Negative numbers already don't work in the POST. If you try to create an invoice with a negative value nothing will happen. Invoice doesn't even show up.

dennisreimann (approved):
tACK.

PR #4483, opened 2022-12-27
add extension points for dashboard

Kukks:


dennisreimann:
Not sure if having start and end points is the right approach. I think we should plan in the direction of letting the user specify the ordering of the tiles.

dstrukt:
can't speak to start/end points, but +1 to ordering and visible/hidden (fwiw) .. imagine that would cover 90% of needs

Kukks:
Either way we still need at least one extension point on here to add new
ones. And since the widgets have a predefined size and start/end index, it
might be a bit tough to make it modular *just right*

I propose this easy solution until we can tackle this in a grand manner
(resizable, moveable, toggle able, configurable widgets)

dennisreimann:
@Kukks let's make it one extension point as `<vc:ui-extension-point location="dashboard" model="@Model"/>` — your call if at start or end.

BTCPayServer/Views/UIStores/Dashboard.cshtml
```diff
@@ -119,6 +120,7 @@
             <vc:app-sales vm="@(new AppSalesViewModel { App = app, InitialRendering = true })"/>
             <vc:app-top-items vm="@(new AppTopItemsViewModel { App = app, InitialRendering = true })"/>
         }
+        <vc:ui-extension-point location="dashboard-end" model="@Model"/>
```
Kukks, line 123:
```suggestion
```

BTCPayServer/Views/UIStores/Dashboard.cshtml
```diff
@@ -76,6 +76,7 @@
         };
     </script>
     <div id="Dashboard" class="mt-4">
+        <vc:ui-extension-point location="dashboard-start" model="@Model"/>
```
Kukks, line 79:
```suggestion
        <vc:ui-extension-point location="dashboard" model="@Model"/>
```

Issue #4114, opened 2022-09-12
Full Improved Point of Sale Checkout Flow


dstrukt:
This issue will track and highlight an end-to-end flow of the improved POS (specifically, but could be any plug-in, etc..) checkout experience, so we evaluate it's strengths and weaknesses. I will create sub issues to break out the relevant tickets to close as we merge, however this ticket should not be closed until all of the following is done.

Note: There are areas of this that are highly WIP (store settings, form builder, cart) that will be determined by the implementation and constraints, but I wanted to start laying out the pieces so we can have more concrete discussions of trade-offs, expectations, etc..

- [x] Whitelabel support + store icon (https://github.com/btcpayserver/btcpayserver/pull/4131)
- [x] Extend whitelabel support (https://github.com/btcpayserver/btcpayserver/issues/3842)
- [x] Improved file system support
- [x] Dynamic form builder (+ eventual plug-in support)
- [x] Updated checkout invoice
- [x] Updated receipt view
- [ ] Updated Point of Sale + Cart view
- [ ] POS: Make it clearer, how the custim amount works ([details](https://github.com/btcpayserver/btcpayserver/issues/4275#issuecomment-1308843465))

## Full animated flow (WIP - Will be updating)
![CheckoutFlowV0 2](https://user-images.githubusercontent.com/6250771/189557820-23890bcf-49aa-48e4-930d-80935a344bae.gif)

### Store Settings (WIP):
Whitelabeling
![Store Settings - General](https://user-images.githubusercontent.com/6250771/189557256-db2932ea-afa8-4645-ae0e-7e8d9b791772.png)

### Crowdfund Admin (WIP):
Addition of the custom metadata / form builder
The form builder would be encapsulated and modular so it can be leveraged in all plugins (Crowdfund, etc..), as each application might require different metadata.
![Apps - Point of Sale](https://user-images.githubusercontent.com/6250771/189557266-7b57559e-deab-42b1-8764-bb668fa60dc0.png)

### Point of Sale
![POS + Cart](https://user-images.githubusercontent.com/6250771/189557351-9a028231-8b94-4a49-9f3a-6eab2a059972.png)

### Cart + Review
![Cart](https://user-images.githubusercontent.com/6250771/189557588-ab565e67-acea-448c-9937-702ea626cd41.png)

### Metadata (WIP):
[...]

Issue #3842, opened 2022-06-09
Add support for logo white-labeling on Payment Requests, Invoices, Receipts, Pull Payments, etc...


dstrukt:
- [ ] Payment Request
- [x] Checkout Invoices
- [x] Receipts (https://github.com/btcpayserver/btcpayserver/pull/4435)
- [ ] Pull Payments

As an admin... I'd like to be able to promote my brand with more than just my website, but with the POS front-end, Payment Requests, etc...

Ideally this functionality would enable an admin to upload their brand/store's logo in the `Store Settings > General` that we would then add to Invoices, Payment Requests, Pull Payments, and (now)Receipts .. and wherever potentially applicable, possibly as a default or a switch on/off functionality .. working this into mocks as well.

This Issue ties into the larger initiative of making files easier to manage as well, but after configuring your File system (would prompt if you haven't), you could simply upload your logo and we'll handle the rest.

We would still keep "Powered by BTCPay Server" at the bottom on all of these views.

Curious to hear other's thoughts.

tmcphail67:
I think this is a great Idea.  It would also be great to be able to white label the the login screen and store screens with the hosts logo and name.  Providing the service to brick and mortar merchants and web stores, merchants would see the hosts brand with the powered by BTCPay Server in the footer.

dstrukt:
Exactly!

We discussed some other basic optional functionality as well in our design call .. e.g. things like a primary theme color,  "business name or LLC", "address", etc.. for those who may also need that use-case, but one step at a time.

dennisreimann:
We can tackle this once #4459 is merged.

PR #4489, opened 2022-12-30
Checkout v2: Display and copy addresses

### LINKED ISSUE #4442: [Checkout V2] Add Copy Address support
- [ ] Show address(es)
A) Via Copy button
Screen Shot 2022-12-21 at 12 00 32 AM
B) Dumping the payment URL under the QR
Screen Shot 2022-12-21 at 1 08 20 AM

- [ ] Shows addresses / relevant payment details as a modal on `Copy` button click for >1

dennisreimann:
Opted for the simplest variant and displaying the info right away — imho this works best for the case in which one wouldn't want to copy, but just see and confirm the addresses. This also makes it one click less for copying.

I left out the payment method icon and copy button to have more space for the info (mostly relevant for the Bitcoin address) and because it felt kind of redundant for the >1 cases.  

Also to be discussed: Where do we want to have the CTA button? I left it at the botton, because I felt the text underneath the QR explains the functionality of the then following addresses better, but I can also see why one would like to have it right undernath the QR.

Closes #4442.

## Separate payment methods

copy-address

## Unified

unified


dennisreimann:
Maybe we can even omit the full payment link as this isn't really meant for copying and the "Pay in Wallet" button triggers the action.

dennisreimann:
Updated it. Much better.

## Bitcoin and Lightning

copy-address


## Unified

unified

dstrukt:
The latter is much nicer!

I do like the idea

dennisreimann:
Ok, what about this? 

* Removed the explanation text underneath the QR, as it isn't necessary anymore if we have those explicit buttons
* Added the payment method icon to the QR (optional, but I remember someone mentioned it'd be good to have it back)

grafik

dstrukt:
Liking the direction more!

### Payment Method Icon

Agree with the payment method icon being an optional toggle in the admin settings .. makes me wonder if we want to any kind of border, some quick explorations to help decide if it's worth doing, e.g:

Screen Shot 2022-12-31 at 5 48 07 PM

Otherwise, I might also scale up the (bitcoin) icon by like 5% if it's 40px to 48px or 56px to 64px if we can afford that space, something like that.

### Copy Address 

Not sure I love the `Copy` button, so did some text only and icon explorations for alternatives.

Exploration:
Screen Shot 2022-12-31 at 5 42 51 PM

dennisreimann:
> Agree with the payment method icon being an optional toggle in the admin settings

I didn't mean it to be a setting and I wouldn't do it like that. I meant I inserted it as an optional thing we could go with — so I'd say either we have it or we don't, but not make it configurable. I also wouldn't bump up the size too much, as it takes space from the QR data and it might be tricky with the unified one especially. So maybe we should just omit it for now and reevaluate if users really want it.

> Not sure I love the Copy button, so did some text only and icon explorations for alternatives.

Ok, will update it and also adjust colors and font weights (semibold?) for the labels and addresses.

Let me know which variant to go with — icons or text? We should then maybe also unify it across the app, as we aren't consistent in that regard right now.

dennisreimann:
Updated and also checked that the unified QR still works with this icon size. I really like it this way and in conjunction with #4459 it also leaves the option to remove the icon via custom CSS.

grafik

dstrukt:
> I didn't mean it to be a setting and I wouldn't do it like that. I meant I inserted it as an optional thing we could go with — so I'd say either we have it or we don't, but not make it configurable. I also wouldn't bump up the size too much, as it takes space from the QR data and it might be tricky with the unified one especially.

Ahhh I see, let's roll with it enabled, and collect feedback, it's more clear imo. I also like the circular border - looks great! Sounds good w.r.t the size, esp. considering we added the border.

> Ok, will update it and also adjust colors and font weights (semibold?) for the labels and addresses.

I think the label here looks great considering we're not using the input borders, and realize this is all a little inconsistent for for the rest of the app, but that might be acceptable...

Screen Shot 2023-01-01 at 10 02 10 PM

> Let me know which variant to go with — icons or text? We should then maybe also unify it across the app, as we aren't consistent in that regard right now.

Hmmm I'm still a little uncertain which variant I prefer, but agree we should make it as consistent as possible.

Will co-ordinate with you on Mattermost so we can get this sorted and merged.

Otherwise, everything looks great!

Craw69699:
What do I do

dennisreimann:
If the only thing left is the copy icon vs. text matter, let's go with the text as that's what we use in most places — there are a few outliers that use the icon, but reviewing and unifying those should be done separately anyways.

We should then also consolidate those "labels" as we use them in the other public pages as will, e.g. for amounts in payment requests and pull payments iirc.

dstrukt:
> If the only thing left is the copy icon vs. text matter, let's go with the text as that's what we use in most places — there are a few outliers that use the icon, but reviewing and unifying those should be done separately anyways.

Sounds good, let's just go with the text for now, can collect some feedback, or switch it in the future if it makes sense.

> We should then also consolidate those "labels" as we use them in the other public pages as will, e.g. for amounts in payment requests and pull payments iirc.

Yeah I'm debating if we should just use the standard "label" that we are currently and consolidate at a later point, as our current label is sufficient. I do like the uppercase label, and think in this instance it looks better than it might in other places, but i'd need to poke around in Figma before making that call.

Do you have a preference?

dennisreimann:
> Do you have a preference?

I'm fine either way — we'll keep it consistent with the rest once we get to the other public pages in #4114 and #3842.

dstrukt (approved):
Sounds good - let's roll with what we have then, as it's perfectly sufficient, and we can do some minor consistency modifications in the near future!

PR #4511, opened 2023-01-09
Allow DI access to host

Kukks:
Currently, there is no way to view correctly what btcpay is bound to internally.  There is access to `IServer` which has a similar structure and data, but it only provides the bound addresses of the internal Kestrel server, while the `IWebHost` takes precedence.  This is needed to be able to reliably generate loopback calls in case we need to call local endpoints from plugins (beyond greenfield calls)


NicolasDorier:
What property do you want from `IWebHost`? This seems hacky, I would be surprised there isn't a way to get those addresses.

Kukks:
I hacked my way around this weird situation. lets not hack it for now

PR #4513, opened 2023-01-09
Changelog v1.7.4

dennisreimann:
Includes commits until 798553e96afeb3aa7ddac125a830fde7a47be187.

This one should also get added:

- #4520
- #4528
- #4529

Depending on how we feel about shipping those Checkout v2 updates as new features, we could also include the following. They seem often requested and ready: 

- #4489 
- #4471

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## 1.7.4
+
+### Bug fixes
+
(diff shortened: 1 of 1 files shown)


NicolasDorier:
I would like to push https://github.com/btcpayserver/btcpayserver/pull/4528 as well if possible.

PR #4518, opened 2023-01-10
Can load external plugins during dev to debug more easily

NicolasDorier:
I refactored the plugin manager so that we can debug a plugin that is in a specific folder.

Example of `BTCPayServer/appsettings.dev.json`.

```json
{
  "DEBUG_PLUGINS": "C:\\Sources\\btcpayserver-plugin-lnbank\\Plugin\\bin\\Debug\\net6.0\\BTCPayServer.Plugins.LNbank.dll"
}
```
Such plugin will be automatically loaded.
This is a debug only feature, as normally the plugins should be in the plugin directory.
If you then reference the plugin in the `.sln`, you can debug step by step the plugin.

I also refactored the code that was quite hard to review.

Using `DEBUG_PLUGINS` rather than project references has the side effect to break circular dependencies of plugins, allowing plugins to reference BTCPayServer like on https://github.com/dennisreimann/btcpayserver-plugin-lnbank/pull/24

I removed a bunch of shared `static` instances that were a code smell.

There is also a breaking change that I believe we don't depend on:

* I restrict each plugin assembly to have a single plugin whose plugin identifier match the assembly name.

Among the many problems if that happen is that a plugin could crash and BTCPay wouldn't be able to disable it if the assembly was a different name than the plugin.


dennisreimann:
This is working for me, though it changes the workflow a bit: The main difference is, that now whenever the plugin changes, it needs to be build separately — this wasn't the case before as BTCPayServer had the plugin as dependency and did rebuild it whenever itself got build. Also BTCPayServer needs to restart whenever the plugin is build anew. I can live with that — debugging works as before, once the `BTCPayServer/appsettings.dev.json` is set up.

In terms of repositories, the submodule references got reversed. Before, I had the plugins as submodules inside the BTCPay Server repo (on a separate `plugins` branch) — now the solution on the `plugins` branch references the plugin projects, which have BTCPay Server as a submodule. This might interest @woutersamaey too.

We might come across more differences, but for now those changes seem minor to me and I like having the possibility to reference classes from BTCPayServer directly, without the need for Abstractions.

woutersamaey:
Thanks for tagging me @dennisreimann . Very interesting topic indeed.

dennisreimann (approved):
concept ACK.

Kukks:
> This is working for me, though it changes the workflow a bit: The main difference is, that now whenever the plugin changes, it needs to be build separately — this wasn't the case before as BTCPayServer had the plugin as dependency and did rebuild it whenever itself got build. Also BTCPayServer needs to restart whenever the plugin is build anew. I can live with that — debugging works as before, once the `BTCPayServer/appsettings.dev.json` is set up.
> 
> In terms of repositories, the submodule references got reversed. Before, I had the plugins as submodules inside the BTCPay Server repo (on a separate `plugins` branch) — now the solution on the `plugins` branch references the plugin projects, which have BTCPay Server as a submodule. This might interest @woutersamaey too.
> 
> We might come across more differences, but for now those changes seem minor to me and I like having the possibility to reference classes from BTCPayServer directly, without the need for Abstractions.

Maybe we can do this to reduce the build step?
https://stackoverflow.com/a/51907613/275504

NicolasDorier:
> This is working for me, though it changes the workflow a bit: The main difference is, that now whenever the plugin changes, it needs to be build separately

Are you sure about this? In VS, when you hit F5, all projects in the solution get built, the plugin included.

PR #4520, opened 2023-01-10
Greenfield: Add payment hash and preimage to Lightning invoices

### LINKED ISSUE #4475: [Feature] Greenfield: Add the preimage in GET {invoiceId}/payment-methods
A user would need the pre image for a successful payment in [the invoice's payment methods route](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Invoices_GetInvoicePaymentMethods)

I believe it can be add in `LightningListener.AddPayment` however the passed `LightningInvoice` object doesn't contain the pre-image. So it probably require a change in the lightning library as well. Or maybe there is a call in the lightning client to fetch the payment image?

See: https://twitter.com/JohnOnChain/status/1606299203415597058?s=20&t=yEYK8OAUqNpnGkNLX2kXiA

dennisreimann:
Closes #4475.

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningLikePaymentMethodDetails.cs
+++ b/BTCPayServer/Payments/Lightning/LightningLikePaymentMethodDetails.cs
@@ -46,20 +46,13 @@ namespace BTCPayServer.Payments.Lightning
             return null;
         }
 
-        public virtual JObject GetAdditionalData(IEnumerable<PaymentEntity> payments)
+        public virtual JObject GetAdditionalData()
         {
             var result = new JObject();
-            
-            // use set properties and fall back to values from payment data
-            var payment = payments.Select(p => p.GetCryptoPaymentData() as LightningLikePaymentData).FirstOrDefault();
-            var paymentHash = PaymentHash != null && PaymentHash != default ? PaymentHash : payment?.PaymentHash;
-            var preimage = Preimage != null && Preimage != default ? Preimage : payment?.Preimage;
-                
-            if (paymentHash != null && paymentHash != default)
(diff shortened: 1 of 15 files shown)


dennisreimann:
The remaining part here is updating the Lighting lib once the PR mentioned above is merged and adding the data to the payment methods endpoint, as originally requested in #4475.

BTCPayServer/Payments/Lightning/LightningLikePaymentMethodDetails.cs
```diff
@@ -43,9 +46,21 @@ public virtual string GetAdditionalDataPartialName()
             return null;
         }
 
-        public virtual JObject GetAdditionalData()
+        public virtual JObject GetAdditionalData(IEnumerable<PaymentEntity> payments)
```
NicolasDorier, line 49: I'd like to avoid changing `GetAddtionalData` signature, as it is unclear what `PaymentData` correspond to this details, this is also a circular dependency.

Why are you doing this? it seems `PaymentHash` and `Preimage` are already there?

dennisreimann: Most of the time the properties aren't set. I have to check and we can maybe change it to set them properly – this seemed like the most reliable way, but I agree about your concerns and will see if we can solve it using the properties exclusively.

dennisreimann: Done in [70626de](https://github.com/btcpayserver/btcpayserver/pull/4520/commits/70626def420542acb2f313bea36775dd3a22d4bc).

PR #4521, opened 2023-01-11
Fix: Version numbers in UI appeared with a suffix zero x.x.x.0 instead of x.x.x

NicolasDorier:
Saving trees

Before:
image

After:
image


dennisreimann (approved):

dennisreimann:
Should we apply this for the plugins as well?

NicolasDorier:
This is just the versioning we use for btcpay, it's only 3 digits.

For plugin it would be whatever is in the csproj by default if we merge that.
https://github.com/btcpayserver/btcpayserver/pull/4522

PR #4528, opened 2023-01-16
Redesign plugin list items

NicolasDorier:
Since anybody can push plugin, it is important that we show the author of plugins as well as link to source and documentation.

I also add a red line "no documentation", for undocumented plugin in an effort to incentive documentation.

Current:
image

After:
image

image


pavlenex:
Concept ACK, good additions, shall we add this in a major release 1.8.0?

Kukks (approved):

dennisreimann:
Concept ACK, but we should unify the icons with what we have already. Will get to it and ask @dstrukt if we need something else.

dennisreimann (approved):
tACK. Reformatted the view code, because it was all over the place with mixed tabs and spaces.

PR #4530, opened 2023-01-17
Minor UI updates

dennisreimann:
Came across these while working on the plugin list on the signet demo instance.


aim2pleas55:
Could someone help me to receive these funds I would greatly appreciate it and compensate who ever could help me

PR #4541, opened 2023-01-20
Make NFC built in

Kukks:
This PR makes the popular NFC plugin built in.

* Supports Checkout v2
* Uninstall existing plugin if detected
* Fixed missing lnurl option in checkout v2 if using unified view


BTCPayServer.Tests/FastTests.cs
```diff
@@ -1698,7 +1698,7 @@ public void CanCalculateCryptoDue2()
             var networkProvider = new BTCPayNetworkProvider(ChainName.Regtest);
             var paymentMethodHandlerDictionary = new PaymentMethodHandlerDictionary(new IPaymentMethodHandler[]
             {
-                new BitcoinLikePaymentHandler(null, networkProvider, null, null, null),
+                new BitcoinLikePaymentHandler(null, networkProvider, null, null, null, null),
```
dennisreimann, line 1701:
```suggestion
                new BitcoinLikePaymentHandler(null, networkProvider, null, null, null, null, null),
```

Koubek11 (approved):

dennisreimann:
Rebased and fixed the test. 

Thanks to @Kukks this now also includes the LNURL in the `lightning` fallback option in case of the unified QR code. This isn't widely supported yet, but it seems like Zeus, Alby and Wallet of Satoshi (at least partially) support [LNURL in BIP21](https://github.com/lnurl/luds/issues/177#issuecomment-1401477399) via the `lightning=` query string parameter. 

## Unified QR enabled - standard invoice

- Payment URI: `bitcoin:bcrt1qnr4urs7rs9jalswyk4vn7hrhkh9m4r69nyfkj6?amount=0.0009188&lightning=lnurl1dp68gup69uhkcmmrv9kxsmmnwsarzdp3xsez7sj5gvh42j2vfe24ynp0wpshjtmf9ayyycm0wpny6v6cgfpkwatpw568xs2s24v5kmddx3y`
- QR code: `bitcoin:BCRT1QNR4URS7RS9JALSWYK4VN7HRHKH9M4R69NYFKJ6?amount=0.0009188&lightning=LNURL1DP68GUP69UHKCMMRV9KXSMMNWSARZDP3XSEZ7SJ5GVH42J2VFE24YNP0WPSHJTMF9AYYYCM0WPNY6V6CGFPKWATPW568XS2S24V5KMDDX3Y`

bip21

## Unified QR disabled - standard invoice

Intentionally does not include the LNURL, because it isn't a top-up invoice.

grafik

## Top-up invoice case, Unified QR enabled/disabled

Explicit switch for BTC and LNURL, no NFC functionality.

grafik

pavlenex:
Why is a button needed? What's the technical blocker in just having NFC  it just always on? We could then add a simple icon somewhere indicating it's NFC compatible, just like CoinCorner does it? IMO a button is just an extra click.

prusnak:
> What's the technical blocker in just having NFC it just always on?

I think it's the same situation as with the WebUSB. A website needs to ask for a NFC permissions first and it cannot ask for it automatically, but rather while handling a user gesture (e.g. a button click).

It is called a "user-initiated permission prompt" - see https://web.dev/nfc/#security-and-permissions for more information.

pavlenex:
Thanks @prusnak for explaining, I wonder though how do guys at CoinCorner avoid the user-initiated prompt? Does anyone know?

dstrukt:
Conceptually agree @pavlenex, was also wondering how they avoided that in their early production demo videos, but wondered if that detail might have been staged?

Would also love to know.

And appreciate the clarification as well @prusnak!

Kukks:
> Thanks @prusnak for explaining, I wonder though how do guys at CoinCorner avoid the user-initiated prompt? Does anyone know?

They have a native app which does not have this interaction requirement.

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -0,0 +1,98 @@
...
+                ? LightningLikePayoutHandler.LightningLikePayoutHandlerOnionNamedClient
+                : LightningLikePayoutHandler.LightningLikePayoutHandlerClearnetNamedClient);
+            var info = (await
+                LNURL.LNURL.FetchInformation(uri, "withdrawRequest", httpClient)) as LNURLWithdrawRequest;
+            if (info is null)
+            {
+                return BadRequest("Could not fetch info from lnurl-withdraw ");
+            }
+
+            httpClient = _httpClientFactory.CreateClient(info.Callback.IsOnion()
```
NicolasDorier, line 63: does `LNUrl` validate the fact that `info.Callback` isn't null?

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -0,0 +1,98 @@
...
+                ? LightningLikePayoutHandler.LightningLikePayoutHandlerOnionNamedClient
+                : LightningLikePayoutHandler.LightningLikePayoutHandlerClearnetNamedClient);
+
+            try
+            {
+                var destinationuri = LNURL.LNURL.Parse(request.Destination, out string _);
```
NicolasDorier, line 69: nothing check if `request.Destination` isn't null.

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -0,0 +1,98 @@
...
+                if (destinfo is null)
+                {
+                    return BadRequest("Could not fetch bolt11 invoice to pay to.");
+                }
+
+                httpClient = _httpClientFactory.CreateClient(destinfo.Callback.IsOnion()
+                    ? LightningLikePayoutHandler.LightningLikePayoutHandlerOnionNamedClient
+                    : LightningLikePayoutHandler.LightningLikePayoutHandlerClearnetNamedClient);
+                var destCallback = await destinfo.SendRequest(destinfo.MinSendable, Network.Main, httpClient);
+                request.Destination = destCallback.Pr;
```
NicolasDorier, line 83: `destCallback.Pr` can't but null?

pavlenex:
> > Thanks @prusnak for explaining, I wonder though how do guys at CoinCorner avoid the user-initiated prompt? Does anyone know?
> 
> They have a native app which does not have this interaction requirement.

@bolatovumar I guess built in NFC is something to consider for the POC of our mobile app then, seems like a must feature for retailers in the future.

dstrukt:
Wondering .. should we make the  `LNURL` buttons our secondary button style? As it's not the primary method/CTA, and is for specific use cases and users..

Thoughts?

dennisreimann:
We only show LNURL in the top-up case, not as an alternative to regular LN. Hence it should be presented the same way as the rest  imho.

dstrukt:
Hmm I only meant in the `Unified QR disabled - standard invoice` and `Unified QR enabled - standard invoice` cases. I wonder if the `Pay by LNURL-Withdraw`, but you're making me reconsider...
image

dennisreimann (approved):
tACK

PR #4545, opened 2023-01-21
Fix several HTML injections

NicolasDorier:
Several HTML injections need to be patched.
I need additional pairs of eyes that I'm not missing anything.

@dennisreimann most of those injections come from `data-description` for modals where are raw HTML. As such anything coming from user's input need to be escaped with `Html.Encode`.

I don't believe this is a critical issue that require urgent release, as it isn't possible to make JS injection thanks to CSP which save our ass once again!

Credit to @thewhiteevil from https://huntr.dev/bounties/3a73b45c-6f3e-4536-a327-cdfdbc59896f/

Reproduce steps:

* Go to the website https://mainnet.demo.btcpayserver.org/account/apikeys
* Click account.
* Click manage account.
* Then move on to the api endpoint.
* That is https://mainnet.demo.btcpayserver.org/account/apikeys
* Here, create api key with the label as a html payload.
* Use the following HTML payload in the label field: `<a href="https://hackerbro.in">clickhere</a>`
* Generate the api key.
* Now, click delete.
* Check that deleting part rendered the html injection.
* That's the issue.

### DIFF
--- a/BTCPayServer/Controllers/UIAppsController.cs
+++ b/BTCPayServer/Controllers/UIAppsController.cs
@@ -13,6 +13,7 @@ using BTCPayServer.Services.Stores;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Rendering;
 
 namespace BTCPayServer.Controllers
 {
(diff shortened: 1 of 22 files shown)


dennisreimann (approved):
I checked the instances in which we are setting `innerHTML` or using jQuery's `html()`. Haven't found more occurrences. Will make sure to think of this in the future!

Koubek11 (commented):
:)

Issue #3855, opened 2022-06-10
Greenfield API: exchange rates endpoint


ndeet:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/3837

<div type='discussions-op-text'>

<sup>Originally posted by **ndeet** June  8, 2022</sup>
We have a somehow [undocumented exchange endpoint](https://github.com/btcpayserver/btcpayserver-doc/issues/1124) that was inspired by BitPay but needs to be store specific /api/rates/?storeId=XYZ, e.g. https://testnet.demo.btcpay.tech/api/rates?storeId=46y7JuD7dUHEN1BxeAmtReA5zjUzSho7HVXVteQr1XUW

The discussion here is about if we want that endpoint also to be available in Greenfield API. I think it would make sense but eager to hear what you think.</div>

ndeet:
As discussed with @Kukks we should also provide exchange rates in Greenfield API.

Bangalisch:
@Kukks takes this on

bolatovumar:
What would be different in the Greenfield version of this endpoint vs the one that already exists? Is it the route so it's like `/api/v1/stores/{storeId}/rates`?

ndeet:
Aside from a more consistent api endpoint inside Greenfield namespace/paths, the exchange rates will be based depending on the rate provider (and rating rules) from the store settings which would be imo expected and allows for customization. 
While the legacy endpoint works it is unclear which exchange rate source it uses and it is not as flexible and completely outside of the Greenfield endpoint paths structure. It would make ugly hacks necessary for Greenfield libraries (e.g. our PHP library) to use a different root path for the endpoint.

PR #3931, opened 2022-07-05
Greenfield: Store Rates Config

Kukks:

PR #4550, opened 2023-01-22
Greenfield: Add store rates api

Kukks:
Close https://github.com/btcpayserver/btcpayserver/issues/3855


NicolasDorier:
Errors should be a greenfield error and HTTP error.

Kukks:
> 

No because:
* it is a multi rate endpoint, we should return at least the successful ones
* the type was already used in a separate endpoint

NicolasDorier:
Rates aren't string, but decimal because `[JsonConverter(typeof(NumericStringJsonConverter))]` isn't set.
We might fix this one, but different PR.

This come from https://github.com/btcpayserver/btcpayserver/pull/3931

PR #4552, opened 2023-01-23
Checkout v2: UI fixes

dennisreimann:
Fixes for two Checkout v2 UI issues that @astupidmoose reported via Mattermost:

## Fix scientific notation in amount display

grafik

## Fix initial value getting replaced after copying and then switching the payment method

altcoincopy


dstrukt:
LGTM

Just some notes looking over this PR..

- the functional ability to `Copy` the amount of the invoice could use some more of a hint .. an easier fix is we could tweak the cursor's hover ..  but will continue to think on this.
- The `X` overlap when the amount is too long, very minor and edge case-y, will think on this as well.

dennisreimann:
@NicolasDorier we should include this in the fix release as well.

PR #4553, opened 2023-01-23
BIP21: Uppercase addresses only in QR, not in payment URL

dennisreimann:
The uppercased address/BOLT11 should only be used for the QR code, the payment URI for the link should stay as it is.

Previously we had the BOLT11 uppercased in both URL versions in the BIP21 case, but this leads to  [wallet incompatibility](https://chat.btcpayserver.org/btcpayserver/pl/d9b6w3mo47rezefcjq66bb7owh).

References:

- #2110
- https://bitcoinqr.dev/

### DIFF
--- a/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
@@ -88,7 +88,6 @@ namespace BTCPayServer.Payments.Bitcoin
                 //
                 // The keys (e.g. "bitcoin:" or "lightning=" should be lowercase!
 
-
                 // cryptoInfo.PaymentUrls?.BIP21: bitcoin:bcrt1qxp2qa5?amount=0.00044007
                 model.InvoiceBitcoinUrl = model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls?.BIP21 ?? "";
                 // model.InvoiceBitcoinUrl: bitcoin:bcrt1qxp2qa5?amount=0.00044007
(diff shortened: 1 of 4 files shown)


NicolasDorier:
It's worth making a release for this one later.

BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
```diff
@@ -63,23 +63,24 @@ class Prepare
...
-                model.InvoiceBitcoinUrl = (cryptoInfo.PaymentUrls?.BIP21 ?? "") + lightningFallback;
-                model.InvoiceBitcoinUrlQR = (cryptoInfo.PaymentUrls?.BIP21 ?? "") + lightningFallback
-                    .Replace("LIGHTNING=", "lightning=", StringComparison.OrdinalIgnoreCase);
-
+                model.InvoiceBitcoinUrl = model.InvoiceBitcoinUrlQR = cryptoInfo.PaymentUrls?.BIP21 ?? "";
+                
+                if (!string.IsNullOrEmpty(lightningFallback))
+                {
+                    model.InvoiceBitcoinUrl += $"&{lightningFallback}";
+                    model.InvoiceBitcoinUrlQR += $"&{lightningFallback.ToUpperInvariant().Replace("LIGHTNING=", "lightning=", StringComparison.OrdinalIgnoreCase)}";
```
NicolasDorier, line 81: I think you can remove 
```
{lightningFallback.ToUpperInvariant().Replace("LIGHTNING=", "lightning=", StringComparison.OrdinalIgnoreCase)}";
```

because it is already set that way.

dennisreimann: I'm uppercasing the whole, but reverting the `lightning=` key part afterwards, so that only the BOLT11 is uppercase as a result. That should be correct.

NicolasDorier: Damn this code is confusing, can you add in comments a few comment like what to expect from the lightningFallback and invoiceBitcoinUrl and InvoiceBitcoinUrlQR?

This isn't the first time we have bug in this area, and it's not surprising.

dennisreimann: Done in 69299f375e18f291703a19eafd0634109e9c254a.

NicolasDorier:
@dennisreimann I think I saw two bugs:

1. If there is no amount for the invoice (top-up) then there is no lightning fallback. (is it because LNUrl isn't valid in lighting isn't possible) ?
2. If there is payjoin disabled, and no amount for invoice, then `model.InvoiceBitcoinUrl: bitcoin:bcrt1qxp2qa5dhn7` and the concat give an invalid ULR `bitcoin:bcrt1qxp2qa5dhn7&lighting=blah`. This case doesn't happen because of `1.`. Though it seems to me a bit fragile.

```csharp
                    model.InvoiceBitcoinUrl += $"&{lightningFallback}";
                    // model.InvoiceBitcoinUrl: bitcoin:bcrt1qxp2qa5dhn7?amount=0.00044007&lightning=lnbcrt440070n1...
                    model.InvoiceBitcoinUrlQR += $"&{lightningFallback.ToUpperInvariant().Replace("LIGHTNING=", "lightning=", StringComparison.OrdinalIgnoreCase)}";
```

dennisreimann:
1. is fixed by @kukks in #4541 - I think it's in [this part](https://github.com/btcpayserver/btcpayserver/pull/4541/files#diff-9c2f33ab1ceccf73b65af6053e3b2fe574c8bbba826b7bfe4f1c037eef4b80f1R78) which also updates that area of the code. Here's where I deactivated it, [that part](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UIInvoiceController.UI.cs#L655) is missing the zero amount case.
2. should get addressed before we merge this

chendo:
@dennisreimann I just tested a QR code with capitalised `LIGHTNING` in the param and it breaks Muun. I don't have easy CashApp to test, but QR code should also stick with lowercase `lightning` for better compatibility

chendo:
I had a closer look at the PR and see that it's just the address that will be capitalised, never mind me!

PR #4556, opened 2023-01-23
Checkout v2: Re-add LNURL for top-up invoices

dennisreimann:
My bad, this was excluded erroneously.


dennisreimann:
Will fix the test

PR #4557, opened 2023-01-23
Lowercase the `lightning` param in unified Bitcoin payment links due to apps not detecting Lightning payments otherwise

chendo:
This PR changes the `lightning` query param in unified BIP21 URLs to lowercase as popular wallet apps (confirmed on CashApp, Muun) do not detect the Lightning address otherwise.

I've confirmed that CashApp and Muun (as of Jan 2023) do not detect the Lightning component of the unified BIP21 URL if the `lightning` parameter is capitalised.

I initially ran into the issue where it was odd that scanning the QR worked fine, but the "Open in Wallet" / "Pay in Wallet" button in both old and new checkouts didn't work. A bit of digging and investigation that both CashApp and Muun appear to only expect lowercase `lightning`.

I don't have a working .NET environment but this appears to be the changes that need to be made.


chendo:
Related to #4553, which was opened because I reported the bug; however that PR would break QR codes. This PR only addresses the payment URIs and doesn't touch QR codes because they currently work with CashApp and Muun due to lowercase `lightning` param

chendo:
Whoops, I started this PR yesterday and only sent it today, turns out #4553 was opened yesterday.

PR #4560, opened 2023-01-24
Point of Sale: Improve merchant view

### LINKED ISSUE #3843: Improve Point of Sale Merchant View
Going to start listing out some of the todo items, will likely be updating this ticket:

- [ ] Improve the Point of Sale selector, most important decision, deserves the focus .. might do another exploration of this:
Screen Shot 2022-06-08 at 11 46 58 PM
 
 - [ ] Improve the tip percentage amount input
 
 - [ ] Controversial: Lock tip percentage amount (enables the final view to look better) to `3` or nothing

dennisreimann:
Closes #3843. There are two more TODOs in that issue, but they aren't specified in more detail.

I opted for a more reduced POS style selector: I experimented a bit with it and played through some variants, because I found that it takes up quite some space when done like in Figma.

image


dstrukt:
Thoughts:

- I like the minified POS option selections, and I think it makes sense to have them exposed (as opposed to say a dropdown or other equivalent) if we keep them all in one "app/plugin" like this as well. 
- Do we need an additional descriptor for the `Type`?
- Debating if putting the `App Name` and `Display Title` together above the selection options makes more sense...?
- Agree with your decision to split the `Display Title` and `Currency`

Otherwise, looks great!

dstrukt (changes requested):
Thoughts:

- I like the minified POS option selections, and I think it makes sense to have them exposed (as opposed to say a dropdown or other equivalent) if we keep them all in one "app/plugin" like this as well.
- Do we need an additional descriptor for the Type?
- Debating if putting the App Name and Display Title together above the selection options makes more sense...?
- Agree with your decision to split the Display Title and Currency

Otherwise, looks great!

dennisreimann:
> Do we need an additional descriptor for the Type?

Imho merging the short title with the additional description makes it more concise and the labels as they are seem clear to me.

> Realize it was in the mock, but debating if putting the App Name and Display Title together above the selection options makes more sense...?

I could see that make sense :)

dstrukt:
> Imho merging the short title with the additional description makes it more concise and the labels as they are seem clear to me.
Sounds good, happy to roll with this great quicker improvement and address more in the future.

> I could see that make sense :)
Hahah the common A or B .. both work, could make sense to select the type, then the display you want on that type, but just re-evaluating the step-by-step again with fresh eyes.

Screen Shot 2023-01-24 at 8 40 48 AM

Also, should we rename this section to `Checkout` now perhaps? As `Appearance` no longer makes as much sense without the type selector.

BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -782,6 +782,7 @@ public async Task CanCreateAppPoS()
             s.Driver.FindElement(By.Id("Create")).Click();
             Assert.Contains("App successfully created", s.FindAlertMessage().Text);
 
+            s.Driver.FindElement(By.CssSelector("label[for='DefaultView_Cart']")).Click();
             s.Driver.FindElement(By.CssSelector(".template-item:nth-of-type(1) .btn-primary")).Click();
```
Daedaesde, line 786: R

dennisreimann:
I think we could trim the whole bottom section to something like this:

grafik

## Before

appearance-before

dstrukt:
I like the idea of consolidating the options, and even though it sometimes add more height, I like the section headers, makes the content a bit more parsable imo, e.g.  seeing `Custom Payments`,`Tips` & `Discounts` within the `Checkout` header might be more easy to understand.

Hmm, will think on here...

dennisreimann:
The rationale behind removing the subheaders was that it's only a single toggle per header, so I thought the toggle + label might be enough.

dstrukt:
Ahh yeah, that makes sense .. hmm it's still a bit parse to parse, but..

If we go that route, I'd rearrange:
- Request customer data on checkout (most important item)
- Text to display on the button
- Tips (second most important item)
- Discounts
- Custom Amount

pavlenex:
Concept ACK, just want to say that graphically presenting the types of pos is fantastic design choice.

dennisreimann:
@dstrukt Ok, I'm giving in ;)

grafik

dennisreimann:
Now also hides the options that aren;t relevant for the Keypad-only view: Description and products are hidden in that case.

grafik

dennisreimann:
@dstrukt @pavlenex  This is now ready from my point of view.

dstrukt:
Looks great, will take a look asap.

Great work!

dstrukt:
Wow. This is so much better for such a quick iteration!

Only other thoughts and nits I have looking at this, which aren't necessary to wrap this one up:

- [ ] Should we kill the constraint on the `Products`? Or possibly increase it .. feels like we could get 3-4 on the row, feels a bit more constrained now, but not sure the best way to go about this.
- [ ] Eventually we'll probably want to expose the `customer information` in a similar way that we did for the POS type, it's very easy to parse and a pretty important decision to overall experience. 
- [ ] It still makes sense to keep `Products` above the `Checkout` yeah? Don't know why I'm questioning this, but wanted to raise it as a discussion point

dstrukt (approved):

PR #4561, opened 2023-01-24
Generic Forms

Kukks:
This re-adds the custom forms support and reactors things considerable as there were many bugs. 

Things to note:
* the form id is not exposed when using pos/payment request. The UI is shown under a specific route
* no tempdata usage.
* there were bugs where the form would be in a validated state on first view


dennisreimann:
I'm still thinking about how to make it work with #4568 and have the store branding applied to the form whenever possible. 

The form per se doesn't have the connection to a store, but as custom forms are set up store, this might work. However, we still would need something like a parameter for the generic forms. 

It'd certainly be good to have that possibility, because otherwise the form as an intermediate step would miss the branding that the initial and checkout view have.

Kukks:
All forms are attached to a store. There is a "public" property on custom forms, which lets you use them without apps/pay requests, but the hardcoded ones are always set to false.  That means the form always has a stored available to it

BTCPayServer.Data/Migrations/20230125085242_AddForms.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            migrationBuilder.CreateTable(
+                name: "Forms",
+                columns: table => new
+                {
+                    Id = table.Column<string>(type: "TEXT", nullable: false),
+                    Name = table.Column<string>(type: "TEXT", nullable: true),
+                    StoreId = table.Column<string>(type: "TEXT", nullable: true),
+                    Config = table.Column<string>(type: migrationBuilder.IsNpgsql() ? "JSONB" : "TEXT", nullable: true),
+                    Public = table.Column<bool>(type: "INTEGER", nullable: false)
```
dennisreimann, line 36:
```suggestion
                    Public = table.Column<bool>(nullable: false)
```

dennisreimann:
Added the branding aspect to the non-public forms and cleaned up the views a bit. 

The test failures seem to be DB related, but at first glance I don't see what might be wrong …  I [cleaned up one field in the migration](https://github.com/btcpayserver/btcpayserver/pull/4561/commits/7e9d33bc705c8db9ad0200b7edfe6a4b8321ec77), but there are two errors remaining which look like they might be related to the recent blob changes — any idea @Kukks @NicolasDorier?

```
  Failed CanDoRateSourceMigration [7 s]
  Error Message:
   Npgsql.PostgresException : 42883: operator does not exist: text ->> unknown

POSITION: 220
  Stack Trace:
     at Npgsql.Internal.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|215_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
```

```
  Failed CanCreateSqlitedb [1 s]
  Error Message:
   System.NotSupportedException : SQLite does not support this migration operation ('AlterColumnOperation'). See http://go.microsoft.com/fwlink/?LinkId=723262 for more information.
  Stack Trace:
     at Microsoft.EntityFrameworkCore.Migrations.SqliteMigrationsSqlGenerator.Generate(AlterColumnOperation operation, IModel model, MigrationCommandListBuilder builder)
```

dennisreimann (approved):
tACK

dennisreimann:
Added validation and adapted the validation styles to use the Bootstrap 5 classes :)

grafik

BTCPayServer.Data/Migrations/20230125085242_AddForms.cs
```diff
@@ -0,0 +1,53 @@
...
+#nullable disable
+
+namespace BTCPayServer.Migrations
+{
+    [DbContext(typeof(ApplicationDbContext))]
+    [Migration("20230125085242_AddForms")]
+    public partial class AddForms : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
NicolasDorier, line 18: missing explicit length for the Id column if it is mysql

BTCPayServer/Controllers/UIPaymentRequestController.cs
```diff
@@ -213,42 +217,33 @@ public async Task<IActionResult> ViewPaymentRequestForm(string payReqId)
...
 
             var prBlob = result.GetBlob();
+            if (prBlob.FormResponse is not null)
+            {
+                return RedirectToAction("PayPaymentRequest", new {payReqId});
+            }
             var prFormId = prBlob.FormId;
-            var formConfig = prFormId is null ? null : Forms.UIFormsController.GetFormData(prFormId)?.Config;
-            switch (formConfig)
+            var formData = prFormId is null ? null : (await FormDataService.GetForm(prFormId));
```
NicolasDorier, line 225: you can decrease the number of branches and code, by handling the case of `prFormId` is null inside `GetForm`. Because as is the terniary is repeated all over the place.

NicolasDorier: Actually, `GetForm` seems to already handle that.

BTCPayServer/Forms/FormDataService.cs
```diff
@@ -32,4 +50,73 @@ public class FormDataService
...
+    
+    public async Task<List<FormData>> GetForms(string storeId)
+    {
+        ArgumentNullException.ThrowIfNull(storeId);
+        await using var context = _applicationDbContextFactory.CreateContext();
+        return await context.Forms.Where(data => data.StoreId == storeId).ToListAsync();
+    }
+
+    public async Task<FormData?> GetForm(string storeId, string id)
+    {
```
NicolasDorier, line 76: if `id` is null, you should return null. Also make `string id` nullable. You are actually passing a null potentially in some controller.

BTCPayServer/Forms/UIFormsController.cs
```diff
@@ -1,109 +1,203 @@
...
-            },
-            { } formId when formId == GenericFormOption.Email.ToString() => new FormData
-            {
-                Config = FormDataService.StaticFormEmail.ToString(),
-                Id = GenericFormOption.Email.ToString(),
-                Name = "Provide your email address",
-            },
-            _ => null
+            Currency = form.GetFieldByName("internal_currency")?.Value ?? store.GetStoreBlob().DefaultCurrency,
+            Amount = string.IsNullOrEmpty(amt) ? null : decimal.Parse(amt, CultureInfo.InvariantCulture),
```
NicolasDorier, line 196: I like the idea, though why is it called `internal_`?
Also maybe in later PR we can make this more generic such that the form can configure all fields of CreateInvoiceRequest.

NicolasDorier:
I made a few minor observations, after it seems good to be merged.

PR #4575, opened 2023-01-31
Fix a bunch of open redirect and incorrect formatting of wallet dialogs

NicolasDorier:
There is some open redirect vulnerability around there, nothing very severe, but better to clean them up.

Step to Reproduce:
* Login your account on https://mainnet.demo.btcpayserver.org/login
* Click the link  https://mainnet.demo.btcpayserver.org/recovery-seed-backup?cryptoCode=BTC&mnemonic=above&passphrase=&isStored=false&requireConfirm=true&returnUrl=//evil.com
* Check the "I have written down my recovery phrase and stored it in a secure location"
* Then click Done

You will be redirected to evil.com.

Credits to https://github.com/gonzxph @gonzxph
Reference: https://huntr.dev/bounties/1a0403b6-9ec9-4587-b559-b1afba798c86/

Also fix https://github.com/btcpayserver/btcpayserver/issues/4577

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -203,7 +203,6 @@ namespace BTCPayServer.Tests
         {
             var isImport = !string.IsNullOrEmpty(seed);
             GoToWalletSettings(cryptoCode);
-
             // Replace previous wallet case
             if (Driver.PageSource.Contains("id=\"ChangeWalletLink\""))
             {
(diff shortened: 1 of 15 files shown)


NicolasDorier:
Please reviewers, take a look if I'm not missing some open redirect.

dennisreimann:
I did just a quick check, but it looks like that the notifications controller also has [one more open redirect](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UINotificationsController.cs#L154).

dennisreimann (approved):
Just checked, the notification one might link out to GitHub (new release) or a block explorer, so that's OK.

PR #4587, opened 2023-02-02
Add missing CORS to ln address of BTCPay (Compatibility Beach Wallet)

NicolasDorier:
We forgot to add CORS to ln address, resulting in incompatibility with a few wallet such as Beach Wallet.
@openoms

https://twitter.com/openoms/status/1620815072208760834?s=20&t=uOeORe0X_o38ubWQ_r-iQg

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -26,6 +26,7 @@ using BTCPayServer.Services.Rates;
 using BTCPayServer.Services.Stores;
 using LNURL;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Cors;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using NBitcoin;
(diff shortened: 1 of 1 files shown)


openoms:
Good fix! To be clear this change was found to be needed first for the snort.social nostr client.
The Bitcoin Beach Wallet was already able to send to the BTC Pay Server LN addresses, but needed CORS activated also on the server side:
https://github.com/GaloyMoney/charts/pull/2487

openoms:
@NicolasDorier this fixes the ability to fetch the lnurl, but now comes up with an error on the path where the invoice is sent:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://pay.diynodes.com/BTC/UILNURL/pay/i/BFcRdppdTFxzHvgmqrzKWw?amount=500000. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.
```

I was able to fix this by adding
```
add_header 'Access-Control-Allow-Origin' '*';
```
to the path:
```
 location /BTC/UILNURL/pay/i
```
in my nginx config so presume that path (or some above) needs to be added as well.

petzsch:
I observed the same with snort.social
https://github.com/btcpayserver/btcpayserver/issues/4615

NicolasDorier:
Will release new version today fixing this for good!

PR #4588, opened 2023-02-02
Crowdfund: View updates

dennisreimann:
Improve store branding and remove the card styles, because they had borders which seemed like visual clutter. Other than that I made some changes to the header section and cleaned up the markup and indentation.

## Before

before

## After

after

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -893,7 +893,7 @@ namespace BTCPayServer.Tests
             Assert.Equal(2, windows.Count);
             s.Driver.SwitchTo().Window(windows[1]);
 
-            Assert.Equal("currently active!",
+            Assert.Equal("Currently active!",
                 s.Driver.FindElement(By.CssSelector("[data-test='time-state']")).Text);
 
             s.Driver.Close();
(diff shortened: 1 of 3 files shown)


dstrukt:
## UPDATES TO COME

Not 100% sure if we should try to match the logo / store name, been debating rolling with a different direction for the Crowdfund..

Few tasks:

- [ ] Let's bump the description up to whatever our `16px` text class is, currently just a `<p>` as 14px, and it's too small

- [ ] Kill the "card" entirely, and use the darkest/lightest background color for the background of the page

- [x] Add `Details` title to the left section

- [ ] Adjust the Crowdfund Title / Description to sit below the main image

- [ ] Restrict the main image to a landscape (16:9 perhaps?) ratio, and scale down the image accordingly, to help improve the square ratio

- [ ] Update the Custom amount card

- [x] More column padding between the `Details` and `Contribute` columns

- [ ] Center align the `Last Updated copy`

Corresponding Exploratory Mock:

dstrukt (changes requested):
See above.

dennisreimann:
@dstrukt I made the changes except the main image constraint. Imho people should upload and display a size that works and we shouldn't interfere here. Constraining it might also break existing campaigns — let's keep it as it is I'd say.

dstrukt:
> @dstrukt I made the changes except the main image constraint. Imho people should upload and display a size that works and we shouldn't interfere here. Constraining it might also break existing campaigns — let's keep it as it is I'd say.

Fair, and awesome, will try to get this reviewed later this evening + any other design ideas, so we can keep pushing it forward!

PR #4594, opened 2023-02-03
Label Manager component

### LINKED ISSUE #4464: [Feature] Ability to add labels to addresses
When we create a new address, we should allow users to attach labels similar to wasabi wallet.
Those labels will propagate to transactions when money is received to them.

We already laid the ground for it with @Kukks so I think at this point it's just a UX/UI issue.

Kukks:
closes #4464


dennisreimann (approved):
tACK. Also unified the Receive view with what we have in Checkout v2.

grafik

dstrukt:
I'd like to improve our label UI components specifically, but overall, looks great!

dennisreimann:
@dstrukt Sure, go ahead. 

We might also want to add a short info note there, that simply adding the labels already attaches them to the address seen above. That was one thing that wasn't clear to me when I initially used it.

NicolasDorier:
image

NicolasDorier:
There are warnings

NicolasDorier:
Missing selenium tests

NicolasDorier:
When money is received on the address, the label isn't applied to the incoming transaction.

BTCPayServer/wwwroot/main/site.css
```diff
@@ -695,3 +695,78 @@ input:checked + .btcpay-list-select-item {
...
+}
+
+/* Payment Box */
+.payment-box {
+    --qr-size: 256px;
+    --icon-size: 64px;
+    --icon-border-size: var(--btcpay-space-xs);
+    --icon-border-color: var(--btcpay-white);
+    
+    max-width: 300px;
```
dennisreimann, line 723:
```suggestion
    max-width: 320px;
```

NicolasDorier:
I think you should associate the label to the incoming transaction, else this PR doesn't make lot's of sense.
Don't try to do it generically for now, just for the scope of this PR should be enough.

Kukks:
> I think you should associate the label to the incoming transaction, else this PR doesn't make lot's of sense. Don't try to do it generically for now, just for the scope of this PR should be enough.

done

PR #4595, opened 2023-02-03
Adapt cookie auth to work with same API permission system

### LINKED ISSUE #3512: [`Roles`] Error 403 while `issue refund` as `Guest`
**Description:**
Guest users cannot access the refund page anymore.

**Steps to reproduce the reported bug:**
1. Go to any invoice under `/invoices/ID` as a Guest.
2. Click on `Issue Refund`.
3. See `403 - DENIED` page.

image

**Expected behavior:**

Instead of 403 error, a guest user could issue the refund and send the link to the customer in case of a partial paid invoices

image

**Environment:**
- BTCPay Server Version: ` v1.4.6.0`
- Deployment Method: Docker
- Browser: Chrome

Kukks:
fixes #3512 

This PR merges the approach of permissions used in greenfield with the cookie auth used on the website. This will allow us to build a similar experience of fine-grained permissions for users on a server and store level through custom roles and custom  store roles.

### DIFF
--- a/BTCPayServer/Data/StoreDataExtensions.cs
+++ b/BTCPayServer/Data/StoreDataExtensions.cs
@@ -1,8 +1,10 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Linq;
 using System.Text;
+using BTCPayServer.Client;
 using BTCPayServer.Payments;
 using BTCPayServer.Payments.Lightning;
 using BTCPayServer.Services.Rates;
@@ -14,6 +16,33 @@ namespace BTCPayServer.Data
 {
     public static class StoreDataExtensions
     {
+        public static PermissionSet GetPermissionSet(this StoreData store)
+        {
+            ArgumentNullException.ThrowIfNull(store);
+            if (store.Role is null)
+                return new PermissionSet();
+            return new PermissionSet(store.Role == StoreRoles.Owner
+                        ? new[]
+                        {
+                        Permission.Create(Policies.CanModifyStoreSettings, store.Id),
+                        Permission.Create(Policies.CanTradeCustodianAccount, store.Id),
+                        Permission.Create(Policies.CanWithdrawFromCustodianAccounts, store.Id),
+                        Permission.Create(Policies.CanDepositToCustodianAccounts, store.Id)
+                        }
+                        : new[]
+                        {
+                        Permission.Create(Policies.CanViewStoreSettings, store.Id),
+                        Permission.Create(Policies.CanModifyInvoices, store.Id),
+                        Permission.Create(Policies.CanViewCustodianAccounts, store.Id),
+                        Permission.Create(Policies.CanDepositToCustodianAccounts, store.Id)
+                        });
+        }
+        public static bool HasPermission(this StoreData store, string permission)
+        {
+            ArgumentNullException.ThrowIfNull(store);
+            return store.GetPermissionSet().Contains(permission, store.Id);
+        }
+
 #pragma warning disable CS0618
         public static PaymentMethodId? GetDefaultPaymentId(this StoreData storeData)
         {
--- a/BTCPayServer/Security/CookieAuthorizationHandler.cs
+++ b/BTCPayServer/Security/CookieAuthorizationHandler.cs
@@ -13,8 +13,6 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Routing;
-using NBitpayClient;
-using Policy = Amazon.Auth.AccessControlPolicy.Policy;
 
 namespace BTCPayServer.Security
 {
@@ -115,14 +113,8 @@ namespace BTCPayServer.Security
             }
 
             // Fall back to user prefs cookie
-            if (storeId == null)
-            {
-                storeId = _httpContext.GetUserPrefsCookie()?.CurrentStoreId;
-            }
-
-            if (string.IsNullOrEmpty(storeId))
-                storeId = null;
-            if (storeId != null)
+            storeId ??= _httpContext.GetUserPrefsCookie()?.CurrentStoreId;
+            if (!string.IsNullOrEmpty(storeId))
             {
                 store = await _storeRepository.FindStore(storeId, userId);
             }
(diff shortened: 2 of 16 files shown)


dennisreimann (approved):
tACK. @Kukks can you take one more look at ebd3a41eabe03a217bd725252e9de09e469c94fb to verify my changes?

NicolasDorier:
@Kukks I refactored it, seems didnt' break anything:

Basically the same as you did before, but I better handle the `Unscoped` stuff that was confusing.
Also, I added a `store.HasPermission()` method, and use this for Greenfield as well.

The idea is that after we can modify its implementation easily, if we make new roles or custom roles.

BTCPayServer/Data/StoreDataExtensions.cs
```diff
@@ -14,6 +16,33 @@ namespace BTCPayServer.Data
...
+        public static PermissionSet GetPermissionSet(this StoreData store)
+        {
+            ArgumentNullException.ThrowIfNull(store);
+            if (store.Role is null)
+                return new PermissionSet();
+            return new PermissionSet(store.Role == StoreRoles.Owner
+                        ? new[]
+                        {
+                        Permission.Create(Policies.CanModifyStoreSettings, store.Id),
+                        Permission.Create(Policies.CanTradeCustodianAccount, store.Id),
```
Kukks, line 28: Can we group all these custodian permissions under a parent `managecustodian` permission?

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -112,49 +113,49 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
-            if (string.IsNullOrEmpty(storeId))
-                storeId = null;
-            if (storeId != null)
+            if (!string.IsNullOrEmpty(storeId))
             {
                 store = await _storeRepository.FindStore(storeId, userId);
             }
 
-            switch (requirement.Policy)
+            if (Policies.IsServerPolicy(policy) && isAdmin)
```
Kukks, line 132: Note: we should change this and make it generic so that we can have custom roles that allow a subset of server permissions

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -112,49 +113,49 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
-            if (storeId == null)
+            storeId ??= _httpContext.GetUserPrefsCookie()?.CurrentStoreId;
+
+            var policy = requirement.Policy;
+            bool requiredUnscoped = false;
+            if (policy.EndsWith(':'))
             {
-                storeId = _httpContext.GetUserPrefsCookie()?.CurrentStoreId;
+                policy = policy.Substring(0, policy.Length - 1);
+                requiredUnscoped = true;
```
Kukks, line 123: I dont understand this var, is it named right?

BTCPayServer/Data/StoreDataExtensions.cs
```diff
@@ -14,6 +16,33 @@ namespace BTCPayServer.Data
...
+                        Permission.Create(Policies.CanWithdrawFromCustodianAccounts, store.Id),
+                        Permission.Create(Policies.CanDepositToCustodianAccounts, store.Id)
+                        }
+                        : new[]
+                        {
+                        Permission.Create(Policies.CanViewStoreSettings, store.Id),
+                        Permission.Create(Policies.CanModifyInvoices, store.Id),
```
Kukks, line 35: Should a store guest be able to modify invoices, or only create them?

BTCPayServer/Data/StoreDataExtensions.cs
NicolasDorier: I asked to @woutersamaey about it and he NACK'd it.

image

He may have a point, so in doubt I left this point in potential future PR.

BTCPayServer/Security/CookieAuthorizationHandler.cs
NicolasDorier: So the issue is that the `StoreRole` is a different concept from the ASP.NET Role. (where we only have a single one, Admin)
As such, abstracting it isn't easy or possible.

BTCPayServer/Security/CookieAuthorizationHandler.cs
NicolasDorier: Basically the `***Unscoped` policy means that we require that the user/api key has the full unscoped `***` permission, and not a scoped one on a store for example.

IMHO shouldn't have done this way, instead I should have added a different permission such as `btcpay.user.modifystores` rather than piggy bagging `btcpay.stores.modifystores`

BTCPayServer/Data/StoreDataExtensions.cs
NicolasDorier: This keep the current behavior in the UI, where the Guest can change the status. (I think it makes sense, for example you want the Guest to be able to mark an invoice complete)

I will double check to make sure I didn't changed this behavior.

BTCPayServer/Security/CookieAuthorizationHandler.cs
Kukks: Ah yes, but we can apply permissions to both StoreRole and aspnet Roles later on

BTCPayServer/Data/StoreDataExtensions.cs
Kukks: We already have that tree system for permissions. Im asking to create a parent permission "Manage Custodians" which has the others as sub policies

woutersamaey: "Manage Custodians" is meant for managing the accounts and config, not doing anything in the custodian account itself (like making a trade, desposit or withdrawal). At least, that was my intention.

woutersamaey: To present as a tree, I would suggest we do:
- Custodian Accounts (this is a new parent level)
- - Manage Custodian Accounts (add, edit, configure, remove), this is a rename of "Manage Custodians" to "Manage Custodian Accounts"
- - Deposit
- - WIthdraw
- - Trade

BTCPayServer/Data/StoreDataExtensions.cs
NicolasDorier: I think they already can, by the API.

BTCPayServer/Data/StoreDataExtensions.cs
NicolasDorier: I will leave this topic for another PR to keep this one focused.

BTCPayServer/Data/StoreDataExtensions.cs
NicolasDorier: Confirmed, they already can. Actually the UI only need `CanViewInvoice`, so should change this in a future PR.

Bryansadventure:
Will I be able to withdrawl these funds soon?

NicolasDorier:
follow up on https://github.com/btcpayserver/btcpayserver/pull/4780

PR #4596, opened 2023-02-03
POS: Improve Keypad view

### LINKED ISSUE #3668: [Design] Improved POS "Keypad" View
I'm just posting this here for visibility, feedback, or suggestions as I'm still working through iterations. I'll be posting updates to this issue as I improve things. Not intended to be worked on just yet.

v0.1 Notes:
- Not shown in the prototype, but this view would be scrollable as it goes past the bottom row buttons .. but trying to make sure this quick prototype shows off all the current functionality, will work that into the next recording..
- The buttons would be fixed to the bottom, so would scale appropriately on larger height devices .. we could even center the keypad numbers as well.
- This prototype is missing the "Custom" tip input steps (to be added).
- Need a proper "<" icon (to be added)

BTCPay_POS_Keypad

Screen Shot 2022-04-21 at 11 34 04 AM

dennisreimann:
Closes #3668.

https://user-images.githubusercontent.com/886/216688876-358c51db-c55e-40fd-b8c0-332921c5486f.MOV


dstrukt (changes requested):
## STILL REVIEWING

Few Open Questions:

- Should the discount include the tip as well, or just on the amount itself...? Does that even make sense to calculate it like that?
- Should we try to scale some of the buttons (Tips / Discounts / etc...) to be a little larger on larger devices, similar to how we hide / show certain pieces of the UI?
- Do we want remove the `$` / whatever currency on the total amount breakdown? We already have USD above the total.

### Nits:
- [ ] Thinking we'll want to make the selected tip our primary, like the others, instead of the light outline .. thoughts?
Screen Shot 2023-02-05 at 3 39 46 PM

- [ ] Let's make the `<` out white as well
Screen Shot 2023-02-05 at 3 41 18 PM

- [ ] Let's add a `mb-2` or `mt-2` between the Amount and the price change history, it's a little too tight atm (if we have the vertical space to spare)

- [ ] Let's make the editable discount amount text white, e.g. if it's a `70%` Discount, only the `70%` would be white, and Discount would be grey.

- [ ] Debating if we should also increase the size of the editable `70%` by 1-2 increments .. thoughts?

- [ ] Change to `mt-1` between pills and keypad, little tight there as well.

### Comments:
- Handles extremes well.
Screen Shot 2023-02-05 at 3 45 44 PM

- Love that you can't add a Tip / Discount if you haven't entered an Amount

NicolasDorier (approved):

dennisreimann:
@dstrukt Done

grafik

grafik

PR #4600, opened 2023-02-06
Fix: Payjoin wasn't always properly choosing utxo for optimal change

NicolasDorier:
[This paper](https://eprint.iacr.org/2022/589.pdf) assert that lot's of payjoin transaction exhibit UIH2: Unnecessary input trait.

I was surprised as @Kukks made this PR a while ago https://github.com/btcpayserver/btcpayserver/pull/1473

By writing a proper test, I found out that the implementation was actually broken, and not always avoiding UIH2.
On top of it the implementation was suboptimal.

image

I also updated the tests.

Closing https://github.com/btcpayserver/btcpayserver/discussions/4599

### DIFF
--- a/BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
+++ b/BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
@@ -591,8 +592,10 @@ namespace BTCPayServer.Payments.PayJoin
             if (availableUtxos.Length == 0)
                 return (Array.Empty<UTXO>(), PayjoinUtxoSelectionType.Unavailable);
             HashSet<OutPoint> locked = new HashSet<OutPoint>();
-            // We don't want to make too many db roundtrip which would be inconvenient for the sender
-            int maxTries = 30;
+
+            TimeSpan timeout = TimeSpan.FromSeconds(30.0);
+            Stopwatch watch = new Stopwatch();
+            watch.Start();
 
             // BlockSci UIH1 and UIH2:
             // if min(out) < min(in) then UIH1 else UIH2
@@ -604,7 +607,7 @@ namespace BTCPayServer.Payments.PayJoin
 
             foreach (var availableUtxo in availableUtxos)
             {
-                if (locked.Count >= maxTries)
+                if (watch.Elapsed > timeout)
                     break;
 
                 var utxoValue = availableUtxo.Value.GetValue(network);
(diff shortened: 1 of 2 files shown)


Kukks (approved):

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -585,62 +585,62 @@ public enum PayjoinUtxoSelectionType
...
+                        locked.Add(availableUtxo.Outpoint);
+                        continue;
                     }
                 }
-
-                if (invalid)
+                else
                 {
+                    // UIH2: Unnecessary input, let's try to get an optimal change by using a different output
                     continue;
```
DanGould, line 631: I think this will loop without getting any closer to maxTries because it does not add to `locked`. Therefore maxTries is not the hard cap on tries it once was

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
```diff
@@ -585,62 +585,62 @@ public enum PayjoinUtxoSelectionType
...
-        public async Task<(UTXO[] selectedUTXO, PayjoinUtxoSelectionType selectionType)> SelectUTXO(BTCPayNetwork network, UTXO[] availableUtxos, decimal[] otherInputs, decimal mainPaymentOutput,
+        public async Task<(UTXO[] selectedUTXO, PayjoinUtxoSelectionType selectionType)> SelectUTXO(BTCPayNetwork network, UTXO[] availableUtxos, decimal[] otherInputs, decimal originalPaymentOutput,
             IEnumerable<decimal> otherOutputs)
         {
             if (availableUtxos.Length == 0)
                 return (Array.Empty<UTXO>(), PayjoinUtxoSelectionType.Unavailable);
-            // Assume the merchant wants to get rid of the dust
             HashSet<OutPoint> locked = new HashSet<OutPoint>();
             // We don't want to make too many db roundtrip which would be inconvenient for the sender
             int maxTries = 30;
```
DanGould, line 595: If the goal here is to accomodate the sender a timout may make more sense. Default HTTP request range timeouts are 30 to > 100 seconds. A timeout of say 10s would be well within the acceptable range for every http client I know of. That would allow you to account for other slowdowns, too.

NicolasDorier:
@DanGould indeed, will update code to use a timeout instead.

BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs
NicolasDorier: This was on purpose, as this line doesn't make any DB roundtrip.
Irrelevant now, as I use a timeout.

NicolasDorier:
Updated for timeout of 10s

DanGould:
Looks like the update applied a timeout of 30s, not 10s, but if it takes that long to search through utxos selection it would seem extraordinary to find a solution

https://github.com/NicolasDorier/btcpayserver/blob/7f9dd1c2f59c0802eb8fe280c8c0f1b3987682b2/BTCPayServer/Payments/PayJoin/PayJoinEndpointController.cs#L596

PR #4604, opened 2023-02-07
Swagger doc fix: store website nullable

ndeet:
I tested the API and the `website` field on store creation can be null it seems. Updated the swagger file accordingly.


dennisreimann (approved):
Thanks!

PR #4606, opened 2023-02-08
Changelog 1.7.6

NicolasDorier:
@Kukks @dennisreimann @bolatovumar

I am also thinking of doing a breaking change: https://github.com/btcpayserver/btcpayserver/pull/4607

This route https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Stores_PreviewStoreRateConfiguration

Doesn't actually return a string for the rate, unlike what the documentation says, it returns a number.
It should be a string to match other part of the API.

@ndeet @woutersamaey not breaking anything on your side?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,42 @@
 # Changelog
 
+## 1.7.6
+
+There are two vulnerabilities fixed in this release.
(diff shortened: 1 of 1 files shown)


woutersamaey:
Shouldn't break anything on my end.

PR #4607, opened 2023-02-08
Greenfield: Currency rate should be strings

NicolasDorier:
Breaking change on  /api/v1/stores/{storeId}/rates/configuration/preview

That said, the swagger was documenting a string when it was a decimal anyway...


dennisreimann (approved):
utACK

Kukks (approved):

PR #4608, opened 2023-02-08
Plugins can now build apps

Kukks:
See usage from an app here:
https://github.com/Kukks/BTCPayServerPlugins/tree/tt-app


dennisreimann:
Can you write up the gist of how to make an app (Kukks/BTCPayServerPlugins@eac0cfe734286d1f27c8c9737e4a0d317921c7bb) on the plugins doc?

Kukks:
Sure, once we merge

dennisreimann (commented):
This is really cool and I'm looking forward to use it for e.g. PodServer, which I also tried to convert to an app in order to review and test this …

grafik

The basics work well already, but there are some areas that require additional changes:

- The dashboard tiles refer to controllers/actions and we should allow for using Razor Pages as well.
- The `IApp` interface should be fleshed out with some more use cases and also allow for omitting features like the dachboard display. Migrating the existing plugins for which it makes sense to be an app will help with that.

Kukks:
> This is really cool and I'm looking forward to use it for e.g. PodServer, which I also tried to convert to an app in order to review and test this …
> 
> grafik
> 
> The basics work well already, but there are some areas that require additional changes:
> 
> * The dashboard tiles refer to controllers/actions and we should allow for using Razor Pages as well.
> * The `IApp` interface should be fleshed out with some more use cases and also allow for omitting features like the dashboard display. Migrating the existing plugins for which it makes sense to be an app will help with that.

In my Wabisabi Plugin, I used a UI extension to extend the dashboard, we can do the same for apps that want to show more than the standard item/sale stats

IApp should maybe have a bool SupportDashboardStats {get;} and hide the elements in dashboard if false

LNURl also has hardcoded elements which I could not easily abstract and pluginify

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -383,17 +201,14 @@ private class InvoiceStatsItem
...
+      
+        public static string GetAppOrderId(AppData app) => GetAppOrderId(app.AppType, app.Id);
+        public static string GetAppOrderId(string appType, string appId) =>
+            appType switch
             {
-                nameof(AppType.Crowdfund) => $"crowdfund-app_{app.Id}",
-                nameof(AppType.PointOfSale) => $"pos-app_{app.Id}",
-                _ => throw new ArgumentOutOfRangeException(nameof(app), app.AppType)
+                CrowdfundApp.AppType => $"crowdfund-app_{appId}",
+                PointOfSaleApp.AppType => $"pos-app_{appId}",
```
dennisreimann, line 210: This is here solely for backwards-compatibility, right?

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -383,17 +201,14 @@ private class InvoiceStatsItem
...
+                _ => $"{appType}_{appId}"
```
dennisreimann, line 211: We might want to lowercase this one to keep it consistent with the above?

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -573,21 +391,32 @@ public string SerializeTemplate(ViewPointOfSaleViewModel.Item[] items)
...
+        {
+            return Parse(_HtmlSanitizer, _displayFormatter, template, currency);
+        }
+
+
+        public ViewPointOfSaleViewModel.Item[] GetPOSItems(string template, string currency)
+        {
+            return GetPOSItems(_HtmlSanitizer, _displayFormatter, template, currency);
+        }
+        public static ViewPointOfSaleViewModel.Item[] Parse(HtmlSanitizer htmlSanitizer, DisplayFormatter displayFormatter, string template, string currency)
```
dennisreimann, line 405: In a future update we should try and move those things out of the general `AppService`

BTCPayServer/Services/Apps/AppType.cs
```diff
@@ -1,24 +1,23 @@
...
-        [Display(Name = "Product list with cart")]
-        Cart,
-        [Display(Name = "Keypad only")]
-        Light,
-        [Display(Name = "Print display")]
-        Print
+        public string Description { get;  }
+        public string Type { get; }
+        public bool SupportsSalesStats { get; }
+        public bool SupportsItemStats { get; }
```
dennisreimann, line 14: I opted for adding those flags to indicate whether or not an app supports this data for the dashboard tiles: As the methods that provide that data are async and the AppService does some pre-work for them, we should have a clear way to avoid all of that beforehand. Same goes for the rendering of the tiles, because we wouldn't even want to show them if the app doesn't support those stats.

BTCPayServer/Views/UIStores/Dashboard.cshtml
```diff
@@ -118,8 +119,14 @@
...
-            <vc:app-sales vm="@(new AppSalesViewModel { App = app, InitialRendering = true })"/>
-            <vc:app-top-items vm="@(new AppTopItemsViewModel { App = app, InitialRendering = true })"/>
+            @if (AppService.SupportsSalesStats(app))
+            {
+                <vc:app-sales vm="@(new AppSalesViewModel { App = app, InitialRendering = true })"/>
+            }
+            @if (AppService.SupportsItemStats(app))
+            {
+                <vc:app-top-items vm="@(new AppTopItemsViewModel { App = app, InitialRendering = true })" />
+            }
```
dennisreimann, line 129: Do not even render those tiles if the app doesn't support them.

dennisreimann (approved):
Used this in dennisreimann/btcpayserver-plugin-podserver#3 to make PodServer an app. We might want to do some minor cleanups in a follow-up, but this is a great start and we should integrate it, because the main concern is solved and this works very well :)

BTCPayServer/Views/UIApps/ListApps.cshtml
```diff
@@ -89,21 +90,23 @@
...
-                            @if (app.AppType != AppType.Crowdfund.ToString())
+                            @AppService.GetAvailableAppTypes()[app.AppType]
+                            @{
+                                var viewStyle = @app.ViewStyle;
+                            }
+                            @if (!string.IsNullOrEmpty(viewStyle))
                             {
-                                <span>-</span> 
+                                <span>-</span>
+                                @Safe.Raw(viewStyle)
```
NicolasDorier, line 100: why using `@Safe.Raw`?

NicolasDorier:
will adapt stuff in different PR

PR #4616, opened 2023-02-09
BIP21 + LNURL fix

### LINKED ISSUE #4609: [Bug]: Invalid LNURL created for top-up invoices in new checkout

v1.7.6+ed0ccd6f1 

Docker

### What happened?

Breez and many other wallets can't scan the combined QR code including the LNURL for my top-up invoices.

I scanned a top-up invoice QR code, it failed in Breez, I reported the bug to breez and they tested it with all sorts of wallets with the same outcome.

### Relevant log output

```shell
Parsing the LNURL with https://lightningdecoder.com/ gives:
JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
```

### What browser do you use?

Firefox 109

### Additional information

could this be related to the CORS change introduced in this version?
For testing, create a fresh invoice [here](https://otr.4sats.net/?email=markus@petzsch.eu).

dennisreimann:
In case of the unified invoice, the LNURL wasn't correct — with this change we are simply reusing th one that was issued on invoice creation instead of generating it anew on the fly.

Also fixes missing uppercasing for the QR code in case of non-unified QR. And removes the `lightning:` scheme from the LNURL that's displayed to the user (unifies it with what we do for Onchain and Lightning)

Fixes #4609.

### DIFF
--- a/BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs
+++ b/BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs
@@ -41,6 +41,8 @@ namespace BTCPayServer.Payments.Lightning
 
         public override PaymentType PaymentType => PaymentTypes.LightningLike;
 
+        private const string UriScheme = "lightning:";
+        
         public IOptions<LightningNetworkOptions> Options { get; }
 
         public override async Task<IPaymentMethodDetails> CreatePaymentMethodDetails(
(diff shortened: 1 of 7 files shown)


BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs
```diff
@@ -107,12 +109,13 @@ public override IEnumerable<PaymentMethodId> GetSupportedPaymentMethods()
...
             var paymentMethodId = paymentMethod.GetId();
-            var cryptoInfo = invoiceResponse.CryptoInfo.First(o => o.GetpaymentMethodId() == paymentMethodId);
             var network = _networkProvider.GetNetwork<BTCPayNetwork>(model.CryptoCode);
+            var cryptoInfo = invoiceResponse.CryptoInfo.First(o => o.GetpaymentMethodId() == paymentMethodId);
+            var lnurl = cryptoInfo.PaymentUrls?.AdditionalData["LNURLP"].ToObject<string>();
             model.PaymentMethodName = GetPaymentMethodName(network);
-            model.InvoiceBitcoinUrl = cryptoInfo.PaymentUrls?.AdditionalData["LNURLP"].ToObject<string>();
-            model.InvoiceBitcoinUrlQR = model.InvoiceBitcoinUrl;
-            model.BtcAddress = model.InvoiceBitcoinUrl;
+            model.BtcAddress = lnurl?.Replace(UriScheme, "");
```
Kukks, line 116: We cant do this.  LNURL has bech32 mode, where this would work (`lnurl1...`), but the newer uri scheme would mean the protocol identifier would be cut off. (`lnurlp://`)

dennisreimann: That `UriScheme` is a [const defined as `lightning:`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs#L44), so we should be good here.

PR #4617, opened 2023-02-09
LNURL: Add missing CORS

### LINKED ISSUE #4615: [Bug]: CORS-related - snort.social still doesn't play with our lnaddress implementation

 v1.7.6+ed0ccd6f1

Docker

### What happened?

grafik

1. Add lnaddress to my nostr account
2. click on address in profile, 500 sats and Zap
3. "Failed to load invoice" appears + JS console output from above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox 109

### Additional information

my public nostr profile on snort if you wanna try:
https://snort.social/p/npub1l2q2pyyjlt5qsgwnmk3ga0dme2aafj35j25fk878p58uzanyfcsqqzlgc2

dennisreimann:
In addition to #4587. Closes #4615.

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -81,7 +81,6 @@ namespace BTCPayServer
             _btcPayNetworkJsonSerializerSettings = btcPayNetworkJsonSerializerSettings;
         }
 
-
         [HttpGet("withdraw/pp/{pullPaymentId}")]
         public async Task<IActionResult> GetLNURLForPullPayment(string cryptoCode, string pullPaymentId, string pr, CancellationToken cancellationToken)
         {
@@ -212,6 +211,7 @@ namespace BTCPayServer
 
             return Ok(request);
         }
+
         [HttpGet("pay/app/{appId}/{itemCode}")]
         public async Task<IActionResult> GetLNURLForApp(string cryptoCode, string appId, string itemCode = null)
         {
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -81,7 +81,6 @@ public class UILNURLController : Controller
             _btcPayNetworkJsonSerializerSettings = btcPayNetworkJsonSerializerSettings;
         }
 
-
```
Kukks, line 84: We should it Cors here too?

Kukks (changes requested):

BTCPayServer/Controllers/UILNURLController.cs
dennisreimann: Not sure about the use case. @petzsch Do you know where it can be relevant for Pull Payment + LNURL withdrawal?

petzsch: When you click on my lnaddress (the little bolt symbol is infront of it here): https://snort.social/p/npub1l2q2pyyjlt5qsgwnmk3ga0dme2aafj35j25fk878p58uzanyfcsqqzlgc2

you get a dialog where you can chose an amount and create an invoice for the lightning address... would be super lit to get it working.

dennisreimann: Yes, I fixed that. Kukks and my question now is if you can imagine a use case to add CORS for Pull Payment + LNURL withdrawal as well — If so, we'd also add that.

petzsch: OK, I don't currently see a need for it. At least not in conjunction with ln-address as such payment methods are afaik not specified for it.

If someone wants to create or claim pull payments via JavaScript from a remote site, he might as well use our API for that (and a proper programing language). So I think we are good here.

dennisreimann: Agreed :+1:

PR #4632, opened 2023-02-13
Different icons for notifications

### LINKED ISSUE #2510: Different icons/emoji's for notifications
It would be good if we can distinguish notifications with different icons or emojis.

Screenshot 2021-05-06 at 12 02 48

Currently, we notify users on the following:
 
- New version
- All invoice updates 
- Invoice was paid after expiration  
- Invoice expired with partial payments 
- Invoice has payments that failed to confirm on time
- Invoice was confirmed paid
- Payouts

dennisreimann:
Closes #2510.

grafik


dstrukt:
Realizing I might need to adjust some of the colors, but will get this reviewed and assets swapped either this evening or tomorrow - thanks for picking this one up!

dennisreimann:
Yup, the white on yellow is a bit hard to see, but I think that's the only one. I've added an [icon overview](https://design.btcpayserver.org/design/icons/) here — feel free to just send me the updated ones and I'll integrate them.

NicolasDorier:
DOPPPPE

BTCPayServer.Abstractions/TagHelpers/SVGUse.cs
```diff
@@ -27,7 +27,10 @@ public SVGUse(IUrlHelperFactory urlHelperFactory, HtmlEncoder htmlEncoder, IFile
     public override void Process(TagHelperContext context, TagHelperOutput output)
     {
         var attr = output.Attributes["href"].Value.ToString();
-        var filePath = attr.Substring(attr.IndexOf("~") + 1, attr.IndexOf("#"));
+        var symbolIndex = attr!.IndexOf("#", StringComparison.InvariantCulture);
+        var start = attr.IndexOf("~", StringComparison.InvariantCulture) + 1; 
+        var end = symbolIndex != -1 ? symbolIndex : attr.Length - 1;
```
NicolasDorier, line 32: I don't believe this is correct. The second parameter of `Substring` is a length, not an index. As such, it should be `end - start`.

For input: `blah~/t#bla`, then you have `start=5`, `end=7`, you need to pass to `Substring(5,2)` (`end-start`)
""

dennisreimann: Thanks, fixed in [0e020ea](https://github.com/btcpayserver/btcpayserver/pull/4632/commits/0e020eaabdb251a3d4d0d29f05bf1529f28cef93).

pavlenex:
@dennisreimann are the three icons the only ones we have, could you post some more screenshots, on a very old PC so can't even build the PR locally.

Concept ACK, just don't think the second icon in opening screenshot communicates things properly, To me, the icon communicates "wait/pending" it will be fine, but it should rather be "error, requires action" kind of thing, @dstrukt may be able to prove me wrong though.

dennisreimann:
I wasn't able to produce the other states, but you can take a look at the other icons in the icon overview I mentioned above.

dstrukt:
> @dennisreimann are the three icons the only ones we have, could you post some more screenshots, on a very old PC so can't even build the PR locally?
> 
> Concept ACK, just don't think the first icon in opening screenshot communicates things properly, To me, the icon communicates "wait/pending" it will be fine, but it should rather be "error, requires action" kind of thing, there's nothing to wait here, invoice expired and you should take action either accept it or refund; @dstrukt may be able to prove me wrong though.

No I think you're correct, it's not the right icon for this - I think I exported the wrong thing, but fortunately it's an easy swap.

I haven't done a proper round of testing myself - was planning to catch those details, but will make sure that's addressed!

NicolasDorier:
@dstrukt you can take a look soonish? we plan making 1.8 this week.

dstrukt:
Absolutely, sorry for the delay, wasn't in front of the comp much this weekend, will take care of it now.

dstrukt:
Updated icons have been sen to @dennisreimann via Mattermost

dstrukt (approved):
Pending icon swaps, this LGTM!

dennisreimann:
New states after swap:

grafik

PR #4633, opened 2023-02-13
Checkout v2: Fix automatic redirect after paid

dennisreimann:
Should use captured `storeLink` property instead of the `merchantRefLink`.


dstrukt (approved):
tACK

LGTM!

PR #4635, opened 2023-02-14
Fix secondary order XSS: Harden file type inputs

NicolasDorier:
I was thinking we could avoid validating file upload.
But nope... @nayefhmoodh has shown me a secondary order XSS which can even avoid the strictest CSP by crafting a rogue PDF file...

Credit to @nayefhmoodh

https://huntr.dev/bounties/9464e3c6-961d-4e23-8b3d-07cbb31de541/

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.cs
+++ b/BTCPayServer/Controllers/UIStoresController.cs
@@ -27,6 +27,7 @@ using BTCPayServer.Services.Stores;
 using BTCPayServer.Services.Wallets;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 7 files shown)


NicolasDorier:
I can't detect whether a file is CSS. However, I can add some CSS trash as first line of a so called CSS file. If JS run, it will just crash first line... will do this.

NicolasDorier:
Ok I managed to make uploaded JS unexecutable with `X-Content-Type-Options: nosniff`

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -691,7 +705,19 @@ public IActionResult GeneralSettings()
 
             if (model.CssFile != null)
             {
-                if (model.CssFile.ContentType.Equals("text/css", StringComparison.InvariantCulture))
+                if (model.CssFile.Length > 1_000_000)
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = "The uploaded file should be less than 1MB";
```
dennisreimann, line 710: We can lower this limit here as 100KB should already be more than enough. (Bootstrap as a whole is > 300KB)

PR #4654, opened 2023-02-16
Do not generate pj= if no utxo available in the wallet

### LINKED ISSUE #4651: Payjoin payment link generated even when hot wallet has no UTXOs
Since receiver cannot contribute UTXOs for payjoining if the hot wallet is empty, this behaviour results in an error on the sender wallet.

BTCPay version: 1.7.11

Steps to reproduce:
- connect/create hot wallet
- ensure it is empty
- ensure payjoin is enabled
- generate a QR/payment link

You will observe that a BIP-21 URI with pj endpoint link is generated, but if you try to pay to it, the payment fails (screenshots) attached.

Screenshot 2023-02-15 at 11 27 30

Screenshot 2023-02-12 at 23 55 44

IMG_5420

Suggested fix: do not generate a payjoin payment link/QR if the connected hot wallet is empty

Hope this info is useful...

NicolasDorier:
I am not sure if this should be merged. As clients need to be able to handle that. 
But it seems bluewallet crash https://github.com/BlueWallet/BlueWallet/issues/5313

Close #4651.

### DIFF
--- a/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Bitcoin/BitcoinLikePaymentHandler.cs
@@ -10,6 +10,7 @@ using BTCPayServer.Models;
 using BTCPayServer.Models.InvoicingModels;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Invoices;
+using Dapper;
 using NBitcoin;
 using NBitcoin.DataEncoders;
 using NBXplorer.Models;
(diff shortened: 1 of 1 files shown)


yashrajd:
BTCPay is essentially forcing wallets to 'handle that' (how to is another question). Also technically uneducated users of BTCPay would not know what's going wrong as there are no prompts/alerts about empty hot wallet...may think payjoins are unreliable. Agree re: blue wallet

Issue #4497, opened 2023-01-02
[Bug]: Crash when accessing /register endpoint after a fresh install on Ubuntu


pox:
### What is your BTCPay version?

Running v1.7.3

### How did you deploy BTCPay Server?

I have followed the [Extended Manual Setup](https://docs.btcpayserver.org/Deployment/ManualDeploymentExtended/) guide to install on Ubuntu. All the steps seemingly succeeded.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
```

### What happened?

Once the server is up and running (`syslog` does not show any errors) I try to GET the main page, get redirected to `/register` and when the browser requests `/register` the server crashes.

Logs attached below (trimmed because the stack dump is huge).

### How did you encounter this bug?

See above.

### Relevant log output

```shell
Jan 01 18:11:52 btcnode run.sh[55888]: info: Configuration:  Data Directory: /home/bitcoin/.btcpayserver/Main
Jan 01 18:11:52 btcnode run.sh[55888]: info: Configuration:  Configuration File: /etc/btcpay/btcpay.config
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Loading plugins from /home/bitcoin/.btcpayserver/Plugins
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Adding and executing plugin BTCPayServer - 1.7.3.0
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Adding and executing plugin BTCPayServer.Plugins.Shopify - 1.7.3.0
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Adding and executing plugin BTCPayServer.Plugins.Crowdfund - 1.7.3.0
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Adding and executing plugin BTCPayServer.Plugins.PayButton - 1.7.3.0
Jan 01 18:11:53 btcnode run.sh[55888]: info: BTCPayServer.Plugins.PluginManager: Adding and executing plugin BTCPayServer.Plugins.PointOfSale - 1.7.3.0
[...]

PR #4655, opened 2023-02-16
Fix build and run scripts

NicolasDorier:
Fix https://github.com/btcpayserver/btcpayserver/issues/4497


NicolasDorier:
Ping @raucao can you check those scripts? pushing that on master.

raucao:
Confirmed to be working with that commit. :+1:

logancom (commented):
Na

PR #4668, opened 2023-02-20
POS improvements

dennisreimann:
Two additions for the POS:

- Prevent redirect URL from containing query parameters (which might be added by the generic forms integration)
- Add POS data for keypad view: Uses the same data model as the cart view and allows for better accounting, because the values are tracked individually.

### DIFF
--- a/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
+++ b/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
@@ -276,7 +276,7 @@ namespace BTCPayServer.Plugins.PointOfSale.Controllers
                             string.IsNullOrEmpty(notificationUrl) ? settings.NotificationUrl : notificationUrl,
                     RedirectURL = !string.IsNullOrEmpty(redirectUrl) ? redirectUrl
                         : !string.IsNullOrEmpty(settings.RedirectUrl) ? settings.RedirectUrl
-                        : Request.GetDisplayUrl(),
+                        : Url.ActionLink(nameof(ViewPointOfSale), "UIPointOfSale", new { appId, viewType }),
                     FullNotifications = true,
                     ExtendedNotifications = true,
                     PosData = string.IsNullOrEmpty(posData) ? null : posData,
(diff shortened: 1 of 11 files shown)


Kukks (approved):

NicolasDorier:
> Prevent redirect URL from containing query parameters (which might be added by the generic forms integration)

Forms shouldn't add query strings, how is that possible?

NicolasDorier:
> Add POS data for keypad view: Uses the same data model as the cart view and allows for better accounting, because the values are tracked individually.

Does it works with form? Because afaik, Form is setting the metadata, so it would probably override the posData you are passing there.

dennisreimann:
> Does it works with form? Because afaik, Form is setting the metadata, so it would probably override the posData you are passing there.

Yes, I just checked, `posdata` is a separate property.

> Forms shouldn't add query strings, how is that possible?

This is now fixed in [d6a72cc](https://github.com/btcpayserver/btcpayserver/pull/4668/commits/d6a72cc026ef2b36e4edfb70f738fb21b941a95b): Before, all incoming parameters were appended to the query string, now  the query and form params are separated — we are using the PostRedirect here, which already supports that separation.

dennisreimann:
@kukks can you check/review once more? The most important part that changes is in the [POSForm action](https://github.com/btcpayserver/btcpayserver/pull/4668/files#diff-365a5712673fc33daef8d1d5fc72b795cebdef338a3d770ecb1c1efb6fc28ef8R321) where we now distinguish between the query and form params. That way the form values don't end up in the URL.

BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
```diff
@@ -331,10 +331,9 @@ public async Task<IActionResult> POSForm(string appId, PosViewType? viewType = n
                 return RedirectToAction(nameof(ViewPointOfSale), new { appId, viewType });
             }
             
-            var myDictionary = Request.Form
+            var formParameters = Request.Form
                 .Where(pair => pair.Key != "__RequestVerificationToken")
                 .ToDictionary(p => p.Key, p => p.Value.ToString());
```
NicolasDorier, line 336: `ToMultiValueDictionary` is better as there might be multiple form with same key.

Kukks:
We need to check what happens in cases where the form has fields that conflict with pos form parameters.

dennisreimann (commented):
Tested and added one more. As far as I can see, this one is Rekukksified and it Whacks The Mole™

PR #4669, opened 2023-02-20
UI: Remove highlight for valid fields

dennisreimann:
I recently updated the validation defaults to use the Bootstrap 5 classes. This removes the highlight for validated fields, because imho it's too much to explicitely highlight fields with correct input and it also made these fields flash green on form submit, when the validation is done.


dstrukt (approved):
LGTM!

PR #4671, opened 2023-02-20
Checkout: Allow NFC/LNURL-W whenever LNURL is available

### LINKED ISSUE #4443: [Checkout V2] Add NFC support
Will update

dennisreimann:
With what we have in master right now, we display NFC only for top-up invoices. With these changes, we display NFC in all cases, where LNURL is available. Note that this hides LNURL from the list of selectable payment methods, it's only available to use the NFC — and explicitely selectable only for the edge case of top-up invoice + non-unified QR (as before).

Rationale: Now that we got NFC tightly integrated, it doesn't make sense to support the NFC experience only for top-up invoices. With this we bring back LNURL for regular invoices as well, but don't make it selectable and use it only for the NFC functionality.

Closes #4443.

There are many different cases. Basically it's a test matrix permutating:

- Unified/non-unified
- Regular vs. top-up invoice
- LNURL available as payment method

### Without LNURL

unified-without-lnurl

### With LNURL

unified-with-lnurl

### Top-up

unified-topup

### Without LNURL

non-unified-without-lnurl-bitcoin non-unified-without-lnurl-lightning png

### With LNURL

non-unified-with-lnurl-bitcoin non-unified-with-lnurl-lightning

### Top-up

non-unified-topup-bitcoin non-unified-topup-lnurl

## Lightning with LNURL

ln-with-lnurl

## Just Lightning without LNURL

lightning-without-lnurl

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -849,6 +849,7 @@ namespace BTCPayServer.Controllers
             paymentMethodHandler.PreparePaymentModel(model, dto, storeBlob, paymentMethod);
             model.UISettings = paymentMethodHandler.GetCheckoutUISettings();
             model.PaymentMethodId = paymentMethodId.ToString();
+            model.PaymentType = paymentMethodId.PaymentType.ToString();
             var expiration = TimeSpan.FromSeconds(model.ExpirationSeconds);
             model.TimeLeft = expiration.PrettyPrint();
             return model;
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -652,17 +652,19 @@ namespace BTCPayServer.Controllers
             var lnurlId = PaymentMethodId.Parse("BTC_LNURLPAY");
             if (paymentMethodId is null)
             {
-                var enabledPaymentIds = store.GetEnabledPaymentIds(_NetworkProvider)
-                    .Where(pmId => storeBlob.CheckoutType == CheckoutType.V1 ||
-                        // Exclude LNURL for Checkout v2 + non-top up invoices
-                        pmId != lnurlId || invoice.IsUnsetTopUp())
-                    .ToArray();
+                var enabledPaymentIds = store.GetEnabledPaymentIds(_NetworkProvider).ToArray();
 
                 // Exclude Lightning if OnChainWithLnInvoiceFallback is active and we have both payment methods
-                if (storeBlob is { CheckoutType: CheckoutType.V2, OnChainWithLnInvoiceFallback: true } &&
-                    enabledPaymentIds.Contains(btcId) && enabledPaymentIds.Contains(lnId))
+                if (storeBlob is { CheckoutType: CheckoutType.V2, OnChainWithLnInvoiceFallback: true })
                 {
-                    enabledPaymentIds = enabledPaymentIds.Where(pmId => pmId != lnId).ToArray();
+                    if (enabledPaymentIds.Contains(btcId) && enabledPaymentIds.Contains(lnId))
+                    {
+                        enabledPaymentIds = enabledPaymentIds.Where(pmId => pmId != lnId).ToArray();
+                    }
+                    if (enabledPaymentIds.Contains(btcId) && enabledPaymentIds.Contains(lnurlId))
+                    {
+                        enabledPaymentIds = enabledPaymentIds.Where(pmId => pmId != lnurlId).ToArray();
+                    }
                 }
 
                 PaymentMethodId? invoicePaymentId = invoice.GetDefaultPaymentMethod();
@@ -688,7 +690,7 @@ namespace BTCPayServer.Controllers
                 if (paymentMethodId is null)
                 {
                     paymentMethodId = enabledPaymentIds.FirstOrDefault(e => e.CryptoCode == _NetworkProvider.DefaultNetwork.CryptoCode && e.PaymentType == PaymentTypes.BTCLike) ??
-                                      enabledPaymentIds.FirstOrDefault(e => e.CryptoCode == _NetworkProvider.DefaultNetwork.CryptoCode && e.PaymentType == PaymentTypes.LightningLike) ??
+                                      enabledPaymentIds.FirstOrDefault(e => e.CryptoCode == _NetworkProvider.DefaultNetwork.CryptoCode && e.PaymentType != PaymentTypes.LNURLPay) ??
                                       enabledPaymentIds.FirstOrDefault();
                 }
                 isDefaultPaymentId = true;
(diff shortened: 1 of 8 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -703,9 +705,13 @@ public async Task<IActionResult> CheckoutNoScript(string? invoiceId, string? id
...
                 var paymentMethodTemp = invoice
                     .GetPaymentMethods()
-                    .FirstOrDefault(c => paymentMethodId.CryptoCode == c.GetId().CryptoCode);
+                    .FirstOrDefault(pm =>
+                    {
+                        var pmId = pm.GetId();
+                        return paymentMethodId.CryptoCode == pmId.CryptoCode && pmId != lnurlId;
+                    });
                 if (paymentMethodTemp == null)
-                    paymentMethodTemp = invoice.GetPaymentMethods().FirstOrDefault();
```
dennisreimann, line 708: This and the part above prevents LNURL from showing up as initial payment method. We don't want this, as LNURL also isn't explicitely selectable.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -768,6 +774,7 @@ public async Task<IActionResult> CheckoutNoScript(string? invoiceId, string? id
                 BrandColor = storeBlob.BrandColor,
                 CheckoutType = invoice.CheckoutType ?? storeBlob.CheckoutType,
                 HtmlTitle = storeBlob.HtmlTitle ?? "BTCPay Invoice",
+                OnChainWithLnInvoiceFallback = storeBlob.OnChainWithLnInvoiceFallback,
```
dennisreimann, line 777: We now need this to discern on the UI-level, whether or not to have LNURL available as payment method — which is the case only for top-up + non-unified.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -803,9 +810,6 @@ public async Task<IActionResult> CheckoutNoScript(string? invoiceId, string? id
                 IsMultiCurrency = invoice.GetPayments(false).Select(p => p.GetPaymentMethodId()).Concat(new[] { paymentMethod.GetId() }).Distinct().Count() > 1,
                 StoreId = store.Id,
                 AvailableCryptos = invoice.GetPaymentMethods()
-                                          .Where(i => i.Network != null && storeBlob.CheckoutType == CheckoutType.V1 ||
-                                              // Exclude LNURL for Checkout v2 + non-top up invoices
-                                              i.GetId() != lnurlId || invoice.IsUnsetTopUp())
```
dennisreimann, line 808: Instead of completely removing LNURL for non-topup invoices, this leaves LNURL in, but hides it from the payment method switches.

BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml
```diff
@@ -26,9 +25,7 @@ Vue.component("lnurl-withdraw-checkout", {
     },
     computed: {
         display: function () {
-            return (
-                this.model.paymentMethodId === 'BTC_LNURLPAY' || (
-                    this.model.paymentMethodId === 'BTC' && this.model.invoiceBitcoinUrl.match(/lightning=lnurl/i)));
+              return !!this.model.availableCryptos.find(pm => pm.paymentMethodId === 'BTC_LNURLPAY');
```
dennisreimann, line 28: Allows NFC whenever LNURL is available.

dennisreimann:
> I am for completely removing it where it makes no sense, graying it out, may just add confusion. Oh, could be that you're actually already doing that removing lnurl button when estimated that it's not functional or just per payment method?

@pavlenex Yes, it already isn't available in Checkout v2 anymore — we completely removed it for non-top-up invoices. And this PR reintroduced it kinda in "hidden" mode, so that it is available for NFC beyond just top-up invoices.

In that sense we kinda went to far by removing it completely, because without the changes here, LNURL (and hence NFC) wouldn't be available for regular invoices. I think the above screenshots explain it better — LNURL isn;t shown as a payment method other than for the top-up + non-BIP21 case.

pavlenex:
Got it, I have to test this and sync up, just to make sure I understood correctly, if the invoice is 10$, and lightning is available as a payment method, all other criterias (browser, etc) are met, the user would be able to checkout by tapping let's say boltcard? I just want to be completely sure I understand 

> we completely removed it for non-top-up invoices

NicolasDorier:
Code ACK for me, however, I am a bit overwhelmed by all of this, so I'll let  @pavlenex  judge :D

dennisreimann:
@NicolasDorier yeah, we'll need proper testing. I just turned this into a draft again, because I noticed that it still isn't fully correct and breaks again what I fixed in #4660. 

Will also add checks for the button in the Selenium tests.

dennisreimann:
> we'll need proper testing. I just turned this into a draft again, because I noticed that it still isn't fully correct and breaks again what I fixed in https://github.com/btcpayserver/btcpayserver/pull/4660.

Fixed that, added test cases and updated the screenshots above.

dennisreimann:
> just to make sure I understood correctly, if the invoice is 10$, and lightning is available as a payment method, all other criterias (browser, etc) are met, the user would be able to checkout by tapping let's say boltcard?

@pavlenex With the changes in this PR - yes, and that's the purpose of this pull request. Before it wouldn't show. 

Screenshots for your test case:

## This PR

pav-pr

## On current master

pav-master

ndeet:
Nice work! 

When I tested NFC with checkout v1 it was a bit annoying to always scroll down to have the NFC button show up. Maybe a bit OT and for a follow up:
I think it would be useful to have an option to show the NFC button more at the top. If someone runs a pos terminal and mostly has face2face customers, they either scan the QR or pay with NFC. The "pay in wallet" and "LN invoice" button/text is not really useful to them. So maybe a "pos mode" for checkout?

NicolasDorier:
@ndeet I think we should eventually have a Checkout "Retail mode" automatically stripping stuff. Like the `Pay in wallet`.

Maybe not for this PR, but I think checkout pages could get more optimized for retails.

dennisreimann:
@NicolasDorier Yes, I still have it on my list for v1.8.x — didn't get to it in time for this release.

Kukks (changes requested):
the NFC option is available regardless of LNURL. If there is a bolt11 or an lnurl on the screen, it should be shown.

dennisreimann:
@Kukks thanks for the pointer, I've fixed this in [3474fd5](https://github.com/btcpayserver/btcpayserver/pull/4671/commits/3474fd5cd80b8f63097a90d099af38ad9ac12f02) and also added a test for that case.

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
```diff
@@ -11,7 +11,10 @@
     ViewData["Title"] = Model.HtmlTitle;
 
     var hasPaymentPlugins = UiExtensions.Any(extension => extension.Location == "checkout-payment-method");
-    var paymentMethodCount = Model.AvailableCryptos.Count;
+    // Show LNURL as selectable payment method only for top up invoices + non-BIP21 case
+    var displayedPaymentMethods = Model.IsUnsetTopUp && !Model.OnChainWithLnInvoiceFallback
```
Kukks, line 15: We should really make this glue more flexible so thta we dint hardcode  these things in a view

Kukks (approved):

PR #4680, opened 2023-02-22
Greenfield: Admins can create/delete API keys of any user

### LINKED ISSUE #4673: [Bug]: Greenfield API: Create API Key allow to set/limit store scope

v1.7.12

docker

### What happened?

On the [create API key endpoint](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/ApiKeys_CreateApiKey) you can set the permissons but not limit the API key to one or more stores like you can do in the UI. 

Creating an API key with the Greenfield API

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

### LINKED ISSUE #4675: [Bug]: Greenfield API: allow to create API key for different user

v1.7.12

docker

### What happened?

When you set the API key with basic auth then you [generate that API key](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/ApiKeys_CreateApiKey) for the user. If you do that with an unrestricted API key then it would be great to create the keys for a specific user. E.g. your webapp has an unrestricted API key and creates users and separate API keys for them. At the moment the unrestricted API key will always create the new API key for that same user (admin) and therefore does not allow 3rd parties to generate users and api keys automatically. (only if they use basic out for each specific user which means they would need to know the password)

Tried to automate user creation and give them an API key they can use to control their BTCPay Server account. 

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

NicolasDorier:
Close #4675

Improved doc to close 
Close #4673


BTCPayServer/wwwroot/swagger/v1/swagger.template.api-keys.json
```diff
@@ -27,7 +27,47 @@
                 },
                 "security": [
                     {
-                        "API_Key": ["unrestricted"],
+                        "API_Key": [ "unrestricted" ],
```
ndeet, line 30: I see you introduced the `canmanageuser` permission. I wonder if it still needs to say `unrestricted` here or `canmanageuser` is enough now?

BTCPayServer.Client/Permissions.cs
```diff
@@ -206,8 +208,15 @@ private static bool ContainsPolicy(string policy, string subpolicy)
...
             PolicyHasChild(Policies.CanModifyStoreSettings,
-                Policies.CanManageCustodianAccounts, Policies.CanManagePullPayments, Policies.CanModifyInvoices, Policies.CanViewStoreSettings, Policies.CanModifyStoreWebhooks, Policies.CanModifyPaymentRequests );
-            
+                Policies.CanManageCustodianAccounts,
+                Policies.CanManagePullPayments,
+                Policies.CanModifyInvoices,
+                Policies.CanViewStoreSettings,
+                Policies.CanModifyStoreWebhooks,
+                Policies.CanModifyPaymentRequests,
+                Policies.CanManageUser);
```
Kukks, line 217: No, this is wrong. You must have meant to put it under CanmanagerServerSettings

BTCPayServer.Client/Permissions.cs
```diff
@@ -28,6 +28,7 @@ public class Policies
         public const string CanViewNotificationsForUser = "btcpay.user.canviewnotificationsforuser";
         public const string CanViewUsers = "btcpay.server.canviewusers";
         public const string CanCreateUser = "btcpay.server.cancreateuser";
+        public const string CanManageUser = "btcpay.server.canmanageuser";
```
Kukks, line 31:
```suggestion
        public const string CanManageUsers = "btcpay.server.canmanageusers";
```

Better plural so that it is clearer that it is a server policy, else it sounds like a current user edit permission

Kukks (changes requested):

BTCPayServer/wwwroot/swagger/v1/swagger.template.api-keys.json
ndeet: Got it, thanks for the explanation.

BTCPayServer.Client/Permissions.cs
NicolasDorier: I was wondering about this. Because just above it's singular.

NicolasDorier: Made it plural, even if it doesn't match above, it's more logical.

BTCPayServer.Client/Permissions.cs
```diff
@@ -28,6 +28,7 @@ public class Policies
         public const string CanViewNotificationsForUser = "btcpay.user.canviewnotificationsforuser";
         public const string CanViewUsers = "btcpay.server.canviewusers";
         public const string CanCreateUser = "btcpay.server.cancreateuser";
+        public const string CanManageUsers = "btcpay.server.canmanageuser";
```
Kukks, line 31:
```suggestion
        public const string CanManageUsers = "btcpay.server.canmanageusers";
```

PR #4681, opened 2023-02-22
Add payment info for invoice endpoints

bolatovumar:
https://github.com/btcpayserver/btcpayserver/discussions/4656


NicolasDorier:
The CSV export gives the payments, not the invoices.

I think payments shouldn't be exposed in this route, but rather through a separate endpoint such as 

`/api/v1/stores/{storeId}/invoices/{invoiceId}/payments`

As this endpoint would typically be used for accounting purpose.

NicolasDorier:
Actually, it should better be `/api/v1/stores/{storeId}/payments` with a filter on date.

bolatovumar:
@Atitlanio thoughts on the above ^^^^?

bolatovumar:
> The CSV export gives the payments, not the invoices.

This makes sense

> I think payments shouldn't be exposed in this route, but rather through a separate endpoint such as
> 
> `/api/v1/stores/{storeId}/invoices/{invoiceId}/payments`
> 
> As this endpoint would typically be used for accounting purpose.

`/api/v1/stores/{storeId}/invoices/{invoiceId}/payments` makes sense as we have an invoice resource which then has payments. 

@NicolasDorier Is there a specific reason we wouldn't want to include payments on `/api/v1/stores/{storeId}/invoices/{invoiceId}` though?

PR #4693, opened 2023-02-23
add invoicemetadata as a tab

Kukks:

### DIFF
--- a/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
@@ -126,7 +126,7 @@ namespace BTCPayServer.Models.InvoicingModels
         public string NotificationEmail { get; internal set; }
         public Dictionary<string, object> PosData { get; set; }
         public Dictionary<string, object> Metadata { get; set; }
-        public IDictionary<string, object> AdditionalData { get; set; }
+        public Dictionary<string, object> AdditionalData { get; set; }
         public List<PaymentEntity> Payments { get; set; }
         public bool Archived { get; set; }
         public bool CanRefund { get; set; }
--- a/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
+++ b/BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
@@ -125,6 +125,7 @@ namespace BTCPayServer.Models.InvoicingModels
         public List<Data.InvoiceEventData> Events { get; internal set; }
         public string NotificationEmail { get; internal set; }
         public Dictionary<string, object> PosData { get; set; }
+        public Dictionary<string, object> Metadata { get; set; }
         public List<PaymentEntity> Payments { get; set; }
         public bool Archived { get; set; }
         public bool CanRefund { get; set; }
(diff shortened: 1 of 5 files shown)


dennisreimann:
I think @dstrukt had a mockup already of how an updated invoice detail page could look like. We can discuss this in the design meeting.

dennisreimann:
Here's a [link to the mockup](https://www.figma.com/file/lE9dYRMqoABkJrn0VdXTE4/App?node-id=5788%3A114430&t=M5Qibjad0yFSFuJu-4), @dstrukt can confirm if this is the latest version.

tbh I don't like the tab approach that much, because that data should be present on th einitial view imho.

I think a quickwin optimization for that page would be to just have everything underneath each other. Right now the floating boxes are at random places depending on whether or not certain data is present. I can try to do a minimum effort cleanup version, if we think that's worth it.

pavlenex:
I don't think Kukks is opposed to a one view when we chatted in dm's about it, this was just a quick fix, but I agree with you @dennisreimann I mean if you can scope it and do a minimum cleanup we can try to get it into 1.8.0.

Kukks:
@dennisreimann  again the issue is data duplication in the view. The main view reads its info from metadata and shows it selectively. We would have to grab the metadata, carefully remove any items we already present and then show that, and at that point, it might be a malformed view (eg imagine asking for an additional buyer specific field, it would be split into 2 sections)

dstrukt:
ACK a single page view, agree that the information just needs to be consolidated and grouped better as opposed to being tabbed.

I had started work on this a while back, but we had other priorities at the time, but can work with @dennisreimann to get something suitable in the short term for 1.8.0.

dennisreimann:
Made it so that we exclude the well-known/typed keys from the additional information — this works quite nicely.

I still find that page a bit hard to parse, but that's something for a more involved cleanup. This quickwin version works good enough I think.

## Desktop

grafik

## Mobile

grafik

Kukks:
I am worried of existing integrations and their usage of the Metadata will
cause this page to get very messy

NicolasDorier:
@Kukks can you fix the test?

BTCPayServer/Views/Shared/PosData.cshtml
```diff
@@ -1,4 +1,4 @@
-@model (Dictionary<string, object> Items, int Level)
+@model (IDictionary<string, object> Items, int Level)
```
Kukks, line 1:
```suggestion
@model (Dictionary<string, object> Items, int Level)
```

BTCPayServer/Models/InvoicingModels/InvoiceDetailsModel.cs
```diff
@@ -125,6 +125,8 @@ public string Fiat
         public List<Data.InvoiceEventData> Events { get; internal set; }
         public string NotificationEmail { get; internal set; }
         public Dictionary<string, object> PosData { get; set; }
+        public Dictionary<string, object> Metadata { get; set; }
+        public IDictionary<string, object> AdditionalData { get; set; }
```
Kukks, line 129:
```suggestion
        public Dictionary<string, object> AdditionalData { get; set; }
```

dstrukt:
> I am worried of existing integrations and their usage of the Metadata will cause this page to get very messy

That's a really valid point, and it might be worth trying to run a few tests if possible to make sure it's not completely botched at this time.

Ultimately it might make sense to add a tab for the Integrations, but I think if we consolidate the data and structure the redesign well, it shouldn't be a problem .. will try to document some of the Integrations and make sure we're looking at all of the edge cases that could arise!

cc @dennisreimann

NicolasDorier:
Merging that, as I have another PR which will conflict a bit.

PR #4700, opened 2023-02-24
Update transaction label display

dennisreimann:
Right now we have a full-colored background for the transaction labels. Now that we introduce the label manager, there will likely be more labels. Once that's the case, the full-colored background gets noisy. Additionally, highlighting selected labels is difficult, because there's so much going on visually. A screenshot …

## Before

label-display-bg

## After

The updated state uses just an outline by default and highlights active/selected labels with the background. I also added the + and x signs as indicators that an action is going to add/remove a label. 

label-display

Also updates the coin selection to fix a visual bug that @Bangalisch cam across and mentioned it in the design meeting:

coin-selection

And one more bugfix, in which the label filter didn't get displayed anymore if there are no transactions with the selected label. This led to a situation in which one could not clear the label filter anymore.


pavlenex:
Concept ACK, I love how it looks.

dstrukt:
Will take a look here this evening -- looking good!

Kukks:
What about making use of the label manager component directly on each
transaction?

dennisreimann:
@Kukks yes, there are more potential optimizations, but I left the out of this PR on purpose to make it more likely to ship this with v1.8

Using AJAX for adding and removing labels would make for a much nicer UX, I agree.

NicolasDorier (approved):

pavlenex:
> Using AJAX for adding and removing labels would make for a much nicer UX, I agree.

Could someone log an issue for this one then? @dennisreimann @Kukks ? We can add it in one of the next releases.

This can be added for 1.8.0 but let's not extend the scope.

PR #4703, opened 2023-02-25
Remove JSON in strings from JObjects

NicolasDorier:
There are two places where we were embedding JSON as string properties.

One is `InvoiceMetadata.PosData`, and the other is `InvoiceBlob.DerivationStrategies`.

This code make sure that future invoices are using JObject rather than String of json object.
It also make sure that old format are properly parsed.

I needed to simplify the `PosParser` which was handling some corner case that shouldn't be supported anymore (for example, having a JArray in PosData). This only mean that non JObject PosData will not show up in the UI.


Webbdlee74 (commented):
It could be faster a little faster.

PR #4709, opened 2023-02-27
Add currency code to payment request table

bolatovumar:
https://github.com/btcpayserver/btcpayserver/discussions/4619

Some currencies have exactly the same formatting even though their values are different. Showing them with exactly the same formatting can be misleading even though it's technically correct for each of these currencies.

Not sure what the best way of dealing with this is but adding a column with the currency code seems like the most straightforward way to deal with this even though the currency code display can be redundant in many cases.

|Before|After|
|---|---|
|before|Capture|


NicolasDorier:
I changed your PR to just add (USD) suffix. That said, your way of doing a new column is more in line with what we did in pull payments page.
Maybe we should change it as well there later.

PR #4717, opened 2023-02-27
show all plugins regardless of version

Kukks:
We used to show all plugins available even if your btcpay did not support it, allowing users to discover plugins and also incentivizing them to update their BTCpay (they were not able to install unless criteria was met).


dennisreimann (approved):
tACK.

PR #4721, opened 2023-02-28
Unset X-Frame-Options header correctly

### LINKED ISSUE #4666: [Bug]: Cannot obtain user data on checkout

BTCPay Server v1.7.12+9344113ae

AWS EC2

### What happened?

I want to collect an email or shipping address in the point of sale

Screenshot 2023-02-19 at 5 39 15 PM

this is the error encountered on front end when data collection is set in point of sale

Screenshot 2023-02-19 at 5 41 14 PM

configure either collect email or shipping address options and get error on ui

### Relevant log output

```shell
2023-02-19 04:21:55.187 +00:00 [WRN] Error while contacting exchange coingecko: The operation was canceled.
2023-02-19 17:10:04.131 +00:00 [INF] BTC (Lightning): Start listening unix://etc/clightning_bitcoin/lightning-rpc
2023-02-19 17:40:48.303 +00:00 [INF] BTC (Lightning): No more invoice to listen on unix://etc/clightning_bitcoin/lightning-rpc, releasing the connection.
```

### What browser do you use?

chrome and safari

### Additional information

_No response_

dennisreimann:
According to the [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) there are onlye the `DENY` and `SAMEORIGIN` options, `ALLOW-FROM` being deprecated. Hence we have to actively unset the header, as we made `DENY` the default.

This also unsets the X-Frame-Options header for the public form pages, which is supposed to fix #4666. Still a draft, because I couldn't figure out yet, why the `X-Frame-Options` for those requests still get set to `SAMEORIGIN` if this is requested within an iframe.

### DIFF
--- a/BTCPayServer/Controllers/UIPaymentRequestController.cs
+++ b/BTCPayServer/Controllers/UIPaymentRequestController.cs
@@ -208,6 +208,7 @@ namespace BTCPayServer.Controllers
         [HttpGet("{payReqId}/form")]
         [HttpPost("{payReqId}/form")]
         [AllowAnonymous]
+        [XFrameOptions(XFrameOptionsAttribute.XFrameOptions.Unset)]
         public async Task<IActionResult> ViewPaymentRequestForm(string payReqId, FormViewModel viewModel)
         {
             var result = await _PaymentRequestRepository.FindPaymentRequest(payReqId, GetUserId());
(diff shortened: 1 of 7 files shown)


NicolasDorier:
The error of this guy says that the header is ignored, so setting it to `unset` shouldn't have any effect.
I don't think this is related

NicolasDorier:
testing this one.

NicolasDorier:
One thing was missing, it works now. Let's add this one for 1.8.

PR #4726, opened 2023-03-02
Support Form Select option

Kukks:


bolatovumar (approved):

NicolasDorier:
@Kukks it's impossible for people to discover this,

Can you add an example template like you did for address/email?

Kukks:
We need to build a UI for building the config, or maybe fleshing out a cheatsheet. I also need to improve some stuff because current the form doesn't support multiple values which kind of sucks for multiselect or other future items

NicolasDorier:
@Kukks no need of fancy UI, the default templates are good enough.
I just want this to be integrated to default template

image

Like here if there was another option which use the Select, it's good enough.

Issue #4364, opened 2022-11-27
[Checkout V2] Feedback & Iteration


dstrukt:
Will be updating this list, but will coordinate with @dennisreimann to make sure we're addressing everything.

- [x] Round all the `full-width` buttons
- [x] Add `back to store` button back to `Expired` view
- [x] Option: [Confetti](https://www.kirilv.com/canvas-confetti/#realistic) for successful payments - #4727 
- [x] Option: Display amount in sats when using BIP21 - #4730 

### NFC Support
- [x] https://github.com/btcpayserver/btcpayserver/issues/4443

### Copy Address CTA
- [x] https://github.com/btcpayserver/btcpayserver/issues/4442

### Custom CSS
- [x] Custom CSS option: #4459

### Unit Displays & Copy Functionality

- [x] Improve the BTC => USD conversion key/pair label `$16,543.45 (USD)` => `16,543.45 USD` #4736
- [x] Improve USD display `$8,000.00 (USD)` => `8,000.00 USD` #4736
- [ ] Should copying the fiat amount just copy the number `16,543.45`, imo, yes ... or everything, e.g. we currently copy this `$16,543.45 (USD)`  — can be done in  #4734 once #4736 is done

### Paid Invoice
- [x] Add copy icons to `Paid Invoice` view, and any other views #4734
- [x] Remove gap, if possible #4473
<img width="399" alt="Screen Shot 2022-12-05 at 4 24 55 PM" src="https://user-images.githubusercontent.com/6250771/205775761-04113418-29b4-4c7f-a563-3193153b0207.png">

### Copy Suggestions & Improvements

- [x] `Return to {Store} if you like to resubmit a payment.` => `Return to {Store} if you would like to resubmit a payment.`  #4734

### Translation

- [x] #4444
- [x] Add Language selector button #4723

PR #4730, opened 2023-03-03
Checkout v2: Option to display amount in Sats in BIP21 case

dennisreimann:
We have this option for Lightning invoices and it makes sense to reuse it in the BIP21 case as well. Addresses one of the points in #4364.

This PR also
- contains some refactoring to extract the part that does the model modifications
- mirrors the setting from the Lightning Settings on the Checkout Appearance page

grafik

grafik

### DIFF
--- a/BTCPayServer.Rating/Currencies.json
+++ b/BTCPayServer.Rating/Currencies.json
@@ -1305,7 +1305,7 @@
       "name":"Satoshis",
       "code":"SATS",
       "divisibility":0,
-      "symbol":"Sats",
+      "symbol":"sats",
       "crypto":true
    },
    {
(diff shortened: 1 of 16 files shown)


dstrukt:
General thought, I've always done `sats` in lowercase for mocks, etc..  but I see we do `Sats`, and have wondered is there a general preference across the space? Or even amongst the team?

Otherwise, checking out the branch now, will add review shortly!

dstrukt (approved):
Only have that remaining question above, otherwise functional tACK.

LGTM!

dennisreimann:
The [Bitcoin Design Guide](https://bitcoin.design/guide/designing-products/units-and-symbols/) is using the lowercase version (e.g. 615 sats) consistently.

dstrukt:
Good additional validation -- if no one is opposed, I'd vote we make it lowercase!

dennisreimann:
Good to go as discussed in the design meeting.

PR #4736, opened 2023-03-04
Update price display

dennisreimann:
As proposed by @dstrukt in #4364.

## Before 

before

## After

after


BTCPayServer.Rating/CurrencyNameTable.cs
```diff
@@ -122,10 +122,10 @@ public string DisplayFormatCurrency(decimal value, string currency)
...
             }
 
-            if (currencyData.Crypto)
-                return value.ToString("C", provider);
-            else
-                return value.ToString("C", provider) + $" ({currency})";
+            var formatted = value.ToString("C", provider)
+                .Replace(provider.CurrencySymbol, "").Trim();
+
+            return currencyData.Crypto ? formatted : $"{formatted} {currency}";
```
dennisreimann, line 128: @NicolasDorier What's the reason for distinguishing here? Can we make this consistent and either apply the currency code or not for all cases?

NicolasDorier:
I believe there are several things wrong in how we deal with formatting.

### Format depends on currency

Right now, we decide how to format the currency depending on the culture of the currency.
It shouldn't be like that. Instead, it should be based on the store's settings.

### Use of model to format currencies

Everywhere, we are relying on the controller + viewmodel to have a property such as FormattedAmount.
I believe this is wrong, how to format should be the concern of the view, we should instead use a tag helper.
Only exception is when we need the value JS side.

### XXX USD versus XXX $ versus XXX $ (USD)

We have at many place, different way of formatting.
#### XXX USD

This one is clear, but dry. Local people who aren't intertionally minded would prefer the use of symbol.
I think using this format in the backend everywhere is fine.

#### XXX $

This one isn't clear. What `$` refers to? But for local people the meaning is clear. In particular, in retail payments.
As such I believe that for checkouts, the merchant should be able to configure the format.
Once we have a "Retail mode", we could set that automatically.

#### XXX $ (USD)

Ugly alternative I ended up with to avoid having debates about which form to use.

NicolasDorier:
About this PR, I am not opposed to convert all `XXX $ (USD)` to `XXX USD`.

But I think a second step would be to optimize the checkout page for retail mode, where we use symbol instead.

Btw, for BTC, I believe it should always appear in checkout as the symbol.

dennisreimann:
@NicolasDorier In [7c34599](https://github.com/btcpayserver/btcpayserver/pull/4736/commits/7c345996e24bcfce15e8e859ab3000f1968a39be) I've added a `DisplayFormatter`, which can produce the different formats like this:

grafik

A tag helper might not work, as we rather need a function, because we use the formatted strings as text content as well as for HTML attributes.

dstrukt:
Some valid considerations @NicolasDorier appreciate you breaking it down, tend to lean towards symbol `$ XXX` or `XXX USD`.

But like the flexibility that this enables @dennisreimann!

NicolasDorier:
@dennisreimann I don't understand why the `DisplayFormatter`, I think you can just get rid of `DisplayFormatCurrency` and `FormatCurrency` completely and replace it by what you have put in the `DisplayFormatter`.

dennisreimann:
Alright, now the `DisplayFormatCurrency` method got replaced by the `DisplayFormatter` and I've applied it in several places. There might be more places to tackle, but I wanted to get the alignment for the changes so far before I clean it up fully:

- Backend (merchant): We display values with the currency code (e.g. `100.00 USD`) across the backend.
- Frontend (customer): We display values with the currency symbol (e.g. `$100.00`) on all user/customer-facing views.

### Example from the payment request view

grafik

Ironically, I had to add some `…Formatted` properties for the invoice, because those values are re-used in JavaScript as well. We also have that for the payment request (and potentially other views that use JS, like the apps). We could clean up all of that and maybe implement a JS-equivalent of the `DisplayFormatter`, but let's first align on what we have so far.

dennisreimann:
I can fix the tests and continue once #4730 is merged, because there'll be some overlap there which I'd like to avoid.

dennisreimann:
We discussed this on the design meeting and think it'd be best to use the currency symbol for fiat only – meaning we'd switch back the latest changes in which I used the symbol for Bitcoin also. 

Reasoning:
- The BTC unicode symbol does not really look good
- Only BTC has a symbol, we don't even have one for sats
- Backwards-compatibility and visual consistency

TL;DR
- crypto: Always currency code for 
- fiat: Currency code in backend, symbol on customer-facing pages

@NicolasDorier Ok like that?

dennisreimann:
Updated it to what was discussed in the design meeting. Also fixed that we were somehow showing fiat order amount and exchange rate for invoices denominated in crypto.

### BTC denomination

btc

### Sats denomination

sats

### USD denomination

usd

USD_SATS

### EUR denomination

EUR_BTC

EUR_SATS

dstrukt:
ACK to the above, it looks great!

dstrukt (approved):

NicolasDorier:
@dennisreimann we can't do a JS equivalent of what we do in C#, the rules for currency formatting are complex, and only present in server side.

What should be done for Views is that the JS only display a `amountFormatted` text field that already got formatted by the backend.

BTCPayServer/Views/UIPullPayment/ViewPullPayment.cshtml
```diff
@@ -141,18 +143,18 @@
...
                                 </div>
                                 <div class="d-flex d-print-inline-block flex-column mb-4 me-5 d-sm-inline-flex mb-sm-0">
-                                    <dt class="h4 fw-semibold text-nowrap order-2 order-sm-1 mb-1">@Model.AmountCollectedFormatted</dt>
+                                    <dt class="h4 fw-semibold text-nowrap order-2 order-sm-1 mb-1">@DisplayFormatter.Currency(Model.AmountCollected, Model.Currency)</dt>
                                     <dd class="order-1 order-sm-2 mb-1">Already claimed</dd>
                                 </div>
                                 <div class="d-flex d-print-inline-block flex-column mb-0 d-sm-inline-flex float-sm-end">
-                                    <dt class="h4 text-sm-end fw-semibold text-nowrap order-2 order-sm-1 mb-1">@Model.AmountFormatted</dt>
+                                    <dt class="h4 text-sm-end fw-semibold text-nowrap order-2 order-sm-1 mb-1">@DisplayFormatter.Currency(Model.Amount, Model.Currency)</dt>
```
NicolasDorier, line 157: I would prefer you keep the `@Model.AmountFormatted` approach.

The reason is that I'd like to allow the user to chose whether to show public facing pages with Currency or symbol.

In retails mode, I believe this should show symbol.

dennisreimann: Should we really make this an additional option? It fells like there is already so much to configure, which leads to many scenarios as the screenshots above show. This would introduce one more setting which and to me it seems we should just make a decision and not bother the user with.

Mentioning @dstrukt and @pavlenex here, because we talked about these things yesterday.

dennisreimann:
@NicolasDorier This is good to go from my side and if you want to make this configurable, we can do so in #4756 or another separate PR. Imho this one contains enough already and it solves what we aligned on in the design meeting.

In [7874d77](https://github.com/btcpayserver/btcpayserver/pull/4736/commits/7874d77cb35633a0cd6a5a9bc422c0d77aa5fe5e) I added a bunch of test cases and found one more thing, which also wasn't correct in the screenshots above: The exchange rate still contained the currency code instead of the symbol. Also fixed now.

PR #4746, opened 2023-03-07
Store branding: Add complementing text and accent colors

dennisreimann:
Currently we use the current theme text color, independent of the brand color: This is problematic with rather light or dark branding colors, because we might show light text on a light background. This fixes that circumstance by calculating the complementing text color and also adds a slight adjustment for the hover/focus states. The accent color is calculated by decreasing the brightness a bit, which works for light and dark brand colors.

### Before

grafik

### After

after

### DIFF
--- a/BTCPayServer/ColorPalette.cs
+++ b/BTCPayServer/ColorPalette.cs
@@ -59,5 +59,44 @@ namespace BTCPayServer
                     return Labels[num % Labels.Length];
             }
         }
+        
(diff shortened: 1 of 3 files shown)


NicolasDorier:
it should probably be rather a non static method inside the color palette, but that's not really important. Good for me.

dstrukt (approved):
ACK, LGTM!

Love the update

PR #4750, opened 2023-03-08
Greenfield: Allow Bearer token in Authorization header

dennisreimann:
WIP, needs testing and docs.

Implements the [LNDhub-compatible API](https://github.com/BlueWallet/LndHub/blob/master/doc/Send-requirements.md#get-getbtc) for stores via our Greenfield API, so that one can connect BlueWallet, Zeus and Alby. The URL for that would be `lndhub://{userId}:{apiToken}@mybtc.pay/api/v1/stores/{storeId}/lndhub/BTC`.

Adapted this based on the [LNbank LNDhub controller](https://github.com/dennisreimann/btcpayserver-plugin-lnbank/blob/master/BTCPayServer.Plugins.LNbank/Controllers/API/LndhubController.cs), because @Kukks had the good idea that this would be useful to have for our stores as well. See dennisreimann/btcpayserver-plugin-lnbank#25


BTCPayServer/Security/GreenField/APIKeyExtensions.cs
```diff
@@ -15,11 +15,19 @@ public static class APIKeyExtensions
...
+                if (auth.StartsWith("token ", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    apiKey = auth.Substring("token ".Length);
+                    return true;
+                }
+                if (auth.StartsWith("Bearer ", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    apiKey = auth.Substring("Bearer ".Length);
+                    return true;
+                }
```
dennisreimann, line 30: As BlueWallet (and compatible wallets) assume this to be an OAuth token, they pass it around with the `Bearer` prefix. For us it's the general API token, which we just give back from the `/auth` endpoint, because we don't do OAuth.

BTCPayServer/Controllers/GreenField/GreenfieldLndHubApiController.Store.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        switch (type)
+        {
+            case "auth":
+                // login = user id, password = api key
+                var apiKey = await _apiKeyRepository.GetKey(req.Password, true);
+                if (apiKey != null && apiKey.UserId == req.Login && !await _userManager.IsLockedOutAsync(apiKey.User))
+                {
+                    result = new AuthResponse { AccessToken = apiKey.Id, RefreshToken = apiKey.Id };
+                }
```
dennisreimann, line 73: Since we don't do OAuth, just pass back the API token we got for authentication purposes. We get this token for following requests as the `Bearer` token in the `Authentication` header.

BTCPayServer/Controllers/GreenField/GreenfieldLndHubApiController.Store.cs
```diff
@@ -0,0 +1,407 @@
...
+                break;
+
+            // fake this case as we don't do OAuth
+            case "refresh_token":
+                result = new AuthResponse { AccessToken = req.RefreshToken, RefreshToken = req.RefreshToken };
```
dennisreimann, line 78: `req.RefreshToken` is also just the users regular API token.

NicolasDorier:
this should be a plugin

dennisreimann:
Fair point, I think this is also what @kukks initially proposed.

dennisreimann:
I removed the controller and will extract it into a separate plugin. However, can we get the remaining change in the `APIKeyExtensions` in, so that it will work out of the box?

dennisreimann:
Any objections, @Kukks?

Kukks:
I don't know, I don't feel very comfortable doing a hack on this layer for
the sake of plugins.

Lets figure out an alternative for the plugin to get authorized.

Personally I would not bother with creating an actual api key to use the
plugin, I would maybe
* add a toggle for enabling lndhub on a store
* when toggled a secret passphrase is generated
* the passphrase enables the lndhub controller for the store

dennisreimann:
Alright!

Issue #4753, opened 2023-03-09
[Feature Request]: Merchant scans QR rather than Customer scans QR


NicolasDorier:
The generic idea of LN Url Withdraw is adapted for retail use: Rather than having the customer scanning a QR, it's the merchant which "pull" the payment from the customer.

The advantage of it is:

* Customer doesn't need a camera
* Customer doesn't need internet

This UX is popular in Japan with payment services such as PayPay.
Because the camera is controlled by the merchant, it is also more reliable.

We allow this throught two ways currently:

* NFC support, 
* Akward LN Withdraw button in the page. (not adapted for retail)

![image](https://user-images.githubusercontent.com/3020646/223886924-bd65ac2b-7fd6-49e7-b2b3-35a1653c49de.png)

I propose that we remove the LN Withdraw button, and instead we replace with a QR code scanner.
Somebody having a wallet supporting LN Withdraw can then just present the QR code so the BTCPay can pull the payment out.

pavlenex:
Which wallets would support this?

Kukks:
> Which wallets would support this?
![](https://i.imgur.com/yCtCFaQ.png)

siwatanejo:
But I guess that after the merchant scans, the user would still need to swipe or verify the payment? Otherwise, merchant could technically swipe any amount it wants.

NicolasDorier:
not different from the NFC cards.

dstrukt:
Do we still want to explore this @pavlenex?

PR #4756, opened 2023-03-09
Add presets in the checkout appearance

### LINKED ISSUE #4712: `Retail checkout mode`
Suggested by @NicolasDorier, we should have a way for retailers using btcpay for in-person payments can abstract all the unnecessary options such as `address copy`, `open in a wallet`,

NicolasDorier:
This PR add three checkbox in the checkout appearance:
* Show store header
* Show Pay in wallet button

On top of this, it allows the merchant to create a new store optimized for in-store purchases.

Currently, this preset will create a store on V2 and hidding store header, and Pay in wallet button from the checkout page.

The reason why we need to add this settings, is that most merchant will not take the time to dive into the hundreds of checkboxes available for customizing their checkout in a fine grained way.
It's better to just have simple presets at the store's creation.

Checkout appearance:

image

Ecommerce: (By default, e-commerce still use checkout v1, contrary to physical preset)
image

Physical:
image

Closes #4712


dennisreimann:
We need a proper ecommerce icon, but I think something like this works better for the selector component:

grafik

NicolasDorier:
I believe another thing in retail: "View details" should be expanded by default.

dennisreimann:
Summary for @dstrukt and @pavlenex, because we discussed this yesterday ….

This introduces three new settings:

- `ShowPayInWalletButton` 
- `ShowStoreHeader`
- `ShowLNWithdrawButton`

Do they have to be separate settings or would a single one combining them suffice? I think so ­— if not, what would be the case to set them individually? 

Our main goal here should be to make it more intuitive for the user and having these three settings imho seems like we are introducing more ambiguity. And that seems counter-intuitive to the main reasoning behind this PR: Letting the user have a simple choice what kind of experience they'd like to go with.

As a sidenote, which we've also discussed yesterday: I think having that decision on the Create Store view might also confuse the user, because it feels like a choice with deep impact — but the outcome is rather simple, as there aren't that many things that change. Tbh I'm still kinda torn on how to best solve this.

NicolasDorier:
I am also unsure about the best UX for that, I just know this is a step in right direction though

BTCPayServer/Views/Shared/_StoreHeader.cshtml
```diff
@@ -7,10 +7,10 @@
...
 }
-<header class="store-header" v-pre>
+<header v-if="srvModel.showStoreHeader" class="store-header">
     @if (!string.IsNullOrEmpty(logoUrl))
     {
-        <img src="@logoUrl" alt="@Model.Title" class="store-logo"/>
+        <img v-pre src="@logoUrl" alt="@Model.Title" class="store-logo"/>
     }
-    <h1 class="store-name">@Model.Title</h1>
+    <h1 v-pre class="store-name">@Model.Title</h1>
```
dennisreimann, line 15: This partial is used in a lot of places and not always within the context of a Vue app. Imho the proper approach would be to do this check on the view that references the partial.

BTCPayServer/Views/UIUserStores/CreateStore.cshtml
```diff
@@ -41,6 +41,25 @@
                 <div class="form-text mt-2 only-for-js">The recommended price source gets chosen based on the default currency.</div>
                 <span asp-validation-for="PreferredExchange" class="text-danger"></span>
             </div>
+
+            <div class="form-group">
+                <label asp-for="Preset" class="form-label"></label>
+                <div class="btcpay-list-select">
+                    <input asp-for="Preset" id="Preset_ECommerce" value="ECommerce" type="radio">
+                    <label for="Preset_ECommerce" class="btcpay-list-select-item">
+                        <vc:icon symbol="pos-cart" />
```
dennisreimann, line 50: @dstrukt Can you prepare a proper eCommerce icon?

BTCPayServer/Data/StoreBlob.cs
```diff
@@ -218,6 +218,15 @@ public RateRules GetDefaultRateRules(BTCPayNetworkProvider networkProvider)
...
         public string CssFileId { get; set; }
+        [DefaultValue(true)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        public bool ShowPayInWalletButton { get; set; } = true;
+        [DefaultValue(true)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        public bool ShowStoreHeader { get; set; } = true;
+        [DefaultValue(true)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        public bool ShowLNWithdrawButton { get; set; } = true;
```
dennisreimann, line 229: We should rename this to e.g. `ShowPayWithNfcButton`, because that's the primary use case of the button.

BTCPayServer/Models/StoreViewModels/CheckoutAppearanceViewModel.cs
```diff
@@ -26,6 +26,13 @@ public void SetLanguages(LanguageService langService, string defaultLang)
         [Display(Name = "Unify on-chain and lightning payment URL/QR code")]
         public bool OnChainWithLnInvoiceFallback { get; set; }
 
+        [Display(Name = "Show \"Pay in wallet\" button")]
+        public bool ShowPayInWalletButton { get; set; }
+        [Display(Name = "Show the store header")]
+        public bool ShowStoreHeader { get; set; }
+        [Display(Name = "Show \"LN Withdraw\" button")]
```
dennisreimann, line 33:
```suggestion
        [Display(Name = "Show \"Pay by NFC\" button")]
```

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -86,13 +86,25 @@
...
             <div class="form-check">
                 <input asp-for="OnChainWithLnInvoiceFallback" type="checkbox" class="form-check-input" />
                 <label asp-for="OnChainWithLnInvoiceFallback" class="form-check-label"></label>
                 <a href="https://bitcoinqr.dev/" target="_blank" rel="noreferrer noopener">
                     <vc:icon symbol="info" />
                 </a>
+            </div>
+			<div class="form-check">
+                <input asp-for="ShowPayInWalletButton" type="checkbox" class="form-check-input" />
+				<label asp-for="ShowPayInWalletButton" class="form-check-label"></label>
```
dennisreimann, line 107: We should show all of them together, even though the store header setting might not aply to v1.

dennisreimann:
One more reason to unify the whole concept into a single setting is that we'd otherwise also need one more option for hiding the copy buttons of the addresses.

BTCPayServer/Views/UIUserStores/CreateStore.cshtml
dstrukt: Yeah I'll add it to the list!

BTCPayServer/Data/StoreBlob.cs
NicolasDorier: I don't understand how this button is related to NFC. When you click on it it show a alert box to copy paste something. How is it related to NFC?

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
NicolasDorier: I've just put them next to each other, while still putting the two button specific to v2 inside the expander.

BTCPayServer/Models/StoreViewModels/CheckoutAppearanceViewModel.cs
NicolasDorier: There is no "Pay by NFC" button

BTCPayServer/Data/StoreBlob.cs
dennisreimann: What you see on the desktop is the fallback functionality that @Kukks added for the case that NFC isn't available. At least in retail mode this will most likely show up as the "Pay by NFC" button.

NicolasDorier:
@dennisreimann can you elaborate on the LN Withdraw button? because you suggest to rename it to NFC, while this button isn't called "NFC button", nor doing it does any NFC related things, it just open a popup to copy/paste a link.

NicolasDorier:
> One more reason to unify the whole concept into a single setting is that we'd otherwise also need one more option for hiding the copy buttons of the addresses.

Actually, we should probably remove completely the text address. If it's for retail, it should be QR code only.

dennisreimann:
The only sense showing the address  would have is for the customer to verify that the address they see in their wallet is the one from the QR.

dennisreimann:
> @dennisreimann can you elaborate on the LN Withdraw button? because you suggest to rename it to NFC, while this button isn't called "NFC button", nor doing it does any NFC related things, it just open a popup to copy/paste a link.

Ye,s the code is [here](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml#L8) and there are two scenarios: The one you are seeing on desktop, where NFC isn't `supported` — the [`btnText`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml#L53) then says "Pay by LNURL-Withdraw" and offers the fallback with the prompt you mentiond (and which we are going to replace in #4753).

The other case is on devices where NFC is `supported` — the button will say "Pay by NFC (LNURL-Withdraw)" then and trigger the NFC functionality, which is the original purpose of that button. @kukks just added the previously mentioned prompt as a fallback.

NicolasDorier:
@dennisreimann I don't think the NFC button should ever be hidden. Only the fallback should.
Also `Pay by NFC (LNURL-Withdraw)` should be only `Pay by NFC` imho.

NicolasDorier:
Ping @Kukks why did you add the fallback "Pay by LNURL withdraw"? (not NFC)

I think we should hide it by default, for both ecommerce and retail case.

Kukks:
> Ping @Kukks why did you add the fallback "Pay by LNURL withdraw"? (not NFC)
> 
> I think we should hide it by default, for both ecommerce and retail case.

Because it's still usable, although the UI could use some work.

dennisreimann:
Isn't the plan anyways to replace the prompt fallback with the merchant scan functionality in #4753?

NicolasDorier:
@Kukks usable by who? which wallet benefit from it?

NicolasDorier:
@dstrukt what do you think, should the Bitcoin Address be removed in retail mode?

NicolasDorier:
@dennisreimann I talked with @Kukks `Show "LN Withdraw" button` is the right term.
This isn't about NFC, it's about LN Withdraw (Copy/paste UX).
While NFC should be available on retail, the LN Withdraw button should not.

dennisreimann:
I'm OK with that. My main point was to give people a clearer way to communicate what that setting does and I had the impression that NFC might ring a bell for them, whereas LNURL-Withdraw might not.

NicolasDorier:
@dennisreimann yes, it's because those are two different things. (even if they use same protocol, it's not something users should be concerned about)

NicolasDorier:
I think for e-commerce a globe was a good idea.
Should we wait @dstrukt icon in the svg, or just reuse the font awesome globe?

NicolasDorier:
I think the two modes should be called `In-store checkout` versus `Web checkout`.
I will change it.

NicolasDorier:
I improved the text:

image

NicolasDorier:
Anything left here?

dennisreimann:
@NicolasDorier we'll discuss it in todays design call. @pavlenex @dstrukt

NicolasDorier:
I am open to negotiate: For example we can probably remove the exchange rate source dropdown, and peek the recommended one by default instead.

I think this option is more important.

pavlenex:
I still am not sold on this. If the only feature we're optimizing for is removing stuff from the checkout page, then this should be in the checkout experience (in-store checkout and web-checkout). There's still unclarity for people that perhaps have both? Or people that don't have stores? The proposed screen optimizes  for two use-cases and ignores all the rest.

NicolasDorier:
>  then this should be in the checkout experience

It is also present in the checkout experience, if you want to customize further.

> The proposed screen optimizes for two use-cases and ignores all the rest.

There isn't any "all the rest". There is only 2 use case to consider:
1. the checkout page will be shown on the customer's device.
2. the checkout page will be shown on the merchant's device. In which case there is a bunch of button that can't be used, clutter the UX, and worse: Can be accidently pushed.

There isn't any other choice, because it's binary and clearcut.

pavlenex:
> There isn't any other choice, because it's binary and clearcut.

No it's not. There's a ton of other use-case besides online vs retail merchant. People use BTCPay as their own wallet or a node, in this scenario what would they select? And what happens if merchant has both online and retail store, what would they select, the UI is ambitious.


> In this case they don't have to care and just push Next.

Right, but UI doesn't  convey this, there's no skip option in the screenshot you posted?

I really don't think this is a good idea, I'd have to concept NACK this. I just don't see a reason to introduce another step in the store creating, to cater for only a **single** improvement (web vs physical checkout), if someone can post potentially more options we could bundle, and optimize for this, then I'll happily be proven wrong, but this particular PR imo introduces potentially more confusion, video/docs editing, and all of that for the sake of catering a single checkout experience checkbox. I just don't see how intorducing an entirely new wizard beats the checkbox, unless it bundles more than one options and reduces the step, there's no UX improvement from checking a box vs selecting the wizard option, it's adds no value unless more options are bundled.

dennisreimann:
@NicolasDorier We discussed this in yesterday's design meeting and everyone felt like that having this choice on the Create Store view might do more harm than good. 

As the options only concern the checkout view, we think a better fit would be to have that choice on the Checkout Appearance config view. And maybe that would also solve another issue I'd see with introducing three more options: Can't we just unify those into a single choice (we might keep them separate in the code, but abstract the three into a single setting for the user, which sets them simultaneously).

TL;DR Imho keeping the Create Store view as it is and moving that switch to Checkout Appearance would make the most sense. 

grafik

dstrukt:
Sorry for the delay in comment, been thinking about this + discussion in today's design call.

Overall, love the thinking behind the idea, and agree we should bundle a configuration for specific use-cases, but this particular implementation, NACK.

For a few reasons:
- I think this is a feature that would be best suited solely within the Point of Sale (plugin) setup flow or possibly a Checkout configuration wizard (or something along these things)
- As @dennisreimann point's out, it introduces another option, and more confusing choice imo, at the Store level when the user may want to do something else, other than the 2 use cases listed .. or even be using in the ways listed at all
- Users will have access to plugins which open an even wider array of choice, and don't think we should push the user to make this choice at the Store level, as this may never be relevant to them if they say, just use the Pay Button plugin

Have more thoughts, but wanted to get this out there to push the discussion!

NicolasDorier:
Should I remove the preset for this PR and keep the three textboxes? At least it will remove by default the horrible "Pay with LNURL WITHDRAW" button when there is no NFC.

NicolasDorier:
> No it's not. There's a ton of other use-case besides online vs retail merchant. 

Once again, there is no other use cases, this isn't about retail vs online, it's about whether the checkout page show up on the customer's device or on the merchant's device.

dennisreimann:
@NicolasDorier imho moving what's on the Create Store to the Checkout Appearance would be a good solution. 

By using `Preset` instead of the three checkboxes we also simplify it for the user and make the whole concept more understandable. See the screenshot above.

dennisreimann:
As discussed with Nicolas, I moved the switch to the Checkout Appearance section.

grafik

One remaining question is whether or not we want to display the address/destination for the `Physical` option.

NicolasDorier:
@dennisreimann in-store should use checkout v2.

NicolasDorier:
I changed it to be a dropdown rather than radiobutton.

Radio button was not adapted as the preset isn't a stored setting. It also should set the checkout in V2.

image

My alternative use Javascript to manually set the settings appropriate for the preset.

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -19,6 +18,18 @@
...
                 : ''
         });
+		delegate('click', '#Presets_InStore', e => {
+			$("#UseNewCheckout").prop('checked', true);
+			$("#NewCheckoutSettings").addClass('show');
+			$("#ShowPayInWalletButton").prop('checked', false);
+			$("#ShowStoreHeader").prop('checked', false);
+		});
+		delegate('click', '#Presets_Online', e => {
+			$("#UseNewCheckout").prop('checked', false);
```
dennisreimann, line 28:
```suggestion
```

I don't think this should set it to V1 — it should not touch that option in that case.

BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -19,6 +18,18 @@
...
+			$("#NewCheckoutSettings").removeClass('show');
```
dennisreimann, line 29:
```suggestion
```

This should also go then.

NicolasDorier:
@dennisreimann merging this for now.
I think we should make v2 the default in a different PR and all at once. Maybe for 1.9.

PR #4764, opened 2023-03-11
Replace text in copy buttons with icon

### LINKED ISSUE #4699: Replace `Copy` text with an icon across all views
UPDATE: We're widening the scope of the issue, and instead of just the checkout page, we should add copy icons across all views and unify this.

Having a text means we have to translate it. In certain languages the text may get lengthy and the experience may seem odd. A simple copy icon is something everyone understands these days, so I suggest we use copy icon instead of the copy text.

Screenshot 2023-02-24 at 10 19 10

dennisreimann:
Closes #4699. I think I covered all ocurrences. We have two pattern for (indicating) copying:

- Button with icon
- Hover icon on text

Both using the icon, which gets replaced for 2500ms with the checkmark icon when the button/text is clicked.

copy

These are the occurances besides Checkout v2….

## Wallet Settings 
wallet-settings

## Wallet Receive 
wallet-receive

## Show QR Modal
showqr

## Public Lightning Info
lninfo

## Lightning Addresses
lnaddress

## Pay Button
paybutton

## 2FA: Enable Authenticator
2fa

# API Keys list
apikeys

## Footer
footer

## Custodians

Two occurances on the View Custodian Account template, which I couldn't reproduce, but it uses the same code as everywhere else, so it should be fine.


dennisreimann:
@dstrukt @pavlenex This one is ready for review now :)

NicolasDorier:
Can you keep the translations as before? I understand we don't use it, but we might need later, if we remove it from translation, it will remove it from transifex as well.

NicolasDorier:
It also seems you changed this icon not only on v2? maybe you should tell us all the places you changed so people can test.

dennisreimann:
Alright, these are the occurances besides Checkout v2….

## Wallet Settings 
wallet-settings

## Wallet Receive 
wallet-receive

## Show QR Modal
showqr

## Public Lightning Info
lninfo

## Lightning Addresses
lnaddress

## Pay Button
paybutton

## 2FA: Enable Authenticator
2fa

# API Keys list
apikeys

## Footer
footer

## Custodians

Two occurances on the View Custodian Account template, which I couldn't reproduce, but it uses the same code as everywhere else, so it should be fine.

NicolasDorier:
ACK for me

dstrukt (approved):
Sorry for the delay. 

tACK!

I think you got everywhere, but will keep looking, great job!

PR #4775, opened 2023-03-15
Dashboard: Fix app stats tiles

dennisreimann:
They broke with #4747, because they contain script blocks that are loaded asynchronuosly and need to get run once the chart data is added to the page.

### DIFF
--- a/BTCPayServer/Components/AppSales/AppSales.cs
+++ b/BTCPayServer/Components/AppSales/AppSales.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Security.AccessControl;
 using System.Threading.Tasks;
 using BTCPayServer.Data;
 using BTCPayServer.Models.AppViewModels;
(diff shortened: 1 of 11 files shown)


NicolasDorier:
I really want to avoid unsafe-eval. 
There is a lot's of way to shoot ourselves in the foot, and I have trouble to catch them all.
While we can't stop relying on vue.js, if we have easy ways to avoid using `eval` we should do this.

I refactored the code so that rather than generating a different method for each app tile tied together with C# razor, we instead generate the same method everytimes, and pass a json object to it.

As side effect, it makes the component more easily usable elsewhere, as it only need the `app-id` and the `app-type`.

Note that we should be careful in exposing too much data to the viewmodel, (before we were exposing the whole `AppData` tot he view model), because if we serialize it to json by mistake, we end up leaking all the object graph... which can contains the store lighting creds and xpub.

dennisreimann (commented):
tACK. The changes here will most likly conflict with #4608, but I can clean that up afterwards. Let's get this in and include it in the next bugfixes release.

PR #4778, opened 2023-03-15
Checkout v2: Payment processing state

### LINKED ISSUE #4742: Have `processing` checkout screen / state
When a user makes an on-chain payment, we show that the invoice is paid (settled state), when in fact, it's `processing`. This may lead to confusion and perhaps we should have a checkout scree/message that can communicate this more accurately.

This issue has extensively been discussed in #1319 but it got sidetracked with multiple topics (RBF for example).

The scope of this issue is straightforward, let's design a checkout screen, that will communicate that `Payment is processing and is waiting for confirmation on blockchain`. 

> Your payment has been received and is currently processing. Once it confirms on the blockchain, the status will be updated to settled.

dennisreimann:
Adds the processing state for onchain payments, in which the progress gets displayed while waiting for the required number of confirmations.

Also contains some improvements for the cheat mode, which I came across.

Closes #4742. 

## Notes on the implementation

The properties `RequiredConfirmations` and `ReceivedConfirmations` got added to the `PaymentModel`. Ideally one would populate them in the `BitcoinLikePaymentHandler.PreparePaymentModel` method, but that doesn't have access to the `InvoiceEntity` and I didn't want to touch those parts.

I also didn't want to touch the `InvoiceWatcher`, which does not report received confirmations, unless they chnage the state of the invoice. That's why I opted to simply poll for updates with a 30 second interval, once the payment enters the processing state.

Maybe @NicolasDorier has ideas about to solve those things better and then I can refactor it.

## Screenshot

One nitpicky thing: For the text we cannot do proper pluralization of "confirmations" (because the localization lib doesn't support that), maybe someone is creative and finds a way how we can circumvent it with other wording.

paid

## Recording

For this I lowered the polling interval to 5 seconds, otherwise this recording would be long and boring ;)

processing


dennisreimann:
@NicolasDorier can you check my question about the implementation above? Let me know if there's a better way to achieve this – otherwise this is good to go and can be merged if you think this os OK.

Kukks:
Hey there is a property `ParentEntity` which gives you access. I'll take a look and try to reduce the footprint on the payment model, we should strive to make it smaller not bigger

BTCPayServer/wwwroot/locales/checkout/en.json
```diff
@@ -26,6 +26,9 @@
   "address": "Address",
   "lightning": "Lightning",
   "payment_link": "Payment Link",
+  "payment_sent": "Payment Sent",
```
NicolasDorier, line 29: uh, this should really be `Payment sent`, but more `Payment received` no?

NicolasDorier: Same point on all the `*_sent_*` stuff below

NicolasDorier:
I'm fine for the current solution. Polling is good enough.

Except if @Kukks has a good idea to improve something, this can be merged.

NicolasDorier (approved):

BTCPayServer/wwwroot/locales/checkout/en.json
dennisreimann: How shall we phrase it, @dstrukt @pavlenex?

Kukks (approved):
gave up refactorting, we will have to do it later as we open up payment methods to plugins.

PR #4781, opened 2023-03-17
Derivation scheme parsing incorporates fingerprint and key path

### LINKED ISSUE #4777: [Bug]: Improve Support for Seedsigner Airgapped HW Wallet

v1.8.4+fadb05e97

Docker

### What happened?

When setting up my airgapped wallet, I was told to use the Export XPUB > BlueWallet option.

However this lead to the  Master fingerprint  and  Account key path not being populated. And as a result I was unable to sign the PSBTs from BTCPay with the device.

I've added the two missing fields from Specter where I also have the HW Wallet connected to and was able to do the signing in BTCPay. However when I wanted to broadcast the transaction, it kept on sending me back to this page:

grafik

and then:

grafik

(beginn at screenshot 1 after clicking "DECODE PSBT"

It doesn't make any difference if I tick "Don't create UTXO change".

I also noticed a UI problem:

When on the "Send BTC" Dialog, clicking "Sign transaction", then "Partially Signed Bitcoin Transaction" and then "Sign transaction", I end back on the "choose your signing method" dialog. (I know I should have chosen "Export PSBT for signing" here.

Also we are lacking documentation for this hardware wallet currently. Kind of sad, cause it's such an awesome little device. :-)

see above.

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox

### Additional information

Seedsigner version 0.6.0

dennisreimann:
Fixes #4777.

### DIFF
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -709,19 +709,41 @@ namespace BTCPayServer.Tests
         [Fact]
         public void ParseDerivationSchemeSettings()
         {
+            var testnet = new BTCPayNetworkProvider(ChainName.Testnet).GetNetwork<BTCPayNetwork>("BTC");
             var mainnet = new BTCPayNetworkProvider(ChainName.Mainnet).GetNetwork<BTCPayNetwork>("BTC");
             var root = new Mnemonic(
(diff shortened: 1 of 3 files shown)


Kukks (approved):

petzsch (approved):
fixes halve of my issues (spending is still unintuitiv UI/UX wise imho)

wallet import tested and works as it should. 👍

PR #4782, opened 2023-03-17
Receipt: Add payment proof

### LINKED ISSUE #4685: `Receipt` instead of BOLT11 show preimage
Currently on the receipt page we show bolt11 as a proof of payment ,which not only is irrelevant information for anyone willing to check if the invoice is paid, but due to it's length makes receipts oddly long. I think we do the same for payment requests.

I propose we add preimage/hash as an actual veritable proof of payment. 

With this embedded, I don't think we need `Show the payment list in the public receipt page` in settings, as any recepit should have a proof of payment, that's the point of a receipt. However the later can be split up in a different issue.

https://user-images.githubusercontent.com/36959754/220588069-da19d907-3107-4e6b-94f1-c3864f304608.MOV

dennisreimann:
We now display a payment proff on the receipt page:

- Onchain: Transaction Id
- Lightning: Preimage

Closes #4685.

grafik

### DIFF
--- a/BTCPayServer/Components/TruncateCenter/TruncateCenter.cs
+++ b/BTCPayServer/Components/TruncateCenter/TruncateCenter.cs
@@ -1,21 +1,29 @@
 using Microsoft.AspNetCore.Mvc;
 
-namespace BTCPayServer.Components.TruncateCenter
+namespace BTCPayServer.Components.TruncateCenter;
+
+/// <summary>
+/// Truncates long strings in the center with ellipsis: Turns e.g. a BOLT11 into "lnbcrt7…q2ns60y"
+/// </summary>
+/// <param name="text">The full text, e.g. a Bitcoin address or BOLT11</param>
+/// <param name="link">Optional link, e.g. a block explorer URL</param>
+/// <param name="classes">Optional additional CSS classes</param>
+/// <param name="padding">The number of characters to show on each side</param>
+/// <param name="copy">Display a copy button</param>
+/// <returns>HTML with truncated string</returns>
+public class TruncateCenter : ViewComponent
 {
-    public class TruncateCenter : ViewComponent
+    public IViewComponentResult Invoke(string text, string link = null, string classes = null, int padding = 7, bool copy = true)
     {
-        public IViewComponentResult Invoke(string text, string link = null, string classes = null, int padding = 7, bool copy = true)
+        var vm = new TruncateCenterViewModel
         {
-            var vm = new TruncateCenterViewModel
-            {
-                Classes = classes,
-                Padding = padding,
-                Copy = copy,
-                Text = text,
-                Link = link,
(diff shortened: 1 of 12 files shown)


dennisreimann:
TODO:

- [ ] Onchain: Remove duplicate Payment Proof
- [ ] Offchain: Remove Transaction Id

dennisreimann:
Improved the view and added a view component for displaying truncated text — it's configurable, displays things properly when printed and optionally shows a copy button. We can re-use it elsewhere once this is merged.

grafik

dstrukt:
Only thought w.r.t to the TX id's ( i ) is that it might not be immediately obvious it's a click away link .. wondering if we'll want to use a more common icon for that .. thoughts?

Otherwise, looks much better and cleaner .. going to explore some more options tomorrow when I'm in Figma.

dennisreimann:
I reused what we have on the labels, but am willing to adapt both if you find a better solution :)

NicolasDorier:
Something is strange.

The `Transaction Id` in on-chain term is only the tx hash, here it is the outpoint. (`txid-idx`)
The `Proof` should be the outpoint.

Here they are the same.

BTCPayServer/Components/TruncateCenter/TruncateCenter.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.AspNetCore.Mvc;
+
+namespace BTCPayServer.Components.TruncateCenter
+{
+    public class TruncateCenter : ViewComponent
```
NicolasDorier, line 5: small comment to understand what is that?

BTCPayServer/Services/Invoices/InvoiceEntity.cs
```diff
@@ -1330,5 +1330,6 @@ public interface CryptoPaymentData
 
         PaymentType GetPaymentType();
         string GetDestination();
+        string GetPaymentProof();
```
NicolasDorier, line 1333: Not sure it's good idea to make this interface bigger. If anything, we should probably break it down later.
Anyway, leave it like that for now, I may change your PR later.

BTCPayServer/Components/TruncateCenter/TruncateCenter.cs
dennisreimann: Added in [83eadc2](https://github.com/btcpayserver/btcpayserver/pull/4782/commits/83eadc2bcb9b6014f9f073c6376e7a3d765b09a3).

dennisreimann:
@NicolasDorier I've now removed the Transaction ID, as we use `GetPaymentId()` in several places and I didn't want to touch it, because it most likely assumed the outpoint.

Also this let's us display everything consistent and independent of the payment method. Added the link to the payment proof, so that it is also still present. 

grafik

You can take over now :)

NicolasDorier:
I removed interface and just hard coded for LN and bitcoin payments.

This is less flexible, but we will soon refactor all of this anyway.

PR #4783, opened 2023-03-17
Optimize invoice print view

### LINKED ISSUE #4729: Optimize invoice print view
Invoice is critical peace of information for merchant to deliver and fulfil the order. Some merchants that don't use CMS, may find it useful to have an easy way to print out the invoice and attach it to the order, or use it externally (for example storage unit worker receives and invoice to pack up the order).

Currently, you can print `invoice` just like any other web-page, but it's unoptimized, lengthy and requires a lot of paper waste which can end up being costly. 

I would like to propose that we add `print` button on the invoice, and optimize the print view to contain only order-critical data such as:

-  General Information
- Invoice Summary
- Additional information

We should not include `events/invoice log` as this information isn't typical for an invoice.

dennisreimann:
Closes #4729.


dennisreimann:
@NicolasDorier Let's do another round of optimizations here once #4782 is merged, because we can then also integrate the payment proof rather than the BOLT11. I'll also add a screenshot then.

Issue #4744, opened 2023-03-06
Apps and Integrations(store plugins) become plugins


pavlenex:
The exact way in which this will be done is TBD.

For now

- [ ] Moving PoS and Crowdfund from Apps to plugins, which means we can remove Apps page and in the sidebar.
- [ ] Moving Integrations Shopify as a plugin? and adding it to sidebar instead when activated?
- [ ] Moving Lightning Address? (TBD)
- [ ] Nuking Store Plugins page formerly known as Integrations?

dennisreimann:
![image](https://user-images.githubusercontent.com/886/224123985-34f2f7db-26b5-4231-8fe3-798deba6ef52.png)

PR #4784, opened 2023-03-18
Apps become plugins

dennisreimann:
Unified navigation for apps that are now plugins. Part of #4744, builds on #4780.

grafik


dstrukt:
Still trying to think of how we can differentiate things better .. by making the high level plugin name take up less space, differing icons, etc..

ShawnRthom88:
nothing

ShawnThomas

Kukks (approved):

PR #4792, opened 2023-03-20
Switch  Apps to json not YML

### LINKED ISSUE #2369: Switch App Item Template to JSON
> I think without proper YAML parsing we are setting us up for problems with edge cases here. For instance if the string contains `"`, it'll also break – a description like this will result in an invalid template:
> 
> ```
> This is "THE BEST TEA EVAR" someone said.
> 
> I tend to agree. 
> ```
> 
> With a YAML parser we could use [multiline YAML](https://yaml-multiline.info/) and get rid of the other hacks as well. Otherwise we need to at least patch this by escaping the quotes `\"` … not optimal.

If anything, we should get rid of the YAML parser and switch over to JSON. YAML is a superset of JSON anyway so it shouldn't be that hard. We're also already hiding the raw editor now so any benefits of yml readability has gone a long time ago.

_Originally posted by @Kukks in https://github.com/btcpayserver/btcpayserver/issues/2366#issuecomment-799479118_

Kukks:
closes #2369
As we're cleaning up many aspects of apps, we should clear out the yml mess. This PR changes the structure of app items to json from yml. A migration converts existing data. This also makes app item data expandable from plugins, allowing enhanced functionality to pos/crowdfund without building it into btcpay core

### DIFF
--- a/BTCPayServer.Data/Data/LightingAddressData.cs
+++ b/BTCPayServer.Data/Data/LightingAddressData.cs
@@ -1,6 +1,8 @@
 using System;
+using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Data;
@@ -41,4 +43,7 @@ public class LightningAddressDataBlob
     public decimal? Max { get; set; }
 
     public JObject InvoiceMetadata { get; set; }
+    
+    [JsonExtensionData] public Dictionary<string, JToken> AdditionalData { get; set; }
+    
 }
--- a/BTCPayServer/Views/Shared/PointOfSale/Public/Print.cshtml
+++ b/BTCPayServer/Views/Shared/PointOfSale/Public/Print.cshtml
@@ -1,11 +1,12 @@
 @using BTCPayServer.Payments.Lightning
 @using BTCPayServer.Plugins.PointOfSale.Models
+@using BTCPayServer.Services
 @using BTCPayServer.Services.Stores
 @using LNURL
 @inject BTCPayNetworkProvider BTCPayNetworkProvider
 @inject StoreRepository StoreRepository
 @model BTCPayServer.Plugins.PointOfSale.Models.ViewPointOfSaleViewModel
-
+@inject DisplayFormatter DisplayFormatter
 @{
     var store = await StoreRepository.FindStore(Model.StoreId);
     Layout = "PointOfSale/Public/_Layout";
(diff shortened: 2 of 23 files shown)


BTCPayServer/Views/Shared/PointOfSale/Public/Print.cshtml
```diff
@@ -90,16 +87,20 @@ else
                             <p class="card-title text-center">@Safe.Raw(item.Description)</p>
                         }
                         <div class="w-100 mb-3 fs-5 text-center">
-                            @switch (item.Price.Type)
+                            @{
+                                
+                                var formatted = DisplayFormatter.Currency(item.Price??0, Model.CurrencyCode);
```
dennisreimann, line 92:
```suggestion
                                var formatted = DisplayFormatter.Currency(item.Price ?? 0, Model.CurrencyCode, DisplayFormatter.CurrencyFormat.Symbol);
```

BTCPayServer/Views/Shared/PointOfSale/Public/Static.cshtml
```diff
@@ -17,8 +19,9 @@
             @for (var x = 0; x < Model.Items.Length; x++)
             {
                 var item = Model.Items[x];
-                var buttonText = string.IsNullOrEmpty(item.BuyButtonText) ? item.Price.Type != ViewPointOfSaleViewModel.Item.ItemPrice.ItemPriceType.Fixed ? Model.CustomButtonText : Model.ButtonText : item.BuyButtonText;
-                buttonText = buttonText.Replace("{0}", item.Price.Formatted).Replace("{Price}", item.Price.Formatted);
+                var formatted = DisplayFormatter.Currency(item.Price??0, Model.CurrencyCode);
```
dennisreimann, line 22:
```suggestion
                var formatted = DisplayFormatter.Currency(item.Price ?? 0, Model.CurrencyCode, DisplayFormatter.CurrencyFormat.Symbol);
```

dennisreimann (commented):
The converted description was wrapped in additional `"` for me, like this:

```
    "description": "\"Lovely, fresh and tender, Meng Ding Gan Lu ('sweet dew') is grown in the lush Meng Ding Mountains of the southwestern province of Sichuan where it has been cultivated for over a thousand years.\"",
```

Looking at the code I cannot find why this is, because it wasn't wrapped in the previous YAML.

NicolasDorier:
Concept ACK.

However, you should use YamlDotnet for the conversion. Your hand written converter will probably fail to cover many corner cases.

We'll be able to cleanup the YamlDotnet a while after.

NicolasDorier:
You also need to rebase.

BTCPayServer.Data/Data/LightingAddressData.cs
```diff
@@ -41,4 +43,7 @@ public class LightningAddressDataBlob
     public decimal? Max { get; set; }
 
     public JObject InvoiceMetadata { get; set; }
+    
+    [JsonExtensionData] public Dictionary<string, JToken> AdditionalData { get; set; }
```
dennisreimann, line 47: The IDE marks this as not used anywhere. Did this slip in?

dennisreimann (approved):
Manually tested the conversion once more, now it works :)

dennisreimann:
@Kukks there is one failing test, but other than that I think you can give this PR a final review, @NicolasDorier :)

NicolasDorier:
fix test @Kukks

NicolasDorier:
Please keep using YamlDotnet for the migration, not home grown parsing library. We will remove it in a few years.

Kukks:
> Please keep using YamlDotnet for the migration, no home grown parsing library. We will remove it in a few years.

It is already using the yamldonet for migration

Issue #4304, opened 2022-11-18
[Tracker] Improve recovery seed backup page


pavlenex:
This is a tracker issue for improving the view where user creates a wallet and recovery seed is shown to them. This one is potentially for 1.8.0 and later. 

Below is a list of tasks that we've discussed on Mattermost:

- [x] Display seed in a standard two-column way, the way in which we display seed currently is confusing #4793
- [ ] Iron out the copy and assure it's concise and clear
- [ ] Create a seed validation page, inspiration for design can be found in the bitcoin design guide [here](https://bitcoin.design/guide/daily-spending-wallet/backup-and-recovery/manual-backup/#how-to-display-a-recovery-phrase)
![Screenshot 2022-11-18 at 10 56 50](https://user-images.githubusercontent.com/36959754/202674924-68f96c25-b819-4afe-9323-e5c479a7ca5f.png)

daviogg:
I would love to work on this issue 👍

TChukwuleta:
Hello team, I would love to work on this issue.

dennisreimann:
@dstrukt can confirm, but I think these are the [latest mocks](https://www.figma.com/file/lE9dYRMqoABkJrn0VdXTE4/App?node-id=2253%3A77350&t=InneebFaRUx5C1VT-4):

![grafik](https://user-images.githubusercontent.com/886/224340493-c599288f-888b-409f-a99f-50403a6622a6.png)

![grafik](https://user-images.githubusercontent.com/886/224340542-86490003-0925-4121-81f3-3d1c08fab25d.png)

![grafik](https://user-images.githubusercontent.com/886/224340660-66cdff6f-2bff-4f78-a597-87175df94c1e.png)

TChukwuleta:
Hello @dennisreimann, just to confirm, these images are in order according to the flow yea?

dennisreimann:
Yes, we present the seed words once a new wallet is created within BTCPay Server — you can get to that via the Wallet Setup and choosing the create option.

Right now we only have the first screen of those mocks, which we'd need to alter the order of word on to make it more intuitive. Currently we display them in column order and it is supposed to go from left to right in rows.

The second and third mock shows the confirmation screen, which should be interactive and have success and error state as shown in the third image.

Let me know in case you have questions or need help with the implementation — you can also reach out on [our Mattermost chat](https://chat.btcpayserver.org). Have fun!

TChukwuleta:
[...]

PR #4793, opened 2023-03-21
Bg recoverseedupdate

TChukwuleta:


dennisreimann:
Thanks for the contribution and hanging in there ;)

@pavlenex @dstrukt If we want to keep it in small chunks, we could get this UI rework in already and have the rest of #4304 done in a separate PR. This one adjust the recovery seed view so that it is closer to the mocks and also has the improved order of the words — I'd go with that.

dennisreimann:
Added a screenshot to the description.

dennisreimann:
I think this one should cover the first two points of #4304:

- Display seed in a standard two-column way, the way in which we display seed currently is confusing
- Iron out the copy and assure it's concise and clear

The first one is already done and the second one can be incorporated here as well to have that page complete.

The then only left point would be "Create a seed validation page", which can be seen as a separate thing to be tackled in a later PR.

NicolasDorier (approved):

dstrukt:
Will give it a review this evening as well!

dstrukt:
Initial thought:

Do we want to order the seed phrase with two columns top to bottom?

e.g.

COL LEFT
1.
2.
3.
4.
5.
6.

COL RIGHT
7.
8.
9.
10.
11.
12

dennisreimann:
After even some more [discussion on Mattermost](https://chat.btcpayserver.org/btcpayserver/pl/m8brw93fbfg13kydcn1ee8kn8w) we choose to go with the simplest possible way to display the seed: One column, no ambiguity.

Visually it is less appealing, but then again that page is rarely shown and has a very clear purpose — and tbh looking good isn't the most important part here. This is very concise, clear and works across cultures.

grafik

PR #4796, opened 2023-03-21
Wallet transactions: Add label manager

### LINKED ISSUE #4706: Use label manager component in wallet transactions list
In addition to #4700 we should reuse the new label management component (see wallet receive page) on the transactions list. 

This will also improve the UX, because managing labels would then work via AJAX instead of complete page reloads when assigning/removing labels.

### LINKED ISSUE #4755: [Feature Request]: Ability to label destination addresses
In the Send Wallet form:

image

It is possible to pay multiple destination. As somebody who wants to make a payment to 10 differents people, I'd wish I could add a label to every destination to make it easier to review my action.

In the "Transaction Review" screen before broadcast, those label could be shown, making it more easy to review the outputs.

dennisreimann:
Uses the label manager component in the wallet transactions list. 
Closes #4706
Closes #4755

labels-inline

Also improved the label manager UI to work better with the list: Introduced an inline display mode, which removes the form input appearance. Also adjusted the dropdown and options to fit what we have elsewhere.

### DIFF
--- a/BTCPayServer/Components/LabelManager/LabelManager.cs
+++ b/BTCPayServer/Components/LabelManager/LabelManager.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using BTCPayServer.Services;
 using Microsoft.AspNetCore.Mvc;
@@ -6,15 +7,17 @@ namespace BTCPayServer.Components.LabelManager
 {
     public class LabelManager : ViewComponent
     {
-        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels, bool excludeTypes = true, bool displayInline = false, Dictionary<string, RichLabelInfo> richLabelInfo = null)
+        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels, bool excludeTypes = true, bool displayInline = false, Dictionary<string, RichLabelInfo> richLabelInfo = null, bool autoUpdate = true, string selectElement = null)
         {
             var vm = new LabelViewModel
             {
                 ExcludeTypes = excludeTypes,
                 WalletObjectId = walletObjectId,
-                SelectedLabels = selectedLabels,
+                SelectedLabels = selectedLabels?? Array.Empty<string>(),
                 DisplayInline = displayInline,
-                RichLabelInfo = richLabelInfo
+                RichLabelInfo = richLabelInfo,
+                AutoUpdate = autoUpdate,
+                SelectElement = selectElement
             };
             return View(vm);
         }
--- a/BTCPayServer/Components/LabelManager/LabelManager.cs
+++ b/BTCPayServer/Components/LabelManager/LabelManager.cs
@@ -5,12 +5,14 @@ namespace BTCPayServer.Components.LabelManager
 {
     public class LabelManager : ViewComponent
     {
-        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels)
+        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels, bool excludeTypes = true, bool displayInline = false)
         {
             var vm = new LabelViewModel
             {
-                ObjectId = walletObjectId,
-                SelectedLabels = selectedLabels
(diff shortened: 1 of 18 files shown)


BTCPayServer/Components/LabelManager/LabelManager.cs
```diff
@@ -5,12 +5,14 @@ namespace BTCPayServer.Components.LabelManager
 {
     public class LabelManager : ViewComponent
     {
-        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels)
+        public IViewComponentResult Invoke(WalletObjectId walletObjectId, string[] selectedLabels, bool excludeTypes = true, bool displayInline = false)
```
dennisreimann, line 8: Options that got added for the trasactions list, on which one was able to manage type labels as well. Inline mode is also for the list.

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -1353,17 +1355,26 @@ public async Task<IActionResult> UpdateLabels([ModelBinder(typeof(WalletIdModelB
...
-               label = tuple.Label,
-               color = tuple.Color,
-               textColor = ColorPalette.Default.TextColor(tuple.Color)
-           }));
+            var walletObjectId = !string.IsNullOrEmpty(type) && !string.IsNullOrEmpty(id)
+                ? new WalletObjectId(walletId, type, id)
+                : null;
+            var labels = walletObjectId == null
+                ? await WalletRepository.GetWalletLabels(walletId)
+                : await WalletRepository.GetWalletLabels(walletObjectId);
```
dennisreimann, line 1369: It turned out I didn't need this, but this makes it more flexible so that one can request either labels for the whole wallet or individual wallet objects.

BTCPayServer/Views/UIWallets/_WalletTransactionsList.cshtml
```diff
@@ -12,31 +17,38 @@
...
-                                   data-bs-html="true"
-                                   data-bs-toggle="tooltip"
-                                   data-bs-custom-class="transaction-label-tooltip">
-                                    <vc:icon symbol="info"/>
-                                </a>
-                            }
-                        </div>
-                    }
-                </div>
+                <noscript>
```
dennisreimann, line 20: Moved the previous approach into `noscript` parts — shall we keep it or remove that?

Kukks: this is fine

Kukks (changes requested):
just tested and it feels good but we now lost a key feature: the descriptions of the labels are gone (for example payout ids, invoice link, etc)

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -122,6 +124,7 @@
             
             $indicator.classList.add('d-none');
             formatDateTimes(document.getElementById('switchTimeFormat').dataset.mode);
+            initLabelManagers();
```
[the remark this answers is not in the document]
dennisreimann: It doesn't, see the `:not(.tomselected)` in the selector. It only inits the ones that aren't initialized already.

dennisreimann:
> we now lost a key feature: the descriptions of the labels are gone (for example payout ids, invoice link, etc)

Good point, added in [e5c5cdc](https://github.com/btcpayserver/btcpayserver/pull/4796/commits/e5c5cdc55eca81ac39abce949a8892907803931f).

grafik

BTCPayServer/Views/UIWallets/_WalletTransactionsList.cshtml
NicolasDorier: imho the `noscript` should go. Burning trees for nothing.

NicolasDorier:
I think you should get ride of `<noscript>` stuff, which only work if no JS is present.
While I think it's cool to have it on public pages, backend pages can just require javascript.

Kukks (changes requested):
weird red pixels?

also there is no link on the badges that used to have links

PR #4799, opened 2023-03-21
Wallet Transactions Export: Add BIP-329 support

### LINKED ISSUE #4536: [Feature] Support BIP 329: Wallet Labels Export Format
Now that [BIP 329](https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki) exists, we can integrate it and help unify the export of labels that may be attached to various common types of records in a wallet.

dennisreimann:
Adds support for the [BIP 329](https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki) export format.

Sample transactions export:

```jsonl
{"type":"tx","ref":"9eb4b1f69dc0aad88f2530fdd124c77021c1633500e14369d0896bf75a025596","label":"microtransaction"}
{"type":"tx","ref":"98ee2e20e45967eeda330b191d881584d1e02ea05e633e4cb7c48aea8980fc63","label":"test"}
{"type":"tx","ref":"84ce8b198d85e715588da114b7b441833cfc4892ba9aa1f7f3cfbfdcf2ca1235","label":"Mining Income, microtransaction, test"}
```

Currently only the transaction data is exported, but I think we can also leverage the new labeling system to export other types, like `addr` as well. @NicolasDorier 

Closes #4536.

grafik

### DIFF
--- a/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
+++ b/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
@@ -64,10 +64,10 @@ namespace BTCPayServer.Services.Wallets.Export
             {
                 if (tx.Labels is { Count: > 0 })
                 {
-                    tx.Labels.ForEach(label =>
-                        res.Add(JsonConvert.SerializeObject(
-                            new { Type = "label", Ref = tx.TransactionId, Label = label }, Formatting.None,
-                            serializerSett)));
+                    var label = string.Join(", ", tx.Labels);
+                    var obj = new { Type = "tx", Ref = tx.TransactionId, Label = label };
+                    var json = JsonConvert.SerializeObject(obj, Formatting.None, serializerSett);
+                    res.Add(json);
                 }
                 return res;
             });
--- a/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
+++ b/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
@@ -108,18 +108,18 @@ namespace BTCPayServer.Services.Wallets.Export
             Map(m => m.Labels).ConvertUsing(row => row.Labels == null ? string.Empty : string.Join(", ", row.Labels));
         }
     }
-#nullable restore
+
     public class ExportTransaction
     {
         [Name("Transaction Id")]
-        public string TransactionId { get; set; }
+        public string TransactionId { get; set; } = string.Empty;
         public DateTimeOffset Timestamp { get; set; }
-        public string Amount { get; set; }
-        public string Currency { get; set; }
+        public string Amount { get; set; } = string.Empty;
+        public string Currency { get; set; } = string.Empty;
 
         [Name("Is Confirmed")]
         public bool IsConfirmed { get; set; }
-        public string Comment { get; set; }
-        public List<string> Labels { get; set; }
+        public string? Comment { get; set; }
+        public List<string>? Labels { get; set; }
     }
 }
--- a/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
+++ b/BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
@@ -1,8 +1,10 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
 using BTCPayServer.Data;
 using CsvHelper.Configuration;
 using CsvHelper.Configuration.Attributes;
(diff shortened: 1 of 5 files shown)


BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
```diff
@@ -47,12 +42,33 @@ public string Process(IEnumerable<TransactionHistoryLine> inputList, string file
...
+            {
+                ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
+                ContractResolver = new Newtonsoft.Json.Serialization.CamelCasePropertyNamesContractResolver()
+            };
+            var lines = txs.Select(tx => 
+                JsonConvert.SerializeObject(new
+                {
+                    Type = "tx",
+                    Ref = tx.TransactionId,
+                    Label = tx.Labels is { Count: > 0 } ? string.Join(", ", tx.Labels): string.Empty
```
NicolasDorier, line 68: I believe you need 1 line per label. If a tx has several label, there should be several lines.

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -165,6 +165,7 @@
             <div class="dropdown-menu" aria-labelledby="ExportDropdownToggle">
                 <a asp-action="Export" asp-route-walletId="@walletId" asp-route-format="csv" asp-route-labelFilter="@labelFilter" class="dropdown-item export-link" target="_blank" id="ExportCSV">CSV</a>
                 <a asp-action="Export" asp-route-walletId="@walletId" asp-route-format="json" asp-route-labelFilter="@labelFilter" class="dropdown-item export-link" target="_blank" id="ExportJSON">JSON</a>
+                <a asp-action="Export" asp-route-walletId="@walletId" asp-route-format="bip329" asp-route-labelFilter="@labelFilter" class="dropdown-item export-link" target="_blank" id="ExportBIP329">JSON (BIP-329)</a>
```
NicolasDorier, line 168: `Wallet Labels Export (BIP-329)`

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
dennisreimann: Is it like that, @craigraw? I couldn't find that case in the BIP.

craigraw: There is no specific instruction on how to handle duplicate references, but since most wallets only show one label per reference I believe comma-separated is the way to go.

dennisreimann: Done in [5a904f2](https://github.com/btcpayserver/btcpayserver/pull/4799/commits/5a904f2872fe0605b5a802673a13c9b5f5649167).

BTCPayServer/Views/UIWallets/WalletTransactions.cshtml
dennisreimann: Done in [351f492](https://github.com/btcpayserver/btcpayserver/pull/4799/commits/351f49234439f6e94a51454c98b4ecada907b3be)

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
dennisreimann: @craigraw Does Sparrow already support this, so that we can test it?

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
```diff
@@ -41,18 +36,44 @@ public string Process(IEnumerable<TransactionHistoryLine> inputList, string file
...
+                ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
+                ContractResolver = new Newtonsoft.Json.Serialization.CamelCasePropertyNamesContractResolver()
+            };
+            var lines = txs.Aggregate(new List<object>(), (res, tx) =>
+            {
+                if (tx.Labels is { Count: > 0 })
+                {
+                    tx.Labels.ForEach(label =>
+                        res.Add(JsonConvert.SerializeObject(
+                            new { Type = "label", Ref = tx.TransactionId, Label = label }, Formatting.None,
```
Kukks, line 69: This doesn't look right according to the bip. There is no type `label`, just a property `label`.
I think the spec says you need to create a record of type `tx`, for each label applied to it.

Kukks (changes requested):

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
```diff
@@ -64,10 +64,10 @@ private static string ProcessBip329(List<ExportTransaction> txs)
             {
                 if (tx.Labels is { Count: > 0 })
                 {
-                    tx.Labels.ForEach(label =>
-                        res.Add(JsonConvert.SerializeObject(
-                            new { Type = "label", Ref = tx.TransactionId, Label = label }, Formatting.None,
-                            serializerSett)));
+                    var label = string.Join(", ", tx.Labels);
```
Kukks, line 67: No, I think you need to create a new row for each label.

Kukks (changes requested):

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
dennisreimann: I reverted the change from making it one per line because see Craig's comment above.

BTCPayServer/Services/Wallets/Export/TransactionsExport.cs
```diff
@@ -41,18 +36,44 @@ public string Process(IEnumerable<TransactionHistoryLine> inputList, string file
...
-            throw new Exception("Export format not supported");
+        // https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki
+        private static string ProcessBip329(List<ExportTransaction> txs)
+        {
+            var serializerSett = new JsonSerializerSettings
+            {
+                ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
+                ContractResolver = new Newtonsoft.Json.Serialization.CamelCasePropertyNamesContractResolver()
+            };
+            var lines = txs.Aggregate(new List<object>(), (res, tx) =>
```
NicolasDorier, line 63: use of Aggregate instead of foreach is really weird.

NicolasDorier:
I give up implementing address export as well.
I need to refactor the export stuff so we don't need another hack overload.

NicolasDorier (approved):

PR #4802, opened 2023-03-22
Introduce Rate limit for Public rates endpoint

Kukks:
cc @kaloudis


NicolasDorier:
NACK, I assume lot's of people using it from the backend, and not from JS. If that's the case this PR will make them DDoS themselves.

I can't believe some users are intentionally spamming this link, and even if they did, it should not DDoS as there isn't really anything heavy there.

PR #4803, opened 2023-03-22
Reconstruct issue template

pavlenex:
This PR adds a bit more context and adds a few more options for people looking to open new issues.

Preview available here 

https://github.com/pavlenex/btcpayserver/issues/new/choose


NicolasDorier:
I really think discussions sucks

pavlenex:
@NicolasDorier What sucks about discussions? What else you propose?

Bangalisch:
> @NicolasDorier What sucks about discussions? What else you propose?

To add to pavs, 
They already exist and get utilized to a point where things sporadically move quicker than we can spot(think of) these requests ourselves. Wouldn't it make more sense to guide them properly and funnel them down into tasks once we know it's legit requested by the community? ( we can measure this by participants, upvotes and other arbitrary stuff we decide on)

PR #4819, opened 2023-03-27
NFC improvements

### LINKED ISSUE #4807: [Bug]: ndefreader should't stop scanning immediately after the card is read.

I'm not actually running BTCPay, just reporting an issue.

The issue is on a POS device in the browser.

### What happened?

Some people run into the issue that another app opens when someone pays with an NFC card. This is because when the ndefreader scan is aborted, the OS takes over the nfc scan. The card will be read again and the default app to handle the link on the card will be opened (sometimes it's the 'choose app' screen).

You'll probably want to stop scanning after a few seconds or so, when the card has been removed from the reader, before stopping the scan.

I'm not sure about the exact steps.

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Two changes which fix #4807:

- Once permissions are granted we start scanning immediately, no need to ask for permissions or have the user click the button again
- We don't abort the scan, which gets rid of the cases in which the OS took over after the scan, because the user left the card on the device (I think @Bangalisch also reported that problem before)

Also adds feedback for the NFC states scanning and submitting.

Video: https://youtu.be/CvckVuSwxis

### DIFF
--- a/BTCPayServer/Plugins/NFC/NFCController.cs
+++ b/BTCPayServer/Plugins/NFC/NFCController.cs
@@ -175,18 +175,18 @@ namespace BTCPayServer.Plugins.NFC
                 }
             }
 
-            if (bolt11 is null)
+            if (string.IsNullOrEmpty(bolt11))
             {
                 return BadRequest("Could not fetch BOLT11 invoice to pay to.");
             }
 
             var result = await info.SendRequest(bolt11, httpClient);
-            if (result.Status.Equals("ok", StringComparison.InvariantCultureIgnoreCase))
+            if (!string.IsNullOrEmpty(result.Status) && result.Status.Equals("ok", StringComparison.InvariantCultureIgnoreCase))
             {
                 return Ok(result.Reason);
             }
 
-            return BadRequest(result.Reason);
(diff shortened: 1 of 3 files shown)


dennisreimann:
Regardless of the changes in #4756 I think we should remove the LNURL-Withdraw input fallback. I don't think it gets used and it would simplify things a bit in this otherwise solely NFC-related component.

Based on what we come up with in #4756 we will tackle an improved fallback in #4753.

Thoughts @Kukks @pavlenex ?

Kukks (approved):

dennisreimann:
@Kukks one more thing I need to check and that came to my mind just now: We need to deregister/abort when the component is removed from the DOM. 

I'll do that in a new PR, shall I also remove the LNURL-W fallback then?

Kukks:
> @Kukks one more thing I need to check and that came to my mind just now: We need to deregister/abort when the component is removed from the DOM.
> 
> I'll do that in a new PR, shall I also remove the LNURL-W fallback then?
Don't remove, just make it hidden unless a querystring key of `nfctest` is there

Issue #3787, opened 2022-05-28
Improve security.md and better document our response processes.


pavlenex:
Filing this issue as to-do to myself, as I think we can do a bit better job documenting our vulnerability response processes.

We should put a bit more context on how we handle security and reporting.
It's been recommended that we follow the following runbooks:

- https://github.com/google/oss-vulnerability-guide/blob/main/runbook.md
- https://github.com/google/oss-vulnerability-guide/blob/main/guide.md#the-vulnerability-response-process
- https://github.com/google/oss-vulnerability-guide/blob/main/guide.md#if-you-are-using-github-2

- [ ] Re-write the [security.md ](https://github.com/btcpayserver/btcpayserver/blob/master/SECURITY.md)page with proper introduction on why reporting sec vuln is important
- [ ] Add vulnerability response process and steps we take
- [ ] Discuss about how to handle vulnerability patching, privately via GitHub advisory vs publicaly, a resource [here](https://github.com/google/oss-vulnerability-guide/blob/main/guide.md#if-you-are-using-github-2)
- [ ] Define bug bounties and add https://huntr.dev as a resource

Zaxounette:
Just commenting to align everybody that's assigned to this issue.
I'm taking it on.

Just FYI, won't be done for 1.8.0 though, that's 100% certain.

pavlenex:
@Zaxounette We release 1.8.0 on Thursday, seems a bit of a tight time-frame, but we can do this for 1.8.x release? Of course if you have it ready, we can include it for 1.8.0, but I don't think it's essential to be added in a major release, since it's just process documentation end-users probably don't care much about, but us and security researchers do.

Zaxounette:
Sorry, didn't get the mention for this.
I should have something to push for review not this monday (27/02), but the next (06/03) TOPS.

Zaxounette:
Sorry for the delay!

PR up for the reporting process and refactor of security.md: https://github.com/btcpayserver/btcpayserver/pull/4815

Regarding completing this issue entirely, on my TODO list:
* Add a fully fledged vulnerability response and management process to our documentation, as a dedicated page, to not bloat the purpose of the security.md page

pavlenex:
@Zaxounette since @NicolasDorier now merged https://github.com/btcpayserver/btcpayserver/pull/4815 is there something else you plan to do in order to consider the issue complete? Or should we close it if everything was addressed?

Zaxounette:
[...]

PR #4821, opened 2023-03-28
Create codeql.yml

Zaxounette:
Part of a larger effort around #3787 

This adds codeql.yml, the native github code scanning tool.

I set it to scan on a weekly basis, with option to manually trigger the action if needed.
Scan on commit or PR can be uncommented in the future, as the need arises.

All the rest is the default configuration that github provides. This should work on most repositories. 
Autobuild needs to be tested. If it fails, we'll have to figure out how to add build steps. 

Enhancements:
- [ ] Define repo paths to exclude from the scan (testing libraries, development code only, etc ...) if applicable

Requirements to be able to push this and test it:
- [x] Create a `Github Security` group in the BTCPay Server org
- [x] Add `Github Security` group as a `Security Manager` (gives read, write and admin permissions to the **Security tab only** in all repos of the org)
- [ ] Add Zaxounette to `Github Security` group so he can:
  - [ ] Verify actions are set up appropriately to use codeql
  - [ ] Test codeql action manually and verify the SARIF files are properly uploaded to github in the Security tab


NicolasDorier:
hey @Zaxounette I just did what you asked

PR #4823, opened 2023-03-28
Checkout: Fix language select width in modal

### LINKED ISSUE #4820: [Bug]: Language Selector on Modal (light theme) is barely visible

v.1.8.4

Docker

### What happened?

grafik

grafik

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox

### Additional information

none

dennisreimann:
Initially delays the width calculation to the next render frame, so that the page is safely rendered. Closes #4820.

grafik

### DIFF
--- a/BTCPayServer/wwwroot/checkout-v2/checkout.js
+++ b/BTCPayServer/wwwroot/checkout-v2/checkout.js
@@ -62,7 +62,11 @@ function updateLanguageSelect() {
     $languageSelect.parentElement.appendChild(element);
     const width = element.offsetWidth;
     $languageSelect.parentElement.removeChild(element);
-    $languageSelect.style.setProperty('--text-width', `${width}px`);
+    if (width && width > 0) {
+        $languageSelect.style.setProperty('--text-width', `${width}px`);
+    } else { // in case of modal this might not be rendered properly yet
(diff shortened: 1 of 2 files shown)


BTCPayServer/wwwroot/checkout-v2/checkout.js
```diff
@@ -224,7 +224,7 @@ function initApp() {
             if (this.isProcessing) {
                 this.listenForConfirmations();
             }
-            updateLanguageSelect();
+            window.requestAnimationFrame(updateLanguageSelect);
```
Kukks, line 227: Shouldnt we use Vue's nextTick instead? https://vuejs.org/api/general.html#nexttick

Kukks (changes requested):

BTCPayServer/wwwroot/checkout-v2/checkout.js
dennisreimann: @kukks I think they amount to the same outcome, but I found a better way to solve it in [11ea96b](https://github.com/btcpayserver/btcpayserver/pull/4823/commits/11ea96b3982fdbd03c4bab14b2eb70fe2da778b4). There we should use the native DOM API there, because it's not in the Vue context.

PR #4827, opened 2023-03-28
Checkout v2: Improve expired paid partial state

### LINKED ISSUE #4303: Improve checkout page error state messages
Post 1.7.0, but we should use this issue as a tracker to improve the copy to better explain invoice error states to the end user.

- [ ] When invoice is `ExpiredPaidPartial` we provide same text and CTA as if it was `expired`. Expired paid partial should have a different screen state telling customer to contact merchant to paid the remaining due or request a refund. More context [in this comment.](https://github.com/btcpayserver/btcpayserver/issues/4261#issuecomment-1319765058)

dennisreimann:
Adds an email property to the store blob and displays that email as a "Contact us" button on the expired view if the invoice was paid partially. Closes #4303.

grafik

### Sample email

The subject is prefilled with `Invoice {InvoiceID}` and the email body looks like this:

```
Hi {StoreName},

the invoice {InvoiceID} (Order ID: {OrderId}) was paid partial.
Please let me know how to proceed.

Best regards!
```


Kukks:
Are emails the right approach for this? I don't have any suggestions yet, but every time I'm personally greeted with a mailto link, I cry a little

dennisreimann:
@Kukks It's the easiest approach by far (the alternative was having an email form and requiring the store to have SMTP set up) and this is mostly for getting customer and merchant into contact.

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
```diff
@@ -204,9 +204,10 @@
                             <span class="fw-semibold" v-t="'view_details'"></span>
                             <vc:icon symbol="caret-down" />
                         </button>
-                        <p class="text-center mt-3" v-html="replaceNewlines($t('invoice_expired_body', { storeName: srvModel.storeName, minutes: @Model.MaxTimeMinutes }))"></p>
+                        <p class="text-center mt-3" v-html="replaceNewlines($t(isPaidPartial ? 'invoice_paidpartial_body' : 'invoice_expired_body', { storeName: srvModel.storeName, minutes: srvModel.maxTimeMinutes }))"></p>
                     </div>
                     <div class="buttons">
+                        <a v-if="isPaidPartial && contactLink" class="btn btn-primary rounded-pill w-100" :href="contactLink" v-t="'contact_us'" id="ContactLink"></a>
```
Kukks, line 210: Why only paid partial?  Probably makes sense for any situation that does not result in a processing/settled state but a payment has been made. (partial, over, late)

Kukks (changes requested):

BTCPayServer/Views/UIInvoice/CheckoutV2.cshtml
dennisreimann: That's basically the condition, as we also don't have that concrete state:

```
isPaidPartial () {
    return this.btcPaid > 0 && this.btcDue > 0;
},
```

dstrukt:
> @Kukks It's the easiest approach by far (the alternative was having an email form and requiring the store to have SMTP set up) and this is mostly for getting customer and merchant into contact.

The only alternative would be including a .. 
`Please email 
{email}` (with a copy icon
at the end of the copy.

I too hate the mailto: but realize we're pretty limited here on the UX.

Kukks:
OK let's do this for now, maybe we can do a plugin that adds a mini embedded chat ojnth checkout to spice things up without bloating core

NicolasDorier:
@dennisreimann I don't think email is a good idea.
I suggest you instead allow input any URL, so merchants who wants to be contacted by mail can just enter `mail:blah@blag.org`, while those who wants to redirect to some custom page can put the URL. If it is an URL, we can later add a query string so the page knows why this "Contact Us" is reached.

dennisreimann:
@NicolasDorier Ok, but what about the prefilled email body? Would we also remove that or append it in case it's a `mailto` link? I don't see a general link as that useful, because the context would get lost and we might need to explain how this is different from the Store Website URL, which is also present.

@pavlenex I think this just wasn't specified enough to be worked on. Maybe we should postpone it until the next release before rushing it.

pavlenex:
I am fine with postponing this one for until we iron out the details, may as well then postpone #4850 just to pack it all up together and ship in 1.9.10 @NicolasDorier @Kukks what do you guys think?

BTCPayServer/wwwroot/checkout-v2/checkout.js
```diff
@@ -172,6 +175,16 @@ function initApp() {
                     ? this.srvModel.merchantRefLink
                     : null;
             },
+            contactLink () {
+                if (!this.srvModel.storeEmail) return null;
+                
+                const email = this.srvModel.storeEmail;
+                const subject = `Invoice ${this.srvModel.invoiceId}`;
+                const details = this.srvModel.invoiceId + (this.srvModel.orderId ? ` (Order ID: ${this.srvModel.orderId})` : '');
+                const body = `Hi ${this.srvModel.storeName},\n\nthe invoice ${details} was paid partial.\nPlease let me know how to proceed.\n\nBest regards!`;
```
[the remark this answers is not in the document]
Moxita33: ****

NicolasDorier:
Idea from @Kukks : for a url, we can use the placeholder like `{OrderId}` so the merchant can know automatically which OrderId is impacted by the problem.

dennisreimann:
Updated to make it an URL instead of email address.

grafik

Kukks (approved):

Kukks (approved):

dstrukt (approved):
Like the latest proposed solution, LGTM!

Kukks (approved):

PR #4828, opened 2023-03-28
Improve pagination

### LINKED ISSUE #3835: Improve pagination
Something I've wanted to do for a little while, as I'm not a fan of the current pagination, visually or functionally.

I'd like to propose we do away with the "show: X/Y/Z" component, and simplify to a centered "Prev/1/2/3/Next" with a default of 50 or 100 items per page / call. 

Few mocks outlining the look. The bottom styling is the closest inline with our current button component, but if there's a preference for 1 of the 3, we can go in that direction.

Screen Shot 2022-06-07 at 8 11 41 AM

This a low priority item. Thoughts?

dennisreimann:
Based on the design repo updates by @benalleng in btcpayserver/btcpayserver-design#61. Closes #3835.

In order to imprve performance we don't fetch the total page count anymore, so most of the paging options do not exist. However, this styles the pagination remainders according to @dstrukt's mocks.

grafik

### DIFF
--- a/BTCPayServer/Components/Pager/Default.cshtml
+++ b/BTCPayServer/Components/Pager/Default.cshtml
@@ -6,51 +6,53 @@
 
 @if (Model.Total is null || Model.Total.Value > pageSizeOptions.Min())
 {
-    <nav aria-label="..." class="w-100 clearfix">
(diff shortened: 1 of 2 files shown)


NicolasDorier:
I merged, as we will release 1.9 RC soon, @dstrukt I think better you take a look then

PR #4833, opened 2023-03-29
Checkout Appearance: Move language settings into the general section

dennisreimann:
The language settings work for both versions of the checkout, so let's make them available for both. Request by @petzsch.


Kukks (approved):

dstrukt (approved):
LGTM

PR #4843, opened 2023-03-31
Store rule emails in HTML + Test rule emails

Nisaba:
Rules emails are a frustrating experience because HTML is not handled and carriage returns that are set are not taken into account.
I therefore propose this development which allows the support of HTML in the rules emails as well as a button allowing them to be tested.

Capture


petzsch:
You will probably need to update the tests as well. Selenium is [failing](https://circleci.com/gh/btcpayserver/btcpayserver/31767?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary):

BTCPayServer/Controllers/UIStoresController.Email.cs
```diff
@@ -61,17 +61,55 @@ public async Task<IActionResult> StoreEmails(string storeId, StoreEmailRuleViewM
...
+                else
+                {
+                    try
+                    {
+                        var emailSettings = blob.EmailSettings;
+                        using var client = await emailSettings.CreateSmtpClient();
+                        var message = emailSettings.CreateMailMessage(MailboxAddress.Parse(rule.To), "(test) " + rule.Subject, rule.Body, true);
+                        await client.SendAsync(message);
+                        await client.DisconnectAsync(true);
+                        TempData[WellKnownTempData.SuccessMessage] = $"Rule email ssaved and sent to {rule.To}. Please verify you received it.";
```
Kukks, line 89:
```suggestion
                        TempData[WellKnownTempData.SuccessMessage] = $"Rule email saved and sent to {rule.To}. Please verify you received it.";
```

Nisaba:
@Kukks : fixed, thank you
@petzsch : I'm not yet very familiar with the BTCPay test environment and I don't quite understand the error generated by Selenium since the Rules_0__Trigger element is indeed generated in the page... Any idea?

Kukks:
I think you should remove the validation checks on the body and subject as they are validated beforehand with the model state check. I think you should also not save when testing (or otherwise the status message is wrong)

Nisaba:
Ok, it works now.
I had to rework the controllers for better command management.
And modify the test so that the new control on the Body HTML can pass.

PR #4850, opened 2023-04-04
Make Checkout V2 the default

NicolasDorier:
The PR:

* New stores are V2
* Preset for ecommerce and retail both use V2
* Old stores which never have set checkoutType will keep using V1 so we don't change their settings.
* In checkout experience, rename the checkbox from `Use new checkout` to `Use classic checkout`

Ping @dennisreimann

### DIFF
--- a/BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
+++ b/BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
@@ -95,7 +95,7 @@
 				</div>
 			</h3>
             
-            <div class="d-flex align-items-center mb-3">
+            <div class="d-flex align-items-center mb-4">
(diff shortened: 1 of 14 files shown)


BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml
```diff
@@ -94,21 +94,12 @@
 			</h3>
             
             <div class="d-flex align-items-center mb-3">
```
dennisreimann, line 96:
```suggestion
            <div class="d-flex align-items-center mb-4">
```

dstrukt:
Hell. Yes.

Ryleeryatt:
How so I get paied

petzsch:
> How so I get paied

https://blog.btcpayserver.org/bitcoin-bonus-scam/

PR #4851, opened 2023-04-04
Add cart items to receipt

### LINKED ISSUE #4716: expose point of sale and custom form data in the receipt (product, form, tip, discount)
Yo, 

Would it be possible to create an abstract view to print usefull data only from invoice page? 

Imagine the following : 
I run a restaurant, with the custom form i now have to power to add table number to invoices, which in theory would make a perfect receipt for the kitchen. If the view printed would show the Custom fields Data + Items ordered only. 

Currently the only way to do this is to print the full invoice and get between 5-6 pages A4 size.. However kitchen printers are 9/10 times thermal receipt printers, this turns into a Loooooong receipt. So the workaround now is to copy all data from invoice view to notepad, dump unnecessary data and send to printer.

Kukks:
closes #4716

This adds the items bought in a pos to the receipt page.


NicolasDorier:
Seems good to me

PR #4854, opened 2023-04-05
Fixed nav menu active states + Better Fake Custodian

woutersamaey:
Fixes issues reported in Mattermost for the 1.9 release

- Fixed nav menu active states
- Fixes a bug where assets were not visible for the Fake Custodian (and other custodians that did not have a `fiatValue` known. Assets without `fiatValue` went lost during the sorting of the table rows)
- Adds deposit support to fake custodian, so this part of the UI becomes visible during dev and we can see the "copy to clipboard" issue.
- Added withdrawal + trade support to Fake Custodian. Upon each action, balances are updated to reflect what you hold.
- Custodian UI fully visible and functional when selecting the Fake Custodian

Only thing missing is a bug where the config form appears to have empty balance fields after a balance was changed programmatically. This is only a display issue affecting devs and should not hold this PR back.

### DIFF
--- a/BTCPayServer/Controllers/UICustodianAccountsController.cs
+++ b/BTCPayServer/Controllers/UICustodianAccountsController.cs
@@ -432,7 +432,8 @@ namespace BTCPayServer.Controllers
                             }
                             catch (WrongTradingPairException)
                             {
-                                // Cannot trade this asset, just ignore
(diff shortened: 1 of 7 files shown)


woutersamaey:
Ping @NicolasDorier

PR #4858, opened 2023-04-06
Bugfix: Cannot create new Fake Custodian

woutersamaey:
Minor bugfix that fixes the issue where you can't create a new Fake Custodian because you can't fill out the form due to a required field you cannot answer.

Expanded the interface with a bool `isNew` so it knows the config form is for a new custodian (and not an update/edit). Tried several alternatives, but there was simply no other, reliable way of detecting create vs edit.

### DIFF
--- a/BTCPayServer.Abstractions/Custodians/ICustodian.cs
+++ b/BTCPayServer.Abstractions/Custodians/ICustodian.cs
@@ -21,6 +21,6 @@ public interface ICustodian
      */
     Task<Dictionary<string, decimal>> GetAssetBalancesAsync(JObject config, CancellationToken cancellationToken);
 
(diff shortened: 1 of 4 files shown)


woutersamaey:
> You can probably already know if it is new by inspecting the `config`.
> If it is empty, then you know it's new.

This was my first thought as well, but because we pass data around it is possible to have values in the `JObject` while it is still unsaved.

The only alternative I see is to took into a post-save event and modify the config there, then re-saving it bypassing validation. Can we hook into post-save + avoid an endless loop?

NicolasDorier:
Then just do not make it required, it's just for the fake, we shouldn't add fields lightly to interfaces that will be used by many people for such niche case.

woutersamaey:
Closed in favor of new PR #4861

PR #4864, opened 2023-04-10
Refactor: Dot not make LNURLPaymentMethodDetails depends on BTCPayInvoiceId

NicolasDorier:
Ping @Kukks

If LNURL was activated during the invoice creation with lazy payment, then we would end up with a LNURP which has an incorrect value because `BTCPayInvoiceId` wasn't set.

This PR fixes this by just passing the `InvoiceEntity` when rendering LNURP in AdditonalData, since we always have access to it.

This part of the code requires soon a big refactoring, but that's after 1.9.

Also abstracted `GetPaymentProof` as it doesn't make sense to have a hardcoded switch there.


BTCPayServer/Controllers/GreenField/GreenfieldInvoiceController.cs
Coldman420, line 1: How do I do this

PR #4888, opened 2023-04-17
Fix crash if auto detect language on checkout page, and the language couldn't be detected (Fix #4881)

### LINKED ISSUE #4881: Detect browser language breaks the GET INVOICE STATUS API
0.1.9

 The GET INVOICE STATUS  API  return a 500 error code and a message of "1" when the auto-detect language option is on.    "BTC_PAY_BASE_URL/i/INVOICE_ID/status"

image

If we enable the "auto-detect language" then the GET request to this API fails. if we turn it OFF, it works.
Also, if we pass along the parameters as documented: "lang=auto", it fails

NicolasDorier:
Fix #4881


frances601 (commented):
I haven't had none of these payments ?? I've had nothing at all 😩

Issue #200, opened 2018-06-01
Automatic Withdrawal - Lightning Network to Bitcoin On-Chain


bitcoinshirt:
One of the other payment processors has a feature that allows automatic withdrawal of the LN balance to the on-chain.

Right now our biggest problem is not knowing how much we have off-chain and the inability to appropriately manage those funds without putting the liquidity in danger.

![scheduled-withdrawals-full](https://user-images.githubusercontent.com/36959754/40847090-ecb14176-65aa-11e8-8fd4-155c0e4c7598.png)

They are on the LND, not sure if it's possible for the c-lightning.

NicolasDorier:
what is this payment processor?

NicolasDorier:
mmh this solution is only possible because they are custodian.

The best we can do for BTCPay is to provide clear visualization on channel state so you can balance by yourself and close the channel you want.

PR #4891, opened 2023-04-18
Add missing param, fix response code.

ndeet:
- The parameter `disableCertificateCheck` exists but was not documented
- When sending mail through API the response code is 200 and not as documented 201

Open questions:
1) the field `fromDisplay` seems to have no effect whatsover (it is not set as from display name on the actual email), was this a mistake? On the UI you also only have one field where you can enter mail and display name together?

2) I tried to add new lines (in PHP you add `\n` to the body text but they were not shown in the email, the text was sent as everything in one line. Not sure if this should have worked or not.

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-email.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-email.json
@@ -143,7 +143,7 @@
                     "x-position": 1
                 },
                 "responses": {
-                    "201": {
+                    "200": {
                         "description": "The email was sent (scheduled) successfully"
                     },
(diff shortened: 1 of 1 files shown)


Carolynmaspn:
#200 email ***@***.***
  Screenshot_2023-04-17-12-43-06.png
<https://drive.google.com/file/d/1xl982awKY6ovkEP1LkH5LZfz1FnAse0t/view?usp=drivesdk>

NicolasDorier:
I believe the description should actually be HTML

ndeet:
By description you mean the mail body content supports html? What about the `fromDisplay` it seems to be not used.

PR #4907, opened 2023-04-24
Fix: Form in Payment Requests was not setting its values to the invoices metadata

NicolasDorier:
When creating a new PR with a Form, we would save the form's metadata in the PR's metadata, but wouldn't set the invoice's metadata.


Kukks (approved):

Chixonya22:
Я не могу вывести деньги и не открывается мой Айди

PR #4908, opened 2023-04-24
Prevent an NRE in LNURL

### LINKED ISSUE #4904: Error scanning Print Only POS Scan QR code since v1.9.1 upgrade.

BTCPay Server v1.9.1

Manual deploy

### What happened?

When using Print only option in POS app, scanning the QR causes error in WoZ and Zeus apps.

Choose "Print Display" option in POS app, SAVE and then VIEW, try scan QR code presented.

### Relevant log output

_No response_

### What browser do you use?

Brave

### Additional information

Error regarding decoding QR code.

dennisreimann:
In addition to f05a7f9f14e0a2f906c79951a15311085774db85. Fixes #4904.

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -468,10 +468,9 @@ namespace BTCPayServer
             lnUrlMetadata ??= new Dictionary<string, string>();
 
             var pm = i.GetPaymentMethod(pmi);
-            var paymentMethodDetails = pm?.GetPaymentMethodDetails() as LNURLPayPaymentMethodDetails;
+            var paymentMethodDetails = (LNURLPayPaymentMethodDetails)pm.GetPaymentMethodDetails();
             bool updatePaymentMethodDetails = false;
-            if (lnUrlMetadata?.TryGetValue("text/identifier", out var lnAddress) is true &&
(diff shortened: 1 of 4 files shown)


Kukks (approved):

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -468,9 +468,10 @@ public async Task<IActionResult> ResolveLightningAddress(string username)
             lnUrlMetadata ??= new Dictionary<string, string>();
 
             var pm = i.GetPaymentMethod(pmi);
-            var paymentMethodDetails = (LNURLPayPaymentMethodDetails)pm.GetPaymentMethodDetails();
+            var paymentMethodDetails = pm?.GetPaymentMethodDetails() as LNURLPayPaymentMethodDetails;
```
NicolasDorier, line 471: I don't understand why this one can be null.

dennisreimann: Me neither. I could not reproduce it in dev unfortunately, so no debugger — I can reproduce it on prod though, that's where I got the line that was causing it and I figured from your recent commit, that this might be a way to prevent it.

NicolasDorier:
Ok so I will merge this as is just fixing the NRE.

PR #4911, opened 2023-04-24
Remove LNURLStandardInvoiceEnabled

NicolasDorier:
This PR removes the lightning setting `LNURL enabled for standard invoices`, which was set to `false` by default.

It also adds the ability to set `lazyPaymentMethod` at the invoice level through Greenfield (see notes for the reason it cannot be done in a separate PR).

## Current situation

The motivation for disabling LNURL by default for standard invoice was to avoid complicating the checkout page with both BOLT11 payment and LNURL options.

While BOLT11 can't be used on TopUp invoices, for standard invoices showing two usable QR code to pay on lightning to the user was confusing. LNURL would be a redundant, less supported QR Code.

Recent change made checkout v2 displays only BOLT11 QR code for lightning payment even if LNURL is also possible.
However, Checkout v1 was still showing both options.

## Motivation of this PR

The main issue is that LNURL is now useful for NFC/Print view for PoS.
If this obscure checkbox isn't checked, NFC and print view won't work, since LNURL is disabled for standard invoices.

This PR completely removes the option and updates Checkout v1 to not show the LNURL payment in the displayed payment methods if BOLT11 is available.
Now, in both checkout, LNURL QR code will not be displayed if BOLT11 is available, however it will still be supported by NFC.

Additionally, a number of bugs have been fixed where LNURL wouldn't work well with lazy payment methods, as it wasn't checking whether the payment method was activated.

## Notes

Support for setting `lazyPaymentMethod` at the invoice level has been added because it shouldn't be set when creating an invoice originating from an LNURL request (e.g., for a PoS app print view).

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -671,7 +671,7 @@ namespace BTCPayServer.Controllers
             var displayedPaymentMethods = invoice.GetPaymentMethods().Select(p => p.GetId()).ToList();
 
             // Exclude Lightning if OnChainWithLnInvoiceFallback is active and we have both payment methods
-            if (storeBlob is { CheckoutType: CheckoutType.V2, OnChainWithLnInvoiceFallback: true } &&
+            if (storeBlob is { OnChainWithLnInvoiceFallback: true } &&
                 displayedPaymentMethods.Contains(btcId))
             {
                 displayedPaymentMethods.Remove(lnId);
@@ -787,11 +787,13 @@ namespace BTCPayServer.Controllers
                 Request.Host,
                 Request.PathBase) : null;
 
+            var isAltcoinsBuild = false;
+#if ALTCOINS
+                isAltcoinsBuild = true,
+#endif
+
             var model = new PaymentModel
             {
-#if ALTCOINS
-                AltcoinsBuild = true,
-#endif
                 Activated = paymentMethodDetails.Activated,
                 CryptoCode = network.CryptoCode,
                 RootPath = Request.PathBase.Value.WithTrailingSlash(),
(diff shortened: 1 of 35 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -666,23 +666,23 @@ public async Task<IActionResult> CheckoutNoScript(string? invoiceId, string? id
             var btcId = PaymentMethodId.Parse("BTC");
             var lnId = PaymentMethodId.Parse("BTC_LightningLike");
             var lnurlId = PaymentMethodId.Parse("BTC_LNURLPAY");
-            if (paymentMethodId is null)
+            var enabledPaymentIds = store.GetEnabledPaymentIds(_NetworkProvider).ToList();
+            // Exclude Lightning if OnChainWithLnInvoiceFallback is active and we have both payment methods
+            if (storeBlob is { CheckoutType: CheckoutType.V2, OnChainWithLnInvoiceFallback: true } &&
```
dennisreimann, line 671: I think we should then also remove the `CheckoutType: CheckoutType.V2` part here, so that we unify V1 and V2.

NicolasDorier: btw, there is a bug with this: If the store is checkout v2, doesn't mean the invoice is. The invoice can opt in checkout v1 overriding the store settings.

So we should definitely unify this later.

Kukks (changes requested):
You will need to remove `DisableBOLT11PaymentOption` option in `LightningSupportedPaymentMethod`; else, there will be installs with LN fully disabled for standard invoices.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
dennisreimann: Unified it in [8de55ce](https://github.com/btcpayserver/btcpayserver/pull/4911/commits/8de55cef313ef5cd58f67fb32d73cb5aa57065e7).

dennisreimann:
> You will need to remove DisableBOLT11PaymentOption option in LightningSupportedPaymentMethod; else, there will be installs with LN fully disabled for standard invoices.

Done in [c95f75b](https://github.com/btcpayserver/btcpayserver/pull/4911/commits/c95f75bc6c157da43ec88fd1dc86af780703d03d).

BTCPayServer/Models/StoreViewModels/LightningSettingsViewModel.cs
```diff
@@ -24,13 +24,7 @@ public class LightningSettingsViewModel : LightningNodeViewModel
         [Display(Name = "LNURL Classic Mode")]
         public bool LNURLBech32Mode { get; set; } = true;
 
-        [Display(Name = "LNURL enabled for standard invoices")]
-        public bool LNURLStandardInvoiceEnabled { get; set; } = true;
-
         [Display(Name = "Allow payee to pass a comment")]
         public bool LUD12Enabled { get; set; }
```
Kukks, line 28: Maybe we can remove this too then and have it always enabled?

BTCPayServer/Views/UIInvoice/CheckoutNoScript.cshtml
```diff
@@ -24,13 +25,13 @@
...
         }
-        @if (Model.AvailableCryptos.Count > 1)
+		@if (displayedPaymentMethods.Count > 1)
         {
             <div>
                 <hr />
                 <h2>Pay with:</h2>
                 <ul style="list-style-type: none;padding-left: 5px;">
-                    @foreach (var crypto in Model.AvailableCryptos)
+					@foreach (var crypto in displayedPaymentMethods)
```
Kukks, line 34: nit: weird ident (tabs)

Kukks (approved):
i think it's ok, but we should def test before a public release

BTCPayServer/Models/StoreViewModels/LightningSettingsViewModel.cs
dennisreimann: Concept ACK. I think with the approach we took for Checkout v2 (and now also unified in v1 as of this PR) it makes sense to not bother the user with this.

dennisreimann (approved):

NicolasDorier:
@Kukks 

> You will need to remove DisableBOLT11PaymentOption option in LightningSupportedPaymentMethod; else, there will be installs with LN fully disabled for standard invoices.

This isn't the case. This PR activates LNURL for all invoices even standard.
If BOLT11 isn't available, then the LNURL QR Code is shown.

PR #4914, opened 2023-04-24
Improve invoice filtering UI

### LINKED ISSUE #3664: Improve invoice filtering UI
Currently, invoice filtering is cryptic and people aren't really aware they can toggle the filter.

Screenshot 2022-04-21 at 12 56 28

Some good suggestions were made in this discussion, but TLDR:
- Add a multiselect drop down for store & status filtering (bear in mind user should be able to filter by mutliple stores, statuses).
- instead of typing startDate, endDate -> show 2 datepickers for start and end date;

This PR from 2019 can server as an inspiration, but ideally we should tend towards simplicity. https://github.com/btcpayserver/btcpayserver/pull/771

Another question would be how to make the toggle more obvious since people weren't really aware of these options.

dennisreimann:
Closes #3664.

### DIFF
--- a/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
+++ b/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
@@ -249,6 +249,20 @@
             <a asp-action="ListInvoices" asp-route-storeId="@Model.StoreId" asp-route-count="@Model.Count" asp-route-searchTerm="@Model.Search.Toggle("includearchived", "true")" class="dropdown-item @(HasBooleanFilter("includearchived") ? "dropdown-item-custom-active" : "")" id="StatusOptionsIncludeArchived">Archived</a>
         </div>
     </div>
+    @if (Model.Apps.Any())
+    {
+        <div class="dropdown">
+            <button id="AppOptionsToggle" class="btn btn-secondary dropdown-toggle dropdown-toggle-custom-caret" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+                App
+            </button>
+            <div class="dropdown-menu" aria-labelledby="AppOptionsToggle">
+                @foreach (var app in Model.Apps)
+                {
+                    <a asp-action="ListInvoices" asp-route-storeId="@Model.StoreId" asp-route-count="@Model.Count" asp-route-searchTerm="@Model.Search.Toggle("orderid", app.AppOrderId)" class="dropdown-item @(HasArrayFilter("orderid", app.AppOrderId) ? "dropdown-item-custom-active" : "")">@app.AppName</a>
+                }
+            </div>
+        </div> 
+    }
     <div class="dropdown">
         <button id="DateOptionsToggle" class="btn btn-secondary dropdown-toggle dropdown-toggle-custom-caret" type="button" data-bs-toggle="dropdown" aria-expanded="false">
             Time
--- a/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
+++ b/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
@@ -247,7 +247,7 @@
             }
             else
             {
-                <span>Status</span>
+                <span>All Status</span>
             }
         </button>
         <div class="dropdown-menu" aria-labelledby="StatusOptionsToggle">
@@ -273,7 +273,7 @@
                 }
                 else
                 {
-                    <span>Plugin</span>
+                    <span>All Plugins</span>
                 }
             </button>
             <div class="dropdown-menu" aria-labelledby="AppOptionsToggle">
@@ -307,7 +307,7 @@
             }
             else
             {
-                <span>Time</span>
+                <span>All Time</span>
             }
         </button>
         <div class="dropdown-menu" aria-labelledby="DateOptionsToggle">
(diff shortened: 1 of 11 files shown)


pavlenex (approved):
tACK This is a good cleanup, concept ACK, will leave it up to @dstrukt to comment on UI, but imo this is great solution, good job!

dstrukt (changes requested):
Overall, love it, very clean change!

However, few suggestions:

- [ ] Push the `Export` button (it's disappeared?) next to the Create button
- [ ] Let the elements go full width (sans Export / Actions)
- [x] Rename `Date` to `Time`? Thoughts?
- [x] min-width the `Status` and `Time` buttons to 120px?
- [x] Reduced the left padding (looked better) to 16px vs. 24px?

i.e:
Screen Shot 2023-05-04 at 5 12 43 PM

- [ ] Thoughts on pushing the mass actions to the next line? i.e:
Screen Shot 2023-05-04 at 8 55 51 PM

Do you think we could remove the `Search` button, and instead add `Search...` instead?

Otherwise, this looks great - awesome work! We could easily translate this to other views too.

dennisreimann:
Did the updates: 

grafik


@dstrukt The export/actions buttons are only present when there are results, so I think the first point of that list doesn't apply?

dstrukt:
Ahh yes, it was intented for if >1 invoice, just knew I could show what the search/time/status would look like sans Export/Actions.

I've been wanting to improve the `Actions` button/CTA for a little while, so thought this might be a good time, but if you think it falls outside of the scoe of this PR, feel free to ignore for now.

dennisreimann:
@dstrukt I added the app filter and also indicators to see which filter is active:

grafik

dennisreimann:
@dstrukt Regarding "Export button next to Create Invoice": We are a bit constraint here, because there are actually two HTML forms on this page: One for search/filtering and another one for export/archive. The latter has to include the checkboxes in the list and we cannot move the Export button out of the form.

Something like this would work though, if we move the Export button next to Archive again. This would also make it more logical imho, because we made it so that only the checked invoices get exported — so having a direct connection to the list would make sense for the user as well.

grafik

dstrukt:
Updated two snippets of text (Plugin instead of App) as we're moving to plug-ins, and `Search...` instead of `Search ...`.

Overall it looks fantastic, this is all much easier to parse!

- I still think we should try to address mass actions in a sep. PR. 
- I see where you're coming from with <form> concerns for Export, so perhaps maybe we can add the export as part of the mass action

The current implementation work, but thought/suggestion for "selected" filters:
For the Status and Plugin filters, we could surface the active count to the parent text, e.g.
- 2 Plugin(s) or 1 Plugin
- 3 Status or 1 Status 
But for Time it could just change to the selected range, e.g simply:
- 24 Hours, 3 Days, 7 Days or Custom
Just a thought.

Otherwise, think we're ready to go!

dennisreimann:
> I suppose if you have 5K invoices, and could only select 500, that might be a problem though... 

If there are none selected, everything will be exported.

> suggestion for even more clarity wrt "selected" filters

Implemented it …

grafik

dstrukt:
> If there are none selected, everything will be exported.

Ahh, good to know, and something to consider with the mass action mock proposal, in its current form, in order to see the options at least 1 result needs to be selected.

> Implemented it …

Looks great in the screenshot! will review this evening .. what do you think of it?

dennisreimann:
> what do you think of it?

I think it gives even more overview, so good suggestion and full ACK 👌

dstrukt (approved):
LGTM!

Only change we could consider the new filtering update @dennisreimann is `All Status` `All Plugins` `All Time` as the default states, but I realize this adds more horizontal space .. so could be something we could address when we tackle Export / Actions stuff, but very minor.

BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
```diff
@@ -200,83 +234,111 @@
...
+    {
+        <div class="dropdown">
+            <button id="AppOptionsToggle" class="btn btn-secondary dropdown-toggle dropdown-toggle-custom-caret" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+                @if (appFilterCount > 0)
+                {
+                    <span>@appFilterCount Plugin@(appFilterCount > 1 ? "s" : "")</span>
+                }
+                else
+                {
+                    <span>Plugin</span>
```
dennisreimann, line 276:
```suggestion
                    <span>All Plugins</span>
```

BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
```diff
@@ -200,83 +234,111 @@
...
+    <input asp-for="SearchText" class="form-control" placeholder="Search…" />
+    <div class="dropdown">
+        <button id="StatusOptionsToggle" class="btn btn-secondary dropdown-toggle dropdown-toggle-custom-caret" type="button" data-bs-toggle="dropdown" aria-expanded="false">
+            @if (statusFilterCount > 0)
+            {
+                <span>@statusFilterCount Status</span>
+            }
+            else
+            {
+                <span>Status</span>
```
dennisreimann, line 250:
```suggestion
                <span>All Status</span>
```

BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
```diff
@@ -200,83 +234,111 @@
...
+                    <span>7 Days</span>
+                }
+                else
+                {
+                    <span>Custom</span>
+                }
+            }
+            else
+            {
+                <span>Time</span>
```
dennisreimann, line 310:
```suggestion
                <span>All Time</span>
```

dennisreimann:
> All Status All Plugins All Time as the default states

Applied it :)

dstrukt:
LGTM!

PR #4916, opened 2023-04-25
Receipt: Optimize print view

### LINKED ISSUE #4902: [Bug]: When printing receipt on Thermal printer, Powered By BTCPay Logo doesn't print.

1.9.1

Lunanode

### What happened?

Printed receipt on PoS device and thermal printer. 
Looking at the receipt there's small issues, one of them is the Powered By Logo is to light gray to actually be printed. 

Print receipt on thermal printer. 

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

As you can see there's also something with alignment of text, however this seems to be different on every printer. So might just be settings thing, or we could maybe center it? Don't know for sure. 

Action would be to make the Powered by logo darker at print? 

dennisreimann:
Fixes #4902. I also tried to do more general improvements, but to me it seems the device has its own rules when formatting things: It doesn't respect spacings or alignments set via CSS, it cuts text and the outcome seems rather random.

image


Kukks (approved):

dstrukt:
Wow, some nightmare formatting unfortunately.

Only section that really bothers me is the item list table stuff, but this gets the job done and it seems pretty futile to keep tweaking.

dstrukt (approved):

BTCPayServer/Views/Shared/_StoreFooterLogo.cshtml
```diff
@@ -1 +1,2 @@
...
+<span class="d-none d-print-inline">BTCPay Server</span>
```
NicolasDorier, line 2: why is there HTML in the SVG?

dennisreimann: It's a partial, the first line is the SVG, the second HTML.

PR #4940, opened 2023-05-01
Store Custom Roles

Kukks:


BTCPayServer/Extensions/StoreExtensions.cs
```diff
@@ -1,11 +1,40 @@
 #nullable enable
 using System.Linq;
+using BTCPayServer.Client;
 using BTCPayServer.Data;
 
 namespace BTCPayServer
 {
     public static class StoreExtensions
     {
+        public static StoreRole? GetStoreRoleOfUser(this StoreData store, string userId)
```
NicolasDorier, line 10: All of those need to be `async` and not use non async versions of EF.

BTCPayServer.Data/Data/UserStore.cs
```diff
@@ -9,14 +10,18 @@ public class UserStore
 
         public string StoreDataId { get; set; }
         public StoreData StoreData { get; set; }
-        public string Role { get; set; }
+        [Column("Role")]
+        public string LegacyRole { get; set; }
```
NicolasDorier, line 14: We should just create two pre defined role in the DB with the permissions.

BTCPayServer.Data/Data/StoreRole.cs
```diff
@@ -0,0 +1,51 @@
...
+                .HasForeignKey(e => e.StoreDataId)
+                .OnDelete(DeleteBehavior.Cascade)
+                .IsRequired(false);
+            
+            entity.HasIndex(entity => new {entity.StoreDataId, entity.Role}).IsUnique();
+        });
+        
+        
+        
+        if (!databaseFacade.IsNpgsql())
```
NicolasDorier, line 38: Should be JSONB if Npgsql

BTCPayServer/Extensions/StoreExtensions.cs
Kukks: This is already done elsehwere. This extension method is simply fetching the roles a user has on a loaded store

BTCPayServer.Data/Data/UserStore.cs
Kukks: It's already done

BTCPayServer.Data/Data/StoreRole.cs
Kukks: It is a `text[]` column when postgres, else it does not type mapper for everything else

dstrukt (approved):
LGTM!

I think the sections could be differentiated a bit more .. mostly to make it easier to quickly scan the high level section.

Maybe a little bit of word-smithing, but it looks great!

BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -2460,6 +2462,133 @@ public async Task CanUseLNURLAuth()
...
+            Assert.Equal(3, existingServerRoles.Count);
+            IWebElement ownerRow = null;
+            IWebElement guestRow = null;
+            foreach (var roleItem in existingServerRoles)
+            {
+                if (roleItem.Text.Contains("owner", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    ownerRow = roleItem;
+                   
+                }else if (roleItem.Text.Contains("guest", StringComparison.InvariantCultureIgnoreCase))
```
dennisreimann, line 2483:
```suggestion
                }
                else if (roleItem.Text.Contains("guest", StringComparison.InvariantCultureIgnoreCase))
```

BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -2460,6 +2462,133 @@ public async Task CanUseLNURLAuth()
...
+            s.FindAlertMessage();
+            
+            existingServerRoles = s.Driver.FindElement(By.CssSelector("table")).FindElements(By.CssSelector("tr"));
+            foreach (var roleItem in existingServerRoles)
+            {
+                if (roleItem.Text.Contains("owner", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    ownerRow = roleItem;
+                   
+                }else if (roleItem.Text.Contains("guest", StringComparison.InvariantCultureIgnoreCase))
```
dennisreimann, line 2509:
```suggestion
                }
                else if (roleItem.Text.Contains("guest", StringComparison.InvariantCultureIgnoreCase))
```

BTCPayServer/Models/StoreViewModels/CreateStoreViewModel.cs
```diff
@@ -21,5 +21,8 @@ public class CreateStoreViewModel
         public string PreferredExchange { get; set; }
 
         public SelectList Exchanges { get; set; }
+
+        [Display(Name = "Preferred Price Source")]
+        public string RoleId { get; set; }
```
dennisreimann, line 26: Why is this in here? It does not get used in the view.

BTCPayServer/StorePolicies.cs
```diff
@@ -1,19 +1,21 @@
...
-            }
-        }
+        // public static IEnumerable<String> AllRoles
+        // {
+        //     get
+        //     {
+        //         yield return Owner;
+        //         yield return Guest;
+        //     }
+        // }
```
dennisreimann, line 19:
```suggestion
```

BTCPayServer/Controllers/UIUserStoresController.cs
```diff
@@ -69,7 +69,7 @@ public async Task<IActionResult> CreateStore(CreateStoreViewModel vm)
             blob.DefaultCurrency = vm.DefaultCurrency;
             blob.PreferredExchange = vm.PreferredExchange;
             store.SetStoreBlob(blob);
-            await _repo.CreateStore(GetUserId(), store);
+            await _repo.CreateStore(GetUserId(), store, vm.RoleId);
```
dennisreimann, line 72: The `vm.RoleId` doesn't get set from the view.

NicolasDorier:
There is nothing preventing an attacker to make a HTML request with the policy `btcpay.server.canmodifyserversettings`?

NicolasDorier:
The "Search by role" and pagination isn't really tested. IMHO it should be completely remove, the number of roles in those page will never reach a number that need pagination.

dstrukt:
> The "Search by role" and pagination isn't really tested. IMHO it should be completely removed, the number of roles in those page will never reach a number that need pagination.

Great point, tend to agree.

Will do some search testing as well.

Kukks:
> I think it would be better if `RoleData` had two column: `StoreId` and `Id`. Where `Id` instead of being a random generated horrible GUID, would be the name of the role. Primary key would be (StoreId, Id). A system wide role would have an empty string StoreId (because all NULLs are different in SQL)

I actually tried this initially but it was an issue with having storeid/role as an index for some reason.  and when you do not use generated ids, the entity tracker gets confused and makes logic messier

Kukks:
> The "Search by role" and pagination isn't really tested. IMHO it should be completely removed, the number of roles in those page will never reach a number that need pagination.

While true, I wanted to try and at least stay using one unified approach across the board for these list entities.

Kukks:
> There is nothing preventing an attacker to make a HTML request adding policy `btcpay.server.canmodifyserversettings`? There need some validation that the policies added exists and that they are store's policies.

See latest commit

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -81,50 +81,57 @@ public PermissionSet ToPermissionSet(string storeId)
...
         }
 
-        public async Task<string> GetDefaultRole()
+        public async Task<StoreRoleId> GetDefaultRole()
         {
-            return (await _settingsRepository.GetSettingAsync<PoliciesSettings>())?.DefaultRole!;
+            var r = (await _settingsRepository.GetSettingAsync<PoliciesSettings>())?.DefaultRole;
+            if (r is not null)
+                return new StoreRoleId(r);
+            return StoreRoleId.Owner;
```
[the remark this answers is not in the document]
NicolasDorier: It can't if it is the default.

Issue #4695, opened 2023-02-23
Add UI to `Custom Forms`


pavlenex:
Currently the way to create a custom form requires you to know JSON. We should add the UI so users can easier select the fields they want in the form. This would require a joined effort of devs and designers to figure out simple, yet effective interface.

![Screenshot 2023-02-23 at 1 00 29 PM](https://user-images.githubusercontent.com/36959754/220899935-0270f604-1be0-42c1-ad0d-d7af76fecad5.png)

dstrukt:
Few Questions & Thoughts:

1. Do we still plan to support JSON input?
2. If so, would it be in addition to, or more of a binary? i.e. you choose the forms or the JSON

3. With regard to scope, are we simply create form UI for the templates for this iteration, or are we attempting to create a a flexible form builder? We touched on the idea of a form builder, but I wanted to determine involved this iteration would be, as it can get quite complex very quickly. Can design for either, just want to check with you @dennisreimann before I proceed.

4. The `Allow form for public use` switch is a little confusing in it's current form, and I think we should try to reword it to make it a little more clear.

dennisreimann:
It will be a builder for the forms JSON, so we can look at it similar to the item/perks editor of the POS/Crowdfund apps. I'd say we still offer the "advanced mode" of editing the JSON manually (see the "raw template" option of the apps), this will also help us with debugging (at least in the first version of the new builder).

Basically we need a rather simple version to add and configure the form fields line by line. We don't need all the fancy features other form builders have, we should start with something simple but support reordering of the fields. For the option per field/line you can take a look at the [Forms documentation](https://docs.btcpayserver.org/Forms/).

The `Allow form for public use` switch enables a standalone mode, with which the form can be the entry/start for triggering an invoice. Similar to a pay button, but with additional information. I think one can also directly provide the amount. 

/cc @kukks

PR #4942, opened 2023-05-02
Forms: Add multiline input

dennisreimann:
Up to now we don't have a way to accept multiline input values via the forms. The input type text strips the newlines and replaces them with whitespace, so there wasn't even a way to solve #4941 correctly.

This add the textarea form provider and modifies the PosData partial, so that it properly displays multiline values.

Closes #4941.


BTCPayServer/Views/Shared/PosData.cshtml
```diff
@@ -26,13 +28,13 @@
                     }
                     else
                     {
-                        @value?.ToString()
+                        @Safe.Raw(Regex.Replace(str.Trim(), @"(\r\n)|\n|\r", "<br/>"))
```
dennisreimann, line 31: @NicolasDorier OK like this?

learntheropes: Guys, sorry for asking, but I don't understand anything of this programming language.

I just want to make sure that the replacement happens only in the rendered field and does not manipulate the data actually stored in the metadata property, which are correct at the moment.

If we recall the invoice, we should find the same text we have added when we created it.

Kukks: This seems excessively complicated, I think we should do as suggested with `white-space: pre-wrap;` styling

Kukks (changes requested):

BTCPayServer/Views/Shared/PosData.cshtml
dennisreimann: Done in [5501b91](https://github.com/btcpayserver/btcpayserver/pull/4942/commits/5501b916897bfd63489260979d01cae558ba003e).

dstrukt (approved):
LGTM!

At least visually / functionally.

BTCPayServer/Views/Shared/Forms/TextareaElement.cshtml
```diff
@@ -0,0 +1,28 @@
+@model BTCPayServer.Abstractions.Form.Field
+@{
+	var isInvalid = ViewContext.ModelState[Model.Name]?.ValidationState is Microsoft.AspNetCore.Mvc.ModelBinding.ModelValidationState.Invalid;
+	var errors = isInvalid ? ViewContext.ModelState[Model.Name].Errors : null;
+}
+<div class="form-group">
+    <label class="form-label" for="@Model.Name"@(Model.Required ? " data-required" : "")>
+        @Model.Label
+    </label>
+    <textarea id="@Model.Name" type="@Model.Type" class="form-control @(errors is null ? "" : "is-invalid")"
```
dennisreimann, line 10:
```suggestion
    <textarea id="@Model.Name" class="form-control @(errors is null ? "" : "is-invalid")"
```

dennisreimann:
@Kukks Your feedback has been addressed. Can we merge it so that I can use the textarea in #4695?

Kukks (approved):

Issue #4889, opened 2023-04-17
[Bug]: standard Tea shop (PoS app) first item breaks after going from shop + cart to just shop 


Bas02:
### What is your BTCPay version?

BTCPay Server v1.9.0 and BTCPay Server v1.9.1

### How did you deploy BTCPay Server?

Luna node standard deployment

### What happened?

After trying to figure out the new receipt situation, i switched between shop + cart to just a shop without cart.

Since this swap, the first item in the shop list brings a 404 page. 

### How did you encounter this bug?

Create a PoS shop. 

Switch to a Product list + cart. 
Save
Buy product 1 in the list ( or at least mark as settled on that invoice) 
Switch to just Product list. 
Save

it seems however, it breaks only in the same view i had open before. 
If you reopen the view from the PoS it seems fine?

### Additional information

petzsch and i were going through some motions of the shop and encountered this bug. 
Petzsch runs 1.9.1 and i'm running on the 1.9.0 version. 

![image](https://user-images.githubusercontent.com/102448109/232529657-ad9e8663-6bff-43d7-b076-78e6cb5e3507.png)

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

petzsch:
Problem does not occur when Hard-Reloading (Ctrl - Shift - R in Firefox) the page after changing from product list with cart to product list.

Also doesn't occur when closing the shop tabs and re-opening them through the "View" button from the backend.

So it could be a caching issue. Perhaps something with LocalStorage? 404 (NotFound) seems to be the kind of [standard reply](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs#L198) if something went wrong on this PoS.

NicolasDorier:
I can't reproduce this.

1. Create a new PoS with cart view
2. Buy an item, see the receipt, go back to PoS view
3. Change the PoS to product via
4. F5 on the PoS view page

Then the first item works. That said, the HTML @petzsch  show isn't right, but I can't reproduce it.

NicolasDorier:
I tried the other way around, where I create a pos without cart, pay and then switch on the pos with cart.
It works.

dennisreimann:
I cannot reproduce this either.

dennisreimann:
I tested in Firefox

Bas02:
[...]

PR #4950, opened 2023-05-04
Resolving weird Firefox form autocomplete behavior on our POS

rockstardev:
Fix: #4889 

Firefox has _specific_ autocomplete behavior on forms that was causing problems by replacing values, as described in the issue referenced above. Here is screen recording of the behavior and how it updates HTML.

https://user-images.githubusercontent.com/5191402/236257578-fd56430e-5c17-4068-bbf6-8c1c71d6e8d5.mp4


dennisreimann (approved):
Wow, this is weird and I've never come across Firefox altering/autocompleting a hidden input before.

NicolasDorier:
wtf

PR #4951, opened 2023-05-04
Improve create first store case

### LINKED ISSUE #3900: Make "Create your store" CTA on new store more prominent
A small UX improvement when creating a new store on a fresh BTCPay server would be to have a much more prominent call to action (CTA) in the centre of the screen that really drives that user towards taking the action they should be taking next. Small example here (could also remove the button on the left hand menu as it's not necessary with the large, more prominent, CTA button.

image

dennisreimann:
As discussed in the design meeting:

- Home redirects to the Create Store page if there are no stores yet
- Enhances the copy on that page for the first store case
- Removes the  redundant Create Store CTA button in the sidebar

Closes #3900.

grafik

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -180,7 +180,7 @@ namespace BTCPayServer.Tests
             {
                 Driver.FindElement(By.Id("StoreSelectorToggle")).Click();
             }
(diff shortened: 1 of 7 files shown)


pavlenex:
Concept ACK. Good idea.

dstrukt (approved):

Kukks (approved):

PR #4966, opened 2023-05-08
Hide sensitive info

### LINKED ISSUE #3779: Hide dashboard balance / hide sensitive data (lurking wife mode)
Currently, our dashboard flashes your balance. It's a good privacy practice to allow users to optionally hide the wallet balance. I think similar has to be applied to the charts in the dashboard.

This may relate to #3752, as we can allow toogle between a currency, btc and hide balance.

Bitcoin Design Guide has some good practice on how to hide the balance  https://bitcoin.design/guide/daily-spending-wallet/privacy/#quickly-hiding-balances

dennisreimann:
Adds the `data-sensitive` attribute, which replaces the content of that HTML tag with `***` when the "Hide Sensitive Data" mode is on. As this is a feature that only works on the frontend level, the mode is persistsed in the local storage, similar to the chosen color mode.

Closes #3779.

hide-sensitive-info

### DIFF
--- a/BTCPayServer/Views/Shared/LayoutHeadTheme.cshtml
+++ b/BTCPayServer/Views/Shared/LayoutHeadTheme.cshtml
@@ -25,6 +25,7 @@ else
 {
     <link href="~/main/themes/default.css" asp-append-version="true" rel="stylesheet" />
     <link href="~/main/themes/default-dark.css" asp-append-version="true" rel="stylesheet" id="DarkThemeLinkTag" />
+    <script>if (window.localStorage.getItem('btcpay-hide-sensitive-info') === 'true') { document.documentElement.setAttribute('data-hide-sensitive-info', 'true')}</script>
     <script src="~/js/theme-switch.js" asp-append-version="true"></script>
     <noscript><style>.btcpay-theme-switch { display: none !important; }</style></noscript>
 }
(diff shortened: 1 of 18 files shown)


dstrukt:
Oh hell yea! Now everyone down at the coffee shop won't going to be able to see my 0.001 bitcoin stack activity 😂

OP looks great, will get a review in later this evening!

dstrukt (approved):
Don't have anything to add here, think you cover all the places that are relevant.

LGTM!

BTCPayServer/Views/Shared/LayoutHeadTheme.cshtml
```diff
@@ -25,6 +25,7 @@ else
 {
     <link href="~/main/themes/default.css" asp-append-version="true" rel="stylesheet" />
     <link href="~/main/themes/default-dark.css" asp-append-version="true" rel="stylesheet" id="DarkThemeLinkTag" />
+    <script>if (window.localStorage.getItem('btcpay-hide-sensitive-info') === 'true') { document.documentElement.setAttribute('data-hide-sensitive-info', 'true')}</script>
```
Kukks, line 28: not: maybe move this to where the logic is too?

Kukks (approved):

BTCPayServer/Views/Shared/LayoutHeadTheme.cshtml
dennisreimann: It has to be right at the top, to ensure the attribute is set before the content is rendered/shown. If this would occur at the bottom where the JS is included, the hidden data would flash in cleartext on subsequent page loads.

Issue #4982, opened 2023-05-17
[Bug]: BTCPay Server: Top 10 Mistakes in Bitcoin Financial Application Code


onlineapps-cloud:
### What is your BTCPay version?

last version

### How did you deploy BTCPay Server?

from source code

### What happened?

found 10 errors

### How did you encounter this bug?

long to explain

### Additional information

hi on habr.com, guys from pvs-studio analized btc pay server code and found 10 erros, in link you will find all relevand info. best regards.
https://habr.com/ru/companies/pvs-studio/articles/735726/

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

pavlenex:
Thanks for notifing us about this! We're on it, fixing in #4983

NicolasDorier:
Seems a good static analyzer, if we could integrate that to our CI we would do

PR #4983, opened 2023-05-18
Fix a bunch of minor bugs

NicolasDorier:
This fix bugs reported on https://pvs-studio.com/en/blog/posts/csharp/1051/

There isn't really anything critical.
However, I think we should definitely use some static analyzer. I'd like to use those analyzer in CI if possible later.
Maybe we can contact them to get access to it, it should be on CI.

Fix https://github.com/btcpayserver/btcpayserver/issues/4982

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.cs
@@ -133,7 +133,7 @@ namespace BTCPayServer.Controllers.Greenfield
                     "A valid node info was not provided to open a channel with");
             }
 
-            if (request.ChannelAmount == null)
+            if (request?.ChannelAmount is null)
             {
                 ModelState.AddModelError(nameof(request.ChannelAmount), "ChannelAmount is missing");
             }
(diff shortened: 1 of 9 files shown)


BTCPayServer/HostedServices/BitpayIPNSender.cs
```diff
@@ -261,7 +261,6 @@ public Task StartAsync(CancellationToken cancellationToken)
                 {
                     if (e.Name == InvoiceEvent.Expired ||
                        e.Name == InvoiceEvent.PaidInFull ||
-                       e.Name == InvoiceEvent.FailedToConfirm ||
```
NicolasDorier, line 264: They suspect it should have been `PaidAfterExpiration`. This is maybe true, but if true, nobody ever complained about it, and this is the legacy API... so let's not touch it and let it fade away.

BTCPayServer/Controllers/GreenField/GreenfieldLightningNodeApiController.cs
```diff
@@ -133,7 +133,7 @@ public virtual async Task<IActionResult> OpenChannel(string cryptoCode, OpenLigh
                     "A valid node info was not provided to open a channel with");
             }
 
-            if (request.ChannelAmount == null)
```
NicolasDorier, line 136: A bunch of NRE could have happen

Kukks (approved):

Issue #4980, opened 2023-05-17
[Feature]: Not creating new address for partial payments


NicolasDorier:
We have reports that some customer and support people are confused when a partial payment happen because the address change.

I think this is such a corner case that we shouldn't make a new address. It's marginally bad for privacy, but will simplify support for merchants.

gutek2:
Hello I did not notice the change of address Thanks for the info

PR #4984, opened 2023-05-18
Do not generate new address when a new payment is detected

NicolasDorier:
When a bitcoin address is used for a payment, it was being switched with a new address.

We got feedback of customers and merchants very confused about it, and thinking they got scammed.
I believe we should just reuse addresses. This isn't good for privacy, but this is a corner case that almost never happen. It isn't really worth bothering support with this.

Fix https://github.com/btcpayserver/btcpayserver/issues/4980

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -2154,6 +2154,7 @@ namespace BTCPayServer.Tests
                 txFee = localInvoice.BtcDue - invoice.BtcDue;
                 Assert.Equal("paidPartial", localInvoice.ExceptionStatus.ToString());
                 Assert.Equal(1, localInvoice.CryptoInfo[0].TxCount);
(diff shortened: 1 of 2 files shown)


BTCPayServer.Tests/UnitTest1.cs
```diff
@@ -2154,7 +2154,6 @@ public async Task InvoiceFlowThroughDifferentStatesCorrectly()
                 txFee = localInvoice.BtcDue - invoice.BtcDue;
                 Assert.Equal("paidPartial", localInvoice.ExceptionStatus.ToString());
                 Assert.Equal(1, localInvoice.CryptoInfo[0].TxCount);
```
Kukks, line 2156:
```suggestion
                Assert.Equal(1, localInvoice.CryptoInfo[0].TxCount);
                Assert.Equal(localInvoice.BitcoinAddress, invoice.BitcoinAddress); //Same address
```

Kukks (approved):

PR #4992, opened 2023-05-22
Add What's New in v1.10.0

dennisreimann:
grafik

### DIFF
--- a/BTCPayServer/Views/UIStores/Dashboard.cshtml
+++ b/BTCPayServer/Views/UIStores/Dashboard.cshtml
@@ -34,7 +34,7 @@
             </div>
             <div class="modal-body">
                 <h5 class="alert-heading">Updated in <a href="https://blog.btcpayserver.org/btcpay-server-1-10-0/" target="_blank" rel="noreferrer noopener">v1.10.0</a></h5>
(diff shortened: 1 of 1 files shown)


BTCPayServer/Views/UIStores/Dashboard.cshtml
```diff
@@ -34,17 +34,17 @@
                 </button>
             </div>
             <div class="modal-body">
+                <h5 class="alert-heading">Updated in <a href="https://blog.btcpayserver.org/btcpay-server-1-10-0/" target="_blank" rel="noreferrer noopener">v1.10.0</a></h5>
+                <p class="mb-2">We've added some more options for refunds and an editor for creating custom forms. A propos custom: You can now also define your own roles with customized access to store features.</p>
```
Bas02, line 38:
```suggestion
                <p class="mb-2">We've added some more options for refunds and an editor for creating custom forms. You can now also define your own roles with customized access to store features.</p>
```

Bas02 (changes requested):
Couldn't figure out what you tried to say with what i took out.

dstrukt:
`A propos custom` is less common in the states, usually see `with regard to`, or `speaking of...`

Might consider dropping, or editing, but just my 2sats, otherwise, looks great!

dennisreimann:
Yup, @Bas02 already took that out, good to have two second pairs of eyes on it 😁

pavlenex (approved):

PR #4999, opened 2023-05-25
Pluginify on chain wallet setup

Kukks:
This PR fixes a few logical points in the wallet setup flow to allow more extensive plugin flexibility; It also fixes an issue when building plugins that requires an Altcoin config profile. Here is an example showcasing the Liquid+ plugin using this to enforce that it is a hot wallet (a requirement it has) and that import to RPC is always set, and a new option that is used to configure the wallet further https://i.imgur.com/pDPQ73v.gif

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.Onchain.cs
+++ b/BTCPayServer/Controllers/UIStoresController.Onchain.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -250,7 +251,7 @@ namespace BTCPayServer.Controllers
                 CryptoCode = cryptoCode,
                 Method = method,
                 SetupRequest = request,
-                Confirmation = string.IsNullOrEmpty(request.ExistingMnemonic),
+                Confirmation = !isImport,
                 Network = network,
                 Source = isImport ? "SeedImported" : "NBXplorerGenerated",
                 IsHotWallet = isImport ? request.SavePrivateKeys : method == WalletSetupMethod.HotWallet,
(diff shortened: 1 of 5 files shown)


BTCPayServer/Views/UIStores/_LayoutWalletSetup.cshtml
```diff
@@ -19,3 +19,5 @@
 }
 
 @RenderBody()
+
+<vc:ui-extension-point location="onchain-wallet-setup-post-body" model="@Model"/>
```
dennisreimann, line 23: This adds it to all pages of the wizard process, which might be a bit unspecific. Shouldn't we rather introduce more specialized extension points and put them on the individual pages/inside the specific forms?

Kukks: > This adds it to all pages of the wizard process, which might be a bit unspecific. Shouldn't we rather introduce more specialized extension points and put them on the individual pages/inside the specific forms?

This is actually great, as one hook allows me to modify the entire flow from one place (especially because they all use the same ViewModel)

Kukks: https://github.com/Kukks/BTCPayServerPlugins/blob/master/Plugins/BTCPayServer.Plugins.LiquidPlus/Views/Shared/OnChainWalletSetupLiquidExtension.cshtml

Kukks: yeah, cant do much, but also dont want to add too much friction to core development

dennisreimann: Fair enough.

BTCPayServer/Controllers/UIStoresController.Onchain.cs
```diff
@@ -270,6 +271,8 @@ public async Task<IActionResult> GenerateWallet(string storeId, string cryptoCod
             GenerateWalletResponse response;
             try
             {
+                if(request.AdditionalOptions is not null)
```
Kukks, line 274:
```suggestion
				//this is required because AdditionalOptions comes from NBXPlorer and is a readonly dictionary, so we overwrite it
                if(request.AdditionalOptions is not null)
```

PR #5020, opened 2023-05-30
prevent adjusting a role that would result in store deletion

Kukks:


NicolasDorier:
Created https://github.com/btcpayserver/btcpayserver/pull/5025 instead.
I don't really like passing errors as tuple Go-like, also, I don't think we should stop users updating roles for this reason, because it might be caused by rogue or unused stores.

My other PR just remove the cleanup of unreachable store. Let's address this when we have some data rentention feature.

PR #5029, opened 2023-06-01
Payment Type Registry

Kukks:
Part of a major effort to abstract the payment system to allow plugins to provide their own payment methods


NicolasDorier:
Played a bit on alternative, but I think this is the right approach.

Only thing I would change.

* Renaming `PaymentType` to `PaymentTypeHandler`
* Removing all singletons
* Renaming `PaymentTypeRegistry` to `PaymentTypeHandlers`
* Directly inject `PaymentTypeHandler` in DI. Both as their base type, and concrete type, so payment type specific controllers can inject that directly.

Maybe can be done in additional PR or this one?

Kgomib:
How do I withdraw

Kukks:
obsolete through invoice payment refactoring

PR #5033, opened 2023-06-02
Support NFC on modal

### LINKED ISSUE #4251: [Bug]: WebNFC Doesn't Work In iFrames

I'm running BTCPayServer v1.6.12.0 and LNURL NFC extension v1.0.5.0.

Manual

### What happened?

The NFC plugin doesn't work when the invoice checkout is loaded in a modal (like with the crowdfunding app). This is happening because the WebNFC spec restricts its usage to top-level contexts. 

Here's a screenshot from a debugging session on an Android device:

photo_2022-10-30 07 59 41

I think we should set `this.supported` to `'NDEFReader' in window && window.self === window.top` in `lnurlwnfc.js`. I can make a PR if the is the desired change.

1. Create a Crowdfund app
2. Make a donation using this app
3. When the invoice modal loads up, click the "Pay by NFC" button
4. Try and tap a NFC tag
5. The tag tap isn't captured by the browser, but the host device

### Relevant log output

_No response_

### What browser do you use?

Chrome v106

### Additional information

_No response_

Kukks:
fixes #4251

### DIFF
--- a/BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml
+++ b/BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml
@@ -22,7 +22,38 @@
     </template>
 </template>
 <script type="text/javascript">
-// https://developer.chrome.com/articles/nfc/
+class NDEFReaderWrapper {
+    constructor() {
+        this.onreading = null;
+        this.onreadingerror = null;
+    }
+
+    async scan( opts) {
+        
+        if (opts && opts.signal){
+            opts.signal.addEventListener("abort", () => {
+                window.parent.postMessage("abortNfc", "*");
+            });
+        }
+        window.parent.postMessage("startNfcScan", "*");
+    }
+}
+
+window.addEventListener("message", (event) => {
+    const { action, data } = event.data;
+    if (action === "nfcData") {
+        if (ndef.onreading) {
+            ndef.onreading({ message: { records: [{ data }] } });
+        }
+    } else if (action === "nfcError") {
+        if (ndef.onreadingerror) {
+            ndef.onreadingerror();
+        }
+    }
+});
+
+
+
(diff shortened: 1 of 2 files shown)


dennisreimann (approved):
@Kukks Can you review my refactoring? I changed a few things with the message handling, it should be working now :)

BTCPayServer/Views/Shared/NFC/CheckoutEnd.cshtml
```diff
@@ -166,25 +150,42 @@ Vue.component("lnurl-withdraw-checkout", {
...
+                if (inModal) {
+                    // receive messages from iframe
+                    window.addEventListener('message', async event => {
+                        // deny messages from other origins
+                        if (event.origin !== window.location.origin) return
+                        
+                        const { action, data } = event.data
+                        switch (action) {
+                            case 'nfc:data':
+                                await this.sendData(data)
```
dennisreimann, line 180: This previously sent data to the `onreading` function (`ndef.onreading({ message: { records: [{ data }] } });`), which led to the data being decoded twice: The data we recedive here is already decoded, hence it should call `sendData` directly.

pavlenex:
@Kukks @dennisreimann is there something we can do to help you guys here, or is it just a matter of code refactoring that's left? Unsure if testing is needed, would be a bit tricky to test maybe in dev env.

Kukks:
> @Kukks @dennisreimann is there something we can do to help you guys here, or is it just a matter of code refactoring that's left? Unsure if testing is needed, would be a bit tricky to test maybe in dev env.

The only thing we need is to find a way to test this

dennisreimann:
I tested it in dev mode at least so far that the NFC card could get read and the state is the same as without a modal. I don't have a regtest setup for LNURL Withdraw via NFC card, but that part works the same as before.

Kukks:
Going to merge and we will test on rc

PR #5041, opened 2023-06-04
Pull Payment: Support LNURL Withdraw with SATS denomination

dennisreimann:
Currently the LNURL Withdraw feature is only active for pull payments denominated in BTC. This PR extends it to the SATS denomination as well.

sats-denominated


154313203002499:
Zaplatil som blockhain a žiadnu výplatu som nedostal.

PR #5061, opened 2023-06-10
Swap crypto amount for fiat amount in checkout v2 header

leesalminen:
The #1 piece of user feedback we've received about the point of sale checkout flow was to display the fiat amount prominently. Normies don't care about the sat amount due, they care about the fiat amount due. This change shows the fiat amount above the fold, and interested users can click View Details to view the sat amount due.


dennisreimann:
Concept nACK: The user does not pay fiat though, so the main thing to display is still the bitcoin amount. In early mocks we had the fiat equivalent as a secondary/muted subline in addition to the bitcoin amount — might be worth exploring that again. @dstrukt @pavlenex

dstrukt:
Agree with @dennisreimann here. The primary amount should always be display in BTC/sats, with the secondary amount being the fiat equivalent at the point of sale.

As for where this is displayed it could be placed directly under the BTC amount, if we feel it deserves the real estate, which there's certainly a case for this level of important, and what was in the original amount OR tucked into the View Details, making it a little more hidden.

Issue #3872, opened 2022-06-15
Improved Payment Requests List View


dstrukt:
Proposal:
- Consolidating Actions: `View` -> Name/ID column (https://github.com/btcpayserver/btcpayserver/pull/3873)
- `Edit` or `View` as the Primary action 
- "..." for all additional actions, this pattern could be applied elsewhere
- Pushing `Status` to either a tab menu filter (Top) or a filterable dropdown (Bottom), both have their pros and cons. Pull Payments & Payouts use the tabbed structure, Invoices use the dropdown filter structure, so we should make a call here, keep different for each, or try to unify for consistent functionality across.
- Improving the `Status` tags of the Payment Requests
- Simplify `Search` input component

<img width="889" alt="Screen Shot 2022-06-14 at 11 02 44 PM" src="https://user-images.githubusercontent.com/6250771/173754840-69b302a0-7c73-4f77-b031-0e0b8c859ccb.png">

PR #5065, opened 2023-06-11
Payment Requests: List view improvements

TChukwuleta:
#3872 

Improved Payment Requests List View fix


myezanebo:
Help me please about how I withdraw my account.phone number 27 0825415835.

pavlenex:
Thanks for picking up on this issue @TChukwuleta! 

It seems when you click on the payment request link it leads to edit, instead of view. The concern here is that users may not immediately figure out that three dots abstract something as simple as being able to view a payment request.

@dstrukt did you envision that edit link edits payment request, and the name link leads to viewing it or is edit a correct direction here? Should view and `...` be different actions? Also with introduction of this, we'd have to add it to pull payments to ensure consistency in the UI.


https://github.com/btcpayserver/btcpayserver/assets/36959754/b8237bcb-9021-49ed-938d-7c999cc7b8fe

TChukwuleta:
Hello @pavlenex 

Yea, the previous implementation was actually an edit. I think making it a 'view' is good for a better user experience as against 'edit'
I'd update that.

pavlenex:
Gotcha. 

> I'd update that. In that case I'd also remove 'view' from the dropdown

Let's keep it there for now, until people get used to it, may not be obvious for people yet, and we also need to consolidate other views, then we can optimize them all and remove view from dropdown.

TChukwuleta:
Hello @pavlenex  this has been updated

dennisreimann (approved):
tACK

dennisreimann:
@dstrukt you can review this now.

dstrukt (approved):
tACK!

Great work @TChukwuleta!

Agree with @pavlenex wrt Edit/View, if we get feedback that the current iteration is more difficult or confusing, we can always adjust as we work through all of the views.

Apologies my review took so long!

PR #5070, opened 2023-06-13
Do not crash when an invoice have an amount that is too big

NicolasDorier:
User reported

```
fail: PayServer:      Unhandled error on watching invoice CUvkPDYMaxJUeEi8eG7dN8
System.OverflowException: Value was either too large or too small for an Int64.
   at System.Decimal.ToInt64(Decimal d)
   at NBitcoin.Money.Coins(Decimal coins)
   at BTCPayServer.Services.Invoices.PaymentMethod.Calculate(Func`2 paymentPredicate)
   at BTCPayServer.HostedServices.InvoiceWatcher.GetNearestClearedPayment(PaymentMethodDictionary allPaymentMethods, PaymentMethodAccounting& accounting) in /source/BTCPayServer/HostedServices/InvoiceWatcher.cs:line 213
   at BTCPayServer.HostedServices.InvoiceWatcher.UpdateInvoice(UpdateInvoiceContext context)
   at BTCPayServer.HostedServices.InvoiceWatcher.StartLoop(CancellationToken cancellation) in /source/BTCPayServer/HostedServices/InvoiceWatcher.cs:line 320
```

I can reproduce.

My fix is just to catch such exception and return max values so it doesn't crash. (Calls to `Calculate` aren't expecting exceptions)
Those exception only happen if a too high amount is set on the invoice, as such I think it is safe to just return garbage which doesn't throw in this case.

I also prevent a value that is too big in Greenfield. If we call `CreateInvoiceRaw` without validating and the value is too big, I just clamp the value to the max amount `ulong.Max`.

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -39,6 +39,7 @@ using BTCPayServer.Plugins.PayButton;
 using BTCPayServer.Plugins.PointOfSale;
 using BTCPayServer.Plugins.PointOfSale.Controllers;
 using BTCPayServer.Security.Bitpay;
+using BTCPayServer.Security.Greenfield;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Apps;
 using BTCPayServer.Services.Invoices;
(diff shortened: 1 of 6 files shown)


dennisreimann (approved):

NicolasDorier:
Will add some check preemptively to make sure the values are sane in API and UI.

PR #5078, opened 2023-06-16
Add category feature to the PoS with Cart

NicolasDorier:
This add a way for merchants to group items in the cart view.

You can add one or several categories to items in the PoS:

image

Then, you can easily filter it in the PoS

image

Ping @dennisreimann for small brush up: I think the labels aren't exactly aligned with the checkbox. (Maybe I overthink it?)

Also design get weird when there is a single item in the category:

image

https://github.com/btcpayserver/btcpayserver/discussions/2624
https://github.com/btcpayserver/btcpayserver/discussions/4468
https://github.com/btcpayserver/btcpayserver/discussions/4597
https://github.com/btcpayserver/btcpayserver/issues/5088

@Kx29PM @yannick10800 @Zaxounette

### DIFF
--- a/BTCPayServer/Views/Shared/TemplateEditor.cshtml
+++ b/BTCPayServer/Views/Shared/TemplateEditor.cshtml
@@ -90,6 +90,11 @@
                                 <label class="form-label">Description</label>
                                 <textarea rows="3" cols="40" class="form-control mb-2" v-model="editingItem.description" ref="txtDescription"></textarea>
                             </div>
+							<div class="form-group">
+								<label class="form-label">Groups</label>
+								<input type="text" placeholder="drinks, tea" required pattern="[^\*#]+" class="form-control mb-2" v-model="editingItem.groups" ref="txtgroups" />
+								<div class="form-text">Easily filter the different items using groups, used only in the product list with cart.</div>
+							</div>
                             <div class="form-group">
                                 <label class="form-label">Inventory</label>
                                 <input type="number" inputmode="numeric" min="0" step="1" class="form-control mb-2" v-model="editingItem.inventory" ref="txtInventory" />
@@ -174,6 +179,10 @@ document.addEventListener("DOMContentLoaded", function () {
                     this.editingItem = {index:-1, id:"", title: "", price: 0, image: "", description: "", priceType: "Fixed", inventory: null, disabled: false};
                 }else{
                     this.editingItem = {...this.items[index], index};
+					if (this.editingItem.groups)
+					{
+						this.editingItem.groups = this.editingItem.groups.join(", ");
+					}
                 }
                 
                 this.editingItem = this.unEscape(this.editingItem);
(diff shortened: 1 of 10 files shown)


dennisreimann (approved):
Done, nice and simple, good job!

grafik

pavlenex:
I tested this one out, good work, very simple solution.

Some feedback:

- I still think calling this `categories` instead of `groups` - product category is pretty common term. 
- It doesn't seem we somehow save `groups` or categories. So we expect owner to keep them in their head. Imagine having 20 categories, and when you add a new product you'd have to remember the exact name of group/category. We should try to re-use how we approach tags in addresses?
- Bear in mind, product buttons will be used on different devices, and we should ensure they are easy to tap on, to avoid fat finger. Can you check how Tiankii stylied them here, it seems less prone to fat-finger error with more padding around them and better alignment to make them quite prominent and easy to distinguish. 
- 
246678291-0d1e7da7-643d-4a35-9125-4b171a16657b

NicolasDorier:
Renamed it to categories

NicolasDorier:
Added a category selector @dennisreimann do you think you need to brush it up?

image

NicolasDorier:
@pavlenex the buttons that Dennis changed are still prone to fat finger issues?

pavlenex:
> @pavlenex the buttons that Dennis changed are still prone to fat finger issues?

I think we can add a bit more padding and ensure they're all spaced out properly, I'll let @dennisreimann decide on this one, perhaps @dstrukt has some ideas.

dennisreimann:
Updated the spacing and refactored the Template Editor. We can now use TomSelect for the categories.

grafik

dstrukt:
Looking great - will get a proper review in here in the next hour!

First thoughts:

Preferred @dennisreimann's original implementation with the pills for the end-user filter.

The admin-facing UI looks/works great.

BTCPayServer/Views/Shared/TemplateEditor.cshtml
```diff
@@ -131,7 +131,7 @@
         <div class="col-xxl-constrain">
             <div class="form-group pt-3">
                 <label for="@Model.templateId" class="form-label">Template</label>
-                <textarea id="@Model.templateId" name="@Model.templateId" rows="10" cols="40" class="form-control" v-model="configJSON" v-on:change="updateFromJSON">@Model.template</textarea>
+                <textarea id="@Model.templateId" name="@Model.templateId" rows="10" cols="40" class="form-control" v-model="configJSON" v-on:change="updateFromJSON"></textarea>
```
NicolasDorier, line 135: @dennisreimann if `@Model.templateId` is controlled by attacker, he can inject. Why is it a property at all? there isn't need of "templateId" I think.

NicolasDorier: I checked. This is actually fine for properties, as view only bind an attribute if it starts by `:`. This would have been problematic if it was in the content of the element though.

grosblanc (approved):

NicolasDorier:
@dstrukt 

> Preferred @dennisreimann's original implementation with the pills for the end-user filter.

What do you mean "Preferred"? It's what is already done no?

dennisreimann:
@dstrukt Can you give this a final review?

dstrukt (approved):
Sorry I misunderstood, I thought we were opting for square filter buttons vs. the pill filters. @NicolasDorier 

LGTM! 

Have some more thoughts, but we can follow up with on the future POS PR!

PR #5079, opened 2023-06-16
Fix: Applying a discount in PoS with cart wasn't working

NicolasDorier:
This fix a bug somebody reported me in prague.
We were calculating the price to pay client side, then testing server side if the amount wasn't less than expected.
However, due to the discount, the amount was less than expected and the route didn't create an invoice.

## How to reproduce

* Create a PoS with card
* Select an item
* Add a discount
* Attempt to pay

## Expected result

A new invoice should show up

## Actual result

You get redirected to the PoS page, no invoice is created

## Proposed fix

The solution is to calculate the price to pay server side with the tip and discount amount.


dennisreimann (approved):

Kukks (approved):

Brooksie18:
That’s awesome thanks so much

Brooksie18 (approved):
I approve these changes.

PR #5080, opened 2023-06-16
Add extension point to template editor

Kukks:
This allows you to extend the items with arbitrary fields. In my case it is a file selector to offer a file upon payment


Winsway88:
WinsBaby

PR #5081, opened 2023-06-16
Make file management UI more useful

Kukks:
Before

After

### DIFF
--- a/BTCPayServer/Views/UIServer/Files.cshtml
+++ b/BTCPayServer/Views/UIServer/Files.cshtml
@@ -42,6 +42,53 @@
                 </form>
             }
 
+            @if (Model.DirectUrlByFiles is { Count: > 0 })
+            {
+                foreach (var fileUrlPair in Model.DirectUrlByFiles)
+                {
+                    var fileId = fileUrlPair.Key;
+                    var file = Model.Files.Single(storedFile => storedFile.Id.Equals(fileId, StringComparison.InvariantCultureIgnoreCase));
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):
Nice one!

pavlenex:
Feedback we received wrt this one is:

1. If I have 100 photos, newly uploaded one is below the fold, have to know that it's on the bottom
2. When I want to add photo to POS app and copy the file, it doesn't work (absolute vs relative paths)

@Kukks any chance we can resolve these two as well?

dennisreimann:
> When I want to add photo to POS app and copy the file, it doesn't work (absolute vs relative paths)

@NicolasDorier and I talked about this, I'll add direct file upload in the POS. No need to consider it here then.

pavlenex:
@dennisreimann Makes sense, will log as a separate issue, should we still tackle above the fold vs below the fold item in this PR?

Kukks:
> > When I want to add photo to POS app and copy the file, it doesn't work (absolute vs relative paths)
> 
> @NicolasDorier and I talked about this, I'll add direct file upload in the POS. No need to consider it here then.

How do you intend to do this considering the item editor is a js enhancement?

dennisreimann:
@pavlenex good point, moved it to the top:

grafik

> How do you intend to do this considering the item editor is a js enhancement?

I'll check that and will continue that topic in #5086.

pavlenex:
Thanks @dennisreimann this was exactly what Paralelni Polis cashier asked us to accommodate. cACK

PR #5083, opened 2023-06-17
Lightning: Relax GetInfo constraint for LNDhub connections

### LINKED ISSUE #4482: [Bug]: LNBank and LNDHub (via lnbits extension) Stopped to work

I'm running BTCPayServer 1.7.13

Running on MynodeBtc via Docker

### What happened?

After upgrade to version 1.7.13 all stores with wallets using LNBits LNDHub Extension or LNBank stopped to work. 

When you create a store, and setup a lightning wallet to use LNBank wallet or a LNBits wallet (via LNDHUb extension), BTCPayServer shows the node as offline and it can't generate invoices. 

image

No Idea how to fix. I think is a bug! 

Any help is welcome!  

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
The LNDhub-compatible implementation by LNbits does not support the `GetInfo` call for all their funding sources — see lnbits/lnbits#1182. By catching that exception in combination with the `LndHubLightningClient`, we give people the ability to still use their LNbits-based LNDhub as a Lightning node.

Talked about this with @Kukks and @callebtc at BTCPrague, because people keep asking for a workaround. Fixes #4482.

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -48,7 +48,7 @@
     <PackageReference Include="YamlDotNet" Version="8.0.0" />
     <PackageReference Include="BIP78.Sender" Version="0.2.2" />
     <PackageReference Include="BTCPayServer.Hwi" Version="2.0.2" />
-    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.4.26" />
+    <PackageReference Include="BTCPayServer.Lightning.All" Version="1.4.27" />
     <PackageReference Include="CsvHelper" Version="15.0.5" />
     <PackageReference Include="Dapper" Version="2.0.123" />
     <PackageReference Include="Fido2" Version="2.0.2" />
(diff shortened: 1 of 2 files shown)


NicolasDorier:
@dennisreimann A `NullReferenceException` is normally a bug, can you update the lightning lib to send a `NotSupportedException` instead?

NicolasDorier:
By `NotSupportedException` I mean, you need to catch where the `NullReferenceException` is thrown, and instead check for null and send the `NotSupportedException` and catch this.

dennisreimann:
Yup, I like the way that Nicolas suggested. Will update.

dennisreimann:
Updated it. Depending in the LNbits version, the endpoint is either not supported or returning incomplete data — both cases are handled now and lead to the node getting registered, but without public addresses.

grafik

massmux:
Just to confirm that the setup that finally works is this one

BTCPay Server v1.11.7+a921504bc 

LNBits ver 0.11.1 lndhub plugin 0.3.2

PR #5100, opened 2023-06-20
Ajaxify the wallet transaction list to avoid timeout (Fix #4987)

### LINKED ISSUE #4987: [Bug]: Do not throw error 500 if the transaction list of wallet is timing out

1.9.3

docker

### What happened?

A big wallet timeout on the wallet page.

Have a big wallet, then go to the transaction list of the wallet, see an error 500

We should have a proper fallback in this situation. There are some pattern of wallet that may timeout like this, it is caused because we do not paginate based on the date.

Another idea would just to ajaxify the page. @dennisreimann do you think we can put that in next release?

### Relevant log output

```shell
fail: Microsoft.AspNetCore.Server.Kestrel: Connection id "0HMQG63U0LBDI", Request id "0HMQG63U0LBDI:00000002": An unhandled exception was thrown by the application.
Npgsql.NpgsqlException (0x80004005): Exception while reading from stream
 ---> System.TimeoutException: Timeout during reading attempt
   at Npgsql.Internal.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|215_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at Npgsql.NpgsqlDataReader.<ReadMessage>g__ReadMessageSequential|44_0(NpgsqlConnector connector, Boolean async)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command) in /_/Dapper/SqlMapper.Async.cs:line 418
   at BTCPayServer.Services.Wallets.BTCPayWallet.FetchTransactionHistory(DerivationStrategyBase derivationStrategyBase, Nullable`1 skip, Nullable`1 count, Nullable`1 interval) in /source/BTCPayServer/Services/Wallets/BTCPayWallet.cs:line 246
```

### Workaround

1. Sweep your wallet by browsing directly to `https://yourserver/wallets/<walletid>/send`
2. Create a new emtpy wallet by browsing `https://yourserver/stores/<storeid>/onchain/BTC/settings`

NicolasDorier:
Some wallets could timeout on this page. This is problematic as it removes any UI path to reach the wallet settings or the send wallet UI.

To fix this:
1. The HTML returned when browser request the page isn't querying transactions anymore
2. When the page load, we start infinite scrolling. This trigger a call to `loadMoreTransactions` as soon as the page is loaded.
3.  I removed the `Load More` button that appeared before. This only role of this button was to start infinite scrolling to load the transactions asynchronously. However, this PR start infinite scrolling automatically after page load.

Fix #4987

### DIFF
--- a/BTCPayServer.Tests/Extensions.cs
+++ b/BTCPayServer.Tests/Extensions.cs
@@ -126,7 +126,7 @@ retry:
         {
             var wait = new WebDriverWait(driver, SeleniumTester.ImplicitWait);
             wait.UntilJsIsReady();
-            wait.Until(d => d.ExecuteJavaScript<bool>("return window.areTransactionsLoaded;"));
+            wait.Until(d => d.WaitForElement(By.CssSelector("#WalletTransactionsList[data-loaded='true']")));
         }
         public static IWebElement WaitForElement(this IWebDriver driver, By selector)
         {
--- a/BTCPayServer.Tests/Extensions.cs
+++ b/BTCPayServer.Tests/Extensions.cs
@@ -122,6 +122,12 @@ retry:
             driver.ExecuteJavaScript($"document.getElementById('{element}').{funcName}()");
         }
 
+        public static void WaitWalletTransactionsLoaded(this IWebDriver driver)
+        {
+            var wait = new WebDriverWait(driver, SeleniumTester.ImplicitWait);
+            wait.UntilJsIsReady();
+            wait.Until(d => d.ExecuteJavaScript<bool>("return window.areTransactionsLoaded;"));
+        }
         public static IWebElement WaitForElement(this IWebDriver driver, By selector)
         {
             var wait = new WebDriverWait(driver, SeleniumTester.ImplicitWait);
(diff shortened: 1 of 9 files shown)


dennisreimann (approved):
tACK, though I think this is a slight degradation in terms of complexity and also UX.

NicolasDorier:
@dennisreimann why do you feel so?
We deleted more line that we added. Also, it avoid the user to have to push a button.

dennisreimann:
I think it's a combination of …

- we don't use this "initial load with AJAX" pattern elsewhere – even on the dashboard we have at least the first tile present on load. 
- we made the edge case fallback the default and imho the way we were doing it like before worked well for the standard use case. 
- there's not a good way to recover from a loading failure other than reloading the page. We could add that, but it would make the frontend code more complex. 

No hard feelings, not really trying to change the approach, just expressing my thoughts.

PR #5101, opened 2023-06-20
Do not create an invoice on every lnurl query

Kukks:


NicolasDorier:
What problem are you trying to solve? Why is it a problem lnurl query create invoice?

dennisreimann:
@NicolasDorier it spams the invoices list and wastes resources.

grafik

Kukks:
> What problem are you trying to solve? Why is it a problem lnurl query create invoice?

I have tens of thousands of expired invoices created this way. it's incredibly spammy at this point because of how nostr clients work

pavlenex:
I can confirm this is an issue with Zaps, literally tens of thousands of invoices are generated. Let's hope this change doesn't break anything for LNURL payments though.

Kukks:
> I can confirm this is an issue with Zaps, literally tens of thousands of invoices are generated. Let's hope this change doesn't break anything for LNURL payments though. My workaround thus far was having a separate store for zaps.

Making it seamless and that nothing changes in terms of usage

NicolasDorier:
I'd like to understand, how does the Zap work? which link get hit?

Kukks:
> I'd like to understand, how does the Zap work? which link get hit?

It hits the ln address endpoint. Some nostr clients pre-emptively hit it as soon as they see a post by a new user, and fetch that user's profile, which has a link to an ln address. They then hit the ln address to see if it works. Then a user may want to "zap" a post by a user, and physically click on a zap button, which triggers then flow again

NicolasDorier:
So in all case, a lightning invoice is created and there is no way to prevent this right?

The only thing you want to prevent is the creation of invoice in BTCPay?

Kukks:
> So in all case, a lightning invoice is created and there is no way to prevent this right?
> 
> The only thing you want to prevent is the creation of invoice in BTCPay?

No, I don't want to generate a btcpay invoice unless the user signals an amount to pay (which also generates a lightning invoice through the btcpay invoice). Btcpay invoices should still be created but not on every query

dennisreimann:
Can you explain why the changes around `ExplicitRateRules` are necessary? I don't get it from reviewing the code.

Kukks:
> Can you explain why the changes around `ExplicitRateRules` are necessary? I don't get it from reviewing the code.

If you intend to generate an invoice of 10USD, and the price is 1USD = 10BTC, the lnurl request will show the minsendable to be 100BTC. Now the user agrees to these terms and initiates the payment action, which creates the invoice then. In the meantime however the price has changed to 1USD  = 12BTC, meaning the minsendable is now 120BTC, erroring the request out as it no longer fits the original lnurl request.

dennisreimann:
Ok, thanks! The `CanUseLNAddress` test failure is related to the changes.

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -545,19 +618,57 @@ PaymentMethodId GetLNUrlPaymentMethodId(string cryptoCode, Data.StoreData store,
...
+            
+            InvoiceEntity i;
+            try
+            {
+               var store = await _storeRepository.FindStore(lnurlReq.StoreId);
+                
+                lnurlReq.CreateInvoice.Checkout ??= new InvoiceDataBase.CheckoutOptions();
+                lnurlReq.CreateInvoice.Checkout.LazyPaymentMethods = false;
+                lnurlReq.CreateInvoice.Checkout.PaymentMethods = new[] {  lnurlReq.PaymentMethodId.ToStringNormalized() };
+                i = await _invoiceController.CreateInvoiceCoreRaw(lnurlReq.CreateInvoice, store, Request.GetAbsoluteRoot(), lnurlReq.AdditionalTags);
```
dennisreimann, line 645: @Kukks The LN Address info gets lost here — this is also why the test fails. In d4828f8d0edb0878ed1cca5cbafc061b3a46e2f0 I adapted some things and also updated the tests to check that the BTCPay invoices get created only for the subsequent LNURL requests, however the LN Address info doesn't make it from here to the invoice and I couldn't figure out how to achieve it.

NicolasDorier:
This solution might not work. Everytimes there is a LNUrl generated, you are saving stuff in RAM. Somebody with lots of follower will end up OOM very quickly.

NicolasDorier:
I think I found a better way: Do not create the invoice in the `ResolveLightningAddress`. Instead, craft the LNURLRequest manually with callback having `address=lnaddress@fwfw.com`.

Then only call `GetLNURLRequest` on the Callback` route.

Kukks:
How do you guarantee the rate then? The user can just modify the callback
and lie on the rate

NicolasDorier:
I propose https://github.com/btcpayserver/btcpayserver/pull/5117 instead that is more simpler.

Kukks:
> I propose https://github.com/btcpayserver/btcpayserver/pull/5117 instead that is simpler.
> This doesn't change the logic for the rest of LNUrl, focusing on our problem with Nostr and LNAddress instead.
> 
> @Kukks with my PR, I don't lie on the rate, the rate is calculated when the callback get hit.

That's not good, because the lnurl request has a min/max in btc, and if the currency is not btc, you will potentially have a different min / max at the callback.

Think of the static print use case on the pos app

NicolasDorier:
yes, @Kukks but this doesn't happen in the case of ln addresses.
min/max is set up by the ligthning address blob.

Kukks:
But my solution is meant for all lnurl invoice generation scenarios not
just Ln Address

NicolasDorier:
Why do we need a solution to all lnurl invoice? The rest is working fine.

Kukks:
It is the identical issue where spam Invoices are created upon scanning the
qr code to check it out but without intent to pay

NicolasDorier:
Nobody complained about that and it is unlikely to provoke spam at the level of the nostr issue.

The fact that the LNURL request is locking on the rate only for a short period of time is exactly as it should be. Else, a malicious party could just play the system by extending the rate longer than it should be.

I am also not sure about adding the ability to set explicit rate at the API level. Invoice creation is a very limited permission, but as soon as you allow the modification via explicit rate, then you give the right for invoice creator to basically set any invoice as settled. (they can just set an explicit rate saying 1 BTC = 0.01 USD)

Also, I'd like to fix the nostr issue in the minor release. This PR scope is wider than just a bug fix.

Kukks:
> Nobody complained about that and it is unlikely to provoke spam at the level of the nostr issue. It's also not proper to LNUrl. When somebody click to pay something on the PoS, it creates an invoice even if the user decides not to pay and it has never been an issue.
> 
> The fact that the LNURL request is locking on the rate only for a short period of time is exactly as it should be. Else, a malicious party could just play the system by extending the rate longer than it should be.
> 
> I am also not sure about adding the ability to set explicit rate at the API level. Invoice creation is a very limited permission, but as soon as you allow the modification via explicit rate, then you give the right for invoice creator to basically set any invoice as settled. (they can just set an explicit rate saying 1 BTC = 0.01 USD)
> 
> Also, I'd like to fix the nostr issue in the minor release. This PR scope is wider than just a bug fix.

I 

> Nobody complained about that and it is unlikely to provoke spam at the level of the nostr issue. It's also not proper to LNUrl. When somebody click to pay something on the PoS, it creates an invoice even if the user decides not to pay and it has never been an issue.
> 
> The fact that the LNURL request is locking on the rate only for a short period of time is exactly as it should be. Else, a malicious party could just play the system by extending the rate longer than it should be.
> 
> I am also not sure about adding the ability to set explicit rate at the API level. Invoice creation is a very limited permission, but as soon as you allow the modification via explicit rate, then you give the right for invoice creator to basically set any invoice as settled. (they can just set an explicit rate saying 1 BTC = 0.01 USD)
> 
> Also, I'd like to fix the nostr issue in the minor release. This PR scope is wider than just a bug fix.

The issue spans beyond just ln address - one can set their nostr profile/notes to point to an arbitrary lnurl -- such as the app items, and this issue would be back then.

I updated the PR, the tests are working, and there is a check on invoice creation via greenfield to prevent explicit rate usage if the key does not have CanModifyInvoices

> Also, I'd like to fix the nostr issue in the minor release. This PR scope is wider than just a bug fix.
If we can't agree on this, we should not rush it. This issue has been around for months now.

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -545,19 +619,58 @@ PaymentMethodId GetLNUrlPaymentMethodId(string cryptoCode, Data.StoreData store,
...
             var i = await _invoiceRepository.GetInvoice(invoiceId, true);
+            return await GetLNURLForInvoice(i, cryptoCode, amount, comment);
+        }
+        
+        [NonAction]
+        private async Task<IActionResult> GetLNURLForInvoice(InvoiceEntity i, string cryptoCode,
+            [FromQuery] long? amount = null, string comment = null)
+        {
+            if (!NetworkSupportsLightning(cryptoCode, out var network))
+                return null;
```
NicolasDorier, line 672: `return NotFound()`

grosblanc:
﻿

pavlenex:
https://github.com/btcpayserver/btcpayserver/pull/5117

PR #5107, opened 2023-06-21
Support LNURL in pay button

Kukks:
Needs some UI love but what this does is:
* shows a QR code for the pay button link and HTML they can copy for show the qr code on their website (acts as a link too)
* adds LNURL link option, which would work directly with qr scanner to lightning wallet flows (and adds previous point feature to it)

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -443,28 +443,37 @@ namespace BTCPayServer
         }
 
 
-        [HttpGet("pay")]
+        [HttpGet("{storeId}/pay")]
         [EnableCors(CorsPolicies.All)]
         [IgnoreAntiforgeryToken]
         public async Task<IActionResult> GetLNUrlForStore(
             string cryptoCode,
             string storeId,
-            string currencyCode = null)
+            string currency = null, 
+            string orderId = null, 
+            decimal? amount = null)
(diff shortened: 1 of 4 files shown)


pavlenex:
What's the use-case here? Nobody requested this afair?

Kukks:
> What's the use-case here? Nobody requested this afair?

I was fixing some code in #5101 and ran into a piece of code that allowed direct pay button to lightning wallet, but was never exposed in the pay button UI. This exposes it, and also gives a quality-of-life improvement by giving the user HTML to copy when using the link format (users always asked how to do the qr code and I just told them to use a qr code generator with the link)

pavlenex:
I am not sure pay button is a proper place to have a static QR code. And we already have one, so it looks a bit cluttered now imo. We'd have to discuss it.

Kukks:
> I am not sure pay button is a proper place to have a static QR code. And we already have one, so it looks a bit cluttered now imo. We'd have to discuss it.

That one is added in this PR. Like I said, I have not polished it off, it needs to be in our standard toggle ux style

dennisreimann:
Adapted the UI to use the same patterns we have in other places.

grafik

Also moved it to the bottom, next to the generated code.

dennisreimann (approved):

dstrukt (approved):
LGTM!

PR #5111, opened 2023-06-22
Fix: The current preimage of a invoice's lightning payment method should be available via API

NicolasDorier:
We are serializing payment method details in two places: Payments blob and inside the Invoices blob.

Sadly, we aren't using the NBitcoin's serializer for the payment method detail in the Invoices blob. resulting in the unuseful

```json
{
  "PaymentHash": { "Size": 32 }
}
```

Being saved in the database. Notice the casing is wrong.

We will really need to refactor everything concerning payment methods and payment methods details very soon.


Kukks (approved):
Damn so we have not had any of this data properly persisted?

NicolasDorier:
@Kukks we actually had the data, but at the payment level (we receive preimage when we receive the payment), not invoice level.

dennisreimann (approved):

NicolasDorier:
Giving up the other PR, need complete refactor of the PaymentType stuff, it's too big. Will probably just push this PR.

PR #5113, opened 2023-06-23
Checkout v2: Play sound when invoice is paid

### LINKED ISSUE #5085: Add Invoice settled sound
We discussed in Prague that having some sort of sound feedback once the invoice is settled is very helpful for UX, but also just addictive. We did something similar  for crowdfunding.

Tasks

- [ ] Explore a few different sounds we could use @dstrukt
- [ ] Have a discussion if we want one orginal sound (perhaps even @webwworthy can help with this one) or allow customization 
- [ ] Implement a sound in the checkout
- [ ] Discuss if we want sounds for different states or just settled.

dennisreimann:
Adds the option to upload an audio file that gets played once a payment is made. Closes #5085.

grafik


pavlenex:
Is there a default sound?

pavlenex:
Sure yes, but I think we need to have a default sound, if we  have a simple unique sound it can also become the part of the brand. I'm okay with having option for custom sounds.

dennisreimann:
Agreed, we have the code part in place already — @dstrukt and @webwworthy can take over and come up with a btcpay'y sound :musical_note:

NicolasDorier:
When we have a default sound, we should use it for the "Retail preset"

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -475,6 +476,58 @@ public async Task<IActionResult> CheckoutAppearance(CheckoutAppearanceViewModel
...
+                }
+                else if (!model.SoundFile.ContentType.StartsWith("audio/", StringComparison.InvariantCulture))
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = "The uploaded sound file needs to be an audio file";
+                }
+                else
+                {
+                    var formFile = await model.SoundFile.Bufferize();
+                    /* TODO: Check if we can detect audio 
+                    if (!FileTypeDetector.IsAudio(formFile.Buffer, formFile.FileName))
```
NicolasDorier, line 498: This is important to do, got many vulnerabilities if not the case in the past.

NicolasDorier:
Pick up my commit https://github.com/btcpayserver/btcpayserver/commit/f2a034741096739eedfe52dbfab5d825f5ec12e0

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -475,6 +476,58 @@ public async Task<IActionResult> CheckoutAppearance(CheckoutAppearanceViewModel
...
+
+            if (model.SoundFile != null)
+            {
+                if (model.SoundFile.Length > 1_000_000)
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = "The uploaded sound file should be less than 1MB";
+                }
+                else if (!model.SoundFile.ContentType.StartsWith("audio/", StringComparison.InvariantCulture))
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = "The uploaded sound file needs to be an audio file";
```
NicolasDorier, line 492: Need to use the modelstate for that rather than global messages like this.

dennisreimann:
Integrated a default sound (to be replaced? ;)) and the option to customize it by uploading another sound file.

Peek 2023-06-30 12-10

BTCPayServer/Controllers/UIStoresController.cs
dennisreimann: Done in [5c4ae19](https://github.com/btcpayserver/btcpayserver/pull/5113/commits/5c4ae19b114159951178635494f8e14256c0f720).

dennisreimann:
> When we have a default sound, we should use it for the "Retail preset"

Enabled the default sound by default for new stores in both cases.

webwworthy:
Got a few ideas started for a default sound :) there are sound clips to listen too in the link below. While I was working on these I was wondering could if we give them a few default sounds to pick from? Maybe when you click the check box to play sound on payment a list of choices comes up inbetween the "play sound on payment" and custom file input? @dennisreimann @dstrukt @pavlenex  

5x separate mp3's in this folder:
https://drive.google.com/drive/folders/14wIb97h6U0XDnCfmVOb61KFOf3aRGZuF?usp=sharing

pavlenex:
Good work @webwworthy, personally I like the simplest one, "default sound bels".

dennisreimann:
tbh I don't think there's an ideal one there yet. To me they seem to dry and imho we should have something a bit more cheerful, especially since this can be used with the confetti. My 2 sats …

> I was wondering could if we give them a few default sounds to pick from?

We could, but imho it'd be good to have one iconic default sound.

pavlenex:
I am with @dennisreimann on having a default "icon" sound. It could become a part of the brand!

webwworthy:
for sure! @dennisreimann @pavlenex I can work on those to make some thing a bit brighter and more interesting. That is a good note to match the confetti, I will keep that in mind working on the next version 👍

NicolasDorier:
I don't have stong opinion, but I believe by default the sounds should be off. And the retail preset should activate sound.

webwworthy:
@dennisreimann @pavlenex round 2 of default sounds to review, these still might not be one epic default sound yet and will keep working on them.  https://drive.google.com/drive/folders/1f3Cg-dfyAaDniUSH6kNW_OkSa-y3WCCm?usp=sharing

dennisreimann:
The forth one (hoorm + chime) is a bit too dark/oomphy imho, but it goes in a good direction. A version which is a bit more cheerful would be nice. It could also be a bit longer, maybe aim for three seconds — to fit the start of the grafitti animation, which in whole takes about five seconds.

webwworthy:
I made some new sounds and kept them simple this round. The sounds I like the most don't have the "_alt" extension. Those files are another option for each one. There is also a clip of all the runner-ups that are additional alternatives.

Success, Error, NFC sounds are all in this folder:
https://drive.google.com/drive/folders/1eBAlnrGQ5Gfaa13Of085u93yLir8w18k?usp=sharing

dennisreimann:
I've now prepared everything so that only the sound files need to be changed to whatever gets chosen. The three files are located in the `BTCPayServer/wwwroot/checkout-v2` directory.

dennisreimann:
Updated the label wording to better indicate, that the setting enables all sounds, while the custom file is only for successful payments.

grafik

pavlenex:
tACK for implementation. I haven't tested the NFC sounds but they're available. 

We still have to agree on the sounds, but @dennisreimann made it easy to replace these sounds so whoever takes over this par can easily do it, since Dennis will be afk.

dennisreimann:
You'll for sure need to replace the default success sound 😂

pavlenex:
@webwworthy It's up to you, I still kinda feel we need something very simple and mimic credit card payments. I would not use complicated coin sounds, it's just too much. Do you have any more ideas or should we vote on existing ones? To me suggested success sound is problematic, sounds more of an error to me, but could be just me?

webwworthy:
Here is a clip all of the simple success sounds I've made with out any coin sound effects:
https://drive.google.com/file/d/1lMgBz0Au5gc-VEfQdTmLVvGyr4RiscvM/view?usp=sharing

I think someone should choose one of those four choices. It doesn't have the be the end all be all default sound forever and there was a simple "bells only" clip you liked early on, that is the first sound in this new success sounds clip. My vote is the current success sound or other than that the "bells only" sound would be my preferred second choice for now.

webwworthy:
link to all 4 sounds as individual .mp3's:
https://drive.google.com/drive/folders/1vgTJbfH54f72vguPBlMpwN84uiKaD8rt?usp=sharing

pavlenex:
Success-1.Mp3 is dope! I vote for that one.

PR #5117, opened 2023-06-24
Fix: Exposing ln address on Nostr should not result in lots of invoice being created

NicolasDorier:
### Motivation

When using ln address on Nostr, lot's of invoices get created because of clients resolving the lightning addresses pre-emptively without user interaction.

image

## Solution

Do not create BTCPay invoice when the ln address get resolved. Instead create the invoice only if the callback get called.

This supersede https://github.com/btcpayserver/btcpayserver/pull/5101
This achieve the same result, without changing a lot of code.

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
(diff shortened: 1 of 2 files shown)


BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -396,6 +436,10 @@ public async Task<IActionResult> ResolveLightningAddress(string username)
                {
                    { "text/identifier", $"{username}@{Request.Host}" }
                });
+            if (result is not OkObjectResult ok || ok.Value is not LNURLPayRequest payRequest)
+                return result;
+            var invoiceId = payRequest.Callback.AbsoluteUri.Split('/').Last();
```
Kukks, line 441: this is kind of scary

Kukks (approved):
I'll compromise on my PR for this because these changes are atomic (we can update the flow in a newer version without breaking usage for users)

Issue #5054, opened 2023-06-07
[Bug]: For POS Plugin, Receipt Contains Redundant  OrderID Link to POS App


yuriygeorge:
### What is your BTCPay version?

I'm running BTCPay 1.10.0

### How did you deploy BTCPay Server?

manual

### What happened?

I'm running BTCPay v1.10.0
I have created a POS Plugin App - For example (https://btcpay.kapitalex.com/apps/2kXpfKAZfBxzU3J34wmmQfjfD9T5/pos)

And generated an invoice, and the invoice has been paid. When I send the receipt to the buyer:
https://btcpay.kapitalex.com/i/3aYoYz8R9jmsFTqjgASpVF/receipt

It contains an Order ID, automatically generated by BTCPay, with a link to the POS app.
I think this is a bug. Because I don't want the buyer to have access/Link to the Store's POS plugin.

Also, this Order ID is copied over and over again for all invoices. So even if it was some Order ID, it is not uniquely generated. 
We would like this Order ID to be blank by default. Or at the very least, can it not include a private link to the POS Plugin?

One solution would be for us to call the API later and to update all invoices to remove such Order ID's. 

But, I believe this is a bug, it should
1) not include the POS Link as an order ID, leave it blank
2) or at the very least, it should create a unique order ID for every invoice.

Can you please leave this ORDER ID field blank? 

Thanks!

### How did you encounter this bug?

1) Create BTCPay Plugin POS .
2) Generate invoice via the plugin.
3) Pay the invoice.
4) Send the Receipt to the buyer, and notice that Receipt contains link to private POS Plulgin app.

### What browser do you use?

any browser

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

pavlenex:
@dennisreimann should we remove the order ID though? I am unsure, but customer should have some sort of invoice id on the receipt to be able to distinguish? App id is def not helpful, thoughts?

dennisreimann:
@pavlenex in the PR I did not remove the order id, just the link  associated with it. I get that the order id in its current form isn't good and not an id, changing it would be breaking though – something to discuss in the next dev meeting.

NicolasDorier:
@dennisreimann we shouldn't remove the link in all case though.
I saw somebody relying on the fact that the receipt has a link to the `orderUrl` as part of the flow for the customer. (In woocommerce, this would redirect to the woocommerce order page)

pavlenex:
[...]

PR #5127, opened 2023-06-27
Generate unique order IDs for PoS and Crowdfund sales

dennisreimann:
Before, the order IDs for the apps weren't unique as they all shared the same app order ID, e.g. `pos-app_4BPMc3oqazLim9H67qc36JLvRmMt`. This appends a randomized value to make it a unique order ID, like `pos-app_4BPMc3oqazLim9H67qc36JLvRmMt-5NpZgWgYwE1QNQ9eT8rAhw`.

Also modifies the search to incorporate those changes.

Part of #5054.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -1142,6 +1142,7 @@ namespace BTCPayServer.Controllers
                 StoreId = fs.GetFilterArray("storeid"),
                 ItemCode = fs.GetFilterArray("itemcode"),
                 OrderId = fs.GetFilterArray("orderid"),
+                AdditionalSearchTerms = fs.GetFilterArray("plugin"),
                 StartDate = fs.GetFilterDate("startdate", timezoneOffset),
                 EndDate = fs.GetFilterDate("enddate", timezoneOffset)
             };
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -1100,10 +1100,9 @@ namespace BTCPayServer.Controllers
 
             // Apps
             var apps = await _appService.GetAllApps(GetUserId(), false, storeId);
-            var appIds = fs.GetFilterArray("appid");
-            if (appIds is { Length: > 0 })
+            var appTags = fs.GetFilterArray("appid")?.Select(AppService.GetAppInternalTag);
+            if (appTags != null)
             {
-                var appTags = appIds.Select(AppService.GetAppInternalTag);
                 list = list.Where(inv => inv.Version < InvoiceEntity.InternalTagSupport_Version ||
                                 inv.InternalTags.Any(it => appTags.Contains(it))).ToArray();
             }
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -16,6 +16,7 @@ using BTCPayServer.Data;
 using BTCPayServer.Filters;
 using BTCPayServer.HostedServices;
 using BTCPayServer.Models;
+using BTCPayServer.Models.AppViewModels;
 using BTCPayServer.Models.InvoicingModels;
 using BTCPayServer.Models.PaymentRequestViewModels;
 using BTCPayServer.Payments;
@@ -1089,8 +1090,9 @@ namespace BTCPayServer.Controllers
             }
             model.Search = fs;
             model.SearchText = fs.TextSearch;
-            
-            InvoiceQuery invoiceQuery = GetInvoiceQuery(fs, timezoneOffset);
+
+            var apps = await _appService.GetAllApps(GetUserId(), false, storeId);
+            InvoiceQuery invoiceQuery = GetInvoiceQuery(fs, apps, timezoneOffset);
             invoiceQuery.StoreId = storeIds.ToArray();
             invoiceQuery.Take = model.Count;
             invoiceQuery.Skip = model.Skip;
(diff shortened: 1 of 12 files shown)


HadiAreeshi (commented):
bc1qjj9v2hj4l2mn5fukhnkprywtxr9x5daew2t0sq

Kukks (commented):
I am worried that this might break integrations that depend on that same order id

NicolasDorier:
@dennisreimann I think we should completely remove the part `app_4BPMc3oqazLim9H67qc36JLvRmMt-` .

I believe that you'll still be able to find everything in search. If not, we have a property in `CreateInvoiceRequest` to add indexed values. We can put the PoS there.
Also we can put the pos id in the Metadata of the invoice.

> I am worried that this might break integrations that depend on that same order id

We never documented it, I think this is OK.

yuriygeorge:
I think it would be great also, if the app_4BPMc3oqazLim9H67qc36JLvRmMt--5NpZgWgYwE1QNQ9eT8rAhw orderId link did not point back to the POS app. Although some use it as a link to navigate back home. Except, if we are sharing this receipt with consumers/buyers, and we do NOT want them to have the link/access to OUR/merchant POS app/terminal interface, then we do not want them to be able to click back on the link.

dennisreimann:
Updated it to only use the randomly generated ID as Order ID and put the `app_4BPMc3oqazLim9H67qc36JLvRmMt` part into the `AdditionalSearchTerms` of the invoice object. 

The invoice filtering and dashboard tiles work backwards-compatible with these changes.

NicolasDorier:
the code seems good. I need to make sure the SQL queries aren't fucking up the perfs though. Will do

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -1142,6 +1142,7 @@ private InvoiceQuery GetInvoiceQuery(SearchString fs, int timezoneOffset = 0)
                 StoreId = fs.GetFilterArray("storeid"),
                 ItemCode = fs.GetFilterArray("itemcode"),
                 OrderId = fs.GetFilterArray("orderid"),
+                AdditionalSearchTerms = fs.GetFilterArray("plugin"),
```
NicolasDorier, line 1145: why you call it `plugin`? additional search terms aren't related to plugin. It's just another way to find invoices.

NicolasDorier: If you want to make sure this come from `plugin`, you need to do an additional tests on the returned invoices. (The internal tags)

NicolasDorier:
The request generated by EF is

```SQL
SELECT t."Id", t."Archived", t."Blob", t."Blob2", t."Created", t."CurrentRefundId", t."CustomerEmail", t."ExceptionStatus", t."ItemCode", t."OrderId", t."Status", t."StoreDataId", p."Id", p."Accounted", p."Blob", p."Blob2", p."InvoiceDataId", p."Type", t0."InvoiceDataId", t0."PullPaymentDataId", t0."Id", t0."Archived", t0."Blob", t0."EndDate", t0."Period", t0."StartDate", t0."StoreId"
FROM (
    SELECT i."Id", i."Archived", i."Blob", i."Blob2", i."Created", i."CurrentRefundId", i."CustomerEmail", i."ExceptionStatus", i."ItemCode", i."OrderId", i."Status", i."StoreDataId"
    FROM "Invoices" AS i
    WHERE NOT (i."Archived") AND EXISTS (
        SELECT 1
        FROM "InvoiceSearches" AS i0
        WHERE ((i."Id" <> NULL) AND (i."Id" = i0."InvoiceDataId")) AND i0."Value" = ANY (@__queryObject_AdditionalSearchTerms_0))
    ORDER BY i."Created" DESC
    LIMIT @__p_2 OFFSET @__p_1
) AS t
LEFT JOIN "Payments" AS p ON t."Id" = p."InvoiceDataId"
LEFT JOIN (
    SELECT r."InvoiceDataId", r."PullPaymentDataId", p0."Id", p0."Archived", p0."Blob", p0."EndDate", p0."Period", p0."StartDate", p0."StoreId"
    FROM "Refunds" AS r
    INNER JOIN "PullPayments" AS p0 ON r."PullPaymentDataId" = p0."Id"
) AS t0 ON t."Id" = t0."InvoiceDataId"
ORDER BY t."Created" DESC, t."Id", p."Id", t0."InvoiceDataId", t0."PullPaymentDataId"

```

I tested it manually in postgres, and couldn't make it works.
Reason is that `i."Id" <> NULL` is always NULL and thus the whole expression always NULL....
This seems to be a bug of entity framework.

NicolasDorier:
I am scratching my head as the `TextSearch` behave the same.... maybe EF doesn't show me the same request it is actually sending to server?

NicolasDorier:
FUCKING ENTITY FRAMEWORK.

EF has a `.ToSql` method to know which SQL query it is generating... EXCEPT IT IS NOT THE ONE IT SENDS TO THE SERVER.

FWipeu)*UWf903

BTCPayServer/Controllers/UIInvoiceController.UI.cs
NicolasDorier: 1. What I mean is that `AdditionalSearchTerms` isn't coupled with anything plugin/PoS related, so it shouldn't be called `plugin`.
2. If you insist to use `plugin`, then you need to take a look at the internal tags. This is some internal tags saved in the InvoiceEntity, that can't be modified through API, and that we use to get accounting of PoS/Crowdfund. Without this, anybody with invoice creation permission could create an invoice that fake a payment on PoS/Crowdfund.

NicolasDorier (changes requested):

dennisreimann:
@NicolasDorier I addressed your feedback in 2a2265215864d0f821938da1f26b704683707d32 — if you know a way to filter the invoices on the DB-level let me know, I had to resort to [doing it after querying](https://github.com/btcpayserver/btcpayserver/pull/5127/files#diff-f08ba1f2067dd60345bba335ad6c8b02e6880656f7d0cadf6183885c62d8c6d0R1103), which isn't ideal I think.

dennisreimann:
@pavlenex @dstrukt How do you feel about this? @NicolasDorier brought up to simply remove the apps filter, because we are constrained in how we can lookup/filter invoices by app — we cannot do it on the database-level, hence paging and additional filters (e.g. times) might not work well.

Please share your thoughts.

dstrukt:
I'm comfortable removing it, but can see instances in the future where you want to filter by the "app" or plugin .. or perhaps being able to Search the plugin's name, or done in the latest Reporting feature ..? 

Apologies for my ignorance here, just trying to figure out an alternative solution, if possible.

NicolasDorier:
I preserved the `appid` filter so we don't lose features. The downside of it is that somebody malicious can include an invoice in the app's invoice list even if the invoice wasn't created by the app.  This is however a bug that was already present before, and there is no way to fix it in this paginated page.

dennisreimann:
@NicolasDorier Is there something left to do here or can we review (and merge) this?

PR #5128, opened 2023-06-28
Fix: Lightning wallet such as Alby should be able to access LND config

NicolasDorier:
Trying to setup Alby on my node, I have the following error.
This PR enable CORS so it should then work.

image

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.cs
+++ b/BTCPayServer/Controllers/UIServerController.cs
@@ -29,6 +29,7 @@ using BTCPayServer.Storage.Services;
 using BTCPayServer.Storage.Services.Providers;
 using BTCPayServer.Validation;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Cors;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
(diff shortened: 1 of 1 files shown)


dennisreimann:
I cannot reproduce it with a valid session, neither in Chromium nor in Firefox. Even though the console errors you posted clearly indicate a CORS issue, it works for me without problems. I can only reproduce the "Connection failed" error once the LND config validity period expired.

NicolasDorier:
The strange thing is that on my side it also worked at one point, because I had it setup before.

NicolasDorier:
Tried again, and it worked. wtf is this

NicolasDorier:
I can't reproduce.

NicolasDorier:
Note that there was an error 404. So that may be the real issue. That said, I don't see any CORS error anymore.

NicolasDorier:
Even with a wrong URL I don't get CORS error anymore.

I don't know if we should merge that or not then...

dennisreimann:
Let's close it, we can bring it up again if it occurs again and we get t o kn ow about the actual problem.

NicolasDorier:
Fuck it happened again

image

NicolasDorier:
I don't think this is an issue on btcpay side see https://github.com/getAlby/lightning-browser-extension/issues/2522

NicolasDorier:
I will merge this.

Activating CORS on LND fixed part of the issue: https://github.com/btcpayserver/btcpayserver-docker/commit/6dc2c783b193e96c8ce6345977c98fd6a080b5ae

However, I still have the error on the pairing part, which this PR fix.

MV-801200320:
Bitcoin

PR #5131, opened 2023-06-29
add extension point for pull payment view

Kukks:


dennisreimann:
Just because I'm curious and this extends the view below the footer: What's the use case? Is it to insert scripts? If so we should name it differently imho.

SAGIE0724:
How can I withdraw my funds from the bitcoin account

Kukks:
> Just because I'm curious and this extends the view below the footer: What's the use case? Is it to insert scripts? If so we should name it differently imho.

Yes, adding additional scripts. What should we name it? We should come up with a standard name for this. Ideally also introduce extension points that can touch as many pages as possible too

dennisreimann:
@Kukks For the crowdfund we have `crowdfund-head`, which matches the naming of our layout partials (`LayoutHead`, `LayoutFoot`) and the sections to customize styles/scripts for individual pages (`PageHeadContent`, `PageFootContent`).

If you want to add more general extension points for scripts and styles, I'd say a good place to add them would be at the end of the layout partials, so that they can extend/override what core provides already.

Going with this, a good name for this extension point would be `pullpayment-public-foot`. Adding the `public` part to distinguish it from the backend editing view … unfortunately we'd have to rename `crowdfund-head` to `crowdfund-public-head` consequently.

dennisreimann:
@Kukks Waht do you think about the `public` part? Yay or nay?

Kukks:
> @Kukks Waht do you think about the `public` part? Yay or nay?

If it means changing existing ones to be aligned, may. We can just have a specific one for backend side

PR #5132, opened 2023-06-29
Update to Bootstrap v5.3

dennisreimann:
Maintenance upgrade, based on btcpayserver/btcpayserver-design#63. 

This mostly needs another pair of eyes, please simply go through the pages and see if something looks weird/unfamiliar.


NicolasDorier:
I will merge it now. We just made a release, so the next one will take time and we are likely to find any issue by then.

dennisreimann:
Thanks, good to have even more eyes on it!

Issue #5130, opened 2023-06-29
Sticky header for POS Plugin


pavlenex:
When a merchant has multiple items, it's currently difficult to navigate to the cart. The simplest solution would be to add sticky header.

https://github.com/btcpayserver/btcpayserver/assets/36959754/311ec8a1-30cf-487f-a179-cf4f83175bd4

Issue #5109, opened 2023-06-22
POS Product Cart view - make it consistent with our design guidelines and dstruktify it!


pavlenex:
Our POS Cart view seemed to be used quite a lot, however the design/styling of it looks quite outdated compared to our design system . This is a tracker issue that will try to log all the improvements we need to make, but those improvements will be posted here most likely  next week once we determine course of action.

Bear in mind that there will be category/group changes coming in that we should consider. #5078 
```[tasklist]
### Tasks
- [x] Come up with design @dstrukt
- [x] Implement proposed design
- [x] Ensure we only touch on cosmetic side of things, but also think if we'd like to improve the checkout flow for the next major release
```

dstrukt:
I love that I've been given an adverb to describe my pixel pushing!

It's not kukksification, but it's honest work.

dennisreimann:
Should incorporate #5130 and the changes made in #5088

PR #5133, opened 2023-06-29
#5130

TChukwuleta:
#5130 

The card and the 'find product' input fields are static while scrolling down the product list

image


BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
leesalminen, line 1: I was also taking a look at this issue and also added `background-color: var(--btcpay-body-bg);` to the sticky container.

BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
```diff
@@ -212,7 +212,7 @@
 <div class="wrapper">
     <!-- Page Content  -->
     <div id="content">
-        <div class="p-2 p-sm-4">
+        <div class="p-2 p-sm-4 sticky-top">
             <partial name="_StatusMessage" />
             <div class="d-flex gap-3 mb-4">
```
leesalminen, line 217: Also when I was playing around with this I think it looked better with `mb-4` removed.

dennisreimann:
We should do this as part of #5109.

pavlenex:
> We should do this as part of #5109.

@dennisreimann can you elaborate what that means? Do we wait for #5109 and then rebase this PR? Would be a shame to waste PR efforts here, instead let's see how we can work together towards the desired outcome?

dennisreimann:
@pavlenex This here is basically a one line change PR, which doesn't fully cover the use case. I don't mind merging it, but spending more work here imho doesn't make sense, as we will redo the whole UI in #5109 anyways

pavlenex:
@dennisreimann How does it not cover the use-case? I'm fine with opening a new PR, just wanted to make sure there's clarity when communicating these things.

dennisreimann:
The background is missing (see screenshot) and we need to ensure it works across breakpoints. I'll wait with #5109 then until this is merged.

TChukwuleta:
Sure @pavlenex. No problem.

PR #5136, opened 2023-06-30
Custom Forms: Allow HTML in labels and help text

### LINKED ISSUE #5003: [Bug]: Custom Forms HTML in Label get's stripped but not in Error message for mandatory fields

master branch (dated 25th may 2023)

docker+docker build

### What happened?

A user in a feature request discussion asked for the ability to ack/nack to his TOS with a link. Trying to build something usable with what we have, turned up this bug: I entered html in the label of the field and made it mandatory. The html gets escaped when displaying it as label but in the error message when the mandatory field is not filled out, it gets shown as a link. (see screenshot)

grafik

json of the form element I created:

```json
{
  "topMessages": [],
  "fields": [
    {
      "name": "newField1",
      "constant": false,
      "type": "select",
      "value": "newOption2",
      "required": true,
      "label": "Agree to our <a href=\"tos.html\">TOS</a>",
      "originalValue": null,
      "helpText": null,
      "validationErrors": [],
      "fields": [],
      "options": [
        {
          "text": "NACK",
          "value": ""
        },
        {
          "text": "ACK",
          "value": "ack"
        }
      ]
    }
  ]
}
```

link to feature request comment: https://github.com/btcpayserver/btcpayserver/discussions/5002#discussioncomment-6002926

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox

### Additional information

none

dennisreimann:
We partially allowed it already for the help texts — up for discussion if we want to support this. Fixes #5003.


Kukks:
Maybe it's ok if we sanitize it?

dennisreimann:
> Maybe it's ok if we sanitize it?

THe `Safe.Raw` method sanitizes the output — or do you mean we should restrict it further and only allow some formatting and links?

Kukks (approved):

NicolasDorier:
the use of `v-html`  in VueJS for user input is a XSS vector. You can try it by adding `<script>alert("lol")</script>` in your label.

If you really want to do this, you need to run the sanitizer before affecting the property `model.Label = Safe.Raw(app.Label);` .

NicolasDorier (changes requested):
Need sanitize the `v-html` bound properties

BTCPayServer/Views/Shared/Forms/FieldSetElement.cshtml
```diff
@@ -5,7 +5,7 @@
 @if (!Model.Constant)
 {
     <fieldset>
-        <legend class="h3 mt-4 mb-3">@Model.Label</legend>
+        <legend class="h3 mt-4 mb-3">@Safe.Raw(Model.Label)</legend>
```
NicolasDorier, line 8: is this rendered in a form with VueJS? if so, you need to add `v-pre` to avoid injection.

dennisreimann: The form page that includes this doesn't use Vue.

dennisreimann:
> Need sanitize the v-html bound properties

Added [vue-sanitize-directive](https://github.com/leopiccionia/vue-sanitize-directive/) in [28cb01f](https://github.com/btcpayserver/btcpayserver/pull/5136/commits/28cb01f788c320190c3b04d06e0432507432028e). Guess we can reuse that elsewhere too.

NicolasDorier:
@dennisreimann will need doc update on https://docs.btcpayserver.org/AdvancedForms/ indicating helptext can be html.

dennisreimann:
@NicolasDorier Added a PR for the doc update.

PR #5137, opened 2023-06-30
Invoice lists: Show icons for payment methods

### LINKED ISSUE #5084: Show icons for payment methods in invoice view
Feedback we received from paralelni polis POS cashier is that they would really love to have crypto icons shown for their payment methods.

For now we should LN but could we show icons in the invoice view for plugins for example fixed float?

Image

So:

- [ ] Show cryptoicons for all payment methods supported in the invoice view
- [ ] Check if we can showcase at least plugin icon when external plugin was used

dennisreimann:
Closes #5084.

## Invoices

invoices

## Dashboard

dashboard

### DIFF
--- a/BTCPayServer/Components/StoreRecentInvoices/Default.cshtml
+++ b/BTCPayServer/Components/StoreRecentInvoices/Default.cshtml
@@ -3,6 +3,7 @@
 @using BTCPayServer.Services
 @using BTCPayServer.Services.Invoices
 @inject DisplayFormatter DisplayFormatter
+@inject PaymentMethodHandlerDictionary PaymentMethodHandlerDictionary
 @model BTCPayServer.Components.StoreRecentInvoices.StoreRecentInvoicesViewModel
 
 <div class="widget store-recent-invoices" id="StoreRecentInvoices-@Model.Store.Id">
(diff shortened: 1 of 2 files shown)


dstrukt:
Love the icon treatment, but curious why we're adding it to the `Status`? vs. the `Amount`, or even a separate column for `Payment` or `Payment Method` so it could be sorted?

Also wondering if it makes sense to merge the Bitcoin/Lightning icons into a singular icon, or just opt for Lightning ... hmmm.

dstrukt (commented):
Love the icon treatment, but curious why we're adding it to the `Status`? vs. the `Amount`, or even a separate column for `Payment` or `Payment Method` so it could be sorted?

Also wondering if it makes sense to merge the Bitcoin/Lightning icons into a singular icon, or just opt for Lightning ... hmmm.

dennisreimann:
> On second thought, I do appreciate how compact this iteration is thinking

Yeah, I think so too — information-wise I think of it like an addition about how the invoice got settled, so it makes sense to have it there imho.

> Also wondering if it makes sense to merge the Bitcoin/Lightning icons into a singular icon, or just opt for Lightning ... 

We don't have that, because e.g. Litecoin also supports LN. We can disregard that, but we'd have to come up with separate icons then and maybe even add one for LNURL, if we also want to distinguish that. Not worth the effort though imho.

pavlenex:
uTACK.

dstrukt:
Hmm all valid points, will give it a proper review today!

dstrukt (approved):
Might want to revisit the points brought up, possibly in a different way, but atm...

LGTM - tACK

PR #5140, opened 2023-07-01
Apps: Add direct file upload in item editor

### LINKED ISSUE #5086: Add direct file upload in POS
> When I want to add photo to POS app and copy the file, it doesn't work (absolute vs relative paths)

@NicolasDorier and I talked about this, I'll add direct file upload in the POS. No need to consider it here then.

_Originally posted by @dennisreimann in https://github.com/btcpayserver/btcpayserver/issues/5081#issuecomment-1595863402_

dennisreimann:
Closes #5086.

### DIFF
--- a/BTCPayServer/Controllers/UIAppsController.cs
+++ b/BTCPayServer/Controllers/UIAppsController.cs
@@ -1,6 +1,9 @@
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
+using BTCPayServer.Abstractions.Contracts;
+using BTCPayServer.Abstractions.Extensions;
 using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client;
 using BTCPayServer.Data;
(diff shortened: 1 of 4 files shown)


BTCPayServer/Controllers/UIAppsController.cs
```diff
@@ -184,13 +191,50 @@ public async Task<IActionResult> DeleteAppPost(string appId)
...
         async Task<string> GetStoreDefaultCurrentIfEmpty(string storeId, string currency)
         {
             if (string.IsNullOrWhiteSpace(currency))
             {
-                currency = (await _storeRepository.FindStore(storeId)).GetStoreBlob().DefaultCurrency;
+                var store = await _storeRepository.FindStore(storeId);
+                currency = store?.GetStoreBlob().DefaultCurrency;
             }
-            return currency.Trim().ToUpperInvariant();
+            return currency?.Trim().ToUpperInvariant();
```
Kukks, line 237: nit: not required

Kukks (approved):
Code ACK. Would be good if we could find a way to repackage this as a reusable component.

BTCPayServer/Controllers/UIAppsController.cs
dennisreimann: Rider complains about a possible null reference, that's why I added it.

dennisreimann:
> Would be good if we could find a way to repackage this as a reusable component.

We don't have it any other place in Vue and the partial is shared across the apps already, so I figured that's not necessary right now. 

I'd see potential for making the backend upload functionality more general though — let's evaluate that when we add the next endpoint to accept direct file uploads.

BTCPayServer/Controllers/UIAppsController.cs
NicolasDorier: it is actually required. If the `storeId` is wrong, the currency also is.

PR #5142, opened 2023-07-02
Compare domains in lowercase

AaronDewes:
Domains are case-insensitive, so this comparison should be too.

I encountered this issue with a Citadel user who accidentally named their domain an uppercase name (Pay.example.com), but browsers automatically converted it to pay.example.com. They got the ` BTCPay is expecting you to access this website from ...` message in this case.


dennisreimann (approved):
Thanks for the contribution.

Issue #5095, opened 2023-06-19
Improve invoice reporting


pavlenex:
This issue will be updated with proper context and description as soon as we get specific feedback from people who requested better invoice exporting/reporting.

Update:

![Screenshot 2023-06-20 at 10 37 50](https://github.com/btcpayserver/btcpayserver/assets/36959754/222cac99-06bb-48cd-83c4-f0f5d44d87bc)

pavlenex:
Possibly we can add items from POS 

https://github.com/btcpayserver/btcpayserver/issues/937

EscapeToElSalvador:
I haven't found it discussed elsewhere, but automating the reporting (e.g. send email with report attached periodically) would be helpful, imo

Issue #937, opened 2019-07-21
Sold items from POS app not exported


Kixunil:
**Is your feature request related to a problem? Please describe.**
It's impossible to process sold items in automated fashion because the item data is missing from exported invoices. The only way to do that is to manually open each invoice and copy-paste the item data.

**Describe the solution you'd like**
Add the items into the exported data. In case it is undesirable to mix them, create a separate export items feature.

**Describe alternatives you've considered**
We're thinking about writing some simple SQL commands that will do that directly on the server. This will be easy to do once we understand the data in the database, but a nicer, official solution would be useful for other merchants. If you decide against this feature, it'd be very helpful if you could hint us at what exactly is in the database (the proper SQL command to get the data)

pavlenex:
Related slightly to #590

Kixunil:
Yes, resolving that one would be also helpful for us. I thought that while similar, it makes sense to keep the issues separate.

bolatovumar:
@Kixunil what status do the invoice that are missing have? I'm trying to repro the issue but can't as I'm not sure what status the invoice should be in in order for me to repro the issue.

The status that exist as far as I can see are: complete, paid, confirmed, invalid, expired, new.

And just to confirm I'm talking about this screen:
![Screen Shot 2019-08-21 at 6 07 27 PM](https://user-images.githubusercontent.com/1934678/63444298-00fbda80-c43f-11e9-9180-bd8e35407754.png)

Kixunil:
The status is complete. The Export button is there of course, but if you try to export the data, you'll see that the POS items aren't there - just invoices.

FYI, I've implemented a hacky ugly script to do that, but it'd be better if we could do it directly.

bolatovumar:
@Kixunil, oh, okay, I misunderstood your issue then. I thought you meant that invoices for items marked as "complete" weren't getting exported but the issue seems to be that there is no indication as to which POS items these invoices correspond to, correct?

bolatovumar:
@Kixunil what data specifically would you want to see? I have "invoice item code" and "invoice item description" in my export. Is there anything else you would like to see alongside these two or am I missing the point?
![Screen Shot 2019-08-21 at 7 31 45 PM](https://user-images.githubusercontent.com/1934678/63450427-a8cad580-c44a-11e9-842e-0ee20211e82d.png)

Kixunil:
You're correct now. Those two are enough. (I guess the time is already there, right?)

[...]

PR #5155, opened 2023-07-05
Add reporting feature

NicolasDorier:
## Features

* [x] Export to CSV. This can be easily interpreted by excel or google sheet for more complex use cases.
* [x] Some cells are links to easily navigate to details (a transaction on the blockchain explorer or an invoice for now)
* [x] Extendable by plugins
* [x] Sorting on all columns
* [x] Payments report
* [x] On-chain wallets report
* [x] Date inputs and outputs using the browser's timezone.
* [ ] Filters
* [ ] Basic grouping features, so user doesn't have to use excel to see some aggregates (like the total of payments in a month)
* [ ] Brush up by @dennisreimann and @dstrukt 
* [ ] Tests

## Out of scope

While the Greenfield route has been done, I think we should disable it for now, as we are still unsure of proper API design.

Additional reports will be in different PRs after we could validate the design of this PR.

## To do later

* Report for Off-Chain payments (will need to have date base filtering of invoices on lightning library)
* Point of Sale/Crowdfunding stats on items sold (see https://github.com/btcpayserver/btcpayserver/discussions/4598)
* Refund reports
* Payouts reports

## UX

This add a new page at store level: Reporting:

image

When clicking on it, a new page open

image

Note on the UX: The hard part is that the designer is unable to predict how many columns will be there. That said, cell content can be customized depending on the type of the column.

### Backend

A new class `ReportService` is exposing report related methods.
A base class `ReportProvider` can be subclassed to implement new types of reports.
At the moment, only two implementations exists `PaymentsReportProvider` and `OnChainWalletsReportProvider`, but many more will come for next major release.

### Front end

I am using VueJs, the whole page is coded as a single page application.

## Linked discussions

https://github.com/btcpayserver/btcpayserver/issues/5095
https://github.com/btcpayserver/btcpayserver/issues/937
https://github.com/btcpayserver/btcpayserver/discussions/2561
https://github.com/btcpayserver/btcpayserver/pull/2601
https://github.com/btcpayserver/btcpayserver/discussions/4598


BTCPayServer.Client/BTCPayServer.Client.csproj
```diff
@@ -33,6 +33,8 @@
       <PackageReference Include="BTCPayServer.Lightning.Common" Version="1.3.21" />
       <PackageReference Include="NBitcoin" Version="7.0.24" />
       <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+      <PackageReference Include="NodaTime" Version="3.1.9" />
```
Kukks, line 36: Do we really need a whole new library on the client library level?

BTCPayServer/wwwroot/js/store-reports.js
```diff
@@ -0,0 +1,199 @@
...
+            {
+                sortBy: "",
+                sortByTitle: "",
+                sortIconClass: "fa fa-sort"
+            };
+        }
+    }
+};
+
+$(function () {
```
Kukks, line 85: reduce jquery usage:
```suggestion
document.addEventListener("DOMContentLoaded", () => {
```

BTCPayServer/wwwroot/js/store-reports.js
```diff
@@ -0,0 +1,199 @@
...
+    $(".flatdtpicker").on("input", function () {
```
Kukks, line 86:
```suggestion
    document.querySelectorAll('.flatdtpicker').forEach(el => el.addEventListener("input", ()=>{
```

BTCPayServer/wwwroot/js/store-reports.js
```diff
@@ -0,0 +1,199 @@
...
+        // We don't use vue to bind dates, because VueJS break the flatpickr as soon as binding occurs.
+        var to = $("#toDate").val();
```
Kukks, line 88:
```suggestion
        const to = document.querySelector("#toDate").value
```

NicolasDorier:
What is the problem with jquery?

NicolasDorier:
@dstrukt @dennisreimann I just added a commit to generate fake data in the reporting.

* When chosing a new view, I believe it is better to not show the raw data upfront unlike I am doing now. Reasons are:
  *  For 10000 rows, the UI seems to be slow to render for some reason. (Probably something that can be fixed)
  * Few people would want to see the raw data, in general I expect they are more interested in the aggregates.
  * The raw data having many column, it can't look nice

dennisreimann:
> UX: This add a new page at store level: Reporting

I think we could either put it up between Dashboard and Settings or at the end of the Payments section. Regardless of where we position it, we need a new icon for this feature :) @dstrukt

dennisreimann:
Introduced a fixed header and made the tables auto-width, so that thgey read a bit better. Will experiment with sticky table header rows next.

Peek 2023-07-11 18-36

NicolasDorier:
The make me more happy

BTCPayServer/Views/UIReports/StoreReports.cshtml
```diff
@@ -0,0 +1,129 @@
...
+@using BTCPayServer.Views.Invoice;
+@using BTCPayServer.Views.Stores;
+@using BTCPayServer.Abstractions.Services;
+@using Microsoft.AspNetCore.Routing;
+@inject Safe Safe
+@inject BTCPayServer.Security.ContentSecurityPolicies Csp
+@model StoreReportsViewModel
+@{
+	ViewData.SetActivePage(StoreNavPages.Reporting, "Reporting");
+	Csp.AllowUnsafeHashes("javascript:void(0)");
```
dennisreimann, line 13: What is this for?

BTCPayServer/wwwroot/js/datatable.js
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    function applyFilters(rows, fields, filterStrings) {
+        if (!filterStrings || filterStrings.length === 0)
+            return rows;
+        // filterStrings are aggregated into one filter function:
+        // filter(){ return filter1 && filter2 && filter3; }
+        var newData = [];
+        var o = {};
+        eval('function filter() {return ' + filterStrings.join(' && ') + ';}');
```
dennisreimann, line 160: Yikes 😬

NicolasDorier: The basic idea is that the report provider can decide to filter rows before building the table.
I use it for the summary of lightning addresses payments. Without this, the datatable would show a row with empty ligthning address if none are present. Instead, I want this line to be gone.

NicolasDorier: Note that the conversion from tabular data to a representation bindable to VueJS for summary tables is done JS side.
I am unsure this is good idea, I believe we could just do that C# side.

BTCPayServer/Views/UIReports/StoreReports.cshtml
NicolasDorier: Note that this trick doesn't work in safari...

dennisreimann: Removed and fixed in [f07b85a](https://github.com/btcpayserver/btcpayserver/pull/5155/commits/f07b85a9defcbe19f07673e7227b0b1638822ab0).

dstrukt (changes requested):
# Still Reviewing & Updating Mocks

Few initial thoughts after reviewing, some might be out of scope, others are minor nits and suggestions:

Happy to make some of these changes post-merge as well, if it makes more sense.

## Search

- [ ] Do we want the ability to search? e.g. something like this:
- Calendar icons to the left of the inputs for the date selection could be nice too
Screen Shot 2023-07-17 at 3 02 49 PM


## Balance Tables

- [ ] Any reason we don't want to conserve a bit of space by doing two columns like so. Am I missing something? Might be nice to consolidate in the future, but for the time being.
Screen Shot 2023-07-17 at 1 49 54 PM

### Payments
- [ ] Could be improved with updated labels:
Screen Shot 2023-07-17 at 2 29 22 PM

- [ ] Lightning Address label updates
- Makes me think some of the other Amount labels could be Balance instead...hmm
Screen Shot 2023-07-17 at 3 51 15 PM


### On-Chain
- [ ] Could be improved with updating the label and pushing the header into the table key value for the on-chain view:
- Context: I found the Balance Change aspect of the On-Chain table a little confusing... as the test data amounts were mostly negative values?
Screen Shot 2023-07-17 at 3 00 53 PM

- Still exploring ways of consolidating the tables for more space and ease of parsing the data, just sharing, might be a later follow up if any of the explorations make sense:
Screen Shot 2023-07-17 at 4 18 18 PM

e.g: On-Chain/Lightning, doesn't hold up as well, but interesting nonetheless:
Screen Shot 2023-07-17 at 5 56 21 PM

works in a simple context as well:
Screen Shot 2023-07-17 at 6 27 36 PM


## Raw Data
### Payments
- [ ] Is the date formatting messed up? `2023-07-05T01:42:41-06:00` 
Screen Shot 2023-07-17 at 3 34 10 PM

### On-Chain
- [ ] Suggestions for improvements
- Unconfirmed would be a red `X` icon instead of the green checkmark
- Balance Change decrease would be `-1.06150021` and be colored red
Screen Shot 2023-07-17 at 3 41 50 PM



## Filter Dropdown

- [ ] Copy suggestion: `Items sold` -> `Products` or `Products Sold`? To keep it consistent with the POS label
- [ ] Copy suggestion: `On-Chain Wallets` -> `On-Chain` or `Wallets`? On second thought, might be fine with the original..
Screen Shot 2023-07-17 at 1 37 15 PM

## General

- [ ] I noticed there's an additional scrollbar at the bottom of the page, can we remove this? I think the table specific scrollbar makes sense, given the width of it's contents.
Screen Shot 2023-07-17 at 1 34 33 PM


- [ ] The main CTA could arguably just be `Export` cleaner, as the CSV bit isn't necessary imo
Screen Shot 2023-07-17 at 2 14 43 PM

## Really great work!

NicolasDorier:
About wallet balance confirmed/unconfirmed: I don't believe a bar make sense. In most case, unconfirmed will be ridiculously low compared to confirmed, so this doesn't add lot's of information.

For on-chain versus lightning, it might make more sense.

Also, I was thinking: Some tables are grouping on Invoice currency. I believe this is overkill, as most of the time, the invoice currency is the same.

Rather than grouping per currency, I will just make additional chart per currency?

NicolasDorier:
@dstrukt I want to add, the aggregated tables I am showing could potentially be replaced by any chart that make sense.

BTCPayServer/Controllers/UIReportsController.CheatMode.cs
```diff
@@ -0,0 +1,114 @@
...
+        if (f.Name == "CryptoAmount")
+            return GenerateDecimal(0.1m, 5m, 8);
+        if (f.Name == "LightningAddress")
+            return TakeOne("satoshi", "satosan", "satoichi") + "@bitcoin.org";
+        if (f.Name == "BalanceChange")
+            return GenerateDecimal(-5.0m, 5.0m, 8);
+        if (f.Type == "datetime")
+            return DateTimeOffset.UtcNow - TimeSpan.FromHours(rand.Next(0, 24 * 30 * 6)) - TimeSpan.FromMinutes(rand.Next(0, 60));
+        if (f.Name == "Currency")
+            return rand.NextSingle() > 0.2 ? "USD" : TakeOne("JPY", "EUR", "CHF");
```
Kukks, line 93: These hardcoded currencies are a very small subset

BTCPayServer/Services/Reporting/PaymentsReportProvider.cs
```diff
@@ -0,0 +1,187 @@
...
+        var rows = await conn.QueryAsync(command);
+        foreach (var r in rows)
+        {
+            var values = queryContext.CreateData();
+            values.Add((DateTime)r.created);
+            values.Add((string)r.invoice_id);
+            values.Add((string)r.order_id);
+            if (PaymentMethodId.TryParse((string)r.payment_type, out var paymentType))
+            {
+                if (paymentType.PaymentType == PaymentTypes.LightningLike || paymentType.PaymentType == PaymentTypes.LNURLPay)
```
Kukks, line 136: use the payment type tostring  instead of hardcoding

Kukks (changes requested):
Missing api docs and tests! 😂

BTCPayServer/Services/Reporting/PaymentsReportProvider.cs
NicolasDorier: I did this because because the difference between lightning versus LNURL doesn't matter at all for merchants.

They want the consolidated view of how much they received on lightning, they don't care whether it come from LNURL or not.

dstrukt:
> About wallet balance confirmed/unconfirmed: I don't believe a bar make sense. In most case, unconfirmed will be ridiculously low compared to confirmed, so this doesn't add lot's of information.

Tend to agree, it's a bit much for what would only be a pending state at best, and rarely visible for instances with low volume.

> For on-chain versus lightning, it might make more sense.

Also agree, but wasn't intended to be included in this PR, just a quick visual exploration to see if it made sense / worth the time to implement!

> Also, I was thinking: Some tables are grouping on Invoice currency. I believe this is overkill, as most of the time, the invoice currency is the same.

> Rather than grouping per currency, I will just make additional chart per currency?

That could work, I do appreciate the flexibility/clarity of the current implementation, but I think I'd have to see it.

pavlenex:
@NicolasDorier Any chance fake data can generate items for POS? I think this is one of the highest-requested reports, so would be good to test it out.

dstrukt:
ACK to fake data, if possible! Was my biggest challenge in providing more feedback as well.

NicolasDorier:
@dstrukt @pavlenex done

pavlenex:
@NicolasDorier I've tested draft a bit, so far so good.
Question, is export always meant to give raw data, not data according to tables?
When do you plan to add fitlers and groups?
When do you think you'll be able to get this one out of draft?

We're thinking of RC for Monday, but unsure how much you have to do on this one.

dstrukt (approved):
Overall, looks really solid, there's some more finessing of labels and UI I'd like to do, but I don't want to hold up the release to make these minor tweaks.

Especially if our RC deadline is monday, this is something @dennisreimann and I could easily do a follow up PR to address.

NicolasDorier:
it's bit tight, so let's merge this one and incrementally improve on dstrukt feedback

NicolasDorier:
> is export always meant to give raw data, not data according to tables?`

@pavlenex  I don't know what you mean, but the CSV export is exporting the raw data, because all other structures can be trivially created again through Excel. The idea is to provide just enough feature to be useful for simple cases, delegating complex cases to Excel.

Filtering for now is only on date, I believe allowing some simple client side filters on the raw data.
We can do that in minor release or major release later.

About Grouping: The summary tables that you see is grouping. If what you want is letting the user drag and drop components to make their own group, I wished to do that but the existing open source components for pivot table just sucks and aren't maintained anymore.

dstrukt:
> it's bit tight, so let's merge this one and incrementally improve on dstrukt feedback

Sounds good - will make an issue with some of the remaining items post-release. Great work!

JamesMessick1965:
Good

PR #5182, opened 2023-07-18
Introduce Payout metadata for api and plugins

Kukks:


BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
```diff
@@ -196,7 +196,14 @@
                                 <span>@pp.Date.ToBrowserDate()</span>
                             </td>
                             <td class="mw-100">
-                                <span>@pp.PullPaymentName</span>
+                                @if (pp.ProofLink is not null && pp.Source is not null)
```
NicolasDorier, line 199: `ProofLink`? Don't you mean SourceLink?

NicolasDorier (changes requested):
You made one typo with `ProofLink` rather than `SourceLink` in the payout list view.

Also, can you modify `PayoutExtensions.GetBlob(this PayoutData data, BTCPayNetworkJsonSerializerSettings serializers)` so that after deserialization, you initialize `Metadata` to empty object?

That way, even for old payouts, we don't have to test nullability of `Metadata` and risk some null ref exception.

Issue #5195, opened 2023-07-26
[Bug]: Receipt page shows wrong amount paid


ndeet:
### What is your BTCPay version?

BTCPay 1.11.0-rc4

### How did you deploy BTCPay Server?

docker

### What happened?

On the receipt page the fiat amount is wrong. e.g. on below screenshot the amoutn was $1.20 but it shows $0.000041 on the receipt page

![image](https://github.com/btcpayserver/btcpayserver/assets/1136761/d5957563-e7d2-438f-b32a-5a7f47efc5ea)

### How did you encounter this bug?

Paid invoice
clicked on receipt

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #5197, opened 2023-07-26
Reverting logic of how paid amount is displayed on the receipt

rockstardev:
Fix: #5195

In 22435a2bf510773afce0a47e9bc7929f30b94c8d @NicolasDorier has changed the display of amount on the receipt; I've reverted the logic in order to unblock release.

Here is how now invoice receipt page looks like

image

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -227,14 +227,18 @@ namespace BTCPayServer.Controllers
 
                     string txId = paymentData.GetPaymentId();
                     string? link = GetTransactionLink(paymentMethodId, txId);
-                    
+                    var paymentMethod = i.GetPaymentMethod(paymentMethodId);
+                    var amount = paymentData.GetValue();
+                    var rate = paymentMethod.Rate;
+                    var paid = (amount - paymentEntity.NetworkFee) * rate;
+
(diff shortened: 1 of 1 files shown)


Kukks (approved):

rockstardev:
Seeing that @Kukks approved, I'm merging to unblock testing and release, when you are back you can review changes @NicolasDorier @dennisreimann.

NicolasDorier:
Checking, the original code seems right.

NicolasDorier:
@rockstardev @Kukks fyi, take a look at https://github.com/btcpayserver/btcpayserver/compare/957fb09%5E..583813883cfb9741156b41a608a3a5df38c6f172#diff-f08ba1f2067dd60345bba335ad6c8b02e6880656f7d0cadf6183885c62d8c6d0L234-L236

The problem was that `.Paid` should be the invoice currency amount, not the crypto amount. So I replaced `PaidAmount` by `InvoicePaidAmount`. The former is in the crypto of the payment, the later in the currency.

I checked if I did a similar mistake elsewhere, and it seems I didn't.

It is important you understand the new `PaidAmount` stuff, it has been done to avoid difficult calculations like was before.

ndeet:
@NicolasDorier not sure it is fixed though, if @rockstardev only reverted the commit likely the weird 1,002 $ rounding errors (instead of showing $1.00) on the receipt page are back too?

NicolasDorier:
no, he reverted only part of the PR

Issue #5192, opened 2023-07-26
[Bug]: PoS cart, tips can be negative


ndeet:
### What is your BTCPay version?

BTCPay 1.11.0-rc4

### How did you deploy BTCPay Server?

docker

### What happened?

![image](https://github.com/btcpayserver/btcpayserver/assets/1136761/021b6e30-4941-4271-8f39-0c7c265c5952)

1. Tip field allows negative inputs which makes the invoice cheaper.
2. Discount field also allows -10% input which will make the invoice more expensive.

Maybe we can prevent users from inputting `-` sign at all?

### How did you encounter this bug?

PoS app on 1.11.0-rc4 

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #5198, opened 2023-07-26
Preventing entering of negative tips and discounts in POS

rockstardev:
Fix: #5192

I have added server side check that will prevent tips and discounts being sent in as negative numbers.
Also, I've added client side checks that will prevent these amounts being negative during calculations.

`min` validation on `input` still prevented form from being submitted as you can see from screenshot

image

but calculations were running, so to prevent confusion I've added extra check.


Kukks (approved):

rockstardev:
Seeing that @Kukks approved, I'm merging to unblock testing and release, when you are back you can review changes @NicolasDorier @dennisreimann.

ndeet:
Nice, thanks a lot, can confirm it is fixed with rc5. :partying_face:

Issue #5200, opened 2023-07-26
[Bug]: Payment sounds enabled on existing stores


ndeet:
### What is your BTCPay version?

BTCPay 1.11.0-rc5

### How did you deploy BTCPay Server?

docker

### What happened?

The sounds do play by default on existing stores. It works as expected on newly created stores.

For existing stores the only fix is to select online preset and hit save again.

Seems if nothing is set as preset it falls back to **sound** instead of **no-sound**. 

### How did you encounter this bug?

Create an invoice (no matter of pos or directly over invoices UI) and for existing stores it will play sounds. Expected behaviour is to play no sounds.

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #5201, opened 2023-07-26
Disabling playing of the invoice sound for existing stores

rockstardev:
Fixes: #5200

Database/json default is now aligned with property default to `false`


ndeet:
Can confirm it works as expected now, thank you :green_heart:

PR #5202, opened 2023-07-26
add opensats and update strike logo

Kukks:


pavlenex (changes requested):
Tested, but 
@Kukks can you please change order of OS and Spiral, Spiral should be before OS since overall they supporter us more significantly.

pavlenex (approved):
tACK now

PR #5229, opened 2023-08-04
Fixing the language selector box on checkout

evanc-ole:
This issue had been marked as resolved but still didn't show the entire word. The first picture is the original issue, the second is what is currently up, and the third is mine.
image
image
Screen Shot 2023-08-03 at 5 54 22 PM


dennisreimann (approved):
tACK. Thanks for the fix!

dennisreimann:
https://github.com/btcpayserver/btcpayserver/pull/5210

PR #5238, opened 2023-08-07
POS: Handle flexible price items in cart view

### LINKED ISSUE #5212: [Bug]: PoS cart view, custom amount item always 0.00

BTCPay 1.11.1

docker

### What happened?

With 1.11.1 the custom amount items are shown again but there is no way to add a custom amount and they will be added as 0.00 $ items.

This needs more thinking if they make sense or not at all. If not then we should show an info on the item or on top of item editor that custom amount items won't show there.

If they should show then there needs to be a way to enter the custom amount.

- Go to PoS app and add a custom amount item
- Add the item to cart

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

### LINKED ISSUE #5218: [Bug]: POS Cart View Custom Price hides item

1.11.1

Docker, Locally & https://master.btcpay.host/

### What happened?

Setting a custom price on an amount causes the item to hide in the shop view, and is no longer visible or searchable.

1. Create Shop
2. View Shop, to confirm all items appear
3. Change item price to `Custom`
4. Reload shop, to show items are no longer visible.

### Relevant log output

```shell
N/A
```

### What browser do you use?

Chrome Version 115.0.5790.114 (Official Build) (x86_64)

### Additional information

Will try to upload the video

dennisreimann:
Allows to set prices for minimum and custom amount items and allows to add multiple of them to the cart. This keeps track of the individual items on a per price basis

Closes #5218. Fixes #5212.

### DIFF
--- a/BTCPayServer/Services/Apps/AppService.cs
+++ b/BTCPayServer/Services/Apps/AppService.cs
@@ -411,7 +411,6 @@ namespace BTCPayServer.Services.Apps
                 return false;
             if (cartObject is null)
                 return false;
-            
             cartItems = new();
             foreach (var o in cartObject.OfType<JObject>())
             {
--- a/BTCPayServer/Services/Apps/AppService.cs
+++ b/BTCPayServer/Services/Apps/AppService.cs
@@ -402,27 +402,24 @@ namespace BTCPayServer.Services.Apps
             }
         }
 #nullable enable
-        public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)] out Dictionary<string, int> cartItems)
+        public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)] out List<PosCartItem> cartItems)
         {
             cartItems = null;
             if (posData is null)
                 return false;
             if (!posData.TryGetValue("cart", out var cartObject))
                 return false;
-            if (cartObject is null)
-                return false;
 
-            cartItems = new();
+            cartItems = new List<PosCartItem>();
             foreach (var o in cartObject.OfType<JObject>())
             {
                 var id = o.GetValue("id", StringComparison.InvariantCulture)?.ToString();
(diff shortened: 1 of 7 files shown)


BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -402,27 +402,24 @@ private static bool TryParseJson(string json, [MaybeNullWhen(false)] out JObject
             }
         }
 #nullable enable
-        public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)] out Dictionary<string, int> cartItems)
```
Kukks, line 405: this will break some plugins, can you create an overload instead

Kukks (changes requested):

dstrukt:
So we ultimately decided to support custom amounts?

dstrukt (approved):
tACK.

Going to do some more browser testing, as that seems to be the biggest issue we've had.

BTCPayServer/Services/Apps/AppService.cs
dennisreimann: Re-added the original implementation and kept the new one as an addition that we now use internally.

BTCPayServer/Services/Apps/AppService.cs
```diff
@@ -427,6 +427,29 @@ public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)]
             }
             return true;
         }
+        
+        public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)] out List<PosCartItem> cartItems)
```
Kukks, line 431: Let's do it later, but to reduce double code we can make `public static bool TryParsePosCartItems(JObject? posData, [MaybeNullWhen(false)] out Dictionary<string, int> cartItems)` use this method internally

PR #5239, opened 2023-08-07
Receipt improvements

### LINKED ISSUE #5196: [Bug]: PoS receipt order id links to PoS

BTCPay 1.11.0-rc4

docker

### What happened?

On the PoS receipt page there is the new unique order ID, but the order ID link is pointing to the PoS URL.

Not sure if there needs to be a link at all or it should link to the receipt page? 

image

Paid PoS invoice, clicked on receipt

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Removes the link from the order ID and adds it to the bottom of the page as a "Return to store" button. Also adjusts the first column width of the POS data, to give it a bit more space.

Closes #5196.

grafik

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -123,6 +123,7 @@ namespace BTCPayServer.Controllers
             var additionalData = metaData
                 .Where(dict => !InvoiceAdditionalDataExclude.Contains(dict.Key))
                 .ToDictionary(dict => dict.Key, dict => dict.Value);
+            
             var model = new InvoiceDetailsModel
             {
                 StoreId = store.Id,
@@ -149,7 +150,6 @@ namespace BTCPayServer.Controllers
                 StatusException = invoice.ExceptionStatus,
                 Events = invoice.Events,
                 Metadata = metaData,
-                AdditionalData = additionalData,
                 Archived = invoice.Archived,
                 CanRefund = invoiceState.CanRefund(),
                 Refunds = invoice.Refunds,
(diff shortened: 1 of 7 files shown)


pavlenex:
Concept ACK

dstrukt:
YES!

Have some thoughts on how we should frame this via the wording, will give more thorough feedback this evening.

Immediate thoughts are the "Summary" should be ordered before the Payment Details, and we should format the the section similarly to how we deal with the Cart, i.e:

```item - quantity -cost (1x)
item - quantity -cost (2x)
discount
tip
total```

dennisreimann:
The additional data isn't structured in a way that would allow to sort/display it in a special way, because this isn't just for the cart/POS, but rather a very generic key/value list.

dstrukt:
Ahhh I see, it seems like it would be worth the effort to format that data into a more digestible format though, no? Maybe not in this PR, but for a future iteration. Itemized receipts for both post-payment and email are as valuable as the custom form, information collection imo.

dennisreimann:
Ok, let 's first get #5238 merged and then I'll update this one.

dstrukt:
Looking much better!

I think the 
`Item - Quantity - Price - Total` is a bit more digestible, but if we're still limited or the LOE at this time is too much, what we have is definitely sufficient, e.g:

Screen Shot 2023-08-09 at 11 34 59 AM

Also for the Invoice Detail view, we had planned to update this, so will hold back comments on this, but was wondering if showing both the receipt and the POS data (or Additional Data in this case) is still necessary?

dennisreimann:
> Also for the Invoice Detail view, we had planned to update this, so will hold back comments on this, but was wondering if showing both the receipt and the POS data (or Additional Data in this case) is still necessary?

POS data can be any kind of structured data that comes via the API, so need to keep it in there.

Kukks (approved):

PR #5251, opened 2023-08-11
Improve invoice filter wording

### LINKED ISSUE #5250: reposting here if it got missing on mattermost
reposting here if it got missing on mattermost

The problem is fixed, thank you. But there is another problem now :innocent: 

When you select e.g. Settled partial but there is no result the right part of the dropdowns disappear (the info icon + Export and Actions dropdown)

Also another bug I missed is the "Archived" filter does not do anything, just reloads the page with same results; also if you actually have archived invoices 
(this bug is also present in 1.11.1 and nothing introduced by the fix)

_Originally posted by @ndeet in https://github.com/btcpayserver/btcpayserver/issues/5246#issuecomment-1673979957_

dennisreimann:
Closes #5250, which is mostly a wording issue I think: The archived filter is now named "Include archived" and already did just that …

grafik

### DIFF
--- a/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
+++ b/BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
@@ -234,7 +234,7 @@
     </div>
 </div>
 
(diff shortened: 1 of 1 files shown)


ndeet (approved):
Agree with the new wording it makes more sense. So this is not an exclusive filter but just adding the archived ones to the result set. :+1:

PR #5252, opened 2023-08-11
POS: Unify item display

dennisreimann:
Display unifications for static and cart view, which @prusnak made me aware of.

## Cart

grafik

## Static

grafik


prusnak:
CI is failing

Issue #5253, opened 2023-08-11
When clicking on `view` payment request link it adds store ID in the link which is unnecessary 


pavlenex:
When you create a new payment request and click on the link to preview it it it appends `storeId=xxxxx` and makes link you're sharing with a potential client uglier, there's no need to append store id there.

![Screenshot 2023-08-11 at 16 32 09](https://github.com/btcpayserver/btcpayserver/assets/36959754/5308b025-dd9d-49e3-97a2-0f6f07db1f4c)

dstrukt:
I'll take this one.

PR #5256, opened 2023-08-11
Remove Store ID from view request URL

dstrukt:
Really simple change.

closes: https://github.com/btcpayserver/btcpayserver/issues/5253


dennisreimann (approved):
Thanks for taking care here!

TheEmpress11 (commented):
Great

Issue #5266, opened 2023-08-21
[Bug]: Pay in Wallet button does not work with certain browsers


strnly:
### What is your BTCPay version?

BTCPay Server v1.11.2

### How did you deploy BTCPay Server?

Running on a VPS. CLN instance

### What happened?

https://chat.btcpayserver.org/btcpayserver/pl/xxd89e7gi7gk9nw4c9jpp6pjzr

When opening an invoice in modal, the "Pay in Wallet" button will not work with Chromium, Brave and Safari.

Likely a scheme compatibility issue on BTCPay's side when in modal using Chromium, Brave and Safari. Possibly something analogous to the issue with the QR code not appearing, that we figured a couple months ago with @dennisreimann 

It works fine using Firefox. 

Tested https://wordpress.demo.btcpay.tech/ and I could replicate the same bug that I experience on my BTCPay server

This is a known issue fixed by @ndeet in the past and that came back with recent release. 

### How did you encounter this bug?

https://chat.btcpayserver.org/btcpayserver/pl/xxd89e7gi7gk9nw4c9jpp6pjzr

When opening an invoice in modal, the "Pay in Wallet" button will not work with Chromium, Brave and Safari.

Likely a scheme compatibility issue on BTCPay's side when in modal using Chromium, Brave and Safari. Possibly something analogous to the issue with the QR code not appearing, that we figured a couple months ago with @dennisreimann 

It works fine using Firefox. 

Tested https://wordpress.demo.btcpay.tech/ and I could replicate the same bug that I experience on my BTCPay server

This is a known issue fixed by @ndeet in the past and that came back with recent release. 

### What browser do you use?

Chromium, Brave and Safari.

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

ndeet:
I tested Brave on Linux and Brave and Safari on MacOS using checkout v2 on that wordpress demo site of mine connected to BTCPay 1.11.2. It worked as expected.

ndeet:
The connected BTCPay server to that wordpress demo site is also running CLN.

strnly:
In modal ?

ndeet:
Yes on my testsite https://wordpress.demo.btcpay.tech

Also checked now on Android with DuckDuckGo and Chrome and it worked as expected.

Would be great if you could open up developer tools or javascript console to check for any error. Without beeing able to reproduce some more information would help.

[...]

PR #5270, opened 2023-08-22
Unset link targt for payment scheme links

dennisreimann:
Potential fix for #5266 — see the discussion in that issue for details.

This change should be non-invasive, I tested the links in regular as well as modal mode and they worked in Firefox, Brave and Chrome.

### DIFF
--- a/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
@@ -24,7 +24,7 @@
                 <label v-t="'lightning'"></label>
             </div>
         </div>
-        <a v-if="model.invoiceBitcoinUrl && model.showPayInWalletButton" class="btn btn-primary rounded-pill w-100 mt-4" target="_top" id="PayInWallet"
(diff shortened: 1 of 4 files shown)


Esduke:
How much do I have in my account ???

dennisreimann:
Yes, I tested it and maybe @ndeet can also give uit another pair of eyes to confirm.

It was weird that the links had the target attribute in the first place, because usually we'd only use `_top` if in a modal.

ndeet:
@strnly I added the case of no target at all as suggested by @NicolasDorier, would be great if you could confirm or deny it works too:
https://jsbin.com/bajebezuco/edit?html,output

ndeet:
> It was weird that the links had the target attribute in the first place, because usually we'd only use _top if in a modal.

The issue was actually when using it in a modal context when integrating btcpay.js to show the modal but in this case the modal is already open and it makes no sense to make the button open another modal? I think the safe bet here is to use target="_blank" because it would actually open an external app so a new browser window makes most sense? _top afaics would try to open it on the body of the current page which is not what we want for sure?

strnly:
> @strnly I added the case of no target at all as suggested by @NicolasDorier, would be great if you could confirm or deny it works too: https://jsbin.com/bajebezuco/edit?html,output

NO TARGET works fine. Thanks!

PR #5273, opened 2023-08-24
Small improvements

Kukks:
* Bump LNURL
* how app view link in nav when not enough permission to modify
* small readability improvement on CanRefund
* Make Create pull payment action usable if create pull payment permission is available
* Pass invoice id in context as plugin relied on that data in route value which is gone now (after lnaddress optimizations)

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -54,7 +54,7 @@
     <PackageReference Include="Fido2" Version="2.0.2" />
     <PackageReference Include="Fido2.AspNet" Version="2.0.2" />
     <PackageReference Include="HtmlSanitizer" Version="5.0.372" />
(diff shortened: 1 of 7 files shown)


NicolasDorier (changes requested):
in `NewPullPayment` you need to check if `CanCreatePullPayments` is granted if `AutoApproveClaims` is true. The `permission` tag helper is only affecting the UI.

Kukks:
> in `NewPullPayment` you need to check if `CanCreatePullPayments` is granted if `AutoApproveClaims` is true. The `permission` tag helper is only affecting the UI.


It's handled here
https://github.com/btcpayserver/btcpayserver/pull/5273/files#diff-4aece436762bb74d8745a92c07974b0959bf860d9b24aa778817ea129a75303aR141

PR #5275, opened 2023-08-24
Parse POS string data for invoice details display

### LINKED ISSUE #5240: [Bug]: Invoice details pos metadata not parsed anymore

BTCPay 1.11.1 (same with 1.10.3)

docker

### What happened?

The PoS metadata set via API was always rendered in a nice way in the past, this seems to have gone missing. Not sure if deliberately or just a sideeffect of another change.

image

- Set PoS metadata via the API
- Visit invoice details page

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Fixes #5240 and adds some display improvements as well.

## Before

grafik

## After

after

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -172,6 +172,20 @@ namespace BTCPayServer.Controllers
                 model.ReceiptData = (Dictionary<string, object>)additionalData["receiptData"];
                 additionalData.Remove("receiptData");
             }
+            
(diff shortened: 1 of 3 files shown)


Kukks (approved):

dstrukt (changes requested):
Looks great - much cleaner!

Only minor nits would be formatting the `posData` label to `Point of Sale` or `Point of Sale Data` and potentially pushing the Point of Sale (WooCommerce, etc..) to the right to make the key/pair view consistent, if possible.

Issue #4936, opened 2023-04-28
[Bug]: Payment incorrectly marked as "settled" in csv after payjoin error


Liongrass:
### What is your BTCPay version?

BTCPay Server v1.9.2+6193835ea 

### How did you deploy BTCPay Server?

Lunanode installer

### What happened?

I'm using BTCPayServer to process on-chain and off-chain payments. I have payjoin enabled. A customer tried to make a payjoin payment but it failed (the payment was never made), yet in the csv the payment shows up as paid, although the invoice is marked as expired. The relevant line in the csv can be found below.

### How did you encounter this bug?

1. Create invoice
2. User tries to pay with Bluewallet
3. User sees error message and lets invoice time out
4. Invoice is marked as expired in the BTCPay dashboard, but the csv marks the amount as paid

### Relevant log output

```shell
ReceivedDate	StoreId	OrderId	InvoiceId	InvoiceCreatedDate	InvoiceExpirationDate	InvoiceMonitoringDate	PaymentId	Destination	PaymentType	CryptoCode	Paid	NetworkFee	ConversionRate	PaidCurrency	InvoiceCurrency	InvoiceDue	InvoicePrice	InvoiceItemCode	InvoiceItemDesc	InvoiceFullStatus	InvoiceStatus	InvoiceExceptionStatus	BuyerEmail	Accounted
```

```
04/28/2023 02:30:57 | HoyPithshvkRMADckPTCx1ZmvUmbZNYwPTjeKyBX2CED | pos-app_4C3Ei2pnpG463BAMwyHwBNhMaHvW | PKCxdf4wjJt37FeGqDMBq4 | 04/28/2023 02:27:21 | 04/28/2023 02:42:21 | 04/29/2023 02:42:21 | 593206df1398bd99d614565328c43a312945be4bcc518f2508030395dcfbbd7a-0 | bc1qyysm5qleqrxc38fqp6ml8wx85f3rpd9ac36t9g | On-Chain | BTC | 0.0021 | 0 | 100000000 | 210000 | SATS | 0 | 210000 | join | Join | Expired | expired |   | redacted@gmail.com | False
```
The lifecycle of this invoice:

```
27/4/2023, 7:27 pm | Creation of invoice starting
-- | --
27/4/2023, 7:27 pm | BTC_SATS: The rating rule is 1 / 0.00000001
27/4/2023, 7:27 pm | BTC_SATS: The evaluated rating rule is 1 / 0.00000001
[...]

PR #5293, opened 2023-09-06
Remove legacy confusing export

NicolasDorier:
This remove the export feature in the invoice list that was very confusing. Some of confusion show up #4936
We will need to rewrite https://docs.btcpayserver.org/Accounting/ to document the reporting feature.

Instead of export, you see now a link to the various reports

image


dennisreimann (approved):
concept ACK.

PR #5296, opened 2023-09-06
Archive stores and apps

dennisreimann:
The ability to archive stores got requested at Baltic Honeybadger by our friends from the HRF (@freedomisuniversal). Archiving for apps got requested by Satoshi Store (@rolandschellhorn) .

# Archived Stores

archived-stores

## Archived apps

Once an app is archived, it is also no more publicly available.

archived-apps

archived-app

### DIFF
--- a/BTCPayServer.Client/Models/CreateAppRequest.cs
+++ b/BTCPayServer.Client/Models/CreateAppRequest.cs
@@ -37,6 +37,7 @@ namespace BTCPayServer.Client.Models
         public string RedirectUrl { get; set; } = null;
         public bool? RedirectAutomatically { get; set; } = null;
         public bool? RequiresRefundEmail { get; set; } = null;
+        public bool? Archived { get; set; } = null;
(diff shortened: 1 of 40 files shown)


dstrukt:
WRT archiving "apps" which I'm a bit confused why/if we're still referring to anything as apps anymore, are we not putting the Archived tab underneath the Plugin specifically because the Plugin (i.e. POS) can be uninstalled, and the archives will no longer be visible? Just trying to understand the thinking of dropping them to the bottom of the list.

NicolasDorier:
I merged this one, as the backend is done and unlikely to change due to @dstrukt comments. The UX comments can be adressed on another PR. Ping @dennisreimann

dennisreimann:
> Just trying to understand the thinking of dropping them to the bottom of the list.

I had to put it somewhere and this felt like the best spot. Nevertheless we can address this in another PR, especially if you think there are wording improvements as well.

PR #5312, opened 2023-09-11
Add Blazor server

NicolasDorier:
This PR focus in adding Blazor server as a dependency. It will make the development of components easier.
Blazor is inserted in the `_Footer.cshtml`, so it will be available to all the backend code.

## UX Challenge for Blazor

Blazor is a front end technology from Microsoft that will make the development of component way easier and modular.
However, all Blazor components instantiated in a Razor page have their state both on the client and on the server for each user.

The state of blazor components on the server is called "Circuit".

If the user disconnect or navigate to a new page, then the circuit get removed from server's memory after 3 minutes, and a new circuit get created on the new page.

When the user is disconnected from the server, then any blazor components can't be interacted with (let's say a client try to click on a button from a blazor component, then nothing will happen).

If the user reconnect before 3 minutes, when it is reconnected, it will work again as expected.

If the user doesn't reconnect before 3 minutes, then the circuit on the server get destroyed and the user need to refresh the page of his browser to make a new circuit.

So we need a non intrusive way to tell to the user "Hey, you aren't connected to the server anymore things might not work anymore, we are trying to reconnect...", or "Hey we can't reconnect to the server, please refresh your page.".

I added an indicator in the bottom.
image

I am not sure @dstrukt  would approve this.

Note: It isn't possible to create a new circuit without refreshing the page. (due to encryption of blazor components on the server rendered page)

## Notes on several hacks to make it works

I documented two hacks I needed to do in `Startup.cs` in order for blazor to work without side effect.

One hack was needed because Blazor need the default StaticFile middleware, but we aren't using any default. They fixed it in .NET 8.0, so we'll be able to remove it later. There is a github thread about it on https://github.com/dotnet/aspnetcore/issues/19578 

The other hack was needed because blazor assumes the page is the root page of the website. They documented a way to work around with the `<base>` directive. However, we shouldn't use this, as this may impact code which reasonably assume that the paths in HTML are relative to the current page. (ie: If I am on https://btcpay/store/A then a `<a href="Settings"` refers to `https://btcpay/store/A/Settings`. Or a `<form>` without action, assume the same action as the current page)
Github thread about it is https://github.com/dotnet/aspnetcore/issues/43191

## Security notes

Blazor server makes it easy to run denial of services by creating lot's of circuit. (ie: Refresh the page many time)
Also, Blazor server exposes to Javascript some callable dotnet code. As such, I believe it is safer to not expose Blazor to unauthenticated user.

This mean we can't use Blazor server for unauthenticated pages such as point of sale.

### DIFF
--- a/BTCPayServer/Blazor/_Imports.razor
+++ b/BTCPayServer/Blazor/_Imports.razor
@@ -1,0 +1,7 @@
+@using System.Net.Http
+@using Microsoft.AspNetCore.Authorization
+@using Microsoft.AspNetCore.Components.Authorization
+@using Microsoft.AspNetCore.Components.Forms
+@using Microsoft.AspNetCore.Components.Routing
+@using Microsoft.AspNetCore.Components.Web
+@using Microsoft.JSInterop
--- a/BTCPayServer/Hosting/Startup.cs
+++ b/BTCPayServer/Hosting/Startup.cs
@@ -25,6 +25,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Rewrite;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -56,7 +57,6 @@ namespace BTCPayServer.Hosting
         }
         public ILoggerFactory LoggerFactory { get; }
         public Logs Logs { get; }
-
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddMemoryCache();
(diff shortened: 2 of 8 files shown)


dennisreimann:
> I added an indicator in the bottom.

We can do it more unobtrusively by showing an overlay in case of disconnect. This overlay can guide the user towards refreshing and be hidden as long as the connection persists.

NicolasDorier:
@dennisreimann the overlay is the default, but it is disruptive.
I don't believe we should bother the user too much about it, users are already used to refresh browser to see updated data.
For example, if the user has the transaction list of a wallet open, and blazor push new transactions automatically to the UI, it isn't really a problem if the connection is lost. You still want to let the user interact with the page (moving to another page and copy/pasting stuff.)

NicolasDorier:
Another idea is to show an alert only if there is a disconnection.
(I'd like to put it on the top over all the width, but I don't know how)

image

dennisreimann:
> Another idea is to show an alert only if there is a disconnection.

That's what I meant. I can take care of the styling part.

dennisreimann:
## Desktop

desktop

## Mobile

mobile

dennisreimann (approved):

NicolasDorier:
I slightly improve it.

* The `Please refresh the page`, now `refresh` is a clickable link that will refresh the page.
* Every times it was retrying to reconnect, the popup would pop up again even if it wasn't closed. Now I don't pop the toaster if it's already open.

PR #5313, opened 2023-09-11
Wallet Receive: Update address formatting

### LINKED ISSUE #5311: UX: Address formatting shouldn't truncate the end of the address
The address below is tb1q8u3t0rswaddwh09lz2k87y4v3qg4c94vgv5kwn

It is hard to compare, because the address is truncated.
In general people compares the few first letter and the last letters.

If possible, it should be truncated in the middle (say `tb1q8u3t0rsw....v3qg4c94vgv5kwn`)

image

dennisreimann:
Closes #5311.

## Before

before

## After

after

### DIFF
--- a/BTCPayServer/Views/UIWallets/WalletReceive.cshtml
+++ b/BTCPayServer/Views/UIWallets/WalletReceive.cshtml
@@ -70,12 +70,9 @@
                     </div>
                     <div class="input-group mt-3">
                         <div class="form-floating">
-                            <input id="PaymentLink" class="form-control-plaintext" readonly="readonly" value="@Model.PaymentLink">
-                            <label for="PaymentLink">Payment Link</label>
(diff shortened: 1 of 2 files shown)


Kukks (approved):

NicolasDorier:
Checking if I can't fix this small gap

image

NicolasDorier:
What do you think of instead just truncating exactly the number of characters?
The ellipsis not being in the middle feel a bit weird.

dennisreimann:
It is done this way to handle flexible width scenarios. Imho it's good enough and solves the problem – we're using this in most places for a while now and no one complained. 😀

dstrukt (approved):
tACK - LGTM!

Agree with @NicolasDorier on the minor nit of removing the extra space.

PR #5314, opened 2023-09-11
Email Rules: Add default texts and document placeholders

dennisreimann:
Applies default subject and body text on editing to simplify email rule setup. Once the text is edited manually, the defaus  aren't applied on switching the rule type.

Also documents the placeholders that can be used.

emil-rules


dstrukt (approved):
tACK - nice quality of life improvement!

PR #5322, opened 2023-09-13
Plugins: Add hook for resolving Lightning Address

dennisreimann:
Allows plugins to provide Lightning Addresses, which aren't provided on the store level already. If the LN Address lookup in the core database fails, the plugin hook is invoked with the username portion, which can then resolve the LNURL-Pay request for the Lightning Address.

Using this, I can provide Lightning Addresses for LNbank wallets. The counterpart in LNbank looks like this:

```csharp
namespace BTCPayServer.Plugins.LNbank.Hooks;

public class ResolveLightningAddressHandler : IPluginHookFilter
{
    public string Hook { get; } = "resolve-lnurlp-request-for-lightning-address";

    private readonly WalletRepository _walletRepository;
    private readonly LinkGenerator _linkGenerator;
    private readonly IHttpContextAccessor _httpContextAccessor;

    public ResolveLightningAddressHandler(
        IHttpContextAccessor httpContextAccessor,
        WalletRepository walletRepository,
        LinkGenerator linkGenerator)
    {
        _httpContextAccessor = httpContextAccessor;
        _walletRepository = walletRepository;
        _linkGenerator = linkGenerator;
    }

    public async Task<object> Execute(object args)
    {
        var obj = (LightningAddressResolver)args;
        var username = obj.Username;
        var wallet = await _walletRepository.GetWallet(new WalletsQuery
        {
            LightningAddressIdentifier = new[] { username },
        });
        if (wallet == null) return obj;

        var request = _httpContextAccessor.HttpContext.Request;
        var metadata = new Dictionary<string, string>
        {
            ["text/identifier"] = $"{username}@{request.Host}"
        };
        obj.LNURLPayRequest = new LNURLPayRequest
        {
            Tag = "payRequest",
            CommentAllowed = 2000,
            Metadata = JsonConvert.SerializeObject(metadata.Select(kv => new[] { kv.Key, kv.Value })),
            Callback = new Uri(_linkGenerator.GetUriByAction(
                action: nameof(LnurlController.LnurlPay),
                controller: "Lnurl",
                values: new { walletId = wallet.WalletId }, request.Scheme, request.Host, request.PathBase) ?? string.Empty)
        };

        return obj;
    }
}
```

### DIFF
--- a/BTCPayServer/Plugins/LightningAddresssResolver.cs
+++ b/BTCPayServer/Plugins/LightningAddresssResolver.cs
@@ -1,22 +1,14 @@
 using LNURL;
-using Microsoft.AspNetCore.Http;
 
 namespace BTCPayServer.Plugins;
 
 public class LightningAddressResolver
 {
     public string Username { get; set; }
-    public HttpContext HttpContext { get; }
-    public LNURLPayRequest LNURLPayRequest { get; private set; }
+    public LNURLPayRequest LNURLPayRequest { get; set; }
 
-    public LightningAddressResolver(HttpContext httpContext, string username)
+    public LightningAddressResolver(string username)
     {
-        HttpContext = httpContext;
         Username = username;
-    }
-
-    public void ResolveLNURLPayRequest(LNURLPayRequest lnurlPayRequest)
-    {
-        LNURLPayRequest = lnurlPayRequest;
     }
 }
--- a/BTCPayServer/Plugins/LightningAddresssResolver.cs
+++ b/BTCPayServer/Plugins/LightningAddresssResolver.cs
@@ -1,0 +1,22 @@
+using LNURL;
+using Microsoft.AspNetCore.Http;
+
+namespace BTCPayServer.Plugins;
+
+public class LightningAddressResolver
+{
+    public string Username { get; set; }
+    public HttpContext HttpContext { get; }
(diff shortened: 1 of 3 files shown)


BTCPayServer/Plugins/LightningAddresssResolver.cs
```diff
@@ -0,0 +1,22 @@
+using LNURL;
+using Microsoft.AspNetCore.Http;
+
+namespace BTCPayServer.Plugins;
+
+public class LightningAddressResolver
+{
+    public string Username { get; set; }
+    public HttpContext HttpContext { get; }
```
Kukks, line 9: I would exclude the HttpContext since you can almost always use the HttpContextAccessor

Kukks (changes requested):

BTCPayServer/Plugins/LightningAddresssResolver.cs
dennisreimann: Updated in [cc89e23](https://github.com/btcpayserver/btcpayserver/pull/5322/commits/cc89e230dbfa5cd2f6f3390d57e04448d43f88b0)

Kukks (approved):

Djritz04 (approved):

PR #5325, opened 2023-09-15
Rewrite the Notification dropdown with Blazor

NicolasDorier:
This PR re-developed the Notification Dropdown by using Blazor (when we click on the bell icon on the top left left).

The move to blazor bring improvements:
* The page will load faster, because on the pre-rendering phase (when the server generate the HTML page), we aren't fetching anything from database. Instead, information is fetched only once the page is loaded.

I also added a small improvement:
* We are capping the number of unread notification to `99`. If the user has more, we will show `99+`.
* If there is more than 99 notification, we will automatically mark as read anything more than 100. This is because our request to fetch unseen notification isn't really optimized, as such we should keep the number low.

I have been happy with Blazor.

## Some Blazor lessons

Each components have 4 different rendering.

1. The **pre-rendering**: This is when the server generates the HTML for the first request. It happens before a circuit is created, as such the component state at this stage isn't saved server side. (I use undocumented hack `JSRuntime.IsPreRendering()` for detecting this)
2. The **first rendering**: This is when the client create the Blazor circuit. It then call `OnParametersSetAsync()` if this call complete synchronously, this will be the only rendering. If the call isn't complete (database call), the rendering happens immediately **without waiting for the call to complete**.
3. The **second rendering**: If `OnParametersSetAsync()` isn't completed synchronously (says because there is a database call), the second rendering will happen when the call finally complete.
4. **Manual renderings**: You can call `StateHasChanged()` to ask for a re-rendering of the component if its state changed. I am using this when I see our `EventAggregator` has some events about updated notification. You need to wrap the call with `InvokeAsync`.

The first rendering and second rendering is documented on the [lifecycle of components](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle).

I tried to document `OnParametersSetAsync` to explain the situation.
* I do not make blocking call if this is a pre-rendering.
* The first rendering and the pre-rendering uses cached value
* The second rendering happen after database fetch


dennisreimann (approved):
tACK. Nice! 
grafik

PR #5332, opened 2023-09-20
API: Document payment method IDs

dennisreimann:
This seems to be a source of confusion (see e.g. #5330), so I thought it'd be best to document the payment method IDs as an enum, so that we can refer to it in the several places they are used.

pmids


NicolasDorier:
The problem with putting an enum is that automated tools will constrain the input as values in the enum.

But we can't enumerate all possibilities. You have LTC-OnChain for example. Maybe more in the future provided by plugins.

dennisreimann:
Alright, simplified it.

grafik

PR #5350, opened 2023-09-25
Improve receipt info display

dennisreimann:
Displays the info in correct order and adds optional info if tip was given with a percentage.

## Before

before

## After

after


ndeet:
Much better readable compared to the previous versions. You got to try figure out how those numbers match up for a few minutes :laughing:

dstrukt (approved):
tACK, much better!

PR #5357, opened 2023-09-27
Email Rules: Require either recipients or customer email option

### LINKED ISSUE #5345: [Bug]: Email rules, Recipients required even if "send mail to buyer"?

BTCPay 1.11.5-rc1

docker

### What happened?

On the Email Rules page you have the option "Send the email to the buyer, if email was provided to the invoice" 

I wonder if we should make the above "Recipients" field non-required if the option is checked (and do not send any mail if there is no buyer email available) otherwise you will always send a copy to yourself which could be kinda anoying with lots of invoices.

image

Go to Store -> Settings -> Emails -> Email rules

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Fixes #5345.

grafik

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.Email.cs
+++ b/BTCPayServer/Controllers/UIStoresController.Email.cs
@@ -62,6 +62,14 @@ namespace BTCPayServer.Controllers
 
                 return View(vm);
             }
+
+            for (var i = 0; index < vm.Rules.Count; index++)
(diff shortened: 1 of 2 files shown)


dstrukt (approved):
tACK

Kukks (approved):

PR #5360, opened 2023-09-28
Invoice: Unify status display and functionality

dennisreimann:
Consolidates the invoice status display and functionality (mark setted or invalid) across the dashboard, list and details pages.

## Dashboard

dashboard

## List

list

## Details

details

detals

### Before

grafik

PR #5362, opened 2023-09-28
Invoice: Improve payment details

### LINKED ISSUE #5207: [Bug]: Overpaid shows amount in all cryptocurrencies, not just the one which was used to pay with

Running 1.11.0

Lunanode docker

### What happened?

When an invoice is paid with Monero and overpaid, it shows correctly amount paid, due, but not overpaid for BTC:

image

There was NO overpayment in BTC, so it should say 0 on that line.

Same thing happens vice versa, when overpaid in BTC, it shows the amount in Monero as well:

image

Happens when multiple cryptocurrencies are accepted and there is an overpayment

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Clearer description and display, especially for overpayments:

- Makes overpayment case more prominent and adds a sentence about the total excess amount if more than one payment method was used, which addresses the initial issue.
- Highlights paid and overpaid amounts.
- Displays Total Due value per payment method.
- Omits Still Due column if the invoice is fully paid.
- Omits overpaid/paid values for payment methods that weren't used.  
- Hides Rates column if there's no fiat rate.

Closes #5207.

## Before

before

## In List Table

grafik

## Regular invoice, no overpayment

grafik

### DIFF
--- a/BTCPayServer.Tests/ThirdPartyTests.cs
+++ b/BTCPayServer.Tests/ThirdPartyTests.cs
@@ -298,7 +298,7 @@ retry:
             var fetcher = new RateFetcher(factory);
             var provider = new BTCPayNetworkProvider(ChainName.Mainnet);
             var b = new StoreBlob();
-            string[] temporarilyBroken = { "UGX" };
+            string[] temporarilyBroken = { "COP", "UGX" };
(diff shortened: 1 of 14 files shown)


dennisreimann:
This will also play nicely with #5360.

PR #5363, opened 2023-09-29
Reporting: Improve rounding and display

### LINKED ISSUE #5193: [Bug]: Reporting: rounding and output issue

BTCPay 1.11.0-rc4

docker

### What happened?

**1. Rounding issue:**
total of lightning address seems to have a rounding isssue 
image

**2. Formatting issue**
Here the 0,00000001 are output as 1e-8 which also ends up in the CSV export. Imo we should enforce decimals here so it is consistent and not make the consumer of the export handle formatting of the values? 
image

**3. Inconsistency / clarification on "products sold" export:**
image

1)  products sold:
"sold" a bit misleading as it shows unconfirmed invoices with status new? (or maybe do not show state new at all here?)

2) state + quantity:
state "new" is ok if it should be shown at all, in this case but the quantity is not, there are 3 rooibos sells open 

3) raw data:
also misses the other new invoices, seems to only count it once per product?

To discuss/decide:
question is if new invoices should be shown at all;

also if e.g. 2 rooibos invoices expire and others are not the state "expired" is not shown at all, only state "new"; unsure if all states should be there or only a few? 

**4. CSV output thoughts (maybe non-issues)**
Currently the CSV output does not put strings into quotes and uses `,` as a separator. 
```
Date,InvoiceId,OrderId,PaymentType,PaymentId,Confirmed,Address,Crypto,CryptoAmount,NetworkFee,LightningAddress,Currency,CurrencyAmount,Rate
2023-03-31 12:15:51,STyQudaVQJDvzGepwmcMGs,729,On-Chain,72a2dfc9b5418663b2bfea1bc48aa24173948aead47bcd2c7821a89143a007d6-1,true,mwX1BypzYcFtfNMy3X1oquTtP7HPQWrjZh,BTC,0.00036571,0,,USD,10.19,27863>
2023-03-31 01:55:36,7J7hGs29id2BaY5TdBC5B4,731,On-Chain,04cc6904ed14c78a1b7418ac68fe633477bcb08a0d68c7eadfbdd073c7818bee-0,true,mks8bxfePzaUD8xAXYFzwBRewy8H4iVeVn,BTC,0.00071459,0,,USD,19.98,27960>
2023-06-27 08:54:24,JPqugmiUdQPtor7g4KLQjk,,Lightning,4b4953cc3ece9a6d0d6b64798d8bc7bbf3f35268a484fc3bca9ff00900a98239,true,,BTC,1e-8,0,test@testnet.demo.btcpay.tech,USD,0.00031,30718
2023-06-27 08:54:28,Mca5izZambMnamQBwKrDHT,,Lightning,5340bb680f7de19e25ddd373f014b33afcc5e7c29c03d5258ea36fe86f1bfbf8,true,,BTC,0.00196674,0,test@testnet.demo.btcpay.tech,USD,60.41,30718

```

First, no `"` for strings could cause issues if your product name on Products sold output has `,` in it as it would be interpreted as sparator if there is no automatic escaping happening on the export.

Second, using `,` as separator maybe problematic if big fiat values such as 10,000.30 could cause problems. This is a non issue if the export escapes the output or does no amercan style amout formatting at all. But maybe less error prone to use a `;` as separator? 

In new reporting feature

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Sums up and displays the amount values according to the currencies divisibility. 

Fixes #5193.

## Example

grafik


NicolasDorier:
I refactored the code:
1. Make the JSON for amount smaller
2. Always format with the invariant format (12345.00 not 12,345.00)

PR #5372, opened 2023-10-04
Greenfield: allow delete user by email too

Kukks:


dennisreimann (approved):
utACK, the [docs already include it](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Users_DeleteUser) like this.

PR #5388, opened 2023-10-11
Fix pay report

Kukks:


NicolasDorier:
I added 11 digits precision for lightning.

PR #5391, opened 2023-10-11
POS Cart: Horizontal scrollable filters

### LINKED ISSUE #5255: [POS] Horizontal scrollable filters on mobile
Takes up a considerable amount of space when there are more than 4-8 filters, and could hypothetically cover the whole screen with 12(?)+. Additionally, with search + keyboard, an item is barely visible as well.

We should consider a horizontal scrolling filter implementation OR potentially a dropdown/drawer selector if the horizontal scroller proves to be too finicky on a mobile browser.

IMG_0613

dennisreimann:
In case the list of categories is too long, we make the list scrollable instead of displaying them stacked.

scrollable

Closes #5255.


dstrukt (approved):
Looks great!

tACK on Desktop Chrome, FF & mobile iOS Safari & Chrome.

Have an old Android phone i've been meaning to dig out for more testing on the mobile app side as well.

PR #5396, opened 2023-10-12
POS Keypad: Add plus and change clear functionality

### LINKED ISSUE #5299: Add plus + functionality in Point of Sale
In practice it has been shown that merchants sometimes have multiple items that they're charging for, in this scenarios they'd had to use calculator or do it on top of their head. Instead a pos device can shave a simple + functionality that would allow merchant to add 2+2+3+4+55+6+7+12.30 and then generate an invoice.

In this scenario we don't really need `.` just replace it with `+` I think?

Some examples:
Screenshot 2023-09-07 at 11 46 31
IMG_1706

It would be worth checking how POS in fiat industry handle this and if they have dot as a decimal separator or is the flow more similar to Breez and WOS. 

I don't recommend tackling this issue until @dstrukt takes a look and decides how to approach this UX wise.

dennisreimann:
Updated the fucntionality and hope everything works intuitively now. A few notes:

* No more decimal input (`.`)
* The `+` adds an additional amount to the stack. If there is more than one amount, it is shown in the calculation.
* Removed the Clear button at the bottom in favor of the new `C` button
    * `C` clears the current value if there's input already, if not it removes the current input, so that you get to edit the previous input again
    * Double-clicking `C` clears out everything

Closes #5299.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -2218,7 +2218,7 @@ namespace BTCPayServer.Tests
             Assert.Contains("EUR", s.Driver.FindElement(By.Id("Currency")).Text);
             Assert.Contains("0,00", s.Driver.FindElement(By.Id("Amount")).Text);
             Assert.Equal("", s.Driver.FindElement(By.Id("Calculation")).Text);
-            Assert.True(s.Driver.FindElement(By.Id("ModeTablist-amount")).Selected);
+            Assert.True(s.Driver.FindElement(By.Id("ModeTablist-amounts")).Selected);
             Assert.False(s.Driver.FindElement(By.Id("ModeTablist-discount")).Enabled);
             Assert.False(s.Driver.FindElement(By.Id("ModeTablist-tip")).Enabled);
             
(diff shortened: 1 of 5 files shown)


pavlenex (approved):
tACK, I love how you handled the deletion and clearing out, good one 👍

dstrukt (approved):
tACK

iOS Safari & Desktop Chome/FF

Agree with @pavlenex, and love how you handled the addition with multi step Clear functionality, not just a giant sweep Clear, it's the current, then the previous of what appears to be 2 or 3 levels deep .. could make the argument we want to support 3 (+s) levels deep or maybe even 4 +s deep ..? but that's a minor bit I'm sure we'll get more feedback on in the future.

Feels much cleaner, great work!

Still looking for any bugs, but if I find any I'll open a new issue, this is ready to go!

dennisreimann:
@dstrukt it should support all levels as it always clears out the current, then goes back to the previous one – all the way down the stack. 

Double clicking C clears it out all at once.

dstrukt:
Ahh yes, I'm still playing around with it, and you're correct .. I think I accidentally misclicked a double click and it completely cleared.

Subtle, but very nice .. gotta be patient with clearing 6 entries :joy:

I guess users will learn that subtle distinction in time, but was thinking if there's an equally smart way of communicating the functionality.

PR #5399, opened 2023-10-13
Unify list views

### LINKED ISSUE #3473: Make `Payment Request` list view consistent with `Invoices` by adding mass action (Archive, Clone?)
**Describe the bug**
In Design call it's been discussed that we should aim towards consistency when it comes to invoices and payment requets lists.

**Additional context**

Screenshot 2022-02-17 at 18 20 23
Screenshot 2022-02-17 at 18 20 29

A follow up from #3471

dennisreimann:
Unifies the table display and most of the mass action functionality for:

- Invoices
- Payment Requests (no mass actions yet)
- Wallet Transactions
- Pull Payments (no mass actions yet)
- Payouts

This is mostly generalizations (functionality and display code moved to site.js and site.css), unifications (e.g. removing the plus icon from add buttons) and legacy cleanups (`sticky-header-setup` wasn't needed anymore).

Closes #3473.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -634,58 +634,56 @@ namespace BTCPayServer.Controllers
         [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanViewInvoices)]
         public async Task<IActionResult> MassAction(string command, string[] selectedItems, string? storeId = null)
         {
-            if (selectedItems != null)
+            IActionResult NotSupported(string err)
             {
-                switch (command)
-                {
-                    case "archive":
-                        await _InvoiceRepository.MassArchive(selectedItems);
-                        TempData[WellKnownTempData.SuccessMessage] = $"{selectedItems.Length} invoice{(selectedItems.Length == 1 ? "" : "s")} archived.";
-                        break;
+                TempData[WellKnownTempData.ErrorMessage] = err;
+                return RedirectToAction(nameof(ListInvoices), new { storeId });
+            }
+            if (selectedItems.Length == 0)
+                return NotSupported("No invoice has been selected");
 
-                    case "unarchive":
-                        await _InvoiceRepository.MassArchive(selectedItems, false);
-                        TempData[WellKnownTempData.SuccessMessage] = $"{selectedItems.Length} invoice{(selectedItems.Length == 1 ? "" : "s")} unarchived.";
-                        break;
-                    case "cpfp":
-                        if (selectedItems.Length == 0)
-                            return NotSupported("No invoice has been selected");
-                        var network = _NetworkProvider.DefaultNetwork;
-                        var explorer = _ExplorerClients.GetExplorerClient(network);
-                        IActionResult NotSupported(string err)
-                        {
-                            TempData[WellKnownTempData.ErrorMessage] = err;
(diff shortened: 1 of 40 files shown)


dstrukt:
Is this ready for review?

dennisreimann:
> Is this ready for review?

Yes, curious to hear what ou think :)

dstrukt (changes requested):
This looks fantastic, and the consistency updates are incredible!

# Still reviewing

Discussed a few things on the last design call, so documenting a few items

- [ ] Experiment with keeping the table headers fixed on scroll to keep the actions visible on long page scrolls

- [ ] As discussed, we should probably hide (mass) actions if there are no options available, in this case the `Cancelled` state for the Payouts

Screen Shot 2023-10-18 at 11 54 09 AM

## For discussion:

- Connecting the selection and the (mass) action is valuable and the direction to move, but  are the updated actions obviously enough in their current placement (top right of the table)?

dennisreimann:
Wow, the sticky header part was way trickier than expected, but got it working now :D

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -634,58 +634,56 @@ public async Task<IActionResult> ToggleArchive(string invoiceId)
...
-                {
-                    case "archive":
-                        await _InvoiceRepository.MassArchive(selectedItems);
-                        TempData[WellKnownTempData.SuccessMessage] = $"{selectedItems.Length} invoice{(selectedItems.Length == 1 ? "" : "s")} archived.";
-                        break;
+                TempData[WellKnownTempData.ErrorMessage] = err;
+                return RedirectToAction(nameof(ListInvoices), new { storeId });
+            }
+            if (selectedItems.Length == 0)
+                return NotSupported("No invoice has been selected");
```
dennisreimann, line 643: Pulled this up, because it can be applied to all mass actions.

BTCPayServer/Views/UIInvoice/ListInvoices.cshtml
```diff
@@ -99,41 +95,11 @@
...
-                        }
-                    });
-                })
-                .on("click", ".invoice-row", function (e) {
-                    const $invoiceRow = $(e.currentTarget);
-                    if ($(e.target).is("td")) {
-                        $invoiceRow.find(".selector").trigger("click");
-                    }
-                });
-        });
```
dennisreimann, line 130: Removed this because we can leverage Bootstrap's `collapse` functionality. The row click got generalized into site.js

Kukks (approved):
I reviewed it as best as I could, but this touches everything everywhere. Maybe we need to develop reusable components to make this less of a chore in the future, especially because plugins will need the same mods to be in line

dstrukt (approved):
tACK, Chrome FF Desktop .. and mobile iOS Safari, works pretty well on mobile despite the sizing/scrolling

Think this is as far as we can push this at the moment functionally. I might make some aesthetic, alignment, etc.. changes in the future, but agree with @Kukks, would be great to make a reuseable component so we can keep this functionality consistent where it make sense.

Great job!

dennisreimann:
I find it hard to see how we can turn it into a reusable component while dealing with the complexity/diversity of the content we want to display.

The changes this PR introduces unified the display and functionality with just a few custom `mass-action` table classes, which is about as lean as it gets. I don't see how we can abstract this without basically adding components for tables/rows/cells, which would simply preapply those classes. Imho the added complexity wouldn't yield much benefit.

PR #5405, opened 2023-10-16
Forms: Make zip code a required field in predefined address form

### LINKED ISSUE #5401: [Bug]: State and Zip fields are not set as required fields.

v1.11.6

LunaNode VPS

### What happened?

During checkout when customer is prompted for their shipping address, "State" and "Postcode" are not required fields. 

If left blank by the customer it can cause confusion for the merchant when shipping items to the correct address. 

Ensuring that, "State" and "Postcode" are required fields will save the merchant time in finding the correct shipping the address. 

[Field.Create("Postcode", "buyerZip", null, false, null),](https://github.com/btcpayserver/btcpayserver/blob/018e4c501d459a40041e0b56d708b9f93a994548/BTCPayServer/Forms/FormDataService.cs#L46-L47)

During customer checkout when prompted for shipping address. 

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Closes #5401.

### DIFF
--- a/BTCPayServer/Forms/FormDataService.cs
+++ b/BTCPayServer/Forms/FormDataService.cs
@@ -43,9 +43,9 @@ public class FormDataService
             Field.Create("Address Line 1", "buyerAddress1", null, true, null),
             Field.Create("Address Line 2", "buyerAddress2", null, false, null),
             Field.Create("City", "buyerCity", null, true, null),
-            Field.Create("Postcode", "buyerZip", null, false, null),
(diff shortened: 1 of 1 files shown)


dstrukt (approved):
tACK - America!

knocte:
nACK: there are countries without post-codes, e.g. Hong Kong

dennisreimann:
@knocte True, nevertheless it can be turned off easily and since in the majority of cases it'll be required, we decided to make this the default.

PR #5412, opened 2023-10-18
switch pos to metadata in invoice create view

Kukks:
Since posdata is a legacy thing, we may  as well expose the propert metadata that we recommend. Pos data fits inside the metadata so no functionality is lost.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -1260,6 +1260,19 @@ namespace BTCPayServer.Controllers
             model.CheckoutType = storeBlob.CheckoutType;
             model.AvailablePaymentMethods = GetPaymentMethodsSelectList();
 
+            JObject? metadataObj = null;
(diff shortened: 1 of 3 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -1272,17 +1285,31 @@ public async Task<IActionResult> CreateInvoice(CreateInvoiceModel model, Cancell
...
+
+                if (!string.IsNullOrEmpty(model.BuyerEmail))
+                {
+                    metadata.BuyerEmail = model.BuyerEmail;
+                }
+
+                if (!string.IsNullOrEmpty(model.OrderId))
+                {
+                    metadata.OrderId = model.OrderId;
+                }
```
dennisreimann, line 1307: This block is redundant, see above:

```suggestion
```

dennisreimann (commented):
Guess this was just a c&p thing?

NicolasDorier:
Remove the redundant block, but except this I think it's good.

dennisreimann (approved):

PR #5413, opened 2023-10-18
Payment Request: Improve public view

### LINKED ISSUE #4450: [Design] Improved Payment Request view
July 25, 2023 - Still a few smaller todos, working actively on them

# Motivation

It's been a minute since the original mocks were posted, so I thought I'd rework them a little bit given the latest improvements to the checkout experience, the receipt views, and an attempt to make them a little more responsive and email/print friendly.

## Questions:

- Should we be showing the Payment History?
- Think we can get away with a more minimal `N amount due on T` with the payment history being displayed below, didn't think there was a much value in seeing the progress bar, but still debating...would actually be something nice to carry over to the admin view instead. i.e:

Screen Shot 2023-07-24 at 10 08 57 PM

TODO:
- [ ] Add custom form back button to this view .. could be added to the top left view
- [ ] Add expires in back to the view
- [ ] Add states (Settled, Expired, etc...) to the views

## Empty State
Screen Shot 2022-12-15 at 2 29 03 PM
Screen Shot 2022-12-15 at 2 55 00 PM
Screen Shot 2022-12-15 at 2 29 20 PM

## Filled State
Screen Shot 2022-12-15 at 2 33 19 PM

## Error
Screen Shot 2022-12-15 at 9 54 24 PM

## Custom Input
Screen Shot 2022-12-16 at 9 25 17 AM

dennisreimann:
This is mostly done, but I'd wait with the payments section updates until #5399 is merged. 

I will follow up on this with another PR that unifies more of the public page aspects, especially the page wrap and tiles.

Closes #4450.

Screenshot 2023-10-18 at 14-39-08 Tasty Beef

Screenshot 2023-10-18 at 14-45-08 Tasty Beef

Screenshot 2023-10-18 at 14-11-14 Tasty Beef


dstrukt:
Damn, you're on a roll with these. Trying to keep up - reviews for both today!

dennisreimann:
@dstrukt I updated the payment history, now it's ready for review.

grafik

dstrukt:
Will give this a proper review and feedback this evening, as I'm sure I'll have a few little nits.

One thought that I've been debating wrt Payment History .. I continue to wonder if it ever just makes sense to show the Invoice or the Transaction data solely, as opposed to attempting to show both. Arguably you get the info you need glancing at the transaction info, but I suppose you're missing out on the Invoice ID for potentially tracking it down.. thoughts?

dennisreimann:
@dstrukt I think it makes sense the way it is right now: You have the invoice data as a high-level overview plus the transaction details to dig deeper. I think at least for the print view we'd need the latter anyways, kinda like as with the receipt.

Also keep in mind, thet there even might be multiple transactions connected with a single invoice. Whet I'm showing in the screenshot is the extreme case, where the customer also could specify the amount to pay, so that we ended up with multiple invoices here, paid for with multiple transactions. The most common/best case will still be one invoice with one transaction though.

dennisreimann:
Also pdated the amount display to show trailing zeros as we do elsewhere and to [fix this](https://chat.btcpayserver.org/btcpayserver/pl/ppxfrzrc9bbbxcy84wj1rtx7re).

## Unified receipt and payment request tables

grafik

grafik

pavlenex:
LGTM, one minor thing is that click copy link button doesn't have any feedback when clicked, so it's not obvious it's `copied`, we should add that and it should be consistent with what we have on other places.

Besides that I like the direction where you guys went with this, going forward I think we need to make print view and fields a bit more traditional, where people can add client information for example on the backend, and print view can look more like an actual traditional invoice. Also, I think eventually we'll inevitably have a discussion about renaming payment requests to payment links but all of that is unrelated to this hub.

dennisreimann:
> LGTM, one minor thing is that click copy link button doesn't have any feedback when clicked, so it's not obvious it's copied, we should add that and it should be consistent with what we have on other places.

Done in [97dff71](https://github.com/btcpayserver/btcpayserver/pull/5413/commits/97dff7174e48461e1955da2a8a49151da7e79758).

> going forward I think we need to make print view and fields a bit more traditional, where people can add client information for example on the backend, and print view can look more like an actual traditional invoice.

Let's add a new issue for that.

dstrukt (changes requested):
It looks fantastic!

## STILL REVIEWING

Sorry this one took a little longer, wanted to be thorough.

- [ ] We should stick with `description` or `memo` and be consistent on the edit view, I personally like `Memo`.

Screen Shot 2023-11-13 at 2 43 59 PM

- [ ] The `View` button on the list view should `target="_blank"` into a new window imo, functionality might have been removed when we adjusted the UI 

- [ ] If there is a `Due Date`, the default `.mb-4` under the amount + state row should be `.mb-2`, so it looks a little cleaner:

Screen Shot 2023-11-13 at 2 47 36 PM

## Thoughts:

- Should we put any other state than `Paid` or `Expired' on this view, e.g. the `Pending`, etc..? I think it adds confusion to see `Pending`.
- Following up on the above, maybe we should put `No Due Date` if there is none selected, which would make the margin for all `.mb-2` under the amount

dennisreimann:
@dstrukt Adjusted. And also made it so that from S on the Print and Copy Link buttons sit besides each other instead of being stacked. Imho that works well with the now always large Pay Invoice button.

grafik

dstrukt (approved):
tACK - LGTM - thanks for the edits!

PR #5418, opened 2023-10-22
POS and Crowdfund: Improve item editor

dennisreimann:
Makes it work the same way as the form editor: Drag and drop for reordering and inline editing without modal.

Peek 2023-10-22 13-49


dstrukt:
Love this change, will give it a proper review this evening!

I had been considering this for mobile (as well as the products list on the POS) sans the draggable aspect for the admin, however the version in the OP is closer to what the end use will see, so there are trade-offs for both I suppose...but something to consider:

Screen Shot 2023-10-25 at 12 59 04 PM

dstrukt (approved):
functional tACK. 

Still wondering if we want to do something closer to the mobile mocks, even on desktop, think it might make the sorting/editing a little easier .. but we can cross that bridge at another time I think, and it's probably a bit out of scope for this PR.

Also wondered if it does make sense on mobile (and how I'm approaching) to do a modal or a slide-in drawer to edit product details once clicked .. leaving all of the other functionality intact, mostly to prevent scrolling on mobile.

One nit - do you think the POS item should be updated if you hit save for the Point of Sale as well, as opposed to just `Apply` on the item? Minor functional piece, but something to consider if it's an easy edit?

Great improvement!

dennisreimann:
> we can cross that bridge at another time I think, and it's probably a bit out of scope for this PR.

Yes, let's discuss this separately as I can also see benefits in having the card match what the user gets to see.

> Also wondered if it does make sense on mobile (and how I'm approaching) to do a modal or a slide-in drawer to edit product details once clicked .. leaving all of the other functionality intact, mostly to prevent scrolling on mobile.

I think this would be doable, but also be a more involved change and I'm not sure if editing the items is a big use case on mobile. Keep in mind that the current solution with the modal isn't ideal on mobile as well and we haven't received any feedback about it being suboptimal on mobile yet. 

> One nit - do you think the POS item should be updated if you hit save for the Point of Sale as well, as opposed to just Apply on the item? Minor functional piece, but something to consider if it's an easy edit?

Not that easy, because we'd have to turn the whole page into a Vue app or at least make the outer form part of it.

dstrukt:
I see, all valid reasons and points for why it exists the way it does today. Just wanted to surface the thinking as we're evaluating how to best re-use components, code, etc.. on these views!

Seems GTG to me then!

Kukks (approved):

PR #5420, opened 2023-10-23
Update de-DE.json remove strange punctuation after "Lightning"

petzsch:
grafik

The "." in the screenshot is not required as it's not really a sentence.


dennisreimann:
As this is a generated file, I fixed it on Transifex. Thanks for the heads-up!

PR #5422, opened 2023-10-24
Support the new LN lib

Kukks:


BTCPayServer/Payments/Lightning/LightningSupportedPaymentMethod.cs
```diff
@@ -17,20 +17,20 @@ public class LightningSupportedPaymentMethod : ISupportedPaymentMethod
...
         {
 #pragma warning disable CS0618 // Type or member is obsolete
             if (string.IsNullOrEmpty(LightningConnectionString))
                 return null;
-            if (!BTCPayServer.Lightning.LightningConnectionString.TryParse(LightningConnectionString, false, out var connectionString, out var error))
-            {
-                throw new FormatException(error);
-            }
-            return connectionString;
+            // if (!BTCPayServer.Lightning.LightningConnectionString.TryParse(LightningConnectionString, false, out var connectionString, out var error))
```
Kukks, line 25: we have these checks in btcpay, but we should just remove the conn string support for grpc?

dennisreimann (approved):
Tested this with LNbank and it works. There was only one occurrence in the tests, where I was using the `LightningConnectionType` type, which I could simply replace with the `string` type to make it build and work again.

Kixunil:
Looks like this broke something - it instructs the users to drop the `certfilepath` argument which is definitely wrong and it fails to connect. However it still works in some other instance on regtest so I'm not really sure what's going on. Do you see anything obvious?

PR #5423, opened 2023-10-24
Update Passport tooltips

sethforprivacy:
Fixes some small issues in language used in the tooltip to properly match current language on Passport.

### DIFF
--- a/BTCPayServer/Views/UIStores/ImportWallet/File.cshtml
+++ b/BTCPayServer/Views/UIStores/ImportWallet/File.cshtml
@@ -55,7 +55,7 @@
     </tr>
      <tr>
         <td>Passport</td>
-        <td>Wallet ❯ Pair Wallet ❯ BTCPay ❯ microSD ❯ Save wallet file</td>
+        <td>Wallet ❯ Connect Wallet ❯ BTCPay ❯ Choose wallet type ❯ microSD ❯ Save wallet file</td>
     </tr>    
     </tbody>
 </table>
--- a/BTCPayServer/Views/UIStores/ImportWallet/File.cshtml
+++ b/BTCPayServer/Views/UIStores/ImportWallet/File.cshtml
@@ -55,7 +55,7 @@
     </tr>
      <tr>
         <td>Passport</td>
-        <td>Wallet ❯ Connect Wallet ❯ BTCPay ❯ Choose wallet type ❯ microSD ❯ Save wallet file</td>
+        <td>Wallet ❯ Connect Wallet ❯ BTCPay ❯ microSD ❯ Save wallet file</td>
     </tr>    
     </tbody>
 </table>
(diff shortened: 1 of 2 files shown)


Kukks (approved):

sethforprivacy:
@Kukks One question -- I tested Taproot export but BTCPay still shows Segwit addresses. Do you know if BTCPay currently supports Taproot xPUBs via QR/microSD?

sethforprivacy:
For reference, here is the exported format:

```
{"ExtPubKey": "xpubREDACTED", "MasterFingerprint": "REDACTED", "AccountKeyPath": "86'/0'/1'", "FirmwareVersion": "2.3.0"}
```

Kukks:
> @Kukks One question -- I tested Taproot export but BTCPay still shows Segwit addresses. Do you know if BTCPay currently supports Taproot xPUBs via QR/microSD?

we do support taproot, but there is no specific "X"pub format for taproot, unlike the xpub/ypub/zpub encoding for legacy,segwit and p2sh. 

The only way to provide an xpub and indicate it is taproot is to use output descriptors  (or take the xpub and add a -[taproot] at the end and feed that into btcpay)

sethforprivacy:
Good to know, thank you! Will get that fixed up on our end specifically for BTCPay then.

Kukks:
> AccountKeyPath

I see. Yes, that will be a very invalid config in btcpay. It will create addresses in segwit but on the taproot derivation path. 

PS. I would suggest adding a way to know it came from a passport device. Right now a passport wallet export is marked as a wasabi  wallet as it's the final catch all we have

sethforprivacy:
> > AccountKeyPath
> 
> I see. Yes, that will be a very invalid config in btcpay. It will create addresses in segwit but on the taproot derivation path.
> 
> PS. I would suggest adding a way to know it came from a passport device. Right now a passport wallet export is marked as a wasabi wallet as it's the final catch all we have

Any specific way you prefer that being passed? We can add that too at the same time.

Kukks:
> > > AccountKeyPath
> > 
> > 
> > I see. Yes, that will be a very invalid config in btcpay. It will create addresses in segwit but on the taproot derivation path.
> > PS. I would suggest adding a way to know it came from a passport device. Right now a passport wallet export is marked as a wasabi wallet as it's the final catch all we have
> 
> Any specific way you prefer that being passed? We can add that too at the same time.

maybe add  a "Source" property with value "Passport"

for the xpub, change "ExtPubKey" to "Descriptor" 
so it becomes something like
```
{
"Descriptor": "tr([FINGERPRINTHERE/86'/0'/0']XPUBHERE/0/*)"
"Source": "Passport"
}
```

sethforprivacy:
Will take that to the team to add in and test heavily, as it's a big change from how we normally export data. Thanks for all of the details!

Once we release Taproot export support I'll open another PR to update the tooltips appropriately. For now this one is accurate and good to go on my end.

PR #5432, opened 2023-10-26
Reporting: UI improvements

### LINKED ISSUE #5189: [Reporting] View Improvements
# Still updating, but starting to track

# Filter Dropdown

- [x] Copy suggestion: `On-Chain Wallets` -> `On-Chain` or `Wallets`?
- [x] `Products sold` -> `Products Sold` or simply `Products`

# General

- [x] I noticed there's an additional scrollbar at the bottom of the page, can we remove this? I think the table specific scrollbar makes sense, given the width of it's contents.
Screen Shot 2023-07-17 at 1 34 33 PM

- [x] The main CTA could arguably just be `Export` cleaner, as the CSV bit isn't necessary imo
Screen Shot 2023-07-17 at 2 14 43 PM

- [x] If complete empty states exist, we shouldn't show all the tables
Screen Shot 2023-07-24 at 10 19 06 PM
Screen Shot 2023-07-24 at 10 19 15 PM
Screen Shot 2023-07-24 at 10 19 21 PM

- [x] Would be nice to consolidate into two columns
Screen Shot 2023-07-17 at 1 49 54 PM

- [ ] Adds calendar icons to the Start / End inputs
Screen Shot 2023-07-17 at 3 02 49 PM

# Tables

- [ ] Improved table header states (https://github.com/btcpayserver/btcpayserver-design/issues/17)

# Search

- [ ] Do we want the ability to search? e.g. something like this:
Screen Shot 2023-07-17 at 3 02 49 PM

# Payments
- [x] Could be improved with updated labels:
Screen Shot 2023-07-17 at 2 29 22 PM

- [x] Lightning Address label updates
- Makes me think some of the other Amount labels could be Balance instead...hmm
Screen Shot 2023-07-17 at 3 51 15 PM

# On-Chain
- [ ] Could be improved with updating the label and pushing the header into the table key value for the on-chain view:

- Context: I found the Balance Change aspect of the On-Chain table a little confusing... as the test data amounts were mostly negative values?
Screen Shot 2023-07-17 at 3 00 53 PM

Screen Shot 2023-07-17 at 5 56 21 PM

works in a simple context as well:
Screen Shot 2023-07-17 at 6 27 36 PM

# Products

- [x] `Summary by items` -> `Summary` Could this label just be shortened, and still make sense?
- [ ] Rename `App ID` to `App Name` or `Plugin Name`

# Raw Data
### Payments
- [x] Is the date formatting messed up? `2023-07-05T01:42:41-06:00` 
Screen Shot 2023-07-17 at 3 34 10 PM

### On-Chain
- [x] Suggestions for improvements
- Unconfirmed would be a red `X` icon instead of the green checkmark
- Balance Change decrease would be `-1.06150021` and be colored red
Screen Shot 2023-07-17 at 3 41 50 PM

dennisreimann:
Closes #5189. Marked what could be done in the issue, left the rest out on purpose.

reporting-payments

reporting-products

reporting-wallets


dstrukt:
How do you feel about the various fixed headers? Overall I like the changes, feels a bit tighter, but thinking about how this is all going to translate to the app as well, so evaluating a few things while I review.

dstrukt (changes requested):
Maybe not something we can easily patch, but noticed a few irregularities I thought I should surface:

- [ ] Scroll bar and the spacing, doesn't fill 100% .. should it? I guess that's just the difference of `w-auto` class .. hmmm

Screen Shot 2023-10-26 at 2 36 12 PM

Screen Shot 2023-10-26 at 3 28 43 PM

## Thoughts

- [ ] Does `Products` or `Sales` make more sense? I keep debating this as I think about how we're framing these filters

- [ ] Could see a world in which we enable App/Plugin based filtering for the Export, e.g. I want to export all of my sales from POS A or POS B, etc... but maybe outside of the scope of this PR. I imagine it would just be an added filter directly right of `Products` before the time filters, and the horizontal sizings adjust as needed.

NicolasDorier:
`Aggregated fiat amount` the issue with this terminology is that's its not necessarily true.
If the invoice was in BTC for examlpe, the fiat amount would be bitcoin which isn't fiat.

dennisreimann:
> How do you feel about the various fixed headers

We cannot have it perfect, as long as we combine this with `table-responsive` and have those headers inside scrollable containers. I've experimented with that long enough to say, that there will always be some kind of trade-off and in generasl I'm not a fan of that approach to solve "responsiveness" of data tables. We'll have to find another solution for that at least for the app.

> Overall I like the changes, feels a bit tighter, but thinking about how this is all going to translate to the app as well, so evaluating a few things while I review.

I don't think we'll translate everything directly 1-to-1 from the desktop to the app and vice versa. I'd see different use cases and usage patterns for reporting on both platforms: On desktop you want the details, more finegrained functionality (e.g. tables and sorting) and export, whereas on mobile you'll most likely want more of an overview or dashboard.

dennisreimann:
>  Scroll bar and the spacing, doesn't fill 100% .. should it? I guess that's just the difference of w-auto class

I removed the `w-auto` class for the Raw Data table, which most likely has enough data to not make it look very weird even on big screens :)

> you have Products you're selling, but when you're looking at reports you're looking at Sales

Agreed and updated.

dstrukt:
> We cannot have it perfect, as long as we combine this with `table-responsive` and have those headers inside scrollable containers. I've experimented with that long enough to say, that there will always be some kind of trade-off and in general I'm not a fan of that approach to solve "responsiveness" of data tables. We'll have to find another solution for that at least for the app.

Fair point, and yeah I don't love it either, but for the time being, it works + having custom responsive tables also adds future work if we add columns or make changes, etc... Will proceed with what is best for the app specifically, and maybe we'll find some kind of in-between solution as we learn more in that implementation.

Kukks (approved):

PR #5435, opened 2023-10-27
Do not activate Blazor in Wizard screens

NicolasDorier:
Blazor isn't used in Wizard screens (such as vault screen)

As such, it is better to not have connection to Blazor in them, as they have their own websocket connection.
I noticed browser tends to have very slow websocket connection if we create too many of them.

Same thing, PullPayment and PaymentRequest shouldn't attempt to use Blazor.


dennisreimann (approved):
utACK, my only concern would be the naming of `IsBlazorAllowed(this ViewDataDictionary viewData, bool allowed)`, which imho should rather be `SetBlazorAllowed`.

PR #5438, opened 2023-10-29
POS Cart: Add options for search and categories display

dennisreimann:
Allows to hide the search bar and categories list. Keeps both on/displayed by default.


dstrukt (approved):
tACK - LGTM!

All variations work perfectly, agree with keeping both ON by default as well - nice quality of life update for merchants!

Kukks (commented):
Should these really be settings inside the app? Wouldn't this make more sense to do as UI toggle (some kind of caret collapse?) and persist to local storage, considering these are only for cart view?

dennisreimann:
I think it should be an the backend and be a decision for the merchant. I added these settings, because I missed them for a cart we have at Einundzwanzig. Imho having a search for a cart view with just a few items is a distraction — same goes for categories, which might be useful from a data perspective, but don't make much sense for a shop with just a few items.

Kukks (approved):

PR #5441, opened 2023-10-31
Make wallet object system much more performant

Kukks:
With the current way of how labels are automatically added to wallets, the database would continously get hammered with a connection + insert query for each. This is even more apparent if you use the coinjoin plugin, which adds many, many labels on every round, which sometimes causes btcpay's postgres connection to drop for everything, causing some data loss potentially.

In this PR I optimize by creating a new method that allows bulk insertions of both wallet objects and wallet object links. Bulk insertions are done through a binary copy from btcpay to postgres directly (the most performant way based on many benchmarks online) into a temporary table and then inserted into the main table (this is done so that duplicate records can be handled with `on conflict`, as the copy command cannot handle that).


NicolasDorier:
Temporary table have annoying side effect I'd prefer not to use. (for example, if somebody use pgbouncer, it will stop working properly)

I think I can refactor this to make just a single DB roundtrip. Several roundtrip is what is causing issues.

NicolasDorier:
image

PR #5449, opened 2023-11-03
Responsive editor improvements

dennisreimann:
As discussed with @dstrukt in yesterdays design meeting, this is one more iteration on #5418.

It turns the POS/Crowdfund items into a more condensed list version and moves the editor into an offcanvas for all breakpoints below XL. Apllied this for the form editor as well.

### Mobile version

pos

### Breakpoint L

This would still be too narrow because of the nav sidebar, so on L this pattern is still applied.

pos-l

pos-l-editor

### Breakpooint XL

From XL on the space is sufficient, so we keep it as before.

pos-xl

## Form

form

### DIFF
--- a/BTCPayServer/Views/Shared/Crowdfund/Public/ViewCrowdfund.cshtml
+++ b/BTCPayServer/Views/Shared/Crowdfund/Public/ViewCrowdfund.cshtml
@@ -38,6 +38,10 @@
         #crowdfund-body-description {
             font-size: 16px;
         }
+        .perk.card .card-img-top{
+            max-height: 210px;
+            object-fit: scale-down;
+        }
     </style>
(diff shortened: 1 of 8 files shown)


prusnak:
Love it! ❤️

dstrukt:
As great as we discussed! 

Giving it a proper testing and review this evening, but now that it's implemented - what do you think? Do you like the change?

prusnak:
> Do you like the change?

Absolutely, it's definitely a change for the better.

dstrukt (approved):
Overall, this is fantastic, very smooth, and agree with @prusnak, it's definitely a change for the better.

Was debating if we should kill the `xxl-constraint` on the section, but the more I test, the more I think it makes sense to keep.

Excellent work!

PR #5450, opened 2023-11-03
Send notification when a new plugin version is available

Kukks:


dennisreimann:
Good idea and concept ACK, but why `GitHubVersionFetcher`? Shouldn't it get the info from our plugin builder API?

Kukks:
> Good idea and concept ACK, but why `GitHubVersionFetcher`? Shouldn't it get the info from our plugin builder API?

I only refactored the btcpay version checker, the plugin version checker uses the plugin builder api

dstrukt:
conceptual ACK! love the idea - looking forward to my rockstar hairstylist plugin notification

PR #5456, opened 2023-11-09
Dynamic Reports

Kukks:
A bit advanced so may put it in a plugin instead. This allows a server admin to create custom reports, which are saved and appear as normal resports in the Reporting features. These reports are only available to Server admins by default but can be marked viewable by non-admins too. This feature only works on Postgres deployments. Security-wise, all queries are run in a transaction and rolled back every time to prevent queries that modify or corrupt. Filters on the rpoerting page can be applied within the sql as parametrs (such as `@StoreId`, `@From`, and `@To`)

### DIFF
--- a/BTCPayServer/Controllers/UIReportsController.cs
+++ b/BTCPayServer/Controllers/UIReportsController.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using System.Linq;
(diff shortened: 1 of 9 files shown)


Kukks:
Made it a plugin instead: Dynamic Reports

PR #5460, opened 2023-11-10
Minor cleanups

dennisreimann:
More or less random stuff I stumbled across. The `sr-only` class got replaced in Bootstrap 5, we still have it in the CSS because `font-awesome.css` also contains it.

Replacing Font Awesome is something we can also tackle when updating our own icons, fyi @dstrukt.


dstrukt (approved):
LGTM!

Kukks (approved):

PR #5462, opened 2023-11-10
Unify public page styles

dennisreimann:
Based on https://github.com/btcpayserver/btcpayserver/pull/5413 and needs it to get merged first.

- Uses `--wrap-max-width` on `.public-page-wrap` rather than inner `.container` classes
- Applies `.tile` class to boxes and makes them connect to the edge of the screen below `400px` width

unify


dennisreimann:
Rebased and ready for review, @dstrukt.

dstrukt (approved):
Still trying to find places where we might have missed, or could be used, but in all of the relevant places, it looks great!

PR #5463, opened 2023-11-10
Support adjusting form invoice amount by multiplier

Kukks:
closes #5425


dennisreimann (commented):
concept ACK, but it could use a test :)

dennisreimann (approved):
tACK. Nice one. Example code for a form with three promo codes:

- `huge` -> 50% discount
- `medium` -> 10% discount
- `tiny` -> 1% discount

```json
{
  "fields": [
    {
      "name": "promo",
      "type": "text"
      "label": "Promo Code"
    },
    {
      "name": "invoice_amount_multiply_adjustment_promo",
      "type": "mirror",
      "value": "promo",
      "label": "Promo Codes",
      "valuemap": {
        "tiny": "0.99",
        "medium": "0.90",
        "huge": "0.5"
      }
    }
  ]
}
```

Kukks:
NOTE: @pavlenex  we need to adjust docs

ascisco:
Just tested this feature and instead of adjusting the invoice amount by multiplied number it replaced the invoice amount with adjusted amount.

For example, I set a promo code for 0.1 (10% off).  During checkout for a $10 order, the invoice I am given has a final price of $1 instead of $9. The error is reflected in my BTCPay order history as well.

Kukks:
It's a multiplier. If the current price is 10 and you multiply by 0.1, then
yes it is correctly calculated as 1.

PR #5471, opened 2023-11-14
Domain mapping: Redirect root app to canonical URL

dennisreimann:
We already redirect public app URLs to the canonical URL if there's a domain mapping — this adds the same behaviour for apps that are defined as root app as well.


NicolasDorier:
I believe you need to adapt https://github.com/btcpayserver/btcpayserver/pull/5471/files#diff-4d62f7096a48ee972a859684203bbbde66e92dcdf4ddc42f2e9f641114948924L49-L65

If the user navigate through the app through the root URL, you need to set the `appId` in the custom route I believe.

PR #5473, opened 2023-11-14
Theme Switch: Refactor and add system option

dennisreimann:
Before, we had no way to reset the theme option to the system default. This introduces that option and refactors the theme switch to work in a simpler manner.

## Light

light

## Dark

dark

## System

system


dstrukt:
I thought about this as well when working on the Payment Request PR, man, like minds!

Thoughts FWIW. I do wonder if this option makes more sense in the Branding section of the Store Settings, as it mostly affects the public views and the overall Store theme, less so a viewing mode like the Light/Dark. I saw it more of a `Reset to default` type of button. Thoughts?

dennisreimann:
> I do wonder if this option makes more sense in the Branding section of the Store Settings, as it mostly affects the public views and specific parts of the overall Store theme

Nope, it's a user preference, which affects just the signed in user. This isn't tied to the store or server themes, it simply chooses which one should be applied for the current user.

dstrukt (approved):
Had to properly test system level display changes, as I originally misunderstood the goal of this PR (clarified with @dennisreimann on the design call) and appears to work well for macOS (10.15.7), but that's the only environment I can test at the moment.

PR #5478, opened 2023-11-15
Keypad: List recent transactions

### LINKED ISSUE #5379: Add local transaction history to keypad-only POS
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/5349

<sup>Originally posted by **vindard** September 25, 2023</sup>
In a nutshell, it's sometimes difficult for a cashier to check if a previous invoice actually went through in instances where they might be doing lots of transaction quickly or there is some network blip or accidental page refresh.

Since POS apps already have unique order ids for each invoice, and there's a dashboard button for seeing that POS's invoices only, it may not be a heavy lift to expose this in some way to the POS end-user as well.

An example could be like how Ibex does it with a little "tx history" button in the top-right corner ([try live demo here](https://ibexpay.ibexmercado.com/bpt/6398c017-f8e3-40aa-9c22-54224ece366b))

### Example screenshot from IbexPay

image
</div>

dennisreimann:
Closes #5379. 

## List

I played with various forms of data and display and ended up with the list group instead of a table and omitting the invoice ID, because the space is rather limited. In its current form it degrades nicely on even smaller displays, wrapping the status to a second line and aligning it on the right (see screenshot).

The whole row is clickable and will bring the user to the invoice details — we can then incorparate the refund process from there or link to the receipt and come up with a new and better approach of handling the refund flow in a separate PR.

trans

grafik

## Empty state

trans-none


dstrukt:
FUGGG yeah!

Will review this evening, was gonna say maybe the table header was necessary, but suppose because there are so few items and their contents are clear enough, we don't need it.

dstrukt (changes requested):
Overall, looks incredible! This is a great addition!

Still reviewing a bit, but few thoughts:

- [ ] Don't need the additional `Close` button on the footer of the modal imo

- [ ] Think we should push the `Refresh` button as an icon on the header OR as a row line item either above or below the transactions in the modal body (could stay text)

- [ ] I need to get you a proper transactions icon, was a placeholder, which works great, but it reads more like a menu icon (on me)

Minor tweak consideration:

- [ ] The transaction icon could be 40x40, slightly larger than the current 32x32 .. thoughts?

- [ ] Can we center align horizontally the icon based on if a store logo or POS title is present, e.g: the transaction icon sits at the top of both, 

Screen Shot 2023-11-16 at 9 43 46 AM
Screen Shot 2023-11-16 at 9 44 23 AM

Mguadalupe08 (approved):

dennisreimann:
Updated the list according to what we discussed, @dstrukt 

Bildschirmfoto am 2023-11-20 um 11 07 59

## No border version

Also added a no border version for discussion.

Bildschirmfoto am 2023-11-20 um 11 12 54

NicolasDorier:
Came for the sovereign open source payment processor. Stayed for the Keypad.

dstrukt:
Getting you a final review - upgrading to .NET8

dstrukt (approved):
tACK!

Might want to adjust some of our modal UI a bit in the future, and we can tackle the icons in one swoop once I wrap them up, but the two we currently do work well!

Have a slight preference for the borderless numbered buttons, but could go either way, thoughts?

pavlenex:
> Have a slight preference for the borderless numbered buttons

Same, but not a blocker for me either.

pavlenex (approved):
tACK

rawgir-7Xoqca-nezgej (commented):
imageimage

rawgir-7Xoqca-nezgej (commented):
imageimage

PR #5490, opened 2023-11-20
Use Mempoolspace fees

Kukks:
Since bitcoind's fee estiomates are horrible, I would use an altenrative, but that adds a third party to the mix. We can either:
* Accept the risk (it is only for fee estimation anyway)
* Offer a toggle in the server settings
* Move this code to a plugin


ndeet:
Yes lots of people complaining that the core fee estimations are off a lot when they change, either too low or too high. 

If you are already on it, what do you think of not displaying the fees as a decimals? Having a suggestion of 8,203 sat is imo very confusing, why not just show 8 sat?

dstrukt:
Tend to agree with @ndeet wrt rounding the decimal, for the sake of simplicity.

I like the toggle or plugin approach, gives optionality and a fallback in case API call failures, dependency risk, etc..

dennisreimann:
Let's not make it a plugin but keep it in core as this fulfills an essential function and we are seeing this right now.

NicolasDorier:
I think it makes sense

Kukks:
> I think it makes sense. What happen when estimations aren't available for example? Maybe we should fallback on bitcoin core?

Done. There is now a mempool-> nbx -> static fallback flow

dennisreimann (approved):
LGTM. 👍

NicolasDorier:
Refactored a bunch of stuff:

* Use C# 12 features with primary ctor to make the code more conscise
* Removed `BaseFeeProvider` class, doesn't make sense for all fee provider to have a fallback
* Fix a bug, where if `economyFee==minimumFee`, then `recommendedFees.ToDictionary` would throw an exception
* Add a test for mempool fee provider
* Removed semaphore access to protect the cache, it's not needed
* Introduced `FallbackFeeProvider` which accept ordered list of fee providers to try
* Do not use mempool space fee rate on regtest, it makes tests unreliable and dependent on external service

harding:
A couple comments:

- I did a quick git grep of BTCPay Server's code and I don't see a maximum fee or fee rate in the code.  (But I could have easily missed it.)  If one doesn't exist, I would suggest adding an "absurd" fee check to avoid the case when a bug or exploit of the API provider accidentally leads BTCPay users to massively overpay fees.

- When using feerates from a third-party API provider, they may be able to associate a transaction with a feerate of _x_ with the IP address which they told to use a feerate of _x_, especially if the request for a feerate occurs only a short time before the broadcast of the transaction using that feerate.  Besides requesting feerates using an ephemeral Tor connection, I don't know of a completely satisfactory solution to this problem, but periodically requesting feerates (e.g. once per block) and adding some noise into the feerate (e.g. + or - a random number from 0% to 10%) might help break the link.  You could also compare the API feerate to the Bitcoin Core feerate and use the Bitcoin Core feerate if the difference is less than 10%.

pavlenex:
Interesting suggestion, thoughts @NicolasDorier @Kukks @dennisreimann?

dennisreimann:
Agreed, two good and valid points, let's check and apply them.

pavlenex:
> Agreed, two good and valid points, let's check and apply them.

Should we log new issue(s) for these?

PR #5502, opened 2023-11-22
enhance fine grain permissions

### LINKED ISSUE #5344: [Bug]: Minimal permissions not working as intended

BTCPay 1.11.5-rc1 but also true for previous verison 1.11.4 and likely earlier

docker

### What happened?

**1.) no "view invoices" permission causes no store access at all**
if that permission is not set then the store won't be accessible at all, also if you have the "manage pull payments" permission

**2.) Sub-permissions won't work:**
E.g. if you have a serverwide role of "pull payment manager" that only has the permission of "create non-approved pull payments" and "archive pull payments" then the "Pull payments" menu entry won't show up at all. But it should?

**3.) Reporting menu item leads to 403**
When you have the "view invoices" permission the "Reporting" menu items shows up but leads to a 403 permission denied page

see above description

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

Kukks:
fixes #5344


BTCPayServer/Views/UIStorePullPayments/PullPayments.cshtml
```diff
@@ -128,12 +128,15 @@
...
-                           asp-route-pullPaymentId="@pp.Id">
+                        <a 
+                            permission="@Policies.CanManagePullPayments"
+                            asp-action="EditPullPayment"
+                            asp-controller="UIPullPayment"
+                            asp-route-storeId="@storeId"
+                            asp-route-pullPaymentId="@pp.Id">
                             @pp.Name
                         </a>
+                        <spam not-permission="@Policies.CanManagePullPayments"> @pp.Name</spam>
```
dennisreimann, line 139:
```suggestion
                        <span not-permission="@Policies.CanManagePullPayments">@pp.Name</span>
```

BTCPayServer/Views/UIStorePullPayments/PullPayments.cshtml
```diff
@@ -149,12 +152,14 @@
                     <td class="actions-col">
                         <div class="d-inline-flex align-items-center gap-3">
                             <a asp-action="ViewPullPayment"
+                                permission="@Policies.CanViewPullPayments"
```
dennisreimann, line 155:
```suggestion
                               permission="@Policies.CanViewPullPayments"
```

PR #5504, opened 2023-11-23
Fix qemu package in ARM Docker files

greimela:
With the new debian bookworm, the `qemu` package has been split into one package per architecture.


arm32v7.Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM mcr.microsoft.com/dotnet/sdk:8.0.100-bookworm-slim AS builder
 ENV DOTNET_CLI_TELEMETRY_OPTOUT=1
 RUN apt-get update \
-	&& apt-get install -qq --no-install-recommends qemu qemu-user-static qemu-user binfmt-support
```
NicolasDorier, line 5: `qemu-system-arm` is for 64 bit only it seems.

NicolasDorier: it works!

PR #5509, opened 2023-11-24
NFC improvements

### LINKED ISSUE #5507: [Enchancement]: Continue scanning for NFC after payment to prevent accidental redirects out of the checkout
When paying at checkout with NFC card, customers need to remove it before payment is processed. If they do not remove the card fast enough they will be redirected out of checkout page, Android's default NFC scanning functionality will take over.

Video illustration:

https://drive.google.com/file/d/1cpOVV9H54npy4BjK4sZP49ZL0jNinM7M/view?usp=sharing

### LINKED ISSUE #5508: [Enchancement]: More prominent visual feedback on NFC operations
Our NFC status messages are currently located at the bottom of the checkout page - probably the fact they are in `CheckoutEnd.cshtml` played part in this. This placement, however, leads to a confusing UX on smaller screens, as these messages often remain unseen. Here's a video demonstrating the issue:

https://github.com/btcpayserver/btcpayserver/assets/5191402/592e37f5-fa16-4d51-88c0-03df31f595f4

In this particular example, there is also an issue with NFC detection on the device, but - you get the problem.

When a user pays using an NFC card, it's crucial to have a status message displayed prominently (e.g., "Payment Submitting...", "Payment Failed", "Payment Succeeded", etc.). 

As an initial step, moving the message label to the top of the checkout page, above the QR code, would improve visibility. In the long term, implementing some form of overlay to show the status of NFC card payments would be ideal.

dennisreimann:
Moves most of the NFC code out of the LNURL component and on Vue app level, so that 

- it keeps the scanning active even on the result page, so that the native scanning functionality does not take over in case the card is held too long on the scanner. Closes #5507
- it can provide more prominent visual feedback on NFC operations. Closes #5508

Downside: I had to duplicate that code in V1 and V2, as the shared partial/component isn't used anymore. However, this can also be seen as a good thing moving forward, as it got rid of some plase where we were using a check f the componnet is used in V2 context.

v1
 
GitHub does not allow me to upload the video I recorded for v2, but @rockstardev can test and see if it works the way he envisioned it.

### DIFF
--- a/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
@@ -32,7 +32,7 @@
 
 <script>
     Vue.component('BitcoinLikeMethodCheckout', {
-        props: ["model"],
+        props: ['model', 'nfcSupported', 'nfcScanning'],
         template: "#bitcoin-method-checkout-template",
         components: {
             qrcode: VueQrcode
--- a/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout-v2.cshtml
@@ -32,7 +32,7 @@
 
 <script>
     Vue.component('BitcoinLikeMethodCheckout', {
-        props: ['model', 'nfcSupported', 'nfcScanning'],
+        props: ['model', 'nfcSupported', 'nfcScanning', 'nfcErrorMessage'],
         template: "#bitcoin-method-checkout-template",
         components: {
             qrcode: VueQrcode
--- a/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout.cshtml
+++ b/BTCPayServer/Views/Shared/Bitcoin/BitcoinLikeMethodCheckout.cshtml
@@ -79,7 +79,7 @@
 <script type="text/javascript">
     Vue.component('BitcoinLikeMethodCheckout',
         {
-            props: ['srvModel', 'nfcSupported', 'nfcScanning'],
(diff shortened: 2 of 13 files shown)


Kukks:
If we're going into this direction, please take a look at https://github.com/Kukks/btcpayserver/tree/ultra-nfc

rockstardev:
Great job all around @dennisreimann! Here are my observations, illustrated by a video:

[Video Link](https://github.com/btcpayserver/btcpayserver/assets/5191402/5688e7ea-dc08-4893-978b-cff29b493edb)

- The issue of navigating out of checkout when holding the NFC card over the device after payment completion is resolved and works great (#5507).
- The placement of the toast notification is good, but since it doesn't expire, it eventually obstructs the rest of the UI.
- On the first device with a smaller screen, as you can see in the video, I'm unable to turn off NFC, leaving me stuck with the message. Unfortunate.
- Even after successful payment, the "Paid out" message remains on the dialog, which detracts from the overall look.

To streamline this and move forward with the merge, my suggestion is to relocate the toast-container into CheckoutEnd.cstml. We can then display errors and successes within that area for Lightning. Also we use a regular div instead of a fixed or floating toast, which will shift the Scanning NFC button downwards. This should work well with all the large-screen devices we're currently using, ensuring visibility of any errors.

Post-AfroBitcoin, when we have more time, we should regroup to brainstorm and refine the UI/UX for this feature.

rockstardev (changes requested):
Just changing error/success div from toast into regular div that shift Scanning NFC button down and I would merge this.

rockstardev:
Also, a question for @dennisreimann and @Kukks – when debugging locally on regtest, how do you handle it? 

When I connect to the internal lightning node, I only see the option to pay with LNURL, not Lightning. Additionally, I'm lacking NFC scanning functionality (I connect with browser on the POS terminal over an IP like 192.168.1.50).

dstrukt (approved):
Just wanted to poke around - everything looks great - agree with @rockstardev, would be nice to revisit in the future to refine a few things!

PR #5513, opened 2023-11-27
Dev env: Fix Lightning config warning

dennisreimann:
Gets rid of this warning on startup:

```
warn: Configuration:  Setting BTC.lightning is a deprecated format (type=clightning;server=tcp://127.0.0.1:30993), it will work now, but please replace it for future versions with 'type=clightning;server=tcp://127.0.0.1:30993/'
```


Kukks (changes requested):
Might need to fix this on the ln lib instead

NicolasDorier:
I don't think this is a bug of LNLib, this is just because of how we check that the connection string is "Legacy" in the btcpay.

PR #5519, opened 2023-11-28
Store Branding: Refactoring and logo as favicon

dennisreimann:
- Encapsulates store branding properties into their own view model
- Uses the logo as favicon on public pages


NicolasDorier:
@dennisreimann does this PR will also use the logo when you add the app to the home screen of your phone? This would be very useful.

pavlenex:
Oh nice idea @dennisreimann!

dennisreimann:
@NicolasDorier Now it does, thanks for the pointer!

NicolasDorier (approved):

dstrukt (approved):
tACK - LGTM as far as I can tell (Payment Requests / Pull Payments / POS & CF), but will submit a bug if I find anything else.

dstrukt:
I was having issues making a new store though, so used existing stores I had, but might not be related to this PR...?

Kukks (approved):
This may break plugins that have used the StoreHeader partial?

PR #5537, opened 2023-12-05
Allow scheduling installs/updates of future plugins

Kukks:
Sometimes we have to release a BTCPay version that will break all current plugins, and the ony way to fix is to release a new plugin version that nly allows btcpay to be the new version onwards. This means one cannot update plugins to a future version to avoid plugins from breaking. This PR introduces mechanics so that plugin installs and updates can be queued even if the dependencies are not met, so that once they are, they will be installed on startup. If you queue a plugin, and a new update is available, you will be able to replace that with the new version

### DIFF
--- a/BTCPayServer/Plugins/PluginManager.cs
+++ b/BTCPayServer/Plugins/PluginManager.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Collections.Specialized;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
@@ -10,7 +9,6 @@ using System.Reflection;
 using System.Text.RegularExpressions;
 using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Configuration;
-using BTCPayServer.Logging;
 using McMaster.NETCore.Plugins;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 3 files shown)


AdilElFarissi:
I was sure you would do this after the plugins notifications... good job and thanks :)

dennisreimann (approved):
tACK. @kukks please also review my update, especially eff5b38c09792a82f5136e9ce4dd7ada9de28a10. I had to do these changes because during startup `ExecuteCommands` is initially invoked with the `installed` argument being `null` ([refereence](https://github.com/btcpayserver/btcpayserver/pull/5537/files#diff-3681d8b09897aa17cda6a030b4f31bcda6310d64b20132aee4ccb2f8048c1ea1R102)), which lead to crashes. 

I also moved the `ToLowerInvariant` conversion to the relevant places where comparisons happen, so that we ensure it's always lowercase when comparing, which wasn't the case for the comparison on startup ([reference](https://github.com/btcpayserver/btcpayserver/pull/5537/files#diff-3681d8b09897aa17cda6a030b4f31bcda6310d64b20132aee4ccb2f8048c1ea1R108)).

Screenshot from the UI with FixedFloat and LNbank having versions which will be available with v1.12. 

plugins

Once scheduled, the commands list keeps them in case the dependencies aren't met.

However, one more thing we could add is the ability to also select the version to be installed, so thnat one could pick an updatable version compatible with the current BTCPay Server version. Right now we show inly the latest plugin version to update to.

dennisreimann:
@Kukks I think we can merge this an progress with the "pick an update to install" part in a separate PR once btcpayserver/btcpayserver-plugin-builder#9 is merged.

Issue #5191, opened 2023-07-25
[Dashboard] Priority ordering on the dashboard widgets


dstrukt:
While testing the CoinJoin plugin, I realized we may want to implement a priority ordering, if possible, to prevent certain widgets from displacing the global widgets by default, we can address specific reordering as an override at a later date though.

Ideally the wallet + transactions + lightning widgets sit first always, unless explicitly overridden. Not 100% sure this is the right place to file this issue, or on the plugin repo itself.

<img width="1429" alt="Screen Shot 2023-07-25 at 11 39 26 AM" src="https://github.com/btcpayserver/btcpayserver/assets/6250771/9ac60c91-1301-4449-a22c-bbd6e74ed737">

pavlenex:
Related https://github.com/btcpayserver/btcpayserver/issues/3822

Kukks:
Related #5742

PR #5543, opened 2023-12-10
dashboard widget priority ordering

TChukwuleta:


Kukks (commented):
What's the reasoning for this change? Did anyone request it?

TChukwuleta:
> What's the reasoning for this change? Did anyone request it?

It was mentioned here #5191 that  the wallet balance, transactions, lightning balance widgets should always sit first, hence the reason for the re-ordering and the changes

dennisreimann:
Ideally we make it so, that the users can reorder and prioritize themselves. Otherwise tgis will always be subjective and limited.

Kukks:
Let's not reorder this until we actually have widget customization support.
I want users to notice plugin specific features on the dashboard faster

Kukks:
closing in favor of #5742

Issue #5549, opened 2023-12-11
[BUG] Low-priority - Server admin gets 404 on a payment request generated by the user of another store


pavlenex:
A low priority bug, but should be addressed.

To replicate:
- Registed as a regular user to a server (non admin)
- Generate payment request
- Share that payment request with server admin or log in into that same server as admin and try to open a payment request you'll get 404

PR #5551, opened 2023-12-12
Allow users with CanViewPaymentRequests to view payment requests

dennisreimann:
I originally tried to reproduce #5549, which I couldn't, but came across this problem where a store user with role `Guest` wasn't able to access the list of payment requests, even though that item appears in the nav list. The `CanViewPaymentRequests` permission on th list action is correct, but access was prevented by the `CanModifyStoreSettings` policy requirement on the controller level, which is adjusted here.

### DIFF
--- a/BTCPayServer/Controllers/UIPaymentRequestController.cs
+++ b/BTCPayServer/Controllers/UIPaymentRequestController.cs
@@ -26,8 +26,8 @@ using StoreData = BTCPayServer.Data.StoreData;
 
 namespace BTCPayServer.Controllers
 {
-    [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
     [Route("payment-requests")]
+    [Authorize(Policy = Policies.CanViewPaymentRequests, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
     public class UIPaymentRequestController : Controller
     {
         private readonly UIInvoiceController _InvoiceController;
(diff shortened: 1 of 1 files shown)


BTCPayServer/Controllers/UIPaymentRequestController.cs
```diff
@@ -363,6 +362,7 @@ public async Task<IActionResult> ViewPaymentRequestForm(string payReqId, FormVie
         }
 
         [HttpGet("{payReqId}/cancel")]
+        [Authorize(Policy = Policies.CanModifyPaymentRequests, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
```
Kukks, line 365: No, AFAIK this should be anonymouse and users can cancel these?

Kukks (changes requested):

BTCPayServer/Controllers/UIPaymentRequestController.cs
dennisreimann: I think you are right and this is the user facing cancellation, so it'd need to `[AllowAnonymous]`. This was broken before then, because it inherited the policy from the controller level, which I simply moved here. Will check and correct it.

dennisreimann: Resolved in [1f11dae](https://github.com/btcpayserver/btcpayserver/pull/5551/commits/1f11dae33aa5ccd4f8580f06aaa9719652f0f763).

Kukks (approved):

PR #5553, opened 2023-12-12
Remove bittrex

Kukks:
bittrex is dead https://www.coindesk.com/business/2023/11/21/crypto-exchange-bittrex-global-to-shut-down/#:~:text=Crypto%20exchange%20Bittrex%20Global%20is,only%20withdrawals%20will%20be%20available.


dennisreimann (approved):
tACK

PR #5555, opened 2023-12-13
POS: Remove forced center alignment for description

dennisreimann:
Allows to specify the text alignment in the description container via the richt text editor. Before it was center aligned, no matter what one did in the editor.

This is feedback we got in yesterdays call with @MattDHill.


dstrukt (approved):
Never noticed this, nice catch @MattDHill!

tACK

PR #5556, opened 2023-12-13
Random feerate and ensure sanity

Kukks:
Suggested at https://github.com/btcpayserver/btcpayserver/pull/5490#issuecomment-1851066223 We can also configure this httpclient to use tor


dennisreimann (approved):
tACK. Randomizing by 3-5 percent might be enough though.

PR #5558, opened 2023-12-14
Greenfield: Fix invoice refund permission

Kukks:


dennisreimann (approved):
utACK. Verified this matches what we have in the UI.

ndeet (approved):
Can confirm the non-approved pull payments permission works now to create refunds based on invoice refund endpoint over the API. Thank you :partying_face:

PR #5567, opened 2023-12-19
Lightning: Allow LND to be used with non-admin macaroons

dennisreimann:
Requires btcpayserver/BTCPayServer.Lightning#152. Using `invoice.macaroon` suffices, only downside is we cannot display the connection details on the public Lightning node info page.

### DIFF
--- a/BTCPayServer/BTCPayServer.csproj
+++ b/BTCPayServer/BTCPayServer.csproj
@@ -50,7 +50,7 @@
     <PackageReference Include="YamlDotNet" Version="8.0.0" />
     <PackageReference Include="BIP78.Sender" Version="0.2.2" />
     <PackageReference Include="BTCPayServer.Hwi" Version="2.0.2" />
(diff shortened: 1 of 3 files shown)


pavlenex:
What's the purpose of doing this?

dennisreimann:
Right now we require the admin.macaroon and we can get by with less permissive macaroons, like invoice.macaroon. basically this is a security measure to e.g. not expose withdraw access for the Lightning node.

dennisreimann:
@pavlenex this was requested by @jlopp on Nostr as response to the LNbank vulnerability recap: https://njump.me/nevent1qqsvv537t96xlsqshkamdalqlss6vlrz5f77zkpzqjw5wvc87sez82spremhxue69uhkummnw3ez6ur4vgh8wetvd3hhyer9wghxuet59upzpaegm8nwwpyrtrnsjv84efjtp9mhpkvfenvxs487vx8d48y28qgxqvzqqqqqqyslv99p

PR #5577, opened 2023-12-21
Various plugin fixes

Kukks:
This will:
* fix plugin updates only causing an uninstall
* Clear any previous pending actions of a plugin if you click uninstall
* Show the plugin version that was disabled
* Show an update button on disabled plugins instead of install
* if a plugin is scheduled to be installed/updated, it will show which version was scheduled to be updated. If a newer version if available than the scheduled one, it will show an option to switch to that


Kukks:
Note: While we should get this out in the next release, I need to redo how we build this plugin page as it's been constantly added on as complex version features were added to plugins.  cc @pavlenex

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -108,6 +108,7 @@ public async Task UploadPlugin(IFormFile plugin)
         public void UninstallPlugin(string plugin)
         {
             var dest = _dataDirectories.Value.PluginDir;
+            PluginManager.CancelCommands(dest, plugin);
```
Kukks, line 111: This is actually a bugfix where you could end up in a situation of scheduling installs/updates and then scheduling a delete, causing the UI to look unintuitive on what's going to happen until they execute

PR #5579, opened 2023-12-21
Lightning Address: Use lowercased username when resolving

dennisreimann:
This wasn't handled consistently and casing shouldn't matter here.


BTCPayServer/Plugins/LightningAddressResolver.cs
```diff
@@ -0,0 +1,9 @@
+using LNURL;
+
+namespace BTCPayServer.Plugins;
+
+public class LightningAddressResolver(string username)
+{
+    public string Username { get; set; } = username?.ToLowerInvariant();
```
Kukks, line 7: Make `NormalizeUsername` public static and use here

Kukks (changes requested):

PR #5593, opened 2023-12-22
Startup: List configured networks in non-altcoin build warning

dennisreimann:
In my standard Docker build I see these log entries: `Configuration:  This build of BTCPay Server does not support altcoins`. I'd like to be able to better debug this, because as far as I see there are no other networks configured:

```
BTCPAY_CRYPTOS=btc
BTCPAYGEN_CRYPTO1=btc
```

The rest of the BTCPAYGEN_CRYPTO0X is empty. The log also yields `Configuration:  Supported chains: BTC` at the same time.


NicolasDorier:
Those environment variable are only used by the `btcpayserver-docker`.

My guess is that

```
BTCPAY_CRYPTOS=btc
BTCPAYGEN_CRYPTO1=btc
```

Is counted as two crypto "btc".

PR #5618, opened 2023-12-26
Removed what's new button and info #5608

### LINKED ISSUE #5608: Remove what's new button and info
image

This one hasn't received love in the past couple of releases so consensus is to remove it and clean up dashboard.

hiluan:
This update resolves #5608 by removing the 'What's New' button and the information that comes with it.


dstrukt (approved):
Functional tACK.

There might be some tests or additional model code that needs to be removed in addition to what you've done, but I'll leave that to @dennisreimann.

@pavlenex and I also discussed potentially removing the `Title`, but decided to keep it, in the likely case that we'll have a `Configure` button (that will allow the user to re-order/hide specific Dashboard cards) which would replace the What's New button.

dennisreimann (approved):
Thanks @hiluan!

PR #5619, opened 2023-12-26
Make tips and discount properties disabled in POS setting

### LINKED ISSUE #5600: Disable `tips` and `discounts` by default in POS
For some reason, by default we add discount and tips features, which in most cases should be left to the user to enable if needed, and we should show a straightforward barebone keypad pos by default.

TChukwuleta:
Closes #5600. 

This PR sets tips and discound properties in POS settings by default

### DIFF
--- a/BTCPayServer.Client/Models/CreateAppRequest.cs
+++ b/BTCPayServer.Client/Models/CreateAppRequest.cs
@@ -27,10 +27,10 @@ namespace BTCPayServer.Client.Models
         [JsonConverter(typeof(StringEnumConverter))]
         public PosViewType DefaultView { get; set; }
         public bool ShowCustomAmount { get; set; } = false;
-        public bool ShowDiscount { get; set; } = true;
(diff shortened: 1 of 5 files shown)


dennisreimann (changes requested):
As there are API methods for the POS as well, this also needs to get adapted in the [`CreatePointOfSaleAppRequest`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Client/Models/CreateAppRequest.cs#L30). While we are at it, can you also add the new `"default": "false"` to [its API docs](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json#L468)?

dennisreimann:
The POS tests now need adaptation too, because in there we test both the discount and the tips feature. We should keep those tests, but add a new one first which checks that those features aren't present, then enable them and continue with the rest of the tests.

dennisreimann (approved):
tACK

PR #5620, opened 2023-12-26
Prevent payment request to be created when a wallet is not set up

### LINKED ISSUE #5602: Don't allow payment request to be created when a wallet is not set up
Context:
Currently, when wallet is not configured we don't allow the ability to generate the invoice. 
We should do the same for payment request, because otherwise, user can generate a payment link, send it to a client, client clicks pay and gets an error that wallet is not configured. 

Screenshot 2023-12-22 at 16 58 25
Screenshot 2023-12-22 at 17 01 12
Screenshot 2023-12-22 at 17 01 21

It's a bit more productive if we use same protection mechanism as we do for invoices?

Screenshot 2023-12-22 at 16 58 09

TChukwuleta:
Closes #5602. 

This prevents the user from accessing the payment requests creation form if a wallet is not set up

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -1199,7 +1199,6 @@ namespace BTCPayServer.Tests
             s.Driver.FindElement(By.Id("StoreNav-PaymentRequests")).Click();
             s.Driver.FindElement(By.Id("CreatePaymentRequest")).Click();
             // Should give us an error message if we try to create an invoice before adding a wallet
-            var rr = s.Driver.PageSource;
             Assert.Contains("To create an invoice, you need to", s.Driver.PageSource);
             s.AddDerivationScheme();
             s.Driver.FindElement(By.Id("StoreNav-PaymentRequests")).Click();
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -1195,8 +1195,13 @@ namespace BTCPayServer.Tests
             await s.StartAsync();
             s.RegisterNewUser();
             s.CreateNewStore();
-            s.AddDerivationScheme();
 
+            s.Driver.FindElement(By.Id("StoreNav-PaymentRequests")).Click();
+            s.Driver.FindElement(By.Id("CreatePaymentRequest")).Click();
+            // Should give us an error message if we try to create an invoice before adding a wallet
+            var rr = s.Driver.PageSource;
+            Assert.Contains("To create an invoice, you need to", s.Driver.PageSource);
+            s.AddDerivationScheme();
             s.Driver.FindElement(By.Id("StoreNav-PaymentRequests")).Click();
             s.Driver.FindElement(By.Id("CreatePaymentRequest")).Click();
             s.Driver.FindElement(By.Id("Title")).SendKeys("Pay123");
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -565,9 +565,11 @@ namespace BTCPayServer.Tests
             s.RegisterNewUser(true);
             s.CreateNewStore();
             s.GoToInvoices();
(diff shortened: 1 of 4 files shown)


dstrukt:
Thank you for all of the latest work @TChukwuleta tagging @dennisreimann and I for reviews - apologies that we're a bit slower to respond around the holidays!

BTCPayServer/Controllers/UIPaymentRequestController.cs
```diff
@@ -102,17 +107,40 @@ public async Task<IActionResult> GetPaymentRequests(string storeId, ListPaymentR
             return View(model);
         }
 
+        private bool AnyPaymentMethodAvailable(StoreData store)
+        {
+            var storeBlob = store.GetStoreBlob();
+            var excludeFilter = storeBlob.GetExcludedPaymentMethods();
+
+            return store.GetSupportedPaymentMethods(_networkProvider).Where(s => !excludeFilter.Match(s.PaymentId)).Any();
+        }
```
dennisreimann, line 116: Can you make this an [extension method on the `StoreData`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Data/StoreDataExtensions.cs)? You can then refactor it so, that it is also [reused in the `InvoiceController`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UIInvoiceController.UI.cs#L1164) and we avoid duplication.

Otherwise it looks good to me! 👍

TChukwuleta: Yea, I was thinking about that, but wasn't so sure on where to place it.

Would do this

TChukwuleta:
@dennisreimann  it has been resolved.

dennisreimann:
It looks like the invoice test for that case fails now. Can you fix it and also add a similar one for the payment requests?

BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -1195,8 +1195,13 @@ public async Task CanCreatePayRequest()
             await s.StartAsync();
             s.RegisterNewUser();
             s.CreateNewStore();
-            s.AddDerivationScheme();
 
+            s.Driver.FindElement(By.Id("StoreNav-PaymentRequests")).Click();
+            s.Driver.FindElement(By.Id("CreatePaymentRequest")).Click();
+            // Should give us an error message if we try to create an invoice before adding a wallet
+            var rr = s.Driver.PageSource;
```
dennisreimann, line 1202:
```suggestion

            // Should give us an error message if we try to create an invoice before adding a wallet
```

BTCPayServer/Controllers/UIPaymentRequestController.cs
```diff
@@ -107,12 +112,27 @@ public async Task<IActionResult> GetPaymentRequests(string storeId, ListPaymentR
...
+            {
+                TempData.SetStatusMessageModel(new StatusMessageModel
+                {
+                    Severity = StatusMessageModel.StatusSeverity.Error,
+                    Html = $"To create an invoice, you need to <a href='{Url.Action(nameof(UIStoresController.SetupWallet), "UIStores", new { cryptoCode = _networkProvider.DefaultNetwork.CryptoCode, storeId = store.Id })}' class='alert-link'>set up a wallet</a> first",
+                    AllowDismiss = false
+                });
+                HttpContext.SetStoreData(store);
+                return RedirectToAction(nameof(GetPaymentRequests), new { storeId = store.Id });
+            }
```
dennisreimann, line 134: To match the behaviour in the invoice controller, can you please also add this to the POST action of creating a payment request? And maybe we can then also refactor both occasions to use a private method for this case (per controller).

TChukwuleta: noted

dennisreimann (approved):
tACK. I modified it so that both controllers redirect to the list in that case.

dstrukt (approved):
LGTM, apologies for the delay in review - tACK!

pavlenex (approved):
tACK

PR #5628, opened 2023-12-28
Checkout v1: Apply custom style

### LINKED ISSUE #5615: Classic checkout - custom style doesn't apply to the checkout page
Context:
In BTCPay Server we have two types of checkouts: classic and new one. Classic (Legacy) is kept for backward compatibility with older checkout pages, as some users are using custom css and we want to avoid breaking changes for them.

Problem:
The classic page doesn't render the custom css anymore, not even default themes dark or legacy.

To replicate:
1. In Checkout Appearance, select classic
2. Select one of the pre-defined themes (for example dark)
3. Create an invoice
4. Visit the checkout paeg
5. You'll see that custom styling doesn't' apply.
6. BTCPay Server v1.12.3+f23078df1

### LINKED ISSUE #5616: When switching from new checkout to classic, use a custom css and go back to new checkout, the new checkout renders legacy css instead of branding
This is a low-priority issue that I encountered when trying to play around with #5615, but basically if user had branding options with new checkout, then in an unlikely scenario goes back to classic checkout, adds custom css, tries it out, doesn't like it and, switches to new checkout again, the new checkout is a mixture of that legacy checkout css and new branding.

Video shows this better.

It seems that somehow custom css embedded in classic still influences the css in new checkout, which shouldn't be the case as they're distinct. 

https://github.com/btcpayserver/btcpayserver/assets/36959754/5f8e591f-d8a2-488d-9a3d-a369864b4f6e

dennisreimann:
Applies the custom CSS in Checkout v1 and prevents that it interferes with the styling of Checkout v2.

Fixes #5615 and fixes #5616.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -856,10 +856,8 @@ namespace BTCPayServer.Controllers
                 DefaultLang = lang ?? invoice.DefaultLanguage ?? storeBlob.DefaultLang ?? "en",
                 ShowPayInWalletButton = storeBlob.ShowPayInWalletButton,
                 ShowStoreHeader = storeBlob.ShowStoreHeader,
-                StoreBranding = new StoreBrandingViewModel(storeBlob)
-                {
-                    CustomCSSLink = storeBlob.CustomCSS
-                },
+                StoreBranding = new StoreBrandingViewModel(storeBlob),
+                CustomCSSLink = storeBlob.CustomCSS,
                 CustomLogoLink = storeBlob.CustomLogo,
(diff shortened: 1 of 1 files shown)


dstrukt (approved):
To test, I just changed some very basic text & background colors, as well as targeting a specific class (all standard CSS) to see if the `custom CSS` input works for the V1 Checkout - all works as expected!

tACK!

Made me curious how many people are still using the OG checkout experience.

Kukks (approved):

pavlenex (approved):
tACK, tested and it resolves both  https://github.com/btcpayserver/btcpayserver/issues/5615 **and** https://github.com/btcpayserver/btcpayserver/issues/5616.

PR #5629, opened 2023-12-28
Store Email Settings: Improve configuration

### LINKED ISSUE #5623: Introduce a setting for Server Setting SMPT to be used for all stores in Server Settings > Email
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/5621

<sup>Originally posted by **iBobik** December 26, 2023</sup>
Currently each store requires to configure own SMTP settings, but it is very impractical if I am ok with using same settings for more stores - password cannot be recovered from existing store to copy it into new store and many SMTP servers cannot recover password too (just generate new password).

Could it be optional to configure SMTP? When not configured, it will use server global SMTP.</div>

dennisreimann:
This works with the existing settings and provides better guidance about the different store email cases. Closes #5623.

## No Fallback

- Either the SMTP settyings on the server level are not configured or the "Disable stores from using the server's email settings as backup" policy is active. 
- Requires configuration of SMTP settings on the store level: No option for using the server fallback.

no-fallback

## Fallback is present and used

- Toggle indicates the availability of providing custom SMTP settings for the store
- By default the server's SMTP settings will be used

uses-fallback

## Fallback is present but custom settings are provided

with-fallback-but-custom

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -468,8 +468,16 @@ namespace BTCPayServer.Tests
             s.RegisterNewUser(true);
             s.CreateNewStore();
             
+            // Ensure empty server settings
+            s.Driver.Navigate().GoToUrl(s.Link("/server/emails"));
+            s.Driver.FindElement(By.Id("Settings_Login")).Clear();
+            s.Driver.FindElement(By.Id("Settings_Password")).Clear();
+            s.Driver.FindElement(By.Id("Settings_From")).Clear();
+            s.Driver.FindElement(By.Id("Save")).Submit();
+            
             // Store Emails without server fallback
             s.GoToStore(StoreNavPages.Emails);
+            s.Driver.ElementDoesNotExist(By.Id("UseCustomSMTP"));
             s.Driver.FindElement(By.Id("ConfigureEmailRules")).Click();
             Assert.Contains("You need to configure email settings before this feature works", s.Driver.PageSource);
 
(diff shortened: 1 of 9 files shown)


BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -3110,15 +3123,13 @@ private static void CanBrowseContent(SeleniumTester s)
 
         private static void CanSetupEmailCore(SeleniumTester s)
         {
+            s.Driver.ScrollTo(By.Id("Settings_Server"));
```
dennisreimann, line 3126:
```suggestion
            s.Driver.ScrollTo(By.Id("QuickFillDropdownToggle"));
```

dstrukt (approved):
This is so much better!

I didn't run it through a rigorous technical review, but functionally it worked as expected.

Will be a nice quality of life improvement for admins - tACK!

Kukks (approved):

dennisreimann:
Feedback from design call: @pavlenex suggested to move (or mirror) the policy setting on the server SMTP setup view.

I think we should also invert the wording to make it "enable" instead of "disable" on the server level, as this makes it easier to think about it.

dennisreimann:
## Server policies 

n1

## Store email settings

n2

dstrukt:
Few nits:

- [ ] Shouldn't this be under `Email Settings` instead of `Notification Settings` in the Server Policies?

Copy suggestion, but don't necessarily know if it's any better, but is more consistent with the custom messages..`Use the Server's SMTP settings across all stores to send emails. This can be customized at the store level.` cc @pavlenex 

Otherwise, tACK!

dstrukt (changes requested):
Just flagging the above, this could be shipped as is!

dennisreimann:
As discussed yesterday. This is now ready to be merged. 

grafik

PR #5630, opened 2023-12-29
Payments Report: Add POS data columns

### LINKED ISSUE #5290: [Bug]: Tips and invoiced amount are inseparable in csv

`BTCPay Server v1.11.4+79b2f1652`

Lunanode deployment

### What happened?

I'm unable to separate tips from total amounts in the csv.

- I, the admin, enable tipping for a store, using the Keypad-only checkout
- The cashier enters the amount of the service, e.g. $1
- The customer enters the amount of the tip, e.g. 20%
- The customer pays the invoice
- I download the csv for the store
**- I'm unable to find out how much the service cost, how high of a tip the customer paid**

This makes it very hard to run a store with tipping enabled. How am I supposed to pay out the tips to the servers at the end of the day if I don't know how many tips were paid?

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Extends the payments report data with subtotal, tip and discount amounts. This works for keypad and cart sales. Closes #5290.

data


dstrukt (approved):
If I understand the scope of the PR, functional tACK!

BTCPayServer/Services/Reporting/PaymentsReportProvider.cs
```diff
@@ -180,15 +183,22 @@ public override async Task Query(QueryContext queryContext, CancellationToken ca
             values.Add(invoiceBlob.Currency);
```
Kukks, line 183:
```suggestion
            values.Add(currency);
```

Kukks (changes requested):
I don't know how I feel about this, it's so specific to the POS that it would be a shame to pollute such a wide-scoped report. I would see if if is possible to create a dedicated report that extends this one.

dstrukt:
> I don't know how I feel about this, it's so specific to the POS that it would be a shame to pollute such a wide-scoped report. I would see if if is possible to create a dedicated report that extends this one.

That's valid. Makes future plug-in (tip splitting, podcast splits, whatever other stuff people come up..) more easily extendible, but can also see the case for simply including the core POS data breakdown here as well.

pavlenex:
> I don't know how I feel about this, it's so specific to the POS that it would be a shame to pollute such a wide-scoped report. I would see if if is possible to create a dedicated report that extends this one.

So you're suggesting we have a dedicated app report or a pos report where we could display data for the POS app in particular? I agree with you on that one if so, thinking about `refund` report we want to have, this probably shouldn't be in the main one as it's quite app-specific from what I understood.

NicolasDorier:
I believe that we should allow the ability to add or remove columns in the report page.
The backend just fetch all the data, and the front end hide/show columns based on user's preferences.

pavlenex:
@NicolasDorier Could we display all data including refunds & tips with that approach? I think that's quite sensible idea for this problem, and would allow people to customize things, we're getting a lot of complaints in general that people aren't getting info they used to from the reporting, I am still working on their specific feedback.

dstrukt:
cACK to showing/hiding columns as per @NicolasDorier's suggestions, seems like an elegant approach to customizing the view for the user's use case

cc @dennisreimann are we in a position to do this with tables? imagine it'll just be an additional bootstrap column with some kind of `+` button with a dropdown of potential items to add. conceptually it seems like it might work, but without actually looking at the view concretely (might do this evening and edit my post), it's hard to say

BTCPayServer/Services/Reporting/PaymentsReportProvider.cs
```diff
@@ -56,6 +56,9 @@ ViewDefinition CreateViewDefinition()
                 new ("InvoiceCurrency", "string"),
                 new ("InvoiceCurrencyAmount", "amount"),
                 new ("Rate", "amount")
```
dennisreimann, line 58:
```suggestion
                new ("Rate", "amount"),
```

BTCPayServer/Services/Reporting/PaymentsReportProvider.cs
```diff
@@ -137,7 +137,7 @@ public override async Task Query(QueryContext queryContext, CancellationToken ca
                 var prompt = invoice.GetPaymentPrompt(PaymentTypes.LNURL.GetPaymentMethodId("BTC"));
                 var consumerdLightningAddress = prompt is null || handler is not LNURLPayPaymentHandler lnurlHandler ? null : lnurlHandler.ParsePaymentPromptDetails(prompt.Details).ConsumedLightningAddress;
                 values.Add(consumerdLightningAddress);
-                values.Add(invoice.Currency);
+                values.Add(currency);
```
dennisreimann, line 140:
```suggestion
                var currency = invoiceBlob.Currency;
                values.Add(currency);
```

PR #5639, opened 2024-01-02
Hide LN Balance when using internal node and not server admin

Kukks:


BTCPayServer/Components/StoreLightningBalance/Default.cshtml
```diff
@@ -1,5 +1,8 @@
 @model BTCPayServer.Components.StoreLightningBalance.StoreLightningBalanceViewModel
-
+@if(!Model.InitialRendering && Model.Balance == null)
+{
+   return;
+}
```
dennisreimann, line 5: Wouldn't it be better to handle this on the Dashboard view-level?

dennisreimann: I get this is component internally and it makes sense, but imho we should then also hide the complete LN section on the dashboard and prevent those tiles from showing that way as well.

dennisreimann: Ok, checked and it doesn't work for this tile, because we also need to distinguish external/internal node in the check.

dennisreimann (approved):
tACK

dstrukt (approved):
tACK!

PR #5640, opened 2024-01-02
Server Users: More precise message when inviting users

dennisreimann:
This lets the admin who invited a new user know whether or not an email has been sent. If the SMTP server hasn't been set up, they need to share the invite link with the user.

- Old: `Account created without a set password. An email will be sent (if configured) to set the password. You may alternatively share this link with them: INVITE_LINK`
- New (no email): `Account created without password. An email has not been sent, because the server does not have an email server configured. You need to share this link with them: INVITE_LINK`
- New (with email): `Account created without password. An email has been sent to set the password. You may alternatively share this link with them: INVITE_LINK`

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.Users.cs
+++ b/BTCPayServer/Controllers/UIServerController.Users.cs
@@ -9,6 +9,7 @@ using BTCPayServer.Data;
 using BTCPayServer.Events;
 using BTCPayServer.Models.ServerViewModels;
 using BTCPayServer.Services;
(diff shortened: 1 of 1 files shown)


dstrukt:
Nice update - will give it a look this evening!

dstrukt (approved):
tACK - LGTM!

We can push this further when we start to address work for the App as well.

dennisreimann:
Made this a part of #5714, we can close this one.

PR #5643, opened 2024-01-03
Refactor fee provider

### LINKED ISSUE #5641: [Bug]: Gateway Timeout (504) on Invoice Creation because Mempool.space is down

 © BTCPay Server v1.12.3+f23078df1 

I enabeld the BTC onchain store wallet (which I had prior disabled because of high fees) and then created an invoice in my store. This loaded until it hit a nginx proxy timeout which resulted in this message:

grafik

### What happened?

504 Gateway Timeout

It seems that this problem coexists with a downtime of the mempool.space website.
This shouldn't happen as we have a fallback with the internal BTC node.

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox

### Additional information

none

Kukks:
The fee provider ended up glued with a hardcoded factory. This PR:
* removes this glue and uses the DI to register fee provider for a network. (allows plugins to add their own fee providers, for any network
* Add a 10 second timeout to mempoolspace fee fetching as they are slow at times (many reports of downtime issues within past 24 hours)
Fixes #5641.

### DIFF
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -28,6 +28,7 @@ using BTCPayServer.Plugins.Bitcoin;
 using BTCPayServer.Rating;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Apps;
+using BTCPayServer.Services.Fees;
 using BTCPayServer.Services.Invoices;
 using BTCPayServer.Services.Labels;
 using BTCPayServer.Services.Rates;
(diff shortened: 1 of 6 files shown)


NicolasDorier:
I will remove the refactor you did to focus on the bug fix.

The refactor still hardcode the mempool fee provider, but on top of this limit fee providers to depends solely on bootstrap services. We should probably do something later, but for now let's just focus on fixing the bug.

NicolasDorier:
* Rewrote the randomizer: This was adding or removing exactly 10% rather than picking random value on a range of +-10%
* Added tests of `InterpolateBound` and `RandomizeByPercentage`
* Rewrote the `InterpolateBound` to make it more concise, removing branches on the way
* Removed the registration of providers in DI for the moment.

PR #5647, opened 2024-01-05
Admins can approve registered users

### LINKED ISSUE #5548: Admins should have an option to approve registered users to their instance
If someone decides to expose their instance to public, I think we should have an option where admin first has to approve registered users. This is to prevent abuse and perhaps malicious users registering. I think the ability for admin to allow registertaion to anyone, can still be an option, but by default, I believe that admins should have the ability to approve users manually. 

This feature can be re-used in BTCPay App as well. If we expect ambassadors to onboard people, we should ensure they have sufficient mechanisms to pick who registers on their public instance.

dennisreimann:
High-level changes:

- Introduces the `RequiresUserApproval` flags for the server policies
- Introduces the `RequiresApproval` and `Approved` flags for the user
- Introduces the `UserService.TryCanLogin` method for checking the various aspects of whether or not a user is allowed to sign in (approval, email confirmation, lock out state) and uses it consistently
- User list reflects the toggle states (approval, email confirmation, lock out) better and omits indicators in case a setting isn't required (e.g. don't show a user is approved when their approval wasn't required on account creation)
- Adds support for (un)approving via the Greenfield API
- Cleans up the account controller

Closes #5548.

## Policies

I made it so that the sub-settings for registration aren't displayed, if public registration isn't available.

policies

## Users List

Here the requirements for confirming the email address and approval only existed when `unapproved@btcpay.local` registered, so these indicators are only shown for that user.

users

## Unapproved user login

unapproved


BTCPayServer.Tests/BTCPayServerTester.cs
```diff
@@ -245,6 +245,9 @@ public async Task StartAsync()
                 rateProvider.Providers.Add("kraken", kraken);
             }
 
+            // reset test server policies
```
dennisreimann, line 248: This was made to ensure a consistent setup. It helps in case the individual tests do not (re)set the policies themselves or when the server policies were edited during development.

BTCPayServer.Tests/SeleniumTester.cs
```diff
@@ -76,6 +77,11 @@ public async Task StartAsync()
                     // A bit less than test timeout
                     TimeSpan.FromSeconds(50));
             }
+            
+            // reset test server policies
```
dennisreimann, line 81: This was made to ensure a consistent setup. It helps in case the individual tests do not (re)set the policies themselves or when the server policies were edited during development.

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -437,32 +441,37 @@ public async Task<IActionResult> LoginWith2fa(LoginWith2faViewModel model, bool
             {
                 throw new ApplicationException($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
             }
+            if (!UserService.TryCanLogin(user, out var message))
```
dennisreimann, line 444: Added this in places where the basic checks were missing.

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -504,30 +513,35 @@ public async Task<IActionResult> LoginWithRecoveryCode(LoginWithRecoveryCodeView
             var user = await _signInManager.GetTwoFactorAuthenticationUserAsync();
             if (user == null)
             {
-                throw new ApplicationException($"Unable to load two-factor authentication user.");
+                throw new ApplicationException("Unable to load two-factor authentication user.");
+            }
+            if (!UserService.TryCanLogin(user, out var message))
```
dennisreimann, line 518: Added this in places where the basic checks were missing.

BTCPayServer/Models/ServerViewModels/UsersViewModel.cs
```diff
@@ -11,7 +11,8 @@ public class UserViewModel
             public string Id { get; set; }
             public string Name { get; set; }
             public string Email { get; set; }
-            public bool Verified { get; set; }
+            public bool? EmailConfirmed { get; set; }
```
dennisreimann, line 14: Renamed `Verified` to `EmailConfirmed` to make it more precise.

BTCPayServer/Services/PoliciesSettings.cs
```diff
@@ -14,6 +14,19 @@ public class PoliciesSettings
...
         [Display(Name = "Disable new user registration on the server")]
         public bool LockSubscription { get; set; }
+        
+        [JsonIgnore]
+        [Display(Name = "Enable new user registration on the server")]
+        public bool EnableRegistration
+        {
+            get => !LockSubscription;
+            set { LockSubscription = !value; }
+        }
```
dennisreimann, line 24: Inverted the property for the view, because it makes it easier to reason about from that perspective.

BTCPayServer/Services/PoliciesSettings.cs
```diff
@@ -32,6 +45,14 @@ public class PoliciesSettings
...
         [Display(Name = "Disable non-admins access to the user creation API endpoint")]
         public bool DisableNonAdminCreateUserApi { get; set; }
+        
+        [JsonIgnore]
+        [Display(Name = "Non-admins can access the user creation API endpoint")]
+        public bool EnableNonAdminCreateUserApi
+        {
+            get => !DisableNonAdminCreateUserApi;
+            set { DisableNonAdminCreateUserApi = !value; }
+        }
```
dennisreimann, line 55: Inverted the property for the view, because it makes it easier to reason about from that perspective.

dstrukt:
Amazing - going to try to break it this evening!!

Kukks (changes requested):
I haven't reviewed the entire thing yet but the first thing that comes to mind is:
Can we reuse the authorization policies we have in place instead of so much custom logic? 

* If the server has a policy for user approval, assign a role to the user "PendingApproval"
* Create a policy "CanLogin", that checks the user's roles

NicolasDorier:
code review ACK.

@Kukks there isn't actually lot's of custom logic. The only place that check for the property is actually  [APIKeysAuthenticationHandler.cs](https://github.com/btcpayserver/btcpayserver/pull/5647/files#diff-5a05a94dc966aa5c9f63431786c5e0cb4fad8b0c331d36515ab97ff4d663b408) for the API and the HTTP POST Login.

There is lots of code impacted mainly because there is a new Greenfield Route for this, and some UI changes, which can't really be better.

NicolasDorier:
@dennisreimann the greenfield API can be used through Basic auth rather than API, so you need to also `BasicAuthenticationHandler`

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -567,52 +581,63 @@ public async Task<IActionResult> Register(RegisterViewModel model, string return
...
+                    if (policies.RequiresConfirmedEmail)
                     {
-                        if (logon)
-                            await _signInManager.SignInAsync(user, isPersistent: false);
-                        return RedirectToLocal(returnUrl);
+                        TempData[WellKnownTempData.SuccessMessage] = "Account created, please confirm your email.";
+                        return RedirectToAction(nameof(Login));
                     }
-                    else
+                    if (policies.RequiresUserApproval)
```
NicolasDorier, line 626: If there is both `RequiresConfirmedEmail` and `RequiresUserApproval`, then the only message shown will be about user approval.

BTCPayServer/Services/UserService.cs
```diff
@@ -46,26 +46,64 @@ public async Task<List<ApplicationUserData>> GetUsersWithRoles()
...
+            return user.Approved || !user.RequiresApproval;
+        }
+
+        private static bool IsDisabled(ApplicationUser user)
         {
             return user.LockoutEnabled && user.LockoutEnd is not null &&
                    DateTimeOffset.UtcNow < user.LockoutEnd.Value.UtcDateTime;
         }
+        
+        public static bool TryCanLogin(ApplicationUser? user, out string? error)
```
NicolasDorier, line 81:
```suggestion
        public static bool TryCanLogin([NotNullWhenAttribute(true)] ApplicationUser? user, out string? error)
```

This way, if the method returns `true` you don't have to check `user` for null and can remove stuff like `user!`

NicolasDorier: Just a note for this method, I would prefer if you used a discriminated unions pattern via records rather than sending error back as a string, this way the caller can check what error is this without having to parse the error string.

That said, don't change it now this is fine.

BTCPayServer/Controllers/UIAccountController.cs
dennisreimann: Fixed in 51df12d.

dennisreimann:
> the greenfield API can be used through Basic auth rather than API, so you need to also BasicAuthenticationHandler

Added in 0ac5e25.

dennisreimann:
Feedback from design call: We should have an additional email which informs the user about approval.

dennisreimann:
Now an email is sent on approval: "Your account has been approved and you can now [login here](#loginlink)".

pavlenex:
I've asked two third-party hosts who use custom roles to provide us more details on what they use so we can flash these out more precisely, will update here when I get a reply.

dennisreimann:
Added a notification for the admins, which leads to the user details view, where they can approve the new user.

Bildschirmfoto vom 2024-01-15 17-09-46

dstrukt:
> Added a notification for the admins, which leads to the user details view, where they can approve the new user.
> 
> We could use a new notification icon for that, @dstrukt
> 
> Bildschirmfoto vom 2024-01-15 17-09-46

I think we could probably reuse the "Account" or profile icon (from the sidebar) for the time being.

dennisreimann:
Thought so too, but wanted to check this first, because the other notification [icons](https://design.btcpayserver.org/design/icons/) have a different style (rounded and colored).

pavlenex:
Upon testing this PR, I was a bit confused, since now we introduce three properties for users:

- Verified (Email verification successful)
- Approved (Approved by admin)
- Enabled (Active)

While I don't think they're confusing in terms of naming, the UI for approving and enabling users isn't unified. To approve the user you need to open user view, where as to enable them, you can do that from users list which leads to a bit of ambiguity of admins enabling users (since UX is in place), thinking they've approved them.

I don't have a solution on how this can be solved, but logging it so we can discuss it on a dev call.

dstrukt:
> Amazing - going to try to break it this evening!!
> 
> First thoughts are the various checkmarks/Xs.
> 
> Feels like the Enabled/Approved/Verified (think this is for the email verificiation IIRC) rows could all be consolidated into a Status column:
> 
> `Disabled` (red), `Awaiting Email Verification`(yellow), `Pending Admin Approval`(yellow), or `Active` (green - for enabled) .. The middle 2 could be wordsmithed a bit, but you get the idea.
> 
> Obviously outside of the scope of this PR, but wanted to float the idea while things are in flux. Especially as my brain starts to tune out when I see a tic-tac-toe board with more than 3+ users 😂

@pavlenex To address your concern, still think consolidation of ideas as just the Status of a user makes the most sense, imo.

A user is either `Pending Approval`, `Pending Email Verification`, `Disabled`, or `Active` (passed 1 of the first 2 stages).

dennisreimann:
grafik

dstrukt:
Love it - can parse the state of the user and then make a determinate action! Are we missing the `Reject` functionality as well in the ss above though? 

Was great before, but with the latest changes, will give it a final proper review this evening!

Last thought, in the future, we'll probably want to distinguish the style of a status vs. a label vs. a tag, etc.. a bit more so they don't all appear so similar - noticed it more as we clean up some of the views.

dstrukt:
> > are we missing the Reject functionality as well in the ss above though?
> 
> I think that could be simply the "Remove", right?

Ahh true, valid point!

dstrukt (approved):
All the latest changes look great - functional tACK!

pavlenex (commented):
UX-wise these improvements are amazing, much better than I hoped, and something that can only be experienced through testing.

It seems there's no distinction between disapproved and disabled UX wise, which I think is good, but if someone was to create a plugin that would now allow users who haven't paid an invoice to log in, which state should be used in that case, disapproved or disabled or even technically it doesn't really make a big difference?

dennisreimann:
> if someone was to create a plugin that would now allow users who haven't paid an invoice to log in, which state should be used in that case, disapproved or disabled or even technically it doesn't really make a big difference?

Technically it doesn't make much difference for the end user, because they won't be able to log in. The wording for the failure messages is different though (need approval vs. locked out).

But from the admin perspective there is a hierarchy: If the approval policy is active, users need approval before they can even be disabled. I think it makes sense that way and the workflow for the use case you described would be, that an admin would first need to approve a user and the user would then need to pay the invoice to be enabled — granted the plugin takes care of setting the initial user state to disabled until an invoice is paid.

dstrukt:
Agreed, good way to frame the status.

Additional thought I had while reviewing this, and I suppose it's a bit of scope, but could use more discussion in a design call, is the user `Type`. It could arguably also be its own column and more explicitly defined, e.g. currently we have `Server Admin` OR no label. It could also be helpful for the sake of ordering, and/if we introduce other Server User types .. `Ambassador`, `Maintainer`, etc..

dstrukt:
Follow up thought on the CTA above, could arguably shorten `Resend verification email` to `Resend email` to get a little space back, as the context + the CTA are clear enough. Minor nit.

pavlenex:
@dennisreimann I've made a table comparing what you outline vs what one of the ambassadors proposed.


Permission | Bitcoin People | D11's Suggestion
-- | -- | --
Owner |   |  
btcpay.store.canmodifystoresettings | ✅ | ✅
btcpay.store.cantradecustodianaccount | ✅ | ✅
btcpay.store.canwithdrawfromcustodianaccount | ✅ | ✅
btcpay.store.candeposittocustodianaccount | ✅ | ✅
Store Admin / Brand Admin |   |  
btcpay.store.canmodifystoresettings | ✅ | ❌
Employee |   |  
btcpay.store.canmodifyinvoices | ✅ | ✅
btcpay.store.canmodifypaymentrequests | ✅ | ✅
btcpay.store.cancreatenonapprovedpullpayments | ✅ | ✅
btcpay.store.canviewpayouts | ✅ | ✅
btcpay.store.canviewpullpayments | ✅ | ✅
Store Manager |   |  
btcpay.store.canviewstoresetting | ✅ | ❌
btcpay.store.canmodifyinvoices | ✅ | ✅
btcpay.store.webhooks.canmodifywebhooks | ✅ | ✅
btcpay.store.canmodifypaymentrequests | ✅ | ✅
btcpay.store.canmanagepullpayments | ✅ | ✅
btcpay.store.canmanagepayouts | ✅ | ✅
btcpay.store.canviewcustodianaccounts | ✅ | ✅
btcpay.store.candeposittocustodianaccount | ✅ | ✅
Employee Readonly |   |  
btcpay.store.canviewinvoices | ✅ | ❌
Guest |   |  
btcpay.store.canviewstoresettings | ❌ | ✅
btcpay.store.canmodifyinvoices | ❌ | ✅
btcpay.store.canviewcustodianaccounts | ❌ | ✅
btcpay.store.candeposittocustodianaccount | ❌ | ✅

<br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">

BTCPayServer.Tests/SeleniumTester.cs
NicolasDorier: Why is it present in `SeleniumTester`? The reset is already done by `BTCPayServerTester.StartAsync`

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -107,7 +106,7 @@ public async Task<IActionResult> CheatPermissions([FromServices] IAuthorizationS
         [AllowAnonymous]
         public async Task<IActionResult> Login(string returnUrl = null, string email = null)
         {
-            if (User.Identity.IsAuthenticated && string.IsNullOrEmpty(returnUrl))
+            if (User.Identity is { IsAuthenticated: true } && string.IsNullOrEmpty(returnUrl))
```
NicolasDorier, line 109: it isn't more readable than `User.Identity.IsAuthenticated`

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -437,32 +440,37 @@ public async Task<IActionResult> LoginWith2fa(LoginWith2faViewModel model, bool
...
             var result = await _signInManager.TwoFactorAuthenticatorSignInAsync(authenticatorCode, rememberMe, model.RememberMachine);
-
             if (result.Succeeded)
             {
-                _logger.LogInformation("User with ID {UserId} logged in with 2fa.", user.Id);
+                _logger.LogInformation("User with ID {UserId} logged in with 2fa", user.Id);
                 return RedirectToLocal(returnUrl);
             }
-            else if (result.IsLockedOut)
+            if (result.IsLockedOut)
```
NicolasDorier, line 460: I think it can't happen anymore, because `TryCanLogin` check for it

BTCPayServer/Controllers/UIServerController.Users.cs
```diff
@@ -74,44 +70,67 @@ public partial class UIServerController
...
         {
             var user = await _UserManager.FindByIdAsync(userId);
             if (user == null)
                 return NotFound();
 
+            bool? propertiesChanged = null;
+            bool? adminStatusChanged = null;
+            bool? approvalStatusChanged = null;
+
+            if (user.RequiresApproval && viewModel.Approved.HasValue && user.Approved != viewModel.Approved)
```
NicolasDorier, line 102: `user.Approved != viewModel.Approved` seems superflous, as after you set `approvalStatusChanged`.

BTCPayServer/Controllers/UIServerController.Users.cs
```diff
@@ -74,44 +70,67 @@ public partial class UIServerController
...
+        [HttpPost("server/users/{userId}")]
         public new async Task<IActionResult> User(string userId, UsersViewModel.UserViewModel viewModel)
         {
             var user = await _UserManager.FindByIdAsync(userId);
             if (user == null)
                 return NotFound();
 
+            bool? propertiesChanged = null;
+            bool? adminStatusChanged = null;
+            bool? approvalStatusChanged = null;
```
NicolasDorier, line 100: please just condense this into a single boolean `isUpdated=false` and you set it like this `isUpdated |= await _userService.SetUserApproval(user.Id, viewModel.Approved.Value, Request.GetAbsoluteRootUri())`

BTCPayServer/Controllers/UIAccountController.cs
dennisreimann: It's because the IDE hinted at a potential null pointer problem with `User.Identity`. But agree and will revert this.

BTCPayServer/Controllers/UIServerController.Users.cs
dennisreimann: I made it like this, because depending on what changed, different methods need to get called later on in the code. If the `adminStatusChanged` we invoke `_userService.SetAdminUser`, the `approvalStatusChanged` is handled by `_userService.SetUserApproval` and `propertiesChanged` will invoke `_UserManager.UpdateAsync`.

BTCPayServer/Controllers/UIServerController.Users.cs
dennisreimann: Done in 12b9d7e.

BTCPayServer/Controllers/UIAccountController.cs
dennisreimann: Done in 12b9d7e.

BTCPayServer/Controllers/UIAccountController.cs
dennisreimann: Done in 12b9d7e.

NicolasDorier (approved):

PR #5649, opened 2024-01-08
Store Emails: Fix test email with multiple recipients

### LINKED ISSUE #5648: [Bug]: Store emails - "Error sending test email: Unexpected token at offset 20"

I'm running the latest version BTCPay Server v1.12.3+f23078df1

Dockerized

### What happened?

When trying to send 1 email to multiple email addresses, separated by commas, it throws an error

"Error sending test email: Unexpected token at offset 20" (Note: The "Offset" value changes as I change the string of the combined email addresses, so this has something to do with the parsing of the string and looking for the "comma" value?

The workaround is to create multiple entries, copies of each other, and send each one to one of the required email addresses.

Tried leaving spaces between the email addresses besides the commas, but makes no difference.

image

Also, users have been reporting that they don't get the email that's being sent to them by using the "Send the email to the buyer if email was provided to the invoice" option. I have it configured like this ( it worked until a previous BTCPayServer version), to send it to the storeowner, and also send it to the buyer ( who provides the email address )

image

Explained above. This worked ok until the last or previous to last upgrade.

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

Nothing is generated in the "docker logs -f generated_btcpayserver_1" log when the error happens

```[tasklist]
### Tasks
```

dennisreimann:
Fixes #5648.

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.Email.cs
+++ b/BTCPayServer/Controllers/UIStoresController.Email.cs
@@ -74,11 +74,10 @@ namespace BTCPayServer.Controllers
                 var rule = vm.Rules[i];
 
                 if (!string.IsNullOrEmpty(rule.To) && (rule.To.Split(',', StringSplitOptions.RemoveEmptyEntries)
-                        .Any(s => !MailboxAddressValidator.TryParse(s, out var mb))))
+                        .Any(s => !MailboxAddressValidator.TryParse(s, out _))))
                 {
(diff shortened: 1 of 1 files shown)


Kukks (approved):

dstrukt (approved):
tACK!

worked on a hotmail/protonmail/gmail split 😂

PR #5650, opened 2024-01-08
Dashboard: Tooltip for balance on a particular day

### LINKED ISSUE #5617: Show balance on a particular day in dashboard
Several times already, users have echoed feedback that it's not possible for them to check their earnings on a particular day. 

Screenshot 2023-12-26 at 00 49 35

There are two possible way to approach this, or both:

1. Currently we show a chart line but if you hover over dots you don't really see the balance on a day. So we could show balance on particular day there Screenshot 2023-12-26 at 00 44 13
2. Add ability to show balance on particular day from the calendar

We should discuss this first before implementing.

dennisreimann:
Closes #5617.

Peek 2024-01-08 19-16

### DIFF
--- a/BTCPayServer/Components/StoreWalletBalance/Default.cshtml
+++ b/BTCPayServer/Components/StoreWalletBalance/Default.cshtml
@@ -64,13 +64,14 @@
             
             const id = `StoreWalletBalance-${storeId}`;
             const baseUrl = @Safe.Json(Url.Action("WalletHistogram", "UIWallets", new { walletId = Model.WalletId, type = WalletHistogramType.Week }));
+            const valueTransform = value => rate
+                ? DashboardUtils.displayDefaultCurrency(value, rate, defaultCurrency, divisibility).toString()
+                : value
             const chartOpts = {
                 fullWidth: true,
                 showArea: true,
                 axisY: {
-                    labelInterpolationFnc: value => rate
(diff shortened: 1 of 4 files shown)


Kukks (approved):
ACK, but I think we usually have a test that verifies the hash of the js files?

dennisreimann:
That's problematic in this case, because afaik it isn't distributed and I had to do a modification to it, because the [source](https://github.com/ant7/chartist-plugin-tooltip2) is six years old.

matman7x (approved):

dstrukt (approved):
Debated if I thought we should also put the unit `{X} {USD/BTC/etc..}` on the tooltip, but think it's very clear given the context. 

tACK - nice update!

BTCPayServer/Components/StoreWalletBalance/Default.cshtml
```diff
@@ -64,32 +64,40 @@
...
-                    labelInterpolationFnc: value => rate
-                        ? DashboardUtils.displayDefaultCurrency(value, rate, defaultCurrency, divisibility).toString()
-                        : value
+                    labelInterpolationFnc: valueTransform
                 }
             };
             
             const render = data => {
                 let { series, labels } = data;
+                console.log(series, labels)
```
NicolasDorier, line 80: debug statement

PR #5659, opened 2024-01-13
Crowdfund : Add Buyer information / Additional information(forms) like POS

Nisaba:
Add the ability tho get buyer information or additional information via forms to a crowdfunf.
Same thing than POS.

### DIFF
--- a/BTCPayServer/wwwroot/crowdfund/app.js
+++ b/BTCPayServer/wwwroot/crowdfund/app.js
@@ -28,8 +28,8 @@ document.addEventListener("DOMContentLoaded",function (ev) {
         },
         computed: {
             canExpand: function(){
-                return !this.expanded 
-                    && this.active && 
+                return !this.expanded
+                    && this.active &&
                     (this.perk.inventory==null || this.perk.inventory > 0)
             }
         },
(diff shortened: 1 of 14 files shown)


dstrukt:
Thanks for the PR! Will take a look this weekend - tagging @dennisreimann for review as well.

BTCPayServer/wwwroot/crowdfund/app.js
```diff
@@ -220,7 +220,11 @@ document.addEventListener("DOMContentLoaded",function (ev) {
             },
             contribute() {
                 if (!this.active || this.loading) return;
-                
+
+                if (this.srvModel.hasFormForExtraValues) {
+                    window.location.href = this.srvModel.formUrl;
```
dennisreimann, line 225: This could be condensed to just formURL, also in the view model.

Nisaba: HasFormExtraValues boolean is set in the Crowdfundplugin.cs file, depending on the settings.
In this file, we don't have access to controller to get the Url.Action method.
Would need extra code more complicated I think...

dennisreimann: I updated it in d3dab94.

dennisreimann (changes requested):
Currently it only works for the "no perks" case. If there are perks, they need to be selected first and afterwards the form would need to be filled.

Nisaba:
OK, noted.
I will fix this before the end of this week.

dennisreimann:
Please also add a test for both cases (pers and no perks), so that we can ensure the crowdfunding with form case will also be considered moving forward. Imho that's important, because the flow is very JS heavy and can easily be missed.

Nisaba:
@dennisreimann fix for the perk has been made. You can check it.
I have to add tests; i hope this week-end...

dennisreimann (approved):
tACK. Thanks for the contribution!

dstrukt (approved):
tACK - LGTM.

+1 Thank you for the contribution - great work!

Kukks (approved):

PR #5662, opened 2024-01-16
Allow translations of BTCPay Server Backend by admins

NicolasDorier:
## Abstract

Sourced from https://github.com/btcpayserver/btcpayserver/discussions/5468

To achieve our goal of widespread adoption of BTCPay Server by local communities, it is crucial to focus efforts on translating BTCPay. As it stands, only the checkout pages are translated, implying that BTCPay merchants must know English. This requirement significantly hinders local adoption.

The proposed solution should possess these characteristics:

* Enable administrators to modify or add translations during runtime.
* Utilize standard types such as `IStringLocalizer`, `IHtmlLocalizer`, or annotations for implementation.
* Facilitate the easy sharing of translations.
* @woutersamaey has already contributed to this endeavor, as seen in https://github.com/btcpayserver/btcpayserver/pull/972/, where he enabled these interfaces to be supported by a JSON file along with dedicated tag helpers.

We can reasonably presume that the backend operates in a specific language, which can be designated in the server's policies.

Translations should be stored in the database and loaded into memory during runtime.

## UX

This PR introduces backend translation capability in BTCPay Server for administrators. A new tab in server settings presents a large text field for updating translations, formatted as `[English] => [Translation]`. The simplicity of this UI facilitates easy modification and sharing of translations. A dedicated GitHub repository will host all translations, linked directly from this page.

Under `Server Settings -> Translations`, you can find the list of dictionary available. By default for this PR only English is available, I plan to ship more default in a later PR.

image

Administrators can effortlessly import translations from other people into the text area and share their translations.
For this, they copy the dictionary of their choice, then edit the lines needed.

image

You can then select this new dictionary either in the dictionary list page, or in the Server Policies page.

image

Untranslated strings from a dictionary will fallback back to the parent dictionary.
I plan to add a link to a github repository later so the community can share translations.
Prior to major release, we should probably show case a few languages translated through ChatGPT in such repository.

## Localizable content

Developers have multiple options for making page content translatable:
* `<label asp-for="Property">` in views are automatically translatable if the target property has a `[Display]` or `[DisplayName]` attribute.
* A `text-translate="true"` tag helper in views marks tag text content as translatable.
* * A `html-translate="true"` tag helper in views marks tag HTML content as translatable.
* In Views' code blocks, a `ViewLocalizer` object is injected for use as `ViewLocalizer["My sentence"]`.
* Controllers can use `IStringLocalizer`, `ViewLocalizer`, or `IHtmlLocalizer`.

## Implementation

The system employs standard MVC interfaces like `IStringLocalizer`, `ViewLocalizer`, and `IHtmlLocalizer`, customized to meet the abstract's outlined requirements. Translations differing from default English versions are stored in the database, ensuring immediate reflection of localized string updates for English-speaking administrators (unless overridden).

A `translations` table has been added with `key` and `value` columns.

### Default translations

Ensuring discoverability of translatable strings is vital. These strings are predefined in `Translations.Default.cs`.

To ease maintenance, a `UpdateDefaultTranslations` script functions similarly to `PullTransifexTranslations`. This script, which requires periodic execution, updates hardcoded assets by scanning *.cs files for Display and DisplayName usage, and .cshtml files for `ViewLocalizer` or `text-translate` tag helper use. Detected translations are then incorporated into `Translations.Default.cs`.

It is possible for developer to manually edit `Translations.Default.cs` to add more strings.

## Out of scope

* `Translations.Default` could be modularized into a plugin format. This would enhance the discoverability of translations for plugins.
* Currently, there is no integration with the Greenfield API. While this could be considered for future implementation, the practical benefits of such integration are yet to be fully ascertained.
* `text-translate` should be used in other parts of the source code, however, we can do this gradually rather than everything in this PR
* @woutersamaey make a taghelper on #972 which allow the translation of arbitrary attributes of tags (such as `tooltip=`), we should probably do this later too.


woutersamaey:
Great work @NicolasDorier  Thanks for picking up my (very old) work.

Some things I'd still like to see are:
- More than 1 language, because users can be international and not just from 1 region.
- A `locale` setting in the user's profile, which controls not only language, but also number format and date/time notation (AM/PM vs 24h time, which is still an issue for me since very long).
- Use the `Accept-Language` header for the user's preferred language if no `locale` is set in his profile, like we do with the checkout. The `locale` setting could default to "auto", which takes this header into consideration.

NicolasDorier:
The choice of a single language is because the goal of this PR is to improve local adoption in Bitcoin.

The expectation is that a Bitcoin ambassador onboard local businesses or friends on their local instance, and set the proper language for them so their first time experience with BTCPay is less intimidating.

I am also not comfortable to put translations in merchant level, because at some place, we will want to allow HTML injection. (For example if the string to translate contains bold letters in some part)

For the `locale`, the formatting should be done by the browser already. Apparently chrome take the data from the format set at the OS level? We should probably copy what the other websites do. But that should be a separate PR.

> Bundled translations that can update along with BTCPay

So if you are using English it's already the case. In other language it's more difficult, as we shouldn't override something that the user edited.

This will trigger weirdness in other language:
1. A language isn't translated in bundled form already, so the admin edit the translation directly to BTCPay.
2. The bundled form is released: Should we override the strings from the admin? It isn't clear we should, because the admin took time to translate things.

woutersamaey:
> The choice of a single language is because the goal of this PR is to improve local adoption in Bitcoin.

Makes sense. Just consider if we have an easy upgrade path after this, or if it would be major refactoring down the line? If the latter, might make sense to have basic structure already in place. At the bare minimum, I think we need to add a dropdown to indicate which language is entered in the textarea. If we don't know the language, we cannot upgrade later to multi-language because we won't know what was entered.

> The expectation is that a Bitcoin ambassador onboard local businesses or friends on their local instance, and set the proper language for them so their first time experience with BTCPay is less intimidating.

If we bundle 1 language (i.e. Spanish) + provide a way to load others, it wouldn't be intimidating at all.

> I am also not comfortable to put translations in merchant level, because at some place, we will want to allow HTML injection. (For example if the string to translate contains bold letters in some part).

The way I solve this typically, is by allowing some tags + escape everything else. Example: the translated output is escaped, except for `<strong>` and `<a>` tags. This means updating templates + careful consideration which tags you want to allow and which one's not to allow. In practise I use 2 groups: single line text can have `<strong>` + `<a>` and multiline text can have more like `<h1>`, `<ul>`, `<li>`, etc, so there's a whitelisting approach.

> Some parts are also not tied to a user or a store. (Login screen for example) This mean that there would be a need for different screen to update the "User level strings" and the "Server level strings". This makes things more complicated for everyone.

Why can't all translations be in 1 place, only accessible by the admin? The merchant is just a user and has to use the translations set by the admin.
 
> For the `locale`, the formatting should be done by the browser already. Apparently chrome take the data from the format set at the OS level? We should probably copy what the other websites do. But that should be a separate PR.
> 
> > Bundled translations that can update along with BTCPay
> 
> So if you are using English it's already the case. In other language it's more difficult, as we shouldn't override something that the user edited.
> 
> This will trigger weirdness in other language:
> 
> 1. A language isn't translated in bundled form already, so the admin edit the translation directly to BTCPay.
> 2. The bundled form is released: Should we override the strings from the admin? It isn't clear we should, because the admin took time to translate things and we don't want to lose his work.
> 
> The situation is the same if the admin edit the translation in BTCPay to fix a typo: When BTCPay update the translation in the bundle, maybe the sentence is a bit different, but we can't really override what the admin did, as we don't want to lose his work.
> 
> At the end of the day, there are too much edge case to consider for bundling without good solution. As such, I think simple is better and shouldn't spend too much time doing something too elaborate.

This is not an issue at all if we consider the bundled translations as defaults + use the textarea as a means to override specific lines. Most people don't want to bother with translations and will just use whatever is bundled or contibute via Transiflex. If a translation is missing or particulaly troublesome, they can override.

woutersamaey:
If you feel bundling translations is risky, we can always make it "opt-in", so nobody is forced into this...

matman7x (approved):

pavlenex:
Ok @NicolasDorier finally had some time to dedicate to user-testing this one, sorry for delays.

My initial thoughts, is that once I landed on `Server Settings > Translations` I've noticed this page. But it was unclear what I need to do.

Screenshot 2024-07-05 at 15 14 27

Is user  expected to add new translation, or do what?  

The only button I saw was `Copy` so I clicked on the `Copy`

Then I see a page where I am adding something called `Dictionary` To me that terminology was a bit confusing.

Ok I create a new dictionary. Then I see a bunch of strings. I assume now these strings need to be translated, but we don't inform user about them.

Screenshot 2024-07-05 at 15 19 25

In general I think the approach you're taking is okay, but the UX needs some improvement.

Some suggestions:
1. Keep consistency in call to action, on translation view, would it be possible to have add new language option as a call to  action? Once user clicks on add new translation they can select default or other strings. 
2. We should add a bit more context on each view. User can't know what this Translations are for, so adding a sentance that explains it on /dictionaries check how we add . 
Screenshot 2024-07-05 at 15 25 34
3. I am not really sold on the term dictionary, do you have any specific reason why you didn't call it a language?
4. Then on editing dictionary I would add a sentance on top explaining what user is expected to do. Perhaps even advising them to use check what i wrote in  2)

pavlenex:
One more note, it seems the sidebar is not translated or included in the lis of strings.

NicolasDorier:
> 3.  I am not really sold on the term dictionary, do you have any specific reason why you didn't call it a language?

Imagine than you have a Serbian language pre-shipped in BTCPay called "Serbian".
A new version of BTCPay get out, but some string aren't translated yet. (Or maybe some sentence in plugins need to be translated)

The user need to create a new dictionary whose fallback is "Serbian" let's call it "Custom Serbian", in this custom they can edit, or add translations from the base pre-shipped "Serbian" dictionary.

I am not sure using `Language` instead of `dictionary` better convey the idea that they can have fallback.

pavlenex:
Hey @NicolasDorier I gave https://github.com/btcpayserver/btcpayserver/pull/5662/commits/2dd5bcea3b22dd672153ed2d80a84828754a016e another test, the improvements are really great and it's way more clearer how to use translations now. 
I like how you added explanation text. I think at this point @dstrukt can review it, to ensure things we've added are consistent and standardized. For example, text boxes you added are great, but we don't have them on other places, but tbhI prefer them like  this (would be good if I could close the explanation at some point).  Same goes for actions on the side.

I think we should somehow ensure in UX that user understands that there's only one translation that can be active, and this is server admin's choice.  Still not sold on `dictionary` but with these UX improvements I don't think it's a blocker. 

One thing still, I am missing a lot of strings, dashboard, sidebar, etc, do you have any estimate on when we can have most of the strings, should that be the goal or this or a follow up PR?

pavlenex:
So my summary thus far is that biggest blocker is a lot of missing strings, if we can get a clarity on when and how these should be added - that would be great. UX wise, I must admit i was sceptical when I heard how you intend to approach this, but now I really like it suo easy to dump it into chatgpt and get a new language, no overhead!
One suggestion for the the future would be to allow user eventually to manually select a language that admin offers for their instance. For example, in Switzerland they may want to offer French and Italian, based on merchant's preference.

So tACK on my end, and now up to @dstrukt to review consistency with design system, review copy and perhaps provide another pair of eyes for this great feature!

NicolasDorier:
Having most of the strings will be an iterative process.
But I plan to do both:
1. Added more strings before 2.0
2. Adding pre-shipped translations for several languages

> Still not sold on dictionary but with these UX improvements I don't think it's a blocker.

If the term is not good it's better we fix it now, as fixing it later is harder.

dstrukt (changes requested):
Overall, this is really fantastic, it works as expected (for the words that can be translated at least)! It's very light, customizable, and is quite intuitive. Functional tACK - great addition to BTCPay!

Might add 1-2 more points to this list over weekend as I continue to think it over, but...

## Nits:

I can also help out with some of these small formatting changes.

- [ ] We should change the action from `Copy` to `Duplicate` or `Clone` cc @pavlenex, thoughts? 

- [x] `Dictionary Name` to just `Dictionary` in table header

- [ ] Kill the black section, and just add ```Dictionaries enable you to translate the BTCPay Server backend into different languages.``` as the description text under the main header.

Screen Shot 2024-07-19 at 9 42 22 PM

- [ ] We should also have a clearer indicator of the Currently Selected, outside of Bold text-weight, need to think on this - first thought is another column `In Use` with either a green checkmark, or a red X if not - but there might be a better implementation - the current CTA to select which Dictionary to use would work with this

- [ ] If we do the above, the `Edit` CTA can just be the title of the Dictionary (this is a consistent pattern we do with other views

- [ ] The `Delete` CTA should be our red, it's too similar to the other functions next to it

- [ ] Not sure the `Source` column is necessary, think we could remove to reduce noise, as it's either Custom or Default
Screen Shot 2024-07-19 at 9 46 37 PM

Suggestions:
- [ ] Can we color co-ordinate the key/pair values like a JSON `code` block? For example. I realize we might not be able to do this if it's not proper code/syntax, or at least as far as I'm aware we can't... but it's something to consider, as the text is currently a little hard to read and could be formatted better.

Screen Shot 2024-07-19 at 9 30 59 PM

- [ ] Alternatively, if we can't color code the text, we could also bump the font-size / line-height up slightly to make it a little easier to read a long list of very similar text, i.e: (there's probably more we can do here as well, but i need to tinker in the editor a bit more)

Screen Shot 2024-07-19 at 9 36 04 PM

- [ ] I do like all of the individual names `Translations` and `Dictionaries` and think they logically make sense, but it is a bit strange to click on the sidebar for `Translations` and then see with a `Dictionary` title. I do think we *could* get away with just calling them all `Translations` in the respective places, but I want to think on this a bit more... I'm curious to hear how others feel about the discrepancy?

NicolasDorier:
> The Delete CTA should be our red, it's too similar to the other functions next to it

We aren't using red for this anywhere in the UI. It doesn't seem consistent. I changed to `Remove`.

NicolasDorier:
What about something like that:

image

dstrukt:
> We aren't using red for this anywhere in the UI. It doesn't seem consistent. I changed to `Remove`.

Ahh you're right. I suppose I was thinking of @dennisreimann's LNBank plugin, but you're right - we're not using different color CTAs for Remove/Delete. I guess that change really stuck in my memory?

> > We should also have a clearer indicator of the Currently Selected, outside of Bold text-weight, need to think on this - first thought is another column In Use with either a green checkmark, or a red X if not - but there might be a better implementation - the current CTA to select which Dictionary to use would work with this
> 
> I only have put in bold, I don't know having a checkbox for selecting a single item of a list seems a bit weird to me.

Sorry, I should have been more specific here. To match what we have done in this instance, you're using (checkmark) or not (x). However, this was a minor, and easy aesthetic change if we decide to go that way in the future. There may be a better solution, but this was my first thought upon reviewing, and no need to address right at this moment.

Screen Shot 2024-07-23 at 9 16 38 PM

> > Can we color co-ordinate the key/pair values like a JSON code block?
> 
> Not possible. This isn't a code block, but a textarea. Having an editor here is a bit above my pay grade, it would require quite a lot of effort.
> 
> > Alternatively, if we can't color code the text, we could also bump the font-size / line-height up
> 
> I tried on my computer, but it looked worse.
> 
> IMHO we shouldn't put too much effort into this, I expect people will just copy/paste ChatGPT the hell out of it. Or maybe, just use a different editor.
> 

No worries, and you're right, and I agree wrt not putting too much effort at this stage - I'm not 100% aware of the potential limitations wrt textareas / code block on the front-end, so I posed a few suggestions in hopes it *could* be an easy change, or that maybe @dennisreimann could work some of his magic. But for getting this first iteration out, I completely agree, let's get it shipped and iterate!

> Maybe what we could do though, is removing the text-area, replacing it with a code block with JSON formatting, like you suggested, then only allow the user to download/upload the translations rather than edit in the browser. What do you think?

Mmm, that's an interesting idea. I suppose it depends how users intend to use the tool. I could see some preferring to edit in a desktop editor, and uploading - but I could also easily see some users preferring direct browser editing (say on mobile or just smaller/minor edits).

I'd say, let's go with what you've done (editor in browser), as it's very easy to copy/paste for chatGPT, etc.. and it works for 100% of users with little effort, and if @dennisreimann and I do any follow up changes, we can explore the upload idea further!

dstrukt (approved):
All that said, tACK - LGTM!

This is an excellent addition, and minor changes and suggestions mentioned above shouldn't delay getting this shipped.

dstrukt:
For anyone looking, current state:
Screen Shot 2024-07-23 at 9 41 51 PM, cur
Screen Shot 2024-07-23 at 9 42 04 PM
Screen Shot 2024-07-23 at 9 41 37 PM

pavlenex:
uTACK based on convo and screenshots. UX wise seems ready to go, with minor improvements we can do with @dennisreimann and @dstrukt independently as already mentioned by @dstrukt, excited for this one LFG 🚀

PR #5673, opened 2024-01-18
Add plugins test

dennisreimann:
WIP: Right now this isn't fully working, because the incompatible version is loaded from the assemblies, even after it is deleted from the plugins folder.


BTCPayServer.Tests/PluginTests.cs
```diff
@@ -0,0 +1,85 @@
...
+            Assert.Contains("Disabled Plugins", s.Driver.PageSource);
+            Assert.DoesNotContain("Installed Plugins", s.Driver.PageSource);
+
+            var disabledPlugins = s.Driver.FindElements(By.CssSelector("#DisabledPlugins li"));
+            Assert.Single(disabledPlugins);
+            disabledPlugins.First().FindElement(By.CssSelector(".uninstall-plugin")).Click();
+            Assert.Contains("Plugin scheduled to be uninstalled", s.FindAlertMessage().Text);
+
+            // Restart
+            await RestartAfterInstall(s);
```
dennisreimann, line 57: This should restart cleanly, because the plugin gets uninstalled before restart. Yet it fails, because the plugin now gets added in `LoadPluginsFromAssemblies`. I don't know why that is, especially as this isn't the case for previous app starts.

lethiaGa (approved):

PR #5683, opened 2024-01-20
Create RELEASE-CYCLES.md

pavlenex:
This PR adds documentation around release cycles in BTCPay and tends to outline processes and ensures there's documented structure on roles and responsibilities. Feedback welcome.

Initially I wanted to put it into `docs` folder, but then I noticed we have `RELEASE-CHECKLIST.MD` so I've just created the `.md` there.

### DIFF
--- a/RELEASE-CYCLES.md
+++ b/RELEASE-CYCLES.md
@@ -1,0 +1,33 @@
+# Release Cycle Documentation
+
+## Introduction: 
+This document outlines the release cycle process for BTCPay Server, detailing the types of releases and the procedures associated with each.
+
+## Release Types: 
+BTCPay Server categorizes its releases into three main types:
+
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):

dstrukt (approved):
I like the standardization of this process - LGTM!

NicolasDorier (approved):

NicolasDorier:
Btw, I think I should document the release process as well if somebody need to do it.

If I put that later, do you think it should be in the same document or different document?

pavlenex:
@NicolasDorier We already have release-checklist.md is that what you are referring, or is it something else?

NicolasDorier:
doh, forgot about it! Yes it is this one, I'll need to elaborate a bit.

PR #5686, opened 2024-01-22
Webhook tests + FIXES + DOCS

Kukks:
Users were reporting issues with webhooks on telegram, so I wrote tests. Turns out there were none of the issues they mentioned (they later reported it was user errors), BUT I did find a number of bugs in that
* There was no docs around the new webhooks
* Updated Payout webhooks would not always be executed 
* The C# client would fails to serialize the webook data


dennisreimann (approved):
LGTM. What about the TODOs?

NicolasDorier:
@Kukks I added a commit as to prevent modification or funny business with StoreId and FormResponse. Please review and merge.

Kukks:
> LGTM. What about the TODOs?

Let's do it in a separate PR

PR #5689, opened 2024-01-23
Guest role: Fix redirect after store creation

### LINKED ISSUE #5688: [Bug]: Creating a Store as a Guest generates a 403 error
Not 100% if this supposed to be the intended behavior, but it seems like a bug...

Running:
BTCPay Server v1.11.2 on Chrome Desktop

Repro:

1. Set Guest role as default
2. Create new user (will be a Guest)
3. Login as new user
4. Attempt to create a Store
5. Get met with a 403 error
6. Still create the Store after `Navigating to home`

dennisreimann:
This ensures that guests land on the invoices list, which tehy are allowed to see — rather than the dashboard, which they don't have permissions for.

Fixes #5688.

### DIFF
--- a/BTCPayServer/Components/StoreSelector/Default.cshtml
+++ b/BTCPayServer/Components/StoreSelector/Default.cshtml
@@ -60,7 +60,7 @@ else
                 {
                     <li><hr class="dropdown-divider"></li>
                 }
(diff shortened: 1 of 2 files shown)


dstrukt (approved):
Resolves the error. 

tACK!

Kukks (approved):

PR #5694, opened 2024-01-23
Wallet file parsing: Add Wasabi test case and re-add Electrum distinction

### LINKED ISSUE #5691: [Bug]:  v1.12.4-RC Importing a wallet file from Wasabi defaults to legacy wallet

 v1.12.4 RC

https://mainnet.demo.btcpayserver.org

### What happened?

Importing a wallet file from Wasabi defaults to legacy, even though the wallet is SegWit native.

1. Create a brand new wallet in Wasabi
2. Import a wallet file in BTCPay
3. Check addresses that are legacy

### Relevant log output

_No response_

### What browser do you use?

RC IMPORT 

https://github.com/btcpayserver/btcpayserver/assets/36959754/8a4a72c8-8ede-4f8d-aab8-0181883ef1b9

NON RC IMPORT
Screenshot 2024-01-23 at 12 46 36

### Additional information

_No response_

dennisreimann:
I had to re-add the Electrum distinction to make it work, otherwise it would fail for Wasabi. Reading [this comment](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Extensions.cs#L81) it might make sense to reintroduce the distinction.

Fixes #5691.


BTCPayServer/DerivationSchemeParser.cs
```diff
@@ -143,7 +143,7 @@ public DerivationStrategyBase Parse(string str, bool ignorePrefix = false, bool
                         throw new FormatException(
                             $"Invalid xpub. Is this really for {BtcPayNetwork.CryptoCode} {Network.ChainName}?");
 
-                    if (!ignorePrefix && !hasLabel && BtcPayNetwork.ElectrumMapping.TryGetValue(prefix, out var type))
+                    if (!ignorePrefix && !hasLabel && electrum && BtcPayNetwork.ElectrumMapping.TryGetValue(prefix, out var type))
```
dennisreimann, line 146: This is the actual fix, which skips the default legacy assignment.

BTCPayServer/Extensions.cs
```diff
@@ -50,7 +48,7 @@ public static DerivationSchemeParser GetDerivationSchemeParser(this BTCPayNetwor
         }
 
         public static bool TryParseXpub(this DerivationSchemeParser derivationSchemeParser, string xpub,
-            ref DerivationSchemeSettings derivationSchemeSettings, out string error)
+            ref DerivationSchemeSettings derivationSchemeSettings, out string error, bool electrum = true)
```
dennisreimann, line 51: Keep Electrum format as default, which has been like that previously as well.

BTCPayServer/Services/WalletFileParsing/WasabiWalletFileParser.cs
```diff
@@ -30,7 +30,7 @@ public bool TryParse(BTCPayNetwork network, string data, [MaybeNullWhen(false)]
             Network = network
         };
 
-        if (jobj is null || !derivationSchemeParser.TryParseXpub(jobj.ExtPubKey, ref result, out var error))
+        if (jobj is null || !derivationSchemeParser.TryParseXpub(jobj.ExtPubKey, ref result, out var error, false))
```
dennisreimann, line 33: Set Electrum to `false` for the Wasabi import.

PR #5700, opened 2024-01-25
In Wallet Send, label were not applied to transactions

NicolasDorier:
If you send BTC to an address (not belonging to the wallet) and add a label to it, the outgoing transaction wouldn't be labelled.


dennisreimann (approved):
tACK.

Issue #5699, opened 2024-01-24
Implement Bitnob rate provider to expand support for African currencies


rockstardev:
This should be pretty straightforward, since all it's needed is to implement new IRateProvider to parse the data from `https://api.bitnob.co/api/v1/rates/bitcoin/price/`

PR #5705, opened 2024-01-25
Bitnob rate provider

TChukwuleta:
#5699 

Implement Bitnob rate provider among other rate providers to expand support for African currencies

### DIFF
--- a/BTCPayServer.Rating/Providers/BitnobRateProvider.cs
+++ b/BTCPayServer.Rating/Providers/BitnobRateProvider.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


NicolasDorier:
@TChukwuleta is there any currency where this exchange should be recommended? If so, you should also add it in `StoreBlob.RecommendedExchanges`.

TChukwuleta:
Hey @NicolasDorier  


I can't say for the others, I have only done interaction with NGN, and even with that, BTC is usually benchmarked against USD and not NGN

NicolasDorier:
@TChukwuleta so you don't think this should be the recommended exchange for NGN?

TChukwuleta:
I probably misunderstood the question. Yea.. it should be

I'd add it to the recommendaed exchange and update the PR

TChukwuleta:
@NicolasDorier added Bitnob as recommended exchange for NGN

Issue #5708, opened 2024-01-26
Adjust precision and remove decimals for Colombian `COP` and Argentina's Peso `ARS`


pavlenex:
Feedback from telegram:

> Is there a way to not show decimal places (cents) in a PoS when creating an invoice?

> cents disappeared from our national shitcoin (aka colombian peso) a long time ago... and it will disappear from yours eventually lol

![photo_2024-01-26_14-03-33](https://github.com/btcpayserver/btcpayserver/assets/36959754/1916b384-495b-4db5-983a-c2f0f4dcd9bb)

pavlenex:
@TChukwuleta This may be something good to tackle if you're up for it.

Kehinde13:
Hello, I would like to work on this 
Here's my email balogunkehinde3@gmail.com

TChukwuleta:
Hey @Kehinde13 

If you are up for this, sure. Take a go at it

manasadepu:
Is this still available?

TChukwuleta:
Not anymore @manasadepu 

You can go through the issues and see if you would find anyone that you can tackle

PR #5710, opened 2024-01-29
remove decimals for Colombian (COP) and Argentina's Peso (ARS)

TChukwuleta:
Fix for issue #5708 

Remove decimal places for COP and ARS currencies

### DIFF
--- a/BTCPayServer.Rating/Currencies.json
+++ b/BTCPayServer.Rating/Currencies.json
@@ -1,4 +1,4 @@
-﻿[
+[
    {
       "name":"Afghan Afghani",
       "code":"AFN",
@@ -58,7 +58,7 @@
    {
       "name":"Argentine Peso",
       "code":"ARS",
-      "divisibility":2,
+      "divisibility":0,
       "symbol":null,
       "crypto":false
    },
@@ -289,7 +289,7 @@
    {
       "name":"Colombian Peso",
       "code":"COP",
-      "divisibility":2,
+      "divisibility":0,
       "symbol":null,
       "crypto":false
    },
--- a/BTCPayServer.Rating/CurrencyNameTable.cs
+++ b/BTCPayServer.Rating/CurrencyNameTable.cs
@@ -77,19 +77,15 @@ namespace BTCPayServer.Services.Rates
                             continue;
                         try
                         {
-                            if (new RegionInfo(culture.LCID).ISOCurrencySymbol.Equals("ARS") || new RegionInfo(culture.LCID).ISOCurrencySymbol.Equals("COP"))
+                            var symbol = new RegionInfo(culture.LCID).ISOCurrencySymbol;
+                            var c = symbol switch
                             {
-                                var modifiedCulture = new CultureInfo(culture.Name);
-                                NumberFormatInfo modifiedNumberFormat = (NumberFormatInfo)modifiedCulture.NumberFormat.Clone();
-                                modifiedNumberFormat.CurrencyDecimalDigits = 0;
-                                modifiedCulture.NumberFormat = modifiedNumberFormat;
-
-                                _CurrencyProviders.TryAdd(new RegionInfo(culture.LCID).ISOCurrencySymbol, modifiedCulture);
-                            }
-                            else
-                            {
-                                _CurrencyProviders.TryAdd(new RegionInfo(culture.LCID).ISOCurrencySymbol, culture);
-                            }
+                                // ARS and COP are officially 2 digits, but due to depreciation,
+                                // nobody really use those anymore. (See https://github.com/btcpayserver/btcpayserver/issues/5708)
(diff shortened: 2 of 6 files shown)


BTCPayServer/wwwroot/pos/common.js
```diff
@@ -93,10 +93,16 @@ const posCommon = {
         formatCurrency (value, withSymbol) {
             const currency = this.currencyCode
             if (currency === 'BTC' || currency === 'SATS') return this.formatCrypto(value, withSymbol)
-            const { divisibility } = this.currencyInfo
+
+            // Check if the currency is COP or ARS (exclude decimal places)
+            const excludeDecimalPlaces = ['COP', 'ARS'].includes(currency);
```
Kukks, line 98: This should not be hardcoded in the js but rather decides based on divisibility, which is set in a currency service in c#.

Kukks (changes requested):

BTCPayServer/wwwroot/pos/common.js
TChukwuleta: Noted @Kukks. Would look into it

matman7x (approved):

NicolasDorier:
why does `numberFormatInfo.CurrencyDecimalDigits` doesn't returns 0 for those two currencies?

TChukwuleta:
@NicolasDorier 

numberFormatInfo (a method under system.Globalization) has default value for CurrencyDecimalDigits to be 2

I am not sure if  or how it is tied to currencies

NicolasDorier:
@TChukwuleta , please instead of doing this hack here, add those currencies to https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Rating/Currencies.json

Then extend the test `RoundupCurrenciesCorrectly` to make sure COP and ARS have the right `CurrencyDecimalDigits`

BTCPayServer.Rating/CurrencyNameTable.cs
```diff
@@ -77,7 +77,19 @@ public IFormatProvider GetCurrencyProvider(string currency)
                             continue;
                         try
                         {
-                            _CurrencyProviders.TryAdd(new RegionInfo(culture.LCID).ISOCurrencySymbol, culture);
+                            if (new RegionInfo(culture.LCID).ISOCurrencySymbol.Equals("ARS") || new RegionInfo(culture.LCID).ISOCurrencySymbol.Equals("COP"))
```
NicolasDorier, line 80: Remove this, you shouldn't have anything hard coded.

BTCPayServer.Rating/Currencies.json
```diff
@@ -58,7 +58,7 @@
    {
       "name":"Argentine Peso",
       "code":"ARS",
-      "divisibility":2,
```
NicolasDorier, line 61: https://en.wikipedia.org/wiki/Argentine_peso it says there is 2 subunit.

BTCPayServer.Rating/Currencies.json
```diff
@@ -289,7 +289,7 @@
    {
       "name":"Colombian Peso",
       "code":"COP",
-      "divisibility":2,
```
NicolasDorier, line 292: https://en.wikipedia.org/wiki/Colombian_peso it says there is 2 subunit

BTCPayServer.Tests/FastTests.cs
```diff
@@ -790,7 +790,7 @@ public void RoundupCurrenciesCorrectly()
                 (0.0005m, "0.0005 USD", "USD"), (0.001m, "0.001 USD", "USD"), (0.01m, "0.01 USD", "USD"),
                 (0.1m, "0.10 USD", "USD"), (0.1m, "0,10 EUR", "EUR"), (1000m, "1,000 JPY", "JPY"),
                 (1000.0001m, "1,000.00 INR", "INR"),
-                (0.0m, "0.00 USD", "USD")
```
NicolasDorier, line 793: Note that you should add a test on `CurrencyNameTable.Instance.GetNumberFormatInfo` after as well.

`GetNumberFormatInfo` is using the json file, but the formatting come from CultureInfo from the OS.

The OS says ARS and COP have 2 decimals, and wikipedia seems to agree, so I am not sure why it should be changed.

NicolasDorier:
Please just add a test on the `GetNumberFormatInfo` https://github.com/btcpayserver/btcpayserver/pull/5710#discussion_r1475590651 in the same test method (below after the loop).

BTCPayServer.Tests/FastTests.cs
TChukwuleta: Well according to the issue, https://github.com/btcpayserver/btcpayserver/issues/5708#issue-2102859157, they no longer run cents

Issue #5563, opened 2023-12-18
New user onboarding flow


dennisreimann:
For the new user onboarding flow we assume there is an instance to connect to already. 

## User Types

There are at least three roles/use cases for the flow of onboarding/initial login:

1. Admin
    - Their account got created when setting up the instance.
    - They could either set up merchant accounts (and stores) in advance or simply offer a registration.
    - Related: #5548
2. Merchant
    - Their account might have been created by the admin or they get to the instance for registering.
    - Manage the store and access roles for their employees.
    - They provide employee accounts by setting them up in advance or let their employees register.
3. Employee
    - Roles like cashier, with limited store access.

## Onboarding Approaches

1. Open Instance
    - Users register themselves via **shared register link**
    - Admin can require email confirmation and approval
2. Invitation based
    - Admin or store owner creates basic user account with email address
    - User gets an **email with invite link** request to setup the account (confirm email, set password)
        - "You have been invited to INSTANCE/STORE, setup your account here: LINK/QRCODE"

In both scenarios the admin has the option to deactivate user later on.
 
## Invite links

Store Owners and Employees can be provided with a link/qrcode, which is basically an encoded version of ...

- The instance URL to connect to
- An access token we can use on the instance to initially identify them and their setup state, which can either be:
    - Pre-setup by the supervisor role (Admin -> Merchant, Merchant -> Employee)
    - Registration and assignment to the defined role for new users, no prior setup

This is very similar to the flow of connecting a LNDhub-compatible wallet: A custom link scheme which can be handled by our app plus the connection details. Something like this, which can be shared as a link or QR code: 

```
https://SERVER_URL/invite/USER_ID/TOKEN
```

The app can then transfer this into an API call to `https://SERVER_URL/api/app//USER_ID/TOKEN`, which handles the setup and responds with the account details for the app. 

## TODO
- #5714
  - [x] Separate self-registration and invite cases
  - [x] Add invitation email for users created by the admin (right now there is no email when an account is setup with a password)
  - [x] Add handler action for invite links
- #5719
  - [x] Add user creation/invite on the store level (right now one can only add existing users to stores)
  - [x] Improve default role structure

### Optional and related
- #5718 — can be incorporated if/when merged.
  - [x] Make instance name configurable to allow for better branding
  - [x] Add contact email address for instance (see #4806)

- [ ] Integrate everything
- [ ] Wording adjustments, e.g. in emails

pavlenex:
[...]

PR #5714, opened 2024-01-31
Onboarding: Invite new users

dennisreimann:
First part of #5563:

- Separates the user self-registration and invite cases
- Adds invitation email for users created by the admin
- Adds invitation tokens to verify user was invited
- Adds handler action for invite links
- Refactors `UserEventHostedService`


BTCPayServer/Events/UserRegisteredEvent.cs
```diff
@@ -2,14 +2,18 @@
...
+public class UserRegisteredEvent
+{
+    public ApplicationUser User { get; set; }
+    public bool Admin { get; set; }
+    public Uri RequestUri { get; set; }
+
+    public TaskCompletionSource<Uri> CallbackUrlGenerated;
+}
+
+public class UserInvitedEvent : UserRegisteredEvent
```
dennisreimann, line 16: This was added to distinguish self-registration vs. invite and so that we can also at least log, who invited the new user.

BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
```diff
@@ -233,7 +234,9 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
                     await _settingsRepository.FirstAdminRegistered(policies, _options.UpdateUrl != null, _options.DisableRegistration, Logs);
                 }
             }
-            _eventAggregator.Publish(new UserRegisteredEvent() { RequestUri = Request.GetAbsoluteRootUri(), User = user, Admin = request.IsAdministrator is true });
+
+            var currentUser = await _userManager.GetUserAsync(User);
+            _eventAggregator.Publish(new UserInvitedEvent { RequestUri = Request.GetAbsoluteRootUri(), User = user, Admin = isNewAdmin, InvitedByUser = currentUser });
```
dennisreimann, line 239: I opted to change this to invite, because the user does not self-register and no password is set. It will trigger an email and the user can setup (email confirmation, set password) via the new invite handler action.

BTCPayServer/Controllers/UIAccountController.cs
```diff
@@ -784,6 +783,33 @@ public async Task<IActionResult> SetPassword(SetPasswordViewModel model)
             return View(model);
         }
 
+        [AllowAnonymous]
+        [HttpGet("/invite/{userId}/{code}")]
+        public async Task<IActionResult> AcceptInvite(string userId, string code)
```
dennisreimann, line 788: One action to handle different cases (email confirmation, set password) uniformly.

BTCPayServer/Extensions/EmailSenderExtensions.cs
```diff
@@ -29,11 +29,17 @@ public static void SendApprovalConfirmation(this IEmailSender emailSender, Mailb
                 $"Your account has been approved and you can now <a href='{HtmlEncoder.Default.Encode(link)}'>login here</a>");
         }
 
-        public static void SendSetPasswordConfirmation(this IEmailSender emailSender, MailboxAddress address, string link, bool newPassword)
+        public static void SendSetPassword(this IEmailSender emailSender, MailboxAddress address, string link, bool hasPassword)
```
dennisreimann, line 32: Changed the name and argument to be a bit clearer.

dennisreimann:
Will add tests, but wanted to publish the current state so that we can test and discuss the open points from #5563, @dstrukt @pavlenex.

We could also split the issue between two PRs and the user invite on store-level could be tackled in a follow-up.

BTCPayServer/Extensions/EmailSenderExtensions.cs
```diff
@@ -19,21 +19,26 @@ private static string CallToAction(string actionName, string actionLink)
...
 
         public static void SendApprovalConfirmation(this IEmailSender emailSender, MailboxAddress address, string link)
         {
-            emailSender.SendEmail(address, "Your account has been approved",
+            emailSender.SendEmail(address, "BTCPay Server: Your account has been approved",
                 $"Your account has been approved and you can now <a href='{HtmlEncoder.Default.Encode(link)}'>login here</a>");
         }
 
-        public static void SendSetPasswordConfirmation(this IEmailSender emailSender, MailboxAddress address, string link, bool newPassword)
+        public static void SendSetPassword(this IEmailSender emailSender, MailboxAddress address, string link)
```
dennisreimann, line 32: This isn't used for initial password setting anymore, because the invitation action takes care of that.

dennisreimann:
I think it makes sense to keep the scope of this PR to what it is right now. I can do the role updates and invite on the store level in another PR building on this. 

So feel free to review @pavlenex and @dstrukt. Main points from the user/admin perspective:

- Separation of user self-registration and invite cases
- Invitation email for users created by the admin
- Accounts created by an admin are now auto-approved
- Admins receive an email if a new account requires approval

pavlenex:
cACK. 

Is there anything specific that needs to be user-tested here? In general what is proposed sounds good to me.

pavlenex:
Tested:

- Invitation email for users created by the admin  ✅ 
- Accounts created by an admin are now auto-approved  ✅ 
- Admins receive an email if a new account requires approval 🚧 (still testing)
   - @dennisreimann do all admins receive email? 

### Bugs encountered 🐛 
#### When changing user to admin role or removing and admin role, we spit out an error,`Error updating the user` but the changes still apply successfully. See video. May be unrelated to this PR.
https://github.com/btcpayserver/btcpayserver/assets/36959754/74dd4407-5155-450e-a09b-759e9284127e

dennisreimann:
> do all admins receive email once user is pending approval?

Yes. All admins should receive th email and notification about the new user requiring approval.

> BUG 1 When changing user to admin role or removing and admin role, we spit out an error,Error updating the user but the changes still apply successfully.

Fixed.

> BUG 2 When email verification is enabled, and user is added manually by admin, email verification email is never sent, but user receives account creation email.

This is now also fixed. There is no separate "Confirm your email address" email, as we can make this part of the process when accepting the invitation — so everything is handled by the invitation email the user got. However, the problem was that the email confirmation in that action did not work. Now it does — thanks for spotting this error! :)

dennisreimann:
Updated as discussed with @pavlenex: Send notification to admins only if the user does ...

- either not require email confirmation
- or when they confirmed their email address
 
Rationale: We want to inform admins only about qualified users and not annoy them with bot registrations in case email confirmation is required.

pavlenex:
@dennisreimann Found one minor thing. These are my server settings
Screenshot 2024-02-09 at 12 35 22

I invite the user and Admin gets notification that user is waiting approval, but I think all users invited by admin are approved, so this is probabaly notification code remaining from one of previous changes, can you check please? Notification says user waiting approval, but the user is already approved basically when you go to that user.

dstrukt (approved):
Finally got this working on my local machine, basically nuked and started fresh.

Tested the 3 cases @pavlenex mentioned previously that were supposed fixed - looks like they work properly for me.

I'm sure there's more testing and edge cases that can be done, but maybe kukks or nico will find something!

BTCPayServer/Events/UserRegisteredEvent.cs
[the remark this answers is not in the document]
dennisreimann: Imho this variant is more explicit and feels better than cluttering up the more general event with a flag plus extra property. I've [added the subclass to the list of events that are subscribed](https://github.com/btcpayserver/btcpayserver/pull/5714/files#diff-28cf1ff2a014e577e74f91b7437d1e360b2968aafe2e835941a313c1e5e9762cR31) as well.

BTCPayServer/Security/IdentityBuilderExtension.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.AspNetCore.Identity;
+
+namespace BTCPayServer.Security;
+
+public static class IdentityBuilderExtension
+{
+    public static IdentityBuilder AddInvitationTokenProvider(this IdentityBuilder builder)
+    {
+        var userType = builder.UserType;
```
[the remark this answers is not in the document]
dennisreimann: Done in 3fe482a.

BTCPayServer/Events/UserRegisteredEvent.cs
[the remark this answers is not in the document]
dennisreimann: Updated in a823aff.

PR #5717, opened 2024-02-01
Store: Move support URL to Checkout Appearance and improve wording

dennisreimann:
As discussed in the recent design meeting.

supporturl

PR #4258, opened 2022-11-01
Add option to customize the instance logo

dennisreimann:
Custom logo for BTCPay instances, plus some view simplifications.

settings

login

hrf

login-spiral

PR #5718, opened 2024-02-01
Server Settings: Customize instance name and add contact URL

dennisreimann:
- The custom instance name would improve #5563
- Added contact URL closes #4806

## Server Settings

The two settings are added to the Server Settings > Maintenance page.

serversettings

## Public pages

If set, the custom instance name appears in the title of the logged out views. The "Contact Us" link is displayed in a subtle way below the main box on the logged out views.

public

### DIFF
--- a/BTCPayServer/Views/UIHome/Home.cshtml
+++ b/BTCPayServer/Views/UIHome/Home.cshtml
@@ -1,11 +1,15 @@
+@using BTCPayServer.Services
+@using Microsoft.AspNetCore.Mvc.TagHelpers
 @model HomeViewModel;
+@inject SettingsRepository SettingsRepository
 @{
-    ViewData["Title"] = "BTCPay Server";
+    var settings = await SettingsRepository.GetSettingAsync<ServerSettings>() ?? new ServerSettings();
+    ViewData["Title"] = string.IsNullOrWhiteSpace(settings.ServerName) ? "BTCPay Server" : settings.ServerName;
 }
 
 <partial name="_StatusMessage" />
 
-<h2>Welcome to your BTCPay Server</h2>
+<h2>Welcome to @(string.IsNullOrWhiteSpace(settings.ServerName) ? "your BTCPay\u00a0Server" : settings.ServerName)</h2>
 
 @if (!Model.HasStore)
 {
--- a/BTCPayServer/Components/MainNav/Default.cshtml
+++ b/BTCPayServer/Components/MainNav/Default.cshtml
@@ -252,7 +252,7 @@
     {
         <ul id="mainNavSettings" class="navbar-nav border-top p-3 px-lg-4">
             <li class="nav-item" permission="@Policies.CanModifyServerSettings">
-                <a asp-area="" asp-controller="UIServer" asp-action="ListUsers" class="nav-link @ViewData.IsActivePage(ServerNavPages.Users) @ViewData.IsActivePage(ServerNavPages.Emails) @ViewData.IsActivePage(ServerNavPages.Policies) @ViewData.IsActivePage(ServerNavPages.Services) @ViewData.IsActivePage(ServerNavPages.Theme) @ViewData.IsActivePage(ServerNavPages.Maintenance) @ViewData.IsActivePage(ServerNavPages.Logs) @ViewData.IsActivePage(ServerNavPages.Files)" id="Nav-ServerSettings">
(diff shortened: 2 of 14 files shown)


dstrukt:
AWWW yes! it's all starting to take shape! Will try to get a review in today / see if we catch everywhere.

pavlenex:
Nice, will be useful for the app onboarding.
@dennisreimann can you elaborate a bit on where `contact us` will be displayed, is it only on the login page when user is logged out or somewhere else as well?

dennisreimann:
Right now it's on the logged out pages, so also e.g. forgot password. 

We can add this elswhere to, wanted to keep it simple and discuss these things and whether or not we want to have something like that. I think it's useful and we could add it to the menu as you mentioned. 

I opted against putting it into the footer, because there we have only BTCPay related links and I didn't want to mix this with instance specific info.

pavlenex:
> Right now it's on the logged out pages, so also e.g. forgot password.

So it will be visible on payment request, crowdfunding, checkout or?

dennisreimann:
It's not visible there, because it would be more appropriate to put the store's contact URL there (see #5717) instead of the server one.

pavlenex:
Yeah I want to make sure it's not visible there, because it makes no sense for it to be visible anywhere but the login page imo.

NicolasDorier:
Weird to see that in Maintenance, we don't have a "Branding" page for the server?

dennisreimann:
I agree, but didn't want to add another page. If we agree this is worthwile to add, I could add it here. /cc @dstrukt

pavlenex:
But we do have Server Settings > Theme > Branding

Perhaps for consistency we can rename theme > branding

Screenshot 2024-02-05 at 12 43 40

dennisreimann:
Good idea and it makes for a good fit! Moved the new settings over and renames that page to Branding.

grafik

dennisreimann:
The server logo is less constraint in size, because we do not require a square image.

pavlenex:
Another thing, the copy says `Contact link for support requests related to this BTCPay Server instance. This link will appear on public facing pages. Can be any valid URI, such as a website, email, and Nostr.`

I've added test.com and it redirects to internal link once clicked which leads to 404.

pavlenex:
> The server logo is less constraint in size, because we do not require a square image.

Are we sure we don't want to unify that as well, because when square logo is added it can lead to this?
Screenshot 2024-02-05 at 19 30 08

dennisreimann:
> Are we sure we don't want to unify that as well, because when square logo is added it can lead to this?

Good catch, but I'd rather fix this than constraining the server to use a square logo as well. Imho we should also think about removing that requirement on the store side. /cc @dstrukt

dennisreimann:
> Hmm, we should definitely keep standardized image sizings imo, a 1:1 square for logo marks, admin labels, social icons, etc.. (so we can serve up something for those cases), and perhaps a horizontal logo option that has a constrained height, so a user could upload a 4:1, 3:1, 2:1, etc.. but the height doesn't change / break the view.

I'd vote against constraining it and would rather see the store logo also being more adaptable. For now it is OK to use a square dimension as we introduced it as such, but there's no necessity to impose the same restriction on the server logo.

The approach with defining a max height (and width) works well with all kinds of logos (see #4258) and we could use it for the store as well, especially now that we've made it so that the logo has it's own row and isn't used horizontally next to other items.

dstrukt:
I think we're in alignment wrt the server logo, as long as it's constrained in the various places - I'm happy to make it more flexible! 

It might be easier to discuss a prototype/PR if you want to proceed, then we can nit pick specifics or edge-cases.

dstrukt:
And cACK to the `Branding` name over `Theme` .. still think there's more we can do on the organizational aspect of the Server / Settings in general, but all of the functional pieces are coming together nicely!

dstrukt (approved):
The only thing I'm uncertain of atm is the `Contact Us` CTA. I think it could use some more thought, but it works well and does what it's intended, so let's ship it and iterate later.

tACK - LGTM!

BTCPayServer/Views/UIHome/Home.cshtml
```diff
@@ -1,11 +1,15 @@
...
 @{
-    ViewData["Title"] = "BTCPay Server";
+    var settings = await SettingsRepository.GetSettingAsync<ServerSettings>() ?? new ServerSettings();
+    ViewData["Title"] = string.IsNullOrWhiteSpace(settings.ServerName) ? "BTCPay Server" : settings.ServerName;
 }
 
 <partial name="_StatusMessage" />
 
-<h2>Welcome to your BTCPay Server</h2>
+<h2>Welcome to @(string.IsNullOrWhiteSpace(settings.ServerName) ? "your BTCPay\u00a0Server" : settings.ServerName)</h2>
```
Kukks, line 12: \u00a0? why not a simple space?

Kukks (approved):
code ack apart from the \u00a0 chars instead of a space? Unless there is a good reason.

pavlenex:
@dennisreimann can we speficy that for emails `mailto:` needs to be added, otherwise if user adds email@email.com and someone cliks contact us it crashes.

PR #5731, opened 2024-02-05
Policies: Cleanup and improvements

### LINKED ISSUE #5660: Cleanup in Server Settings > Policy to reduce ambiguity when enabling/disabling certain settings
Currently, we're inconsistent with certain options in Server settings, and user not reading carefully may enable certiain options by ticking the box (which is what should be default behaviour).

Example 
Screenshot 2024-01-15 at 13 34 44

- `Enabling public user registration`
   - Enabling this allows anyone to register to your server and may expose your BTCPay Server instance to potential security risks from unknown users.
 - `Allowing non-admins to access API endpoints`
 -  Enabling this may expose your BTCPay Server instance to potential security risks from unknown users.

Cc @dstrukt @dennisreimann for suggestions on the copy

dennisreimann:
First step towards improving the Server Settings > Policies page:

- [x] Turn settings into toggles
- [x] Move Email and Maintenance settings to their particular page
- [x] Improve wording for settings

Do we need to keep "Disable notifications from automatically showing (no websockets)"? This seems very niche. /cc @kukks @NicolasDorier 

policies

policies-email

policies-maintenance

Closes #5660.


dstrukt:
Looking good! 

Would love to remove that setting if possible, was going to bring it up in this week's call.

dennisreimann:
As discussed with @NicolasDorier, I've removed the "Disable notifications from automatically showing (no websockets)" setting. 

Ready for your reviews and wording adjustments, @pavlenex @dstrukt.

dstrukt (changes requested):
Generally pretty happy with the copy in this view, i'm sure it could be slightly more succinct, but I'm not always the best at that, so here's my attempt. I also think we could try to capitalize our key values (e.g. Store, Hot Wallet, Lightning Node, etc.. I may have gone a little overboard, but you get the idea)

## Registration:

`Enable public user registration`

`Confirmation email required`

`Admin must approve new users`

`Non-admins can access the User Creation API Endpoint`

## Users:

`Non-admins can use the Internal Lightning Node for their Store`

`Non-admins can create Hot Wallets for their Store`

`Non-admins can import Hot Wallets for their Store`

## Maintenance

Don't feel so strongly about this one, but considering we're trying to do the On/Enabled consistent correlation..

- [ ] Enabled by default
`Search engines can index this site`

The `Enable experimental features` could also use some helper text to explain what's actually going on. I've never enabled it, so I don't actually know :joy: - need to do some research this week. cc @pavlenex

## Email

`All Stores use the Server's SMTP email settings as the default` 
With helper text:
`This can be overridden at the Store level.`

Otherwise, functionally seems to work just fine - will do a proper test once we settle on changes!

dennisreimann:
> The Enable experimental features could also use some helper text to explain what's actually going on. 

For now it is just the Custodian Accounts feature, but we are about to remove it anyways with one of the upcoming releases. So this option might go completely as well, but for now I've added that helptext.

wording

maintenance

email

dstrukt (approved):
Copy looks great - could possibly use more functional testing from someone else, maybe I missed something? Everything seems to work as expected on my end.

tACK - LGTM!

pavlenex:
Really good improvement, I much prefer switch to checkboxes and the copy is much clearer now.

One thing I find a bit problematic is the logic behind things being moved to maintenance, is there a reasoning I am missing? I think initial naming there `Maintenance` in Policies,  might be slightly misleading. 

In my view, search engine indexing, is server policy, don't think it has much to do with maintenance of the server, you enforce it once and that's it, and same applies for experimental features imo. 

On an unrelated note, but it's affected by this PR, the maintenance page should have some sort of prioritization in terms of above the fold actions, for example right now with these changes an important and likely often used function `update` is stuck at the bottom of the page, so perhaps we can discuss architecture and priotitizaion of features there.

@dstrukt For experimental features, currently it's only used by custodian accounts, which are planned to be phased out and remvoed from the code-base so explenation there should probably be `Enable experimental features`, but once we remove custodians I am unsure if we have anything else under this flag.

@Kukks do you know if we use anything under `experimental` flag aside custodians?

Kukks:
> > In my view, search engine indexing, is server policy, don't think it has much to do with maintenance of the server, you enforce it once and that's it, and same applies for experimental features imo.
> 
> Ok, shall I move it back? We can also combine this with the prioritization you talked about, I don't mind doing it all in this PR.
> 
> > do you know if we use anything under experimental flag aside custodians?
> 
> Right now it's only that in core and I don't think any plugin uses it. I'd vote for also removing it once we remove the custodian accounts. Until then we should keep it though.
ack, move it back

Kukks (approved):

PR #5735, opened 2024-02-06
UI: Deprecate the custom CSS options

dennisreimann:
As discussed with @pavlenex we are deprecating the custom CSS options for particular entities (payment request, pull payment, POS, crowdfunding) in favor of the store branding approach.

This displays the custom CSS section only if these values are set already.

### DIFF
--- a/BTCPayServer/Views/Shared/Crowdfund/Public/ViewCrowdfund.cshtml
+++ b/BTCPayServer/Views/Shared/Crowdfund/Public/ViewCrowdfund.cshtml
@@ -14,7 +14,7 @@
 	}
 }
 <!DOCTYPE html>
-<html class="h-100" @(Env.IsDeveloping ? " data-devenv" : "")>
+<html class="h-100" @(Env.IsDeveloping ? " data-devenv" : "") id="Crowdfund-@Model.AppId">
 <head>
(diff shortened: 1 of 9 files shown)


dennisreimann:
@dstrukt this is the custom CSS on the level of individual payment requests, pull payments, etc. 

I think it's rather confusing to have this additional option and imho it's simply a leftover from before, when we didn't have the branding options on the store level.

dstrukt:
Oh I see, will get this tested today as well!

dstrukt (approved):
Seems to work as described - tACK!

Worth a quick note in this week's design call.

dennisreimann:
As discussed in the call, I think this is good to be merged.

@pavlenex can you give it a final review?

pavlenex:
> Users actually use this feature. How would someone fine-tune their pos with custom CSS if they want?

Branding in store settings?

dennisreimann:
Yes, the store branding option of providing a CSS file would cover everything holistically and one could even add CSS targeting extra cases. I think going forward it should be clearer how to customize branding aspects and removing this legacy option would help. However, for the users using this feature nothing changes and the inputs remain in place.

Kukks (approved):

PR #5739, opened 2024-02-07
Bumping LND to 0.17.4-beta

rockstardev:
Updating our devenv to use LND 0.17.4

Used it in BTCPayServer.Lightning library and it's good to go. Tests would occasionally fail, because of Eclair, we need to update it in the future:
https://github.com/btcpayserver/BTCPayServer.Lightning/pull/158

### DIFF
--- a/BTCPayServer.Tests/docker-compose.altcoins.yml
+++ b/BTCPayServer.Tests/docker-compose.altcoins.yml
@@ -224,7 +224,7 @@ services:
       - "5432"
 
   merchant_lnd:
-    image: btcpayserver/lnd:v0.17.3-beta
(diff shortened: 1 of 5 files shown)


rockstardev:
Running into problems with tests... checking what is the issue with LND update

https://github.com/lightningnetwork/lnd/issues/8468

rockstardev:
Part of the reason why the test above failed is because when run in isolation there are not enough funds to open lightning channels. And then we do have a problem with flaky lightning channel opening between implementations. For now, in conversations with Nicolas and Kukks we decided to proceed as is, and will address this if it becomes a bigger issue in the future.

PR #5743, opened 2024-02-08
Multisig/watchonly wallet transaction creation flow proof of concept

Kukks:
This PR introduces a new concept, "pending transactions" (working title 💀).  The idea is that if you have a watch only/multsig wallet, if you need to create a transaction, you currently have to
* Wait until you have access to your (or all) private keys and hope you remember all the details of the transaction
* Hand over a psbt to various parties manually to sign

With this PR, you can create a transaction as usual, but instead of signing you "create a pending transaction". A pending transaction is essentially a PSBT that fleshes out what coins to spend and where to spend them, and collects signatures as signers provide them. Once there are enough signatures, you can broadcast.

If the coins being spent by a pending transaction are spent elsewhere before the signatures are fully collected, the pending transaction is "invalidated".
If the transaction was signed out of band, using other means, BTCPay Server will detect it and mark it complete anyway.

With this system, a payout processor can create pending transactions out of payouts awaiting payments, and mark them as in progress. We can make the system send emails when a pending transaction is created to alert signing parties to come online and sign.  If more payouts are added to the awaiting payment list, and the previous pending transaction did not collect any signatures, we can cancel that pending transaction, and create a fresh pending transaction with all the new payouts too, reducing the number of transactions needed.

Essentially, this system enables functionality that was deemed to be only possible for hot wallets.

There is also an optional expiry so that pending transactions can expire (as the fees are determined at the time of the transaction creation, you do not want the pending transaction to be there for too long and avoid the need of another fee bump transaction).

Currently, pending transactions show at the top of the transactions list.


rockstardev:
rebased branch on latest `master`, will take over this PR

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -160,12 +173,33 @@ async Task<bool> RequireDeviceUnlocking()
...
+                                        if (alreadySigned)
+                                        {
+                                            await websocketHelper.Send("{ \"error\": \"already-signed-psbt\"}", cancellationToken);
+                                            continue;
+                                        }
+                                    }
+                                }
+
+                                // we're adding all coins to the PSBT, in case non witness UTXOs are needed
+                                var wallet = _btcPayWalletProvider.GetWallet(cryptoCode);
```
rockstardev, line 190: Emperor @NicolasDorier says we have to do extra checks. @Kukks bringing up that we can send inputs instead of whole PSBT, but it might be can of worms.

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -142,12 +147,20 @@ async Task<bool> RequireDeviceUnlocking()
...
                                 {
-                                    await websocketHelper.Send("{ \"error\": \"wrong-wallet\"}", cancellationToken);
-                                    continue;
+                                    var signing = derivationSettings.GetSigningAccountKeySettings();
+                                    if (signing.GetRootedKeyPath()?.MasterFingerprint != fingerprint)
+                                    {
+                                        await websocketHelper.Send("{ \"error\": \"wrong-wallet\"}", cancellationToken);
+                                        continue;
+                                    }
                                 }
```
rockstardev, line 161: Switch this check to if any of the master fingerprints match the device fingerprint... and if any inputs are signable

rockstardev: 1d7dbb2fd4a4d6a2eed636fcf15514fac5978fd9

BTCPayServer/Controllers/UIVaultController.cs
rockstardev: Since some wallets may explicitly ask for full transaction, adding user setting that can control this behavior
https://github.com/btcpayserver/btcpayserver/pull/5743/commits/587d309df1e9818e08af574511418c8c19272626#diff-fd467f9ca72efdff817c5a3ab8e533078d1757a6fe7a7615808383e0be2b23c1R186

BTCPayServer.Data/Data/PendingTransaction.cs
```diff
@@ -0,0 +1,74 @@
...
+        public string TransactionId { get; set; }
+        public string CryptoCode { get; set; }
+        public string StoreId { get; set; }
+        public StoreData Store { get; set; }
+        public DateTimeOffset? Expiry { get; set; }
+        public PendingTransactionState State { get; set; }
+        public string[] OutpointsUsed { get; set; }
+
+        [NotMapped][Obsolete("Use Blob2 instead")]
+        public byte[] Blob { get; set; }
```
NicolasDorier, line 20: Remove this, there is no point in having this column... this column is present in some table just because I couldn't really migrate it easily.

BTCPayServer.Data/Data/PendingTransaction.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (databaseFacade.IsNpgsql())
+            {
+                builder.Entity<PendingTransaction>()
+                    .Property(o => o.Blob2)
+                    .HasColumnType("JSONB"); 
+                builder.Entity<PendingTransaction>()
+                    .Property(o => o.OutpointsUsed)
+                    .HasColumnType("text[]");
+            }
+            else
```
NicolasDorier, line 44: Can remove this `else`, we don't support anything else than postgres now.

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -12,12 +12,14 @@
...
 using BTCPayServer.Payments.Bitcoin;
 using BTCPayServer.Services.Invoices;
+using BTCPayServer.Services.Wallets;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using NBitcoin;
 using NBXplorer.DerivationStrategy;
 using Newtonsoft.Json.Linq;
+using Org.BouncyCastle.Asn1;
```
NicolasDorier, line 22: Remove this using

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -160,12 +168,37 @@ async Task<bool> RequireDeviceUnlocking()
...
+                                }
+
+                                // sometimes we need to send the full transaction to the device
+                                if (derivationSettings.ForceNonWitnessUtxo)
+                                {
+                                    var wallet = _btcPayWalletProvider.GetWallet(cryptoCode);
+                                    foreach (var input in psbt.Inputs)
+                                    {
+                                        var txid = input.PrevOut.Hash;
+                                        var tx = await wallet.GetTransactionAsync(txid, false, cancellationToken);
```
NicolasDorier, line 192: Do not fetch it if it already exists in the PSBT.

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -160,12 +168,37 @@ async Task<bool> RequireDeviceUnlocking()
...
+                                        if (alreadySigned)
+                                        {
+                                            await websocketHelper.Send("{ \"error\": \"already-signed-psbt\"}", cancellationToken);
+                                            continue;
+                                        }
+                                    }
+                                }
+
+                                // sometimes we need to send the full transaction to the device
+                                if (derivationSettings.ForceNonWitnessUtxo)
```
NicolasDorier, line 186: IMHO you shouldn't fetch that here.
Instead, I suggest you automatically check `Always include non-witness UTXO if available` (`AlwaysIncludeNonWitnessUTXO`) in the WalletSendViewModel if this is `ForceNonwitnessUtxo`. 

By doing this we have a single path with the logic of fetching the previous transactions, and that is handled already by NBXplorer.

BTCPayServer/Controllers/UIVaultController.cs
```diff
@@ -160,12 +168,37 @@ async Task<bool> RequireDeviceUnlocking()
...
+                                        input.NonWitnessUtxo = tx?.Transaction;
+                                    }
                                 }
+
                                 try
                                 {
                                     psbt = await device.SignPSBTAsync(psbt, cancellationToken);
                                 }
-                                catch (Hwi.HwiException)
+                                catch (Hwi.HwiException hwiex)
```
NicolasDorier, line 201: hwiex isn't used

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -949,19 +1020,29 @@ private IActionResult ViewVault(WalletId walletId, WalletPSBTViewModel vm)
...
                 BackUrl = model.BackUrl
             });
         }
 
-        private IActionResult RedirectToWalletPSBTReady(WalletPSBTReadyViewModel vm)
+        private async Task<IActionResult> RedirectToWalletPSBTReady(WalletPSBTReadyViewModel vm)
         {
+            if (vm.SigningContext.PendingTransactionId is not null)
+            {
+                var walletId = WalletId.Parse(this.RouteData.Values["walletId"].ToString());
```
NicolasDorier, line 1038: No hack, this will blow up in our face. Instead callers to `RedirectToWalletPSBTReady` should pass directly the `WalletId` via an argument.

BTCPayServer/DerivationSchemeSettings.cs
```diff
@@ -106,15 +101,24 @@ public AccountKeySettings[] AccountKeySettings
...
         public string Label { get; set; }
-
+        
+        #region MultiSig related settings
+        //public bool IsMultiSig => AccountKeys.Count > 1;
+        public bool IsMultiSigOnServer { get; set; }
+        
+        // some hardware devices like Jade require sending full input transactions if there are multiple inputs
+        // https://github.com/Blockstream/Jade/blob/0d6ce77bf23ef2b5dc43cdae3967b4207e8cad52/main/process/sign_tx.c#L586
+        public bool ForceNonWitnessUtxo { get; set; }
```
NicolasDorier, line 111: Would rename to `DefaultIncludeNonWitnessUTXO`.

BTCPayServer/DerivationSchemeSettings.cs
```diff
@@ -143,4 +147,19 @@ public bool IsFullySetup()
             return AccountKeyPath != null && RootFingerprint is HDFingerprint;
         }
     }
+
+
+    public class DerivationSchemeSettingsFactory
+    {
+        public DerivationSchemeSettings Create(string label, bool isMultiSigOnServer, bool forceNonWitnessUtxo,
```
NicolasDorier, line 154: Unused code.

BTCPayServer/Services/Wallets/BTCPayWalletProvider.cs
```diff
@@ -64,4 +72,224 @@ public IEnumerable<BTCPayWallet> GetWallets()
                 yield return w.Value;
         }
     }
+    
+    
+    public class PendingTransactionService:EventHostedServiceBase
```
NicolasDorier, line 77: Instead derive from `: EventHostedServiceBase, IPeriodicTask`

You can then remove the `while` loop in `CheckForExpiry`. Instead, you will implement the `IPeriodicTask.Do`, which will call `EventHostedServiceBase.PushEvent` to trigger the handling of an `ExpiryEvent` in your `ProcessEvent`.

You then need to use `services.AddScheduledTask<PendingTransactionService>(TimeSpan.FromMinutes(10))` in the DI.

BTCPayServer.Data/Data/PendingTransaction.cs
rockstardev: It can't be removed since IHasBlob interface depends on it. That's why it's `[NotMapped]` so it doesn't go in the database. We need to first remove it from underlying interface.

BTCPayServer/Controllers/UIVaultController.cs
rockstardev: 211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer.Data/Data/PendingTransaction.cs
rockstardev: 211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer.Data/Migrations/20241029163147_AddingPendingTransactionsTable.Designer.cs
```diff
@@ -0,0 +1,1663 @@
...
+                });
+
+            modelBuilder.Entity("BTCPayServer.Data.WebhookData", b =>
+                {
+                    b.Navigation("Deliveries");
+                });
+#pragma warning restore 612, 618
+        }
+    }
+}
```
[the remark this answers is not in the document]
rockstardev: the fact we remove these files is crazy 😂  211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer/Controllers/UIVaultController.cs
rockstardev: 211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer/Controllers/UIVaultController.cs
rockstardev: OK, I have moved it to generation of PSBT, see: https://github.com/btcpayserver/btcpayserver/commit/211b92cf8b0b83f7a0c0d21b9c1acea55867038c#diff-c21a38f4ba0eb228eb1c2c2dfbccaa926cfcb1f770ed116d2bcea18139d84b75R46

BTCPayServer/Controllers/UIVaultController.cs
rockstardev: https://github.com/btcpayserver/btcpayserver/pull/5743#discussion_r1853256003

BTCPayServer/Controllers/UIWalletsController.cs
rockstardev: 211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer/DerivationSchemeSettings.cs
rockstardev: 211b92cf8b0b83f7a0c0d21b9c1acea55867038c

BTCPayServer/DerivationSchemeSettings.cs
rockstardev: d1f080d20ef6c7130974a2c65e1730b699633240

BTCPayServer/Services/Wallets/BTCPayWalletProvider.cs
rockstardev: 88cc4eabf9967051422edd5db414b9040588e620

rockstardev:
@NicolasDorier addressed all the items you've brought up, take a look and let me know how it looks now.

PR #5745, opened 2024-02-09
Admin overview of the stores on the instance

### LINKED ISSUE #5674: Overview of the stores and apps in user list for administrators
This issue is a portion of https://github.com/btcpayserver/btcpayserver/issues/396.

As we agreed on a dev call, to properly tackle #396 we need to tackle it step by step and ask for ambassador's (third-party hosts) feedback in each steps.

#396 summarizes 3 different use-cases

1.  Overview of the stores
2. Reporting aspect, how much users are earning
3. Managing of store, supporting users, removing malicious stuff and so on

The goal of this ticket is to tackle 1. and give admins better overview on who is using their Server. Currently, admins only have overview of users, but not on stores and public facing pages (such as apps) these users are using. 
As our overall goal to give admins more gradual control over who's using their server

The scope of this issue is to (upon @dstrukt mockup on the UI) is to add an expandable list of stores and apps for every user registered on the instance and also their role in a store.  We agreed that list shouldn't be extendable by default but rather admin can expand it on clicking.

**Out of scope**: add ability for admin to manage, preview stores and store report (2. and 3.)

**Blocker:** Waiting for @dstrukt to provide mockup before it can be tackled

dennisreimann:
Closes #5674.

## Users

The user list now features an overview of stores per user as proposed in #5674.

grafik

### DIFF
--- a/BTCPayServer/Components/StoreSelector/Default.cshtml
+++ b/BTCPayServer/Components/StoreSelector/Default.cshtml
@@ -67,8 +67,10 @@ else
                     <li><hr class="dropdown-divider"></li>
                     <li><a asp-controller="UIUserStores" asp-action="ListStores" asp-route-archived="true" class="dropdown-item @ViewData.IsActivePage(StoreNavPages.Index)" id="StoreSelectorArchived">@Model.ArchivedCount Archived Store@(Model.ArchivedCount == 1 ? "" : "s")</a></li>
                 }
+                @*
                 <li permission="@Policies.CanModifyServerSettings"><hr class="dropdown-divider"></li>
                 <li permission="@Policies.CanModifyServerSettings"><a asp-controller="UIServer" asp-action="ListStores" class="dropdown-item @ViewData.IsActivePage(ServerNavPages.Stores)" id="StoreSelectorAdminStores">Admin Store Overview</a></li>
+                *@
             </ul>
         </div>
     </div>
--- a/BTCPayServer/Components/StoreSelector/Default.cshtml
+++ b/BTCPayServer/Components/StoreSelector/Default.cshtml
@@ -2,6 +2,7 @@
 @using BTCPayServer.Client
 @using BTCPayServer.Components.MainLogo
 @using BTCPayServer.Services
+@using BTCPayServer.Views.Server
 @using BTCPayServer.Views.Stores
 @inject BTCPayServerEnvironment Env
 @inject IFileService FileService
@@ -53,7 +54,7 @@ else
                 @foreach (var option in Model.Options)
                 {
                     <li>
-                        <a asp-controller="UIStores" asp-action="Index" asp-route-storeId="@option.Value" class="dropdown-item@(option.Selected ? " active" : "")" id="StoreSelectorMenuItem-@option.Value">@StoreName(option.Text)</a>
+                        <a asp-controller="UIStores" asp-action="Index" asp-route-storeId="@option.Value" class="dropdown-item@(option.Selected && ViewData.IsActivePage(ServerNavPages.Stores) != "active" ? " active" : "")" id="StoreSelectorMenuItem-@option.Value">@StoreName(option.Text)</a>
                     </li>
                 }
                 @if (Model.Options.Any())
@@ -66,6 +67,8 @@ else
                     <li><hr class="dropdown-divider"></li>
                     <li><a asp-controller="UIUserStores" asp-action="ListStores" asp-route-archived="true" class="dropdown-item @ViewData.IsActivePage(StoreNavPages.Index)" id="StoreSelectorArchived">@Model.ArchivedCount Archived Store@(Model.ArchivedCount == 1 ? "" : "s")</a></li>
                 }
+                <li permission="@Policies.CanModifyServerSettings"><hr class="dropdown-divider"></li>
+                <li permission="@Policies.CanModifyServerSettings"><a asp-controller="UIServer" asp-action="ListStores" class="dropdown-item @ViewData.IsActivePage(ServerNavPages.Stores)" id="StoreSelectorAdminStores">Admin Store Overview</a></li>
             </ul>
         </div>
     </div>
--- a/BTCPayServer/Controllers/UIServerController.cs
+++ b/BTCPayServer/Controllers/UIServerController.cs
@@ -20,6 +20,7 @@ using BTCPayServer.Hosting;
 using BTCPayServer.Logging;
 using BTCPayServer.Models;
 using BTCPayServer.Models.ServerViewModels;
+using BTCPayServer.Models.StoreViewModels;
 using BTCPayServer.Payments;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Apps;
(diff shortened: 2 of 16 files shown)


dstrukt:
Awesome, hyped to get builds of these to keep iterating!

dstrukt:
Thinking through this more this week, but immediate thoughts:

## For the Admin Stores overview:

- Search

- Store Logo (if exists)
- Store Creation Date
- Store Website (if exists)
- Store Balance (perhaps?)
- Store Status (Archived / Active)


## Open Questions:

- Do we want to support mass actions on this view (for archiving, removing, etc..)?
- Do we want to support filters (users, roles, status)
- Do we want to push the Archived Stores into this view, just at the bottom of list, or treat it like the Invoice view .. might help consolidate the number of views.

pavlenex:
I recommend that we approach this in two different PR's, because store overview on a user level is tagged for 1.13, and pretty much has consensus on the approach, where as ad admin overview within store still needs discussion and is tagged for 1.14, though it's great to see them in one PR as well, easier to thinker about the UX.

pavlenex:
In private conversation @rockstardev and @Kukks cACk this approach. 

Kukks did point out that if we intend to refine roles further, we'd have to re-work roles to have ability to be set at a server level, since currently we can't do that, as they're on a store level. For example, a potential use-case here is:
1. Server Admin, who can maintain server and plugins but not access store 
2. Server Support who can access stores but not touch Server Settings or plugins
Currently we don't have those abilities, so in initial approach, only Server Owners would have that option to edit stores, but we need to refine it and make it customizable.

Rockstar mentioned that our plan to roll this out in two stages in 1.13 and 1.14 is okay, since the second stage, of giving ability to Store Owners involves some sensitive and perhaps unwanted UI power for store owners, we should announce it in 1.13. notes that Store Owners upgrading to 1.14 will get that ability, just to give users a bit of time to get used to possibly disruptive updata and prepare them.

dennisreimann:
Ok, so the scope of this PR should be offering the overview of stores on the admin's users list only?

Shall there be links to the stores as well? If so we need to define the capabilities of admins when accessing a store that doesn't belong to them. With the current state of this PR they can manage everything like the store owner, but I think as a temporary measure we could come up with something like a "virtual store role", with which we grant them only limited access. /cc @Kukks

pavlenex:
@dennisreimann What's wrong with just giving them `Guest` access in 1.13 and in 1.14 `Owner` access over stores?

dennisreimann:
Nothing wrong, I just wanted to clarify what it should be like in this PR :)

BTCPayServer/Components/StoreSelector/Default.cshtml
```diff
@@ -66,6 +67,10 @@ else
                     <li><hr class="dropdown-divider"></li>
                     <li><a asp-controller="UIUserStores" asp-action="ListStores" asp-route-archived="true" class="dropdown-item @ViewData.IsActivePage(StoreNavPages.Index)" id="StoreSelectorArchived">@Model.ArchivedCount Archived Store@(Model.ArchivedCount == 1 ? "" : "s")</a></li>
                 }
+                @*
+                <li permission="@Policies.CanModifyServerSettings"><hr class="dropdown-divider"></li>
+                <li permission="@Policies.CanModifyServerSettings"><a asp-controller="UIServer" asp-action="ListStores" class="dropdown-item @ViewData.IsActivePage(ServerNavPages.Stores)" id="StoreSelectorAdminStores">Admin Store Overview</a></li>
+                *@
```
dennisreimann, line 73: Hide the link for now, but keep the functionality as we'll need it in the follow-up PR.

BTCPayServer/Controllers/UIServerController.cs
```diff
@@ -122,6 +123,26 @@ TransactionLinkProviders transactionLinkProviders
             _transactionLinkProviders = transactionLinkProviders;
         }
 
+        [HttpGet("server/stores")]
+        public async Task<IActionResult> ListStores()
```
dennisreimann, line 127: This is the action that is not linked to. Let's keep it nevertheless, so that it is present for the follow-up PR and easier to discuss what the series of PRs will bring as a whole.

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -125,7 +126,10 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
 
             if (!string.IsNullOrEmpty(storeId))
             {
-                store = await _storeRepository.FindStore(storeId, userId);
+                var cachedStore = _httpContext.GetStoreData();
+                store = cachedStore?.Id == storeId
+                    ? cachedStore
+                    : await _storeRepository.FindStore(storeId, userId, isAdmin);
```
dennisreimann, line 132: Optimization as this is called multiple times, depending on how many `permission` attributes are used on the page. Do not query the DB on every hit of this method.

BTCPayServer/Security/CookieAuthorizationHandler.cs
```diff
@@ -164,14 +168,15 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
 
                         // cache associated entities if present
-                        if (app != null)
+                        if (app != null && _httpContext.GetAppData()?.Id != app.Id)
                             _httpContext.SetAppData(app);
-                        if (invoice != null)
+                        if (invoice != null && _httpContext.GetInvoiceData()?.Id != invoice.Id)
                             _httpContext.SetInvoiceData(invoice);
-                        if (paymentRequest != null)
+                        if (paymentRequest != null && _httpContext.GetPaymentRequestData()?.Id != paymentRequest.Id)
```
dennisreimann, line 179: No need to set the data if it has already been set. Also see the optimization above.

BTCPayServer/Security/GreenField/GreenFieldAuthorizationHandler.cs
```diff
@@ -95,7 +97,7 @@ public class GreenfieldAuthorizationHandler : AuthorizationHandler<PolicyRequire
                         {
                             if (string.IsNullOrEmpty(userid))
                                 break;
-                            var store = await _storeRepository.FindStore(storeId, userid);
+                            var store = await _storeRepository.FindStore(storeId, userid, isAdmin);
```
dennisreimann, line 100: Provides the same access on the API-level as in the UI: Admins get `Guest` access to foreign stores.

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -45,16 +45,36 @@ public StoreRepository(ApplicationDbContextFactory contextFactory, JsonSerialize
...
+                        {
+                            ApplicationUserId = userId,
+                            StoreDataId = store.Id,
+                            StoreData = store,
+                            StoreRoleId = guestRole.Id,
+                            StoreRole = guestRole
+                        }
+                    ];
+                }
+            }
```
dennisreimann, line 76: Authorization-wise this is the key part: If an admin user isn't part of the store's users list, then they will be granted `Guest` access.

dennisreimann:
Updated according to the comments:

- Admin is granted `Guest` access to foreign stores
- Link to admin stores overview is removed — I kept the code/actions so that it is present for the follow-up PR and easier to discuss what the series of PRs will bring as a whole.

grafik

pavlenex:
cACK, up to @Kukks to give it a look code-wise and @rockstardev if he wants as well. Then we proceed according to plan, 1.13 we announce admins will have owner access and execute in 1.14. 🚀

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -45,16 +45,36 @@ public StoreRepository(ApplicationDbContextFactory contextFactory, JsonSerialize
             return result;
         }
 
-        public async Task<StoreData?> FindStore(string storeId, string userId)
+        public async Task<StoreData?> FindStore(string storeId, string userId, bool isAdmin = false)
```
Kukks, line 48: this doesn't feel great. I propose we split this part out of this PR   to get the foundation merged

Kukks (changes requested):

BTCPayServer/Services/Stores/StoreRepository.cs
Kukks: I want to change the inner workings. Please separate this code into a separate pull request so I can merge this

dennisreimann: Done in f6878b5.

Kukks (approved):

PR #5748, opened 2024-02-12
Update GeneralSettings.cshtml

BitcoinMitchell:
Removed trailing data in `GeneralSettings.cshtml` that was copied/taken from [`swagger.template.stores.json`](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores.json#L414).


dennisreimann (approved):
Thanks for the fix!

Issue #5755, opened 2024-02-16
Provide ability to directly influence Info popup on Wallet Send page


rockstardev:
Right now, the main way to influence this is through bip21 query string that's passed. However, this results in complicated strings that are generated from logic in `UIWalletsController.LoadFromBIP21` method.

![image](https://github.com/btcpayserver/btcpayserver/assets/5191402/526639ae-e94b-4075-aba2-f58c32ea84cd)

It would be much preferable to directly have parameter directly in `UIWalletsController.WalletSend` (possibly `[FromQuery] string? infoMessage =  null`) and allow those using Wallet Send functionality to pass in string they want.

PR #5756, opened 2024-02-16
Providing additional parameter for info message

rockstardev:
Resolves: #5755

Here is how it looks after this slight change

https://github.com/btcpayserver/btcpayserver/assets/5191402/08ac47da-4e91-4e25-99eb-73c32d64ccab

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.cs
@@ -540,7 +540,6 @@ namespace BTCPayServer.Controllers
                 }
                 catch (Exception ex) { model.RateError = ex.Message; }
             }
-            
             return View(model);
         }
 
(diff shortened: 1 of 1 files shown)


pavlenex:
cACK, sweet addition!

dennisreimann:
Code looks good. Only thought or sidenote: As this would allow providing links to the send page with custom messages I'm thinking if there might be ways to abuse this and trick someone.

rockstardev:
@dennisreimann I've used `Message` in `StatusMessageModel` and not `Html`, exactly to prevent "richly formatted" messages. 

Also, you previously could already pass in strings using `message` property of `bip21`... so you could've already used that to abuse/trick people.

dstrukt:
🔥 looks great

Kukks (commented):
Can't you do TempData.SetStatusMessageModel in your own code  before the redirect? I  believe that Tempdata persists between 1 internal redirect?

rockstardev:
Code updated as we agreed on the meeting, per @Kukks feedback.

dennisreimann (approved):
ACK. Yet this in an indicator for us to refactor the temp message handling after this PR.

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -539,7 +540,7 @@ private async Task<bool> CanUseHotWallet()
                 }
                 catch (Exception ex) { model.RateError = ex.Message; }
             }
-
+
```
Kukks, line 543:
```suggestion

```

Kukks (approved):

PR #5762, opened 2024-02-16
Apps: Remove title and replace it with app name

### LINKED ISSUE #5757: App Name in POS should be the display title
There's no reason to have this duplication. 
App name should be a display name, whoever doesn't like it can change it.
People deploying btcpay stores  at scale are reporting they:
a) either accidentally leave tea shop in POS
b) always have to rename it

Approach, there's no reason to have both, App Name should be what is displayed in Display Title, let's remove redundancy and unify this view.
Screenshot 2024-02-16 at 12 40 46

dennisreimann:
Does it for POS and Crowdfund. Closes #5757.

### DIFF
--- a/BTCPayServer.Client/Models/CreateAppRequest.cs
+++ b/BTCPayServer.Client/Models/CreateAppRequest.cs
@@ -21,7 +21,6 @@ namespace BTCPayServer.Client.Models
     public class CreatePointOfSaleAppRequest : CreateAppRequest
     {
         public string Currency { get; set; } = null;
-        public string Title { get; set; } = null;
         public string Description { get; set; } = null;
         public string Template { get; set; } = null;
         [JsonConverter(typeof(StringEnumConverter))]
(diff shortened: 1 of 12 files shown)


pavlenex:
Thanks for details @rockstardev I didn't expect we'll get a PR this soon, but Dennis is ⚡. What you're proposing seems good, but from what I understand, we'd still keep the `Display Tittle` and `App Name` fields in the UI, or will it also be one field in the UI?

dennisreimann:
Yes it would keep both fields, but correctly populate the Title one right away. 

I like that suggestion and can revert the PR to do that if we agree this is better – it would be the more flexible and also backwards compatible approach, but I get why we would also want to remove ambiguity here. 

I started with this PR, because in the chat it sounded like an urgent issue to me.

dstrukt:
All for keeping backwards compatibility, but removing the ambiguity would also be nice - would be nice to phase the second input out, but cACK.

Kukks:
> Yes it would keep both fields, but correctly populate the Title one right away.
> 
> I like that suggestion and can revert the PR to do that if we agree this is better – it would be the more flexible and also backwards compatible approach, but I get why we would also want to remove ambiguity here.
> 
> I started with this PR, because in the chat it sounded like an urgent issue to me.

lets do this for now and maybe switch to a singular one later on

PR #5772, opened 2024-02-20
Adding link to API usage examples in docs.

ndeet:
I tried to find that text "BTCPay Greenfield Plugins API" but it does not exist in the swagger files. Not even searching for "Plugins" brought up relevant text or link. How it looked before:

screenshot-docs btcpayserver org-2024 02 20-10_32_15

When adding "externalDocs" the text magically disappeared and was replaced with the proper link, how it looks now:

screenshot-localhost_14142-2024 02 20-10_31_23

@pavlenex the text is currently "Check out our examples on how to use the API" and it links to https://docs.btcpayserver.org/Development/GreenFieldExample/

Feel free to suggest different naming and/or link to link to. Maybe we should do it like this now but as discussed have a better page to start (or expand what we have with curl examples). Anyway better than status quo.

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.json
@@ -19,6 +19,10 @@
       "description": "BTCPay Server Greenfield API"
     }
   ],
+  "externalDocs": {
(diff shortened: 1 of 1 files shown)


ndeet:
cc @pavlenex

PR #5777, opened 2024-02-21
Support pluginable rate providers

Kukks:
This PR allows plugins to provide custom rate providers, that can be contextual to a store. For example, if you use the upcoming fiat offramp plugin, or the Blink plugin, you'll probably want to configure the fetch the rates from them since they are determining the actual fiat rrate to you. However, they require API keys. This PR enables these scenarios, even much more advanced ones, but for example:
* Install fiat offramp plugin
* Configure it
* You can now use the fiat offramp rate provider (no additional config steps beyond selecting the rate source from the select, or maybe the plugin would automatically set it for you once configured)

### DIFF
--- a/BTCPayServer.Rating/Services/RateProviderFactory.cs
+++ b/BTCPayServer.Rating/Services/RateProviderFactory.cs
@@ -15,7 +15,6 @@ namespace BTCPayServer.Services.Rates
 {
     public class RateProviderFactory
     {
-        
         class WrapperRateProvider : IRateProvider
         {
             public RateSourceInfo RateSourceInfo => _inner.RateSourceInfo;
@@ -91,8 +90,6 @@ namespace BTCPayServer.Services.Rates
                     AvailableRateProviders.Add(coingecko.RateSourceInfo);
                 }
             }
-
-          
             AvailableRateProviders.AddRange(_dynamicProviders.Select(provider => provider.RateSourceInfo));
             AvailableRateProviders.Sort((a, b) => StringComparer.Ordinal.Compare(a.DisplayName, b.DisplayName));
         }
--- a/BTCPayServer.Rating/Services/RateProviderFactory.cs
+++ b/BTCPayServer.Rating/Services/RateProviderFactory.cs
@@ -15,6 +15,7 @@ namespace BTCPayServer.Services.Rates
 {
     public class RateProviderFactory
     {
+        
         class WrapperRateProvider : IRateProvider
         {
             public RateSourceInfo RateSourceInfo => _inner.RateSourceInfo;
@@ -50,9 +51,10 @@ namespace BTCPayServer.Services.Rates
             public ExchangeException Exception { get; internal set; }
             public string Exchange { get; internal set; }
         }
-        public RateProviderFactory(IHttpClientFactory httpClientFactory, IEnumerable<IRateProvider> rateProviders)
+        public RateProviderFactory(IHttpClientFactory httpClientFactory,IEnumerable<IRateProvider> rateProviders, IEnumerable<IDynamicRateProvider> dynamicProviders)
         {
             _httpClientFactory = httpClientFactory;
+            _dynamicProviders = dynamicProviders;
             foreach (var prov in rateProviders)
             {
                 Providers.Add(prov.RateSourceInfo.Id, prov);
@@ -60,6 +62,7 @@ namespace BTCPayServer.Services.Rates
             InitExchanges();
         }
         private readonly IHttpClientFactory _httpClientFactory;
+        private readonly IEnumerable<IDynamicRateProvider> _dynamicProviders;
         public Dictionary<string, IRateProvider> Providers { get; } = new Dictionary<string, IRateProvider>();
         void InitExchanges()
         {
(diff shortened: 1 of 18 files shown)


BTCPayServer.Rating/Services/RateProviderFactory.cs
```diff
@@ -88,15 +91,25 @@ void InitExchanges()
                     AvailableRateProviders.Add(coingecko.RateSourceInfo);
                 }
             }
+
```
Kukks, line 94:
```suggestion
```

BTCPayServer.Rating/Services/RateProviderFactory.cs
```diff
@@ -88,15 +91,25 @@ void InitExchanges()
...
+
```
Kukks, line 95:
```suggestion
```

BTCPayServer.Rating/Services/RateProviderFactory.cs
```diff
@@ -15,6 +15,7 @@ namespace BTCPayServer.Services.Rates
 {
     public class RateProviderFactory
     {
+
```
Kukks, line 18:
```suggestion
```

NicolasDorier:
I think we should allow rate providers to have settings at the store level, but I am not fan of this solution. It burdens all callers with a cryptic `context` parameter, which may not be interpreted correctly by the dynamic provider receiving it.

It also doesn't take advantage of the background rate fetcher which avoid hammering the rate source which was always resulting in flaky checkout experience.

I'll try to find another solution so the RateFetcher can have hooks attached to it.

NicolasDorier:
I rewrote part of your PR, keeping the overall approach.
* Rename `DynamicRateProvider` to `IContextualRateProvider`, a subclass of `RateProvider`.
* Rather than having a string context, it is now passing an `object`. Only one type is passed into the code today `StoreIdRateContext`. It has an interface `IHasStoreIdRateContext` on which the implementations should rather check. This is made so later we can for example pass a `PullPaymentRateContext` without breaking the existing contextual rate providers
* As `IContextualRateProvider` are `IRateProvider`, they are registered in the same way as current providers.
* Added a test, making sure context get passed down properly.
* Fix a bug: The rates get updated by an hosted service. This hosted service doesn't have any context, as such this would crash on contextual rate provider and spam the logs every 30 seconds.
* Allow rate providers can throw a `NotSupportedException` if an unexpected context is sent to them, this will not spam the logs.

Not that this mean that if two store have blinks, the rate of one might come from the API key of another.

Kukks:
> I rewrote part of your PR, keeping the overall approach.
> * Rename `DynamicRateProvider` to `IContextualRateProvider`, a subclass of `RateProvider`.
> * Rather than having a string context, it is now passing an `object`. Only one type is passed into the code today `StoreIdRateContext`. It has an interface `IHasStoreIdRateContext` on which the implementations should check instead of `StoreIdRateContext`. This is made so later we can for example pass a `PullPaymentRateContext` without breaking the existing contextual rate providers
> * As `IContextualRateProvider` are `IRateProvider`, they are registered in the same way as current providers.
> * Fix a bug: The rates get updated by an hosted service. This hosted service doesn't have any context, as such this would crash on contextual rate provider and spam the logs every 30 seconds. I fixed it by having the `BackgroundRateProvider` remember the last successful context.
> * Allow rate providers can throw a `NotSupportedException` if an unexpected context is sent to them, this will not spam the logs.
> * Added a test, making sure context get passed down properly, and that the background refresh works as expected.
> 
> Note that this mean that if two store have blinks, the rate of one might come from the API key of the other store!
> 
> Also note that BTCPay, before shutting down and before starting up, save the last rates. This is done so the customer, after a restart always have the rates prefetched and checkout come fast. We won't be able to do this with contextual rate provider because we aren't saving the context in the DB.

Note that this mean that if two store have blinks, the rate of one might come from the API key of the other store!

This is really bad and is very dangerous.. A provider may have different rates based on api keys due to different customer agreements

NicolasDorier:
> This is really bad and is very dangerous.. A provider may have different rates based on api keys due to different customer agreements

Is it something that could potentially happen in the future? or something that is actually happening today? (Important, as I don't want to spend too much time designing for something that **might** happen)

The background fetcher is important to make sure that when an invoice is created, the rates don't have to be fetched and thus making the checkout flaky.
The problem is that in the current architecture, the background fetcher apply to a specific rate provider, and there is a single rate provider: `blink`, even if it is a contextual one.

Would we want to fix this problem, we need to allow, at rate provider registration time, a way to opt out of the background fetcher. But if we do this, the providers itself would need to duplicate the logic of the background fetcher. (Or not rely on background fetcher at all and have a flaky checkout experience)

Kukks:
> > This is really bad and is very dangerous.. A provider may have different rates based on api keys due to different customer agreements
> 
> Is it something that could potentially happen in the future? or something that is actually happening today? (Important, as I don't want to spend too much time designing for something that **might** happen)
> 
> The background fetcher is important to make sure that when we are creating an invoice, the rates don't have to be fetched, which would making the checkout slow and flaky. (Rate APIs often have small maintenance windows, crash transiently or are slow as fuck) The problem is that in the current architecture, we automatically wrap every rate provider with a background fetcher. And in our case, `blink`, even if it is a contextual one, isn't treated specially.
> 
> Would we want to fix this problem, we need to allow, at rate provider registration time, a way to opt out of the background fetcher.
> 
> Another possibility is to not wrap the contextual rate providers with a background fetcher.
> 
> But if we do this, the providers itself would need to duplicate the logic of the background fetcher to not be flaky.

Yes, I already have a use-case specifically with the new Bringin plugin, where some users might have a promotion and their rates would be different to others

NicolasDorier:
Got it. On it.

NicolasDorier:
I removed the background fetching for contextual providers

PR #5779, opened 2024-02-21
Apps: Make app name the default title

dennisreimann:
Successor of #5762 with a way simpler approach. Allows the user-facing title to be set, but defaults it to the app name instead of "Tea shop".


pavlenex:
cACK

dstrukt (approved):
functional tACK

pavlenex (approved):
User-tested ACK, needs code review.

Kukks (approved):

PR #5781, opened 2024-02-22
Parallel payout ln

Kukks:


NicolasDorier:
From bitcoin madeira, I think there is a bug, which might only be logging:

```
2024-03-03T22:25:18.203752223Z fail: LightningAutomatedPayoutSenderFactory:4ma1qdRjdSJY8UDCbtYiG87ir9fPBQNxixpmNWtZqNXg:BTC_LightningLike: Could not mark payout BQaQC6oZw92h7wpDhpXXVXPtfzz as Completed because InvalidState
```

Do we need to set the payout to complete? I guess this is handled by another process as soon as the payment is received?

feathers2fly1:
Help me how do I do that

PR #5783, opened 2024-02-23
refund report

Kukks:

PR #5791, opened 2024-02-26
Add refund reports

NicolasDorier:
Refund reports.

I didn't see @Kukks has been working on it, possible alternative to https://github.com/btcpayserver/btcpayserver/pull/5783

Reviewed Kukks's one, outside a minor point (not using the display formatter), it's also good, no strong preferences.

image

Still need tests

### DIFF
--- a/BTCPayServer.Tests/UnitTest1.cs
+++ b/BTCPayServer.Tests/UnitTest1.cs
@@ -3006,7 +3006,7 @@ namespace BTCPayServer.Tests
         }
 
         [Fact]
-        [Trait("Selenium", "Selenium")]
+        [Trait("Integration", "Integration")]
         public async Task CanCreateReports()
         {
             using var tester = CreateServerTester(newDb: true);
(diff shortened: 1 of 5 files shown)


pavlenex:
Unlike https://github.com/btcpayserver/btcpayserver/pull/5783, this one actually worked for me in dev enviroment.

@NicolasDorier Thoughts on giving crypto amount  rate and currency? IMO these should reflect the typical payment report just do it for refunds and provide as much info as possible, you never know what people actually need from them.

Can you explain what fully paid status is? Is it full `CryptoAmount` or just complete refund or something else?

NicolasDorier:
@pavlenex Fully Paid is saying if all the money has been transferred from the pull payment of the refund.
We don't keep information on which way the user decided to reimburse. We only know he did a refund of X amount in X currency, which is all that is shown here.

That said, for BTCPay 2.0, we can start saving additional information for future refunds.

NicolasDorier:
This is now ready and tested.

pavlenex (approved):
Looks good to me, but I have't tested in-depth.

PR #5807, opened 2024-03-03
(feat) monero settlement thresholds

HennyH:
# Overview

This PR adds the ability to configure a monero specific settlement confirmation threshold. The UI for this setting was based of the existing 'speed policy' dropdown. This feature was requested by users in this [bounty item](https://bounties.monero.social/posts/106/2-900m-enable-zero-confirmation-for-monero-on-btcpay-server).

# Bug Fixes

## Treat Monero Wallet Directory Setting as Required

It also fixes a crash which occurs when using the monero feature without having specified a wallet base directory. This is required because the `GetMoneroLikePaymentMethodViewModel` method checks if the wallet file exists and as part of that check it tries to combine the wallet base directory with "wallet" which fails if the wallet base directory is null.

## Labels Without Spaces

It also fixes some of the form labels which were missing spaces.

## Unreadable Validation Messages

Previously the `text-danger` class was being added to the validation summary element causing the text to be red too. This has been fixed and tested on both light and dark themes.

image

# UI Changes

Adds a new 'Consider the invoice settled when the payment transaction ...' field that defaults to 'Store Speed Policy':

image

The options are as follows:

image

These options were based upon the expected use cases (default, 0 for instant, 1, and then 10 as spendable). When the 'Custom' value is selected an additional field appears in the form:

image

This field is restricted to integer values between 0 and 100:

image
image

If you choose custom but don't provide a value an error is displayed:

image

# Testing

To test this change I modified to development docker compose file to run a testnet node. I then created invoices and used the POS app using the different payment thresholds. The zero-confirmation option worked as expected.

A video demonstrating the PR can be viewed [here](https://www.youtube.com/watch?v=HwmNlieh4JI) - the zero conf is demoed first, and then a custom setting of 2 confirmations. It took a long time for the confirmations to come through.

### DIFF
--- a/BTCPayServer/Services/Altcoins/Monero/MoneroLikeExtensions.cs
+++ b/BTCPayServer/Services/Altcoins/Monero/MoneroLikeExtensions.cs
@@ -73,7 +73,7 @@ namespace BTCPayServer.Services.Altcoins.Monero
                 var daemonPassword =
                     configuration.GetOrDefault<string>(
                         $"{moneroLikeSpecificBtcPayNetwork.CryptoCode}_daemon_password", null);
-                if (daemonUri == null || walletDaemonUri == null)
+                if (daemonUri == null || walletDaemonUri == null || walletDaemonWalletDirectory == null)
                 {
                     throw new ConfigException($"{moneroLikeSpecificBtcPayNetwork.CryptoCode} is misconfigured");
                 }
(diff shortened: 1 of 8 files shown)


Kukks (approved):

Kukks:
I believe this PR broke monero integration for many users. @HennyH

HennyH:
Do you have any details about what is breaking for them?

Issue #5730, opened 2024-02-05
[Bug]: Inability to see all payments in `payments` report 


pavlenex:
### What is your BTCPay version?

BTCPay Server v1.12.5+a86165ac0

### How did you deploy BTCPay Server?

Docker

### What happened?

When trying to generate a report, BTCPay is unable to show all results for a specified time-frame, and in some cases no results at all.

We have two users reporting this with 1.12.5 thus far.

### How did you encounter this bug?

1. Check invoice page to see multiple invoices with different payment method
2. Go to reports, selcect time-frame 
3. Not all payments are exported (seems mostly on-chain is missing but this needs to be confirmed)

### Relevant log output

```shell
No server logs, but this is logs in browser console.


[Info] [2024-02-05T11:09:46.970Z] Information: Normalizing '_blazor' to 'https://pay.store/_blazor'. (blazor.server.js, line 1)
[Info] [2024-02-05T11:09:47.313Z] Information: WebSocket connected to wss://pay.store/stores/9BXHj48xGPK4QXktLkJJaxwSXg4XVXWbkaNvDVQxLBYM/_blazor?id=dIld8IxCSsUgackLaNHNtA. (blazor.server.js, line 1)
```
```


### What browser do you use?

Tried Safari Version 17.3 (19617.2.4.11.8)
Google Chrrome 21.0.6167.139 (Official Build) (arm64)


### Additional information

_No response_

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

pavlenex:
It seems reports may not work with some older versions of BTCPay, which is the case here. Meaning users who received payments in 2021/2022 may not be able to see them in reports. @NicolasDorier may be able to provide more details on this one.

PR #5809, opened 2024-03-04
The Big Cleanup

NicolasDorier:
# Abstract

BTCPay Server design has evolved incrementally since its creation in 2017.
In order to minimize breaking changes, and make PR easier to review, we decided that, rather than rewriting the data in the DB, we would hide the legacy behind helper methods or interfaces which would support all data format.

Fast forward today, I believe it is time that we rewrite all the entities in the Invoices/Stores and Payments table to have a new design which take into account the needs of tomorrow, and what we learnt over the years, while cleaning up the years of legacy decision from the main code.

## Data changes

The invoice and payment table are too huge to rewrite during a migration step while the server is starting.
For this reason, the approach chosen is to leverage Entity Framework with an Interceptor, to migrate the rows on the fly when EF materialize a row into an object. (See Migrate method of PaymentData and InvoiceData)

On top of this, an hosted service `InvoiceBlobMigratorHostedService`, is scanning all the invoices and the payments, migrate them and save them in the database in parallel while the BTCPay Server instance is operating normally.

## UI Change

The receipts don't have a "Proof" field anymore. (Outpoint on-chain or pre-image for lightning)
We need to add it back, but need to be extensible.

Crowdfund: When you hover the amount received, a tooltip opened with the amount of BTC/Sats received in LN versus BTC.
I changed it to be the % of contribution of both.

Never refer to Lightning payment method as `Bitcoin (Lightning)` or `Lightning (Bitcoin)`, and instead just use `Lightning`. (altcoins are unchanged, ie `Lightning (Litecoin)`)

## Architecture change

`BTCPayNetwork` is a class representing Bitcoin or a crypto currency forked from Bitcoin. This class was tied to the concept of payment method, as all payments method were historically crypto currencies.

With this PR, the two concepts are now decoupled, as such:

* `PaymentMethodId` doesn't have any concept of `PaymentType` or `CryptoCode` attached to it anymore.
* The features of `BTCPayNetwork` which are generic enough to non crypto related payment method have been extracted into their own extension interface. `IPaymentMethodHandler`, `IPaymentModelExtension`, `IPaymentMethodViewExtension`, `IPaymentMethodBitpayAPIExtension`. Only `IPaymentMethodHandler` is mandatory.
* A `Payment Method` has a `config` in the StoreData, and when an invoice use a payment method, it creates a `Payment Prompt` attached to the invoice containing the information to settle the invoice using this payment method. Such prompt can contains `Prompt Details` to store data specific to this payment method.
* When a payment is made, the payment data contains `Payments Details` to store data specific to a payment made by this payment method.
* A Payment and a Payment Prompt have a specific `Currency`.
* A Payment Prompt have a specific divisibility, which may differ from the normal currency divisibility. (For example, BTC can have 11 decimals in lightning network) This influence the rounding calculation of required payment amount.
* `Destination`, `PaymentMethodFee` and `Currency` are now mandatory properties for all payment prompts and payments.
* A column `Created` datetime has been added to PaymentData.

# Subject to changes for future PRs

The design of the new extension interfaces has been done to avoid having to change them in the future.
Changing an interface is a breaking change for plugins so we should limit this as much as possible.

That said, I believe the correct design of those interface will need more polishing and feedback from @Kukks and other plugin devs.

# Postboned for future PRs

### Changing invoice status to the new invoice status

In the database, the old bitpay status are still used. They should be rewritten to the new states.

### Payment Proof not showing in receipts
I noticed the concept of `Payment Proof` is only showing up the receipts.

So I believe a `IPaymentMethodReceiptExtension` would be useful so the payment methods decide themselves what to show up in receipts.

### Removing CryptoCode from PaymentMethodId

Payouts depends too much on it, and we need to refactor the payouts. So this will be done later.

### Payout refactoring

It turns out that Payouts are tied to PaymentMethodId, I believe there should be a PayoutMethodId as payouts and payments are disjoint concepts.

# Breaking changes

Payment methods id name have been rewritten to be consistent between API and Backend.

* `BTC-OnChain` or `BTC_BitcoinLike` or `BTC-BitcoinLike` or `BTC-BTCLike` -> `BTC-CHAIN`
* `BTC-LightningNetwork` or `BTC_LightningLike` or `BTC_OffChain` or `BTC_Off-Chain`  -> `BTC-LN`
* `BTC-LNURLPAY`->`BTC-LNURL`
* `XMR-MoneroLike`->`XMR-CHAIN`
* `ZEC-ZcashLike`->`ZEC-CHAIN`

Note that it isn't a breaking change for input data to API, as BTCPay will automatically convert the old strings to the new format.
However, data returned by BTCPay API will return the new names.

The API needed to be changed to reflect the fact payment methods doesn't have a PaymentType anymore:

Removed routes:
* `GET /api/v1/stores/{storeId}/payment-methods/LNURL`
* `GET /api/v1/stores/{storeId}/payment-methods/LNURLPay/{cryptoCode}`
* `PUT /api/v1/stores/{storeId}/payment-methods/LNURLPay/{cryptoCode}`
* `DELETE /api/v1/stores/{storeId}/payment-methods/LNURLPay/{cryptoCode}`
* `GET /api/v1/stores/{storeId}/payment-methods/LightningNetwork`
* `GET /api/v1/stores/{storeId}/payment-methods/LightningNetwork/{cryptoCode}`
* `PUT /api/v1/stores/{storeId}/payment-methods/LightningNetwork/{cryptoCode}`
* `DELETE /api/v1/stores/{storeId}/payment-methods/LightningNetwork/{cryptoCode}`
* `GET /api/v1/stores/{storeId}/payment-methods/OnChain`
* `GET /api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}`
* `PUT /api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}`
* `DELETE /api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}`

Modified route:
* `GET /api/v1/stores/{storeId}/payment-methods`
  * Remove the `cryptoCode` field
  * Rename the `paymentMethod` to `paymentMethodId` field
  * Rename the `data` property to `config`
  * Rename `networkFee` to `paymentMethodFee`
  * Rename the query parameter `enabled` to `onlyEnabled`
  * Add `includeConfig` parameter defaulting to `false`. If `true`, the config of the payment method will be returned and the permission `btcpay.store.canmodifystoresettings` is required.
 * `GET /api/v1/stores/{storeId}/invoices/{invoiceId}/payment-methods`
   * Remove the `cryptoCode` field
   * Renamed `paymentMethod` to `paymentMethodId`
   * Add a `currency` field

Added route:
* `PUT /api/v1/stores/{storeId}/payment-methods/{paymentMethodId}`
* `DELETE /api/v1/stores/{storeId}/payment-methods/{paymentMethodId}`
* `GET /api/v1/stores/{storeId}/payment-methods/{paymentMethodId}`

Renamed route:
* `GET api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/preview` to `api/v1/stores/{storeId}/payment-methods/{paymentMethodId}/preview`
* `POST api/v1/stores/{storeId}/payment-methods/onchain/{cryptoCode}/generate` to `api/v1/stores/{storeId}/payment-methods/{paymentMethodId}/generate`
* Removed properties `cryptoCode` and `paymentMethod` from their response.
* Added property `paymentMethodId`, `config` so the format of `generate` is the same as the `GET /api/v1/stores/{storeId}/payment-methods/{paymentMethodId}`

Webhook change:
* Rename `paymentMethod` field to `paymentMethodId`

Legacy Bitpay API:
* Removal of the `confirmed` state, an invoice directly goes to the `complete` state.


NicolasDorier:
I am actually not sure it fix https://github.com/btcpayserver/btcpayserver/issues/5730, will make additional tests

NicolasDorier:
Confirmed, this doesn't work, I need to migrate payments too.

NicolasDorier:
So @Kukks reported 2 points:
1. One payment method should be able to create several prompts.
2. Prompts should not be overidden but replaced.

If we take both into account, payment method prompt should really have their own table.

NicolasDorier:
There are some issues with multiple payment prompt per payment method around activation where there is no clearcut solution.

For example, a payment prompt can be activated or not. But the InvoiceActivator asks for the payment method id to activate, not the prompt. It would require additional refactoring, and also mean that there would need some kind of id for a prompt which currently doesn't exist.

PR #5814, opened 2024-03-05
POS: Add item list to keypad

dennisreimann:
- Refactors the common POS Vue mixin to share most of the functionality between keypad and cart
- Adds an option to show the items list for the keypad; if enabled, aadds the item list to the keypad view
- Extends the keypad receipt data with item info and separate values for manuel entries 
- Offers the ability to add recent transactuions to the cart view as well (not done yet)

## Keypad with items

keypad

## Receipt

receipt


dstrukt:
Man I love seeing this all come together!

The line below `Subtotal` should be 1px imo on the Receipt.

dennisreimann:
> The line below Subtotal should be 1px imo on the Receipt.

There could also be discounts and tips listed. Over here it's a common format to separate list items / subtotal and extras / total. Shall I still change it? 

> Debating if I like the hamburger icon for the Products, but looks great regardless!

Let me know if we have a better one or you wanna do a new one :)

pavlenex:
Haven't looked at this one yet, but there needs to be a call to action to proceed, currently (what video shows) is that call to action to proceed is to `close` which doesn't seem very logical to me.

dennisreimann:
I've now added a bunch of tests for the keypad and cart to also verify that the receipt contains proper information.

dennisreimann:
I think this works well:

pavlenex:
Great solution @dennisreimann !

dstrukt:
I didn't originally find it too confusing, and would be curious if users do, but if we choose to proceed with this direction, @dennisreimann's solution does work. Alternatively, incrementing the number (think notification number) on the products icon could work too to show change. Makes me think a reasonable icon might be a cart / bag.

dennisreimann:
Let's use the current version, imho it works best. If we'd go with a numbered badge, then there could be confusion about the manual entries via the keypad.

dstrukt (approved):
tACK - LGTM with the latest suggestions, we'll continue to collect feedback and adjust as needed!

pavlenex:
@Kukks Feel free to review code when you get a chance and if all good, you can merge it, we user-tested it.

Kukks (approved):

PR #5844, opened 2024-03-16
Specify mailto: prefix for emails in Server Settings

### LINKED ISSUE #5843: Specify `mailto:` prefix for emails in `Server Settings > Branding > Contact`
We should specify that for emails `mailto:` needs to be added, otherwise if user adds email@email.com and someone clicks, `contact us` on login page,  it gives 404. Even better if we can recongize it's an email and no need for mailto.

**To replicate:**

1. In Server Settings > Branding > Contact us add email@test.com
2. Log out
3. Go to login page, click contact us
4. 404 
5. To fix add mailto:email.test.com

##Solution: 
- Recognize it's email and apply mailto so user doesn't have to,
- less ideal: add text that for emails mailto: prefix is required.           

Screenshot 2024-03-15 at 23 08 24
Screenshot 2024-03-15 at 23 08 51

Feature Introduced in this PR https://github.com/btcpayserver/btcpayserver/pull/5718

TChukwuleta:
Resolves #5843 by Specifying mailto: prefix for an contact url in server settings that is an email

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.cs
+++ b/BTCPayServer/Controllers/UIServerController.cs
@@ -1049,11 +1049,22 @@ namespace BTCPayServer.Controllers
             vm.LogoFileId = theme.LogoFileId;
             vm.CustomThemeFileId = theme.CustomThemeFileId;
             
-            if (server.ServerName != vm.ServerName || server.ContactUrl != vm.ContactUrl)
+            if (server.ServerName != vm.ServerName)
             {
                 server.ServerName = vm.ServerName;
-                server.ContactUrl = vm.ContactUrl;
                 settingsChanged = true;
+            }
+            
+            if (server.ContactUrl != vm.ContactUrl)
+            {
+                server.ContactUrl = !string.IsNullOrWhiteSpace(vm.ContactUrl)
+                    ? vm.ContactUrl.IsValidEmail() ? $"mailto:{vm.ContactUrl}" : vm.ContactUrl
+                    : null;
+                settingsChanged = true;
+            }
+            
+            if (settingsChanged)
+            {
                 await _SettingsRepository.UpdateSetting(server);
             }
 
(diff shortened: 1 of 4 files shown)


Kukks (commented):
Maybe the mail to should be part of the source, since you can technically have the same format for nostr but with a nostr: scheme

TChukwuleta:
Hey @Kukks, correct me if I am wrong, but contact using nostr would involve updating the branding with the npub key. Is there any other form of contact for nostr? If none, we can do a double ternary for mail and nostr key

I'm just here thinking how the mailto would be added to input field.  Maybe the business would do this themselves, by adding it before they type out the actual mail. In that case we send a description message and validation across.

What I am just considering is if a business decide to start with mail and after sometimes switches to either website url, nostr or whatever platform, How it would be handled UX wise, that is if mailto added to the source

Kukks:
> Hey @Kukks, correct me if I am wrong, but contact using nostr would involve updating the branding with the npub key. Is there any other form of contact for nostr? If none, we can do a double ternary for mail and nostr key
> 
> I'm just here thinking how the mailto would be added to input field.  Maybe the business would do this themselves, by adding it before they type out the actual mail. In that case we send a description message and validation across.
> 
> What I am just considering is if a business decide to start with mail and after sometimes switches to either website url, nostr or whatever platform, How it would be handled UX wise, that is if mailto added to the source
> 

Yeah, in nostr there is nip05, which is an internet identifier format (kukks@kukks.org), same as emails and lightning addresses. This translates to a url that you do a GET which returns an npub.

Conceptually you can even use lightning addresses as a form on contact, since lnurl allows passing comments on a payment 😅

So to summarize:
Kukks@kukks.org could be:
* an email: mailto:Kukks@kukks.org
* a nostr profile: nostr:Kukks@kukks.org
* a lightning address: lightning:Kukks@kukks.org

And all allow a form of communication to the receiver.
I would suggest that we simply show the merchant a good helper text on how to populate the field, it's better than hardcoded prefixing based on assumptions

TChukwuleta:
Yea... Sounds better. 👌

I'd update the description and revert the previous commit

pavlenex:
Why are we optimizing for 1% of use-case, while it's technically possible for someone to paste lightning address, who would realistically do that? 99% would want to paste either an email or an external link to telegram or something else.

dennisreimann:
It's present on all non-logged in pages. Register, forgot password, etc.

dennisreimann (approved):
tACK.

dstrukt (approved):
Just wanted to see what conclusion was reached - tACK as well!

PR #5849, opened 2024-03-18
Template Editor: Apply item changes directly

### LINKED ISSUE #5847: Item editor: Each item needs to be `saved` separately
When dealing with a bulk of items, adding or changing, our POS currently expect that for every product change user needs to click on the `Apply`

Is there a way to unify these so that `Save` is the only call to action necessary? I'd assume it can be done as part of https://github.com/btcpayserver/btcpayserver/issues/5581 or https://github.com/btcpayserver/btcpayserver/issues/3827 that would need to be completed beforehand.

dennisreimann:
Closes #5847.


dstrukt:
I'm still trying to remember why we didn't do this on the original PR. I feel like I remember bringing up a similar idea.

dstrukt (approved):
I like it!

tACK - LGTM

Issue #5784, opened 2024-02-23
Reposition the camera scan icon is the wallet > send


Transisto:
https://github.com/btcpayserver/btcpayserver/discussions/5366

This is low hanging fruit that would make using the wallet much better on mobile

![271641565-5ea6596e-0947-416f-b114-74d1c7a93b96](https://github.com/btcpayserver/btcpayserver/assets/36959754/5ba074c8-29b9-4cfa-8b84-1b8cc0a9b7fa)

pavlenex:
I think this is a good suggestion. 

@TChukwuleta feel free to tackle it if you wish!

cc @dstrukt i don't see any better way to achieve this than what @Transisto proposed, if you do, let us know.

TChukwuleta:
Sure @pavlenex

dstrukt:
Agree with @Transisto's original proposal, and agree @pavlenex we should use the QR code icon instead of the camera icon, which should already be in the icon set, if not @TChukwuleta feel free to reach out and I'll get you what you need!

PR #5851, opened 2024-03-19
Wallet UI quick wins

### LINKED ISSUE #5850: Coinselection should show more utxo and allow to sort by amount and show fiat value
The way I use coin control is to find an amount that match closely what I want to send and leave no change.

Like when funding another wallet / account.

10 per page is not enough and there should be sorting.  Could switch to sorting by amounts by default since I don't think many people care about the block height.

dennisreimann:
Draft mode, so that we can discuss the following changes:

- Unify single/multiple inputs display
- Destination: Move labels above amount
- Coin Selection: Add sorting by amount and confirmations, add page size. Closes #5850
- Move PSBT and BIP21 buttons up to input related button group
- Turn checkboxes in Advanced Settings into toggles
- Improve spacings and button groups

## Simple

wallet-plain

## Full

Screenshot 2024-03-19 at 13-22-33 Send BTC

### DIFF
--- a/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
+++ b/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
@@ -44,7 +44,7 @@ namespace BTCPayServer.Models.WalletViewModels
 
         public List<FeeRateOption> RecommendedSatoshiPerByte { get; set; }
 
-        [Display(Name = "Fee rate (satoshi per byte)")]
+        [Display(Name = "Fee rate (sat/vB)")]
         [Required]
         public decimal? FeeSatoshiPerByte { get; set; }
 
--- a/BTCPayServer/Views/UIWallets/CoinSelection.cshtml
+++ b/BTCPayServer/Views/UIWallets/CoinSelection.cshtml
@@ -1,6 +1,6 @@
 @model WalletSendModel
 
-<div class="form-group hide-when-js mt-4">
+<div class="form-group hide-when-js">
     <label asp-for="SelectedInputs" class="form-label"></label>
     <select multiple="multiple" asp-for="SelectedInputs" class="form-select">
         @foreach (var input in Model.InputsAvailable)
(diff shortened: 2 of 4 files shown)


nostitos:
Moving the camera button to the address bar is a very low hanging fruit too.

dennisreimann:
> Moving the camera button to the address bar is a very low hanging fruit too.

Re-added.

nostitos:
https://github.com/btcpayserver/btcpayserver/issues/5784

Supposedly it's done.

dennisreimann:
Yes, it got lost in my first version of this draft though. Thanks for the pointer!

dstrukt:
> Moving the camera button to the address bar is a very low hanging fruit too.

ACK, glad to see it was done, have more thoughts, but immediate ones:

- Like the 2 column approach, but think the `Hide Coin Selection` CTA should be more correlated with that section, and `Add Destination` should take up the full line imo

...need to explore a better way to group this, as it'd be nice to push the CTA to the top right of the Coin Selection section, but then it wouldn't look great when hidden

- X selected / Total should be put where the Sort by is, then

- Sort by should be pushed to the right

- Think we should add cards around the destination sections, so we can eliminate the <hr> tags, will help delineate sections better

- Could arguably do that to the Coin Selection section as well

Mind if I take a quick stab at some visual explorations based on this? Love the push, looks much better!

dennisreimann:
> Mind if I take a quick stab at some visual explorations based on this? Love the push, looks much better!

Sure, go ahead, glad to get this going.

Pjrich1313 (changes requested):
Nodes and legally binding contracts only

dennisreimann:
@dstrukt Shall we merge this and you'll take it from there?

dennisreimann:
@NicolasDorier This one is ready for code review and merge. @dstrukt might do additioonal improvements based on this in a separate PR.

dstrukt (approved):
tACK

Going to follow up on this PR with a visual polish. Don't want to delay the improvements for that!

PR #5853, opened 2024-03-20
Refactoring UIStoresController

dennisreimann:
- Use file-scoped namespace
- Use primary constructor
- Extract Rates, Settings and Token actions into partial class
- Move private methods to partial classes
- Simple syntax cleanups


NicolasDorier:
I don't think we should use primary constructor at this stage, those a pain in the ass to debug because the debugger doesn't show the properties...

Also please don't merge it before my massive PR.

PR #5854, opened 2024-03-20
UI: Improve Create First Store view

dennisreimann:
Quick update, which unifies the width and display with the login view.

## Before

before

## After

after


dstrukt (approved):
tACK - LGTM!

PR #5863, opened 2024-03-25
Remove custodians

Kukks:


dennisreimann (approved):
ACK. We could now also hide the Server Settings > Policies > Experimental toggle, as this was the only experimental feature affected.

NicolasDorier:
as soon as 1.13 is released, I'll merge this before https://github.com/btcpayserver/btcpayserver/pull/5809

PR #5877, opened 2024-03-26
POS: Increase size of quantity buttons

### LINKED ISSUE #5873: New item view for keypad (possible improvements)
Leaving this for discussion for design meeting, but I've tested our new keypad pos on bitcoinize, and +- experience is a bit challenging. All other buttons are okay, but these seem to be hard. I assume same exprience would happen on other places where we have thi.

I just imagine burger guy with gloves trying to navigate around to add/remove items, when I had troubles.

This is me attempting to add/remove. (gif may not be that good)

IMG_3687-2

Possible solution could be something that Wolt app does, asically you can click on item itself, or swipe left/right it's pretty snappy 

unsure if we can do it in a same way though.

RPReplay_Final1711456164-2

This issue is just a discussion for design call and shouldn't be tackled until there's consensus on how to approach it (possibly in other pos views) as well. cc @dstrukt @dennisreimann

dennisreimann:
Closes #5873.

items

cart

### DIFF
--- a/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
+++ b/BTCPayServer/Views/Shared/PointOfSale/Public/Cart.cshtml
@@ -163,13 +163,13 @@
                                     <span class="badge text-bg-warning inventory" v-if="item.inventory">
                                         {{ item.inventory > 0 ? `${item.inventory} left` : "Sold out" }}
                                     </span>
-                                    <div class="d-flex align-items-center gap-2">
-                                        <button type="button" v-on:click="updateQuantity(item.id, -1)" class="btn btn-light p-1 rounded-pill d-flex align-items-center justify-content-center">
-                                            <vc:icon symbol="minus" />
+                                    <div class="d-flex align-items-center gap-2 quantities">
+                                        <button type="button" v-on:click="updateQuantity(item.id, -1)" class="btn btn-minus">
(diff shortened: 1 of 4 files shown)


dstrukt:
hmmm, they are indeed quite large, and look out of proportion with the surrounding UI, which in my opinion, we should scale up all together or not, to keep things consistent

alternative solution .. is it possible to increase the padding/margins on the buttons themselves and keep the icons, etc.. the same size, thus increasing the hit area? or is that too hack-y for all the devices/OSs involved?

dennisreimann:
@dstrukt Agreed, that's the better soplution. Updated it.

dstrukt (approved):
Latest patch works well for me - attempted to fat finger and simulate what I could! 

Not sure if @pavlenex can test again on the Bitcoinize device (don't currently have one myself). Added him as a reviewer, but feel free to remove if needed.

The original issue definitely deserves some discussion, wrt if we want to consider multi tap / swipe / etc.. more "native" functionality on mobile, and/or to potentially do away with the counter(s), as these were originally for desktop specifically and they work well enough on a mobile browser + also having an increasing counter / total, etc.. on the product page as well, regardless this will help improve what's out there today!

pavlenex:
I can only test once it's on RC and on mainnet for now. So let's merge this one cc @NicolasDorier

PR #5885, opened 2024-03-30
fix: switch to using get_info call for monerod RPC

Jabster28:
Replace the RPC call to `sync_info` in the monero logic with `get_info`, which isn't a restricted call and allows for using a remote node. I'm also replacing the "peers" field with a "busy_syncing" field, since peers was never used and it clears up some logic.

I've only tested this locally, so please let me know if there are any issues.


plowsof:
Hello @Jabster28 there is now a separate bounty for adding syncing via remote node https://bounties.monero.social/posts/124/5-000m-btcpay-server-allow-syncing-via-remote-node

PR #5896, opened 2024-04-03
Receipt: Improve data display

### LINKED ISSUE #5882: [Bug]: Receipt Data does not show on printed paper version from POS.

BTCPay Server v1.13.0

Manual deployment.

### What happened?

I'm using the POS app and I have a form that collects some data before the invoice is displayed for payment, I want this data on the receipt screen page AND printed paper version, but the data only displays on the screen page. I use the receiptData "fieldset" in the form and the mirror type to get the data on the receipt page, however when i use the [Print] option to create a paper receipt, it only prints the basic, orderid and total on the receipt, no other data. I would expect it to print the form data collected for receipt also. 

Setup form to collect data, added receiptData with a mirror type

### Relevant log output

_No response_

### What browser do you use?

Pos app, used on chrome or brave, same result.

**Screen display showing fields on the page OK.**
image

**Print paper receipt shows no fields**.
I would expect to see name, title, description, guess

image

### Additional information

_No response_

dennisreimann:
Once more an improvement for the receipt, which also fixes #5882:

- Unify data displayed on the web and print version, format date and time on print version
- Split cart and additional data and ensure additional data is displayed
- Do not display extra subtotal row if there are no tips or discounts
- Make PosData partial more universal and backwards-compatible by using case insensitive key lookups

receipt

print

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -2620,9 +2620,9 @@ namespace BTCPayServer.Tests
 
             // Receipt
             s.Driver.WaitForElement(By.Id("ReceiptLink")).Click();
-            var additionalData = s.Driver.FindElement(By.CssSelector("#AdditionalData table"));
-            var items = additionalData.FindElements(By.CssSelector("tbody tr"));
-            var sums = additionalData.FindElements(By.CssSelector("tfoot tr"));
+            var cartData = s.Driver.FindElement(By.CssSelector("#CartData table"));
+            var items = cartData.FindElements(By.CssSelector("tbody tr"));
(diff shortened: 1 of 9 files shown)


dstrukt:
Minor immediate visual nit, let's reduce the bottom/top padding between the cards to maybe half (if 16 total ->, if 8 ->4) of what it is now.

Screen Shot 2024-04-03 at 1 54 32 PM

Looks great though - will give it a proper review this evening!

NicolasDorier:
ping @dstrukt

dstrukt (approved):
tACK - great improvements @dennisreimann! could only find that original nit - excited to see this carry over to the app

Issue #5881, opened 2024-03-29
[Bug]: after updating to version v1.13.0+5b31d4de2 (Altcoins) some stores are inaccessible


Bianconiglio:
### What is your BTCPay version?

i'm running btcpay server v1.13.0+5b31d4de2 (Altcoins) 

### How did you deploy BTCPay Server?

i'm running btcpay server on docker

### What happened?

after updating to version v1.13.0+5b31d4de2 (Altcoins) some stores are inaccessible. Some stores are displayed correctly, others are not accessible and give an HTTP ERROR 500 error

### How did you encounter this bug?

when I log in and enter the f2a code instead of viewing the default store it gives me HTTP ERROR 500 error. If I manually insert the code of another store into the URL then I see the screen. All the other functions of btcpay server continue to function normally, such as the payment form, for example

### Relevant log output

```shell
currently unable to handle this request.
HTTP ERROR 500
```

### What browser do you use?

any browser

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #5909, opened 2024-04-08
(bug) handle null XMR settings

HennyH:
# Overview

This PR resolves a crash which would occur in v1.13.0 whenever the user had no previously configured their XMR wallet settings. The cause of the crash was that code introduced in https://github.com/btcpayserver/btcpayserver/pull/5807 assumed that the `settings` variable was non-nullable but that is not the case. Could be related to #5881 

# Reproduction

Create a new postgres database, then w/ XMR enabled as an altcoin start up btcpayserver. After setting up you admin account and completing the "Create your first store" page the server will error:

image

# After Fix

The error does not occur and you can see the default store page.

image

And an example test transaction:

image
image
image


Kukks (approved):

NicolasDorier:
Releasing 1.13.1 tonight fixing this bug

PR #5918, opened 2024-04-09
allow admin update default rate provider and default currency for stores

### LINKED ISSUE #5653: Server admin can set up default rate provider and default currency for stores on the server
From @NicolasDorier 

Server admin can set up default rate provider and default currency for stores on server

TChukwuleta:
resolve #5653 

Allow server admin can change currencies and providers for stores. 

Cc. @dennisreimann


TChukwuleta:
this is up for review :)

NicolasDorier:
@TChukwuleta the idea isn't to give the admin the ability to change the default of individual stores.

The idea is to give the admin the ability to change what is selected by default in the "Create Store" wizard, so that new users don't have to select their currency and the right exchange manually.

TChukwuleta:
Oh. I see, thank you @NicolasDorier

TChukwuleta:
Hey @NicolasDorier  I updated the policy section in server settings to give the admin the ability to update default currency, which is now used as default for create currency, and in scenarios where there is none, it selects the previous default USD (storeblob.defaultcurrency)

NicolasDorier:
Did a minor rewording of the label.

PR #5919, opened 2024-04-09
Expose QR code generator

TChukwuleta:
Expose QR code generator as htmlstring in order to use it in payroll plugin for QR code display.


Kukks (changes requested):
Why not use the component directly?

```
<div class="qr-container" >
  <vc:qr-code data="your data here"/>
 </div>
```

TChukwuleta:
Thank you @Kukks. This actually solves my need

PR #5922, opened 2024-04-10
update csv export to include full date and time in 12 hour format

### LINKED ISSUE #5915: [Bug]: On CSV export we do have 12h format but no indication of am/pm

v1.12.5

docker

### What happened?

Reported by a user on Telegram: 

On csv export the time is in 12 hour format but am/pm is missing, on the UI it is 24h format. See screenshot:

image
image
image

we should do 24h format and UTC all the way through?

Export under Reporting

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

TChukwuleta:
Resolves #5915

### DIFF
--- a/BTCPayServer/wwwroot/js/store-reports.js
+++ b/BTCPayServer/wwwroot/js/store-reports.js
@@ -183,9 +183,9 @@ function downloadCSV() {
     if (!origData) return;
     const data = clone(origData);
 
-    // Convert ISO8601 dates to YYYY-MM-DD HH:mm:ss so the CSV easily integrate with Excel
+    // Convert ISO8601 dates to "dddd, MMMM Do YYYY, h:mm:ss a" to display full date plus time in 12 hour format so the CSV easily integrate with Excel
     modifyFields(srv.result.fields, data, 'amount', displayValue)
-    modifyFields(srv.result.fields, data, 'datetime', v => v? moment(v).format('YYYY-MM-DD hh:mm:ss'): v);
+    modifyFields(srv.result.fields, data, 'datetime', v => v ? moment(v).format("dddd, MMMM Do YYYY, h:mm:ss a"): v);
     const csv = Papa.unparse({ fields: srv.result.fields.map(f => f.name), data });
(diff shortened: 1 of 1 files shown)


Kukks (changes requested):
This is a breaking change so extra care should be considered.  Maybe we need to see what the old export on the invoices page used to export first

TChukwuleta:
This is the csv export view prior to this requested change. In 24 hours format:

image

BTCPayServer/wwwroot/js/store-reports.js
```diff
@@ -183,9 +183,9 @@ function downloadCSV() {
     if (!origData) return;
     const data = clone(origData);
 
-    // Convert ISO8601 dates to YYYY-MM-DD HH:mm:ss so the CSV easily integrate with Excel
+    // Convert ISO8601 dates to "dddd, MMMM Do YYYY, h:mm:ss a" to display full date plus time in 12 hour format so the CSV easily integrate with Excel
     modifyFields(srv.result.fields, data, 'amount', displayValue)
-    modifyFields(srv.result.fields, data, 'datetime', v => v? moment(v).format('YYYY-MM-DD hh:mm:ss'): v);
+    modifyFields(srv.result.fields, data, 'datetime', v => v ? moment(v).format("dddd, MMMM Do YYYY, h:mm:ss a"): v);
```
dennisreimann, line 188: To not break the curent format I'd suggest to simply use proper ISO format with added timezone info, like this: 

```suggestion
    modifyFields(srv.result.fields, data, 'datetime', v => v ? (new Date(v)).toISOString().replace('T', ' '));
```

Not sure if `v` is a `Date` already, if so it could be `v.toISOString()`.

NicolasDorier: we shouldn't do string manipulations to format dates when we have a library to do things properly.

NicolasDorier:
@TChukwuleta I think you only have to change `'YYYY-MM-DD hh:mm:ss'` to `'YYYY-MM-DD HH:mm:ss'`. This format is making sure excel recognize the cells as date.

PR #5931, opened 2024-04-15
Prevent payout double send

### LINKED ISSUE #5913: [Bug]: UX: Possible double sent on pull-payments/payouts

v1.13.0

docker

### What happened?

When doing pull-payments and having a payout processor enabled it could lead to double sending the transaction as after clicking button "approve + pay" the user is redirected to sign transaction screen where it would sign it a scecond time, the processor already sent the tx on click of the button in the previous step.

Problem description from Nicolas:

> The problem is: The automated payout processor will automatically pay from a hotwallet any payout that become approved. 
> 
> When the user click on the "Approve & Pay", he approves the payout, and thus the automated payout processor send the money. Then the user not noticing it, proceed with doing his own transaction.

1. have payout processor enabled
2. create a pull payment (add payout address)
3. in payouts: approve + pay the pull (sends first tx)
4. on sign tx screen, sign and broadcast (sends second tx)

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Fixes #5913.

- Pulls the check for payout processors out of the view into the controller.
- For the GET action is hides the "send/pay" options
- For the POST action it prevents going to pay after approving, though the hiding of those actions in GET should prevent accessing that anyways.

### DIFF
--- a/BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
+++ b/BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
@@ -7,8 +7,6 @@
 @model BTCPayServer.Models.WalletViewModels.PayoutsModel
 
 @inject IEnumerable<IPayoutHandler> PayoutHandlers;
-@inject PayoutProcessorService _payoutProcessorService;
-@inject IEnumerable<IPayoutProcessorFactory> _payoutProcessorFactories;
 @{
     var storeId = Context.GetRouteValue("storeId") as string;
     ViewData.SetActivePage(StoreNavPages.Payouts, $"Payouts{(string.IsNullOrEmpty(Model.PullPaymentName) ? string.Empty : " for pull payment " + Model.PullPaymentName)}", Context.GetStoreData().Id);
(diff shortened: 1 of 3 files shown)


BTCPayServer/Views/UIStorePullPayments/Payouts.cshtml
```diff
@@ -24,15 +22,16 @@
...
     {
         case PayoutState.AwaitingApproval:
+            if (!Model.HasPayoutProcessor) stateActions.Add(("approve-pay", "Approve & Send"));
             stateActions.Add(("approve", "Approve"));
-            stateActions.Add(("approve-pay", "Approve & Send"));
             stateActions.Add(("cancel", "Cancel"));
             break;
         case PayoutState.AwaitingPayment:
-            stateActions.Add(("pay", "Send"));
+            if (!Model.HasPayoutProcessor) stateActions.Add(("pay", "Send"));
```
Kukks, line 34: i dont think we should remove this one?

dennisreimann: But the payout processor might take over as well here, right?

NicolasDorier: it seems correct to me? if we allow people to pay while there is payout processor, then we run into the risk of paying twice?

dennisreimann: It depends on the config of the processor of how likely it is to interfere, but imho we should prevent that alltogether. WIth the changes here it's either you have a processor to take care or not.

Kukks (approved):

Issue #5819, opened 2024-03-08
Add support for 0-amount bolt11 invoices


pavlenex:
We do support amountless invoices for LNRUL-pay, there's no reason why we shouldn't add this functionality for Bolt11, especially since it's been widely adopted, and even apps like cashapp support it. It will help unify topup experience as well.

pavlenex:
@NicolasDorier I don't think #5943 resolves this. Does this mean that right now if the invoice is topup, we will show Bolt11 as a payment method on the checkout as well? If no, then this doesn't complete the issue.

PR #5943, opened 2024-04-23
fix bolt 11 0 amount check for payouts

Kukks:
If you specified a 0  amount bolt 11 invoice for a payout, it would be incorrectly validated and not accepted.


dennisreimann (approved):

NicolasDorier:
https://github.com/btcpayserver/btcpayserver/issues/5819

PR #5947, opened 2024-04-23
Branding updates for 2.0

### LINKED ISSUE #5953: Branding: Change File IDs to full URL
For the branding properties, which refer to files (e.g. logo and CSS), we save the file ID and generate the URL when needed. Examples:

- Set: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/UIStoresController.Settings.cs#L99
- Get: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Shared/LayoutHeadStoreBranding.cshtml#L5

This makes it incompatible with the API, because the API does not offer file upload and setters for the file IDs.

The most practical way to fix this would be to switch from saving the file IDs to the full URL. This way the URLs could be read and set via the API without the need to also provide file upload via the API.

dennisreimann:
### Public

- Remove deprecated `CustomCSSLink` and `EmbeddedCSS` properties from Payment Request, Pull Payment, Point of Sale and Crowdfund. They are replaced by the unified store branding functionality.

### Internal

- `...FileId` properties (which we use in conjunction with file uploads) are replaced with (relative) URLs, which directly reference the file. This was done to allow retrieval and editing via the API. Closes  #5953.


BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
```diff
@@ -145,16 +143,14 @@ public async Task<IActionResult> NewPullPayment(string storeId, NewPullPaymentMo
...
             {
                 Name = model.Name,
                 Description = model.Description,
                 Amount = model.Amount,
                 Currency = model.Currency,
                 StoreId = storeId,
-                PayoutMethodIds = selectedPaymentMethodIds,
-                EmbeddedCSS = model.EmbeddedCSS,
-                CustomCSSLink = model.CustomCSSLink,
+                PayoutMethodIds = selectedPaymentMethodIds
```
dennisreimann, line 153:
```suggestion
                PayoutMethodIds = selectedPaymentMethodIds,
```

NicolasDorier:
@dennisreimann can you document the breaking changes in the description of this PR?

I will then add it to https://github.com/btcpayserver/btcpayserver/issues/5964

BTCPayServer/Controllers/UIServerController.cs
```diff
@@ -1325,5 +1304,10 @@ public async Task<IActionResult> LogsView(string? file = null, int offset = 0)
 
             return View("Logs", vm);
         }
+
+        private async Task<string?> GetRelativeFilePath(string fileId)
+        {
+            return (await _fileService.GetFileUrl(new Uri("/"), fileId))?.Replace("file://", "");
```
NicolasDorier, line 1310:
```csharp
return (await _fileService.GetFileUrl(Context.Request.GetAbsoluteRootUri(), fileId))?.Replace("file://", "");
```

NicolasDorier:
1. This PR will not work as intended for people using `BTCPAY_ROOTPATH` other than `/` (default).
2. If a reference to a file was made, and then `ROOTPATH` is changed, all the links will be broken.
3. It is also lacking a test on the migration.

It isn't something that is easy to fix, as in the migrations, there isn't any access to the `Request.GetAbsoluteRootUri()`.

Ideally, the `FileService` should just provide relative paths (like `~/blahblah/blahblah.png`), and we would just save this in `LogoUrl`  and resolve the full path when the page is viewed. But this isn't how the API of FileService works and refactoring it is a can of worms.

There is I think two things we could consider:
1. Keeping the `..FileId` and removing the migration. If the FileId is set, the the URL is ignored.
2. Allows the URL to have a special format like `fileid:FILE_ID`, then resolve it when in the view.

I think `2.` is probably better, though the migration is still needed.

About the migration: I will check later on if I can't replace it with just a simple postgres query.

NicolasDorier:
Rebased

NicolasDorier:
@dennisreimann

1. I rewrote the test to directly insert into DB legacy data and migrate from that, and test all the properties impacted.
2. I rewrote the migration to be simple SQL queries, so we can remove obsolete properties, and not fear of memory issues on instances with lot's of stores.
3. Now, `cssUrl`, `logoUrl` accept `fileid:ID` which allow reference to a file id rather than a URL.

Two things I noticed:
1. I noticed that `paymentSoundUrl` can't be changed through the API. Should we add it in this PR?
2. The crowdfund page doesn't actually use the StoreBranding but the ViewModel does. Can we clean that up?

dennisreimann:
>  I noticed that paymentSoundUrl can't be changed through the API. Should we add it in this PR?

Done in c703a4c.

> The crowdfund page doesn't actually use the StoreBranding but the ViewModel does. Can we clean that up?

The [view uses it here](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Views/Shared/Crowdfund/Public/ViewCrowdfund.cshtml#L7) — it gets passed to the `LayoutHead` partial via the `ViewData`. I think I initially implemented it like this because the store branding is optional for the layouts and I didn't want to introduce a separate model for it.

PR #5959, opened 2024-04-26
Guard against running current master

Kukks:
With a longer release cycle for v2, we need to guard people from running master and corrupting their data. This adds a new requirement in that  a special config must be set when running master. We will remove when v2 rc is ready.


dennisreimann (approved):
Concept ACK, good idea!

PR #5962, opened 2024-04-29
Checkout: Fix copying numeric values to clipboard

### LINKED ISSUE #5960: Click to copy amount in SATS has spaces, should not.  1m sat - 1 000 000 , Some wallet like Sparrow do not want to receive spaces.
Clicking here
image

Result : 1 000 000
Expected : 1000000

dennisreimann:
Fixes #5960.


dstrukt (approved):
tACK.

Additional thought - is this the only place this would be applicable?

dennisreimann:
> is this the only place this would be applicable?

I think so.

Issue #5899, opened 2024-04-04
BTCPay Server 2.0: Megathread


NicolasDorier:
This is a mega thread to track everything that should be cleaned up for BTCPay Server 2.0:
Could use project milestone, but at this stage this is more of a brainstorm.

- [x] [The Big Cleanup: Refactor BTCPay internals](https://github.com/btcpayserver/btcpayserver/pull/5809/commits/11068605209b6807c3c86b0be1f18e589a8fa74f) 
```[tasklist]
### Deprecations and Cleanups
- [x] Remove Checkout v1 https://github.com/btcpayserver/btcpayserver/pull/5906
- [x] Cleanup the AddressInvoice table from non on-chain data (https://github.com/btcpayserver/btcpayserver/pull/5905)
- [x] Remove MySql/Sqltite depedency (https://github.com/btcpayserver/btcpayserver/pull/5910)
- [x] Remove any migrations code older than 2020 (https://github.com/btcpayserver/btcpayserver/pull/5939)
- [ ] #5744
- [ ] Remove Guest role
- [ ] Cleanup uppercase/lowercase inconsistencies in POS metadata and receiptData
- [x] Update status field of Invoices to not use legacy bitpay's status, but the BTCPay native one (https://github.com/btcpayserver/btcpayserver/pull/5982)
- [x] Decoupling PaymentMethodId from Payouts with PayoutMethodId https://github.com/btcpayserver/btcpayserver/pull/5944
- [ ] Rename *PaymentData to *PaymentDetails
- [ ] VACUUM (ANALYZE) as part of the migration to 2.0
- [ ] Cleanup data such as old expired invoices logs
- [x] Removing periodicity of pull payments (https://github.com/btcpayserver/btcpayserver/pull/5963)
- [ ] https://github.com/btcpayserver/btcpayserver/issues/5903
- [x] Remove docs on https://docs.btcpayserver.org/Development/Theme/#checkout-page-themes (https://github.com/btcpayserver/btcpayserver-doc/pull/1385)
- [ ] Onchain: Introduce pending transactions (#5743) as new concept/layer between transaction creation and signing
- [ ] Refactor onchain support to be similar to LN (connection string style, allows plugins to provide their own onchain tracker system instead)
- [ ] Cleanup all `[Obsolete]` declarations
- [x] Migration of Payouts
- [ ] https://github.com/btcpayserver/btcpayserver/issues/5485
- [ ] Restrict creation of new stores to Owner role
```

[...]

PR #5963, opened 2024-05-01
Remove period concept from PullPayment

NicolasDorier:
Remove the concept of periodicity in Pull Payments (Part of https://github.com/btcpayserver/btcpayserver/issues/5899)

Pull Payments were meant to allow a freelance to get paid every month by giving the ability for the patron to setup a pull payment that reset every month.

This feature has never been exposed in the UI, never asked, and poorly tested. As such, let's remove it for BTCPay 2.0.

### DIFF
--- a/BTCPayServer.Client/Models/CreatePullPaymentRequest.cs
+++ b/BTCPayServer.Client/Models/CreatePullPaymentRequest.cs
@@ -12,8 +12,6 @@ namespace BTCPayServer.Client.Models
         [JsonProperty(ItemConverterType = typeof(NumericStringJsonConverter))]
         public decimal Amount { get; set; }
         public string Currency { get; set; }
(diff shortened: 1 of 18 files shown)


TChukwuleta:
Hey @NicolasDorier, So no more pull payment from 2.0?

NicolasDorier:
It's maybe cool, but nobody use it, and nobody asked for it. It also make some future stuff harder to implement.

Also, we want really convert the concept of "PullPayments" to the concept of "Balances" (unsure we will rename it), but the basic idea is that those pull payments will be considered a bit like account that can be debited and credited.

In this new framework, the use case of periodicity will be replaced by "Recurring Deposit/Scheduled Deposit" features. This is easier to communicate and understand for a user than maintaining our own concept of periodicity.

TChukwuleta:
haa,.. Got it. 

Welldone @NicolasDorier

PR #5976, opened 2024-05-10
Recommended exchange to be resolved during Invoice Creation

NicolasDorier:
## Recommended exchange to be resolved during Invoice Creation

Previously, a store was required to select a specific exchange rate source during its creation.

We improved the user experience by automatically recommending the appropriate source when the user selects the currency. The selected recommendation was then saved in the store's configuration.

There were two problems with this approach:
1.  In https://github.com/btcpayserver/btcpayserver/issues/5653, we want to allow the server administrator to change the recommended exchange rate. Because the specific exchange rate source was set at the creation time of the store, a change in the recommended exchange would only affect stores created in the future. This means that if the exchange rate source becomes unreliable, most merchants will need to take manual steps.
2. If the merchant decides to switch the default currency, the specific exchange rate source would not change, resulting in a failure to create invoices.

This PR addresses these two issues by allowing the merchant to choose "Use the recommended exchange rate." Instead of selecting a specific exchange rate at the time of the store's creation, the "recommended exchange rate" is now selected at the time of the invoice's creation.

In a future PR (https://github.com/btcpayserver/btcpayserver/issues/5653), this will allow the administrator to easily change these if a rate source becomes unavailable and enables the merchant to switch his default currency without having to think about switching his exchange rate source.

To better reflect this change in meaning, in the store creation wizard and the rate configuration, `Kraken (Recommended)` has been replaced by `Recommendation (Kraken)`. This reflects that the choice is actually to adopt whatever recommendation is given when the invoice is created, and that the current recommendation is Kraken (for USD).

You can select this new option in the Greenfield API by simply setting the `PreferredSource` in the store's config to null.

## UX Changes

* Instead of `Kraken (Recommended)`, it should read `Recommendation (Kraken)`.
* `Recommendation (Kraken)` can now be selected in the rate configuration; previously, it was only selectable in the create store wizard.
* If you change your currency from `USD` to `JPY`, you should see that the rate of new invoices is determined by `bitbank` rather than `kraken`.

## Implementation

A store `PreferredSource` can now be null or empty string to take the current recommendation. If it is the case, the rate rules used when an invoice is created are:

```csharp
X_USD=kraken(X_USD)
X_JPY=bitbank(X_JPY)
...
```

If a specific exchange like `coinbase` is selected, this is instead used.

```csharp
X_X=coinbase(X_X)
```

## Pluginifiable Recommended exchanges

A plugin can now define its own recommended exchanges via `services.AddSingleton<DefaultRules>(new DefaultRules.Recommendation("USD", "kraken"))`.

`DefaultRule.Order` has been added to decide which rule should be given priority if there is a tie.


NicolasDorier:
@pavlenex @dstrukt do you think it makes sense UX wise?

dstrukt:
Logically, this makes a lot of sense, and sounds like it's more flexible than our current implementation. Would want to take a look at the UI to confirm, but sounds good to me AFAICT!

NicolasDorier:
@dstrukt in the current UI, the only change you really see is:

1. Instead of `Kraken (Recommended)`, you see `Recommendation (Kraken)`
2. `Recommendation (Kraken)` can be selected in the rate configuration (this was selectable only in the create store wizard previoulsy)
3. If you change your currency from `USD` to `JPY`, you should see that the rate is now solved by `bitbank` rather than `kraken`

dstrukt:
Oh I see, yes, that sounds pretty close to how I imagined it originally reading the PR. Sounds like great change!

cACK

Kukks (approved):
concept ACK, minimal code review but looks fine

PR #5979, opened 2024-05-11
Setting up defailt exchange provider

### LINKED ISSUE #5653: Server admin can set up default rate provider and default currency for stores on the server

TChukwuleta:
Resolves #5653, allowing admin to setup default exchange provider


NicolasDorier:
It will need a few changes and go after https://github.com/btcpayserver/btcpayserver/pull/5976, but I can take it from here.

I think we will need to talk to @dstrukt. This UX is simple, but lacking flexibility as it only allows to change the recommended exchange for the default currency.

On one side, maybe we don't really care about the lack of flexibility. An admin changing this setting will probably change it once in his life, as all his merchants should use the same default currency.

On the other side, it doesn't allow an admin to change the recommended exchanges for other currencies.

TChukwuleta:
I think the flexibility would be good one to have.

I have a follow up question, if an admin has merchants in different countries (say Nigeria and  Ghana), how would the currencies be managed if the intentions is for all the merchants to use the same default currency

NicolasDorier:
@pavlenex @dstrukt we will need some guidance here

So from https://github.com/btcpayserver/btcpayserver/pull/5976, it is possible to take whatever exchange is recommended for the selected currency. (rather than forcing the same exchange for all currencies)

This PR aims at allowing the admin to setup the recommended exchange for the default currency.

The good side is that this is simple to understand.

The bad side is that it doesn't allow the customization of recommended exchange for non default currency. 

I am tempted to keep it simple like this PR, as our original goal is to allow local admin to streamline the experience for their local merchant. (ie, for the default currency)

dstrukt:
Discussed on today's design call. Not sure if @dennisreimann is going to implement the additional dropdown solution we discussed, so will leave it to him add any additional context, or leave it to @TChukwuleta, but thought I'd provide an update.

TChukwuleta:
Thanks @dstrukt

BTCPayServer/Views/UIServer/Policies.cshtml
```diff
@@ -187,7 +187,11 @@
             <h4 class="mt-5">Customization</h4>
             <div class="form-group mb-3">
                 <label asp-for="DefaultCurrency" class="form-label"></label>
-				<input asp-for="DefaultCurrency" placeholder="@StoreBlob.StandardDefaultCurrency" class="form-control" currency-selection />
+                <input asp-for="DefaultCurrency" placeholder="@StoreBlob.StandardDefaultCurrenc" class="form-control" currency-selection />
```
dennisreimann, line 190:
```suggestion
                <input asp-for="DefaultCurrency" placeholder="@StoreBlob.StandardDefaultCurrency" class="form-control" currency-selection />
```

BTCPayServer/Views/UIServer/Policies.cshtml
```diff
@@ -274,5 +278,12 @@
                 element.classList.remove("show");
             }    
         }
+        function toggleExchangeProviderVisibility() {
+            var currencyInput = document.getElementById("currencyInput");
```
dennisreimann, line 282:
```suggestion
            var currencyInput = document.getElementById("DefaultCurrency");
```

dennisreimann (approved):
tACK. Ideally we'd also combine this with the rate provider recommendations, e.g. Kraken for EUR — but I think this is more complex and would involve some more JS.

BTCPayServer/Views/UIServer/Policies.cshtml
```diff
@@ -265,14 +269,21 @@
...
         for (let element of document.getElementsByClassName("collapse-on-js")){
-            if (element.classList.contains("show")){
+            if (element.classList.contains("show")) {
                 element.classList.remove("show");
             }    
         }
+        function toggleExchangeProviderVisibility() {
+            const currencyInput = document.getElementById("DefaultCurrency");
+            const exchangeDiv = document.getElementById("DefaultExchangeProviderForm");
+            exchangeDiv.classList[currencyInput.value.trim() !== "" || '@Model.DefaultCurrency' !== "" ? "remove" : "add"]("d-none");
```
NicolasDorier, line 280: There is a XSS vulnerability here. It needs to be wrapped with `@Safe.Json`

NicolasDorier:
There is a vulnerability, and it doesn't work for me.

I selected JPY and recommendation `Bitflyer`, but the recommendation is still `Bitbank`.

NicolasDorier (changes requested):
XSS vulnerability, and not actually changing the recommendation.

https://github.com/btcpayserver/btcpayserver/pull/5979/files/d331ec3f3f02d0ee84d61aae024cbd8177245780#r1658737631

TChukwuleta:
Going to look at this

BTCPayServer/Views/UIServer/Policies.cshtml
TChukwuleta: Wrapped it. You can also retry setting up a default exchange provider @NicolasDorier

pavlenex:
@TChukwuleta What's the status on this one?

TChukwuleta:
Hey Pav.. I got this resolved already. 

But let me work on the pipeline failure.

Sure totally agree to moving to 2.0.1

TChukwuleta:
Hey @NicolasDorier  Could you take another look at this and also test. Thanks

PR #5981, opened 2024-05-13
Fix: Error while connecting to websockets without reverse proxy

NicolasDorier:
When not using a web proxy, modern browser connects to the server throught HTTP/2.

A bug prevented browsers to connect to some of websocket routes, especially:
* The websocket refreshing the state of the checkout page when the invoice's status change.
* The websocket used for BTCPay Server Vault integration (Boltcard and Hardware wallet related)

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -1028,11 +1028,11 @@ namespace BTCPayServer.Controllers
             return Json(model);
         }
 
-        [HttpGet("i/{invoiceId}/status/ws")]
-        [HttpGet("i/{invoiceId}/{paymentMethodId}/status/ws")]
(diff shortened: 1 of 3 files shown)


dennisreimann (commented):
We should do this for the other hubs (e.g. Crowdfund) as well. 

Also commenting these cases would be good, I didn't know that `HttpGet` would be different than `Route`.

NicolasDorier:
@dennisreimann we don't need, because Crowdfund is using signalR for its route, which already does that.

Yeah, I needed to look into dotnet core source code to find out the cause of why it wasn't working...

In this loop, 
https://github.com/dotnet/aspnetcore/blob/0e7349e41a112a0c62f364ce9666a1a7dd4dc2fd/src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs#L94 if there is not `HttpGet`, then `metadata` is null and `needs405Endpoint` keeps `false`.

PR #5986, opened 2024-05-16
Search: Display text filters in search input

### LINKED ISSUE #5984: [Bug][UI]: Invoices could be permanently filtered via SearchTerm stored in cookies

I'm running BTCPay Server v1.13.1+ca55e1f30

I'm running BTCPay Server on dedicated VPS deployed with Docker

### What happened?

If you are filtering out Invoices list by clicking "Invoices" in Requests list - it goes into cookies and Invoices list became filtered forever

1. Install BTCPay server
2. Create store and wallet
3. Make some invoices
4. Make a Payment Request
5. Open that request and try to pay it, that should create invoices with Order Id linked to request - e.g. PAY_REQUEST_ID
6. Observe that now we have different invoices in Invoices list
7. Go to Requests page, click three dots at the end of the request row and click "Invoices"
8. Observe Invoices list which only contain invoices that linked to the request
9. Click anywhere else on the web UI and click again on the Invoices - you should return to the Invoices list which will be still filtered by the request with no obvious way to reset this filter

### Relevant log output

_No response_

### What browser do you use?

Chrome 124.0.6367.119

### Additional information

There are three ways to reset filter:
1. LogOut-LogIn
2. Put focus on search input in Invoices page, press Enter, that will bring params into URL, delete SearchTerm content in URL, press Enter to reload page
3. Delete SearchTerm field from cookies

dennisreimann:
This changes the search text input to also display the filters, which don't have a special UI (e.g. dropdown). Those filters (e.g. orderid) were not displayed before and hence could not be reset.

Fixes #5984.

grafik

### DIFF
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -1299,17 +1299,22 @@ bc1qfzu57kgu5jthl934f9xrdzzx8mmemx7gn07tf0grnvz504j6kzusu2v0ku
 
             filter = "status:abed, status:abed2";
             search = new SearchString(filter);
-            Assert.Equal("", search.TextSearch);
+            Assert.Null(search.TextSearch);
+            Assert.Null(search.TextFilters);
             Assert.Equal("status:abed, status:abed2", search.ToString());
             Assert.Throws<KeyNotFoundException>(() => search.Filters["test"]);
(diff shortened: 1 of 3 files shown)


dstrukt (approved):
Seems to work as intended AFAICT. 

Might not be a bad idea to get another pair of eyes on this, if we want to be extra safe

tACK from my end

PR #5992, opened 2024-05-20
Store Branding: Apply brand color to backend as well

dennisreimann:
For now we only apply the store branding on public pages — this adds it to the store's backend as well. Closes #5990.

colors

### DIFF
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -17,6 +17,7 @@ namespace BTCPayServer.Client.Models
         public string Website { get; set; }
 
         public string BrandColor { get; set; }
+        public bool ApplyBrandColorToBackend { get; set; }
         public string LogoUrl { get; set; }
         public string CssUrl { get; set; }
         public string PaymentSoundUrl { get; set; }
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -1622,6 +1622,7 @@ namespace BTCPayServer.Tests
                 CssUrl = "https://example.org/style.css",
                 LogoUrl = "https://example.org/logo.svg",
                 BrandColor = "#003366",
+                ApplyBrandColorToBackend = true,
                 PaymentMethodCriteria = new List<PaymentMethodCriteriaData>
             {
                 new()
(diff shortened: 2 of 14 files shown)


dstrukt (approved):
Looks good AFAICT.

tACK!

Another pair of eyes wouldn't hurt though

Few thoughts: 
- if the user chooses a yellow, it might look bad on Light mode, etc... but i guess the user will figure this out pretty quickly, which leads me to...
- Also realizing i didn't test, but does just removing a custom HEX value default back to the BTCPAY green? we should probably have a way to opt back into the default

dennisreimann:
> does just removing a custom HEX value default back to the BTCPAY green? 

Yes it does.

> if the user chooses a yellow, it might look bad on Light mode

Vice versa with dark brand color in dark mode. Not sure if we want to touch this — in 1b7718e I made an attempt at it and adjust the colors based on the brightness of the brand color. I think it covers it well enough to be safe, but I'm not sure if it's wanted. Yet it might be better than the status quo, in which there are brand color and theme combinations, which would not work.

dstrukt:
Ahh great - will take another look!

dstrukt (approved):
i think your solution handles the variability pretty well, we can always patch if we think of something better

tACK

dennisreimann:
@pavlenex This one is ready for user-testing and review.

dstrukt (approved):
Tested again, it seems it works on all variants and attempts I could throw at it.

My only reservation is if a user might not want their branding colors applied to the back-end too. I can see how you might want your public facing pages be your brand color, etc.. but you don't necessarily want your admin UI to match that too, esp. considering the number of minor UI edge-cases, etc.. it *could* introduce, that we work so hard to keep consistent.

Arguably we could make applying it to the back-end an additional toggle, but I'll save that for tomorrow's call!

pavlenex (approved):
tACK wasn't able to break it.

pavlenex:
@NicolasDorier I think we can merge this one for 2.0.0

dstrukt:
- Add additional context /descriptor text around what items will be affected
- Add toggle that allows the admin to choose if the branding applies to back-end and front-end as well

dennisreimann:
Here we go.

grafik

pavlenex:
lgtm, will leave to @dstrukt to do final copy review.

NicolasDorier:
@dstrukt I will push to the alpha, let dennis know if there is something missing.

PR #6004, opened 2024-05-23
Add MWEB support

hectorchu:
Changes necessary to support MWEB.


losh11:
For anyone who is seeing this pull request, MWEB is a new privacy/fungibility feature added to Litecoin - based on MimbleWimble, which itself is based on Confidential Transactions. We hope that this PR is merged into btcpayserver. This would allow stores to accept private transactions.

NicolasDorier:
Just a note: From our master branch, we are working on BTCPay 2.0.

If you want to add altcoin support, you need to add it as a plugin.
If BTCPay Core is lacking something for your altcoin to work, you will need to create the necessary abstraction which isn't tied to your altcoin directly, and then use it in the plugin.

The code to do what you did via a plugin is essentially the same, outside of the fact that you can't modify files in btcpayserver core.

hectorchu:
@NicolasDorier The change to the BTCPayWallet.ReserveAddressAsync method, how would you reconcile that? Either a hook inside that method, or an ability to override that method in a derived class and allow a way to set a new instance in the BTCPayWalletProvider? I think I like the idea of allowing BTCPayWallet to be derived from and selectively overriding methods.

NicolasDorier:
@hectorchu I think subclassing can be a strategy yes.
What I would advise you is:
1. Try to develop it directly inside BTCPay
2. Make sure that everything belonging to MWEB is inside a system plugin (Check in the Plugins folder)
3. When things work, isolate the changes you did outside of your plugin's folder and propose a PR to inject what you need in core
4. Move your plugin's folder in its own plugin repository.

This is probably the easiest way to get a V1 running.

hectorchu:
I developed a plugin, so closing.

PR #6008, opened 2024-05-25
User: Add name and image URL

dennisreimann:
More personalization options, prerequisite for btcpayserver/app#3.

Additionally:
- Remove ambigious and read-only username from manage view.
- Improve email verification conditions and display.
- Greenfield: Update current user. Prerequisite for btcpayserver/app#13.

grafik


dstrukt (changes requested):
Amazing, functional tACK!

Nit: 

- [ ] Rename `Image` label to `Profile Picture` (my preference) or `Profile` label

- [ ] How are we handling horizontal, non square 1:1 images? Think a 1:1 might be a little more clean, compared to what we're currently doing:

Screen Shot 2024-05-26 at 12 12 53 PM

I was curious how you handled the case in which there is no profile picture (mostly for the app to have a default filler image, which I think would look best), but I do like that it's hidden until you upload an image. Having a default image for the mobile app could be a nice edge case scenario. Regardless either work, just a thought as I review.

dstrukt (approved):
tACK - LGTM!

Thanks for the changes!

louieo:
My first PR review here as you suggested @dennisreimann. tACK - the changes look good to me!

Screenshot 2024-06-13 at 1 43 40 PM

I did however try uploading a larger image (1.3MB jpg) and my profile pic was not set and there was no error message. Is the file size limited and should there be an error message displayed if above max size?

dstrukt:
> I did however try uploading a larger image (1.3MB jpg) and my profile pic was not set and there was no error message. Is the file size limited and should there be an error message displayed if above max size?

Great catch, I'd think we should probably limit the file size or reduce the file size / dimension upon upload for specific assets (profile picture, logos, etc..)!

dennisreimann:
Thanks, the validation was present, but the check for whether the state is valid had to be moved. Fixed it.

grafik

NicolasDorier:
@dennisreimann please look my two commits

NicolasDorier:
Note for [6041](https://github.com/btcpayserver/btcpayserver/pull/6041): you are using MailboxAddressValidator to validate in the controller.
However, this validate addresses in this form: Firstname Lastname <blah@example.com> as well as blah@example.com.
I am not sure what happen if you attempt to use Firstname Lastname <blah@example.com> from the API. (The UI has an additional validator at the VM side)

BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
```diff
@@ -139,7 +139,7 @@ public async Task<IActionResult> UpdateCurrentUser(UpdateApplicationUserRequest
             if (User.Identity is null || user is null)
                 return this.CreateAPIError(401, "unauthenticated", "User is not authenticated");
 
-            if (!string.IsNullOrEmpty(request.Email) && !MailboxAddressValidator.IsMailboxAddress(request.Email))
+            if (!string.IsNullOrEmpty(request.Email) && !request.Email.IsValidEmail())
```
NicolasDorier, line 142: cool, forgot we had this extension method

PR #6023, opened 2024-05-31
Bumping LND to 0.18.0-beta

rockstardev:
Updating our devenv to use LND 0.18.0

Used it in BTCPayServer.Lightning library and it's good to go, passed all the tests:
https://github.com/btcpayserver/BTCPayServer.Lightning/pull/161

### DIFF
--- a/BTCPayServer.Tests/docker-compose.altcoins.yml
+++ b/BTCPayServer.Tests/docker-compose.altcoins.yml
@@ -226,7 +226,7 @@ services:
       - "5432"
 
   merchant_lnd:
-    image: btcpayserver/lnd:v0.17.4-beta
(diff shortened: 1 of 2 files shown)


warioishere:
are there any breaking changes because lnd 0.18.0  and btcpay invoice storage because lnd now gives possibilty for own sql invoice dbs?
will be the first on to test if its out.

Thanks for progressing so fast @rockstardev

rockstardev:
@warioishere I haven't updated LND to use the new SQL invoice storage option, so there shouldn't be breaking changes.

rockstardev:
I've tested it on my dev-env and server extensively, works well, merging to get it in for 1.13.2 release

PR #6027, opened 2024-06-03
Cancel shopify order when invoice payment fails

### LINKED ISSUE #6021: When an invoice expires, the corresponding Shopify order remains pending instead of canceling.
> When an invoice expires, the corresponding Shopify order remains pending instead of canceling.

This has been reported by a user, but we need to try to replicate it and if we manage to replicate it, then we need to address it for 1.13.x as it's tricky UX issue that can potentially mess up the fulfilment for certain type of merchants.

Deployment method: Voltage

Kukks:
fixes #6021


pavlenex:
@ndeet any chance you can deploy and test this one? It's the blocker for 1.13.x and we should prioritize it 🙏

PR #6042, opened 2024-06-07
Fix Monero local dev docker start up

napoly:
This is a follow up for https://github.com/btcpayserver/dockerfile-deps/pull/92.


napoly:
Fixing the https://docs.btcpayserver.org/Development/LocalDevelopment/#altcoin-support-development. This works locally for me. Only problem I see is the (unhealthy) sign next to monero_wallet when doing `docker ps` since the health-check is monerod specific (port 18081).. so I removed it

Kukks (approved):

napoly:
Now watch out here.. as I had to re-base my branch before the commit 88fe28df8db3ef5aff652fbbf710aa7707b34b18 to make this build pass.. 

@Kukks can you please double check here.. it seems `btcpayserver/lightning:v24.05` is unstable.

PR #6045, opened 2024-06-10
Fix null pointer exception on receipt print page

dennisreimann:


rockstardev:
@dennisreimann should we also change the reference on line 116 from `Model.AdditionalData` to `Model.CartData`?

dennisreimann:
> should we also change the reference on line 116 from Model.AdditionalData to Model.CartData?

No, that one is ok and the receipt might contain additioonal data, see https://github.com/btcpayserver/btcpayserver/blob/v1.13.x/BTCPayServer/Controllers/UIInvoiceController.UI.cs#L252

rockstardev:
OK, I've tested it on my local machine and it works as expected:

image

I see you're adding tests for the receipt so let's go with your version. There were also 2 small problems with Fast tests, we can either fix now or open a new issue:

1. https://wasabiwallet.io was returning forbidden
2. CheckJsContent was failing for versions
image

dennisreimann:
Let's do the fast test fixes separately.

rockstardev (approved):
LGTM, tests now in separate issue #6046

rockstardev:
Hah, wait, seems that new test is failing 😂

rockstardev (changes requested):
Let's merge it after this test fix: 
image

PR #6050, opened 2024-06-12
Notificationlist

### LINKED ISSUE #3871: Improved Notifications List View
For the developer who picks up this Issue, feel free to reach out to me on Mattermost @dstrukt for any feedback, additional mocks, or discussion of specific items!

Few things involved in this improvement, but all the changes should be consistent with what we're doing in the Invoice / Payment Request views already, so that could act as a reference for code / implementation:

- [x] Mass Action buttons that replace the table header when `> 1` are selected
- [x] Introduction of the `Store` from which the notification originated to the table column ... might just be adding more noise though, but an option, thoughts?
- [x] Ability to filter based on the `Store`, regardless if we add to the column
- [x] A `Read/Unread` or `Seen/Not Seen` toggle, would be similar styling to the `Store Invoices / All Invoices` in the Invoice List View, but this is also an optional choice .. thoughts?
- [x] Filter by the `type` e.g. `Invoice`, `Payment Request`, etc.. similar functionality to the filtering in other views

.gif might take a second load

Notifications+MassActionsV0 1

TChukwuleta:
Resolves #3871 

@dstrukt did Include the store column, as not all notifications are associated with a store.

But outside of that, open to review.


dstrukt (changes requested):
Overall, this looks great! Awesome work!

# General

- [ ] let's kill the `ms-3` class on all of the buttons
- [ ] we could increase the `gap-3` to `gap-4` at the parent form level

these spacing changes could help solve some of the problems below

- [ ] For the types, let's combine the User related ones to `User Updates`
- [ ] combine Payout+External Payout -> `Payouts`
- [ ] `New Version` to `BTCPay Server Updates` or `Server Updates`
- [ ] `Plugin Update` to `Plugin Updates`

# Mobile

- [ ] Ideally the tabs sit flush to the left
- [ ] Also, it would be nice if everything sat on the same line
Screen Shot 2024-06-12 at 4 53 45 PM

- [ ] Some things get cut off on the tablet breakpoint, not sure what we can do about this?
Screen Shot 2024-06-12 at 4 54 09 PM

- [ ] there's an in-between state between tablet and mobile where the options are hidden, minor, but worth noting

Screen Shot 2024-06-12 at 5 46 01 PM

Otherwise, I think we're good to go!

@dennisreimann It'd be great to get another pair of eyes on this, and @pavlenex it'd be great to get a functional review from you!

TChukwuleta:
Alright. Will update later accordingly

Thank you

BTCPayServer/Controllers/UINotificationsController.cs
```diff
@@ -1,20 +1,26 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net.WebSockets;
 using System.Threading;
 using System.Threading.Tasks;
+using Amazon.Runtime.Internal.Transform;
```
dennisreimann, line 7: This might have slipped in.

dennisreimann (commented):
Code LGTM so far. The addition of the store ID to the notifications will be very helpful for the app as well.

dstrukt:
> * The All/Unread switch could have the primary class for highlighting instead of secondary. There's no CTA on that page and imho that toggle has the most important functionality.

Completely agree, and was going to do a follow up design component issue, but if we can tackle here, that'd be great!

TChukwuleta:
made the update

dstrukt (changes requested):
Thanks for the changes, it looks excellent! All of my requested changes look good - last 2 nits and we're good! Great work! 

- [x] let's make the active tab our `primary` green / variable, not the darker green

Screen Shot 2024-06-14 at 4 38 37 PM


- [ ] can we kill the `gap-4` class on mobile breakpoint to tighten up the height?

Screen Shot 2024-06-14 at 4 39 50 PM

Otherwise, this looks good to go!

Requested re-review from @dennisreimann in case he has anything else.

dstrukt:
Tested the latest, DM'd a quick fix, will approve after!

dstrukt (approved):
tACK - LGTM

dennisreimann (approved):
tACK, good job!

dstrukt:
Was this a new review request @pavlenex? Or due to the potential merge conflicts from the sidebar PR? cc @dennisreimann

dennisreimann:
Will rebase this and update with the changes from the sidebar/subnav PR tomorrow.

dennisreimann (approved):

NicolasDorier:
Seems good, @TChukwuleta can you squash the commits, and put a proper commit message?

PR #6051, opened 2024-06-13
Greenfield: Refactor app endpoints

dennisreimann:
- Do not change unset data
- Clean up difference between request (template) and data (items/perks)
- Add missing properties (form id and custom tip percentage)
- Update docs


Kukks (approved):

dennisreimann:
@Kukks I changed a bunch of stuff can you re-review?

NicolasDorier:
@dennisreimann 

> Do not change unset data

All single API point do change unset data. This is the nature of PUT (as opposed to PATCH that we do not support)
Why are apps the only different?

dennisreimann:
@NicolasDorier I've reverted the ToPointOfSaleSettings/ToCrowdfundSettings changes, so that it works as before and all data needs to get passed.

PR #6055, opened 2024-06-17
POS: Keypad error should sit below the header

### LINKED ISSUE #6054: Keypad error should sit below the header
Noticed while testing, probably an edge case, but imagine this is consistent across both platforms, and in case there's an error, we should handle it better!

Screen Shot 2024-06-13 at 11 38 38 AM

dennisreimann:
Fixes #6054.

grafik


dstrukt (approved):
tACK.

In the future, we may want to add a background to the header bar as well, but let's cross that bridge if/when we need to.

PR #6061, opened 2024-06-20
Handle LNURL Payouts failing due to amount restriction

Kukks:
Payouts to LNURLS that had an amount restriction that did not match the payout amount would get stuck retrying. This cancels them once it happens.

### DIFF
--- a/BTCPayServer/Data/Payouts/LightningLike/LightningLikePayoutHandler.cs
+++ b/BTCPayServer/Data/Payouts/LightningLike/LightningLikePayoutHandler.cs
@@ -89,7 +89,11 @@ namespace BTCPayServer.Data.Payouts.LightningLike
                 {
                     using var timeout = new CancellationTokenSource(TimeSpan.FromSeconds(30));
                     using var t = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, cancellationToken);
-                    var info = (LNURLPayRequest)(await LNURL.LNURL.FetchInformation(lnurl, CreateClient(lnurl), t.Token));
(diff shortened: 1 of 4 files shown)


BTCPayServer/Data/Payouts/LightningLike/LightningLikePayoutHandler.cs
```diff
@@ -159,9 +163,27 @@ public Task BackgroundCheck(object o)
...
 
-        public Task<decimal> GetMinimumPayoutAmount(IClaimDestination claimDestination)
+        public async Task<decimal> GetMinimumPayoutAmount(IClaimDestination claimDestination)
         {
-            return Task.FromResult(Money.Satoshis(1).ToDecimal(MoneyUnit.BTC));
+            if(claimDestination is LNURLPayClaimDestinaton lnurlPayClaimDestinaton)
+            {
+                try
+                {
+                    var lnurl = lnurlPayClaimDestinaton.LNURL.IsValidEmail()
```
NicolasDorier, line 172: I have no clue about what is going on here

NicolasDorier: @Kukks Can't you just fix that by making the `LNURL.LNURL.Parse` smart enough to handle the case when this is a email?

rockstardev: Previously GetMinimumPayoutAmount would always return 1 BTC, now if `claimDestination` is `LNURLPayClaimDestinaton` it queries it to see if `MinSendable` is set.

Just reading the logic of code everything seems fine to merge.

rockstardev (approved):

PR #6071, opened 2024-06-28
bump selenium container

NicolasDorier:
May solve the flaky `OpenQA.Selenium.WebDriverException : Expected to read a START_MAP but instead have: END. Last 0 characters read: ` error we have.

https://github.com/SeleniumHQ/docker-selenium/issues/1507


dennisreimann (approved):
utACK. Let's hope this solves this annoying error :)

PR #6075, opened 2024-07-01
Greenfield: Add file endpoints and upload

### LINKED ISSUE #6074: Add FileUpload to Greenfield API
For profile photo  to be uploadable on the app side, we'd have to expose the File uploading capability in the Greenfield API.

dennisreimann:
- Endpoints for server files
- File upload using `multipart/form-data`

Closes #6074.

Can also be tested using cURL:

- `curl --location 'https://localhost:14142/api/v1/files' --header 'Authorization: token MY_API_TOKEN' --form 'file=@"LOCAL_FILEPATH"'`
- `curl --location 'https://localhost:14142/api/v1/users/me/picture' --header 'Authorization: token MY_API_TOKEN' --form 'file=@"LOCAL_FILEPATH"'`


BTCPayServer/UnresolvedUri.cs
```diff
@@ -1,9 +1,12 @@
+#nullable enable
 using System;
 
 namespace BTCPayServer
 {
-    public record UnresolvedUri
+    public record UnresolvedUri(string? FileId)
```
NicolasDorier, line 6: Not all `UnresolvedUri` have a FileId.
Just use `new UnresolvedUri.FileIdUri(fileId)`.

dennisreimann: I did this to capture the FileId for the `public string? GetFileId() => string.IsNullOrEmpty(FileId) ? null : FileId;` below, because I wanted a way to access the FileId later on without parsing the URI.

NicolasDorier: `var fileId = (record as FileIdUri)?.FileId` then check if `fileId` is null.

PR #6079, opened 2024-07-02
Receipt page fixes

dennisreimann:
Quick fixes for various things we came across while testingh the BTCPay Day 2024 ticket process.

- Don't assign empty values to data; hide present empty values
- Use same URL on "Return to Store" link as on invoice/checkout page

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -249,10 +249,12 @@ namespace BTCPayServer.Controllers
                         receiptData.Remove(key);
                     }
                 }
-                // assign the rest to additional data
+                // assign the rest to additional data and remove empty values 
                 if (receiptData.Any())
                 {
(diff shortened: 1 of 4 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -214,6 +214,7 @@ public async Task<IActionResult> InvoiceReceipt(string invoiceId, [FromQuery] bo
                 InvoiceId = i.Id,
                 OrderId = i.Metadata?.OrderId,
                 OrderUrl = i.Metadata?.OrderUrl,
+                StoreUrl = i.RedirectURL?.AbsoluteUri,
```
NicolasDorier, line 217: Please rename to `RedirectUrl` this isn't the StoreUrl.

dennisreimann: Rebased and updated :)

PR #6081, opened 2024-07-03
Checkout: Add glow gradient to spice it up

dennisreimann:
Adding this as a draft, so that we don't forget about this idea as a potential 2.0 candidate.

light

dark


pavlenex:
Don't really like this suggestion, don't see why we should change aesthetics of this, if there wasn't any functional feedback, since it's such a subjective decision,  leaving it up to @dstrukt.

NicolasDorier:
can't people do this with a custom css?

dennisreimann:
Agreed, people can customize it and we should keep it clean and simple :)

Issue #6065, opened 2024-06-23
[Bug]: Reporting RAW Data View - 2 issues


petzsch:
### What is your BTCPay version?

BTCPay Server v1.13.3+9d84ec4aa 

### How did you deploy BTCPay Server?

Docker

### What happened?

I've clicked on the Reporting link (left menu) and got this:
![grafik](https://github.com/btcpayserver/btcpayserver/assets/1374810/47257f67-d282-482b-890f-eecb8ecc6b8d)

Here are the issues I see:
- default sorting: by Payment Type and then Date -> should be just Date I think
- the shortend label for "BTC-LightningNetwork" is showing "BTC-LightningNe … ightningNetwork" which is kind of unexpected

### How did you encounter this bug?

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox 127.0.1 on Win11

### Additional information

none

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

NicolasDorier:
I removed the text-truncate column from Payment Id, and fixed the bug on sorting.

@dennisreimann will also make a PR for fixing the `text-truncate` components.

PR #6087, opened 2024-07-04
Truncate Center Component: Fix Vue rendering

dennisreimann:
Context: #6065. Had to use `Safe.Raw` instead of `Safe.Json` to render the comparison, which contains an `>` that otherwise would have been escaped.


NicolasDorier:
Mixing vue-js and Razor template is a minefield of vulnerability, please avoid doing this at all cost.
Problem is that a Razor property can inject something like `{{  alert("booh") }}`, which it think it is safe, but then Vue-JS will happily execute the javascript, even if untrusted.

NicolasDorier:
Ok so `v-text="window.alert('lol')"` will show open a poppup. As such anytime you are injecting into VueJS from Razor, you have an XSS.

NicolasDorier:
I simplified the component. I don't know why `Safe` is used in this context, it doesn't prevent injections in Razor over VueJs scenario.

dennisreimann:
Checked, thanks for the additions!

PR #6088, opened 2024-07-04
POS: Improve padding on mobile and unify card look with tiles

dennisreimann:
On mobile, the description content was lacking horizontal padding. This adjusts it while also unifying the cards to work like the tiles on checkout: Below 400px width, we pull the to the edges of the screen, which makes it looks nicer and display better than as if they'd also have an outer margin.

Adjustments take effect on all POS view variants.

grafik


dstrukt (approved):
tACK - LGTM, nice little change!

PR #6093, opened 2024-07-05
Greenfield: Add store id for notifications

dennisreimann:
Follow up to #6050, which adds the store IDs for notifications to Greenfield. This will be of use for the app too.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.Notifications.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Notifications.cs
@@ -9,7 +9,7 @@ namespace BTCPayServer.Client;
 public partial class BTCPayServerClient
 {
     public virtual async Task<IEnumerable<NotificationData>> GetNotifications(bool? seen = null, int? skip = null,
-        int? take = null, CancellationToken token = default)
+        int? take = null, string[] storeId = null, CancellationToken token = default)
     {
         var queryPayload = new Dictionary<string, object>();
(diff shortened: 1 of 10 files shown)


BTCPayServer/PayoutProcessors/BaseAutomatedPayoutProcessor.cs
```diff
@@ -111,29 +110,30 @@ private async Task Act()
...
+                    Stores = [PayoutProcessorSettings.StoreId]
                 }, context, CancellationToken);
 
             await _pluginHookService.ApplyAction("before-automated-payout-processing",
                 new BeforePayoutActionData(store, PayoutProcessorSettings, payouts));
             if (payouts.Any())
             {
-                Logs.PayServer.LogInformation(
-                    $"{payouts.Count} found to process. Starting (and after will sleep for {blob.Interval})");
+                Logs.PayServer.LogInformation("Found {PayoutsCount} payouts to process. Starting (and after will sleep for {BlobInterval})", payouts.Count, blob.Interval);
```
NicolasDorier, line 124: Remove debug logs like here, it really spams

dennisreimann: Dropped the debugging commit. That test failed several times here and I wanted to see why — but it worked on the commit that added the logging statements 🙈

PR #6097, opened 2024-07-09
Fix missing normalization of Settled invoice status

dennisreimann:
Addition to #5982, which fixes lookups for invoices with status `Settled` (e.g. for Sales tiles on the Dashboard).


NicolasDorier:
Good catch!

PR #6103, opened 2024-07-11
Greenfield: App endpoints for sales statistics

dennisreimann:
New Greenfield endpoints to bring the POS statistics to the mobile app's dashboard.


dennisreimann:
Tested with the app :)

grafik

Issue #6105, opened 2024-07-12
[Bug]: Shopify Integration: BTCPay invoice not generated until user clicks the button 


beeforbacon:
### What is your BTCPay version?

This issue is written on behalf of a Voltage user from uncute.com having issues with the shopify checkout.

BTCPay version 1.13.3
Running BTCPayServer from Voltage service.

The User says: 
Main issue: BTCpay invoice is not created until users clicks on the button injected by the js in the order status page.
- Desirable solution: Triggering the creation of the invoice as soon as the order is placed using Webhook. This way it's not dependent on end users's UI.
- Easy solution: in the script added to order status page, call function getOrCheckInvoice when the scripts load (get request to trigger invoice creation).

Worked with @pavlenex and customer on July 12, 2024 conference call.

### How did you deploy BTCPay Server?

2

### What happened?

.

### How did you encounter this bug?

.

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

pavlenex:
Feedback from the user:
 Main issue: BTCpay invoice is not created until users clicks on the button injected by the js in the order status page.
- Desirable solution: Triggering the creation of the invoice as soon as the order is placed using Webhook. This way it's not dependent on end users's UI.
- Easy solution: in the script added to order status page, call function getOrCheckInvoice when the scripts load (get request to trigger invoice creation).

PR #6109, opened 2024-07-15
Shopify: Create invoice when the payment page opens

NicolasDorier:
Fix  https://github.com/btcpayserver/btcpayserver/issues/6105

This will generate the invoice as soon as the page for payment in shopify open.


Bojo75 (commented):
I don't know how to confirm my deposit I've been trying to get help but just keep going around in the same circles

PR #6118, opened 2024-07-25
Uniform primary action id's of pages

NicolasDorier:
Since our UI is more coherent, I believe we can simplify a bit our tests by naming coherently the primary action of a page across the whole project.


dennisreimann (approved):
Concept ACK. Good idea :)

PR #6119, opened 2024-07-25
Add parts of the UI to translate

NicolasDorier:
This cover additional strings in the UI:

I also added a cheat mode to translate all the strings to "OK" so we can easily spot missing one.

image


dennisreimann (commented):
tACK. Follow up in #6120.

Only remark: Why "OK" and not something like "%TODO%"?

PR #6122, opened 2024-07-26
Wallet: Generate receive address automatically

dennisreimann:
This circumvents landing on a blank page with only the "generate address" button and automatically generates a new address, unless the Unreserve action was used.

This was discussed in yesterdays design meeting.


dstrukt:
Nice!

Will try to get some testing in today.

dstrukt (approved):
tACK - Works as expected!

NicolasDorier:
It seems a bit weird to me that you can click on "Unreserve" then have the screen without QR Code.

What about either of those?
1. Would it be better to just redirect to the transaction list with a success alert
2. Just remove the "Unreserve" feature.

I am not super sure Unreserve is really needed.

dstrukt:
Fair.

Unless there's a reason I might not aware of, seems like option 2, removing the Unreserve feature would make the most sense?

dennisreimann:
I'm fine with both, I just didn't want to remove it straight away and if @NicolasDorier thinks this would be fine too, we can remove it.

dstrukt:
Will re-review the latest changes this evening.

dstrukt (approved):
tACK - LGTM!

PR #5828, opened 2024-03-12
HTML lang setting and Head tags for POS and Crowdfund public pages

Nisaba:
Modifications allowing the user to improve the SEO of their pages

PR #6134, opened 2024-08-04
HTML lang setting and Head tags for POS and Crowdfund public pages

Nisaba:
Modifications allowing the user to improve the SEO of their pages


Nisaba:
#5828

PR #6150, opened 2024-08-15
Handle password reset when SMTP isn't configured or validated

### LINKED ISSUE #5597: Handle Reset password link on login when Server's SMTP is not configured
**Context:**
Often time users register to a server which doesn't have email smtp configured, which then leads them into coming to our chat asking they're not receiving email reset password upon submitting the form:

**Solution:**
There are two possible solutions to this:
1. We hide the password link the server doesn't have SMTP configured. It's easy to implement, but still doesn't solve a problem for users and can potentially lead them into coming to chat and asking how to recover the account.
2. Instead of hiding link, we show a different view when user clicks `forgot password`. That view ideally could give a bit more context. 

_Email password reset functionality is not configured for this server. Please contact the server administrator to assist with account recovery. If you are the administrator, please follow these steps to [configure email password resets](https://docs.btcpayserver.org/Notifications/#smtp-email-setup) or reset your admin password through [command line](https://docs.btcpayserver.org/FAQ/ServerSettings/#forgot-btcpay-admin-password)_

So this is not yet to be implemented, let's discuss how we can acheive this.

TChukwuleta:
Resolves #5597 

This PR contains:

- Redirection to a new view if the Email SMTP isn't configured, or if the  smtp configuration cannot be authenticated when a user is attempting password reset.

- Server admin privileges to reset user's password.


dennisreimann (approved):
tACK

dstrukt (approved):
Will double check we covered all tests in this week's design call, but think this is a tACK from me as well!

NicolasDorier:
@TChukwuleta I fixed a merge conflict and simplify a little bit for a case that shouldn't happen.

Can you write a selenium test "CanManageUsers"?

TChukwuleta:
Hey @NicolasDorier  I have added the test

PR #6169, opened 2024-08-26
Update notificaiton counter record

### LINKED ISSUE #6165: Notification counter updates, but the counter doesn't
When receiving a notification, the counter updates, but the actual content (invoice) remains static and requires a page refresh to get it.
Screenshot 2024-08-25 at 10 52 05

TChukwuleta:
Resolves #6165 

Noticed that the counter does update, but the timing doesn't. 

Included Key attribute and status update to force re-rendering of the razor file.


dennisreimann (approved):
tACK.

Issue #5903, opened 2024-04-05
Removal of some DB column that aren't needed


NicolasDorier:
```[tasklist]
### Tasks
- [x] InvoiceEvents contains an unused 10 bytes UniqueId (Cleanup data from the InvoiceEvents table #5904), which consume a lot of space.
- [x] Cleanup AddressesInvoices table (Cleanup AddressesInvoices table #5905)
- [x] Remove CustomerEmail, ItemCode, OrderId, Archived? from Invoices table
```

pavlenex:
@NicolasDorier did you finish this one? Or is that still something remaining, if it's finished let's close it? If there's more we can add it to 2.1.0?

PR #6170, opened 2024-08-26
Delete columns CustomerEmail, OrderId, ItemCode

NicolasDorier:
All of this data is already in the invoice's metadata.
The columns were duplicating the metadata in order to index it, but as we only support postgres we can just use a partial index for `OrderId` and `ItemCode`.

Removing this duplication mean we don't risk potential bug where the two values aren't sync'd. We are also saving space as the new index is partial.

https://github.com/btcpayserver/btcpayserver/issues/5903


BTCPayServer.Data/ApplicationDbContextFactory.cs
```diff
@@ -4,20 +4,25 @@
...
         {
+            LoggerFactory = loggerFactory;
         }
 
+        public ILoggerFactory LoggerFactory { get; }
+
         public override ApplicationDbContext CreateContext(Action<NpgsqlDbContextOptionsBuilder> npgsqlOptionsAction = null)
         {
             var builder = new DbContextOptionsBuilder<ApplicationDbContext>();
+            builder.UseLoggerFactory(LoggerFactory);
```
NicolasDorier, line 25: This makes it possible to log entity framework requests to the console in debug.

Issue #6167, opened 2024-08-25
Sales report on dashboard doesn't change values and shows random number of sales


pavlenex:
The sale report on dashboard showscases really inconsistent data, and on top of that doesn't reflect the change when you switch from monthly to weekly and vice-versa, where as on the right hand size

https://github.com/user-attachments/assets/b1ab9eed-1445-44a0-b0f4-b81092f266a0

umarcodehub:
I noticed the issue #6167 regarding the inconsistent sales report on the dashboard. I’m interested in tackling this and would like to help resolve it. Could you please assign this task to me?

TChukwuleta:
Hello @umarcodehub  Sure, you can go ahead and work on this.

umardevX:
@TChukwuleta  Can you share more info about this. I successfully run the project and signed in as administrator but i'm unable to see the screen which you showed above in the video.

TChukwuleta:
Okay. So first you'd need to create a new Point of sale.. 

And then make a successful transaction via POS. Once that is done, you'd see the chart appear on your dashboard @umardevX

umardevX:
@TChukwuleta is there any documentation for this whole process?

TChukwuleta:
Have you been able to create a Point of Sale? If yes, can you see your point of sale application?

umardevX:
Yes

umardevX:
Because when i go on my POS add a item to cart and tap to Pay in wallet it takes me to the blank next tab of browser.

TChukwuleta:
For now.. select fake a payment (not pay in wallet) and then, mine... 

Once both processes are successful, go back to your dashboard, you'd be able to see the chart in question.

For documentation visit: https://docs.btcpayserver.org/

umardevX:
Yes i got it thankyou

umardevX:
@TChukwuleta  if i have to access database how i can do that?

TChukwuleta:
Postgres is setup with docker. So you can access it with postgres. 

However I dont think you need database to resolve this

umardevX:
@TChukwuleta Write now i added a lot sales from POS and it's not reproducing the error i just check if i can add sales in database on previous dates may be it can work then

umardevX:
@TChukwuleta  Initially i created a branch related to this fix after cloning the code and added my changes and commit that changes. But whenever i go to push that branch to master i got error of 403. I'm new to contributions so if you can help me with that so i can learn.

[...]

PR #6176, opened 2024-08-28
Sales report dashboard issue #6167 

umardevX:
querySelectorAll returns multiple elements that leads to not updating the required element innerText directly, so i change querySelector that will directly update our .sales-count element innerText.

Please review the changes and let me know if any additional modifications are required.


dennisreimann (approved):
utACK

PR #6178, opened 2024-08-29
POS: Fix accounting for manually entered keypad amounts

### LINKED ISSUE #6168: Dashboard: If product view within keypad is enabled, we're not counting manual keypad entries towards sales
During BHB24 merchant had 10 sales, he had 8 sales through product view on the keypad, and 2 they manually typed in the amount on keypad. Our dahboard only showed 8 sales.

@dennisreimann can provide more details.

dennisreimann:
For keypad orders where there are products AND manual amount entries, we didn't account for the latter.

Fixes #6168.


dstrukt:
I know we call it `Manual amount`, but I still prefer the `Custom Amount` wording.

Just a side note.

PR #6186, opened 2024-09-02
Greenfield: Can set the label when generating a wallet for store

NicolasDorier:
Reported by @ndeet. It wasn't possible to set the label of the wallet in the generate wallet route.

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
@@ -24,9 +24,9 @@ namespace BTCPayServer.Controllers.Greenfield
         public async Task<IActionResult> GenerateOnChainWallet(string storeId,
             [ModelBinder(typeof(PaymentMethodIdModelBinder))]
             PaymentMethodId paymentMethodId,
-            GenerateWalletRequest request)
+            GenerateOnChainWalletRequest request)
         {
-
+            request ??= new GenerateOnChainWalletRequest();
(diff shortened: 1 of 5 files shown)


BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
```diff
@@ -24,16 +24,18 @@ public partial class GreenfieldStoreOnChainPaymentMethodsController
...
+            request ??= new GenerateOnChainWalletRequest();
             AssertCryptoCodeWallet(paymentMethodId, out var network, out _);
 
             if (!_walletProvider.IsAvailable(network))
             {
                 return this.CreateAPIError(503, "not-available",
                     $"{paymentMethodId} services are not currently available");
             }
+            if (request.Label is { Length: > 300 })
+                ModelState.AddModelError(nameof(request.Label), "Label is too long (Max 300 characters)");
```
dennisreimann, line 38: We could make this a `[MaxLength(300)]` on the `GenerateOnChainWalletRequest` model class.

dennisreimann (approved):

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
NicolasDorier: No,

1. This assembly doesn't have access to `System.ComponentModel.DataAnnotations` (unsure why)
2. Even if it did, the models with annotations aren't validated by ASP.NET for the API. I think I did that on purpose as it was creating issue with properties using custom json converters.

PR #6188, opened 2024-09-03
Invitation process improvements

### LINKED ISSUE #6157: QR code for user invite

### LINKED ISSUE #6158: A way to disable email for user invite in the settings

dennisreimann:
A combination of three things to enhance the invite flow:

- Make sending invitation email optional when adding a new user on the server-level. Closes #6158.
  - Store-level invites will always send an email, because the user need 
- Generate custom invite token and store it in user blob. Closes btcpayserver/app#46.
  - Up to now, we've generated invite codes using the .NET internals, but this leads to lengthy codes and hence dense QR codes. With this, we can also later on show the invitation URL again to the admin — previously that required a new invite code to be generated.
- Display QR code for user invite. Closes #6157.

invite

### DIFF
--- a/BTCPayServer.Data/Data/ApplicationUser.cs
+++ b/BTCPayServer.Data/Data/ApplicationUser.cs
@@ -46,5 +46,6 @@ namespace BTCPayServer.Data
         public bool ShowInvoiceStatusChangeHint { get; set; }
         public string ImageUrl { get; set; }
         public string Name { get; set; }
+        public string InvitationToken { get; set; }
     }
 }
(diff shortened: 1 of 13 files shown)


dstrukt:
Amazing! 

And this will make the carry over with the admin being able to invite / employee being able to accept the invite all from the App.

Only nit, would this make more sense as a modal (with the different CTA to `View Invite`) vs. via the accordion for the Pending Invitation case? Would hide the accordion functionality...

With fresher eyes, it might make sense to have a proper label on the accordion chevron as well, e.g. `Details` or `View Stores` if > 1 Store / No pending invite (but this is just a side thought that's slightly out of scope)

Regardless, will try to get a proper review this evening.

Kukks (approved):

dstrukt (changes requested):
tACK!

Could ship as is, but wanted to discuss if any of the changes mentioned above are worth tackling in the design call, but still flagging in case we want to chat about them later this week.

PR #6190, opened 2024-09-03
Transactions: Improve TX ID display

dennisreimann:
As per discussion with @NicolasDorier.

tx


NicolasDorier:
@dennisreimann now I have another issue with this solution: CTRL + F searching for the txid in the list doesn't work anymore. :(

Is it something that can be improved in the component?

NicolasDorier:
So I tried to set `elastic="true"` then I can CTRL+F. But the end of the transaction's id repeat itself.

dstrukt (approved):
functional tACK!

NicolasDorier:
HUGE improvement for me

PR #6191, opened 2024-09-04
Make Role Permissions more human legible

### LINKED ISSUE #6183: Make Role Permissions more human legible
Got feedback from a user in Riga that our Role Permissions could be a bit more human readable, as opposed to simply listing them out as we do.

We could simply list out the more human legible permissions on this view as we do on the Create view...? e.g.
`Can View Invoices`
`Can Modify Invoices`

Tracking, will need to adjust the copy.

Screen Shot 2024-08-26 at 1 39 04 AM

dennisreimann:
Had to rename `CanModifyStoreWebhooks` to `CanModifyWebhooks` for this, but the value stayed the same, so I don't think it's a big deal.

Closes #6183.

grafik


dennisreimann:
Will add another PR to clean up the deprecated custodian account permissions.

dstrukt:
Awesome - this looks like what they wanted, and it looks like it takes up roughly about the same amount of space.

I will admit, I know the distinction between Store/Server is necessary so that we must keep it, but 
`Can Modify Store Settings`
`Can Modify {Store} {X}`

reads a bit easier than 
Store: Can Modify {X}

no?

And carries over to 
`Can Modify {Server} {X}` just as easily?

Will review asap as well!

dennisreimann:
> no?

yes, but we cannot infer that from what we have. We'd have to refactor the code to achieve something better.

dennisreimann:
Follow-up PR is #6193.

dstrukt:
> > no?
> 
> yes, but we cannot infer that from what we have. We'd have to refactor the code to achieve something better.

I see, sounds good, this more than satisfies the requirement for the time being, and a refactor for that minor copy detail isn't a good use of time atm.

dstrukt (approved):
tACK!

Thanks for taking care of this so quickly!

Issue #6200, opened 2024-09-06
BTCPay 2.0.0-alpha integrations testing


ndeet:
### What is your BTCPay version?

v2.0.0-alpha2

### How did you deploy BTCPay Server?

docker

### What happened?

Testing WooCommerce and other integrations and collect breaking changes here:

1. LN payment does not pay the full amount but it says "has not been paid in full" when there is a small deviation on 12th decimal point

![image](https://github.com/user-attachments/assets/d90daab4-cd4d-4b11-b31d-3b19ba6e0c04)

### How did you encounter this bug?

pay LN invoice

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

ndeet:
@NicolasDorier the `config` data is missing on the Store PaymentMethods endpoint, it just gives me this:
```
[
    {
        "enabled": true,
        "paymentMethodId": "BTC-LN"
    },
    {
        "enabled": true,
        "paymentMethodId": "BTC-CHAIN"
    },
    {
        "enabled": true,
        "paymentMethodId": "BTC-LNURL"
    }
]
```
on the old datastructure `config` was called `data`, do you think we can keep `data`  or should we stick with `config`?

old data structure:
```
{
    "BTC": {
        "enabled": true,
        "data": {
            "derivationScheme": "tpubD6NzVbkrYhZ4YjkqGkhJB8BCYpy31J5YdSWyJ9p5v9mVadicqyjLgCnL4CyFP3NbQeAR1gci8MhYGQM7dWyr9C17UFBhDrkdJ2BRFzhx5bf-[legacy]",
            "label": "btc testnet",
            "accountKeyPath": null
        },
        "cryptoCode": "BTC"
    },
    "BTC-LNURLPAY": {
        "enabled": true,
        "data": {
            "useBech32Scheme": false,
            "lud12Enabled": false
        },
        "cryptoCode": "BTC"
    },
    "BTC-LightningNetwork": {
        "enabled": true,
        "data": {
            "connectionString": "Internal Node"
        },
        "cryptoCode": "BTC"
    }
}
```

dennisreimann:
@ndeet See the `includeConfig` query parameter for that one.

ndeet:
Oh I missed that one, thank you @dennisreimann 

Need to check how I can make it work without a breaking change in PHP library. I see why it now has `canmodifystoresettings` permission but need to figure out how to make it work with the `canviewstoresettings` permission without breaking anything in integrations.

[...]

PR #6202, opened 2024-09-06
Fix divisibility in invoice details of lightning amounts

NicolasDorier:
* Lightning payment method calculations are performed at the msat level.
* Since Blink rounds down to sats and doesn’t support msats, a negative payment method fee (the "Tweak Fee") is added to cover the missing msats and prevent underpayment.
* To keep things simple for the customer, the Tweak Fee is not displayed at checkout.
* he checkout shows amounts in sats, while the invoice details page shows them in msats.

The Tweak Fee is a fee, hidden from Checkout, meant to be used when a payment method has a service provider which
have a different way of converting the invoice's amount into the currency of the payment method.
This fee can avoid under/over payments when this case happens.

This is useful for Blink but also Strike @rockstardev. I expect this is a recurrent quirk from payment providers that we should account for.

This fix an issue reported by @ndeet on https://github.com/btcpayserver/btcpayserver/issues/6200 that was caused by Blink truncating amounts to sats instead of msats.

### DIFF
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -20,6 +20,7 @@ using BTCPayServer.Data;
 using BTCPayServer.HostedServices;
 using BTCPayServer.Hosting;
 using BTCPayServer.JsonConverters;
+using BTCPayServer.Lightning;
 using BTCPayServer.Logging;
 using BTCPayServer.Payments;
 using BTCPayServer.Payments.Bitcoin;
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -506,7 +506,6 @@ namespace BTCPayServer.Tests
 
             var paymentMethod = entity.GetPaymentPrompts().TryGet(PaymentTypes.CHAIN.GetPaymentMethodId("BTC"));
             var accounting = paymentMethod.Calculate();
-            Assert.Equal(1.0m, accounting.ToSmallestUnit(Money.Satoshis(1.0m).ToDecimal(MoneyUnit.BTC)));
             Assert.Equal(1.1m, accounting.Due);
             Assert.Equal(1.1m, accounting.TotalDue);
 
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -19,6 +19,7 @@ using BTCPayServer.NTag424;
 using BTCPayServer.Payments;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Invoices;
+using BTCPayServer.Services.Rates;
 using BTCPayServer.Services.Wallets;
 using BTCPayServer.Views.Manage;
 using BTCPayServer.Views.Server;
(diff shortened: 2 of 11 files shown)


NicolasDorier:
Showing msats in checkout is distracting, and @dstrukt said we should remove it.
There are two ways to do it:
1. Keep msats in the backend, but don't show it on the checkout
2. Scratch msats, always round to sat.
3. Choose a default, then allow the user to choose

BTCPayServer v1 is using `2.`.

Ping @dstrukt and @pavlenex

dennisreimann:
I'd say we should go with option 1.

dstrukt:
Agree with @dennisreimann, and shared same thought in MM. Option 1 probably makes the most sense at this time, and gives us the most flexibility in the future.

NicolasDorier:
@dstrukt the issue I am worried with option 1 is that if a wallet is showing msats, then they will see different amount from checkout.

dennisreimann:
@NicolasDorier but it's the same with the other options as well, right? As far as I see it, 1 basically implies to do 2 for the checkout and imho it would be understandable enough.

ndeet:
> @dstrukt the issue I am worried with option 1 is that if a wallet is showing msats, then they will see different amount from checkout.

do you have any example for your concern? I tried 3 LN wallets (breez, phoenix, blink) and none of them shows msat. I believe there is no LN wallet out there doing it. You will only see if you use CLI.

Also if there are actual millisat in the invoice then we can still show it but hide by default when they are 000. Seems like the most reasonable approach without crippling it for use cases where it still can happen. I would think that LN wallets to the exact same thing?

dstrukt:
> do you have any example for your concern? I tried 3 LN wallets (breez, phoenix, blink) and none of them shows msat. I believe there is no LN wallet out there doing it. You will only see if you use CLI. If the majority does not show msat then it is imho safe to do the same?

> Also if there are actual millisat in the invoice then we can still show it but hide by default when they are 000. Seems like the most reasonable approach without crippling it for use cases where it still can happen. I would think that LN wallets to the exact same thing?

Agree with these points, also did some research with other LN wallets to see how they handle the mSats (Zeus, Muun, Aqua, will test more...) all appear to hide it as well.

NicolasDorier:
Ok then 1. it is

NicolasDorier:
Ok @dstrukt @ndeet I fixed it this way:
1. The calculations are done down to the MilliSat.
2. The display is shown down to the Sat.
3. The invoice details show the MilliSat.

ndeet:
> I remember seeing somewhere that sometimes financial calculation in dollar are usually done up to 4 digits, but only 2 is shown.

Agree, technically we should/need to support msat but no need to show that in the UI. Seems that's how most LN wallets handle it too.

dstrukt:
Awesome, I think approach gives us maximum flexibility while keeping the UI as simple as we can - we can always address any changes as/if we get feedback!

NicolasDorier:
@dstrukt you can test it on https://testing.btcpay.tech/ ask @ndeet  for admin permission.

dstrukt (approved):
tACK.

This feels like a reasonable trade-off, the checkout is simple and expected, and for those who want to double check down to the last mSat, they can go to the invoice details for confirmation, along with giving us flexibility in the future.

PR #6203, opened 2024-09-09
Edit dictionary should be in JSON format

NicolasDorier:
The format was previously `KEY => Translation`.

It is better to use JSON instead, as it will deal with escape, and end of lines properly.
On top of it, it will be easier to translate on chatgpt.


dstrukt:
*Could* this also mean color coding and better `<code>` formatting as well?

NicolasDorier:
@dstrukt no, can't have a code editor with syntax highlight, this is too difficult.

What we could have instead is that the `Edit` doesn't allow to `Edit`, but only to download and upload a JSON document, and we show that as formatted.

dstrukt:
Ahh okay, nevermind - makes sense it would be too difficult upon looking up a few things. I still think the in-browser editor is the best trade-off at the moment (as opposed to the download/upload UX), as previously discussed.

NicolasDorier:
Merging the change of format for now. A code editor is a monster on his own and deserve probably a separate PR.

PR #6214, opened 2024-09-13
Remove confirmation tracking from backend

NicolasDorier:
This is creating some complexity without clear advantage outside of reporting in InvoiceDetails, and in Checkout.

In a later PR, we might decide to restore this information. But rather than tracking it ourselves, we would directly query NBX.


NicolasDorier:
Supersede by https://github.com/btcpayserver/btcpayserver/pull/6215

PR #6217, opened 2024-09-14
Histograms: Add Lightning data and API endpoints

dennisreimann:
Ported over from the mobile-working-branch.

Adds histogram data for Lightning and exposes the wallet/lightning histogram data via the API. It also add a dashboard graph for the Lightning balance. 

Caveat: The Lightning histogram is calculated by using the current channel balance and going backwards through as much invoices and transactions as we have. The "start" of the LN graph data might not be accurate hence. That's because we don't track (and not even have) the LN onchain data. It is calculated by using the current channel balance and going backwards through as much invoices and transactions as we have. So the historic graph data for LN is basically a best effort of trying to reconstruct it with what we have: The LN channel transactions.

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.Lightning.Internal.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.Lightning.Internal.cs
@@ -21,6 +21,13 @@ public partial class BTCPayServerClient
         return await SendHttpRequest<LightningNodeBalanceData>($"api/v1/server/lightning/{cryptoCode}/balance", null, HttpMethod.Get, token);
     }
 
+    public virtual async Task<HistogramData> GetLightningNodeHistogram(string cryptoCode, HistogramType? type = null,
+        CancellationToken token = default)
+    {
+        var queryPayload = type == null ? null : new Dictionary<string, object> { { "type", type.ToString() } };
+        return await SendHttpRequest<HistogramData>($"api/v1/server/lightning/{cryptoCode}/histogram", queryPayload, HttpMethod.Get, token);
+    }
+
     public virtual async Task ConnectToLightningNode(string cryptoCode, ConnectToNodeRequest request,
         CancellationToken token = default)
     {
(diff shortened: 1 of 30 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.lightning.store.json
```diff
@@ -114,6 +114,63 @@
                 ]
             }
         },
+        "/api/v1/stores/{storeId}/lightning/{cryptoCode}/histogram": {
```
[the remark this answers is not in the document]
dennisreimann: Yes, but I think we shopuld clean that up in a follow-up PR, because it's the same for the rest of the LN-related routes as well — not just in the Swagger docs, but also in the controllers.

BTCPayServer.Client/BTCPayServerClient.Lightning.Internal.cs
```diff
@@ -21,6 +21,13 @@ public virtual async Task<LightningNodeBalanceData> GetLightningNodeBalance(stri
         return await SendHttpRequest<LightningNodeBalanceData>($"api/v1/server/lightning/{cryptoCode}/balance", null, HttpMethod.Get, token);
     }
 
+    public virtual async Task<HistogramData> GetLightningNodeHistogram(string cryptoCode, HistogramType? type = null,
+        CancellationToken token = default)
+    {
+        var queryPayload = type == null ? null : new Dictionary<string, string> { ["type"] = type.ToString() };
```
dennisreimann, line 27:
```suggestion
        var queryPayload = type == null ? null : new Dictionary<string, object> { { "type", type.ToString() } };
```

BTCPayServer/Services/Wallets/WalletHistogramService.cs
```diff
@@ -89,6 +89,6 @@ public class WalletHistogramData
 {
     public WalletHistogramType Type { get; set; }
     public List<decimal> Series { get; set; }
-    public List<string> Labels { get; set; }
+    public List<DateTimeOffset> Labels { get; set; }
```
NicolasDorier, line 92: need to add 

`[JsonProperty(ItemsConverter=typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]`

PR #6219, opened 2024-09-15
fix: check lightning payment status

jackstar12:
The lightning payment status is currently assumed to be `Completed` when the pay result is `Ok`, but this doesn't have to be the case - a pay call can be `Ok` while the actual lightning payment is still `Pending`

### DIFF
--- a/BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
+++ b/BTCPayServer/Data/Payouts/LightningLike/UILightningLikePayoutController.cs
@@ -319,23 +319,34 @@ namespace BTCPayServer.Data.Payouts.LightningLike
                 string message = null;
                 if (result.Result == PayResult.Ok)
                 {
-                    message = result.Details?.TotalAmount != null
-                        ? $"Paid out {result.Details.TotalAmount.ToDecimal(LightMoneyUnit.BTC)}"
-                        : null;
-                    payoutData.State = PayoutState.Completed;
-                    try
+                    payoutData.State = result.Details?.Status switch
                     {
-                        var payment = await lightningClient.GetPayment(bolt11PaymentRequest.PaymentHash.ToString(), cancellationToken);
-                        proofBlob.Preimage = payment.Preimage;
-                    }
-                    catch (Exception)
+                        LightningPaymentStatus.Pending => PayoutState.InProgress,
+                        _ => PayoutState.Completed,
+                    };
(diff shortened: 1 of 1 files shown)


NicolasDorier:
mmmh if this happens, the lightning client shouldn't return Ok...
Probably need to rework the interface on the lightning library, because this API design is really easy to use wrong...

jackstar12:
> mmmh if this happens, the lightning client shouldn't return Ok...

There are situations in the boltz plugin where it makes sense, e.g. if boltz rejects 0 conf of the lockup transaction and waits for one confirmation (liquid). The swap was created successfully and the funds are locked up - so everything is Ok but its just gonna take a few more seconds to confirm.

NicolasDorier:
@jackstar12 do I understand correctly that boltz has its own `LightningClient` with this behavior?
Because if Boltz is the only one with this behavior, we shouldn't change the logic to fit it. Instead I suggest to return `Unknown` rather than `Ok`.

Kukks (approved):
I think it looks ok, and I don't think it should break any existing implementations. This comes from us initially only having a PayResult and then extending with the Details object.

NicolasDorier:
yeah @Kukks I think merging it. However, Boltz client should returns Unknown if the lightning isn't final. I am afraid somebody using the lib step on a mine if they don't do this.

Kukks:
> yeah @Kukks I think merging it. However, Boltz client should returns Unknown if the lightning isn't final. I am afraid somebody using the lib step on a mine if they don't do this.

Their Lightning client impl is exclusively used in a btcpay plugin so not too worried. As long as the plugin has a dependency requirement to run on a version of btcpay AFTER this is merged and released.

NicolasDorier:
@Kukks I am worried. If anyone call `Pay` they expect the payment gone through if it is OK.
It may happen through greenfield, or just in a future feature we are coding. This is a bomb waiting to explode. Fine to fix it here, but Boltz plugin need to fix it there too.

jackstar12:
> @jackstar12 are you involved with Boltz to fix it at the implementation level too?

@NicolasDorier I wrote it and intentionally chose to do it this way since it describes the situation most accurately (Ok in general, but still in flight) and It resilts in better ux this way since returning Unknown causes the manual payout UI to display a big red Failed message - which is not accurate. 

But I understand the concerns regarding greenfield API so we could tackle the problem on the UI side (not displaying failed for Unknown status) and return Unknown in boltz

PR #6226, opened 2024-09-17
Greenfield: Add image upload for app items

dennisreimann:
Upload endpoints for app item images. Follow-up to #6075. 

Tested to work with the app item editor.


dstrukt:
YES!

I was going to make an issue for this, and you already have a PR 💥💥

NicolasDorier:
@dennisreimann 

1. There isn't any UploadImage API for any other data stucture (for example store), as such, it shouldn't be expected for the app.
2. Instead, you should do like we do in Store: Use the `UnresolvedUri` type.

A user who wants to update will need to upload the file via the API to the file service, then reference it through an UnresolvedUri.

dennisreimann:
Hmm, but there are dedicated upload actions for the [store logo](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/GreenField/GreenfieldStoresController.cs#L109) as well as the [user profile picture](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs#L234).

NicolasDorier:
@dennisreimann this really sucks, we don't want to add a separate routes everytimes we want to upload a logo.

Let me think about it.

dennisreimann:
We could do it via the GreenfieldFilesController, but we'd have to relax the permission requirement. When I implemented it, I matched the original UI-based file action policies, which make it available to admins only.

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -249,6 +250,63 @@ public async Task<IActionResult> GetAppTopItems(string appId, [FromQuery] int of
...
+
+        [HttpPost("~/api/v1/apps/{appId}/image")]
+        [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        public async Task<IActionResult> UploadAppItemImage(string appId, IFormFile? file)
+        {
+            var app = await _appService.GetApp(appId, null, includeArchived: true);
+            var userId = _userManager.GetUserId(User);
+            if (app == null || userId == null) return AppNotFound();
+            
+            if (file is null)
```
NicolasDorier, line 262: Can you use `_fileService.UploadImage` created by https://github.com/btcpayserver/btcpayserver/pull/6254?

dennisreimann: There are more places, in which we can leverage the new method. Shall I clean up all of them in one go or in a follow-up PR?

BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs
```diff
@@ -249,6 +250,63 @@ public async Task<IActionResult> GetAppTopItems(string appId, [FromQuery] int of
...
+                return this.CreateValidationError(ModelState);
+            
+            try
+            {
+                var storedFile = await _fileService.AddFile(file!, userId);
+                var fileData = new FileData
+                {
+                    Id = storedFile.Id,
+                    UserId = storedFile.ApplicationUserId,
+                    Uri = new UnresolvedUri.FileIdUri(storedFile.Id).ToString(),
```
NicolasDorier, line 286: Do not return this. The API is already confusing enough :D

woutersamaey:
Are you uploading the files base64 encoded or using HTTP form encoding? From an API POV, the base64 way may be more convenient?

dennisreimann:
HTTP form, that's also how we implemented the other upload endpoints.

PR #6228, opened 2024-09-17
App Service: Validate IDs when parsing items template

### LINKED ISSUE #6227: Adding products with the same ID breaks the POS
# EDIT - REASON:

I figured out how I broke it, if the Product has the same `ID`, it triggers this error? Set ID = 2 for two of them.

----

I added 4 items to my Products list, added them all via POS cart, and it kept triggering this error:

Screenshots, because copy/pasting the log via the browser wasn't formatting well. Hope this is helpful.

Screen Shot 2024-09-16 at 4 13 03 PM
Screen Shot 2024-09-16 at 4 13 14 PM

dennisreimann:
Validates missing and duplicate IDs on the edit actions and when creating/updating apps via the API. Fails gracefully by excluding existing items without ID or with duplicate ID for the rest of the cases.

Fixes #6227.

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -367,6 +367,27 @@ namespace BTCPayServer.Tests
                     }
                 )
             );
+            var template = @"[
+              {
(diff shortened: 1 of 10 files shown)


dstrukt (approved):
Looks like it works for me - at least with the original case I broke it with

tACK

PR #6231, opened 2024-09-19
POS: Option for user sign in via the QR code

dennisreimann:
This is an optimization for the current QR code on the Point of Sale editing page: Right now we only offer the option to scan the QR code with another device, to aid reaching the public page. These changes add the option to also sign in a chosen user, so that the Recent Transactions feature becomes available on the frontend.

This consists of two changes:
- Turn the Login Code into a Blazor component and extend it with data for the app. The latter isn't directly related to the feature of this PR, but I did this in conjunction with the login code changes required for the app: In order to sign in the user, the app needs the base URL of the BTCPay Server instance and also the user's email address — both are now appended to the login code, separated by `;`
- Adds the `PosLoginCode` Blazor component, which replaces the simple URL QR code in the POS backend modal. It still features the URL without sign in option, but also a list of users associated with the store. Best described by the following screenshots ...

### No sign in

Just the URL as before, scan and open.

### Sign in the chosen user and redirect

Scanning the QR code requests the new `GET` version of the login code sign in, redirecting to the POS frontend after login.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -3417,11 +3417,14 @@ namespace BTCPayServer.Tests
             var user = s.RegisterNewUser();
             s.GoToHome();
             s.GoToProfile(ManageNavPages.LoginCodes);
-            var code = s.Driver.FindElement(By.Id("logincode")).GetAttribute("value");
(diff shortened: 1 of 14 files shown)


dstrukt:
Interesting solution - will review and think on this one!

PR #6233, opened 2024-09-19
UI: Additional improvements to the User Invitation flow

### LINKED ISSUE #6224: Additional UI Improvements to the User Invitation flow
Follow to up https://github.com/btcpayserver/btcpayserver/pull/6188

@dennisreimann If you don't think we should make these changes, feel free to close this issue.

- [x] Would this make more sense as a modal with the CTA to View Invite vs. the accordion? (for the Pending Invitation case - would hide the accordion functionality in favor of the CTA)

- [x] With fresher eyes, it might make sense to have a proper label on the accordion chevron as well, e.g. Details or View Stores if > 1 Store / No pending invite (but this is just a side thought that's slightly out of scope)

### LINKED ISSUE #6236: Problems with invitation and accepting invitation to the server by user
What is your BTCPay version?
BTCPay Server 2.0.0

How did you deploy BTCPay Server?
Docker through dev env

I'm inviting users directly from the store and sending them the generated link on the 2.0 version of BTCPay Server, but I've encountered a few issues:

1. **404 Error on Follow-Up Visits:**  
   If the user clicks the link but doesn’t complete the form immediately, follow-up visits return a 404 error.  
   **Expected behavior:** The page should only become unavailable after the user sets their password. Ideally, there should be a clearer error message like "Invite expired" rather than a generic 404.

https://github.com/user-attachments/assets/a520a7da-b4c1-495d-9e8e-f2e51d82edd3

2. **Re-inviting Deleted Users:**  
   When inviting users from the Store page, if you delete them and then try to re-invite, it’s not obvious why the invitation link isn’t working. Additionally, if you only have access to Store Users and not Server Users, you won’t be able to remove the user you added, which complicates managing users.

https://github.com/user-attachments/assets/92949242-73fb-4197-a899-57e90783b40e

3. **Incorrect Invitation URL for Localhost:**  
   When visiting the user page while an invitation is pending, the short Invitation URL appears incorrect for localhost. 

https://github.com/user-attachments/assets/f08b9561-d49b-4d61-a109-357bfefbe5a2

dennisreimann:
Closes #6224. Closes #6236. Closes btcpayserver/app#48.

list

modal


pavlenex:
@NicolasDorier should be done for 2.0.0

dstrukt (changes requested):
- [x] think we should just keep the `Details` CTA our brand green

otherwise, think this looks great!

dennisreimann:
>  think we should just keep the Details CTA our brand green

As you wished, ser!

dstrukt (approved):
Thank you for indulging even the slightest of nits!

LGTM - tACK!

pavlenex:
@NicolasDorier we should include this in 2.0.0 beta as agreed on dev call.

Issue #5588, opened 2023-12-21
Move features that aren't considered "core" into plugins


pavlenex:
Going into 2024, one of the things we discussed is that core-codebase needs to be simpler and accessible to new contributors. One of our major goals in 2024 is to move redundant code into plugins and ensure core is stable, secure and easy enough. 

This issue serves as a tracker, and it's up for discussion on what else can be added.

One thing that was discussed is that we should start moving certain things into external codebases/plugins. This also includes altcoins.

```[tasklist]
### Tasks
- [ ] Move POS to plugin
- [ ] Move CF to plugin
- [ ] Move Shopify to plugin
- [ ] Move lightning addresss to plugin
- [ ] Move pay button to plugin
- [ ] Move all Altcoins to plugins
```

PR #6239, opened 2024-09-21
Allow remote xmr node sync

napoly:


napoly:
thanks for the update!.. the full story here is that using a remote note is not really a good idea if it's not your own one.. especially in the privacy world as [it will spy on you](https://github.com/monero-project/monero/blob/master/src/simplewallet/simplewallet.cpp#L3197).. however, the community is pushing for this feature as there are bounties on this one [here](https://bounties.monero.social/posts/88/11-000m-btcpay-server-additional-optimizations) and [here](https://bounties.monero.social/posts/124/5-000m-btcpay-server-allow-syncing-via-remote-node).. 

so maybe just a better UI for testing purposes could help?! hard to tell.. would be nice to see broader discussion involving both bounty sponsors/creators and users.. also, curious who the future maintainers of the Monero plugin will be..

napoly:
dependent on https://github.com/btcpayserver/btcpayserver/issues/5588

satoshinotdead:
> I don't think this will work when you restart the server.
> 
> Anyway, our plan is to move Monero as a plugin in a different repo soon, so the maintainers can decide to merge that or not rather than us.

That's nice! It will allow anyone to win the bounties (~3k USD) and get Monero really working on BTCPay Servers 👌

napoly:
..have re-based,  bumped version and added a warning for the user with an HTTP/HTTPS protocol drop-down.

remone_node_warn

1. The data is properly persisted and loaded form/to PostgreSQL.

db_view_adminer

2. I followed the existing pattern used for zero confirmation or account index, as it appears to be similar.

NicolasDorier:
how is the monero listener using this?
There is a fundamental architectural issue here: There is only a single monero listener at the server level. Though here, you have a monero listener at the store level.

PR #6245, opened 2024-09-25
Optimize load time of StoreRoles related pages/routes

NicolasDorier:
The `GetStoreRole` was loading all the users of the DB on big instances just because we wanted the `IsUsed` boolean...
On big instances the Roles page were taking minutes to open. (or even timeout crash)


dennisreimann:
This should then consequently also remove the `In Use` display from the UI, right?

NicolasDorier:
It shouldn't

PR #6254, opened 2024-09-26
Add updated image upload support on Crowdfund plugin

### LINKED ISSUE #6242: Add updated image upload support on Crowdfund plugin
Just realized we have image upload form support, but not on the Crowdfund.

For those who pick up, see other examples in the app to apply to this instance.

Screen Shot 2024-09-23 at 10 16 53 PM

TChukwuleta:
Resolves #6242 

By changing the featured image from an input field to a custom upload

@dstrukt  @dennisreimann

### DIFF
--- a/BTCPayServer.Abstractions/Contracts/IFileService.cs
+++ b/BTCPayServer.Abstractions/Contracts/IFileService.cs
@@ -14,4 +14,5 @@ public interface IFileService
     Task<string?> GetTemporaryFileUrl(Uri baseUri, string fileId, DateTimeOffset expiry,
         bool isDownload);
     Task RemoveFile(string fileId, string userId);
+    Task<(bool success, string response, IStoredFile file)> UploadImage(IFormFile file, string userId);
 }
--- a/BTCPayServer.Abstractions/Contracts/IFileService.cs
+++ b/BTCPayServer.Abstractions/Contracts/IFileService.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using BTCPayServer.Abstractions.Models;
 
 namespace BTCPayServer.Abstractions.Contracts;
 
@@ -14,5 +15,5 @@ public interface IFileService
     Task<string?> GetTemporaryFileUrl(Uri baseUri, string fileId, DateTimeOffset expiry,
         bool isDownload);
     Task RemoveFile(string fileId, string userId);
-    Task<(bool success, string response, IStoredFile file)> UploadImage(IFormFile file, string userId);
+    Task<UploadImageResultModel> UploadImage(IFormFile file, string userId, long maxFileSizeInBytes = 1_000_000);
 }
--- a/BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
+++ b/BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
@@ -397,6 +397,7 @@ namespace BTCPayServer.Plugins.Crowdfund.Controllers
 
             var settings = app.GetSettings<CrowdfundSettings>();
             var resetEvery = Enum.GetName(typeof(CrowdfundResetEvery), settings.ResetEvery);
+
             var vm = new UpdateCrowdfundViewModel
             {
                 Title = settings.Title,
@@ -410,7 +411,7 @@ namespace BTCPayServer.Plugins.Crowdfund.Controllers
                 StartDate = settings.StartDate,
                 TargetCurrency = settings.TargetCurrency,
                 Description = settings.Description,
-                MainImageUrl = settings.MainImageUrl,
+                MainImageUrl = await _uriResolver.Resolve(Request.GetAbsoluteRootUri(), new UnresolvedUri.FileIdUri(settings.MainImageUrl)),
                 EndDate = settings.EndDate,
                 TargetAmount = settings.TargetAmount,
                 NotificationUrl = settings.NotificationUrl,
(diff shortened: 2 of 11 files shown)


dstrukt:
Beautiful, will get this reviewed here shortly!

dstrukt:
tACK - looks great!

When you run into a validation issue (i.e. not inputting a description), it removes the uploaded images - technically, is there any way we can keep the image until it's explicitly removed (even on refresh)?

I'm not sure if we do this elsewhere, possibly the Point of Sale products, but @dennisreimann might be able to weigh in.

If not, consider this a tACK, and will update my review.

Great job!

dstrukt (commented):
See above.

TChukwuleta:
I don't think it is possible to pass file back to the view, unless the file has been uploaded (has url), because at this point the file hasn't been uploaded (this is done after validation checks)

image

For POS the image has a CTA button to upload which is different from the save button for the form, do you think we should do something similar.  @dstrukt 

Cc. @dennisreimann

BTCPayServer/Storage/Services/FileService.cs
```diff
@@ -47,6 +47,29 @@ public async Task<bool> IsAvailable()
...
+            {
+                var storedFile = await AddFile(formFile, userId);
+                return (true, "File uploaded successfully", storedFile);
+            }
+            catch (Exception e)
+            {
+                return (false, $"Could not save logo: {e.Message}", null);
+            }
+        }
+
```
dennisreimann, line 72: If we abstract this (which is a good idea I think), we should make the max file size an argument and also rephrase th evalidation messages to use "image" instead of "logo".

BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
```diff
@@ -538,6 +548,22 @@ public async Task<IActionResult> UpdateCrowdfund(string appId, UpdateCrowdfundVi
...
             };
 
+            if (vm.MainImageFile != null)
+            {
+                var imageUpload = await _fileService.UploadImage(vm.MainImageFile, userId);
+                if (!imageUpload.success)
+                {
+                    ModelState.AddModelError(nameof(vm.MainImageFile), imageUpload.response);
+                }
+                newSettings.MainImageUrl = await _uriResolver.Resolve(Request.GetAbsoluteRootUri(), new UnresolvedUri.FileIdUri(imageUpload.file.Id));
```
dennisreimann, line 558: This should be in an else clause then, right?

BTCPayServer.Abstractions/Contracts/IFileService.cs
```diff
@@ -14,4 +14,5 @@ public interface IFileService
     Task<string?> GetTemporaryFileUrl(Uri baseUri, string fileId, DateTimeOffset expiry,
         bool isDownload);
     Task RemoveFile(string fileId, string userId);
+    Task<(bool success, string response, IStoredFile file)> UploadImage(IFormFile file, string userId);
```
dennisreimann, line 17: This interface is a bit unusual I'd say. Can you rework it to be mor elike what we e.g. have in `UserService.TryCanLogin`?

Somethink like: `Task<IStoredFile>(IFormFile file, string userId, out string error)`

BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
```diff
@@ -434,8 +438,14 @@ public async Task<IActionResult> UpdateCrowdfund(string appId)
 
         [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
         [HttpPost("{appId}/settings/crowdfund")]
-        public async Task<IActionResult> UpdateCrowdfund(string appId, UpdateCrowdfundViewModel vm, string command)
+        public async Task<IActionResult> UpdateCrowdfund(string appId, UpdateCrowdfundViewModel vm,
+             [FromForm] bool RemoveLogoFile,
```
dennisreimann, line 442: This should come last, do not reposition the existing command argument.

BTCPayServer/Plugins/Crowdfund/Controllers/UICrowdfundController.cs
dennisreimann: Also: Do not assign the resolved URL, but rather `new UnresolvedUri.FileIdUri(imageUpload.file.Id)`.

Resolve it when handing out the data to the view/API response.

dennisreimann (changes requested):

BTCPayServer.Abstractions/Contracts/IFileService.cs
TChukwuleta: Hey @dennisreimann  I have updated it.. I created a new response model.. async process doesn't work with ref. 

You can review.

dstrukt (approved):
Alright, sounds good, then tACK - LGTM!

dennisreimann (commented):
One more note: The image needs to get resolved here as well: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Controllers/GreenField/GreenfieldAppsController.cs#L435

TChukwuleta:
Hey @dennisreimann  the image has been resolved in the greenfield API

dennisreimann (approved):
tACK.

I've made a few adjustments, can you double-check them, @TChukwuleta?

TChukwuleta:
Ermm.. I notice that the crowdfund public link is now visible even without a description. Just wanted to point that out. 

I dont know if that's the case now @dennisreimann 

If it is, I noticed that the View button then disappears when I tried to update and description validation fails.


image

dennisreimann:
> Ermm.. I notice that the crowdfund public link is now visible even without a description. Just wanted to point that out.

Yes, you are right, should be fixed now.

TChukwuleta:
Okay. I'd take another look

TChukwuleta:
tACK.  All is fine now @dennisreimann

PR #6255, opened 2024-09-26
 prevent app creation without wallet creation

TChukwuleta:
- Added check to ensure a wallet is set up before creating an app
- App type is now included in the wallet setup message to provide more context to the user

image


dstrukt (approved):
tACK - LGTM functionally

Even worked with an existing CF that I was able without payment method before testing this PR.

PR #6258, opened 2024-09-30
Refactoring: Move AppItem to Client lib and use the class for item list

dennisreimann:
This makes it available for the app, which would otherwise have to replicate the model. Also uses the proper class for the item/perk list of the app models.

### DIFF
--- a/BTCPayServer.Client/Models/AppBaseData.cs
+++ b/BTCPayServer.Client/Models/AppBaseData.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Client.Models;
 
@@ -15,7 +16,7 @@ public class AppBaseData
     [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
     public DateTimeOffset Created { get; set; }
     [JsonExtensionData]
-    public IDictionary<string, object> AdditionalData { get; set; } = new Dictionary<string, object>();
+    public IDictionary<string, JToken> AdditionalData { get; set; } = new Dictionary<string, JToken>();
 }
 
 public interface IAppRequest
--- a/BTCPayServer.Client/Models/AppBaseData.cs
+++ b/BTCPayServer.Client/Models/AppBaseData.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Client.Models;
 
@@ -16,7 +15,7 @@ public class AppBaseData
     [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
     public DateTimeOffset Created { get; set; }
     [JsonExtensionData]
-    public IDictionary<string, JToken> AdditionalData { get; set; } = new Dictionary<string, JToken>();
+    public IDictionary<string, object> AdditionalData { get; set; } = new Dictionary<string, object>();
 }
 
 public interface IAppRequest
--- a/BTCPayServer.Client/Models/AppItem.cs
+++ b/BTCPayServer.Client/Models/AppItem.cs
@@ -41,9 +41,6 @@ public class AppItem
     [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
     public int? Inventory { get; set; }
 
-    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
-    public string[] PaymentMethods { get; set; }
-
     [JsonExtensionData]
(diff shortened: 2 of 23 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
```diff
@@ -747,34 +747,31 @@
...
-                                        "description": "Tian Jian Tian Jian means 'heavenly tippy tea' in Chinese, and it describes the finest grade of dark tea. Our Tian Jian dark tea is from Hunan province which is famous for making some of the best dark teas available.",
                                         "id": "black tea",
-                                        "image": "~/img/pos-sample/black-tea.jpg",
-                                        "price": {
-                                            "type": 2,
-                                            "formatted": "$1.00",
-                                            "value": 1.0
-                                        },
                                         "title": "Black Tea",
+                                        "description": "Tian Jian Tian Jian means 'heavenly tippy tea' in Chinese, and it describes the finest grade of dark tea. Our Tian Jian dark tea is from Hunan province which is famous for making some of the best dark teas available.",
```
Kukks, line 771: What if I want a lighter tea?

BTCPayServer.Client/Models/AppBaseData.cs
```diff
@@ -16,7 +15,7 @@ public class AppBaseData
     [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
     public DateTimeOffset Created { get; set; }
     [JsonExtensionData]
-    public IDictionary<string, JToken> AdditionalData { get; set; } = new Dictionary<string, JToken>();
```
NicolasDorier, line 19: why?

BTCPayServer.Client/Models/AppItem.cs
```diff
@@ -0,0 +1,46 @@
...
+    public bool Disabled { get; set; }
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string Description { get; set; }
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string[] Categories { get; set; }
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public string Image { get; set; }
```
NicolasDorier, line 29: good time to change it to This should be a `UnresolvedUri`, so we can link File Ids...

BTCPayServer.Client/Models/AppItem.cs
```diff
@@ -0,0 +1,46 @@
...
+    Minimum
+}
+
+public class AppItem
+{
+    public string Id { get; set; }
+    public string Title { get; set; }
+    public bool Disabled { get; set; }
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
```
NicolasDorier, line 22: Better remove those if this is exposed by the API. In the API, if those fields are shown, it self-document to the user the available fields in the API.

BTCPayServer.Client/Models/AppBaseData.cs
dennisreimann: For the app we try to not use Newtonsoft.Json types and here it doesn't make much of a difference to use `object` I'd say.

BTCPayServer.Client/Models/AppItem.cs
dennisreimann: As far as I see, we do not use `UnresolvedUri` in the client project up to now. We resolve and assign these on the Greenfield controller level.

BTCPayServer.Client/Models/AppItem.cs
dennisreimann: Done in d6b6765.

Kukks (approved):

BTCPayServer.Client/Models/AppBaseData.cs
NicolasDorier: 1. You can't use those types without Newtonsoft.Json, as they use converters.
2. If you change to object, then the `JsonExtensionData` isn't respected anymore. This is a problem, because it means that if the client doesn't know a field of this object, the field won't do a roundtrip (get/change the object/update), effectively erasing the previous value of those fields.

dennisreimann: Ok, I reverted this.

BTCPayServer.Client/Models/AppItem.cs
```diff
@@ -0,0 +1,35 @@
...
+    [JsonConverter(typeof(StringEnumConverter))]
+    public AppItemPriceType PriceType { get; set; }
+
+    [JsonConverter(typeof(NumericStringJsonConverter))]
+    public decimal? Price { get; set; }
+    public string BuyButtonText { get; set; }
+    public int? Inventory { get; set; }
+
+    [JsonExtensionData]
+    public Dictionary<string, object> AdditionalData { get; set; }
```
NicolasDorier, line 34: Should be `JObject`,  not `object`.

PR #5744, opened 2024-02-09
UI: Move section navigation to sidebar

dennisreimann:
Proof of concept for the new sidebar structure, including the active section navigation. As discussed in yesterday's design meeting — this changes the navigation for the Store Settings and Server Settings, so that we can discuss it.

I think it works very well, we'd onlty need to ensure that plugins also update/migrate.

Closes #3827. Closes #5581.

## Desktop

grafik

## Mobile

grafik

PR #6267, opened 2024-10-01
V2 compatibility: Re-add deprecated navigation methods

dennisreimann:
Gives the new methods a new name and re-adds the old ones (Context: #5744) in order to not break plugins. Simple enough backwartds compatible change, makred the old methods as obsolete to make plugin developers aware that new methods are available.

### DIFF
--- a/BTCPayServer.Abstractions/Extensions/ViewsRazor.cs
+++ b/BTCPayServer.Abstractions/Extensions/ViewsRazor.cs
@@ -55,7 +55,7 @@ namespace BTCPayServer.Abstractions.Extensions
             viewData[ACTIVE_CATEGORY_KEY] = activeCategory;
         }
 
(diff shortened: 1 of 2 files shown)


dennisreimann:
Failing test is fixed in #6265.

NicolasDorier:
@dennisreimann plugin breaking isn't really bad, it is detected then they are disabled on restart. Still good to have a fallback like this but we don't have to go out of our way for plugin backward compatiblity.

dennisreimann:
I agree, that's why I removed the old methods at first. 

But as this is view-only code, the errors caused by the removals wouldn't deactivate the plugin, because they do not crash on startup.

Also it is an easy enough fix, that imho we should ease the plugun devs life and not have everything breaking for them when just opening the page. We can transition devs a bit more gracefully here.

PR #5938, opened 2024-04-18
Design system and icon updates for 2.0

dennisreimann:
WIP, will update this in an ongoing manner as we progress with 2.0 and the app.

Closes #5485.

PR #6284, opened 2024-10-08
Add fontawesome back

NicolasDorier:
This commit removed fontawesome: https://github.com/btcpayserver/btcpayserver/pull/5938

This PR adds it back because multiple plugins still depends on it.
I didn't add it to the HeaderLayout, so it won't be available by default for normal BTCPay Server development, but plugins can still reference it.

We can potentially remove it one day, but we first should migrate the plugins.
I myself didn't managed to migrate due to size issues.

Ping @dennisreimann


dennisreimann:
@dstrukt we need to find out which plugins are affected and add replacements for what they need :)

PR #6287, opened 2024-10-09
Support for searching plugins by name

rockstardev:
Depends on: https://github.com/btcpayserver/btcpayserver-plugin-builder/pull/34

In action: 

https://github.com/user-attachments/assets/2e6cdfeb-dcab-4260-88c5-6c0fbf117b87


rockstardev:
@NicolasDorier I've extensively tested this in various setups and confirmed it works, merging it to make sure it is in 2.0.

One final thing to keep in mind - with this change we will also start sending version string, so that in future we can differentiate between queries if needed:

https://github.com/btcpayserver/btcpayserver/pull/6287/files#diff-9ec39991492de8ab4f273cb066c8936961207606d39bf189cb4a6b18078c51aeR51

This is to avoid problems like: https://github.com/btcpayserver/btcpayserver/pull/6287#discussion_r1795192399

PR #6289, opened 2024-10-10
Require non interactivity for boltcard payments

NicolasDorier:
## Old behavior

If you paid via Boltcard and didn't:
* Set auto-approval on the Pull Payment
* Set up the Lightning automated processor with immediate payout
then the checkout would show this:

image

The merchant would be confused, as would the user, who likely wouldn't know what to do with this message.

## New behavior

When paying with a Boltcard backed by BTCPay Server, the pull payment should require auto-approval, and the Lightning automated processor with immediate payout must be enabled for the store.

If that's not the case, the new behavior will display a red alert box indicating that the payment has been canceled.

For regular pull payments, interactivity is allowed, enabling the payer to manually approve an LNURLWithdraw request at a later time

## Rationale

The rationale is that a Boltcard is likely used for merchant payments and shouldn't remain in a 'Pending' state, as this would leave both the merchant and the user uncertain about whether the payment has been sent or received. (or be sent later)

However, someone looking to sweep a pull payment may be fine with waiting for the creator of the pull payment to accept their request at a later time.

### DIFF
--- a/BTCPayServer/Controllers/UIBoltcardController.cs
+++ b/BTCPayServer/Controllers/UIBoltcardController.cs
@@ -65,6 +65,6 @@ public class UIBoltcardController : Controller
         if (!cardKey.CheckSunMac(c, piccData))
             return BadRequest(new LNUrlStatusResponse { Status = "ERROR", Reason = "Replayed or expired query" });
         LNURLController.ControllerContext.HttpContext = HttpContext;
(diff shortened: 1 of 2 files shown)


dstrukt:
ACK, agree it should either succeed or fail (and indicate appropriately), otherwise the pending/limbo state is confusing.

I could see `pending` approval functionality potentially working on the App, but given the current realities, this makes a lot more sense.

Kukks (approved):

PR #6291, opened 2024-10-11
Improve UX for uninstalling disabled plugins

NicolasDorier:
When updating to BTCPay 2.0, a lot of plugins will crash and get disabled.
This PR:
* Add a button to uninstall them all in one click
* Change the state of the "uninstall" button when it has already been clicked.

@pavlenex 

image


pavlenex:
cACK

PR #6299, opened 2024-10-15
UI: Inactive toggle hover color fix

dennisreimann:
@pavlenex noted, that before the inactive state had the primary highlight color on hover, which makes it seem like it is already active. This changes it to use a slightly darker/lighter neutral version for inactive state hover.


dstrukt (approved):
awww yeah, LGTM!

tACK!

PR #6311, opened 2024-10-16
refactor: make `BitcoinCheckoutModelExtension` support other payment handlers

jackstar12:
The bitcoin checkout extension doesn't have to be tied to the native bitcoin handler since it only really needs the payment details to be in a specific format, which can be provided by other handlers aswell, allowing for better code reuse.

The small part for the `BitcoinPaymentData` can be handled inside a check for the standard onchain handler.

### DIFF
--- a/BTCPayServer/Payments/Bitcoin/BitcoinCheckoutModelExtension.cs
+++ b/BTCPayServer/Payments/Bitcoin/BitcoinCheckoutModelExtension.cs
@@ -7,7 +7,6 @@ using BTCPayServer.Models.InvoicingModels;
 using BTCPayServer.Services;
 using BTCPayServer.Services.Invoices;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Localization;
 using NBitcoin;
 using NBitcoin.DataEncoders;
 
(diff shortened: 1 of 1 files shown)


Kukks (approved):

BTCPayServer/Payments/Bitcoin/BitcoinCheckoutModelExtension.cs
```diff
@@ -16,29 +15,26 @@ namespace BTCPayServer.Payments.Bitcoin
...
         public const string CheckoutBodyComponentName = "BitcoinCheckoutBody";
-        private readonly PaymentMethodHandlerDictionary _handlers;
         private readonly BTCPayNetwork _Network;
         private readonly DisplayFormatter _displayFormatter;
         private readonly IPaymentLinkExtension paymentLinkExtension;
         private readonly IPaymentLinkExtension? lnPaymentLinkExtension;
         private readonly IPaymentLinkExtension? lnurlPaymentLinkExtension;
         private readonly string? _bech32Prefix;
+        private PaymentMethodId lnPmi => PaymentTypes.LN.GetPaymentMethodId(_Network.CryptoCode);
+        private PaymentMethodId lnurlPmi => PaymentTypes.LNURL.GetPaymentMethodId(_Network.CryptoCode);
```
NicolasDorier, line 25: This is wasteful, make it local fields and calculate the value in the constructor.

PR #6315, opened 2024-10-17
[Greenfield] Select default payoutMethodId if none are selected in the Refund route

NicolasDorier:
Proposed by @ndeet, as 2.0 is breaking the refund route via a rename of `paymentMethod` to `payoutMethodId`, the call to the endpoint always failed.

This PR will just instead select the default payment method of the invoice.


NAKKIABW:
Approve

PR #6316, opened 2024-10-17
refactor(checkout): displayed payment methods vue component

jackstar12:
The displayed payment methods can change with updates aswell, so it should be rendered as a vue component instead.
As an example, a payment method might have a minimum amount which the invoice was initially above, but isnt anymore after a partial payment, so it deactivates itself.

### DIFF
--- a/BTCPayServer/Views/UIInvoice/Checkout.cshtml
+++ b/BTCPayServer/Views/UIInvoice/Checkout.cshtml
@@ -12,7 +12,7 @@
     ViewData["StoreBranding"] = Model.StoreBranding;
 	Csp.UnsafeEval();
 	var hasPaymentPlugins = UiExtensions.Any(extension => extension.Location == "checkout-payment-method");
-	var displayedPaymentMethods = Model.AvailablePaymentMethods.Where(c => c.Displayed).ToList();
+    var checkoutLink = Url.Action("Checkout", new { invoiceId = Model.InvoiceId });
 }
 @functions {
     private string ToJsValue(object value)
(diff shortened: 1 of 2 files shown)


BTCPayServer/Views/UIInvoice/Checkout.cshtml
```diff
@@ -90,24 +90,20 @@
...
-                            @await Component.InvokeAsync("UiExtensionPoint", new { location = "checkout-payment-method", model = Model })
-                        </div>
+                <div v-if="displayedPaymentMethods.length > 1 || @Safe.Json(hasPaymentPlugins)" class="mt-3 mb-2">
+                    <h6 class="text-center mb-3" v-t="'pay_with'"></h6>
+                    <div class="btcpay-pills d-flex flex-wrap align-items-center justify-content-center gap-2 pb-2">
+                        <a
+                            v-for="crypto in displayedPaymentMethods"
+                            :href="@ToJsValue(checkoutLink) + '/' + crypto.paymentMethodId"
+                           class="btcpay-pill m-0 payment-method"
+                           :class="{ active: srvModel.paymentMethodId === crypto.paymentMethodId }"
```
dennisreimann, line 100: @jackstar12 Can you explain the reasoning behind changing `:href` and `:class`?

BTCPayServer/Views/UIInvoice/Checkout.cshtml
```diff
@@ -90,24 +90,20 @@
...
+                           v-on:click.prevent="changePaymentMethod(crypto.paymentMethodId)"
+                           v-text="crypto.paymentMethodName">
+                        </a>
                     </div>
-                }
+                    @await Component.InvokeAsync("UiExtensionPoint", new { location = "checkout-payment-method", model = Model })
```
dennisreimann, line 105: Moving this outside the div above break the UI for other existing plugins. We need to revert this.

dennisreimann: @jackstar12 Can you check if #6481 works for your case?

PR #6330, opened 2024-10-24
feat: log download button

jackstar12:
Adds a download button to each log file in the Logs view. Sending a file can be a lot easier than pasting a lot of raw content in a lot of messengers.

Preview:
image

Also note that I had to remove the `using` from the `fileStream` since the `FileStreamResult` needs to use it after the function has returned (I assume it takes care of closing it when its done).


BTCPayServer/Controllers/UIServerController.cs
```diff
@@ -1317,13 +1317,21 @@ public async Task<IActionResult> LogsView(string? file = null, int offset = 0)
                     return NotFound();
                 try
                 {
-                    using var fileStream = new FileStream(
+                    var fileStream = new FileStream(
```
NicolasDorier, line 1320: why you removed `using`?

BTCPayServer/Views/UIServer/Logs.cshtml
```diff
@@ -13,6 +13,12 @@
     {
         <li>
             <a asp-action="LogsView" asp-route-file="@file.Name">@file.Name</a>
+            <a asp-action="LogsView" asp-route-file="@file.Name" asp-route-offset="@Model.LogFileOffset" asp-route-download="true" class="mx-2">
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-download" viewBox="0 0 16 16">
```
NicolasDorier, line 17: @dennisreimann @dstrukt we have a download icon?

jackstar12: I didnt find one at https://design.btcpayserver.org/design/icons/ or in the source code. Should I add it as a symbol in `icon-sprite.svg`?

BTCPayServer/Controllers/UIServerController.cs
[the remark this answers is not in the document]
jackstar12: I guess the best way would be a `using` block

BTCPayServer/Views/UIServer/Logs.cshtml
NicolasDorier: Unsure, need to ping @dennisreimann and @dstrukt for that

BTCPayServer/Controllers/UIServerController.cs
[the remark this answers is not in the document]
NicolasDorier: @jackstar12 Yes, but only if not Downloading. For download, indeed you shouldn't do it as the `FileStreamResult` does it after the file is downloaded.

BTCPayServer/Views/UIServer/Logs.cshtml
dennisreimann: Added it in 0ac655384030231780851ef753ed222160e604ea.

PR #6351, opened 2024-11-01
UI: Improve brand color adjustment

dennisreimann:
Uses a more finegrained method of deciding whether or not to adjust the brand color: Instead of simply using the brightness of the color, we calculate the contrast ratio and adjust the color by increasing the contrast until it is sufficient.

The thresholds also try to preserve the brand color in its original form, so that the color only gets adjusted if the contrast is very low.

@dstrukt @pavlenex This was [reported on Mattermost](https://chat.btcpayserver.org/btcpayserver/pl/84oppwxcmfgnjfcmsp4yfpieyr) by a user who noticed that their brand color got adjusted too much. I've fiddled with the thresholds quite a bit, please test and let me know if this works in your opinion.

Example from wrong adjustmnents before:

grafik

grafik

### DIFF
--- a/BTCPayServer/ColorPalette.cs
+++ b/BTCPayServer/ColorPalette.cs
@@ -110,5 +110,10 @@ namespace BTCPayServer
         {
             return ColorTranslator.FromHtml(html);
         }
+
+        public string ToHtml(Color color)
+        {
+            return ColorTranslator.ToHtml(color);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


dstrukt (approved):
tACK, seems like a reasonable approach

AFAICT, it should address the original concern well, and worked well on the tests i had done in the past

Issue #6361, opened 2024-11-04
[Bug]: Pull-Payments / LNURL Withdraw QR Code view is not centered - looks weired


petzsch:
### What is your BTCPay version?

 © BTCPay Server v2.0.1+8a5a16064 

### How did you deploy BTCPay Server?

docker

### What happened?

Created a pull payment and tried to claim it with lnurl withdrawl through my lightning wallet. got this screen which looks a bit weirdly formatted: (QR Code is claimed allready and pull payment is archived):

![grafik](https://github.com/user-attachments/assets/5ffd3db6-56f7-4f78-8141-5f181ccc6897)

### How did you encounter this bug?

1. created a pull payment
2. clicked on view
3. clicked on the QR code icon in the pull payment

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox on MacOS

### Additional information

none

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #6362, opened 2024-11-04
fix: center qr code

jackstar12:
closes: #6361 

example from issue now:
image

have yet to look at the other views which use the view

### DIFF
--- a/BTCPayServer/Views/Shared/ShowQR.cshtml
+++ b/BTCPayServer/Views/Shared/ShowQR.cshtml
@@ -9,8 +9,8 @@
                     </button>
                 </div>
                 <div class="modal-body pt-0">
-                    <div class="payment-box m-0">
-                        <div class="qr-container justify-content-start">
+                    <div class="payment-box">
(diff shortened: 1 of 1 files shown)


petzsch:
schould the title (LNURL Withdraw) be centered as well? Looks slightly off still.

Jacksonearl2468:
Approved

Jacksonearl2468 (approved):

dstrukt:
> schould the title (LNURL Withdraw) be centered as well? Looks slightly off still.

We shouldn't center the title, it will mess up other modals.

The only other way to make this feel more balanced imo would be to expend the `LNURL Withdraw` to be full width of the container, but it's minor, and the current proposal, while I agree, looks a little off, is mostly fine.

dstrukt (approved):
As suggested above, outside of maybe extending the `LNURL Withdraw` to full-width, this LGTM!

Jacksonearl2468 (approved):

Jacksonearl2468 (approved):

PR #6368, opened 2024-11-08
UI: Fix mising navigation links for store managers

dennisreimann:
As reported by @pavlenex, the store settings links were missing for users with the `Manager` role.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTester.cs
+++ b/BTCPayServer.Tests/SeleniumTester.cs
@@ -643,7 +643,15 @@ retry:
             GoToUrl(url);
             Assert.DoesNotMatch("404 - Page not found</h", Driver.PageSource);
             if (shouldHaveAccess)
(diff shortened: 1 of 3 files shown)


dstrukt:
oh wow, so the store settings just didn't appear at all for the Manager role?

will also test

dennisreimann:
> so the store settings just didn't appear at all for the Manager role?

The main page did, but not the subnavigation items/pages.

dstrukt:
> > so the store settings just didn't appear at all for the Manager role?
> 
> The main page did, but not the subnavigation items/pages.

Ahh i see, great catch!

NicolasDorier:
I am not expecting a view only permission to have access to those settings

dstrukt (approved):
LGTM - AFAICT

Kukks (approved):

PR #6372, opened 2024-11-08
Store: Fix missing invitation email when adding new user

### LINKED ISSUE #6369: [Bug]: Creating user at store level doesnt send email invitation

BTCPay v2.0.1

docker

### What happened?

In store level, after adding a totaly new user, there is a message about invitation mail and invitation link displayed, but mail is not sent. 
When user is created at server level everything works and mails are sent. 

Make sure the SMTP for server is set and allowed to be used by stores. 
Login with Owner role and go to Store settings. Under Users add new user. 

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Fixes #6369, a regression introduced with #6188: Sending invite emails required a flag to be set, which defaulted to false. This fixes the code to explicitely set the flag and also defaults it to true (sending an invite email unless declared otherwise).

### DIFF
--- a/BTCPayServer/Controllers/UIStoresController.Users.cs
+++ b/BTCPayServer/Controllers/UIStoresController.Users.cs
@@ -70,6 +70,7 @@ public partial class UIStoresController
                     Kind = UserRegisteredEventKind.Invite,
                     User = user,
                     InvitedByUser = currentUser,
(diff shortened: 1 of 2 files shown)


dstrukt (approved):
tACK

Nice catch of the regression!

Issue #5968, opened 2024-05-03
App: Integrate mobile-working-branch


dennisreimann:
Just a reminder for us to integrate the app-reklated updates from the `mobile-working-branch` into the core project. It contains new APIs for the upcoming BTCPay Server app.

PR #6374, opened 2024-11-09
Extract CommonServer package

dennisreimann:
Part of #5968: Extracts common functionality, which is shared between app and server. This cannot be part of the Abstractions package, because that one depends on ASP.NET, which does not work in MAUI.


BTCPayApp.CommonServer/BTCPayApp.CommonServer.csproj
```diff
@@ -0,0 +1,17 @@
...
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\BTCPayServer.Client\BTCPayServer.Client.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="8.0.8" />
+      <PackageReference Include="VSS" Version="1.0.1" />
```
NicolasDorier, line 15: What is this package?

BTCPayApp.CommonServer/BTCPayApp.CommonServer.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <LangVersion>latest</LangVersion>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\BTCPayServer.Client\BTCPayServer.Client.csproj" />
```
NicolasDorier, line 10: You said that your project can't reference `BTCPayServer.Client`, but this is exactly what you do with `CommonServer` here.

If that it true, your project might not be able to reference CommonServer either. And if it does, why the need of a separate project?

dennisreimann: We cannot reference `BTCPayServer.Abstractions`.

NicolasDorier:
I have thought about it more.

On one side, we want to be able to maintain the app more effectively and as such, making a system plugin would definitely fit the purpose.

On the other side:
1. This is adding a few dependencies that I would rather not see, as most users of BTCPay will never use it.
2. This is creating an undocumented API which will internally expose things which may potentially be abused, and if it is shipped as part of core, then it will expose everybody rather than only people installing the plugin.
3. The development process of the app is different from core, as such it will probably require its own releases independent of BTCPay Server.
4. It is creating yet another project whose goal isn't really obvious to a developer in btcpayserver: Where to put a specific class? Common? Abstractions? CommonServer? I was planning to consolidate those projects rather than spreading it further.

@dennisreimann @Kukks is there any reason, outside of convenience for moving this stuff here?

Issue #6371, opened 2024-11-08
[Bug]: 2.0.1 Crowdfunding balance tracking logic is inaccurate


pavlenex:
### What is your BTCPay version?

2.0.1

### How did you deploy BTCPay Server?

Docker

### What happened?

Several users are reporting that upon update from 2.0 to 2.0.1 they're seeing innacurate pending balance in the crowdfunding bar despite no payments being made.  Users can't detect any stuck payments or suspicious invoices. 

This is likely related to recent changes in 2.0.1 around percentage logic here  https://github.com/btcpayserver/btcpayserver/pull/6347/

Examples:
https://Donate.praiabitcoin.org
https://Savestarlightschool.org

### How did you encounter this bug?

Update existing crowdfunding to 2.0.1 (I was unable to reproduce)

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

PR #6381, opened 2024-11-11
Fix: Incorrect calculation for crowdfund and payment request status

NicolasDorier:
Fix https://github.com/btcpayserver/btcpayserver/issues/6371

In the Crowdfund app, payments for expired or invalid invoices were always recorded as Pending payments.

Additionally, there was a bug in both the Crowdfund app and the Payment Request feature, where a user could send a payment to an old invoice, and it would be included in the total for the payment request or crowdfund.

This behavior is undesirable, as the currency rate may have fluctuated since the invoice's expiration.

Instead, for crowdfund and payment requests, we will now account for every payment:
* Paid before the expiration regardless of the status (ie. even if the invoice is Invalid)
* Paid on an invoice that has been manually marked Settled. (I am not sure this is a good idea, as somebody could then pay the old invoice with an old rate, replicating the bug from 1.0. however, a merchant may expect this behavior?)

### DIFF
--- a/BTCPayServer/PaymentRequest/PaymentRequestService.cs
+++ b/BTCPayServer/PaymentRequest/PaymentRequestService.cs
@@ -66,12 +66,10 @@ namespace BTCPayServer.PaymentRequest
             {
                 var invoices = await _paymentRequestRepository.GetInvoicesForPaymentRequest(pr.Id);
                 var contributions = _invoiceRepository.GetContributionsByPaymentMethodId(blob.Currency, invoices, true);
-                var allSettled = contributions.All(i => i.Value.Settled);
-                var isPaid = contributions.TotalCurrency >= blob.Amount;
 
-                if (isPaid)
+                if (contributions.Total >= blob.Amount)
                 {
-                    currentStatus = allSettled
+                    currentStatus = contributions.TotalSettled >= blob.Amount
                         ? Client.Models.PaymentRequestData.PaymentRequestStatus.Completed
                         : Client.Models.PaymentRequestData.PaymentRequestStatus.Processing;
                 }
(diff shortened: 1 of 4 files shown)


pavlenex:
Okay, this is an interesting corner-case, glad you found it. I am unsure why it was implemented this way originally, but what you outlined makes sense to me. 

I've tested things on `master` first to ensure I understood behavior you're describing. Send funds to expiring invvoice and on the `master` it indeed gets into `pending` status on the UI of the crowdfund. 


https://github.com/user-attachments/assets/51913e18-67dd-439c-bd7f-bfd8d92c65ca



When tested on this PR, the behaviour is now expected. Paid late invoice doesn't get included in the view. Then I manually settled it , and it had no confirmations, then it was shown as `pending` in the UI, upon confirmations it was correctly shown again.



https://github.com/user-attachments/assets/d1f46a11-89cb-4024-824b-5717cc1fe713


So on my end, tACK.

NicolasDorier:
@pavlenex, I am not sure about this part.

```
* Paid on an invoice that has been manually marked Settled. (I am not sure this is a good idea, as somebody could then pay the old invoice with an old rate, replicating the bug from 1.0. however, a merchant may expect this behavior?)
```

The current situation is:
1. If an invoice has been marked as Settled and it has no payment, account the whole invoice's amount to the crowdfund/payreq
2. If an invoice has been marked as Settled and it has payment, account all the payments to the crowdfund/payreq.

I am thinking we should maybe change 2... because if somebody make a payment after the expiration of such invoice, it will be accounted... even if the rate is obsolete.

NicolasDorier:
From pav:

> Hmmm I am leaning towards not changing any behaviour to that comment.
> 
> Because:
> 
> For crowdfunding, it's donation-based people accept and credit whatever you sent them anyway.
> For payment request, it can actually be helpful to see that there's more money then requested on the user-facing page.
> Imagine you invoice me 2000$, you send a link, I pay later, after expiration, you still want to showcase that certain amount is there, but it may be at a different rate

Let's keep it like that then!

PR #6382, opened 2024-11-11
Fix: Do not allow retry of payouts if they are non interactive (Boltcards)

### LINKED ISSUE #6377: [Bug]: Failed payouts (routing issue) aren't canceled but left in awaiting payment status for days/possibly indefinitiatly?

2.0.2

docker

### What happened?

I tried to pay a 100 sats with my boltcard (linked to btcpay and to blink) to a Zeus wallet which is only reachable via private channel from LNBIG (Hub-3).

The payment doesn't happen and in the pull request it remains in " Awaiting Payment " status. As a end user, I have no chance to cancel this payment attempt (owners of boltcards don't have an interface for that). And as server admin I can only cancel the payouts manually.

I would suggest:

1. cancel the payout when the invoice from Zeus that it's trying to pay has expired (by default with this wallet 60 minutes)
2. after 10 retries each after the automatic payment processors interval, not just stop attempting the payment, but also cancel the payout. So that in the case of boltcard users, their balance is no longer blocked.

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox on Windows 11

### Additional information

none

NicolasDorier:
Fix #6377

From version 2.0, a payout that fails to be sent will be automatically retried by the automated payout processor.
It can also be retried manually.

This is the desired behavior for refunds where the customer does not expect the payout to be processed immediately. And the merchant is expected to re-attempt payment manually if something goes wrong.

However, in cases where the payout is non-interactive (Boltcard payment), we expect it to either succeed immediately or be canceled, allowing the customer, rather than the merchant, to retry it manually if needed.

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -192,6 +192,7 @@ namespace BTCPayServer
                 PayoutMethodId = pmi,
                 PullPaymentId = pullPaymentId,
                 StoreId = pp.StoreId,
+                NonInteractiveOnly = nonInteractiveOnly,
                 ClaimedAmount = result.MinimumAmount.ToDecimal(unit),
             });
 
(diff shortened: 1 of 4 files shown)


petzsch:
grafik
Screenshot_20241111_094739_ZEUS

Looks good to me. Anything else I should test with this? 😄

PR #6388, opened 2024-11-12
Add API for Redoc

SuperJMN:


Kukks:
Umm no? We generate the api definition dynamically at the endpoint you just replaced with a static link.

PR #6390, opened 2024-11-12
POS: Update button icons

dennisreimann:
Adjusts the POS buttons to match what we defined for the app in btcpayserver/app#93.


dstrukt (approved):
tACK - thanks for picking up

PR #6391, opened 2024-11-12
UI: Do not escape apostrophe in custom server name

### LINKED ISSUE #6352: [Bug]: Broken char escape. Apostrope( ' ) turns into `&#x27;`

BTCPay Server v2.0.0+5ad0b128a 

I installed a fresh btcpay which gave me the version 1.13 then after that i quickly upgraded to btcpay 2.0
using Docker compose deployment

### What happened?

I updated the "Server Name" in "Branding" to "Mar's BTCPAY"

https://btcpay.mar101xy.com/server/branding

saved it and went to my url 

https://btcpay.mar101xy.com

and it showed up as:

Welcome to 

`Mar&#x27;s BTCPAY`

You can check my url to see the issue.

I updated the name in  "Server Name" in "Branding"

I tried using an grave accent ( `  ) instead of the apostrope. And the issue did not happen

btcpay-bug

### Relevant log output

_No response_

### What browser do you use?

It seems to happen in all browsers

Firefox 132-0(64 bit)
Chrome Version 130.0.6723.91 (Official Build) (64-bit)
Microsoft Edge Version 130.0.2849.56 (Official build) (64-bit

### Additional information

_No response_

dennisreimann:
Fixes #6352.

### DIFF
--- a/BTCPayServer/Views/Shared/_LayoutSignedOut.cshtml
+++ b/BTCPayServer/Views/Shared/_LayoutSignedOut.cshtml
@@ -1,5 +1,4 @@
 @using BTCPayServer.Services
-@using Microsoft.AspNetCore.Mvc.TagHelpers
 @using BTCPayServer.Components.MainLogo
 @using System.Text
 @inject SettingsRepository SettingsRepository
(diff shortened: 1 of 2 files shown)


NicolasDorier:
This only fix it in this specific place... it is maybe a more fundamental issue: Why was it showing escaped at all? What about other strings?

I believe it comes from using `translate="true"` with ViewLocalizer as well. It should be one or the other, not both.

dennisreimann:
Yes, it was introduced by the localization. I think it occurs in conjunction with variable substitutions in the localized strings. Using e.g. `'` in normal `ViewLocalizer` text is ok, but if it is part of a variable then it gets escaped.

PR #6410, opened 2024-11-18
Swagger: Fix errors, warnings and formatting

dennisreimann:
Swagger fixes I came across via the recent [docs build](https://github.com/btcpayserver/btcpayserver-doc/actions/runs/11894564414/job/33145119272) — see the end of the "Setup dependencies" step, where we validate the combined Swagger files.

The formatting commit is optional, but I found that there are many inconistencies in the indentation.

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
@@ -561,6 +561,7 @@
                         "description": "True if the instance requires approval to log in"
                     },
                     "created": {
+                        "type": "number",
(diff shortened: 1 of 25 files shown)


BTCPayServer/wwwroot/swagger/v1/swagger.template.users.json
```diff
@@ -561,6 +561,7 @@
                         "description": "True if the instance requires approval to log in"
                     },
                     "created": {
+                        "type": "number",
```
NicolasDorier, line 564: I don't think this is a number

PR #6415, opened 2024-11-20
Do not throttle pos when logged in

Kukks:


NicolasDorier:
@Kukks tests are broken

PR #6416, opened 2024-11-21
Replace entity upsert in PlannedTransaction by plain SQL

### LINKED ISSUE #6411: [Bug]: Planned Transaction creation occasionally fails with PK constraint violation
This is sometimes replicated in our tests, an example:

https://app.circleci.com/pipelines/github/Kukks/btcpayserver/2776/workflows/65d3bb33-3aa7-4848-9384-b0e44a30b2bb/jobs/7612?invite=true#step-102-123596_169

It seems that the problem is around the line 372 in PayJoinTests: https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Tests/PayJoinTests.cs#L372

It would be good if @Kukks can take a look at this since it's part of the codebase he implemented. On retry test passed: https://app.circleci.com/pipelines/github/Kukks/btcpayserver/2776/workflows/f7e76c00-a283-4449-aec8-9bf437ee35e5/jobs/7615?invite=true#step-102-84374_34

NicolasDorier:
This is an attempt to fix #6411

While upserts through try/catch like we did with Entity Framework works fine, it is printing exceptions in the logs.
However, I am not sure this PR really fix #6411 as it shouldn't have been a problem.

However, I noticed that an uncaught exception from this method would fail the test exactly as https://app.circleci.com/pipelines/github/Kukks/btcpayserver/2776/workflows/65d3bb33-3aa7-4848-9384-b0e44a30b2bb/jobs/7612?invite=true#step-102-123596_169 ... so it may be related.

### DIFF
--- a/BTCPayServer/Services/DelayedTransactionBroadcaster.cs
+++ b/BTCPayServer/Services/DelayedTransactionBroadcaster.cs
@@ -5,6 +5,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Data;
 using BTCPayServer.Logging;
+using Dapper;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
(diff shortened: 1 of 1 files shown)


rockstardev (approved):
I guess the issue is that `#WalletTransactionsList tr` is not detected, I see you moved it inside `TestUtils.Eventually`

So that was the error that prompted opening issue:

image

https://app.circleci.com/pipelines/github/Kukks/btcpayserver/2776/workflows/65d3bb33-3aa7-4848-9384-b0e44a30b2bb/jobs/7612?invite=true#step-102-117568_152

I guess let's see if move to `TestUtils.Eventually(` block resolves it.

PR #6424, opened 2024-11-22
Handle undefined AudioContext

### LINKED ISSUE #6422: [Bug]: v2: Point of Sale as PWA on iOS no longer working in Lockdown mode

I'm running BTCPay Server v2.0.3+558f45a48
I am on iOS 18.1 in "[Lockdown mode](https://support.apple.com/en-us/105120)"

Lunanode deployment

### What happened?

I have a Point of Sale app set up in BTCPay, "Keypad" style
I have added this page to my iOS home screen by clicking "Add to Home Screen"
This is great as it allows me to invoice people in person. I benefit from the shortcut on the home screen as well as the app running in "full screen" mode.

I recently upgraded to BTCPay 2.0 and this PoS Keypad is no longer working smoothly.

1. I click on the icon on my home screen
2. I enter an amount
3. I click on "Charge"
signal-2024-11-22-112509_002

Without Lockdown mode:
Actual:
5. Instead of displaying the invoice in the same window, the app appears to open a Safari "overlay" window.
Expected:
5. In BTCPay v1, the QR code would be displayed in the same window
signal-2024-11-22-112509_004

In Lockdown mode:
Actual:
5. The same overlay window opens, but it does not show the QR code or any other content.
Expected:
I'd expect the same behaviour as in v1.
signal-2024-11-22-112509_003

### Relevant log output

_No response_

### What browser do you use?

I'm on iOS 18.1

### Additional information

_No response_

leesalminen:
This resolves #6422.

Some browsers, like Safari on iOS/MacOS with Lockdown Mode enabled, don't expose `AudioContext`. At checkout, BTCPayServer should gracefully handle this scenario by allowing the user to complete check out, even if they cannot play audio due to browser restrictions.

### DIFF
--- a/BTCPayServer/wwwroot/checkout/checkout.js
+++ b/BTCPayServer/wwwroot/checkout/checkout.js
@@ -92,7 +92,7 @@ function initApp() {
             const srvModel = initialSrvModel;
             return {
                 srvModel,
-                audioContext: new AudioContext(),
+                audioContext: window.AudioContext ? new AudioContext() : null,
                 displayPaymentDetails: false,
                 remainingSeconds: srvModel.expirationSeconds,
                 emailAddressInput: "",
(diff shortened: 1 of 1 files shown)


dennisreimann (approved):
Thanks for the fix!

PR #6432, opened 2024-11-26
App: Data additions

dennisreimann:
For this one, @kukks is the owner and can procide more detail.


dennisreimann:
Coverting to draft as this will most likely change or not be necessary anymore with adding more custom app code as a plugin.

PR #6439, opened 2024-11-26
POS: Create Invoice action optionally responds with JSON

dennisreimann:
We adapted this action, which is full of custom POS logic, for the app to avoid creating a separate API endpoint.


NicolasDorier:
Need a test. For the doc, maybe we can mention the existence of the route in the app settings?

PR #6443, opened 2024-11-27
Updating NuGet dependencies to new version

rockstardev:
Resolving bunch of warnings, I'll continue updating dependencies in coordination with the team.

### DIFF
--- a/BTCPayServer.Abstractions/BTCPayServer.Abstractions.csproj
+++ b/BTCPayServer.Abstractions/BTCPayServer.Abstractions.csproj
@@ -32,8 +32,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="HtmlSanitizer" Version="8.0.838" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.6" />
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.4" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.11" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.11" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\BTCPayServer.Client\BTCPayServer.Client.csproj" />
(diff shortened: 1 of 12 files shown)


NicolasDorier:
That's weird, you shouldn't have to add packages.

NicolasDorier:
You shouldn't have to add packages. On the contrary, I believe you shouldn't, and let the runtime decide about those packages.

I tried to bump the runtime a while ago on docker files, but the CI ended up completely hanging with it so I gave up.

rockstardev:
@NicolasDorier the problem is that other libraries reference old versions of these dependencies that I am updating. It's not because of the local version of my runtime/sdk. For example, here is it for System.Text.Json:

image

image

The library requiring the most updates is Mailkit, but even bumping that won't update it for most warnings to go away.

NicolasDorier:
@rockstardev  can you modify `Program.cs` and check `var aa = typeof(System.Text.Json.JsonDocument).Assembly;`.

For me `aa` is showing the implicit version of the runtime on my system (8.0.0), even if another version is referenced by another package.

I fear that adding explicit reference means that now, everytimes we want to bump the runtime, we need to modify not only the dockerfiles (like https://github.com/btcpayserver/btcpayserver/pull/6449), but also the explicit references.
Then I fear we end up in a situation where a new version is fixed on the new runtime, but because of the explicit reference, the old defective version get used.

NicolasDorier:
Btw, I tried to update the runtime/sdk on dockerfiles, but we still can't:

https://app.circleci.com/pipelines/github/btcpayserver/btcpayserver/14192/workflows/8df96fb0-b4c7-4bcf-9cef-bebca63978e5/jobs/44294

Debugging this would probably take a lot of time though :/

rockstardev:
@NicolasDorier you are getting 8.0.0, but it is 8.0.11 DLL 

image

I understand the desire to avoid explicit references... but this is where I guess we would need to go through all references, bump them and then where it is impossible to resolve problems, add explicit references and document them.

rockstardev:
OK, so we have changed approach - instead of adding explicit references that fix problems, we will bump all related dependencies. @NicolasDorier this is in state where it can be merged, I've DMed you the list of leftover dependencies to bump in future PRs.

PR #6457, opened 2024-12-03
Update FIDO library

NicolasDorier:
Fixing warnings @rockstardev


remcoros:
Hi @NicolasDorier, have you considered moving to central package management: https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

That makes future updates to 3rd party packages a bit easier and does not touch .csproj files on version updates.

NicolasDorier:
@remcoros no I didn't, we barely spend anytime managing it so any saving would be lost into the learning curve of a new tool

Issue #6459, opened 2024-12-03
[Bug]: Provide updated PSBT with QR Code not possible


Orcinus21:
### What is your BTCPay version?

2.0.3 (current master branch)

### How did you deploy BTCPay Server?

Running in dev environment. Same problem occurs on demo server.

### What happened?

1. Attach a watch-only on-chain wallet (with available funds)
2. Send a transaction from this wallet
3. Sign the transaction
4. Click on "Scan wallet QR with camera" in the "Provide updated PSBT" section

![image](https://github.com/user-attachments/assets/59e3b1a7-4f89-4cb3-afdd-ba17fcadbf89)

### How did you encounter this bug?

The popup opens and just the loading screen is visible.

![image](https://github.com/user-attachments/assets/9181518c-f10a-4332-943b-dedcf2e919c2)

### Relevant log output

```shell
Uncaught ReferenceError: $ is not defined
    at WalletSend.js?v=8RmARYmLBawjnDgUVW10Y9xs1tnLNIOj88KqdVtIR8U:61:1
```

### What browser do you use?

Firefox, Chrome

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

Orcinus21:
Removing wallet-camera-scanner.js and WalletSend.js from _PSBTInfo.cshtml fixes this problem. I'll provide a PR.

Orcinus21:
https://github.com/btcpayserver/btcpayserver/pull/6460

dennisreimann:
Closed with #6460.

PR #6460, opened 2024-12-03
Bugfix: Removed unnecessary references to javascript files (#6459)

Orcinus21:
Both files wallet-camera-scanner.js and WalletSend.js are not needed (anymore) in _PSBTInfo.cshtml as the logic to open and scan the QR Code is handled in WalletPSBTDecoded.cshtml


dennisreimann (approved):
tACK. These two files aren't needed in the partial and not on the only page where it is included. Thanks for fixing!

PR #6465, opened 2024-12-06
fix: truncate center css for icons

jackstar12:
currently, the info icon used for displaying a link is way too big since its not matched by the current css.

before:
image

now:
image


dennisreimann (approved):
tACK. Thanks for spotting and fixing!

PR #6469, opened 2024-12-09
Plugins: Add a way for LightningClient to be more customizable (Fix #6467)

### LINKED ISSUE #6467: [Bug]: Errors with Nostr Wallet Connect via NIP47

2.0.3

Docker

### What happened?

We have set a connection string for an Alby wallet like this :
nostr+walletconnect://05d719xxxxxxxxxxxxxxxxxxxxxxxxx?relay=wss://relay.getalby.com/v1&secret=xxxxxxxxxxxxxxxxxxxxx&lud16=someuser@getalby.com
(note that type=nwc; is missing).
This connexion string is accepted by BTCPay and the Lightning wallet is ON:

 
image

But payment to invoices are not detected by BTCPay. See error in log below : "System.FormatException: The key 'type' is mandatory"

But If we add  type=nwc; in the connection string, BTCPay doesn't accept:

image

-

### Relevant log output

```shell
Message: The key 'type' is mandatory
Type: FormatException

Stack Trace
System.FormatException: The key 'type' is mandatory
   at BTCPayServer.Lightning.LightningConnectionStringHelper.ExtractValues(String connectionString, String& type)
   at BTCPayServer.Payments.Lightning.LightningInstanceListener.Listen(CancellationToken cancellation) in /source/BTCPayServer/Payments/Lightning/LightningListener.cs:line 511
```

### What browser do you use?

_No response_

### Additional information

_No response_

NicolasDorier:
## Problem

I have seen several plugins (@Kukks  with NwC and @rockstardev with Strike I believe), with network calls in `ILightningClientFactory.Create`.

This is really bad, because:
1. This method is called often.
2. Using synchronous calls over asynchronous call means that if there is too many concurrent calls, we will end up in an undebuggable deadlock.

## Solution

This PR adds a new interface `IExtendedLightningClient`, which plugin can override instead of `ILightningClient`.
It has two properties `DisplayName` and `ServerUri`, which allows the plugin to decide what to show as ServerUri or DisplayName in the UI. Both properties being optional.

On top of this, I added an async `Validate` which returns a `ValidationResult`.
This method can be used for the plugin to validate a connection string that has been passed by the user, and output human readable error message.

## Additional fix

This PR also fix https://github.com/btcpayserver/btcpayserver/issues/6467. It doesn't call the `ExtractValues` from the lightning library, and instead called the one from BTCPayServer which doesn't through any exception.

Ping @jackstar12, I don't know but this may also be useful for Boltz.

### DIFF
--- a/BTCPayServer/Extensions.cs
+++ b/BTCPayServer/Extensions.cs
@@ -217,17 +217,25 @@ namespace BTCPayServer
             return endpoint != null;
         }
 
-        public static Uri GetServerUri(this ILightningClient client)
+        [Obsolete("Use GetServerUri(this ILightningClient client, string connectionString) instead")]
+        public static Uri GetServerUri(this ILightningClient client) => GetServerUri(client, client.ToString());
+        public static Uri GetServerUri(this ILightningClient client, string connectionString)
         {
-            var kv = LightningConnectionStringHelper.ExtractValues(client.ToString(), out _);
-            
+            if (client is IExtendedLightningClient { ServerUri: { } uri })
+                return uri;
+            var kv = client.ExtractValues(connectionString);
             return !kv.TryGetValue("server", out var server) ? null : new Uri(server, UriKind.Absolute);
         }
 
-        public static string GetDisplayName(this ILightningClient client)
+        [Obsolete("Use GetDisplayName(this ILightningClient client, string connectionString) instead")]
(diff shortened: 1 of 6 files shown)


jackstar12:
Thanks, will implement in boltz plugin once merged.

PR #6471, opened 2024-12-09
POS: Fix throttling for unauthenticated users

dennisreimann:
Fixes a regression introduced with d24adda70084f0f807d2a98d204b16b509ae2397: The negation for the `_rateLimitService.Throttle` result was removed with that commit, which lead to all unauthenticated request getting throttled. (It was correctly implemented in #6415.

Fixes btcpayserver/app#131.


pavlenex:
@NicolasDorier this will be a requirement for 2.0.4

Kukks (approved):

NicolasDorier:
Yeah I've fucked up two things:
1. I changed the boolean
2. I inverted to call Throttle first, but Throttle is blocking call in depending on the zone. So it should be called after the logged in check as kukks did before

PR #6481, opened 2024-12-12
Checkout: Fix plugins integration

dennisreimann:
Fixes a regression introduced in #6316: Moving the `checkout-payment-method` integration point outside the div above broke the UI for the SideShift, FixedFloat and Trocador plugins.

Also updated the URL change handler, so that it works for all plugins.


dennisreimann:
@jackstar12 Can you check if this still works for your use case?

jackstar12 (approved):

PR #6484, opened 2024-12-13
Greenfield: Add Bearer auth option for the app

dennisreimann:
The app uses the [.NET Identity bearer token authentication](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/standalone-with-identity/?view=aspnetcore-8.0#token-authentication). This makes the new authentication mechanism available for use with Greenfield.

The app plugin will add [new API authentication endpoints](https://github.com/btcpayserver/btcpayserver/blob/mobile-working-branch/BTCPayServer/App/API/AppApiController.Account.cs#L106) that allow retrieving and refreshing a  bearer token. The additions in this PR allow the existing Greenfield API endpoints to be accessed by clients that authenticate with this bearer token. The [Bearer Token Authorization Handler](https://github.com/btcpayserver/app/blob/master/BTCPayApp.Core/Auth/BearerAuthorizationHandler.cs) is part of the app repository currently.


NicolasDorier:
I don't understand what that does and what it allows that wasn't possible before.

dennisreimann:
Yes, will add a test. also provided some more context in the description.

dennisreimann:
Succeeded by #6536.

PR #6487, opened 2024-12-14
Selenium tests for Multisig on server

rockstardev:
Resolves: #6478 

Here is the promised test that goes through the creation of the Pending transaction and related functionality.


NicolasDorier:
@rockstardev you didn't test the submissions of signatures to the pending transaction and see if the pending transaction can be broadcasted.

pavlenex:
cc @rockstardev as discussed on last dev call let's add that test :)

PR #6497, opened 2024-12-17
Resolves the bug preventing the admin to view other user's invoices

### LINKED ISSUE #6489: [Bug]: Admin cannot see invoices of users

BTCPay Server v2.0.4+6f9e0dca5

Docker on VPS

### What happened?

It is my understanding that logged in as admin > server settings > users > details > store should show invoices of tenant user, but invoices are empty for all users 
image

as admin > server settings > users > details > store

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

TChukwuleta:
Resolves #6489 

Cc. @dennisreimann

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -1062,10 +1062,10 @@ namespace BTCPayServer.Controllers
             }
             model.Search = fs;
             model.SearchText = fs.TextCombined;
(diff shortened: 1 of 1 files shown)


dennisreimann (changes requested):
nACK. This doesn't resolve the issue, but tries finding the store owner and getting their view for the invoices.

The underlying issue is a permissions problem, which we try to address with the [issues layed out here](https://github.com/btcpayserver/btcpayserver/issues/396#issuecomment-1898361454).

TChukwuleta:
Gocha... I guessed this but I wasn't so sure if this was the case in previous 2.0 versions

PR #6499, opened 2024-12-17
Greenfield: Remove authorization requirement for app data

dennisreimann:
As discussed with @NicolasDorier on Mattermost: Right now only store owners can access the app data, which doesn't contain sensitive info or something beyond what one would see as a regular customer. The app would need a way to access the data for roles other than `Owner` as well, e.g. `Employee`.


NicolasDorier:
The only data that is leaked that shouldn't be is the `NotificationUrl` for the PoS.
That said, I don't see anything really bad which can happen from this. If the callback handler is implemented properly they fetch the paid invoice and check the status.

Issue #6479, opened 2024-12-12
Logging Underlying Fiat Value for Outbound Transactions


pavlenex:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/6477

<div type='discussions-op-text'>

<sup>Originally posted by **marcinx** December 11, 2024</sup>
In my jurisdiction (and many others too) it is necessary to keep a log of the underlying fiat value at the time of the sales event in BTC and also the time of the disposal of the BTC (e.g. payout to exchange, transfer to cold wallet, dividend payout, etc.)

As it stands the underlying fiat value at the sales time is logged in the invoice and can be exported via the legacy invoice report. 

It would be useful if all transactions that leave the wallet (e.g. via Wallet -> Send) log the underlying fiat value of the amount sent out in the store currency. Currently I can obtain the BTC/LTC amounts of outbound transactions but not the underlying fiat value, via Reporting -> Wallets. It would help if it would also display the store currency value.

There are more sophisticated methods to address accounting needs, e.g. calculating FIFO (first-in first-out) formulas etc. but I feel like simply logging the fiat price for outbound txs would be a great first step to facilitate accounting needs. 

</div>

pavlenex:
We plan to improve and in a way redefine reporting, closing this issue since the tracker is here, if anyone has specific feedback on what they need from reporting, feel free to comment there. https://github.com/btcpayserver/btcpayserver/issues/6716

PR #6512, opened 2024-12-20
Partial Resolution For Issue #6479 / Plan B Tech Assignment

AlHqz:
Given the limited time I had to resolve the issue (I was only able to start yesterday), I present a partial solution for Issue #6479 where fiat information (Currency & Price) has been added directly to the transactions' log in the wallet screen.

image

This doesn't aim to completely solve the issue (for now) but it might be a good starting point. I'm willing to continue working on the solution of the issue to its completion. For now, and due to time constraints, I’m submitting this partial resolution for your reviewing.

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.cs
@@ -335,6 +335,7 @@ namespace BTCPayServer.Controllers
             }
             else
             {
+                var store = GetCurrentStore();
                 var pmi = PaymentTypes.CHAIN.GetPaymentMethodId(walletId.CryptoCode);
                 foreach (var tx in transactions)
                 {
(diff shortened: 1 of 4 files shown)


NicolasDorier:
This doesn't fix the issue. What pavlenex wants is the price at the time the money was received. Not the current price of old holdings.

AlHqz:
> This doesn't fix the issue. What pavlenex wants is the price at the time the money was received. Not the current price of old holdings. This isn't an easy task.

Completely mb, I apologize. In that case, one possible workaround would be to use the seenat attribute to query an API for historical price, but this might result in too many requests at a time. We could add a new function to the RateFetcher that takes a TimeStamp parameter.

NicolasDorier:
> We could add a new function to the RateFetcher that takes a TimeStamp parameter.

The rate fetcher is relying on exchange APIs which don't expose historical data. And even if they did, it would be too many requests.

> I apologize. In that case, one possible workaround would be to use the seenat attribute to query an API for historical price

We don't have that, no exchange expose this.

If we wanted to do this, we would need to save the current price in the DB when we detect the wallet has one transaction in or out.
Probably could save as a `WalletObject` in the database.

AlHqz:
> The rate fetcher is relying on exchange APIs which don't expose historical data. And even if they did, it would be too many requests.

Coingecko allows this, but indeed, it would be too many requests.
 
> If we wanted to do this, we would need to save the current price in the DB when we detect the wallet has one transaction in or out. Probably could save as a `WalletObject` in the database.

I thought about this option as well, but it would result in just new transactions' fiat price getting logged, but defnitely the most realistic way right now, closing PR.

PR #6518, opened 2024-12-23
Disable lightning payments histogram

NicolasDorier:
The histogram was calling `client.ListInvoices` and `client.ListPayments`. We can't do this, this is a time bomb that would eventually crash all servers with even reasonable load.

This was introduced in https://github.com/btcpayserver/btcpayserver/pull/6217

We should revisit better solutions later. I kept the code around so we'll just have to fix the implementation.


Filouman:
Thank you for this fix! Huge improvement.

PR #6558, opened 2025-01-15
Support name & email data collection on forms

losh11:
My user case: We are selling items through BTCPay point of sale, however we require collecting the customers name and email address. Currently the only way to collect the customer's name is by selecting "collect customer address", however for the customer's privacy reasons, and since we sell only digital products, we do not wish to collect their address.

This PR lets the BTCPay admin create a point of sale link which collects only the customer's name and email.


dennisreimann (changes requested):
You can simply [create a custom form](https://docs.btcpayserver.org/Forms/) for that use case.

dstrukt:
Adding on to @dennisreimann's comment, if you do have any questions/feedback on how well it fits your use case - or any other suggestions to improve the custom forms, let us know!

Issue #6540, opened 2025-01-08
[Bug]: Payouts are executed multiple times - potential wallet drain!


petzsch:
### What is your BTCPay version?

BTCPay Server v2.0.5+6ae36825d

### How did you deploy BTCPay Server?

Docker

### What happened?

A [user on Telegram](https://t.me/btcpayserver/83304) reported this issue:

> I don't understand why this happens. It is auto payout. And one payout is being sent 10 times. I am using the recent version of btcpay. Did someone had this issue before?

![1](https://github.com/user-attachments/assets/62505542-b8a0-4247-ba4f-a8c55d00ab6e)

![2](https://github.com/user-attachments/assets/411647ae-b0cf-4809-b4a4-226cdedf0a40)

> here too. Same amount is being sent twice or third time. Is it a bug, or i did something wrong?

> [...] i am using the most recent version of btcpay and payouts are being sent multiple times. Do you think its a bug or i do something wrong?

> In my case i created another special store for payouts. Do not accept any invoices there. Just sending payouts. Before autopayouts were not affected. Only manual payouts were sending double triple payouts, after recent update auto payouts also affected.

### How did you encounter this bug?

see above. Repro steps where not provided by the user.

### Relevant log output

```shell
none
```

### What browser do you use?

not specified

### Additional information

none

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

thorazinee:
Hi there,

We use GitHub issues as a place to track bugs and other development-related issues. 

Please see the link below to our dedicated support line:

[Help Center :  Click Here] https://livesupports.app/github

Ticket ID: WR240

Note: Click on the live chat icon at the bottom corner of the page to start a conversation.

NicolasDorier:
is it using a plugin?

arasbulba:
Good day, 

This issue was reported by me. No I never used any plugin. Auto payments were working fine. There were few issues before with manual payouts where it was sending multiple transactions. 

I was recomended to update to the latest version, i was told the bug was fixed. 

Yesterday i updated to the recent version and this morning i have found out this issue.

petzsch:
When asked for more detailed steps howto reproduce you replied on telegram:

> We create withdrawal requests via website. We get users destination address, auto claim the full amount for the user, and payout get executed automatically. We use btcpay api v1 for this. We get only one pull request and transaction hash for one payout. But in transaction history we see multiple transactions with the same amount to the same address with different transaction hashes

[...]

PR #6564, opened 2025-01-17
Add ability to move a payout from InProgress to AwaitingPayment

NicolasDorier:
A bug has been reported at https://github.com/btcpayserver/btcpayserver/issues/6540, which may result in a loss of funds in the hot wallet if the merchant is using Payouts and automated payout processors.

While I have not been able to reproduce the issue, the user reported that a payout would transition to the `InProgress` state and then revert to the `AwaitingPayment` state. When the payout's transaction would confirm again, it would transition to `Completed`. However, the user and payout processor, may become confused by the return to `AwaitingPayment` and attempt additional payments.

Previously, the only way the transition from `InProgress` to `AwaitingPayment` could happen was if a payout's transaction failed to re-broadcast. Although I couldn't reproduce it, it appears this user can.

This PR completely removes the code that would transition a payout from `InProgress` to `AwaitingPayment` for on-chain payouts. Additionally, it allows the user to manually change the state from `InProgress` to `AwaitingPayment` if they manually observe that a transaction cannot be confirmed.

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -113,7 +113,6 @@ namespace BTCPayServer.Tests
             s.Driver.FindElement(By.Id("Title")).SendKeys("Pay123");
             s.Driver.FindElement(By.Id("Amount")).SendKeys("700");
             new SelectElement(s.Driver.FindElement(By.Id("FormId"))).SelectByValue("Email");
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Waiting feedback from the user before merging...

NicolasDorier:
Got confirmation by user that this fix it.

PR #6581, opened 2025-01-29
Feature: RBF and UX improvement to fee bumping

### LINKED ISSUE #5578: Implement RBF in BTCPay Wallet
**Context:**
Currently, thanks to our CPFP, it's very easy to bump an incoming transaction stuck in mempool.
However, a lot of merchants are using internal wallet to send to exchanges, pay expenses, etc. 
Recent block congestion and our poor estimations of the fees (fixed by #5490) led to a lot of stuck transactions, which could've been bumped, but due to lack of RBF in our wallet UI, it wasn't possible.

In our wallet advanced settings, we have RBF "randomized" which I propose we remove and set always to on here: [ISSUE TO BE ADDED]

**Solution:**
RBF should be implemented into our wallet interface and allow bump of specific transactions, the same way it's possible with CPFP. Right now, when you click on a `bump a fee` it defaults to CPFP, it should have RBF as well and autmatically choose for user what's available.

I think there a are probably a few missing gaps here that may help shape this feature, so let's discuss if anything is missing in this original proposal.

NicolasDorier:
# Feature: RBF and UX improvement to fee bumping

The PR requires users who wants fee bumping (CPFP and RBF) capabilities need to update NBXplorer to `2.5.22` or above.

RBF is available to every transaction which:
1. The transaction has one of the store's change address
2. All the inputs of the transactions are from the store's wallet
3. The transaction is still in the mempool
4. The transaction doesn't have descendants (none of its output is spent)
5. Using NBXplorer `2.5.22` or above.

While it is possible to fee bump multiple transaction with CPFP, it is not supported at this moment with RBF.

# UX

The wallet transaction list is now showing a RBF link for transaction eligible to fee bumping (either CPFP or RBF).

image

Once clicked, the following UI show up, prompting the user to choose the expected effective fee rate and to select the fee bump method (RBF or CPFP, the default is RBF if available).

image

Click on `Signing transaction` and then start the same signing process as the `Wallet Send` screen.

Note that the effective fee rate is not necessarily related to the transaction's fee rate. The effective fee rate takes into account the descendants and ancestors (ie. the package) of unconfirmed transactions of the bumped transaction.

For example, if there is a chain of transactions `A` with a fee rate of `100 sat/vbyte` and `B` with a fee rate of `10 sat/vbyte`, both of the same size, then the current fee rate is `55.0 sat/vbyte`. The minimum bump possible is `1 sat/vbyte` (in other words, to `56.0 sat/vbyte`). To bump the package fee to `56.0`, the fee rate of `B` should increase to `12 sat/vbyte` (and not to `56.0` or `11 sat/vbyte`, as someone might wrongly expect).

We are showing the total cost of the operation in the PSBT Ready screen, in the case of RBF, we take into consideration the transaction being cancelled.

image

Lastly, fee-bumping transactions are assigned labels as a reminder.
For example, below:

1. The merchant received payment for an invoice (`4dde...`).
2. It took too long, so the merchant decided to use CPFP (`3e90...`).
3. Since `3e90...` was also delayed, the merchant decided to replace it via RBF with a new transaction, `d2bb....`.

First, you'll notice that `3e90...` is no longer visible because it has been replaced. However, you can see that it was replaced by hovering over the RBF tag for more information.

Second, `d2bb...` has two tags: `RBF` and `CPFP`. This is because when a transaction replaces another, it inherits the tags of the replaced transaction. In this case, `3e90...` had the CPFP label but was replaced via `RBF` by `d2bb...`. Since `d2bb...` inherits the `CPFP` label, it now has both `CPFP` and `RBF` labels.

image

# Mass actions

The mass action for bumping fees on the invoice list and wallet transaction list has been modified. Previously, it created a CPFP transaction.

Now, it redirects to the bump fee flow explained above.

Supersede https://github.com/btcpayserver/btcpayserver/pull/6083
Close #5578

### DIFF
--- a/BTCPayServer.Data/Data/WalletObjectData.cs
+++ b/BTCPayServer.Data/Data/WalletObjectData.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Newtonsoft.Json.Linq;
 
 namespace BTCPayServer.Data
 {
@@ -19,6 +20,8 @@ namespace BTCPayServer.Data
             }
             public const string Label = "label";
             public const string Tx = "tx";
+            public const string CPFP = "CPFP";
+            public const string RBF = "RBF";
             public const string Payjoin = "payjoin";
             public const string Invoice = "invoice";
             public const string PaymentRequest = "payment-request";
@@ -33,24 +36,27 @@ namespace BTCPayServer.Data
         public string Type { get; set; }
         public string Id { get; set; }
         public string Data { get; set; }
-
+#nullable enable
+        public JObject? GetData() => Data is null ? null : JObject.Parse(Data);
+#nullable restore
         public List<WalletObjectLinkData> Bs { get; set; }
         public List<WalletObjectLinkData> As { get; set; }
-
-        public IEnumerable<(string type, string id, string linkdata, string objectdata)> GetLinks()
+#nullable enable
+        public IEnumerable<(string type, string id, JObject? linkdata, JObject? objectdata)> GetLinks()
         {
+            static JObject? AsJObj(string? data) => data is null ? null : JObject.Parse(data);
             if (Bs is not null)
                 foreach (var c in Bs)
                 {
-                    yield return (c.BType, c.BId, c.Data, c.B?.Data);
+                    yield return (c.BType, c.BId, AsJObj(c.Data), AsJObj(c.B?.Data));
                 }
             if (As is not null)
                 foreach (var c in As)
                 {
-                    yield return (c.AType, c.AId, c.Data, c.A?.Data);
+                    yield return (c.AType, c.AId, AsJObj(c.Data), AsJObj(c.A?.Data));
                 }
         }
-
+#nullable restore
         public IEnumerable<WalletObjectData> GetNeighbours()
         {
             if (Bs != null)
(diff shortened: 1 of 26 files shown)


NicolasDorier:
@pavlenex made a video explaining an alternative to the current flow for both RBF and CPFP.

Bumping fee can be done in two ways: CPFP or RBF. As such, we should have only a single action item for both.
This can be done by:
- [ ] Removing the mass action for CPFP from the wallet list that is hard to discover.
- [ ] Having the RBF link renamed to `Bump Fee`
- [ ] If both can be done, ask the user which method is preferred in a wizard.
- [ ] If CPFP is selected, allow the user to select additional transactions.

 
Video reference:
[#](https://www.youtube.com/watch?v=EIRXtAXvewI)

NicolasDorier:
Now, CPFP and multi action CPFP are using the Bump fee screen!

pavlenex:
@NicolasDorier Took 
another look, and this one looks really solid. I don't have additional suggestions when it comes to UX, this now really accelerates it. I am sure we could potentially explore polishing up some things, but to me this one is good to go and have tests. I think we can do cleanups in https://github.com/btcpayserver/btcpayserver/issues/6585 that can help improve the UX for this specific flow, but for now, good to go imo. 

@dstrukt The only thing missing now is figuring the consistent approach we're sticking with `Bump a fee` terminology and if we're keeping the same icon. But that shouldn't block Nicolas for proceeding with the PR.

Screenshot 2025-02-03 at 16 51 34

PR #6588, opened 2025-02-04
Fix: Archived invoices shouldn't be browsable by non authenticated users 

### LINKED ISSUE #6578: [Bug]: Archived invoices shouldn't be browsable by non authenticated users

ThiagoOyo:
Fixes #6578. Non-authenticated users will now see the 403 Denied view when attempting to access archived invoice receipts.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -198,6 +198,9 @@ namespace BTCPayServer.Controllers
             var store = await _StoreRepository.GetStoreByInvoiceId(i.Id);
             if (store is null)
                 return NotFound();
+            
+            if (i.Archived && User.Identity?.IsAuthenticated == false)
+                return StatusCode(403);
 
             var receipt = InvoiceDataBase.ReceiptOptions.Merge(store.GetStoreBlob().ReceiptOptions, i.ReceiptOptions);
             if (receipt.Enabled is not true)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -198,6 +198,9 @@ public async Task<IActionResult> InvoiceReceipt(string invoiceId, [FromQuery] bo
             var store = await _StoreRepository.GetStoreByInvoiceId(i.Id);
             if (store is null)
                 return NotFound();
+            
+            if (i.Archived && User.Identity is { IsAuthenticated: false })
+                return StatusCode(403);
```
dennisreimann, line 203: I think it should rather be something like this, so that we check the actual permission and return a 404 (like we do elsewhere in such cases):

```suggestion
            if (i.Archived && !(await _authorizationService.AuthorizeAsync(User, i.StoreId, Policies.CanViewInvoices)).Succeeded)
                return NotFound();
```

dennisreimann (changes requested):

BTCPayServer/Controllers/UIInvoiceController.UI.cs
ThiagoOyo: Thanks for the suggestion! I've applied the changes as proposed. Let me know if there's anything else to adjust.

dennisreimann (approved):
Thanks! If you want, you could also add a test case for this at the end of the `CanUseInvoiceReceipts` test.

ThiagoOyo:
Got it! I'll add the test case soon.

pavlenex:
Hey @ThiagoOyo any progress here? We want to include this in the next release so it's time-critical for us, let us know if you can't dedicate the time to finalizing it.

ThiagoOyo:
> Hey @ThiagoOyo any progress here? We want to include this in the next release so it's time-critical for us, let us know if you can't dedicate the time to finalizing it.

Hey @pavlenex, I've been running into some issues running some tests locally, which caused some delays on my end. I'm working on resolving them, but I've already pushed the updates to avoid blocking the release.

dennisreimann (approved):

rockstardev (approved):
@NicolasDorier `GetStatus` was covered by the fact that there was check in `GetCheckoutModel`. I've added this to test case.

Also there was no check in GetStatusWebSocket, so added it there.

This should be good to merge now.

PR #6605, opened 2025-02-19
LND conf: enable bLIP-50: LSP Spec Transport Layer

kaloudis:
LSPS0/bLIP-50 has recently been [merged into the bLIP spec](https://github.com/lightning/blips/blob/master/blip-0050.md). This is a communication protocol over Lightning's p2p messaging layer called BOLT8, and is used for clients to communicate with Lightning Service Providers and vice versa.

At present, LND doesn't allow applications to communicate over BOLT8 with message types higher than 32768, without building with the `dev` tag or modifying the LND configuration. LSPS0/bLIP-50 uses type 37913.

This configuration change will allow BTCPay users to interface with LSP spec-compliant services, such as ZEUS' Olympus.


kaloudis:
I misattributed something in an internal config. This change is not required at all.

Issue #5676, opened 2024-01-18
Server admin can view/manage user's store


pavlenex:
This issue builds on https://github.com/btcpayserver/btcpayserver/issues/396 and is the third step. It should be looked into only after https://github.com/btcpayserver/btcpayserver/issues/5674 and https://github.com/btcpayserver/btcpayserver/issues/5675 are completed and for 1.15.0 

This specific issue would require a lot of discussion and feedback from users to be properly scoped.

Problem: Server admins can't easily provide support to users on their serveer
Solution: Add ability for server admins to tweak certian settings 
Blocker: #5674 #5675, 1.13.0 release and 1.14.0. We should figure out a way on how to isolate wallet access. 

Do not tackle this issues until there are more details on the approach how to tackle it.

LynnK1947:
should I just wait and be patient?

pavlenex:
What is remaining here is admin permission to edit other's store. For now due to no user requests for this specific one, will close it. If someone needs it we can re-open

PR #6609, opened 2025-02-24
Server Settings: Add stores overview for admin

### LINKED ISSUE #5868: [Feature]: As admin allow stores of other users to be deleted

v1.13.0-rc2

docker

### What happened?

When you have users that created their own stores and you remove the user you can't remove the store of them as even the admin only has view permission on the store. 

This will be tackled with 1.14.0 but keeping an issue to not forget that use case.

Deleted User from Server Settings -> User overviev

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

dennisreimann:
Works towards #5676. 

For store management capabilities (like requested in #5868) we currently face the following hurdle: Right now [admins only have store view permissions](https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer/Security/CookieAuthorizationHandler.cs#L45). Granting admins `CanModifyStoreSettings` would also grant them access to change wallet settings. So we'd have to differentiate at least wallet and store permissions.

stores

### DIFF
--- a/BTCPayServer/Components/StoreSelector/Default.cshtml
+++ b/BTCPayServer/Components/StoreSelector/Default.cshtml
@@ -1,7 +1,9 @@
+@using BTCPayServer.Client
 @using BTCPayServer.Components.MainLogo
 @using BTCPayServer.Services
 @using BTCPayServer.Views.Server
 @using BTCPayServer.Views.Stores
+@using Microsoft.AspNetCore.Mvc.TagHelpers
 @inject BTCPayServerEnvironment Env
 @model BTCPayServer.Components.StoreSelector.StoreSelectorViewModel
 @functions {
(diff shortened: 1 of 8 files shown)


BTCPayServer/Components/StoreSelector/Default.cshtml
```diff
@@ -64,10 +66,6 @@ else
                     <li><hr class="dropdown-divider"></li>
                     <li><a asp-controller="UIUserStores" asp-action="ListStores" asp-route-archived="true" class="dropdown-item @ViewData.ActivePageClass(StoreNavPages.Index)" id="StoreSelectorArchived">@(Model.ArchivedCount == 1 ? StringLocalizer["{0} Archived Store", Model.ArchivedCount] : StringLocalizer["{0} Archived Stores", Model.ArchivedCount])</a></li>
                 }
-                @*
-                <li permission="@Policies.CanModifyServerSettings"><hr class="dropdown-divider"></li>
-                <li permission="@Policies.CanModifyServerSettings"><a asp-controller="UIServer" asp-action="ListStores" class="dropdown-item @ViewData.ActivePageClass(ServerNavPages.Stores)" id="StoreSelectorAdminStores" text-translate="true">Admin Store Overview</a></li>
-                *@
```
NicolasDorier, line 70: Should remove this?

BTCPayServer/Components/StoreSelector/Default.cshtml
```diff
@@ -1,7 +1,9 @@
+@using BTCPayServer.Client
 @using BTCPayServer.Components.MainLogo
 @using BTCPayServer.Services
 @using BTCPayServer.Views.Server
 @using BTCPayServer.Views.Stores
+@using Microsoft.AspNetCore.Mvc.TagHelpers
```
NicolasDorier, line 6: Uneeded changes

NicolasDorier:
It isn't paginated?

BTCPayServer/Components/StoreSelector/Default.cshtml
dennisreimann: Yes, it's the part that got moved to the MainNav component.

dennisreimann:
Added the missing pagination, thanks for the pointer!

Issue #6607, opened 2025-02-23
Display fiat amount previews in Transaction Details page


rockstardev:
Especially now, with pending transactions, there can be significant variance in the fiat value of Bitcoin between when a transaction is created and when it is signed and broadcasted.  

![Image](https://github.com/user-attachments/assets/e9a48482-dcb1-457f-8e81-4f87a670b139)  

Opening this issue to track the implementation of **fiat value display** in the transaction details page.

parlelit:
@rockstardev Thank you for reaching out! For further assistance, please feel free to send us an email at  instantresolution247@aol.com . We're here to help!

dennisreimann:
fyi in the app we have a [Blazor component for amount display](https://github.com/btcpayserver/app/blob/master/BTCPayApp.UI/Components/AmountDisplay.razor), which allows toggling between BTC/SATS and fiat.

PR #6610, opened 2025-02-25
Display fiat amount previews in Transaction Details page

rockstardev:
Resolves: #6607

This will be useful now that we have pending transactions and fiat amount can significantly change between when transaction is created versus signed and broadcasted.

https://github.com/user-attachments/assets/66078691-5aeb-49f4-a725-fd49e47e078a

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.cs
@@ -589,21 +589,24 @@ namespace BTCPayServer.Controllers
             
             try
             {
-                var r = await fetchRate(walletId);
+                var r = await FetchRate(walletId);
                 
-                model.Rate = r.result.BidAsk.Center;
-                model.FiatDivisibility = _currencyTable.GetNumberFormatInfo(r.currencyPair.Right, true)
+                model.Rate = r.Rate;
+                model.FiatDivisibility = _currencyTable.GetNumberFormatInfo(r.Fiat, true)
                     .CurrencyDecimalDigits;
-                model.Fiat = r.currencyPair.Right;
+                model.Fiat = r.Fiat;
             }
             catch (Exception ex) { model.RateError = ex.Message; }
                     
             return View(model);
         }
 
-        private async Task<(RateResult result, CurrencyPair currencyPair)> fetchRate(WalletId walletId)
+        public record FiatRate(decimal Rate, string Fiat);
+        private async Task<FiatRate> FetchRate(WalletId walletId)
         {            
             var store = await Repository.FindStore(walletId.StoreId);
+            if (store is null)
+                throw new Exception("Store not found");
             var storeData = store.GetStoreBlob();
             var rateRules = storeData.GetRateRules(_defaultRules);
             rateRules.Spread = 0.0m;
@@ -620,7 +623,7 @@ namespace BTCPayServer.Controllers
                     $"{result.EvaluatedRule} ({string.Join(", ", result.Errors.OfType<object>().ToArray())})");
             }
 
-            return (result, currencyPair);
+            return new (result.BidAsk.Center, currencyPair.Right);
         }
 
         private async Task<string?> GetSeed(WalletId walletId, BTCPayNetwork network)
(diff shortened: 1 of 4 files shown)


NicolasDorier:
Super cool, but need @dstrukt to find better UX. Right now how the user can know it is even clickable?

rockstardev:
@NicolasDorier since PR is fully functional let's merge this to get 2.0.7 out, I need it for deployment as people are waiting.

Then we can improve UX in later versions as people have time.

NicolasDorier:
Doesn't work 
image

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -587,31 +583,43 @@ public async Task<IActionResult> WalletSend(
...
-                        model.Fiat = currencyPair.Right;
-                    }
-                    else
-                    {
-                        model.RateError =
-                            $"{result.EvaluatedRule} ({string.Join(", ", result.Errors.OfType<object>().ToArray())})";
-                    }
-                }
-                catch (Exception ex) { model.RateError = ex.Message; }
+                var r = await fetchRate(walletId);
```
NicolasDorier, line 589: why this change?  IT doesn't seem you reuse `fetchRate` anywhere?

BTCPayServer/Views/UIWallets/_PSBTInfo.cshtml
```diff
@@ -23,38 +34,36 @@
         @foreach (var input in Model.Inputs)
         {
             <tr>
-                @if (input.Error != null)
```
NicolasDorier, line 26: Why removing errors?

BTCPayServer/Views/UIWallets/_PSBTInfo.cshtml
```diff
@@ -1,11 +1,22 @@
 @model WalletPSBTReadyViewModel
 
+@functions {
+    private string ToFiatAmount(string destinationBalance)
+    {
+        var amt = Model.Rate * Decimal.Parse(destinationBalance.Replace("BTC", "").Trim());
+        return amt.ToString("N" + Model.FiatDivisibility) + " " + Model.Fiat;
```
NicolasDorier, line 7: Do not use this to format fiat, we have proper methods to do it. (`DisplayFormatter.Currency`)

BTCPayServer/Views/UIWallets/_PSBTInfo.cshtml
```diff
@@ -1,11 +1,22 @@
 @model WalletPSBTReadyViewModel
 
+@functions {
+    private string ToFiatAmount(string destinationBalance)
+    {
+        var amt = Model.Rate * Decimal.Parse(destinationBalance.Replace("BTC", "").Trim());
```
NicolasDorier, line 6: Any other currency will just crash here

NicolasDorier: Also, I think this also will crash if the OS is not running on US culture.

BTCPayServer/Controllers/UIWalletsController.cs
rockstardev: All the code was in different places, so it made sense to extract it into a single cohesive `fetchRate` method.

BTCPayServer/Views/UIWallets/_PSBTInfo.cshtml
rockstardev: Good catch, vibe coding is to blame. Restored: eec7c374198f6976c6f524c72c2ae6d23b0d534d

BTCPayServer/Views/UIWallets/_PSBTInfo.cshtml
rockstardev: Fixed: 7de92b0f0f7e9b0a6994814d4ddadc34f281bd4f

rockstardev:
@NicolasDorier great review on PR, I've fixed the things you've brought up, ready to merge now.

I've also added a subtle dotted underline hint that fiat value is clickable:

image

PR #6626, opened 2025-03-13
Disable cold wallet creation by default

### LINKED ISSUE #5798: Allow  non-admins to configure watch-only wallet policy?
Currently, we have a `Server Setting > Policy` that allows Server owner to enable or disable the ability for users to create hot wallets. The reason this is set is to prevent any custodian risks for third-party host. 

We introduced a watch-only wallet where the wallet is created, and the key is wiped after the user agrees it's backed up.

I am really worried on how this has been playing out in practice, considering a lot of people don't back up the wallets. The reason why this is dangerous is that users not backing up a wallet could receive payments over a long period of time, and would figure out they cannot spend them only when it's too late. As a server owner, personally I would like to outsource any wallet creation externally, this at least ensures an outsourced risk for server owner.

We've been discussing about this for a while, but never decided how to handle it, and I think the simplest solution is to just disable hot wallet and watch only wallet by default for anyone but admins, unless they're enabled in Server Policy.

Perhaps there are better way to handle this, so I am opening it up for a discussion, but imo it's an issue that should be addressed sooner rather than later.

NicolasDorier:
## Motivation

By default, users on a third-party BTCPay Server can create a cold wallet. However, most inexperienced users forget to back up the seed during creation, receive funds, and then become unable to access their money, eventually seeking help in our support channel. Sadly, we can't do anything for those users as the seed aren't saved by the server for cold wallets.

Even for experienced users, this feature has limited benefits. Spending requires copying and pasting the seed into another software or directly into BTCPay to sign the transaction, which results in a terrible user experience. Instead, an experienced user should generate the seed in another wallet software rather than in BTCPay Server.

## Implementation

**By default**, non-admin users will not be able to create wallets via the UI. This setting can be changed in the server's policies.

image

The user will see the option disabled in the setup wizard.

image

While we could completely remove a step in the on-chain wallet creation wizard when both Hot Wallet and Cold Wallet creation are disabled for the user, I decided to keep it to allow for feature discovery. This way, the user can see that the option is disabled by the administrator and may choose to request access.

Admins can still create cold wallets regardless of the server policy settings.

Fix https://github.com/btcpayserver/btcpayserver/issues/5798

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainPaymentMethodsController.WalletGeneration.cs
@@ -43,13 +43,13 @@ namespace BTCPayServer.Controllers.Greenfield
             }
 
             var canUseHotWallet = await CanUseHotWallet();
-            if (request.SavePrivateKeys && !canUseHotWallet.HotWallet)
+            if (request.SavePrivateKeys && !canUseHotWallet.CanCreateHotWallet)
             {
                 ModelState.AddModelError(nameof(request.SavePrivateKeys),
                     "This instance forbids non-admins from having a hot wallet for your store.");
             }
 
-            if (request.ImportKeysToRPC && !canUseHotWallet.RPCImport)
+            if (request.ImportKeysToRPC && !canUseHotWallet.CanRPCImport)
             {
                 ModelState.AddModelError(nameof(request.ImportKeysToRPC),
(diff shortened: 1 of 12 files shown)


hectorbiggins8:
I am using the BTCPay website. I created a shop, with a connected wallet, and there was no recovery phrase at all, now I am stuck and cannot withdraw my funds.
I need assistance on the matter.

slater82:
> I am using the BTCPay website. I created a shop, with a connected wallet, and there was no recovery phrase at all, now I am stuck and cannot withdraw my funds.
> I need assistance on the matter.

I'm here to help you further reach out to this s.slater82@yahoo.com

PR #6631, opened 2025-03-17
Refactor: Move clipboard styling from attributes to CSS classes

thgO-O:
This PR refactors the styling of the clipboard button ([data-clipboard] and [data-clipboard-hover]) across the project. It replaces them with .clipboard-button and .clipboard-button-hover for consistency.

- Replaced [data-clipboard] styling with .clipboard-button
- Replaced [data-clipboard-hover] styling with .clipboard-button-hover
- Updated affected views to use the new classes

closes #6625

PR #6633, opened 2025-03-17
UI: Prevent copy to clipboard double-clicks

### LINKED ISSUE #6632: [Bug]: Copied checkout amount never resets

2.0.7

Locally, manually.

### What happened?

When a user copies the amount, the `Copied` text never resets back to the amount.

Generate an invoice.

Copy the total amount.

Copy text never resets back to the amount.

Image

### Relevant log output

```shell
NA
```

### What browser do you use?

Chrome 128.0.6613.138

### Additional information

_No response_

dennisreimann:
As long as the copy confirmation is shown, it should not allow the handler to be called again.

Fixes #6632. This is also compatible with #6631, so the pull requests can be merged independent of each other.


dstrukt (approved):
tACK

Seems to fix the issue!

@pavlenex tagging you for review as well

pavlenex (approved):
tACK

PR #6649, opened 2025-04-02
Greenfield: Expand user data

dennisreimann:
These two additions are for the user management inside the app: 

- Adds invitation info to the application user data
- Expands the store user data to contain the same properties as the application user data
- Refactors the assembling of the data to consistently use the `UserService.FromModel` method

This way we can offer admins and store owners a detailed overview of their users.


Lcote71 (approved):

BTCPayServer/Services/UserService.cs
```diff
@@ -78,7 +79,7 @@ private static bool IsApproved(ApplicationUser user)
 
         private static bool IsDisabled(ApplicationUser user)
         {
-            return user.LockoutEnabled && user.LockoutEnd is not null &&
+            return user is { LockoutEnabled: true, LockoutEnd: not null } &&
                    DateTimeOffset.UtcNow < user.LockoutEnd.Value.UtcDateTime;
```
NicolasDorier, line 83: You can use this instead:
With CSharp matching you can "extract" the parts of the pattern that are matching. `{}` match anything that isn't null.

```csharp
return user is { LockoutEnabled: true, LockoutEnd: {} lockoutEnd } &&
                   DateTimeOffset.UtcNow < lockoutEnd.UtcDateTime;
```

PR #6659, opened 2025-04-08
Restore behavior: Invite non existing users when adding a store user

NicolasDorier:
This is restoring the behavior before https://github.com/btcpayserver/btcpayserver/pull/6658

However, we are checking the permissions of the user.


BTCPayServer/Controllers/UIStoresController.Users.cs
```diff
@@ -44,35 +54,83 @@ public async Task<IActionResult> StoreUsers(string storeId, StoreUsersViewModel
...
-            
+
+        StoreUsersAction action;
+        string? inviteInfo = null;
         var user = await _userManager.FindByEmailAsync(vm.Email);
         if (user is null)
         {
-            ModelState.AddModelError(nameof(vm.Email), StringLocalizer["This user does not exist"]);
+            action = StoreUsersAction.Invited;
+            if (!_policiesSettings.LockSubscription || User.IsInRole(Roles.ServerAdmin))
```
dennisreimann, line 64: Imho we should use `(await _authorizationService.AuthorizeAsync(User, null, new PolicyRequirement(Policies.CanCreateUser))).Succeeded` rather than `User.IsInRole(Roles.ServerAdmin)`;

NicolasDorier: Ok will change. I just wanted to match what we do in `Register`

Issue #6653, opened 2025-04-05
In Email Rules show "Send the email to the buyer" checkbox only if trigger supports it


rockstardev:
Currently, when creating or updating Email Rules, we have the option to "Send the email to the buyer if an email was provided with the invoice." This feature is intended to dynamically deliver extra notification email (to buyerEmail on invoice for example):

![Image](https://github.com/user-attachments/assets/b51f95d7-9660-452c-8435-90308e5f9753)

This functionality is supported by several triggers, particularly those related to invoices; however, it is not applicable to all triggers. Now that we also have triggers associated with Pending Transactions, these do not include buyer or invoice details, making the checkbox irrelevant for them.

What needs to be done is to review all WebhookDeliveryRequests. If the logic does not utilize the `storeEmailRule.CustomerEmail` field, then it should not be displayed in the UI, as it is not applicable. For instance, the `PaymentRequestWebhookDeliveryRequest` uses this field:

https://github.com/btcpayserver/btcpayserver/blob/c373f1e08f8f5968df78c00634e1fe14e21e3fad/BTCPayServer/HostedServices/Webhooks/PaymentRequestWebhookDeliveryRequest.cs#L26

In contrast, the `PayoutWebhookDeliveryRequest` does not:

https://github.com/btcpayserver/btcpayserver/blob/c373f1e08f8f5968df78c00634e1fe14e21e3fad/BTCPayServer/HostedServices/Webhooks/PayoutWebhookDeliveryRequest.cs#L18

Therefore, in the UI for `Payout Request` triggers, the checkbox should be hidden. This change should be implemented for all relevant triggers in an abstract and scalable manner.

rockstardev:
This is pretty straightforward issue, so let's keep it for new contributors, that want to get familiarized with BTCPay codebase.

AdamWroblewski:
Can I take this task?

NicolasDorier:
@rockstardev I think it isn't a straightforward one.

While you can apply https://github.com/Abhijay007/btcpayserver/commit/db9dbc7dffe2c5bec30486a3b82ad85a1ad99bad I think that this is a bad idea. Because every times you add a field to an interface you break every plugins which implement this interface.

I think somebody has to refactor to avoid the situation of breaking plugins every times somebody an optional field to this form.
Something like rather than having properties such as `SupportXY`, have a single method `Support(string xy)`. (unsure if `string` is the best)

NicolasDorier:
This is already implemented

PR #6660, opened 2025-04-09
Fix/send email checkbox

yemmyharry:
#6653

### DIFF
--- a/BTCPayServer/HostedServices/Webhooks/WebhookSender.cs
+++ b/BTCPayServer/HostedServices/Webhooks/WebhookSender.cs
@@ -87,6 +87,11 @@ namespace BTCPayServer.HostedServices.Webhooks
                 return Task.FromResult(req)!;
             }
 
+            public virtual bool SupportsCustomerEmail()
+            {
+                return false;
+            }
+
             protected  static  string InterpolateJsonField(string str, string fieldName, JObject obj)
             {
                 fieldName += ".";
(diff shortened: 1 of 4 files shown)


BTCPayServer/Views/UIStores/StoreEmailRulesManage.cshtml
```diff
@@ -183,6 +184,9 @@
             const triggerSelect = document.querySelector('.email-rule-trigger');
             const subjectInput = document.querySelector('.email-rule-subject');
             const bodyTextarea = document.querySelector('.email-rule-body');
+            const customerEmailContainer = document.getElementById('customerEmailContainer');
+
+            const supportCustomerEmail = @Html.Raw(System.Text.Json.JsonSerializer.Serialize(webhookTypesSupportingCustomerEmail));
```
NicolasDorier, line 189: Need to use `Safe.Json`.

BTCPayServer/HostedServices/Webhooks/WebhookSender.cs
```diff
@@ -341,5 +346,30 @@ public Dictionary<string, string> GetSupportedWebhookTypes()
...
+        public Dictionary<string, bool> GetWebhookTypesSupportingCustomerEmail()
+        {
+            var result = new Dictionary<string, bool>();
+            var webhookTypes = GetSupportedWebhookTypes();
+
+            foreach (var type in webhookTypes.Keys)
+            {
+                bool supportsCustomerEmail = false;
+                
+                if (type.StartsWith("Invoice"))
```
NicolasDorier, line 359: why isn't it using `Request.SupportsCustomerEmail` ?

yemmyharry: > why isn't it using `Request.SupportsCustomerEmail` ?

i dont have access within the webhook delivery request function. i would need to instantiate it first. GetWebhookTypesSupportingCustomerEmail is a helper function that  only looks at type names, not delivery request instances

BTCPayServer/HostedServices/Webhooks/WebhookSender.cs
```diff
@@ -87,6 +87,11 @@ public class WebhookDeliveryRequest(string webhookId, WebhookEvent webhookEvent,
                 return Task.FromResult(req)!;
             }
 
+            public virtual bool SupportsCustomerEmail()
```
NicolasDorier, line 90: I don't see it used anywhere

BTCPayServer/HostedServices/Webhooks/WebhookSender.cs
NicolasDorier: This is fragile code here. You should do like in `GetSupportedWebhookTypes`, but just check that the provider is the one of payment request or invoices rather than doing string comparisons.

NicolasDorier (changes requested):
See my comments

NicolasDorier:
supersede by https://github.com/btcpayserver/btcpayserver/pull/6815

Issue #6665, opened 2025-04-10
[Bug]: QR Code Scanner in Bitcoin Send dialog should remove "BITCOIN:" prefix pefore putting it in the destination field


petzsch:
### What is your BTCPay version?

BTCPay Server v2.1.0+b09b61251 

### How did you deploy BTCPay Server?

docker

### What happened?

I scanned the BTC onchain deposit address of my Breez wallet via QR Code.

The content of the QR code is "BITCOIN:BC1...."

After scanning the BITCOIN:BC1 line got put in the destination field, however when trying to submit this transaction it's marked as invalid:

![Image](https://github.com/user-attachments/assets/8cbdccf2-634a-4835-a344-ffcf82245f0f)

### How did you encounter this bug?

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox on Win11

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

petzsch:
or maybe BIP21 URLs should be accepted as destination?
The problem accord on a transaction with multiple destinations. (the second destination was the one with the URL)

rockstardev:
This issue should be opened in the Breeze wallet, not in our repository. There is a history behind why QR codes are certain way: https://github.com/btcpayserver/btcpayserver/issues/2110

rockstardev:
I now understand this is about scanner within BTCPay - reopening the issue.

petzsch:
Should I close the issue I just opened with Breez then? If it's the scanner doing something unexpected.

thgO-O:
Hey guys,

I've been reviewing this issue and here’s what I found:

There are currently three ways users can input a BIP21 URI in the Send view:

1- Scanning a QR code

2- Using the "Paste BIP21" button, which opens a dedicated input field

3- Pasting directly into the destination address input field

The third case is not currently supported: when a BIP21 URI (e.g., bitcoin:address?amount=...&pj=...) is pasted directly into the destination field, it results in an "Invalid address" error. This is because the full URI is passed directly to BitcoinAddress.Create(...), which only accepts raw addresses.
This can be fixed by checking for StartsWith("bitcoin:") and if the pasted value starts with bitcoin:, we can assign it to the bip21 variable already handled in the scan/paste logic.

While investigating this, I noticed another design behavior that may be worth reviewing:

[...]

PR #6672, opened 2025-04-12
Fix: strip "bitcoin:" prefix when scanning QR 

sapakus:
Removing everything before the ":" when scanning from QR that contains "Bitcoin:" 

Fixes: #6665 

Manually tested to make sure it works.

### DIFF
--- a/BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
+++ b/BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
@@ -1,11 +1,11 @@
 window.addEventListener("load", () => {
     let $input = null;
     initCameraScanningApp("Scan address or payment link", data => {
-        if (data.includes('?') || $input == null) {
+        if (data.includes(':') || $input == null) {
             document.getElementById("BIP21").value = data;
             document.getElementById("SendForm").submit();
         } else {
-            $input.value = data.includes(":") ? data.split(":")[1] : data;
+            $input.value = data;
         }
     }, "scanModal");
     document.getElementById('scanModal').addEventListener('show.bs.modal', e => {
--- a/BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
+++ b/BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
@@ -5,7 +5,7 @@ window.addEventListener("load", () => {
             document.getElementById("BIP21").value = data;
             document.getElementById("SendForm").submit();
         } else {
-            $input.value = data;
+            $input.value = data.includes(":") ? data.split(":")[1] : data;
         }
     }, "scanModal");
     document.getElementById('scanModal').addEventListener('show.bs.modal', e => {


dennisreimann:
Please keep in mind that this component is used in multiple places and there might be occurences where stripping the prefix isn't wanted. We need to double check this.

sapakus:
[wallet-camera-scanner.js](BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js) only seems to be imported in 

https://github.com/btcpayserver/btcpayserver/blob/18fca7162065823bf53a5aba71794d8dabdafc3d/BTCPayServer/Views/UIWallets/WalletSend.cshtml#L43

and 

https://github.com/btcpayserver/btcpayserver/blob/18fca7162065823bf53a5aba71794d8dabdafc3d/BTCPayServer/Views/UIWallets/WalletBumpFee.cshtml#L43

The import in [WalletBumpFee.cshtml](BTCPayServer/Views/UIWallets/WalletBumpFee.cshtml) seems to be unused. 

Everywhere else, https://github.com/btcpayserver/btcpayserver/blob/18fca7162065823bf53a5aba71794d8dabdafc3d/BTCPayServer/Views/Shared/CameraScanner.cshtml#L70 seems to be called directly from the .cshtml and handled independently. 

I apologize in advance if I'm mistaken and thank you for your time! I'm just getting started to Open-source contribution and not 100% familiar with this codebase.

sapakus:
I realize this is a little more nuanced than I assumed after seeing @thgO-O's comment on #6665. marking this as Draft for now.

sapakus:
The issue was with scanner only treating the scanned QR as bip21 uri if there was a "?" in it, but a valid BIP21 doesn't have to have "?" in it if there is no label or amount. we can detect bip21 by checking to see if it includes ":" instead since that's the only character guaranteed.

BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
```diff
@@ -1,7 +1,7 @@
 window.addEventListener("load", () => {
     let $input = null;
     initCameraScanningApp("Scan address or payment link", data => {
-        if (data.includes('?') || $input == null) {
+        if (data.includes(':') || $input == null) {
```
dennisreimann, line 4: Let's change it to this to test for both options:

```suggestion
        if (data.match(/[:?]/) || $input == null) {
```

dennisreimann (changes requested):

BTCPayServer/wwwroot/js/wallet/wallet-camera-scanner.js
sapakus: Sure, but I'm curious why? There can never be "?" without a ":" in a [BIP21 URI](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)

dennisreimann:
I've tinkered a bit with it... this makes it BIP21-compatible and also solved the problem of clearing other potential destinations, when scanning a BIP21 URI:

```js
window.addEventListener("load", () => {
    let outputIndex = null;
    initCameraScanningApp("Scan address or payment link", data => {
        const bip21 = data.match(/bitcoin:(\w+)(?:\?(.*))?/)
        if (bip21 && outputIndex == null) {
            document.getElementById("BIP21").value = data;
            document.getElementById("SendForm").submit();
        } else if (bip21 && bip21.length >= 2) {
            document.querySelector(`[name="Outputs[${outputIndex}].DestinationAddress"]`).value = bip21[1];
            if (bip21.length === 3) {
                const p = new URLSearchParams(bip21[2]);
                if (p.has('amount')) {
                    document.querySelector(`[name="Outputs[${outputIndex}].Amount"]`).value = p.get('amount');
                }
            }
        } else {
            document.querySelector(`[name="Outputs[${outputIndex}].DestinationAddress"]`).value = data;
        }
    }, 'scanModal');
    const $scanModal = document.getElementById('scanModal');
    $scanModal.addEventListener('show.bs.modal', e => {
        const { index } = e.relatedTarget.dataset;
        outputIndex = index ? index : null;
    });
    $scanModal.addEventListener('hide.bs.modal', e => {
        outputIndex = null;
    });
});
```

Unfortunately I wasn't able to make it also work with the BIP21 label parameter. The [TomSelect input component](https://tom-select.js.org/examples/api/) is kinda gnarly and I wasn't able to set additional labels via JS.

sapakus:
yep, my initial reaction was to also pull this to the client side and ran into the same issue of not being able to update the <select> options from JS and figured that's why this was being done on the server side.

btw, I do have the fix working for scanning the bip21 QR and NOT clearing out the existing outputs, but I still have to work on the Paste Bip21 functionality which is also clearing out the existing outputs. 

Let me know if you'd like me to update this branch with the full fix of all related issues or keep it separate?

dennisreimann:
@sapakus Let's apply either your adapted fix or the version I posted in the comment. Afterwards we can progress with the multiple destinations issue.

PR #6673, opened 2025-04-12
Added inputmode="decimal“ to POS Cart app

iBobik:
So the iOS will show numpad for this fields.


dennisreimann (approved):
Nice catch, thanks for the contribution!

Issue #6662, opened 2025-04-10
Add confirmation for deletion of Email Rules and use the opportunity to create new client side pattern


rockstardev:
Currently, when you delete Email Rules, the HTTP POST is executed immediately without any confirmation. There should be a confirmation step, like on other pages. Please refer to the accompanying video that demonstrates how deletion works for Email Rules compared to the deletion in Vendor Pay when users are removed.

https://github.com/user-attachments/assets/f3c04164-df1d-4181-a5e1-46f70e15ad06

Our existing confirmation pattern requires multiple server-side methods to ensure the command is completed successfully. Instead, we should implement a new client-side dialog that intercepts the form submission using JavaScript. If the user confirms the action, the request will then be submitted to the server. This change will also allow us to clean up the codebase.

The first implementation of this new pattern should be the deletion of Email Rules.

PR #6675, opened 2025-04-12
Fix: adding a confirmation for the deletion of Email Rule

wbalbo:
Fix: https://github.com/btcpayserver/btcpayserver/issues/6662

Added a confirmation modal screen for the deletion of an Email Rule. 

It was also added a template for how we could call the confirmation screen using JavaScript without refreshing the page.

Finally, a UX way to close the modal was added, pressing ESC.

## Summary by CodeRabbit

- **New Features**
	- Added success messages when creating, editing, or deleting email rules, providing clearer feedback to users.
	- Introduced a confirmation dialog for removing email rules, requiring users to type "REMOVE" before deletion.

- **Improvements**
	- Enhanced the email rules management interface with updated action buttons and permission-based visibility.
	- Improved reliability and feedback handling for email rule operations in automated tests.

### DIFF
--- a/BTCPayServer/Views/UIStores/StoreEmailRulesList.cshtml
+++ b/BTCPayServer/Views/UIStores/StoreEmailRulesList.cshtml
@@ -53,7 +53,7 @@
                     <td class="actions-col" permission="@Policies.CanModifyStoreSettings">
                         <div class="d-inline-flex align-items-center gap-3">
                             <a asp-action="StoreEmailRulesEdit" asp-route-storeId="@storeId" asp-route-ruleIndex="@rule.index">Edit</a>
-                            <a asp-action="StoreEmailRulesDelete" asp-route-storeId="@storeId" asp-route-ruleIndex="@rule.index" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-description="@StringLocalizer["This action will remove the rule with the trigger {0}.", Html.Encode(rule.value.Trigger)]" data-confirm-input="@StringLocalizer["REMOVE"]" text-translate="true">Remove</a>
+                            <a asp-action="StoreEmailRulesDelete" asp-route-storeId="@storeId" asp-route-ruleIndex="@rule.index" data-bs-toggle="modal" data-bs-target="#ConfirmModal" data-description="@ViewLocalizer["This action will remove the rule with the trigger <b>{0}</b>.", Html.Encode(rule.value.Trigger)]" data-confirm-input="@StringLocalizer["REMOVE"]" text-translate="true">Remove</a>
                         </div>
                     </td>
                 </tr>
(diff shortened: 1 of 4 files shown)


wbalbo:
Not sure where the tests from this page come from, how are these Razor pages are tested in this project?

NicolasDorier:
Please, use the same pattern as we are using everywhere else. The dialog's design doesn't match anything we use elsewhere.
See how the delete is implemented in `StoreUsers.cshtml` for example. (Which is the `Users` menu item of the store)

NicolasDorier (changes requested):
Please, use the same pattern as we are using everywhere else. The dialog's design doesn't match anything we use elsewhere.
See how the delete is implemented in `StoreUsers.cshtml` for example. (Which is the Users menu item of the store)

wbalbo:
I've implemented the changes based on @NicolasDorier's comment. I wasn't able to run the SeleniumTest for some reason, not related to my change. The error that I got was this:

image

The only change that I've made in my OS was to install .NET 10 for other projects, but this shouldn't be the cause 
of this error.  

I had to remove a small part of code from the CanSetupEmailRules() test that removed the email rules, because following StoreUsers solution I didn't found a way to test in Selenium this specific part. 

Let me know if something else is necessary.

BTCPayServer/Views/UIStores/StoreEmailRulesList.cshtml
```diff
@@ -17,35 +18,19 @@
         </ol>
         <h2>@ViewData["Title"]</h2>
     </nav>
-    <a id="CreateEmailRule" permission="@Policies.CanModifyStoreSettings" asp-action="StoreEmailRulesCreate" asp-route-storeId="@storeId" 
-       class="btn btn-primary" role="button">
-        Create Email Rule
-    </a>
+    <form method="post" permission="@Policies.CanModifyStoreSettings">
```
dennisreimann, line 21: Why the wrapping form here?

wbalbo: I thought was necessary but in this case is really not important. I tested without it and worked fine so I removed it, thanks for looking into it.

NicolasDorier:
@coderabbitai review this

wbalbo:
> @wbalbo in the future, don't remove tests when they don't pass. Fix them, this is half of the fun. (Tests may take even more time to write/debug than the feature itself)
> 
> The test itself was really flaky due to the retrieval of `PageSource` which happened before the page could load. Adding a line waiting for a success message to show up before accessing `PageSource` fixed that... but I needed to add those success message to the UI. The reason it passed on CI is that on CI it is way faster.

Thanks for the explanation. I removed because in my mind there were not needed, since in the original source StoreUsers, they weren't used.

But good to know, thanks for fixing.

PR #6678, opened 2025-04-15
Refactor vault

NicolasDorier:
### New feature
* Support HW Taproot (BIP86) import and signing

### UX
* Can translate the UI text in the HW pairing page
* Remove the `Confirm Addresses` page during HW import, move it to the HW pairing process
* After HW import, set the `Label` to the name of the model of the HW.
* Automatically detect if the HW needs `Default Include NonWitness Utxo in PSBTs`.
* Always show the option `Default Include NonWitness Utxo in PSBTs` in the wallet settings
* Do not show Brave browser warning in HW screens if there is no issue connecting to Vault
* Do not ask passphrase to Trezor One if passphrase protection isn't enabled on it

### Refactoring
* Replaced old WebSocket-based/JS approach by Blazor
* Codebase simplified and cleaned up (Less line of code and no JS)
* VaultBridgeUI logic customizable via `IController` and its UI via custom `VaultElement`

## Motivation

When I implemented hardware wallet integration, we didn't have access to Blazor. As a result, messages were forwarded through a WebSocket between the backend and frontend to achieve the goal.

The resulting code was very difficult to maintain and would have prevented us from implementing more advanced features in the future (e.g., better UI for multisig hardware integration).

On top of that, it was impossible to translate the messages used in the hardware integration.

## Result

The number of lines of code decreased considerably, while readability and user experience improved.

For example, previously, when pairing your store with a hardware wallet, there was a separate "Confirm Addresses" page that let users display the address on the device.

Now, this step is an integral part of the pairing process, making it harder for users to skip.

image

It also sets the wallet label to something more descriptive.

image

Last but not least, it allows the use of Taproot (BIP86) if the hardware wallet supports it.

## Follow-up

I believe @rockstardev built an xpub extractor to help with multisig setup. We could probably move it into core instead of keeping it as a plugin.


dennisreimann:
Blazorify BTCPay 🚀

NicolasDorier:
I am kicking the can for this one until I find a way to allow plugins such as Xpub Extractor of @rockstardev to plug in to it.

NicolasDorier:
@coderabbitai review this

PR #6679, opened 2025-04-16
Fix: Managers could not manage payouts in the UI

NicolasDorier:
Approving, Paying, Marking or Cancelling a payout should be possible to anybody with `btcpay.store.canmanagepayouts` permission.

While the backend was enforcing the proper permission, the UI hid these actions from anyone without `btcpay.store.canmodifystoresettings`.


Deezy33-creator (commented):
[adamspolishes.com](https://www.facebook.com/share/1EHhFGB8mb/?mibextid=wwXIfr)

dennisreimann (approved):

PR #6688, opened 2025-04-22
Automatically set Multisig Server if Output descriptor has several keys

NicolasDorier:
If we detect the output descriptor is a multisig, we should probably activate multi sig server automatically.
Also, if we detect that the output descriptor isn't taproot, we should get the previous txs by default.


NicolasDorier:
@coderabbitai stop editing PR descriptions of this repo

NicolasDorier:
@coderabbitai configuration

PR #6699, opened 2025-04-26
Do not reuse outpoint used in ongoing pending transactions

NicolasDorier:
If you created more than two pending transaction in a multi sig wallet, it was possible they both spend the same UTXO.

This mean that one of the transaction would replace or invalidate the other.


NicolasDorier:
@coderabbitai is there a way to disable walkthrough? I just need any nit, or obvious bugs from PR, not a summary of the changes.

PR #6702, opened 2025-04-28
Restoring reading of PaymentHash from LightningInvoice

rockstardev:
For some reason, we are no longer using PaymentHash returned during invoice creation. One of the biggest issues with this approach is that it hinders our ability to debug mainnet Lightning invoices locally.

Therefore, I have revised the logic - if PaymentHash is present - use it.

Otherwise - if PaymentHash is missing - only then do the extra parsing of BOLT11 invoice and generate PaymentHash.

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningListener.cs
+++ b/BTCPayServer/Payments/Lightning/LightningListener.cs
@@ -527,8 +527,7 @@ retry:
                         continue;
                     if (notification.Id == listenedInvoice.PaymentMethodDetails.InvoiceId &&
                         (notification.BOLT11 == listenedInvoice.PaymentMethod.Destination ||
-                         BOLT11PaymentRequest.Parse(notification.BOLT11, _network.NBitcoinNetwork).PaymentHash ==
-                         GetPaymentHash(listenedInvoice)))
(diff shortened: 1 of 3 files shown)


BTCPayServer/Payments/Lightning/LightningListener.cs
```diff
@@ -592,7 +591,9 @@ public async Task<bool> AddPayment(LightningInvoice notification, string invoice
                 return false;
 
             var handler = _handlers[paymentMethodId];
-            var paymentHash = BOLT11PaymentRequest.Parse(notification.BOLT11, _network.NBitcoinNetwork).PaymentHash;
+            var paymentHash = notification.PaymentHash != null ?
+                uint256.Parse(notification.PaymentHash) :
+                BOLT11PaymentRequest.Parse(notification.BOLT11, _network.NBitcoinNetwork).PaymentHash;
```
NicolasDorier, line 596: You should probably extract this and put it in a `GetPaymentHash(LightningInvoice)` method.

Because there is another place in this file where we extract `PaymentHash` from a `LightningInvoice` in this class that you didn't changed.

PR #6704, opened 2025-04-30
Implement store templates

NicolasDorier:
## Motivation

Ambassadors who onboard local merchants often want new stores to be preconfigured based on local conditions—for example, with their preferred exchange, language, currency, timezone, retail mode, etc.

This PR introduces the ability to define a **default template store** in the `Server Settings`. When a new store is created, its default settings are set from this template instead of the usual USD-centric defaults.

## UX

First, the ambassador sets up a **model store** in BTCPay Server with the desired default settings, including checkout experience and rate sources.

Then, the ambassador makes sure the store is selected in the top left dropdown.

image

Then, under `Server Settings`, they click `Create template from selected store`:

image

This action creates a template based on the selected store’s settings. Future stores will use this template as their default configuration.

Once set, the ambassador has the option to either reset it, or replace it:

image

## Migration

Since the `Default Currency` settings isn't useful anymore, I removed it from the `Server Settings`.

Users who had a `Default Currency` in the server settings different from `USD` are migrated to a template with the same currency.

## Notes

Once the template is created, it becomes independent of the original model store. Any future changes to the model store will not affect the template.

A templates doesn't include any access token, webhooks, emails, payout processors, payment method settings, roles and users of the model store.

Related to https://github.com/btcpayserver/btcpayserver/discussions/2685

#2685 is discussing about removing completely the create store page. I think this is a good follow up in a future PR.
I am adding later an option `Automatically create store for new users` in `Server Settings`: If the default name is already set, there isn't anything to ask to the user for creating a new store.

### DIFF
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -17,7 +17,7 @@ namespace BTCPayServer.Client.Models
         public string Website { get; set; }
 
         public string BrandColor { get; set; }
-        public bool ApplyBrandColorToBackend { get; set; }
+        public bool? ApplyBrandColorToBackend { get; set; }
         public string LogoUrl { get; set; }
         public string CssUrl { get; set; }
         public string PaymentSoundUrl { get; set; }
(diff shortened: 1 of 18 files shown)


NicolasDorier:
@coderabbitai review

dstrukt:
Cool concept - think this could be quite powerful if/when we have plans to expand beyond just basic instancing!

Few immediate thoughts without reviewing in depth:

## 1
w.r.t #2685 i think my original comment (https://github.com/btcpayserver/btcpayserver/discussions/2685#discussioncomment-983463) still holds true today, given that we're not a SPA, as much as i like the functionality suggested / eliminating that view

## 2
> Once the template is created, it becomes independent of the original model store. Any future changes to the model store will not affect the template.

> A templates doesn't include any access token, webhooks, emails, payout processors, payment method settings, roles and users of the model store.

This information should likely be included as helper text or text near the template selection UI.

Will try to get my review in ASAP. Might have a few suggestions!

NicolasDorier:
> This information should likely be included as helper text or text near the template selection UI.

@dstrukt can you tell me what to put in the text exactly?

Issue #5622, opened 2023-12-26
Fallback invoice rate source


pavlenex:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/2267

<div type='discussions-op-text'>

<sup>Originally posted by **astupidmoose** February  9, 2021</sup>
So with all the major exchanges running on windows 95 era servers over the last few weeks and crashing with anything larger than a 2% spike in price I've noticed that quite often our own site will go down because we aren't able to generate rates from the providers we have picked. 

We use the custom rate scripting, and as far as I know I can only define one coin per one exchange. I'd like to know if there is a way, or the possibility to add, the ability to use multiple rate providers in a priority type instance. 

For instance, if you use coinbase for your rate, but have binance set as a failover, then when Coinbase inevitably crashes you are still able to generate invoices using binance, until someone at coinbase turns the power back on at which point it would switch back over. 

Would appreciate ideas and thoughts. </div>

pavlenex:
@NicolasDorier pointed out a significant UX challenge: users might not be aware of or understand how to use the fallback feature until they encounter an outage. This could make the feature less effective, as users would have to navigate the UI during a crisis, which is akin to manually changing the rate provider.

Also, this would mean the fallback would need to be introduced in the store creation screen.

astupidmoose:
I don't know about others, but I'm using Custom rate formulas anyways and would be OK with failovers able to be scripted via the custom rates formulas

PR #6705, opened 2025-05-02
[Feature] Can configure fallback rates

NicolasDorier:
## Motivation

When a rate sources is unreliable due to connectivity issues or plain maintenance, the creation of invoice would fails leading to unwanted downtime for merchants.

This add the ability to configure a fallback rate source. The fallback rate source will be used when the primary rate source fails.

## UX

A new checkbox has been added in the Rates configuration page of the store.

image

When enabled, tabs will appear which allow the merchant to select which rate source he wants to configure.

image

Finally, `Currency pair testing` test both, the fallback and the primary:

image

## API Change

Two routes have been added:
`HTTP GET api/v1/stores/{storeId}/rates/configuration/{rateSource}`
`HTTP PUT api/v1/stores/{storeId}/rates/configuration/{rateSource}`

`rateSource` can either be `primary` or `fallback`.
Legacy `api/v1/stores/{storeId}/rates/configuration` routes have the same behavior as `api/v1/stores/{storeId}/rates/configuration/primary`

Related issue https://github.com/btcpayserver/btcpayserver/issues/5622

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.apps.json
@@ -1225,6 +1225,14 @@
         {
             "name": "Apps",
             "description": "App operations"
+        },
+        {
+            "name": "Crowdfund",
+            "description": "Crowdfund operations"
+        },
+        {
+            "name": "Point of Sale",
+            "description": "Point of Sale operations"
         }
     ]
 }
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.pull-payments.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.pull-payments.json
@@ -1198,6 +1198,10 @@
         {
             "name": "Pull payments payout (Public)",
             "description": "Pull payments payout (Public) operations"
-        }
+        },
+        {
+            "name": "Stores (Payouts)",
+            "description": "Store pull payment payout operations"
+        },
     ]
 }
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-lightning-addresses.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-lightning-addresses.json
@@ -229,5 +229,11 @@
                 }
             }
         }
-    }
+    },
+    "tags": [
+        {
+            "name": "Lightning address",
+            "description": "Lightning address configuration"
+        }
+    ]
 }
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-rates.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.stores-rates.json
@@ -75,7 +75,7 @@
                 "tags": [
                     "Stores (Rates)"
                 ],
-                "summary": "Get store rate settings",
+                "summary": "Get store rate settings for the specified rate source",
                 "parameters": [
                     {
                         "$ref": "#/components/parameters/StoreId"
@@ -84,7 +84,7 @@
                         "$ref": "#/components/parameters/RateSource"
                     }
                 ],
-                "description": "View rate settings of the specified store",
+                "description": "View rate settings for the specified store and rate source (`primary` or `fallback`).",
                 "operationId": "Stores_GetStoreRateConfiguration",
                 "responses": {
                     "200": {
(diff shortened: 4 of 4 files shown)


dstrukt:
concept ACK!

My suggestion would be displaying the Fallback options UI when enabled via accordion, as we do for some of our onion layered functionality, ideally keeping it all on the same view/tab, as opposed to adding a new tab. 

cc @dennisreimann

dennisreimann:
> My suggestion would be displaying the Fallback options UI when enabled via accordion, as we do for some of our onion layered functionality

Yes, let's reuse that UX pattern here — same as with the additional POS options (tips, etc.)

NicolasDorier:
@dstrukt 

So I did it, but the UI looks like shit.

By default it is as always:

image

Once enabled

image

With fallback open

image

rockstardev:
imo, the `primary source` shouldn't be placed in the accordion. If you want to use this design pattern, the accordion should expand when the fallback rate is enabled and collapse otherwise. Including "enabled" in the accordion adds unnecessary friction, as users may find themselves in UI states that do not correspond to enabled or disabled status + you force all these extra clicks.

dstrukt:
@rockstardev agreed, the primary is always visible, and you can opt in / enable a secondary that would then be visible via the accordion

@NicolasDorier no need to scrap, we're very close, let me just mock up a quick visual for us - this is a common ui/ux pattern - will have something shortly!

dstrukt:
First off, ignore the container stylization. I've been experimenting with making our settings views more grouped and legible. I can mock this up without if it's too distracting as well.

## Proposal

Essentially, keep the Primary visible, then just add the `Fallback` or `Secondary Price Source` toggle, not sure if people have a preference. If enabled, show the select option. 

Underneath the two price selection, we keep the two existing options, and push them below. If each price source needs a specific spread or scripting method (not assumed here), just push the `Fallback` toggle to the bottom, and duplicate the price spread and scripting inputs, if enabled.

Screen Shot 2025-05-06 at 3 42 05 PM
Screen Shot 2025-05-06 at 3 42 12 PM

## Changes Requested 

- [ ] Rename the labels for each `Primary Price Source` & `Fallback Price Source`
- [ ] Update the helper text below the Select to `Primary rate provider is {source}`
- [ ] Reorder the inputs to match the mocks, Primary, Fallback (+ expanded), Exchange Spread, Advanced Scripting
- [ ] Remove the confirmation modal
- [ ] Update helper text on the enable switch for the Fallback `Enable Fallback Source` & `Fallback rates will be used when the primary rate is not available.`

I hope all of this makes sense. Feel free to ask any additional clarifying questions if needed!

NicolasDorier:
@dstrukt so the price spread isn't part of the source. It's a "global setting". This is because I don't see why a spread would be different between primary and fallback. But `Advanced Scripting` is per source!

Also, there is the issue of testing: How the user test the pair to see how it's resolve in both cases?

dstrukt:
Makes sense w.r.t spread - being a "global" concept, also makes sense w.r.t scripting being specific just haven't used the scripting personally, so didn't know if it could be applied, or rather.. targeted towards a particular source in the textarea inputs. In that case, we'll likely want to make that explicit in each of the wordings as well i.e.

`Enable advanced rate rule scripting for primary source`

`Enable advanced rate rule scripting for fallback source`

w.r.t Testing, we could approach it:

A) the same way we do the scripting, by duplicating the input and placing it at the end of both the Primary/Fallback UI 
B) keep it at the bottom section, and perhaps let the user select `All`, `Primary`, or `Fallback` via a select input, and then run specific tests on that selection
C) if you have a better suggestion - all ears!

Either way, will get this mocked up.

NicolasDorier:
@dstrukt Another solution is that the test is testing both, and the test result properly show both result.

dstrukt:
The latest iterations based on @NicolasDorier's comments.

@dennisreimann you'e already so swamped with work on the App side, but I've been messing with containerizing Settings content, much like we do in the App, and I think it looks pretty nice - could be something we do here on this view, and we could eventually carry over to other parts of the Server, using this code as the template!

Screen Shot 2025-05-06 at 7 33 14 PM

NicolasDorier:
I've updated the screenshots. The fallback is off by default.

I have reused the UI pattern of the `Enable tips` in the PoS for enabling the fallback.

NicolasDorier:
Merged

PR #6709, opened 2025-05-11
Share server between tests

NicolasDorier:


NicolasDorier:
Not worth the rebases and headaches

PR #6722, opened 2025-05-12
Enable XMR plugin test setup

napoly:
should fix / relates to https://github.com/btcpay-monero/btcpayserver-monero-plugin/pull/4

### DIFF
--- a/BTCPayServer.Tests/BTCPayServerTester.cs
+++ b/BTCPayServer.Tests/BTCPayServerTester.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
+using BTCPayServer.Abstractions.Contracts;
 using BTCPayServer.Configuration;
 using BTCPayServer.HostedServices;
 using BTCPayServer.Hosting;
@@ -260,8 +261,8 @@ namespace BTCPayServer.Tests
 
         private async Task WaitIsFullySynched(CancellationToken cancellationToken)
         {
-            var dashBoard = GetService<NBXplorerDashboard>();
-            while (!dashBoard.IsFullySynched())
+            var o = GetService<IEnumerable<ISyncSummaryProvider>>().ToArray();
+            while (!o.All(d => d.AllAvailable()))
             {
                 await Task.Delay(10, cancellationToken).ConfigureAwait(false);
             }
--- a/BTCPayServer.Tests/TestUtils.cs
+++ b/BTCPayServer.Tests/TestUtils.cs
@@ -6,7 +6,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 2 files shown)


BTCPayServer.Tests/ServerTester.cs
```diff
@@ -83,6 +83,11 @@ public void ActivateLangs()
                 File.Copy(file, Path.Combine(langdir, Path.GetFileName(file)));
         }
 
+        public void ActivateXmrPlugin()
```
NicolasDorier, line 86: Please remove this, I don't want any trace of XMR in core.
You can put those in the constructor of your test class in your plugin.

NicolasDorier:
Can your remove XMR specific code from here?
Also can you better explain the change and why this fix the issue? `AppContext.BaseDirectory` should resolve the same as 

```
var btcPayDirectory = ((OutputPathAttribute)typeof(TestUtils).Assembly.GetCustomAttributes(typeof(OutputPathAttribute), true)[0]).BuiltPath;
```

I would have expected.

napoly:
Not an author of `((OutputPathAttribute)typeof(TestUtils).Assembly.GetCustomAttributes(typeof(OutputPathAttribute), true)[0]).BuiltPath` but they do return different values:

AppContext.BaseDirectory
"/home/user/Git/BTCPayServer/btcpayserver-monero-plugin/BTCPayServer.Plugins.Tests/bin/Debug/net8.0/"
((OutputPathAttribute)typeof(TestUtils).Assembly.GetCustomAttributes(typeof(OutputPathAttribute), true)[0]).BuiltPath
"/home/user/Git/BTCPayServer/btcpayserver-monero-plugin/btcpayserver/BTCPayServer.Tests/bin/Debug/net8.0/"

NicolasDorier:
@napoly make sense thanks

BTCPayServer.Tests/ServerTester.cs
NicolasDorier: You don't need `PayTester.Chains.Add("XMR");`

NicolasDorier: @napoly you can also make virtual the factory methods in the `UnitTestBase` so you can override and add additional logic relevant for your plugin.

You don't need it here.

napoly: @NicolasDorier what you are saying is correct.. i can move this function to test constructor and `dotnet test BTCPayServer.Plugins.IntegrationTests --verbosity normal` runs green.. the problem is I had to switch to [incontainer](https://github.com/btcpay-monero/btcpayserver-monero-plugin/pull/4/files#diff-d5a2cea578a0446aad66faf31bf1076ae94c9916b1a3374e342272a6300e8344R44) style (which i do like as it isolates..) .. but that doesn't run green.. I do understand that in tests the containers are linked by name: `BTCPAY_XMR_DAEMON_URI: http://monerod:18081` .. i can even attach to tests container and ping monerod.. but I'm still not sure what is the correct way to propagate that env var from docker compose.. (digging further here)

Issue #5936, opened 2024-04-16
paidPartial email trigger


pavlenex:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/5920

<div type='discussions-op-text'>

<sup>Originally posted by **earlingeling** April  9, 2024</sup>
Sometimes customers pay insufficient funds and get their invoice marked as New (paidPartial).

The reasoning varies - if they use large crypto-exchanges like binance, coinbase and similar, they often take a rather high fee. Also, they might also take their flat btc fee out of the receivers end of the transaction.

To remedy these situations it would be really helpful to be able to automatically trigger an email rule for paidPartial.
Then the store can give in-depth instructions on how this happens and how they can solve it themselv - and largely remove customers enquiries about paidPartial invoices.

Also, an {Invoice.url} link to their invoice would be helpful so we can link them back to their invoice (especially for stores with modal btcpay overlays at checkout)
</div>

dstrukt:
huge ACK

PR #6723, opened 2025-05-13
Adding InvoiceExpiredPaidPartial and InvoicePaidAfterExpiration webhooks

rockstardev:
Resolves: #5936

### DIFF
--- a/BTCPayServer.Client/Models/WebhookEventType.cs
+++ b/BTCPayServer.Client/Models/WebhookEventType.cs
@@ -5,11 +5,13 @@ public static class WebhookEventType
 {
     public const string InvoiceCreated = nameof(InvoiceCreated);
     public const string InvoiceReceivedPayment = nameof(InvoiceReceivedPayment);
+    public const string InvoicePaymentSettled = nameof(InvoicePaymentSettled);
     public const string InvoiceProcessing = nameof(InvoiceProcessing);
     public const string InvoiceExpired = nameof(InvoiceExpired);
     public const string InvoiceSettled = nameof(InvoiceSettled);
     public const string InvoiceInvalid = nameof(InvoiceInvalid);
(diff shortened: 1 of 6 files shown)


NicolasDorier:
@coderabbitai any potential issue in this code?

rockstardev:
I could create a new `WebhookExpiredExceptionEvent`, but I'd prefer not to introduce an entirely new class. The other two concerns are not relevant because the code that raises events does not involve payments. Additionally, it will never raise events for `AfterExpiration = false`. Payments, in particular, are not included in the webhook, as we only need the invoice ID, which is referenced in the emails sent to customers.

Let me know how you want to proceed here @NicolasDorier.

ChidiChuks:
I could help test and review this PR

rockstardev:
@ChidiChuks sure, also feel free to join on our Mattermost chat to agree on details: https://chat.btcpayserver.org/btcpayserver/pl/zro7g5cnjfbtme919gh1fn5qbe

ChidiChuks:
> @ChidiChuks sure, also feel free to join on our Mattermost chat to agree on details: https://chat.btcpayserver.org/btcpayserver/pl/zro7g5cnjfbtme919gh1fn5qbe

I have joined Mattermost

NicolasDorier:
@rockstardev coderabbitai is right. You shouldn't reuse the PaymentReceived type, it isn't related to payment at all, this is confusing.

rockstardev:
The problem is that whole part of code is mixed up with different webhook classes and accessor properties. To simplify it, I've switched to just using base `WebhookInvoiceEvent` with new events, that's the cleanest way forward.

rockstardev:
@ChidiChuks reply on my comment on open issues in Mattermost and let's move conversation there.

ChidiChuks:
> @ChidiChuks reply on my comment on open issues in Mattermost and let's move conversation there.

@rockstardev  Done

Issue #6712, opened 2025-05-12
Phase 1: Global POS tax implementation


pavlenex:
This feature is urgent, since a large adopter depends on it to demonstrate our POS at a big conference. 

For full context please review https://github.com/btcpayserver/btcpayserver/issues/6711

Scope:

- Add a POS global sales tax (similar to Tips) which in the future, when enabled overwrites Store Settings Global tax (not implemented yet). 
- Once enabled a merchant can add a decimal number 8.375% (please note this is the format in which we want them implemented). 
- Show tax line at the checkout and in receipt (receipt should already show all metadata by default).
- Optional: User can input multiple taxes (for example: sale tax, municipality tax) 

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/d45be892-986b-40ca-88c0-2ca0c18dc89e" />

pavlenex:
This has been completed in [#6724](https://github.com/btcpayserver/btcpayserver/pull/6724)
I'll update other phases accordingly. 
We need to add optionality in the POS to display amount with or without tax inlcuded, and we need to fix the UX of Keypad, which is related to https://github.com/btcpayserver/btcpayserver/issues/6621

PR #6724, opened 2025-05-13
Can apply tax rates to PoS items

NicolasDorier:
## Motivation

For compliance reasons, merchants need the ability to apply tax rates to PoS items and show taxes on receipts.
For more details, see https://github.com/btcpayserver/btcpayserver/issues/6712.

## UX

A new settings is added in the Update Point of Sale.

image

This settings will apply a tax rate to items in the point of sale.

We also allow a per-item customization of this tax rate

image

We then display it in the checkout

image

And in receipts

image

image

We also include it in the legacy invoice export

image

All the point of sales item are considered tax excluded.

The tax is calculated on each line item, after applied discount, and rounded (midpoint up).
Then tip is added to this (not affected by the tax). This follow US accounting rule.

The PoS shows the calculation in real time

image

The invoice details has been updated to show the data.

image

## Implementation

The total amount of taxes is saved in the already existing [.taxIncluded invoice metadata](https://docs.btcpayserver.org/Development/InvoiceMetadata/).


rockstardev (commented):
Looks pretty good to me & it is fully functional. I've tested it with Keypad that includes items and here it is how it went:

https://github.com/user-attachments/assets/a899c0b9-86fc-4f41-9319-eb136da135c5

Ideas on TODOs:
- Add Tax in Payments Report (like it was added to Legacy Invoice Export Report)
- Fix tests, for some reason now calculation logic on 4 selenium tests is failing

NicolasDorier:
@rockstardev 

> Add Tax in Payments Report 

This is not clear. What happen if there are two payments for an invoice? should we split the tax? How do we round up?

rockstardev:
@NicolasDorier OK, if there is no way to do this, we can then use Legacy Invoice Export for starters.

ndeet:
In "Total taxes" the amount is shown but shouldn't also the percentage rate be shown somwhere. At least in Europe I think you need to print the % of the tax too. So if there is only one tax field maybe do "Total taxes (8%)" and the amount?

It get's even crazier in EU as most countries have multiple VAT rates, usally a lower one for food and higher one for the rest. And on receipts you end up with mixed tax rates in the grocery store, e.g.
VAT 10% ...... 0.28
VAT 20% ...... 0.30
Total .......... 12.00

This also implies that the tax rate needs to be set on a per item basis :laughing: But ok this is getting out of hand, let's start with a simple solution.

dstrukt:
> image
> 
> This settings will apply a tax rate to items in the point of sale.

Considering you're not "enabling" this feature, would the default be 0%? 
I like the overridden helper text here as well!


> image

Should we show the %, e.g. `$0.62 (12%)`?

> image

Don't feel very strongly, and could always be adjusted later, but I think i'd push the Taxes bit into the lower tables instead of the primary section here.


> image

Similarly, should we show the %, e.g. `$0.62 (12%)`?


> image

Think we should be consistent with the formatting, e.g. `$0.62 (12%)`. instead of the `15% = X` like above

> image
> 

Similarly, push it to the right e.g. `$0.62 (12%)`. instead of like `Tip (15%)`, should be adjusted for both. Mostly to maintain consistency.

Otherwise, this works really well and covers a lot of bases!

I agree with @ndeet w.r.t to displaying the % wherever possible so the user understands, but if we can't or won't for the conference in this first pass, we should eventually do it!

dstrukt (changes requested):
See comment above. Outside of a few minor nits, this is pretty much ready to go imo!

rockstardev:
@dstrukt heads up that I can't see any of the images you shared

rockstardev:
@coderabbitai any potential issue in this code? 😂

rockstardev:
Yeah, pretty much agree with your conclusions... have few nits that I will add in PR review comment, but this is in mergable state, I would proceed with merging it and user testing.

rockstardev (approved):
No showstoppers, leaving more for documenting:

* I would move POSTests.cs and similar files into the FeatureTests folder to reduce the number of files in Tests.
* The tips + taxes mode is cool, but it brings up the question of how the two interact. I think the current default - calculating tips on the tax-excluded value - is fine, but we should make it configurable.

In conclusion - let's merge this to start with user testing and open issues for follow-up work.

dstrukt:
> @dstrukt heads up that I can't see any of the images you shared

Interesting, just quote replied from Nico's OP.

Let me try to see if I can fix.

NicolasDorier:
> Considering you're not "enabling" this feature, would the default be 0%? I like the overridden helper text here as well!

Yes, default is 0.

> Should we show the %, e.g. $0.62 (12%)?

Sadly we can't. We take this value from `taxIncluded`, which is the sum of taxes... which may (or may not) be from different rates.

> Think we should be consistent with the formatting, e.g. $0.62 (12%). instead of the 15% = X like above

Will fix

NicolasDorier:
I addressed @dstrukt point except the % on taxes which would require too much work at this point.
We can probably address this later when we support other type of taxes.

Issue #6707, opened 2025-05-08
Coinmate as exchange rate provider


NicolasDorier:
### Discussed in https://github.com/btcpayserver/btcpayserver/discussions/4825

* https://coinmate.io/api/tickerAll

rockstardev:
@d4rp4t can you comment on the issue so we can assign it to you?

PR #6452, opened 2024-12-02
Add Rate Providers for BTC/NOK

schjonhaug:
# Add Rate Providers for BTC/NOK

This PR adds support for the Norwegian bitcoin only exchanges Bitmynt (https://bitmynt.no) and Bare Bitcoin (https://barebitcoin.no) as a rate provider for BTCPay Server.

## Changes
- Added new `BitmyntRateProvider` and `BareBitcoinRateProvider` classes to fetch BTC/NOK rates
- Registered Bitmynt and Bare Bitcoin as rate providers in `BTCPayServerServices`

## Details
The providers fetch bid/ask prices from Bitmynt's public API endpoint (https://ny.bitmynt.no/data/rates.json) and Bare Bitcoin's public API endpoint (https://api.bb.no/price) and provide BTC/NOK exchange rates. These additions will improve price accuracy for Norwegian merchants using BTCPay Server.

## Testing
- Verified that the rate providers successfully fetch current BTC/NOK rates
- Confirmed integration with existing rate provider infrastructure
- Tested bid/ask price parsing from API responses

Skjermbilde 2024-12-02 kl  15 01 20
Skjermbilde 2024-12-02 kl  15 01 11

PR #6725, opened 2025-05-13
Add Coinmate rate provider

d4rp4t:
Resolves:  #6707

### DIFF
--- a/BTCPayServer.Rating/Providers/CoinmateRateProvider.cs
+++ b/BTCPayServer.Rating/Providers/CoinmateRateProvider.cs
@@ -1,0 +1,53 @@
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


rockstardev (changes requested):
Looks great, can you just add basic test checks to `BTCPayServer.Tests/ThirdPartyTests.cs`. Similar to how it was done for previous rate providers? See as example:

https://github.com/btcpayserver/btcpayserver/pull/6452/files

Seems that validation should mostly be focused on EUR and CZK from comments @prusnak left last week: https://github.com/btcpayserver/btcpayserver/discussions/4825#discussioncomment-13051270

d4rp4t:
> Looks great, can you just add basic test checks to `BTCPayServer.Tests/ThirdPartyTests.cs`. Similar to how it was done for previous rate providers? See as example:

Done! Added tests for both EUR and CZK.

rockstardev (approved):
LGTM

Issue #6621, opened 2025-03-08
Counter-Intuitive Calculator Display in POS Checkout


pavlenex:
## Description
The calculator interface in the POS Checkout screen displays information in a reverse order compared to standard calculators, causing confusion for merchants. The running total is displayed prominently while the calculation sequence is shown in smaller, secondary text.
Steps to Reproduce

Open the POS Checkout screen
Enter multiple values using the calculator (e.g., 5 + 5 + 0.80)
Observe how the display prioritizes the total over the calculation sequence

## Current Behavior

- Primary (large) display shows the running total: $10.80
- Secondary (small, gray) display shows the calculation: $5.80 + $5.00
- When adding new values, the primary display is updated with new totals instead of showing the current input

## Expected Behavior
The interface should match standard calculator behavior:

- Primary display should show the calculation sequence or current input
- Secondary display should show the running total 

## Suggested UI (up for discussion)

<img width="236" alt="Image" src="https://github.com/user-attachments/assets/0501fed0-4374-4407-bfd3-a698afc72188" />

## User Context
Testing conducted with merchants  who are unfamiliar with Bitcoin and digital payments. The standard calculator mental model is causing expectations that conflict with the current interface design.

GittyGitGitGitty:
The `x` also looks like a multiplication symbol.

Changing the `x` to `DEL` or `BACK` or similar would be appreciated.

dstrukt:
It used to be a `C`, but that's not recognized in many parts of the world, then we changed it to `X`, but we agree.

I do like the idea of a using the standard apple delete/back icon or a `<` chevron, to keep it simple.

We've have been kicking around the idea of adding a `delete icon (Apple)`, an `X`, a `C`, or any other locale based symbol, etc.. as it seems to vary quite a bit around the world, and we want to provide the flexibility, but considering there are problems with simply using the phone's location to display the correct symbol, we would have to introduce an admin feature to manually select the correct symbol in the Checkout settings.

Thanks for the feedback!

PR #6739, opened 2025-05-19
Show tax rate in receipt, adjust keypad display

NicolasDorier:
## Fix keypad display

I want input from @dstrukt and @dennisreimann on this, as I am unsure if I am breaking something that was intended.
Related discussion https://github.com/btcpayserver/btcpayserver/issues/6621

Previously, the bold amount shown on the keypad reflected the full invoice total (including discounts, taxes, and tips), not the value being typed. This was confusing, as the user input didn't match what was displayed.

For example, typing `5.00` would immediately show `5.50` due to tax being applied, and subsequent entries like `8.00` wouldn’t visibly appear on screen, only the running total updated.

*Before:*

Before

This PR updates the keypad to display the exact amount the user is typing, without pre-applying tax or other adjustments. This makes input behavior clearer and more predictable.

Notice that the total is now shown in the calculation `= total`.

*After:*

After

## Add taxes on receipts and cart

Trying to nail down some observation by @dstrukt 
This PR show the tax rate in the receipt

image

And in the cart

image

The percentage will not show if items are mixing different rates.

### DIFF
--- a/BTCPayServer.Tests/POSTests.cs
+++ b/BTCPayServer.Tests/POSTests.cs
@@ -522,23 +522,23 @@ donation:
             Assert.Equal("1.234,00", await s.Page.TextContentAsync("#Amount"));
             Assert.Equal("", await s.Page.TextContentAsync("#Calculation"));
             await EnterKeypad(s, "+56");
-            Assert.Equal("1.234,56", await s.Page.TextContentAsync("#Amount"));
+            Assert.Equal("0,56", await s.Page.TextContentAsync("#Amount"));
             Assert.True(await s.Page.IsEnabledAsync("#ModeTablist-discount"));
             Assert.True(await s.Page.IsEnabledAsync("#ModeTablist-tip"));
-            await AssertKeypadCalculation(s, "1.234,00 € + 0,56 €");
+            await AssertKeypadCalculation(s, "1.234,00 € + 0,56 € = 1.234,56 €");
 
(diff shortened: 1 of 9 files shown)


pavlenex:
Beautiful!

dstrukt (changes requested):
I do like the thinking, and initially thought I might like it more, but the more I sit with it, I don't think it makes sense to push the Total anywhere other than the large bolded number, so in it's current Keypad form, nACK.

Will follow up with more thinking, I think there's a middle ground between keeping the calculated bold total, but only show the value being typed on the history list. 

cc @dennisreimann 

Minor UI nits:

- [ ] make the line between the cart elements and the `Subtotal` thin as well, the bold should be reserved for the Total, imo

image

- [ ] id shift would be pushing the discount input element to the far right to be consistent with the other elements, i think it used to be that way and there might have been a regression or mistake.

image

NicolasDorier:
Here is how it looks now in the cart

image

Receipt

image

And print

image

Is it what you need @dstrukt ?

dstrukt:
Yes this looks great - i know they're minor nits, but while we're looking at this view!

rockstardev:
Hey guys, I've waited for this change over the weekend. Also now waited whole day to deploy it for testing. I can't have it be delayed any longer.

Let's please merge it in this state and then we can open a separate issue to do further changes to the feature.

Thank you!

Issue #6729, opened 2025-05-14
[Bug]: Yadio exchange rate lookup failing


kaloudis:
### What is your BTCPay version?

v2.1.0 + v2.1.1

### How did you deploy BTCPay Server?

Docker + manual deployment

### What happened?

Yadio exchange rates are failing, likely due to a configuration change on their end. We have fallen back to Coingecko which has most of the currencies, but still it degrades the options available to our users on our mobile and web apps.

The funny thing is everything works fine in ZEUS app if you switch away from our BTCPay endpoint and select Yadio directly, so it seems like it must be some sort of server config or CORS thing on their end. Because of this I do not suspect it to be a BTCPay server bug per se, but figured it would be good to document everything here and to be able to track it.

### How did you encounter this bug?

Got a notification that our rates API endpoint was down and manually hit the endpoint

### Relevant log output

```shell

```

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

sutt0n:
fwiw, this is because `XPT` was returning null. related: https://github.com/blinkbitcoin/price/issues/72

sutt0n:
i received this from Yadio via e-mail:

> Thank you for contacting us about this. The XPT ticker is fixed now. Please try again and let's know if you continue to have problems with it.
> 
> Regards,
> Yadio

curling their api seems that the XPT ticker is no longer null.

kaloudis:
Thanks. We're back to fully operational.

Will leave the issue open to the maintainers in case they want to add error handling here. Sounds like other related enhancements are on the way though.

rockstardev:
OK, so essentially someone needs to change from `decimal` to `decimal?` here: https://github.com/btcpayserver/btcpayserver/blob/9afa45088e9979f4ba9f20301f4992114ee66a0e/BTCPayServer.Rating/Providers/YadioRateProvider.cs#L35

And then if value is null, skip the item, otherwise add it as `decimal` into the list.

Good first issue for new contributors @pavlenex.

PR #6743, opened 2025-05-20
(Fix): Fixed Yadio exchange rate lookup failing 

Abhijay007:
Fixes: #6729 

image


Abhijay007:
cc: @rockstardev review required

rockstardev (approved):
Great first PR @Abhijay007. I've just added test on currency that @kaloudis reported was causing problems.

Issue #6744, opened 2025-05-21
[Bug]:Fiat text input disappears if you remove the BTC amount you've previously entered entirely


pavlenex:
### What is your BTCPay version?

v2.1.1+6b5cab65a

### How did you deploy BTCPay Server?

Docker

### What happened?

When clearing out the amount field, the fiat equivalent field gets cleared out.

This can be tricky for users to navigate, especially if by mistake they enter fiat amount in the input field for btc, then realize they've made a mistake removed that amount, now they can't add fiat amount.

### How did you encounter this bug?

1. Enter BTC amount for example 0.001
2. Remove the amount with backspace until there's no amount
3. The fiat field gets cleared out

### Relevant log output

```shell

```

### What browser do you use?

Safari

### Additional information

https://github.com/user-attachments/assets/f5f3a0d9-d2b3-4a7f-81fd-796a70246a9c

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

Abhijay007:
@pavlenex can I take up this?

pavlenex:
Sure, go for it!

PR #6746, opened 2025-05-21
(Fix): Fiat text input disappears if you remove the BTC amount 

Abhijay007:
Fixes: #6744 

image


Abhijay007:
cc : @pavlenex , @rockstardev review required

Abhijay007:
> LGTM. @Abhijay007 a suggestion that you install https://screentogif.com - so you can record videos of how PR works, instead of single shot.

sure, thanks for the suggestion I will share vids from next time instead of screenshots

PR #6749, opened 2025-05-23
Add tip and subtotal to report

NicolasDorier:
Ping @rockstardev is it what you need?

I want to re-iterate something for people using the Legacy report:
** They should not just add up the `Invoice*` columns, because if an invoice has two payment, they will sum up twice the tip or tax. **

I know everybody like the legacy invoice export, but I am pretty sure everybody is using it incorrectly and are just summing columns that would be counted several time if an invoice has several payments.

### DIFF
--- a/BTCPayServer/Services/Invoices/PosAppData.cs
+++ b/BTCPayServer/Services/Invoices/PosAppData.cs
@@ -19,6 +19,17 @@ public class PosAppData
         }
         return null;
     }
+    public static PosAppData TryParse(JObject posData)
+    {
+        try
+        {
+            return posData.ToObject<PosAppData>();
+        }
+        catch
+        {
+        }
+        return null;
+    }
 
     [JsonProperty(PropertyName = "cart")]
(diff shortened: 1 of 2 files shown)


rockstardev:
@NicolasDorier yes, this is a great start. I'd merge it so people have access to the data.

But what we really need is a version of this report where YOU CAN sum the columns and clearly separate subtotal, tips, and taxes. Maybe call it **Settlement Report**, and 100% optimize it to serve the following use case:

1. Provide summary data for a selected range - broken down into subtotal, tips, and taxes
2. Display a detailed view of all invoices and payments - item by item, showing which invoice each payment is tied to, and break down each by subtotal, tips, and taxes

Think of the situation: you've got 10 merchants using the instance, and 1 accountant handling payouts. This report would be what the accountant uses to quickly:

* Pay each merchant exactly what they're owed (in USD, BTC, etc.)
* Allocate tips to staff operating POS terminals
* Transmit the correct sales tax amount to the state

That’s the goal.

Issue #6406, opened 2024-11-16
[Bug]: email activation for new users can't be turned of for private registration via API


petzsch:
### What is your BTCPay version?

BTCPay Server v2.0.3+558f45a48

### How did you deploy BTCPay Server?

docker

### What happened?

A [user on Telegram](https://t.me/btcpayserver/81254) reported that he's missing an API call of activating newly created user accounts (with API created).
After some research by @pavlenex and myself we found that the Option "Email confirmation required" can only be disabled if "Enable public user registration" is on in server settings. However there should be no self registration enabled as users are created via the API.

It seems that this use case was overlooked here.

I think it would be better to allow toggle of email confirmation requirement independently if public registration is enabled.

### How did you encounter this bug?

I didn't, the above mentioned telegram user did.

### Relevant log output

```shell
none
```

### What browser do you use?

not specified

### Additional information

none

### Are you sure this is a bug report?

- [X] I confirm this is a bug report

pavlenex:
> I think it would be better to allow toggle of email confirmation requirement independently if public registration is enabled.

I don't think this should be a desired outcome, and it wouldn't fix the underlaying problem. There should be some sort of explanation in the API or the error that this needs to be enabled, since from what I understand ,the user leveraging API had no context on what's happening.

lightningcheckout:
I wasn't aware of this issue created. 
After your explanation on TG that the activation of the user only impacts GUI, this is not an issue for me. 
It took some time to figure this out, so perhaps some additional info in the docs could be of help. 
Or an option to enable the account (via an additional field or automatically) when password is provided.

dennisreimann:
@lightningcheckout Yes, these are expected to behave this way. I've added #6425 to provide more precise error messages for the approval endpoint to distinguish why approval failed.

Mecanik:
This issue is currently happening to myself with the latest version. I wish to:

- Have public registrations disabled
- Create users via API

However, what happens is exactly what the OP described. I create a user via the API and email confirmation gets sent out. The only current workaround I can see is to break/remove the SMTP configuration details... which I really don't want to do, because I wish to receive email notifications myself.

What is the solution for this? Email confirmations cannot be turned off unless public registration is ON.

Thanks

Mecanik:
Help?

rockstardev:
@Mecanik help is on the way 😂

Mecanik:
[...]

PR #6752, opened 2025-05-24
Adding flag on Greenfield API Create User method to skip email invitation

rockstardev:
Resolves: #6406

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
@@ -249,7 +249,7 @@ namespace BTCPayServer.Controllers.Greenfield
         {
             var user = await _userManager.GetUserAsync(User);
             if (user is null) return this.UserNotFound();
-            
+
             UploadImageResultModel? upload = null;
             if (file is null)
                 ModelState.AddModelError(nameof(file), "Invalid file");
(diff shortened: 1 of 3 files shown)


BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
```diff
@@ -358,6 +358,9 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
                 Created = DateTimeOffset.UtcNow,
                 Approved = isAdmin // auto-approve first admin and users created by an admin
             };
+            if (request.SkipEmailInvite == true)
```
dennisreimann, line 361: This doesn't seem logically correct to me. An email confirmation might still be desirable even though one would like skip sending the invite when creating users via the API.

Mecanik:
Is this going to be merged or not? If the latter, please advise how can someone edit the code manually and remove this check/feature. Thanks.

BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
rockstardev: @Mecanik I'm confused as well. Will talk with @dennisreimann when I have a moment to understand what's the problem. Will also address coderabbitai test comment.

Will also address

Mecanik: Can you merge then please?

rockstardev: @dennisreimann - could you please implement this the way you would prefer? All we need is to have a flag that allows the caller of the API method to ensure no email is sent.

> ... make this feature work different then on the backend, where we also have the option to skip sending the invitation email. There it does only that, skipping the email — but not automatically setting the email to be confirmed.

I am fine with matching the backend logic behavior: skipping sending of invitation email but not automatically setting the email to be confirmed. And if you implement it that way - let's merge it.

PR #6753, opened 2025-05-25
Revert the 2.1.1 behavior of keypad total

NicolasDorier:
Asked by @pavlenex


NicolasDorier:
talked with @rockstardev and @pavlenex closing this.

@pavlenex was concerned how changing this behavior would affect EU users. But since EU users don't use PoS tax feature yet nor tips, this actually isn't really changing anything for them at all.

NicolasDorier:
Supersede by https://github.com/btcpayserver/btcpayserver/pull/6768

Issue #6751, opened 2025-05-24
Sort stores by name when displayed in dropdowns and archive list


rockstardev:
Currently, when generating Greenfield API keys, the stores listed in the dropdown are not sorted. You can navigate to [this link](https://your_btcpay/account/addapikey) to see the issue:

![Image](https://github.com/user-attachments/assets/ee67faf1-db6b-494b-96f6-e3594d2d2fbf)

A similar problem occurs when viewing all archived stores at [this link](https://your_btcpay/stores?archived=true):

![Image](https://github.com/user-attachments/assets/316bb1e8-860a-48f5-9389-708c4d89375e)

To improve usability, please sort the stores alphabetically before displaying them. This will make it easier to find a store by name.

gayuvick:
@rockstardev , I would like to work on this issue.

monicamuyama:
Hello @rockstardev , I am Muyama Monica a bitcoin dada dev and I would like to take this on. Can  I go ahead. Thank you

rockstardev:
Thanks @monicamuyama & congrats on first PR merged in BTCPay Server 🙌

PR #6754, opened 2025-05-25
Fix/6751 sort dropdown

monicamuyama:
This PR ensures that the list of stores displayed in the dropdown when generating a Greenfield API key is sorted alphabetically by store name. It also caters for the archives
Changes Made
- Updated the `SetViewModelValues<T>` method in UIManageController.APIKeys.cs to order stores alphabetically using orderBy before assigning them to the view model.
- Updated the `ListStores` action in the UIUserStoresController.cs controller to include `.OrderBy` using `StringComparer.InvariantCultureIgnoreCase` to sort store names alphabetically.

Fixes: #6751

### DIFF
--- a/BTCPayServer/Controllers/UIManageController.APIKeys.cs
+++ b/BTCPayServer/Controllers/UIManageController.APIKeys.cs
@@ -468,7 +468,7 @@ namespace BTCPayServer.Controllers
         {
             // FIXED: Added OrderBy to sort stores alphabetically by name
             var stores = await _StoreRepository.GetStoresByUserId(_userManager.GetUserId(User));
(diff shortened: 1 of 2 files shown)


TChukwuleta:
Hey @monicamuyama, thanks for your contribution!

It looks like this PR includes a large number of unrelated commits—from previous merges. 
Can you ensure that the PR only reflects your changes?

Nit: You can rebase your branch onto the latest. This will clean up the commit history so your PR only shows the changes you made.

monicamuyama:
Replaced by #6758  to clean up commit history. Only includes relevant changes.

PR #6778, opened 2025-06-10
Refactor greenfield api to follow UI flow when creating stores

### LINKED ISSUE #6773: [Bug]: User removed but email still in use

v2.1.1

btcpay-setup.sh

### What happened?

I am creating users via the API, and I am removing them as I test via the dashboard. Even after 10 minutes, a user removed cannot be re-created with the same email.

```
[▼
  "info" => [▶]
  "response_headers" => [▶]
  "response_json" => [▼
    [▼
      "path" => "Email"
      "message" => "Username '....' is already taken."
    ]
  ]
]
```

How can I resolve this?

Create user via API.

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

TChukwuleta:
Resolve #6773 

The changes now allows for the creation of new users via the greenfield API when adding them to a store if they do not already exist, with appropriate authorization and email invitation logic. This follows the same flow as in the UI dashboard.

Also, I noticed something, If I delete a user from a store, the user is deleted from the store users but exist in the server instance as an active user, hence deleted store users can create their own stores. 

I don't think this would favor ambassadors running membership. The PR then includes a deactivation of user when the user is deleted from the store. The admin can then activate the user at will or remove the user entirely from the system.

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreUsersController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreUsersController.cs
@@ -31,12 +31,14 @@ namespace BTCPayServer.Controllers.Greenfield
         private readonly EmailSenderFactory _emailSenderFactory;
         private readonly IAuthorizationService _authorizationService;
         private readonly PoliciesSettings _policiesSettings;
+        private readonly EventAggregator _eventAggregator;
         private readonly UserManager<ApplicationUser> _userManager;
         private readonly CallbackGenerator _callbackGenerator;
         private readonly UriResolver _uriResolver;
 
         public GreenfieldStoreUsersController(
             StoreRepository storeRepository,
+            EventAggregator eventAggregator,
             PoliciesSettings policiesSettings,
             EmailSenderFactory emailSenderFactory,
             UserManager<ApplicationUser> userManager,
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@dennisreimann had a PR like this, and I prefer we do not do this as the permissions for creating a user on the server are different from the permissions to add a server to a store. I think it was even merged and I needed to revert it.

You also removed some stuff in tests which I don't understand why.

TChukwuleta:
Alright. 

What do you think about this:

"If I delete a user from a store, the user is deleted from the store users but exist in the server instance as an active user, hence deleted store users can create their own stores."

You can test this by adding a user to a store, and removing the user from the store. You can then try and login with the user, the removed user can then be a store owner

NicolasDorier:
@TChukwuleta this is expected, as store users are normal server users which got granted right to access the store. Removing the store user only mean that the permission to the store has been removed.

If you want to make sure a server user do not create new store, then a new UX need to be though: For example, adding the ability to affect permissions during the creation of the new user.

It would also need permission for store creation, which I think do not exist yet.

NicolasDorier:
Closing this PR for the reason enunciated: Creation/Deletion of users isn't same permission as granting or not the access to a store.

I couldn't find where, but I remember @dennisreimann already did this, and it was merged at one point. But during a review, I saw it and reverted.

PR #6781, opened 2025-06-10
Fix: Free items in the PoS were generating top-up invoices rather than settled invoices (Fix #6780)

### LINKED ISSUE #6780: [Bug]: 0 amount invoices no longer work as free items in POS, but generate any amount

 v2.1.2+c6f960cef

Docker 

### What happened?

Since updating to 2.1.2 it seems that we broke how we handle invoice with the amount 0 in the POS.

To replicate go to https://pay.pavlenex.com/apps/2SfptYNoGkDKoc5Sc5C2s6jXpQWD/pos

1. Invoice with amount = 0 is free item and it is automatically settled 
2. Invoice with the amount null is any amount, or topup invoice, which accepts any amount

To replicate you can even generate the invoice manually with amount=0 and it will settle. The only issue is that it seems our recent changes around POS likely made a breaking change there, leading all freebie items in POS to become donation ware (any amount).

We even apply `free tag`

Image

1. Create a POS item with amount = 0
2. Version 2.1.1
3. Checkout and see any amount instead of instantly settled.

### Relevant log output

```shell
  {
    "id": "free-registration",
    "title": "Free Registration (You already have BTCPay Merch)",
    "disabled": false,
    "description": "",
    "categories": null,
    "image": "",
    "priceType": "Fixed",
    "price": "0",
    "buyButtonText": "Free registration",
    "inventory": 2,
    "taxRate": null
  },
```

### What browser do you use?

_No response_

### Additional information

_No response_

NicolasDorier:
Regression from 2.1.2

Fix https://github.com/btcpayserver/btcpayserver/issues/6780

### DIFF
--- a/BTCPayServer.Tests/POSTests.cs
+++ b/BTCPayServer.Tests/POSTests.cs
@@ -172,6 +172,8 @@ orange:
 donation:
   price: 1.02
   custom: true
+goodies:
+  price: 0
 ";
             vmpos.Currency = "EUR";
             vmpos.Template = AppService.SerializeTemplate(MigrationStartupTask.ParsePOSYML(vmpos.Template));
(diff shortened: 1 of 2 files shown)


NicolasDorier:
merged on https://github.com/btcpayserver/btcpayserver/commit/f52cb053feb36ce06e57c9926f2b102a18dba17b

PR #6784, opened 2025-06-11
Fix: Allow Pos Cart view to check out free items exclusively

Kukks:


NicolasDorier:
I already pushed that, but I kept the choiceItem check, because I am not sure why we should remove it.

NicolasDorier:
Closing as it is already on master (minus the check on choiceItem)

PR #6793, opened 2025-06-14
feat(pos): add confirmation dialog for multi-store user selection in POS (#6763)

### LINKED ISSUE #6763: Ask for confirmation to display QR code for Server Admins in POS
In most cases, when selecting a user from the dropdown to log into the Point of Sale, you're typically choosing the dedicated user for that specific store.

However, selecting the Server Admin or any user with multistore access introduces a potential security risk. It's also easy to accidentally select such a user. To mitigate this, the system should prompt a confirmation when a user with access to multiple stores is selected. For example:

"Are you sure you want to generate a QR code for a user who has access to multiple stores?"

This confirmation ensures users are aware they are not logging in under a store-scoped account.

Image
(screenshot illustrating the problem ^)

Ayrabb:
- Adds a confirmation dialog when a multi-store user is selected in the POS.
- Ensures users are aware of and confirm multi-store selection before proceeding with an order.
- Passes the selected userId to the backend for accurate order attribution.
- Updates the POS controller to accept and process the userId field.
- (Closes #6763)

### DIFF
--- a/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
+++ b/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
@@ -106,7 +106,6 @@ namespace BTCPayServer.Plugins.PointOfSale.Controllers
             var store = await _appService.GetStore(app);
             var storeBlob = store.GetStoreBlob();
             var storeBranding = await StoreBrandingViewModel.CreateAsync(Request, _uriResolver, storeBlob);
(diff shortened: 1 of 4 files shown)


rockstardev:
@Ayrabb you need to rebase this on top of latest `master` and then also resolve CodeRabbit suggestions before we proceed with review.

dstrukt:
Thank you for your contribution!

Small note, but for more visual PRs, when possible, it would be nice to post a screenshot or two of the changes on the OP - thanks.

PR #6796, opened 2025-06-17
feat: add Reserved Addresses view with filtering, pagination and labels

### LINKED ISSUE #6285: Implement UI to Display List of Reserved Addresses and Labels in Bitcoin Wallet
We currently support the functionality to reserve and label wallet addresses using the `Receive` button in the Bitcoin wallet. However, viewing a list of all reserved addresses is currently only possible through a Postgres query.

Implementing a user interface for this would enhance usability and accessibility.

Nicolas:
- Wants searchable list
- Rockstar: add paging as well (Nicolas: not a simple problem... make it performant so it doesn't blow up when there is 1 million addresses... we cannot show for example invoice count without blowing up performance)

Kukks:
- We had issue that had to redo wallet page to show and manage and edit labelling attached to items. (Maybe: https://github.com/btcpayserver/btcpayserver/issues/4250)

thgO-O:
This PR implements a new Reserved Addresses view. It allows users to:

- View all addresses generated from the Receive tab
- See associated labels and reservation timestamps
- Add labels to an existing reserved address
- Filter addresses by label, address or date
- Use client-side pagination and customizable page sizes
- Access Reserved Addresses directly by clicking a label in the Wallet Labels screen, which redirects to the Reserved Addresses view filtered by that label

It also includes:

- A new reservedAt metadata field stored when the address is generated via receive tab.
- A new test that validates filtering, pagination, label interactions and redirect from Wallet Labels screen.

https://github.com/user-attachments/assets/9a656b56-a14a-49a0-87a2-fa784b702418

Resolves #6285


BTCPayServer.Tests/SeleniumTests.cs
```diff
@@ -3319,5 +3319,88 @@ private void SetTransactionOutput(SeleniumTester s, int index, BitcoinAddress de
                 checkboxElement.Click();
             }
         }
+
+        [Fact(Timeout = TestTimeout)]
+        public async Task CanUseReservedAddressesView()
+        {
+            using var s = CreateSeleniumTester();
```
NicolasDorier: Can you use the PlaywrightTester instead for new tests? We plan to remove selenium.

This is not that hard to convert your test to it, AI can do a decent job at doing it for you.

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -763,6 +763,26 @@ public async Task<IActionResult> WalletReceive([ModelBinder(typeof(WalletIdModel
...
 
+        [HttpGet("{walletId}/addresses")]
+        public async Task<IActionResult> ReservedAddresses(
+            [ModelBinder(typeof(WalletIdModelBinder))] WalletId walletId)
+        {
+            var paymentMethod = GetDerivationSchemeSettings(walletId);
+            if (paymentMethod == null)
+                return NotFound();
+
+            var labeledAddresses = await WalletRepository.GetReservedAddressesWithDetails(walletId);
```
NicolasDorier, line 774: Wouldn't that blow up if the store has say 1 million addresses?

NicolasDorier (commented):
Great work.

I am concerned about the fact you are loading all addresses in memory, this will end up blowing up for even moderately big stores.

What you are trying to do is to filter only the `received` addresses. The cool part of it is that since it concerns only the addresses generated manually, those shouldn't be too big, and loading them in memory with client side filtering shouldn't be too bad.

The problem is that your query in the repository is fetching all address, then filtering on the server. You should use a custom SQL query (or extending `GetWalletObjects`) to fetch the info instead of relying on

```
var objects = await GetWalletObjects(new GetWalletObjectsQuery(walletId, WalletObjectData.Types.Address)
            {
                IncludeNeighbours = true
            });
```

Other issue is UX wise. Given where the Addresses menu is placed, a user would expect to find ALL addresses included those not manually received! (for example, a customer complains he sent money to an invoice address, user will probably want to check here and not find the address)

For this reason, I believe you should either:
* Keep the current address menu, but expand it to all addresses, and in this case the search must be server side to not blow up.
* Move this address menu inside the "Receive" to make it clears that it only cover those addresses. (and fixing the SQL query as I said)

thgO-O:
Thanks for the review!

I'll work on the improvements you mentioned.

kayyrod21:
## 🐛 Bug Report: Labels not persisting in Reserved Addresses view

Hi @thgO-O — great work on this feature! 👏

I tested this locally using the latest commit (`a4ac0e8`) from the `pr-6796` branch. I can confirm the UI correctly lists reserved addresses and allows labeling. However, I found an issue where the **labels disappear** unless you switch tabs or reload the page.

### ❌ Actual behavior:
- When I add a label to a reserved address, it appears immediately.
- But the label disappears after interacting further on the page (unless I switch tabs or refresh).
- It seems like the UI state is not properly updated reactively after labeling.

### ✅ Expected behavior:
- Labels should persist without needing to reload or switch views.

### 📽️ Screen recording:
I recorded the issue in this video:
[Screen Recording: Label not persisting](https://youtu.be/QAH4CnTj2oQ)

### 🧪 My testing steps:
- BTCPayServer running in Regtest via JetBrains Rider
- Wallet created in Regtest
- I generated blocks and transactions using:
  ```bash
  bitcoin-cli -regtest generatetoaddress 101 <regtest_address>

rockstardev (commented):
I spoke with @NicolasDorier at BTCPrague and we agreed that the best path forward is the second option he suggested:

> Move this address menu inside the "Receive" to make it clears that it only cover those addresses. (and fixing the SQL query as I said)

To save you some time @thgO-O, I went ahead and implemented this and pushed a commit to your branch.

Also, with local testing, I confirmed what @kayyrod21 pointed out. Fixing the labels is important because I’m thinking we add a Reserved Addresses link to the Labels page here:

image

So each Label will have `Reserved Addresses` link and once clicked a filtered page would open:

image

BTCPayServer/Views/UIWallets/ReservedAddresses.cshtml
```diff
@@ -0,0 +1,176 @@
...
+            <tr>
+                <th text-translate="true">Address</th>
+                <th text-translate="true">Label</th>
+                <th text-translate="true">Reserved At</th>
+            </tr>
+            </thead>
+            <tbody>
+            @foreach (var address in Model.Addresses)
+            {
+                <tr v-show="filteredItems.some(a => a.address === '@address.Address')">
```
NicolasDorier: @thgO-O is it possible to use VueJs to generate this table instead of Razor.  (replacing the `@foreach (var address in Model.Addresses)`) and the `@address`. 

As @coderabbitai said (https://github.com/btcpayserver/btcpayserver/pull/6796/files#r2151298905), mixing both is a good way to introduce some XSS vulns. This PR doesn't, but this is easy foot shooting for someone who may modify this page and forget the context.

NicolasDorier (changes requested):

BTCPayServer/Views/UIWallets/ReservedAddresses.cshtml
thgO-O: > @thgO-O is it possible to use VueJs to generate this table instead of Razor. (replacing the `@foreach (var address in Model.Addresses)`) and the `@address`.
> 
> As @coderabbitai said (https://github.com/btcpayserver/btcpayserver/pull/6796/files#r2151298905), mixing both is a good way to introduce some XSS vulns. This PR doesn't, but this is easy foot shooting for someone who may modify this page and forget the context.

Thanks for pointing this out. The tricky part is vc:label-manager is a Razor component and can't be used directly from Vue, so switching fully would mean reimplementing it. Open to discuss the best way forward.

NicolasDorier: @thgO-O for the `<vc:` we can keep razor, no problem, there is none of that risk.

For `<tr v-show="filteredItems.some(a => a.address === '@address.Address')">` you however have another issue, this one is O(n2) complexity, so I think it might crash the browser suddenly if there is say 10 000 addresses. (maybe less)

For the ` '@address.Address'` you can use `@Safe.Json(address.Address)` this should escape it properly.

For the `filteredItems`, I think you should maintain a `filteredItemsSet` of addresses every times you modify the filteredItems. Checking if an address is contained there should be quite fast.

thgO-O: > @thgO-O for the `<vc:` we can keep razor, no problem, there is none of that risk.
> 
> For `<tr v-show="filteredItems.some(a => a.address === '@address.Address')">` you however have another issue, this one is O(n2) complexity, so I think it might crash the browser suddenly if there is say 10 000 addresses. (maybe less)
> 
> For the ` '@address.Address'` you can use `@Safe.Json(address.Address)` this should escape it properly.
> 
> For the `filteredItems`, I think you should maintain a `filteredItemsSet` of addresses every times you modify the filteredItems. Checking if an address is contained there should be quite fast.

Thanks for the review! Changes pushed with filteredItemsSet. Thought pagination slice would avoid issues, but this ensures better perf.

Issue #6783, opened 2025-06-10
[Bug]: Line Breaks <br /> in Dropdown


ghost:
### What is your BTCPay version?

BTCPay Server v2.1.2+c6f960cef

### How did you deploy BTCPay Server?

[Demo](https://mainnet.demo.btcpayserver.org/)

### What happened?

2x `<br />
`

![Image](https://github.com/user-attachments/assets/ea262626-b1fd-433b-a92b-f6db9d7867c0)

### How did you encounter this bug?

Settings -> Checkout Appearance -> Select a preset

### Additional information

The text itself is in two files:
 https://github.com/btcpayserver/btcpayserver/blob/79d6bc9fb54e33ffc1357dc1b0c30d20606786a2/BTCPayServer/Services/Translations.Default.cs#L567-L568
https://github.com/btcpayserver/btcpayserver/blob/79d6bc9fb54e33ffc1357dc1b0c30d20606786a2/BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml#L121
https://github.com/btcpayserver/btcpayserver/blob/79d6bc9fb54e33ffc1357dc1b0c30d20606786a2/BTCPayServer/Views/UIStores/CheckoutAppearance.cshtml#L126

But the problem is probably more likely where you guys render the strings. Sanitization might be the culprit.

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

bhola-dev58:
Hi, 
I'd like to work on this issue as my first contribution. 
I’ve reviewed the code and can submit a fix using `@Html.Raw()` to properly render the line breaks. 
Let me know if I can take this up. Thanks!

dstrukt:
@bhola-dev58 i've assigned you the Issue, feel free to fork the main repo, create a PR with a link to this issue - we'll review the change(s), and if they look good, we'll get it merged!

bhola-dev58:
> [@bhola-dev58](https://github.com/bhola-dev58) i've assigned you the Issue, feel free to fork the main repo, create a PR with a link to this issue - we'll review the change(s), and if they look good, we'll get it merged!


<hr>
Thanks @dstrukt for assigning the issue! 🙌

I’ve forked the repo and started working on the fix. I’ll open a PR shortly with the changes.

Appreciate your guidance!

bhola-dev58:
Hi @dstrukt sir,
I noticed that GitHub’s "Intelligent summaries of build failure messages" feature is not enabled.
[...]

PR #6803, opened 2025-06-21
fix(ui): Line Breaks <br /> in Dropdown

bhola-dev58:
### What this PR does

- Resolves issue #6783: Line breaks were not rendering correctly in 
  Settings -> Checkout Appearance -> Select a preset
- Replaced `<br />` in translation strings with separate `<div>`s for better localization, accessibility, and CSP compliance.
- Updated both the view and the translations file accordingly.

### Why this approach?

- Avoids using @Html.Raw() inside translation strings, which can introduce CSP issues.
- Improves i18n flexibility for different languages.
- Supports better styling and screen reader accessibility.

Let me know if any improvements are needed. Thanks for the guidance on my first contribution!


bhola-dev58:
Hi  👋 @dstrukt sir,
 I've submitted a fix, but one of the test cases is failing (`CheckJsContent`). I'm not sure why this is happening as I'm still new to the project. Any guidance would be appreciated. 🙏

NicolasDorier:
@bhola-dev58 instead of changing the translations can you instead change the `text-translate="true"` to `html-translate="true"`? This should fix the issue.

bhola-dev58:
Hi! @dstrukt sir,

**Resolved:**
 - This PR Resolves issue #6783: Line breaks were not rendering correctly in
 `Settings → Checkout Appearance → Select a preset.`
by integrating html-translate="true" to fix the line break rendering.
 - Successfully fixed by integrating `html-translate="true"` instead of using raw `<br />` in translation strings.
 - Replaced `<br />` with separate `<div>` elements to improve localization, accessibility, and CSP compliance.
 - Updated both the view file and the translation file accordingly.





### Why this approach?
 - Avoids the use of inside translation strings, reducing potential CSP issues.
 - Enhances i18n support for different languages.
 - Enables cleaner styling and improves screen reader accessibility.

All tests pass and visuals are attached above.
Kindly review this PR when convenient.
Thanks again for the guidance on my first contribution — happy to make any further changes if needed! ✅

NicolasDorier:
@bhola-dev58 your removed all your changes and added a bunch of changes such as space changes in files that aren't related. Don't do this. Please review your PR before asking review.

NicolasDorier (changes requested):
Changing unrelated files in the projects, and isn't actually fixing line breaks in Dropdown.

bhola-dev58:
@NicolasDorier 
Thank you for the feedback — I really appreciate it.
I'm still new to open-source contributions, and I realize now that I included unrelated changes by mistake.
I will start fresh from the correct base branch and submit a clean PR with only the required fix for the dropdown line break issue.
Thanks again for your patience and guidance!

NicolasDorier:
Do not open several PR for same thing #6820

PR #6818, opened 2025-07-01
feat: exposed invoice accounting endpoint to facilitate refund on btcpayserver app project

IzyPro:
feat: exposed invoice accounting endpoint to facilitate refund on btcpayserver app project

Linked to: https://github.com/btcpayserver/app/pull/228


NicolasDorier:
Concept ACK.

However three small points:

* You need to add documentation it in the swagger files
* I don't like the endpoint name. It isn't really accounting. What about `~/api/v1/stores/{storeId}/invoices/{invoiceId}/refund/{paymentMethodId}` this would better match with the UI and the intention.
* Add more assertions to your test

@pavlenex @dstrukt  I am wondering if we should go ahead with this (to keep the UX the same as Core), or if we should first rethink the refund flow even on Core.

I feel like the refund process is asking too many steps to the merchant.
Wouldn't it be better to configure in the store settings the `Refund method` and not ask questions for refunds (aside from the ability to override with "Custom")

Or allowing to select the Default Refund Method, with ability to override it in the dialog?

pavlenex:
@NicolasDorier I would go with it to ensure feature parity, and then we can log it as a separate issue and re-think it. We could have refund policy, but bear in mind that you'd still need to offer all these steps in case merchant hasn't set up such policy. If you want we could log a separate issue and discuss there which specific things you want improved, but for not it's far more important to have any sort of refund capability in the app, as we saw user-feedback where merchants had to refund from their own wallets, which caused problems with reporting and settlement between owner and cashier who refunded.

dstrukt:
Tend to agree with @pavlenex. Feature parity at this moment seems to make the most sense so users at least have something, but the refund flow definitely could use some attention, and I do like your line of thinking @NicolasDorier!

We should create a basic ticket @pavlenex to scope out a V2 / optimization.

IzyPro:
> Concept ACK.
> 
> However three small points:
> 
> * You need to add documentation it in the swagger files
> * I don't like the endpoint name. It isn't really accounting. What about `~/api/v1/stores/{storeId}/invoices/{invoiceId}/refund/{paymentMethodId}` this would better match with the UI and the intention.
> * Add more assertions to your test
> 
> @pavlenex @dstrukt I am wondering if we should go ahead with this (to keep the UX the same as Core), or if we should first rethink the refund flow even on Core.
> 
> I feel like the refund process is asking too many steps to the merchant. Wouldn't it be better to configure in the store settings the `Refund method` and not ask questions for refunds (aside from the ability to override with "Custom")
> 
> Or allowing to select the Default Refund Method, with ability to override it in the dialog?

@NicolasDorier this is updated now

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class InvoiceRefundTriggerData
+    {
+        public decimal CryptoAmountNow { get; set; } = 0m;
+        public string CurrentRateText { get; set; }
+        public decimal CryptoAmountThen { get; set; } = 0m;
+        public decimal FiatAmount { get; set; } = 0m;
+        public decimal? OverpaidAmount { get; set; } = null;
+        public string InvoiceCurrency { get; set; } = string.Empty;
+        public string CryptoCode { get; set; } = string.Empty;
```
NicolasDorier:
```suggestion
        public string PaymentCurrency { get; set; } = string.Empty;
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
+namespace BTCPayServer.Client.Models
+{
+    public class InvoiceRefundTriggerData
+    {
+        public decimal CryptoAmountNow { get; set; } = 0m;
```
NicolasDorier:
```suggestion
        public decimal PaymentAmountNow { get; set; } = 0m;
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+        public string CurrentRateText { get; set; }
+        public decimal CryptoAmountThen { get; set; } = 0m;
```
NicolasDorier:
```suggestion
        public decimal PaymentAmountThen { get; set; } = 0m;
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+        public decimal FiatAmount { get; set; } = 0m;
```
NicolasDorier:
```suggestion
        public decimal InvoiceAmount { get; set; } = 0m;
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+        public decimal? OverpaidAmount { get; set; } = null;
```
NicolasDorier:
```suggestion
        public decimal? OverpaidPaymentAmount { get; set; } = null;
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+        public string InvoiceCurrency { get; set; } = string.Empty;
+        public string CryptoCode { get; set; } = string.Empty;
+        public int CryptoDivisibility { get; set; }
```
NicolasDorier:
```suggestion
        public int PaymentCurrencyDivisibility { get; set; }
```

BTCPayServer.Client/Models/InvoiceRefundTriggerData.cs
```diff
@@ -0,0 +1,15 @@
...
+        public int InvoiceDivisibility { get; set; }
```
NicolasDorier:
```suggestion
        public int InvoiceCurrencyDivisibility { get; set; }
```

NicolasDorier (changes requested):
Thanks the PR is good to me. Maybe I'll do a follow up PR to try to remove the duplicated code with the UI later.

I know you tried to use the same terminology as the Refund UI, however, it is an old terminology that is very confusing.

Can you the properties and the doc as suggested?

IzyPro:
Updated @NicolasDorier

PR #6824, opened 2025-07-02
Feat: Alert user if Lightning Node lacks inbound capacity. 

MegalithicBTC:
This pull request relies on changes to `BTCPayserver.Lightning`, shown in this PR: https://github.com/btcpayserver/BTCPayServer.Lightning/pull/172

`BTCPayServer.Lightning` now creates a `LiquidityReport` for a backing Core Lightning node, which allows us to show useful alerts to the user on the Lightning page:

no-inbound-capacity
pending-capacity
good-inbound-capacity

### DIFF
--- a/BTCPayServer/Components/Liquidity/Liquidity.cs
+++ b/BTCPayServer/Components/Liquidity/Liquidity.cs
@@ -1,0 +1,101 @@
+#nullable enable
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using BTCPayServer.Lightning;
+using BTCPayServer.Lightning.CLightning;
+using BTCPayServer.Lightning.JsonConverters;
+using Newtonsoft.Json;
+
+namespace BTCPayServer.Components.Liquidity
+{
+    /// <summary>
+    /// Overall inbound-liquidity health.
+    /// </summary>
+    public enum LiquidityStatus
+    {
+        Good,
+        Pending,
+        Bad
+    }
+
+    /// <summary>
+    /// DTO returned when liquidity could be analysed.
+    /// </summary>
+    public class LiquidityReport
+    {
+        public LiquidityStatus LiquidityStatus { get; set; }
+
+        [JsonConverter(typeof(LightMoneyJsonConverter))]
+        public LightMoney ActiveInboundSatoshis { get; set; } = LightMoney.Zero;
+
+        [JsonConverter(typeof(LightMoneyJsonConverter))]
+        public LightMoney PendingInboundSatoshis { get; set; } = LightMoney.Zero;
+    }
+
+    /// <summary>
+    /// Helper for checking Core-Lightning inbound liquidity.
+    /// </summary>
+    public static class Liquidity
+    {
+        private static readonly LightMoney DefaultThreshold = LightMoney.Satoshis(250_000);
+
+        /// <summary>
+        /// Returns <c>null</c> when the supplied node is not Core-Lightning or
+        /// if an error occurs; otherwise returns a populated <see cref="LiquidityReport"/>.
+        /// </summary>
+        public static async Task<LiquidityReport?> CheckAsync(
+            ILightningClient client,
+            LightMoney? threshold = null,
+            CancellationToken token = default)
+        {
+            // This check is specific to CLN because it's the only implementation
+            // where ListChannels returns pending channels.
+            if (client is not CLightningClient)
+            {
+                return null;
+            }
+
+            var min = threshold ?? DefaultThreshold;
+
+            try
+            {
+                var channels = await client.ListChannels(token);
+
+                if (channels is null)
+                {
+                    return null;
+                }
+
+                // inbound capacity = total – local
+                LightMoney activeInbound = channels.Where(c => c.IsActive)
+                                                    .Aggregate(LightMoney.Zero,
+                                                               (s, ch) => s + (ch.Capacity - ch.LocalBalance));
+
+                LightMoney pendingInbound = channels.Where(c => !c.IsActive)
+                                                     .Aggregate(LightMoney.Zero,
+                                                                (s, ch) => s + (ch.Capacity - ch.LocalBalance));
+
+                var status = LiquidityStatus.Bad;
+                if (activeInbound >= min)
+                    status = LiquidityStatus.Good;
+                else if (pendingInbound >= min)
+                    status = LiquidityStatus.Pending;
+
+                var report = new LiquidityReport
+                {
+                    LiquidityStatus = status,
+                    ActiveInboundSatoshis = activeInbound,
+                    PendingInboundSatoshis = pendingInbound
+                };
+                return report;
+            }
+            catch (Exception)
+            {
+                // Return null instead of re-throwing to allow the UI to handle it gracefully.
+                return null;
+            }
+        }
+    }
+}
--- a/BTCPayServer/Components/Liquidity/LiquidityViewComponent.cs
+++ b/BTCPayServer/Components/Liquidity/LiquidityViewComponent.cs
@@ -1,0 +1,128 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using BTCPayServer.Lightning;
+using BTCPayServer.Services;
+using BTCPayServer.Services.Stores;
+using BTCPayServer.Data;
+using BTCPayServer.Payments;
+using BTCPayServer.Payments.Lightning;
+using BTCPayServer.Configuration;
+using BTCPayServer.Services.Invoices;
+using Microsoft.Extensions.Options;
+using NBitcoin;
+
+
+namespace BTCPayServer.Components.Liquidity
+{
+    /// <summary>
+    /// ViewComponent that displays Lightning Network inbound liquidity information.
+    /// Currently only supports Core Lightning (CLN) nodes.
+    /// </summary>
+    public class LiquidityViewComponent : ViewComponent
+    {
+        private readonly StoreRepository _storeRepository;
+        private readonly LightningClientFactoryService _lightningClientFactory;
+        private readonly BTCPayNetworkProvider _networkProvider;
+        private readonly PaymentMethodHandlerDictionary _handlers;
+        private readonly IOptions<LightningNetworkOptions> _lightningNetworkOptions;
+
+        public LiquidityViewComponent(
+            StoreRepository storeRepository,
+            LightningClientFactoryService lightningClientFactory,
+            BTCPayNetworkProvider networkProvider,
+            PaymentMethodHandlerDictionary handlers,
+            IOptions<LightningNetworkOptions> lightningNetworkOptions)
+        {
+            _storeRepository = storeRepository;
+            _lightningClientFactory = lightningClientFactory;
+            _networkProvider = networkProvider;
+            _handlers = handlers;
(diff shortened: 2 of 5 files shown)


dstrukt:
Awesome work! Love to see a more user friendly push for Lightning liquidity.

Have a few quick formatting suggestions.

Was debating if `Receive Capacity` or `Receive Capacity` made more sense than `Lightning Inbound` or `Lightning Inbound Capacity` for headers? Thoughts? cc @pavlenex

Similarly, `None` vs `Bad`? Don't feel to strongly, just another thought.

I utilized our invoice status labels, looks more like a tag or label than a button.

I formatted a bit of the container and text, using a mixture of 4, 8, 16, and 24px paddings for the fonts, left-aligned the content, and simplified the encasing container, and adjusted/removed a bit of text.

I also kept the `Get Inbound Liquidity` at each state as well, in case the user wants to go back to the plugin, maybe it makes more sense to change the CTA text.

Feel free to use if you want, or choose what's easy to do, also happy to iterate more as well - I can also link you to our Figma if that makes it easier as well.

Screen Shot 2025-07-03 at 5 03 49 PM

BTCPayServer/Components/LSPS1/Default.cshtml
```diff
@@ -0,0 +1,70 @@
...
+@if (Model.HasLiquidityReport && Model.LiquidityReport != null)
+{
+    var status = Model.LiquidityReport.Liquidity_Status.ToString();
+    var activeInboundSats = Model.LiquidityReport.Active_Inbound_Satoshis.MilliSatoshi / 1000;
+    var pendingInboundSats = Model.LiquidityReport.Pending_Inbound_Satoshis.MilliSatoshi / 1000;
+
+    Func<long, string> formatAmount = sats =>
+    {
+        var btc = (decimal)sats / 100_000_000m;
+        return $"{sats:N0} sats ({btc.ToString("0.00000000")} BTC)";
```
NicolasDorier: this isn't how we format currencies. Please look the `CurrencyNameTable`

BTCPayServer/Components/LSPS1/LSPS1ViewComponent.cs
```diff
@@ -0,0 +1,171 @@
...
+            _storeRepository = storeRepository;
+            _lightningClientFactory = lightningClientFactory;
+            _networkProvider = networkProvider;
+            _handlers = handlers;
+            _lightningNetworkOptions = lightningNetworkOptions;
+        }
+
+        public async Task<IViewComponentResult> InvokeAsync(string storeId)
+        {
+            _logger.LogInformation("[LSPS1View] Component invoked for storeId: {StoreId}", storeId);
```
NicolasDorier: Do not add logs everywhere, no logs please.

BTCPayServer/Components/LSPS1/LSPS1ViewComponent.cs
```diff
@@ -0,0 +1,171 @@
...
+                if (client == null)
+                {
+                    _logger.LogWarning("[LSPS1View] Could not get Lightning client.");
+                    viewModel.Message = "Could not connect to Lightning node";
+                    return View(viewModel);
+                }
+                _logger.LogInformation("[LSPS1View] Successfully retrieved Lightning client of type: {ClientType}", client.GetType().Name);
+
+                _logger.LogInformation("[LSPS1View] Calling Liquidity.CheckAsync...");
+                var liquidityReport = await Liquidity.CheckAsync(client, _logger);
```
NicolasDorier: Liquidity shouldn't be in LightningClient.

BTCPayServer/Views/UIStores/Lightning.cshtml
```diff
@@ -44,7 +44,7 @@
             Public Node Info
         </a>
     </div>
-
+        <vc:lsps1 store-id="@Model.StoreId" />
```
NicolasDorier: `lsps1` isn't really descriptive. Please find a better name that explain what this component is doing.

BTCPayServer/Components/LSPS1/Default.cshtml
```diff
@@ -0,0 +1,70 @@
+@model BTCPayServer.Components.LSPS1.LSPS1ViewModel
+@using BTCPayServer.Lightning
+
+@if (Model.HasLiquidityReport && Model.LiquidityReport != null)
+{
+    var status = Model.LiquidityReport.Liquidity_Status.ToString();
+    var activeInboundSats = Model.LiquidityReport.Active_Inbound_Satoshis.MilliSatoshi / 1000;
```
NicolasDorier: Do not convert using your own math. Use `Active_Inbound_Satoshis.Satoshi` is you want the satoshi amount.

BTCPayServer/Components/LSPS1/Default.cshtml
MegalithicBTC: thanks

BTCPayServer/Components/LSPS1/LSPS1ViewComponent.cs
MegalithicBTC: thanks

BTCPayServer/Components/LSPS1/LSPS1ViewComponent.cs
MegalithicBTC: "Liquidity shouldn't be in LightningClient." -- I just want to be sure of that. Because this is basic data we are getting from the the backing Lightning Node -- shouldn't it be abstracted along with other utility methods specific to Lightning? I'm definitely happy to take your directive on this. I just want to be sure that YOU are sure about that.

BTCPayServer/Views/UIStores/Lightning.cshtml
MegalithicBTC: thanks

BTCPayServer/Components/LSPS1/Default.cshtml
MegalithicBTC: thanks

BTCPayServer/Components/Liquidity/Liquidity.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                var report = new LiquidityReport
+                {
+                    LiquidityStatus = status,
+                    ActiveInboundSatoshis = activeInbound,
+                    PendingInboundSatoshis = pendingInbound
+                };
+                return report;
+            }
+            catch (Exception)
```
NicolasDorier, line 94: Do not catch it here, the UI needs to decide what to do when this crash.

BTCPayServer/Components/Liquidity/LiquidityViewComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+                    viewModel.Message = "Liquidity information is available for your Core Lightning node";
+                }
+                else
+                {
+                    viewModel.Message = "Liquidity information is only available for Core Lightning nodes";
+                }
+            }
+            catch (Exception)
+            {
+                viewModel.Message = "Error retrieving liquidity information";
```
NicolasDorier, line 85: At least show the exception's message in this message, so people can figure out what happened.

BTCPayServer/Components/Liquidity/LiquidityViewComponent.cs
```diff
@@ -0,0 +1,128 @@
...
+                if (string.IsNullOrEmpty(paymentMethodDetails.ConnectionString))
+                {
+                    return _lightningNetworkOptions.Value.InternalLightningByCryptoCode.TryGetValue(network.CryptoCode, out var internalClient)
+                        ? internalClient
+                        : null;
+                }
+
+                return _lightningClientFactory.Create(paymentMethodDetails.ConnectionString, network);
+            }
+            catch (Exception)
```
NicolasDorier, line 122: no catch all exception handlers. Let the exception bubble up. Else it is impossible to troubleshot.

NicolasDorier (changes requested):

NicolasDorier:
I talked with @MegalithicBTC .

He doesn't really want to move forward with this PR, but this is something we should include to improve the life of merchants using lightning.
It would be cool if another person can take ownership of this PR.

MegalithicBTC:
@NicolasDorier Thanks. Yes, the devops challenges to work on core have now overwhelmed my capabilities, unfortunately.  

To re-cap: I think the ideal goal is to incorporate our LSPS1 plugin into core: https://github.com/MegalithicBTC/BTCPayserver-LSPS1

The goal would be: If a user sets up a Lightning node, they are alerted that they "need a lightning channel to receive inbound payments" -- and are then directed to the LSPS1 interface you can see in the video linked above. 

I'm very happy to provide any expertise on Lightning and LSPS1 that might be useful. Please feel free to add to this pull request, or contact me via https://megalithic.me/contact

dstrukt:
Appreciate the work @MegalithicBTC! 

Regardless of who picks this up.

MegalithicBTC:
One note I could add here. 

I would suggest, due to the complexities involved, to not attempt to support LND with a first version of this.

Supporting LND will likely require additional code, and there is also an added puzzle as to "where this code should go".

The issue with LND: For this feature to be useful for users, we need to know about "pending" channels from the backing node.

For CLN, there is one nice API that gives us both "pending" and "confirmed" channels, and BTCPayserver is already using that. 

For LND,  there are actually two different REST endpoints that would would need to hit in LND -- 

for confirmed channels -- "ListChannels" -- we already use this here:
https://github.com/btcpayserver/BTCPayServer.Lightning/blob/8c71e11ba67a51d02d1dc957231630efc3e0e2b0/src/BTCPayServer.Lightning.LND/LndClient.cs#L363

for "pending" channels -- which are not yet confirmed..... Here is the LND API..
https://lightning.engineering/api-docs/api/lnd/lightning/pending-channels/

...  I don't believe that `BTCPayserver.Lightning` currently gets these pending channels... so possibly we need to add this if we want to support LND with this feature

The added challenge is that in comments to this pull request @NicolasDorier was clear that no changes should be made to the `BTCPayserver.Lightning` library, so I modified the pull request to pull everything into core.

It might end up being confusing, but I think therefore the suggested strategy would be to add these LND-specific methods directly to core, instead of abstracting them into this external library? 

This strategy would seem to add even more complexity on top of a very complex system, as then LND's `ListChannels` endpoint would be accessed via `BTCPayserver.Lightning`, whereas the `ListPendingChannels` endpoint would be accessed via core -- which is sure to confuse future developers on the project -- so I might suggest to just give up on LND compatibility for now.

NicolasDorier:
> For LND, there are actually two different REST endpoints that would would need to hit in LND --
> 
> for confirmed channels -- "ListChannels" -- we already use this here: https://github.com/btcpayserver/BTCPayServer.Lightning/blob/8c71e11ba67a51d02d1dc957231630efc3e0e2b0/src/BTCPayServer.Lightning.LND/LndClient.cs#L363
> 
> for "pending" channels -- which are not yet confirmed..... Here is the LND API.. https://lightning.engineering/api-docs/api/lnd/lightning/pending-channels/
> 
> ... I don't believe that `BTCPayserver.Lightning` currently gets these pending channels... so possibly we need to add this if we want to support LND with this feature

The correct fix would be to just consolidate the call to those two API so the `ListChannel` in `BTCPayServer.Lightning` is also returning the pending channels like CLN does. By doing that there would be no LND specific code in btcpayserver.

Am I missing something?

MegalithicBTC:
Yes, I think this is right. At your request I abandoned my attempt to make any modifications to `BTCPayserver.Lightning`, see here -- https://github.com/btcpayserver/BTCPayServer.Lightning/pull/172#pullrequestreview-2990055242 -- but maybe someone else will be able to pick this up.

Issue #6792, opened 2025-06-13
Allow updating Payment Request as Settled if there are settled invoices


rockstardev:
If invoices linked to a Payment Request are settled with exceptions (such as being paid late or paid partially), they will not update the Payment Request status to "Settled".

This is even if you mark the invoices as settled. As a result, the Payment Request will remain in "Pending" status.

It should be possible to manually update the Payment Request to "Settled" if there are settled invoices, regardless of the amounts paid on those invoices.

Abhijay007:
Hi @rockstardev you can assign me this one

Abhijay007:
Hi @rockstardev, for this issue, do we need to create similar dropdowns like we have in invoices, so that we can mark them settled or something manually? For payment requests? 

Invoices : 

![Image](https://github.com/user-attachments/assets/403320e5-2828-4ac4-ba5d-e646879088bd)

Payment requests : 

![Image](https://github.com/user-attachments/assets/e1f034b4-311c-41d4-bc7e-df34b07c0191)

rockstardev:
@Abhijay007 yes, if we detect that `Pending` Payment Request can be Settled (has invoices that are settled) - we should provide option to `Mark as settled`.

Abhijay007:
Hi @rockstardev,
I'm trying to retrieve the scenarios for "settled late" and "settled partially" — essentially payments that are settled with exceptions (like being paid late or only partially).

I tested a few cases with partial amounts and after expiry, but I'm not seeing the expected tags. The status remains stuck on "processing", and I'm only getting the option to either mark it as "settled" or "invalid".

![Image](https://github.com/user-attachments/assets/45c5132a-3dcc-4d2f-96e4-61da619e95a8)

![Image](https://github.com/user-attachments/assets/e8bbfd55-7ab0-4fc5-a7fb-9c227b508217)

Could you please guide me on how to properly reproduce these scenarios to test the functionality?

rockstardev:
We talked about this today... but to reply here as well - you need to mark Invoice as Settled.

And then once there is at least 1 Settled invoice, Payment Request can be offered to mark as Settled as well.

PR #6825, opened 2025-07-02
(Feat) : Allowed updating payment requests as settled

Abhijay007:
Fixes: #6792 

https://github.com/user-attachments/assets/00481129-3047-4c81-ab6b-6c792fcd0017

ref Pull : https://github.com/btcpayserver/btcpayserver/pull/6807

### DIFF
--- a/BTCPayServer/Controllers/UIPaymentRequestController.cs
+++ b/BTCPayServer/Controllers/UIPaymentRequestController.cs
@@ -465,47 +465,30 @@ namespace BTCPayServer.Controllers
             return NotFound();
         }
 
-        [HttpPost("{payReqId}/changestate/{newState}")]
+        [HttpPost("{payReqId}/complete")]
         [Authorize(AuthenticationSchemes = AuthenticationSchemes.Cookie, Policy = Policies.CanModifyPaymentRequests)]
-        public async Task<IActionResult> ChangePaymentRequestState(string payReqId, string newState)
+        public async Task<IActionResult> TogglePaymentRequestCompleted(string payReqId)
         {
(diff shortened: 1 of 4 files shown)


rockstardev (approved):
The option to update a Payment Request is now available only if there is a settled invoice with some amount collected. 

This will facilitate the completion of Payment Requests in cases of partial payments that occasionally occur on instances with numerous paymentsion requests.

rockstardev:
Here is the video of the feature in action.

https://github.com/user-attachments/assets/542b414a-b107-4c3f-96b5-bb9ef8fe1281

NicolasDorier:
@rockstardev then rather than updating the payment request state, please just mark the invoice that is partially paid as Settled.

NicolasDorier:
Talking with @rockstardev, I was pointing out that marking an invoice partial paid as `Settled` should account the full amount of the invoice toward the Payment request contribution. (not the sum of the payments)
If that was not the case, it was a bug.

@rockstardev said that if we were doing this, it means that if an invoice of 10$ and only 2$ was paid, then we settled that it would be accounted for 10$, which is not what the user want.

I am not fan of the solution of this PR to just mark the payment request itself as Settled. Because in the view of the PR, the total settled will show less than the payment request value, which is counter intuitive. But better than having pending payment requests forever...

rockstardev:
> if an invoice of 10$ and only 2$ was paid, then we settled that it would be accounted for 10$, which is not what the user want.
> better than having pending payment requests forever...

Since the situation of having partially paid invoices that do not cover the full amount of a payment request is considered an edge case, I have moved the option to mark the invoice as settled as admin link on payment request page.

I am merging the pull request now as we agreed, and I have opened a follow-up issue for testing: https://github.com/btcpayserver/btcpayserver/issues/6827.

PR #6835, opened 2025-07-08
Game changing improvements for the Invoice Export report

NicolasDorier:
Talked with @pavlenex .

A common request we have from our users is that there are no good way of exporting the metadata from invoices in the reports. So I needed to approach the problem in several steps:

* The `Legacy invoice export` report is now called `Invoices`.
* Rewriting of fields in the Invoice Report
* Export all metadata in the invoice export
* Remove invoice specific data from lines if there is more than one payment
* Adding `lightningAddress` and `lightningUsername` to metadata for lightning address payments, so it appears in the report
* Added a button in the invoice list to easily browse the report
* The default report when clicking on Reporting was `Payments`, now it is `Invoices`
* Do not show unaccounted payments (ie. RBF'd payments)
* Do not show `New` and `Expired` invoices which doesn't have any exception status nor any payment.

## Rewriting of fields in the Invoice Report

The old structure was:
| Column Name              | Type        |
|--------------------------|-------------|
|  ~~ReceivedDate~~ ->  PaymentReceivedDate | datetime    |
| ~~StoreId~~ (Removed) | text        |
| ~~OrderId~~ (Removed)                  | text        |
| InvoiceId                | invoice_id  |
| InvoiceCreatedDate       | datetime    |
| ~~InvoiceExpirationDate~~  (Removed)  | datetime    |
| ~~InvoiceMonitoringDate~~ (Removed)   | datetime    |
| PaymentId                | text        |
| ~~Destination~~ -> PaymentAddress              | text        |
| ~~PaymentType~~ -> PaymentMethodId              | text        |
| ~~CryptoCode~~ -> PaymentCurrency               | text        |
| ~~Paid~~ -> PaymentAmount                     | text        |
| ~~NetworkFee~~ -> PaymentMethodFee               | text        |
| ~~ConversionRate~~ -> Rate           | number      |
| ~~PaidCurrency~~ -> PaymentInvoiceAmount   | text        |
| InvoiceCurrency          | text        |
| InvoiceDue               | number      |
| InvoicePrice             | number      |
| ~~InvoiceTaxIncluded~~ (Removed)       | number      |
| ~~InvoiceTip~~     (Removed)           | number      |
| ~~InvoiceSubtotal~~    (Removed)       | number      |
| ~~InvoiceItemCode~~    (Removed)       | text        |
| ~~InvoiceItemDesc~~    (Removed)       | text        |
| InvoiceFullStatus  (Use the new terminology, eg. Settled, Processing, Invalid...)     | text        |
| InvoiceStatus    (New terminology)        | text        |
| InvoiceExceptionStatus  | text        |
| ~~BuyerEmail~~ (Removed)               | text        |
| ~~Accounted~~ (Removed)                | boolean     |

The new structure is

| Column Name              | Type        |
|--------------------------|-------------|
| InvoiceCreatedDate       | datetime    |
| InvoiceId                | invoice_id  |
| InvoiceCurrency          | text        |
| InvoiceDue               | amount      |
| InvoicePrice             | amount      |
| InvoiceFullStatus        | text        |
| InvoiceStatus            | text        |
| InvoiceExceptionStatus   | text        |
| PaymentId                | text        |
| PaymentReceivedDate      | datetime    |
| Rate                     | amount      |
| PaymentAddress           | text        |
| PaymentMethodId          | text        |
| PaymentCurrency          | text        |
| PaymentAmount            | amount      |
| PaymentMethodFee         | amount      |
| PaymentInvoiceAmount     | amount      |

## Export all metadata in the invoice export

The way I am handling this is by flattening the JSON into columns. For the carts, I create columns prefixed by the cart item id.

See here `Black-tea-count` for example.

image

The field `Size` and `New Field 1` comes from forms.

## Remove invoice specific data from lines if there is more than one payment

When an invoice have more than two payments, a user not being careful might use Excel and sum duplicated values  of the same invoice more than once. For example, those two lines are from the same invoice, here was the data before:

Before:

image

Now:

image

The old behavior was very perverse, because for 99.999% of cases, the merchant will not have any issue. But one day, somebody makes two payments to the same invoice, then the merchant will end up counting things double.

For this reason, the metadata fields also only appears for the first payment.

## Added a button in the invoice list to easily browse the report

image


pavlenex:
Reading just the description, the direction of this seems good to me.

dstrukt:
+1 cACK, love to see these changes!

PR #6836, opened 2025-07-08
feat: store sales summary

IzyPro:
feat: Store Sales Summary
Time-based store specific sales metrics 
Provides API endpoint for https://github.com/btcpayserver/app/issues/91 and https://github.com/btcpayserver/app/issues/107

image

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldReportsController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldReportsController.cs
@@ -14,6 +14,8 @@ using BTCPayServer.Services;
 using System.Linq;
 using System.Threading;
 using BTCPayServer.Abstractions.Constants;
+using BTCPayServer.Services.Invoices;
+using BTCPayServer.Services.Rates;
 
 namespace BTCPayServer.Controllers.GreenField;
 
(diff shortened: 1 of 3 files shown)


IzyPro:
@NicolasDorier do let me know if this concept is in alignment
If yes, i'd proceed with documentation and tests. Otherwise, let me know what needs to change

dstrukt:
!!!

Excited to see this one come together.

BTCPayServer/Controllers/GreenField/GreenfieldReportsController.cs
```diff
@@ -66,5 +71,41 @@ public async Task<IActionResult> StoreReports(string storeId, [FromBody] StoreRe
...
+            return Ok(new SalesSummary());
+
+        var summary = new SalesSummary
+        {
+            GrossSales = invoices.Sum(x => x.GetPayments(false).Sum(y => y.PaidAmount.Gross)),
+            NetSales = invoices.Sum(x => x.GetPayments(false).Sum(y => y.PaidAmount.Net)),
+            Refunds = invoices.Sum(x => x.Refunds.Count),
+            SalesCount = invoices.LongLength,
+            Currency = invoices.FirstOrDefault()?.GetPayments(false)?.FirstOrDefault()?.Currency,
+            Taxes = invoices.Sum(x => x.TaxInPaymentCurrency()),
```
NicolasDorier, line 101: Why are you using payment currency here?
You are summing amounts that may be in different currencies. Not only payments can be in different currencies, but invoices too. So you can't just sum them.

NicolasDorier:
@IzyPro if this is for the app, I think you should rather expose this API in the plugin of the app than in Core.
In Core, we should make an API over the concept of reports that we already have.

Reports also have summaries, though they are calculated client side currently, so some non trivial work would need to be done for an endpoint that expose only the summaries.

However, your API as is isn't good, even in the plugin. See my comment (you are adding different currencies together)

IzyPro:
Alright
I will look at the concept of reports that we already have and find a way to work that into the app

PR #6841, opened 2025-07-10
Add rate information in on chain wallet and reports

NicolasDorier:
## Motivation

Currently, we only show the exchange rate between the payment currency and the invoice currency. For example, if a USD invoice is paid with BTC, the merchant can see the BTC/USD rate.

However, there are many other contexts where merchants would benefit from seeing exchange rate information—for example, in the wallet transaction list (where a transaction may or may not be related to an invoice), and in reports.

Additionally, in some cases, a merchant may need to know the rate of more than one currency. For instance, when sending funds to contributors of the BTCPay Server Foundation, contributors upload USD- or EUR-denominated invoices, payments are made in BTC, but tax calculations require the JPY rate. In such cases, the store should be able to track not only the exchange rate for the invoice currency, but also for JPY.

This PR adds a new option to specify additional tracked currencies for a store.

### Configure the additional tracked currency in the store settings

In the example of the BTCPay Server Foundation, we would fill this new option with `JPY`.

image

### Rate information in the wallet transaction list

Rate information now appears in the wallet transaction list.

image

### Rate information in the Wallet report

image

### Rate information in the Invoice report

I decided against truncating the rate, as it may have significant repercussion on accounting if the price of a crypto unit is very small.

image

### Easily go to Wallet report from the wallet transactions list

Similar to what we did in #6835, I added a `Reporting` button.

image

## Implementation details

There are two types of rates:
* Wallet Rates: Which is the rates recorded when we detect a transaction for a wallet.
* Invoice Rates: The rates recorded at the creation of an invoice.

For the "Wallets" report and in wallet transactions list, we are mixing both. When two values conflict, the invoice rate is chosen.

If the tracking of additionally tracked currency fails at invoice creation time, the invoice will still properly generate, and the data will be missing.

### DIFF
--- a/BTCPayServer.Client/Models/StoreBaseData.cs
+++ b/BTCPayServer.Client/Models/StoreBaseData.cs
@@ -46,6 +46,7 @@ namespace BTCPayServer.Client.Models
         public double? PaymentTolerance { get; set; }
         public bool? AnyoneCanCreateInvoice { get; set; }
         public string DefaultCurrency { get; set; }
+        public List<string> AdditionalTrackedRates { get; set; }
 
         public bool? LightningAmountInSatoshi { get; set; }
         public bool? LightningPrivateRouteHints { get; set; }
--- a/BTCPayServer.Tests/CSVInvoicesTester.cs
+++ b/BTCPayServer.Tests/CSVInvoicesTester.cs
@@ -1,0 +1,46 @@
+﻿using System.Linq;
+using Xunit;
+
+namespace BTCPayServer.Tests;
+
+internal class CSVInvoicesTester(string text) : CSVTester(text)
+{
+    string invoice = "";
+    int payment = 0;
+
+    public CSVInvoicesTester ForInvoice(string invoice)
+    {
+        this.payment = 0;
+        this.invoice = invoice;
+        return this;
+    }
+    public CSVInvoicesTester SelectPayment(int payment)
+    {
+        this.payment = payment;
+        return this;
+    }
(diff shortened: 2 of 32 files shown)


pavlenex:
@NicolasDorier Any way we can showcase total in fiat as well?

> This currently only works with transactions paying an invoice.

How can this be unblocked? Most of outgoing or incoming wallet transactions are probably not related to the invoice, why can't we just create on the fly invoice for incoming transactions, just like we do for LNURL payments?

NicolasDorier:
> why can't we just create on the fly invoice for incoming transactions, just like we do for LNURL payments?

We could, that would not solve the rate of outgoing transactions though.

pavlenex:
cACK, thanks for a clear description of situation!

dstrukt:
There's a typo `Addtional` `Additional` on the label.

dstrukt (commented):
There's a typo `Addtional`, should be`Additional` on the label.

Also, would be good to know how well this handles 5+ currencies on mobile, I believe the row is scrollable, but worth checking if you haven't.

NicolasDorier:
> @NicolasDorier Any way we can showcase total in fiat as well?
> 
> > This currently only works with transactions paying an invoice.
> 
> How can this be unblocked? Most of outgoing or incoming wallet transactions are probably not related to the invoice, why can't we just create on the fly invoice for incoming transactions, just like we do for LNURL payments?

This has been addressed, we now show rates for ingoing and outgoing transactions, even those unrelated to invoices.

PR #6846, opened 2025-07-14
Fix: Improve the responsivity of the Reporting page

NicolasDorier:
With this improvement, bye bye to the flakyness of the Reporting page.

Before

### DIFF
--- a/BTCPayServer/Views/UIReports/StoreReports.cshtml
+++ b/BTCPayServer/Views/UIReports/StoreReports.cshtml
@@ -70,25 +70,19 @@
                        data-fdtp='{ "enableTime": true, "enableSeconds": true, "dateFormat": "Y-m-d H:i:S", "time_24hr": true, "defaultHour": 0 }'
                        placeholder="@StringLocalizer["End Date"]" />
             </div>
-        </div>
-
-        <div id="searchGroup" v-cloak class="form-group">
-            <button id="searchBtn" class="btn btn-primary text-nowrap" :disabled="loading" type="button">
-                <span v-if="loading" class="spinner-border spinner-border-sm me-2" role="status"></span>
-                <span v-else class="me-2"><vc:icon symbol="actions-search" /></span>
-                <span text-translate="true">Search</span>
-            </button>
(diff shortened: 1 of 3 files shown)


dstrukt (changes requested):
- [ ] Left align the icon and text inside the button
- [ ] the Search icon also looks a little small? the standard sizings we use are 16x16 or 24x24, ideally the latter, so worth checking if it matches either of those (can't determine until i get a build)
- [ ] Push the Search button to the same line as the start/end inputs to get us a bit of vertical space back, should sit on the same bottom line as the inputs

Otherwise, initial impressions - looking good!

Excited to test this one out.

NicolasDorier:
@dstrukt made it a bit better. About the exact size, I am not sure. Bootstrap define the image size in button through `rem`. I adjusted it to be a bit bigger (1.25 rem) and align to text-bottom, it feels more natural. Will provide you a way to tweak later.

image

@pavlenex told me you didn't really like the tabs in other parts of the site.
Reason I changed to tab is because clicking on it reload the data (as if it was a new page), but changing dates requires now to click on `Search`.

If I kept the current dropdown, it felt strange that selecting in the dropdown cause auto reload, but not the date filter.

dstrukt:
Sounds good, and looks great - thanks for the adjustments!

Only other thing I'd do is left align the icon/text in the Search bar and this would be GTG!

NicolasDorier:
@dstrukt It is already left align, but all buttons have a 24 px padding. This is too much ?

dstrukt:
Typically the interior padding for buttons is `8px 16px` or `16px 24px` for the smaller and larger buttons respectively, or more simply, double the left/right to the top/bottom.

NicolasDorier:
@dstrukt all buttons we use (the action buttons such as Save) have 24px. If I change this, this would need to be changed everywhere?

Or the search button should be a different type of button?

NicolasDorier:
@dstrukt see in the PR description, I updated the button.

The button and logo were centered in the button, but the balance felt visually off. So I applied a negative margin left of half the size of the icon so it feels more balanced to the eyes.

I also fixed the button size jumping when it is loading.

PR #6853, opened 2025-07-17
Make sure that mempool space fee rate doesn't drop below node's minRelayTxFee

NicolasDorier:
Since miners are starting mining sub 1 sat/vbytes transaction, it is possible that in the future mempool space drop the minimum fee rate to less that 1 too.

If that is the case, we should make sure it is at least as high as the Bitcoin node or it will not be broadcastable.

I modified the `FeeProviderFactory` to clamp the value to the minimum mempool fee and 2000 vsat/b.


NicolasDorier:
@coderabbitai  review

PR #6876, opened 2025-07-30
Fix time icon spacing in wallet transactions header

### LINKED ISSUE #6874: [Bug]: Minor visual time icon bug in the wallet transaction view
## currently 
Image

## before 

Image

Regtest

### What happened?

A minor visual inconsistency 

1. Use regtest mode, and use cheat mode to receive transactions to the wallet
2. See the icon (icon is only visible if there are transaction to the wallet)
Reproducible easily on http://mainnet.demo.btcpayserver.org if wallet has transactions

### Relevant log output

```shell

```

### What browser do you use?

Safari

### Additional information

_No response_

bc1cindy:
Increase padding from p-0 to p-1 for better visual spacing between Date text and time format toggle icon.

Fixes #6874


thgO-O (changes requested):
Thanks for the contribution, @bc1cindy! The fix works, but I think p-2 would look better visually.

Screenshot 2025-07-30 at 22 20 57

PR #6878, opened 2025-07-31
Asking for confirmation to display QR code if user is store owner

rockstardev:
Fixes: #6763

Switched to using Store Owner instead of Server Admin in order not to do deeper refactoring.

https://github.com/user-attachments/assets/28ec47d6-3d79-4ad0-8324-186f1048425a


pavlenex (approved):
cACK

PR #6887, opened 2025-08-08
Fix RBF transaction handling in wallet balance and coin selection

### LINKED ISSUE #6884: [Bug]: When incoming transaction is RBF'd don't show it in total balance or coin selection

We received a report on Telegram from a user who tried to sweep their balance but received a vague error message when trying to send `RPC Error while broadcasting: RPC_TRANSACTION_ERROR General error during transaction submission bad-txns-inputs-missingorspent` 

Upon further investigation, user mentioned that the transaction from the wallet isn't visible on memool.space and has been received 2 days ago.

Image

Image

That felt odd, because afair we allow users to construct transaction without confirmation.

I initially suggested user to re-broadcast the tx, but wasn't quite sure what happened there.

Image

After user send a screenshot  above @petzsch realized that the transaction has likely been RBF'd by the customer.

So what happened here is that there was an incoming transaction that got RBF'd, but it seems that we still show this transaction in the total balance and in the transaction list as well as the coin selection.

I am not quite sure what's the right way to handle this, but it would be worth looking into the code trying to understand how we handle this. Natively, to me it seems that once transaction is RBF'd we at least shouldn't show it in the total balance or potentially not in the coin selection. 

Leaving this up to @NicolasDorier to try to examine and reproduce once he's available, but anyone else willing to try to reproduce this and solve it feel free to.

unknown

### What happened?

/

1. Receive a tx that's RBF'd by a customer
2. Try to sweep the wallet
3. You'll get error input message

Instead we probably shouldn't account for RBF'd transaction in the total balance.

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

bc1cindy:
- Exclude RBF'd transactions from available balance calculation
- Filter out UTXOs from replaced transactions in coin selection
- Add visual indicator for replaced transactions in transaction history
- Implement caching for replaced transaction detection
- Prevent 'bad-txns-inputs-missingorspent' errors when attempting to spend

Fixes #6884

### DIFF
--- a/BTCPayServer/Services/Wallets/BTCPayWallet.cs
+++ b/BTCPayServer/Services/Wallets/BTCPayWallet.cs
@@ -256,6 +256,17 @@ namespace BTCPayServer.Services.Wallets
         List<TransactionInformation> dummy = new List<TransactionInformation>();
         public async Task<IList<TransactionHistoryLine>> FetchTransactionHistory(DerivationStrategyBase derivationStrategyBase, int? skip = null, int? count = null, TimeSpan? interval = null, CancellationToken cancellationToken = default)
         {
+            HashSet<uint256> replacedTxs = null;
+            try
+            {
+                replacedTxs = await GetReplacedTransactions(derivationStrategyBase, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                Logs.PayServer.LogWarning(ex, $"Error getting replaced transactions for history: {derivationStrategyBase}");
+                replacedTxs = new HashSet<uint256>();
+            }
+
             // This is two paths:
             // * Sometimes we can ask the DB to do the filtering of rows: If that's the case, we should try to filter at the DB level directly as it is the most efficient.
             // * Sometimes we can't query the DB or the given network need to do additional filtering. In such case, we can't really filter at the DB level, and we need to fetch all transactions in memory.
@@ -274,7 +285,9 @@ namespace BTCPayServer.Services.Wallets
                     if (timestampLimit is DateTimeOffset l &&
                         t.Timestamp <= l)
                         break;
-                    lines.Add(FromTransactionInformation(t));
+                    var line = FromTransactionInformation(t);
+                    line.IsReplaced = replacedTxs.Contains(t.TransactionId);
+                    lines.Add(line);
                 }
                 return lines;
             }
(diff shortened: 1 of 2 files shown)


bc1cindy:
Captura de Tela 2025-08-15 às 10 52 43

BTCPayServer/Services/Wallets/BTCPayWallet.cs
```diff
@@ -509,31 +524,107 @@ public async Task<ReceivedCoin[]> GetUnspentCoins(
...
+        public async Task<GetBalanceResponse> GetBalance(DerivationStrategyBase derivationStrategy, CancellationToken cancellation = default(CancellationToken))
         {
-            return _MemoryCache.GetOrCreateAsync("CACHEDBALANCE_" + derivationStrategy, async (entry) =>
+            var balance = await _MemoryCache.GetOrCreateAsync("CACHEDBALANCE_" + derivationStrategy, async (entry) =>
             {
                 var result = await _Client.GetBalanceAsync(derivationStrategy, cancellation);
                 entry.AbsoluteExpiration = DateTimeOffset.UtcNow + CacheSpan;
                 return result;
             });
+
```
NicolasDorier, line 599: You should not run this everytimes `GetBalance` is called.
Instead, this check need to be done after `_Client.GetBalanceAsync` above, so the cache is actually useful.

Did you also properly confirmed this is the problem? `GetBalance` should already not count what has been RBF'd. If not, does the bug come from the `GetBalanceAsync` to NBX, or does it come from the cache not being invalidated?

BTCPayServer/Services/Wallets/BTCPayWallet.cs
```diff
@@ -509,31 +524,107 @@ public async Task<ReceivedCoin[]> GetUnspentCoins(
...
+                            Total = balance.Total,
+                            Immature = balance.Immature,
+                            Confirmed = balance.Confirmed
+                        };
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Logs.PayServer.LogWarning(ex, $"Error adjusting balance for replaced transactions: {derivationStrategy}");
```
NicolasDorier, line 624: Do not catch. If this call crash, we should crash.

BTCPayServer/Services/Wallets/BTCPayWallet.cs
```diff
@@ -256,6 +256,17 @@ private async Task<UTXOChanges> GetUTXOChanges(DerivationStrategyBase strategy,
         List<TransactionInformation> dummy = new List<TransactionInformation>();
         public async Task<IList<TransactionHistoryLine>> FetchTransactionHistory(DerivationStrategyBase derivationStrategyBase, int? skip = null, int? count = null, TimeSpan? interval = null, CancellationToken cancellationToken = default)
         {
+            HashSet<uint256> replacedTxs = null;
+            try
+            {
+                replacedTxs = await GetReplacedTransactions(derivationStrategyBase, cancellationToken);
+            }
+            catch (Exception ex)
```
NicolasDorier, line 264: Move the handler inside `GetReplacedTransactions`

BTCPayServer/Views/UIWallets/_WalletTransactionsList.cshtml
```diff
@@ -22,7 +22,13 @@
                 display-inline="true"/>
         </td>
         <td>
-            <vc:truncate-center text="@transaction.Id" link="@transaction.Link" classes="truncate-center-id" />
+            <div class="d-flex align-items-center gap-2">
+                <vc:truncate-center text="@transaction.Id" link="@transaction.Link" classes="truncate-center-id" />
+                @if (transaction.HistoryLine?.IsReplaced == true)
+                {
+                    <span class="badge bg-light text-muted" title="Transaction was replaced by RBF">RBF'd</span>
```
NicolasDorier, line 29: A RBF'd transaction shouldn't even show in the list.

NicolasDorier (changes requested):
I've noted several things. I am not convinced this PR, even taking into account my points, is actually fixing the issue. And if it is, is it a bug of NBX that this PR is tapping over, or is it really a bug of BTCPay?

Please, make a test (or improve tests on RBF) trying to replicate the issue so I can be convinced of it.

bc1cindy:
Thanks a lot fot the review @NicolasDorier but i'll need to close this PR for now.

Issue #6872, opened 2025-07-30
[Bug]: Manager role has no Dashboard (only Owner can view Dashboard)


GittyGitGitGitty:
### What is your BTCPay version?

BTCPay v2.1.6

### How did you deploy BTCPay Server?

btcpayserver-docker

### What happened?

The Dashboard contains essential Store information for a Manager: wallet balance, recent transactions, recent invoices and various analytics. This information isn't inherently sensitive (for a store Manager) and I would argue that they are in fact essential for a Manager role.

This seems like a bug as I see no documentation supporting this as a security feature. What would the rationale be to not have essential Store information available to a Manager role?

### How did you encounter this bug?

1. Create user
2. Set role to Manager
3. Login as Manager user
4. See no Dashboard

### Relevant log output

```shell

```

### What browser do you use?

Brave 1.80.124 (Official Build) (64-bit)
Chromium: 138.0.7204.168

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

pavlenex:
Thanks for bringing this one up, I absolutely agree, and don't think this was intentional. @NicolasDorier  @rockstardev this could be a good first issue for a new contributor, provided that we don't have disagreement that store manger should see the dashboard and agree this is a bug and not a desired behaviour.

NicolasDorier:
The problem we have is that in some instance, I believe the dashboard has confidential data.

For example, it exposes the hot wallet, the sales, and I think it also exposes access to LND node management tools. 
I will have to check that.

NicolasDorier:
So here is the problem. Fundamentally the permissions aren't tied to the Dashboard, but to the individual tiles.

For example, if you don't have the permission to setup a store (can modify store), you should not see the tile which show a call to action to setup the wallet.

Another example: If you can't use the lightning node (as it opening channel, pay with it, administer it), we shouldn't show the tile to for connecting to RTL.

While we could adjust the permissions of those tiles individually, the Dashboard is static. We can't just adapt the layout automatically.

Here is an example where a store didn't set up a Bitcoin wallet for a manager. The manager can't setup the bitcoin wallet, so I removed the call to action. But now it looks weird.

<img width="2280" height="934" alt="Image" src="https://github.com/user-attachments/assets/d19c0ade-6060-4253-ab68-272f363535d9" />

It means that for such a feature to be useful, a deep redesign of the dashboard is needed, with a proper layout system.
[...]

PR #6888, opened 2025-08-08
Fix Manager role Dashboard access (#6872)

bc1cindy:
Manager role was unable to access store Dashboard due to incorrect authorization policy. Dashboard endpoints were using CanModifyStoreSettings which requires CanUseLightningNodeInStore permission that Manager lacks.

Changed to CanViewStoreSettings policy which Manager has and is appropriate for read-only Dashboard operations.


bc1cindy:
Added commit to update the Selenium test that was expecting the old behavior where Manager role was denied Dashboard access. The failing test has been updated to reflect the intended behavior described in issue #6872.

bc1cindy:
Need to investigate alternative solutions that don't affect Server Admin permissions.

PR #6896, opened 2025-08-20
fix: ensure unlisted installed plugins appear as updatable

### LINKED ISSUE #6893: Manage plugins page no longer displays available updates for displayed plugins that are not Listed
This issue is likely related to the recent change made by me and @thgO-O. By default, we no longer return a list of all plugins; instead, we only return the Listed ones. This causes the situation that if the installed plugin is not Listed, it won't be offered to be updated.

To address this, we should consider implementing a check on the installed plugins to determine if there are any new versions available.

Example: 

https://github.com/user-attachments/assets/0f5096dc-9c1a-4fb8-a0e5-c23e4708ce43

thgO-O:
This PR ensure installed but unlisted plugins are detected as updatable by skipping core installed plugins and adding logic in GetUpdatesForUnlistedInstalledAsync

Fixes #6893 

https://github.com/user-attachments/assets/6c712a14-b8fe-4069-86cf-ecd2ccc8506a

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.Plugins.cs
+++ b/BTCPayServer/Controllers/UIServerController.Plugins.cs
@@ -26,12 +26,14 @@ namespace BTCPayServer.Controllers
             {
                 availablePlugins = await pluginService.GetRemotePlugins(search);
             }
-            catch (Exception)
+            catch (Exception ex)
             {
+                var errorDetail = ex.GetBaseException().Message;
+
                 TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Message = StringLocalizer["Remote plugins lookup failed. Try again later."].Value
+                    Message = StringLocalizer["Remote plugins lookup failed. Try again later. Error: {0}", errorDetail].Value
                 });
                 availablePlugins = Array.Empty<PluginService.AvailablePlugin>();
             }
--- a/BTCPayServer/Controllers/UIServerController.Plugins.cs
+++ b/BTCPayServer/Controllers/UIServerController.Plugins.cs
@@ -28,12 +28,10 @@ namespace BTCPayServer.Controllers
             }
             catch (Exception ex)
             {
-                var errorDetail = ex.GetBaseException().Message;
-
                 TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Message = StringLocalizer["Remote plugins lookup failed. Try again later. Error: {0}", errorDetail].Value
+                    Message = StringLocalizer["Remote plugins lookup failed. Try again later. Error: {0}", ex.Message].Value
                 });
                 availablePlugins = Array.Empty<PluginService.AvailablePlugin>();
             }
--- a/BTCPayServer/Plugins/PluginBuilderClient.cs
+++ b/BTCPayServer/Plugins/PluginBuilderClient.cs
@@ -53,8 +53,6 @@ namespace BTCPayServer.Plugins
 
     public class PluginBuilderClient
     {
-        #nullable enable
-
         HttpClient httpClient;
         public HttpClient HttpClient => httpClient;
         private readonly ILogger<PluginBuilderClient> _logger;
@@ -99,7 +97,7 @@ namespace BTCPayServer.Plugins
                    ?? throw new InvalidOperationException();
         }
 
-        public async Task<PublishedVersion[]?> GetInstalledPluginsUpdates(
+        public async Task<PublishedVersion[]> GetInstalledPluginsUpdates(
             string btcpayVersion,
             bool includePreRelease,
             IEnumerable<InstalledPluginRequest> plugins)
@@ -123,12 +121,12 @@ namespace BTCPayServer.Plugins
             catch (HttpRequestException ex)
             {
                 _logger.LogWarning(ex, "Failed to check for plugins updates");
-                return null;
+                return Array.Empty<PublishedVersion>();
             }
             catch (JsonException ex)
             {
                 _logger.LogWarning(ex, "Failed to parse plugins updates response");
-                return null;
+                return Array.Empty<PublishedVersion>();
             }
         }
     }
--- a/BTCPayServer/Plugins/PluginBuilderClient.cs
+++ b/BTCPayServer/Plugins/PluginBuilderClient.cs
@@ -53,12 +53,12 @@ namespace BTCPayServer.Plugins
 
     public class PluginBuilderClient
     {
-        HttpClient httpClient;
-        public HttpClient HttpClient => httpClient;
+        private readonly HttpClient _httpClient;
+        public HttpClient HttpClient => _httpClient;
         private readonly ILogger<PluginBuilderClient> _logger;
         public PluginBuilderClient(HttpClient httpClient, ILogger<PluginBuilderClient> logger)
         {
-            this.httpClient = httpClient;
+            _httpClient = httpClient;
             _logger = logger;
         }
         static JsonSerializerSettings serializerSettings = new() { ContractResolver = new Newtonsoft.Json.Serialization.CamelCasePropertyNamesContractResolver() };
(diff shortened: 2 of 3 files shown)


BTCPayServer/Plugins/PluginService.cs
```diff
@@ -20,19 +23,35 @@ public class PluginService
         private readonly IOptions<DataDirectories> _dataDirectories;
         private readonly PoliciesSettings _policiesSettings;
         private readonly PluginBuilderClient _pluginBuilderClient;
+        private readonly ILogger<PluginService> _logger;
+        private static readonly HashSet<string> _builtInPluginIdentifiers = new(StringComparer.OrdinalIgnoreCase)
```
NicolasDorier: Do not hardcode. You have `btcPayServerPlugins`, so you can get them through `btcPayServerPlugins.Where(c => c.SystemPlugin).Select(c => c.Identifier).ToHashSet()`

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -20,19 +23,35 @@ public class PluginService
...
+            ILogger<PluginService> logger
+            )
         {
             LoadedPlugins = btcPayServerPlugins;
             Installed = btcPayServerPlugins.ToDictionary(p => p.Identifier, p => p.Version, StringComparer.OrdinalIgnoreCase);
             _pluginBuilderClient = pluginBuilderClient;
             _dataDirectories = dataDirectories;
             _policiesSettings = policiesSettings;
             Env = env;
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
NicolasDorier: no need to check for null, DI do not inject null.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -54,20 +73,114 @@ public async Task<AvailablePlugin[]> GetRemotePlugins(string searchPluginName)
...
+            CancellationToken ct = default)
+        {
+            var listedIdentifiers = new HashSet<string>(
+                listedPlugins.Select(p => p.Identifier),
+                StringComparer.OrdinalIgnoreCase);
+
+            var installedToCheck = Installed
+                .Where(installedPlugin => !_builtInPluginIdentifiers.Contains(installedPlugin.Key) &&
+                              !listedIdentifiers.Contains(installedPlugin.Key))
+                .ToList();
```
NicolasDorier, line 75: Do not used `Installed` and `_builtInPluginIdentifiers`, as explained before, just used `LoadedPlugins.Where(c => !c.SystemPlugin && !listedIdentifiers.Contains(installedPlugin.Key))`

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -54,20 +73,114 @@ public async Task<AvailablePlugin[]> GetRemotePlugins(string searchPluginName)
...
+
+                    if (latestCandidate == null)
+                        return;
+
+                    var latestAvailable = MapToAvailablePlugin(latestCandidate.VersionInfo);
+
+                    if (latestAvailable is null)
+                        return;
+
+                    if (latestAvailable.Version <= installedVersion)
```
NicolasDorier: Doesn't `GetRemotePlugins` returns all the plugins regardless of their version? Why this check here, when the published plugins doesn't check?

NicolasDorier (changes requested):

BTCPayServer/Plugins/PluginBuilderClient.cs
```diff
@@ -85,9 +92,47 @@ public async Task<PublishedVersion[]> GetPluginVersionsForDownload(string identi
...
+            {
+                _logger.LogWarning(ex, "Failed to parse plugins updates response");
+                return Array.Empty<PublishedVersion>();
+            }
+            catch (TaskCanceledException ex)
+            {
+                _logger.LogWarning(ex, "Timeout while checking plugins updates");
+                return Array.Empty<PublishedVersion>();
+            }
+        }
```
NicolasDorier: Do not catch if there is nothing you can do about it. Just let the exception bubble up.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -49,32 +55,94 @@ public Version GetVersionOfPendingInstall(string plugin)
...
+
+            return plugins.ToArray();
+        }
+
+#nullable enable
+        private AvailablePlugin? MapToAvailablePlugin(PublishedVersion publishedVersion)
+        {
+            if (publishedVersion.ManifestInfo is null)
+            {
+                _logger.LogWarning("Skipping published plugin with missing ManifestInfo (BuildId: {BuildId}, ProjectSlug: {ProjectSlug})", publishedVersion.BuildId, publishedVersion.ProjectSlug);
```
NicolasDorier: Remove the logs, it's noisy... some plugins don't have manifest (for example when you dev a plugin and just reference the dll)

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -49,32 +55,94 @@ public Version GetVersionOfPendingInstall(string plugin)
...
+
+            AvailablePlugin? availablePlugin;
+
+            try
+            {
+                availablePlugin = publishedVersion.ManifestInfo.ToObject<AvailablePlugin>();
+            }
+            catch (JsonException ex)
+            {
+                _logger.LogWarning(ex, "Failed to parse ManifestInfo (BuildId: {BuildId}, ProjectSlug: {ProjectSlug})", publishedVersion.BuildId, publishedVersion.ProjectSlug);
```
NicolasDorier: If that fail, let it throw. It shouldn't happen, if it does, that's a bug, and better to just let it bubble up.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -49,32 +55,94 @@ public Version GetVersionOfPendingInstall(string plugin)
...
+                return null;
+            }
+
+            if (availablePlugin is null)
+            {
+                _logger.LogWarning("Failed to deserialize ManifestInfo to AvailablePlugin (BuildId: {BuildId}, ProjectSlug: {ProjectSlug})", publishedVersion.BuildId, publishedVersion.ProjectSlug);
```
NicolasDorier: don't log, let it crash.

NicolasDorier (changes requested):
Made some comment, but in general: What is unexpected shouldn't be logged and ignored, it should be crashing.

The UI is already taking care of catching crashes of `GetRemotePlugins`. (Though, in `ListPlugins`, we aren't showing or displaying the specific exception information, we should do it...)

thgO-O:
> Made some comment, but in general: What is unexpected shouldn't be logged and ignored, it should be crashing.
> 
> Not doing so add too much verbosity and make it hard to distinguish between an "expected crash" (for example, failure to reach the plugin website because of internet connection issue or server going down) and a bug.
> 
> The UI is already taking care of catching exception thrown by `GetRemotePlugins`. (Though, in `ListPlugins`, we aren't showing or displaying the specific exception information, we should do it...)

Thanks for the review and explanations! I removed logs and also added the specific exception information on ListPlugins view.

Screenshot 2025-09-02 at 14 24 45

rockstardev (approved):
@NicolasDorier this looks good to me, tested on local and it works as expected.

BTCPayServer/Controllers/UIServerController.Plugins.cs
```diff
@@ -26,12 +26,14 @@ public async Task<IActionResult> ListPlugins(
...
-            catch (Exception)
+            catch (Exception ex)
             {
+                var errorDetail = ex.GetBaseException().Message;
+
                 TempData.SetStatusMessageModel(new StatusMessageModel
                 {
                     Severity = StatusMessageModel.StatusSeverity.Error,
-                    Message = StringLocalizer["Remote plugins lookup failed. Try again later."].Value
+                    Message = StringLocalizer["Remote plugins lookup failed. Try again later. Error: {0}", errorDetail].Value
```
NicolasDorier:
```suggestion
                    Message = StringLocalizer["Remote plugins lookup failed. Try again later. Error: {0}", ex.Message].Value
```

BTCPayServer/Controllers/UIServerController.Plugins.cs
```diff
@@ -26,12 +26,14 @@ public async Task<IActionResult> ListPlugins(
             {
                 availablePlugins = await pluginService.GetRemotePlugins(search);
             }
-            catch (Exception)
+            catch (Exception ex)
             {
+                var errorDetail = ex.GetBaseException().Message;
```
NicolasDorier:
```suggestion

```

NicolasDorier (changes requested):
Just fix this, then you can merge

BTCPayServer/Controllers/UIServerController.Plugins.cs
rockstardev: 9caea0958a1cca55b30c0c820cb7a90e45d1c3f2

BTCPayServer/Controllers/UIServerController.Plugins.cs
rockstardev: 9caea0958a1cca55b30c0c820cb7a90e45d1c3f2

rockstardev:
I pushed commit to save Thgoo some time & merged, we good.

PR #6909, opened 2025-09-05
fix: update docker certificates

jackstar12:
- **fix: install `ca-certificates` in dockerfile**
- **chore: bump `dotnet/aspnet` base image to 8.0.18**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN cd BTCPayServer && dotnet publish -p:GitCommit=${GIT_COMMIT} --output /app/ 
 
 FROM mcr.microsoft.com/dotnet/aspnet:8.0.11-bookworm-slim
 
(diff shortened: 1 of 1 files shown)


rockstardev (approved):
@NicolasDorier for context - this fix originated from Boltz indentifying that in their plugin they can't do `curl https://mempool.space` because of invalid certificates.

You can verify it yourself with:

docker run -it mcr.microsoft.com/dotnet/aspnet:8.0.3-bookworm-slim bash
apt update
apt install -y curl
curl https://mempool.space
apt install -y ca-certificates
curl https://mempool.space

So the fix is to:
1. bump base image to 8.0.18
2. also add ca-certificates installed, just for the future

NicolasDorier:
@jackstar12 @rockstardev does the bump of asp.net needed? NBX is using 8.0.11, so this would avoid it to share the same base image.

NicolasDorier:
I will bump both...

Issue #6911, opened 2025-09-07
[Bug]: 0 amount invoice for BTC-LN payment method


w0aw:
### What is your BTCPay version?

i am running bycpay v2.1.6.

we really need to generate the bolt11 invoices without any amount (0).

by default lnurl does not work with few wallets (muun/speed  wallet. etc)

we really need a uniform way to create bolt11 invoice where it is users choice to put in amount or keep it blank

thank you

### How did you deploy BTCPay Server?

docker compose on ec2

Kukks:
Not a bug, but a feature. After extensive testing of many wallets, 0 amount invoices had less support than Lnurl overall, not too mention a series of shortcomings over the years. The experience of lnurl is also superior, and required for certain features.

Your best bet would be to directly message wallets that do not support lnurl and add it. It's a very simple protocol and would take less than a day.

pavlenex:
I haven't encountered a wallet not supporting bolt11 zero amount in ages. @Kukks Which wallets currently don't support bolt11 zero amount? Can you be more specific about the shortcomings, it seems pretty obvious feature to add imo.

Kukks:
> I haven't encountered a wallet not supporting bolt11 zero amount in ages. @Kukks Which wallets currently don't support bolt11 zero amount? Can you be more specific about the shortcomings, it seems pretty obvious feature to add imo.


Any wallet that uses boltz or similar swap mechanisms would not work. 

So from the top of my head, aqua, bull bitcoin, misty, and even the boltz btcpay plugin itself would not work or support generating invoices in that style.

w0aw:
@Kukks @pavlenex Fun fact is btcpay itself cannot pay to bolt11 created by another btcpay. if we add 0 amount bolt11 invoices, it can be paid by another btcpay server.

and it will gave more freedom to online sites. 

each and every wallet out there support creating bolt11 invoice with 0 amount.

please add it. thank you ( it will help us to accept any amount of donations via bolt11)

currently we are relying on lnurl, (which is not working (for whatever reason) with most of wallets we tested - muun, wallet of satoshi, speed, few reported not working with cashApp)

Kukks:
> @Kukks @pavlenex Fun fact is btcpay itself cannot pay to bolt11 created by another btcpay. if we add 0 amount bolt11 invoices, it can be paid by another btcpay server.
> 
> and it will gave more freedom to online sites. 
> 
[...]

PR #6916, opened 2025-09-09
Add 0 amount (Top-up) bolt11 lightning invoice

w0aw:
This PR will remove restrictions and allow creating top-up bolt11 invoice.

please Add this feature. we really need it.

i am not sure if i did it correct or wrong, but i think it will work without any UI adjustments.

Thank you.

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
+++ b/BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
@@ -77,11 +77,6 @@ namespace BTCPayServer.Payments.Lightning
 
         public async Task ConfigurePrompt(PaymentMethodContext context)
         {
-            if (context.InvoiceEntity.Type == InvoiceType.TopUp)
-            {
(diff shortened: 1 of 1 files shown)


w0aw:
related to issue 6911 https://github.com/btcpayserver/btcpayserver/issues/6911

Kukks (changes requested):
1) I do not appreciate the backdoor navigation of the discussion this way.
2) this should be behind a configuration option, and toggled off by default to prevent different behavior to what is expected.
3) you should ping cashapp to support lnurl. It is trivial to implement and it's stupid to not have it in 2025.
4) you need to ensure that payment inwards and outwards work to 0 amount invoices too, from all ln nodes, as well as plugin added variants. 
5) you need to add test coverage

rockstardev:
Closing per explaination here: https://github.com/btcpayserver/btcpayserver/pull/6956?notification_referrer_id=NT_kwDOAE826rMxOTY3Nzk4NTI1Nzo1MTkxNDAy#issuecomment-3416929857

PR #6927, opened 2025-09-23
(Refactor) : Converted Selenium test for `CanUsePairing` and Others to playwright

Abhijay007:
This PR converts a few of the remaining tests: ref PR: https://github.com/btcpayserver/btcpayserver/pull/6861

- CanUsePredefinedRoles() - Tests predefined roles
- CanUsePairing() - Tests pairing functionality

Preview of playwright test for `CanUsePredefinedRoles`

playwrightTest-CanUsePredefinedRoles

Preview of playwright test for `CanUsePairing`

playwrightTest-CanUsePairing

### DIFF
--- a/BTCPayServer.Tests/PlaywrightTester.cs
+++ b/BTCPayServer.Tests/PlaywrightTester.cs
@@ -324,8 +324,6 @@ namespace BTCPayServer.Tests
         }
         public async Task Logout()
         {
-            if (!await Page.Locator("#Nav-Logout").IsVisibleAsync())
-                await GoToUrl("/account");
             await Page.Locator("#menu-item-Account").ClickAsync();
             await Page.Locator("#Nav-Logout").ClickAsync();
         }
--- a/BTCPayServer.Tests/PlaywrightTests.cs
+++ b/BTCPayServer.Tests/PlaywrightTests.cs
@@ -2255,12 +2255,15 @@ namespace BTCPayServer.Tests
 
             // Setup users
             var manager = await s.RegisterNewUser();
+            await s.GoToHome();
             await s.Logout();
             await s.GoToRegister();
             var employee = await s.RegisterNewUser();
+            await s.GoToHome();
             await s.Logout();
             await s.GoToRegister();
             var guest = await s.RegisterNewUser();
+            await s.GoToHome();
             await s.Logout();
             await s.GoToRegister();
 
@@ -2360,6 +2363,7 @@ namespace BTCPayServer.Tests
                 s.TestLogs.LogInformation($"Checking access to store page {path} as employee");
                 await s.AssertPageAccess(false, $"/stores/{storeId}/{path}");
             }
+            await s.GoToHome();
             await s.Logout();
 
             // Guest access
@@ -2384,6 +2388,7 @@ namespace BTCPayServer.Tests
                 s.TestLogs.LogInformation($"Checking access to store page {path} as guest");
                 await s.AssertPageAccess(false, $"/stores/{storeId}/{path}");
             }
+            await s.GoToHome();
             await s.Logout();
         }
 


BTCPayServer.Tests/PlaywrightTester.cs
```diff
@@ -296,8 +296,28 @@ public async Task GoToLogin()
         }
         public async Task Logout()
         {
-            await Page.Locator("#Nav-Account").ClickAsync();
-            await Page.Locator("#Nav-Logout").ClickAsync();
+            // Try to find the account nav or a logout button, or fallback to /logout
+            if (await Page.Locator("#Nav-Account").CountAsync() > 0 && await Page.Locator("#Nav-Account").IsVisibleAsync())
```
NicolasDorier: This is not good. Because the Count might be 0 just because the page hasn't fully loaded yet...
While the `.ClickAsync` will wait for the element to appear.

Abhijay007: dec80b7

BTCPayServer.Tests/PlaywrightTester.cs
```diff
@@ -296,8 +296,35 @@ public async Task GoToLogin()
...
+            {
+                await logoutAnchor.ClickAsync();
+            }
+            else
+            {
+                await GoToUrl("/logout");
+            }
+            // Wait for login form or login page URL as a sign of successful logout
+            await Page.WaitForURLAsync("**/login");
+            await Page.Locator("#LoginButton").WaitForAsync(new() { State = WaitForSelectorState.Visible, Timeout = 15000 });
```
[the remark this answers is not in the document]
Abhijay007: 35e378d

BTCPayServer.Tests/PlaywrightTester.cs
```diff
@@ -296,8 +296,31 @@ public async Task GoToLogin()
         }
         public async Task Logout()
         {
-            await Page.Locator("#Nav-Account").ClickAsync();
-            await Page.Locator("#Nav-Logout").ClickAsync();
+            // Try to find the account nav or a logout button, or fallback to /logout
+            var navAccount = Page.Locator("#Nav-Account");
+            var navLogout = Page.Locator("#Nav-Logout");
+            var logoutAnchor = Page.Locator("a[href='/logout']");
+            if (await navAccount.IsVisibleAsync())
```
NicolasDorier: `IsVisibleAsync()` has the same problem as `CountAsync()` , it doesn't wait. Why are you even calling it?

Abhijay007: 674ed05

Abhijay007:
ping @NicolasDorier

BTCPayServer.Tests/PlaywrightTester.cs
```diff
@@ -296,8 +296,15 @@ public async Task GoToLogin()
         }
         public async Task Logout()
```
NicolasDorier, line 325: why is this different from the Selenium version?

Abhijay007: Hi @NicolasDorier, the thing is, we can keep it like the Selenium version, but in some tests, there’s an exception, they don’t land on the home page or the page where the account elements are accessible before logout, which causes those tests to fail as it get timeout. We might need to modify those tests to navigate back to the home page before accessing Nav-Account.

The Selenium version relies on UI interactions, it clicks on navigation elements, first selecting the "Nav-Account" button, then "Nav-Logout", following the user’s typical workflow through the interface. However, the updated approach uses direct URL navigation, going straight to the /logout endpoint and completely bypassing the UI.

I believe the new approach is better because it goes directly to the logout URL instead of depending on UI elements, making it less brittle and faster since there’s no need to find and click multiple elements. But we can revert to the old approach as well; we’d just need to update some tests to ensure they’re on the home page before performing the logout. I appreciate your thoughts on the same to move forward.

Abhijay007:
ping @NicolasDorier

BTCPayServer.Tests/PlaywrightTester.cs
NicolasDorier: Need to find why this is the case. LogOut expects to be in the main page, if the test isn't in the main page, it means there is a bug either in the test itself or in the code.

Abhijay007: Fixed this. The navigation issue was that we were not using the 
```
 if (!Driver.PageSource.Contains("id=\"Nav-Logout\""))
                GoToUrl("/account");
``` 
statement from the selenium test in the logout method

NicolasDorier: Since it seems to be the only test depending on this behavior, can you move it out of LogOut, and just going to /action directly in the test before calling it?

Issue #6918, opened 2025-09-13
Add option to re-enable disabled plugins after a crash


rockstardev:
Currently, when a plugin crashes the server, it is automatically disabled. The only available action afterward is to uninstall it.

This creates unnecessary friction:

- If the crash was only temporary, the user must uninstall and reinstall the plugin.
- If the plugin is updated to a fixed version, reinstallation is still required instead of simply re-enabling.

It would be great to have another button to re-enable a disabled plugin without uninstalling it, so users can quickly recover from temporary issues or test new versions.

monicamuyama:
I would like to work on this issues

monicamuyama:
@prusnak @rockstardev @ryenski @dennisreimann would you mind reviewing when you get a chance, thank you

PR #6930, opened 2025-09-29
Issue 6918 re enable plugin

monicamuyama:
This PR addresses issue #6918.

Currently, disabled plugins can only be uninstalled. This adds a Re-enable button for plugins that have crashed or were disabled, allowing users to restore them without uninstalling/reinstalling.

Changes:
- PluginService: new EnablePlugin method that queues an 'enable' command.
- ServerController: new action for re-enabling a disabled plugin.
- Plugins.cshtml: added Re-enable button with tooltip indicating restart required.
- Uses existing disabled/commands file workflow; no changes to plugin loader logic.
- This the above changes for the disable button on installed plugin

Testing:
- Manual: Disabled a plugin → Re-enable button appears → clicking queues command → plugin restored on restart.


BTCPayServer/Plugins/PluginManager.cs
```diff
@@ -516,6 +516,11 @@ public static void DisablePlugin(string pluginDir, string plugin)
             QueueCommands(pluginDir, ("disable", plugin));
         }
 
+        public static void EnablePlugin(string pluginDir, string plugin)
```
thgO-O: This function is duplicated and is not being used. You can remove it please

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -287,6 +301,22 @@
                                     </form>
                                 }
                             }
+                            
+                            @if (DependentOn(plugin.Identifier))
```
thgO-O: I think this might be out of scope for the issue. The disabled state here refers specifically to plugins that crashed due to fatal errors and were automatically disabled. The re-enable action should only apply to those cases, not to the regular Installed Plugins list.

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -103,17 +103,31 @@
...
-							@if (uninstalled)
-							{
-								<button type="submit" class="btn btn-sm btn-outline-danger" disabled text-translate="true">Marked for deletion</button>
-							}
-							else
-							{
-								<button type="submit" class="btn btn-sm btn-outline-danger" text-translate="true">Uninstall</button>
-							}
-						</form>
+                        <div class="d-flex gap-2">
```
thgO-O, line 108: there’s an inconsistency here. Clicking Enable keeps the Uninstall button visible, but the reverse isn’t true. I think the behaviour should be consistent on both actions.

thgO-O (commented):
Great job, Monica! Thanks for catching this up. Left some notes, but looks good overall.

BTCPayServer/Views/UIServer/ListPlugins.cshtml
monicamuyama: Thank you this, it should be fine now

BTCPayServer/Views/UIServer/ListPlugins.cshtml
monicamuyama: This makes sense the current scope is for plugins auto-disabled after a crash.
In my implementation, I had also considered the case where an admin might want to manually disable or re-enable a plugin for maintenance or troubleshooting.
I thought it could improve the overall UX by giving admins more control, but I am open to hearing thoughts on whether that fits within the intended behavior or should be handled separately.

thgO-O: Yeah, but if I disable a plugin, the UI shows that it was disabled due to fatal errors, as shown in the image.
So if we go this route, we’d need a refactor to differentiate between a crash and a user-initiated disable.
I’m not sure it’s worth it, since the user can simply uninstall and reinstall the plugin for the similar effect.

Screenshot 2025-10-26 at 23 34 13

monicamuyama: Yeah that makes sense, I have removed the disable button for now to keep things consistent with the current behavior. We can revisit this in the future if there is a need to allow admins to manually disable plugins. At that point a refactor to properly distinguish between crash-related and user-initiated disables would fit in.
Thanks for your insights

BTCPayServer/Controllers/UIServerController.Plugins.cs
```diff
@@ -86,6 +86,35 @@ public IActionResult UnInstallPlugin(
...
+            {
+                Message = StringLocalizer["Plugin scheduled to be enabled."].Value,
+                Severity = StatusMessageModel.StatusSeverity.Success
+            });
+
+            return RedirectToAction("ListPlugins");
+        }
+
+        [HttpPost("server/plugins/disable")]
+        public IActionResult DisablePlugin(
```
thgO-O: As we decided to remove the disable option, this is no longer being used. Please remove it

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -187,6 +187,14 @@ public void UninstallPlugin(string plugin)
             PluginManager.QueueCommands(dest, ("delete", plugin));
         }
 
+        public void EnablePlugin(string plugin){
+            PluginManager.QueueCommands(_dataDirectories.Value.PluginDir, ("enable", plugin));
+        }
+
+        public void DisablePlugin(string plugin){
```
thgO-O: This also can be removed

monicamuyama:
Thank you @thgO-O for your review and comments, appreciate you taking time to look this over.

.dockerignore
```diff
@@ -1,124 +1 @@
-# Build Folders (you can keep bin if you'd like, to store dlls and pdbs)
```
thgO-O, line 1: The .dockerignore change seems unrelated to this PR, and removing the existing ignore rules may send unnecessary files into the Docker build context. What’s the goal of modifying it here?

monicamuyama: I had a technical issue with my device and had to setup the environment afresh and after doing that and addressing the comments, the CI/CD pipeline was failing with an error of concerning unsupported repository version. However the tests were all passing on my end

rockstardev (changes requested):
Hey @monicamuyama please just rebase the PR and revert the change on .gitignore. Then @NicolasDorier can review and test, @thgO-O tinks this is good to go.

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -287,6 +301,22 @@
...
+                                    <span text-translate="true">Disable</span>
+                                    <vc:icon symbol="warning" />
+                                </button>
+                            }
+                            else
+                            {
+                                <form asp-action="DisablePlugin" asp-route-plugin="@plugin.Identifier">  
+                                     <button type="submit" class="btn btn-outline-warning" text-translate="true">Disable</button>  
+                                 </form>  
+                            }
```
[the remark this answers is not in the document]
rockstardev: @monicamuyama please also address this, and if you have - link commit hash.

monicamuyama:
Done with revert and rebasing

rockstardev:
@monicamuyama please DM me if you are on https://chat.btcpayserver.org - would appreciate talking with you about future PRs.

PR #6943, opened 2025-10-09
feat : add functionality to download language pack

Abhijay007:
This PR aims to link the btcpayserver-translator files to the translations page: 

https://github.com/user-attachments/assets/98645236-7338-4842-ad86-7a9552229659

### DIFF
--- a/BTCPayServer/Controllers/UIServerController.Translations.cs
+++ b/BTCPayServer/Controllers/UIServerController.Translations.cs
@@ -134,29 +134,30 @@ namespace BTCPayServer.Controllers
                 return RedirectToAction(nameof(ListDictionaries));
             }
 
+            string translationsJson;
             try
             {
-                var translationsJson = await FetchLanguagePackFromRepository(language);
-                var translations = Translations.CreateFromJson(translationsJson);
-                
-                try
-                {
-                    var dictionary = await _localizer.CreateDictionary(language, Translations.DefaultLanguage, "Custom");
-                    await _localizer.Save(dictionary, translations);
-                    TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' downloaded successfully", language].Value;
-                }
-                catch (DbException)
-                {
-                    var existingDictionary = await _localizer.GetDictionary(language);
-                    await _localizer.Save(existingDictionary, translations);
-                    TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' updated successfully", language].Value;
-                }
+                translationsJson = await FetchLanguagePackFromRepository(language);
             }
             catch (HttpRequestException ex)
             {
                 TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Failed to download language pack: {0}", ex.Message].Value;
+                return RedirectToAction(nameof(ListDictionaries));
             }
 
+            var translations = Translations.CreateFromJson(translationsJson);
+            var existingDictionary = await _localizer.GetDictionary(language);
+            if (existingDictionary is null)
+            {
+                existingDictionary = await _localizer.CreateDictionary(language, Translations.DefaultLanguage, "Custom");
+                TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' downloaded successfully", language].Value;
+            }
+            else
+            {
+                TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' updated successfully", language].Value;
+            }
+            
+            await _localizer.Save(existingDictionary, translations);
             return RedirectToAction(nameof(ListDictionaries));
         }
 
--- a/BTCPayServer/Services/LocalizerService.cs
+++ b/BTCPayServer/Services/LocalizerService.cs
@@ -177,7 +177,7 @@ namespace BTCPayServer.Services
         {
             await using var ctx = _ContextFactory.CreateContext();
             var db = ctx.Database.GetDbConnection();
-            var r = await db.QueryFirstAsync("SELECT * FROM lang_dictionaries WHERE dict_id=@dict_id", new { dict_id = name });
+            var r = await db.QueryFirstOrDefaultAsync("SELECT * FROM lang_dictionaries WHERE dict_id=@dict_id", new { dict_id = name });
             if (r is null)
                 return null;
             return new Dictionary(r.dict_id, r.fallback, r.source ?? "", JObject.Parse(r.metadata ?? "{}"));


Abhijay007:
Hi @dstrukt, We have some translations for different languages [here](https://github.com/Abhijay007/btcpayserver-translator/tree/main/translations) , @NicolasDorier asked me to find a way to link these files in the BTCPayServer translations so that users can easily discover them. I’m not entirely sure whether we want to give users an option to select those strings directly from the dropdown (just like we have for English) or simply link the files on the page.

For now, I’ve just linked them on the page, but I’d really appreciate your guidance on how we can make the translations you worked on more discoverable to users.

dstrukt:
Sorry this took me a minute @Abhijay007. If we're confident about the base level translations, I think a dropdown to select/display would make the most sense from a usability standpoint, that the user can then edit/update/override specific translations, etc.. for their individual use-case. 

That's certainly a nicer experience than a distinct page.

Unless there's something I'm missing here?

Abhijay007:
> Sorry this took me a minute @Abhijay007. If we're confident about the base level translations, I think a dropdown to select/display would make the most sense from a usability standpoint, that the user can then edit/update/override specific translations, etc.. for their individual use-case.
> 
> That's certainly a nicer experience than a distinct page.
> 
> Unless there's something I'm missing here?

okay, thanks a lot for the review. I will update this PR and share it for further review

rockstardev:
So I spoke with @Abhijay007 today and my undestanding is that @NicolasDorier also spoke with him and you guys have plan for merging this PR.

Translator repository has been already migrated to BTCPay Server organization on: https://github.com/btcpayserver/btcpayserver-translator/tree/main/translations

I'll unassign myself from this task and focus on other stuff.

NicolasDorier:
I discussed previously with @Abhijay007 , and said a better UX would be a drop down which download and install, rather than asking the user to copy paste manually.

Abhijay007:
> I discussed previously with @Abhijay007 , and said a better UX would be a drop down which download and install, rather than asking the user to copy paste manually.

Hi @NicolasDorier, sorry for the late update on this. I was experimenting with something so that the user wouldn’t even need to download the language file and could instead load it directly through the interface. I got a bit busy with the Playwright test work, but here’s a sneak peek of the functionality I’ve built.

test

Let me know if it looks good, I can push the changes asap.

NicolasDorier:
Awesome progress. @Abhijay007 I've made some feedback in DM.
Basically instead of adding it here, make a popup in `Dictionaries` that can be open with a button next to `Create` called `Download language pack`.

Selecting one and confirming would make btcpay download the pack and import in the DB, adding the dictionary to the list.

Kind of excited by this. Lot's of people need translations, but were not using it due to this missing piece.

BTCPayServer/Views/UIServer/EditDictionary.cshtml
```diff
@@ -32,6 +32,10 @@
                 <li text-translate="true">Remove the translation from this dictionary.</li>
                 <li text-translate="true">Set the translation to match the string in the fallback.</li>
             </ul>
+			<p>
+				<span text-translate="true">We have some premade translation strings here for several languages you can just copy paste it from</span>
+				<a href="https://github.com/Abhijay007/btcpayserver-translator/tree/main/translations" target="_blank" text-translate="true">Here</a>.
+			</p>
```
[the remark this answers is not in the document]
Abhijay007: c6ccccc

Abhijay007:
updated UI layout : 
image

cc : @NicolasDorier

BTCPayServer/Views/UIServer/ListDictionaries.cshtml
```diff
@@ -65,3 +70,125 @@
...
+					<span class="btn-text" text-translate="true">Download</span>
+					<span class="spinner-border spinner-border-sm d-none ms-2" role="status" aria-hidden="true"></span>
+				</button>
+			</div>
+		</div>
+	</div>
+</div>
+
+@section PageFootContent {
+	<script>
```
NicolasDorier: Do not download from Javascript, download from the server, it's easier and more maintainable.

Abhijay007: [9404932](https://github.com/btcpayserver/btcpayserver/pull/6943/commits/94049323f455fabf96041949f62c73e1951a3c7c)

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+                var url = $"https://raw.githubusercontent.com/btcpayserver/btcpayserver-translator/main/translations/{fileName}.json";
+                var response = await httpClient.GetAsync(url);
+
+                if (!response.IsSuccessStatusCode)
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Failed to download language pack: {0}", response.StatusCode].Value;
+                    return RedirectToAction(nameof(ListDictionaries));
+                }
+
+                var translationsJson = await response.Content.ReadAsStringAsync();
```
NicolasDorier: Encapsulate that in a `DownloadLanguagePack` method for readability.
Also, you need to using `using var response...` so it doesn't result in memory leak.

Use `EnsureSuccessStatusCode` to throw if there is an error.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+
+                if (string.IsNullOrEmpty(translationsJson) || translationsJson == "{}")
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Translation file is empty"].Value;
+                    return RedirectToAction(nameof(ListDictionaries));
+                }
```
NicolasDorier: Useless check. If there is a json back and no error in the query, this should be a proper json. If not, it's fine to crash, as it shouldn't happen in the first place, this catch will kind of hide the problem.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+
+                if (!Translations.TryCreateFromJson(translationsJson, out var translationsObj))
```
NicolasDorier: Same, useless. Use `CreateFromJson`, let it crash if that's not right json.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+                {
+                    TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Invalid translation format"].Value;
+                    return RedirectToAction(nameof(ListDictionaries));
+                }
+
+                var dictionaryName = language;
+                
+                await _localizer.CreateDictionary(dictionaryName, Translations.DefaultLanguage, "Custom");
+                var dictionary = await _localizer.GetDictionary(dictionaryName);
+                if (dictionary is null)
```
NicolasDorier: useless, it shouldn't be null, we just created it... if it is null, crashing is the right answer.

But more importantly, why do you use `GetDictionary`? the `CreateDictionary` already returns the dictionary to you.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+                await _localizer.Save(dictionary, translationsObj);
+                TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' downloaded successfully", dictionaryName].Value;
+            }
+            catch (DbException)
+            {
+                TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Dictionary '{0}' already exists", language].Value;
+            }
+            catch (HttpRequestException ex)
+            {
+                TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Failed to download language pack: {0}", ex.Message].Value;
```
NicolasDorier, line 144: If failure to download, this should be caught when you call the new `DownloadLanguagePack`, as this is the only place where you do HTTP requests.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+            }
+            catch (Exception ex)
+            {
+                TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Failed to import language pack: {0}", ex.Message].Value;
```
NicolasDorier: Outside of the download failure which can legitimately happen and the duplicate error, all other possible exception should just crash as it would be a bug.

BTCPayServer/Controllers/UIServerController.Translations.cs
```diff
@@ -122,5 +124,70 @@ public async Task<IActionResult> DeleteDictionary(string dictionary)
...
+                    TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Failed to create dictionary"].Value;
+                    return RedirectToAction(nameof(ListDictionaries));
+                }
+
+                await _localizer.Save(dictionary, translationsObj);
+                TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Language pack '{0}' downloaded successfully", dictionaryName].Value;
+            }
+            catch (DbException)
+            {
+                TempData[WellKnownTempData.ErrorMessage] = StringLocalizer["Dictionary '{0}' already exists", language].Value;
```
NicolasDorier: I think you should handle this better.

If a user download a dictionary that is already in the list, you should update the existing one it rather than failing.

NicolasDorier (changes requested):

BTCPayServer/Controllers/UIServerController.Translations.cs
Abhijay007: Okay so we need to create a separate helper method for this?

Abhijay007: [66157b8](https://github.com/btcpayserver/btcpayserver/pull/6943/commits/66157b8552b16dc0f692577a64c49b61c3dcc9ee)

PR #6956, opened 2025-10-15
Lightning topup invoice

kngako:
Lightning node implementations have become reliably stable  when handling zero amount invoices. This should allow BTC to support lighting topup invoices.

These changes were tested on CLN v25.05

### DIFF
--- a/BTCPayServer.Tests/SeleniumTests.cs
+++ b/BTCPayServer.Tests/SeleniumTests.cs
@@ -1320,7 +1320,6 @@ namespace BTCPayServer.Tests
             Assert.Contains("Lightning", s.Driver.FindElement(By.CssSelector(".payment-method.active")).Text);
             Assert.Contains("LNURL", s.Driver.FindElement(By.CssSelector(".payment-method:nth-child(2)")).Text);
 
-            s.Driver.FindElement(By.CssSelector(".payment-method:nth-child(2)")).Click();
             TestUtils.Eventually(async () =>
             {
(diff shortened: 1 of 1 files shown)


kngako:
The SeleniumTests changes seem to have become eventually consistent. Probably something I'm missing.

rockstardev:
Please refrain from opening PRs that are not agreed upon through https://github.com/btcpayserver/btcpayserver/issues or through existing contributors, we don't have time to dedicate to as big refactorings as this ones without first agreeing on plan of action.

kngako:
Ops... sorry about that. I only saw the issue https://github.com/btcpayserver/btcpayserver/issues/6911 after I already worked on this PR.  

Seems @w0aw had also submitted a related PR https://github.com/btcpayserver/btcpayserver/pull/6916 but they were limited by technical ability.

As for the size of the PR, I believe it's a reasonable small change besides the changes I've made in the SeleniumTests.cs which I can happily revert but I'd need some assistance to figure out why I can't get these tests to work correctly.

rockstardev:
@kngako, I will also close that other PR for the reasons mentioned by @Kukks.

For changes that alter default behavior (like what @w0aw attempted), we should implement them either through plugins or use a gate before introducing a configuration option. While a one-liner solution might work for a specific user, it may not be effective for others.

Therefore, let's first discuss this either on https://chat.btcpayserver.org or on https://t.me/btcpayserver. Once we reach an agreement on a plan of action, we can proceed with the implementation.

kngako:
Koo.

PR #6958, opened 2025-10-17
Providing full status for review in POS recent transactions

rockstardev:
I've been reviewing POS Recent Transactions and they are missing exception status string, which is critical for some transactions. How it is currently:

image

How it should be:

image

The one liner in PR fixes this, so person operating POS device has more info right away


rockstardev:
The icon for recent transactions also needs to be center aligned. Before: 
image

After:
image

pavlenex:
cACK. Makes a lot of sense.

PR #6961, opened 2025-10-21
GPG validation check

TChukwuleta:
Include GPG verification on plugin install and update

gpg_btcpayserver_core

Dependent on this: https://github.com/btcpayserver/btcpayserver-plugin-builder/pull/96

Also pointing the documentation link to redirect to public pages.. well this happens on the plugin builder PR

Cc. @rockstardev

### DIFF
--- a/BTCPayServer.Abstractions/Contracts/IBTCPayServerPlugin.cs
+++ b/BTCPayServer.Abstractions/Contracts/IBTCPayServerPlugin.cs
@@ -14,6 +14,7 @@ namespace BTCPayServer.Abstractions.Contracts
         Version Version { get; }
         string Description { get; }
         bool SystemPlugin { get; set; }
+        string Fingerprint { get; set; }
         PluginDependency[] Dependencies { get; }
         void Execute(IApplicationBuilder applicationBuilder, IServiceProvider applicationBuilderApplicationServices);
         void Execute(IServiceCollection applicationBuilder);
--- a/BTCPayServer.Abstractions/Contracts/IBTCPayServerPlugin.cs
+++ b/BTCPayServer.Abstractions/Contracts/IBTCPayServerPlugin.cs
@@ -14,7 +14,6 @@ namespace BTCPayServer.Abstractions.Contracts
         Version Version { get; }
         string Description { get; }
         bool SystemPlugin { get; set; }
-        string Fingerprint { get; set; }
         PluginDependency[] Dependencies { get; }
         void Execute(IApplicationBuilder applicationBuilder, IServiceProvider applicationBuilderApplicationServices);
         void Execute(IServiceCollection applicationBuilder);
--- a/BTCPayServer.Abstractions/Models/BaseBTCPayServerPlugin.cs
+++ b/BTCPayServer.Abstractions/Models/BaseBTCPayServerPlugin.cs
@@ -55,7 +55,6 @@ namespace BTCPayServer.Abstractions.Models
             }
         }
         public bool SystemPlugin { get; set; }
-        public string Fingerprint { get; set; }
         public virtual IBTCPayServerPlugin.PluginDependency[] Dependencies { get; } = Array.Empty<IBTCPayServerPlugin.PluginDependency>();
 
         public virtual void Execute(IApplicationBuilder applicationBuilder,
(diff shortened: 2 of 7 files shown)


BTCPayServer.Abstractions/Contracts/IBTCPayServerPlugin.cs
```diff
@@ -14,6 +14,7 @@ public interface IBTCPayServerPlugin
         Version Version { get; }
         string Description { get; }
         bool SystemPlugin { get; set; }
+        string Fingerprint { get; set; }
```
NicolasDorier: Don't change this interface, this is a breaking change that would probably break all plugins.

BTCPayServer.Abstractions/Models/BaseBTCPayServerPlugin.cs
```diff
@@ -55,6 +55,7 @@ public virtual string Description
             }
         }
         public bool SystemPlugin { get; set; }
+        public string Fingerprint { get; set; }
```
NicolasDorier: Add a property to a base class isn't a breaking change, but adding to an interface is. And since you added this because you changed the interface, this should also be removed.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -155,15 +161,78 @@ public async Task<AvailablePlugin> DownloadRemotePlugin(string pluginIdentifier,
...
+                var fingerprintsDestination = Path.Combine(_dataDirectories.Value.PluginDir, "last-trusted-fingerprints.json");
+                if (!File.Exists(fingerprintsDestination)) return null;
+                try
+                {
+                    var map = JsonConvert.DeserializeObject<Dictionary<string, string>>(File.ReadAllText(fingerprintsDestination));
+                    return map?.GetValueOrDefault(pluginIdentifier);
+                }
+                catch (Exception)
+                {
+                    return null;
```
NicolasDorier: Don't catch all. If there is something unexpected, let it should crash. If there is something that can happen but isn't a happy path, handle it procedurally.

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -155,15 +161,78 @@ public async Task<AvailablePlugin> DownloadRemotePlugin(string pluginIdentifier,
...
+                SavePluginFingerprint(pluginIdentifier, plugin.Fingerprint);
+            }
             return manifest;
         }
 
+        public void SavePluginFingerprint(string pluginIdentifier, string fingerprint)
+        {
+            lock (_fingerprintLock)
+            {
+                var fingerprintsDestination = Path.Combine(_dataDirectories.Value.PluginDir, "last-trusted-fingerprints.json");
```
NicolasDorier: Why this piece of code is repeated copy/pasta 3 times?

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -155,15 +161,78 @@ public async Task<AvailablePlugin> DownloadRemotePlugin(string pluginIdentifier,
...
             await resp2.Content.CopyToAsync(fs);
             await fs.FlushAsync();
+            if (plugin?.IsSigned == true)
+            {
+                SavePluginFingerprint(pluginIdentifier, plugin.Fingerprint);
+            }
             return manifest;
         }
 
+        public void SavePluginFingerprint(string pluginIdentifier, string fingerprint)
```
NicolasDorier: `SavePluginFingerprint/GetPluginFingerprint/RemovePluginFingerprint` should be async, and use async versions of methods (File.WriteAllTextAsync etc....)

BTCPayServer/Plugins/PluginService.cs
```diff
@@ -43,6 +43,8 @@ BTCPayServerEnvironment env
         public IEnumerable<IBTCPayServerPlugin> LoadedPlugins { get; }
         public BTCPayServerEnvironment Env { get; }
 
+        private static readonly object _fingerprintLock = new object();
```
NicolasDorier: You should avoid locks. The reason you need it here is because you put every fingerprint in a single file `last-trusted-fingerprints.json`. If you were saving each fingerprint in their own file (or in the database, in dedicated rows), you would avoid this issue entirely.

But you can also completely avoid this problem if instead the fingerprint were stored inside the `ManifestInfo`.

NicolasDorier (changes requested):

TChukwuleta:
Hey @NicolasDorier  

I have refactored the PR.. I am now saving the fingerprint to the plugin directory ....

Kindly take another look

PR #6981, opened 2025-11-07
Add missing 'Close' translation to checkout locales

cubetech126:
Added the 'Close' key to the English and Brazilian Portuguese checkout locale files.


NicolasDorier:
Why? this isn't used in the checkout page.

NicolasDorier:
Closing. First, this isn't used in the Checkout, so this doesn't change anything.
Second, those translations need to be done in transifex. (see at the top of the file, there is a warning)

PR #6982, opened 2025-11-08
Show payment request title in wallet transaction tooltip

AshwinGajbhiye:
## Description
This PR improves the user experience by displaying the payment request **title** instead of the **ID** in wallet transaction tooltips.

### Before

Screenshot 2025-11-08 at 3 59 47 PM

### After

Screenshot 2025-11-08 at 4 00 44 PM

## Changes
- **TransactionLabelMarkerHostedService.cs**: 
  - Added `PaymentRequestRepository` dependency injection
  - Modified `InvoiceEvent.ReceivedPayment` handler to fetch payment request title from blob
  - Store title in attachment data using `JObject`

- **LabelService.cs**:
  - Updated `PaymentRequest` label tooltip to read title from `tag.Data`
  - Falls back to payment request ID if title is unavailable

## Testing
-  Created new payment request with title "Test Payment Request"
-  Paid the request using fake payment
-  Verified tooltip shows: "Received through a payment request: Test Payment Request"
-  Tested fallback when payment request not found
-  Existing invoice and app labels still work correctly

### DIFF
--- a/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
+++ b/BTCPayServer/Models/WalletViewModels/WalletSendModel.cs
@@ -81,7 +81,5 @@ namespace BTCPayServer.Models.WalletViewModels
             public long Confirmations { get; set; }
             public DateTimeOffset? Timestamp { get; set; }
         }
-
-        public string Comment { get; set; }
     }
 }


Abhijay007:
Hi @AshwinGajbhiye, thanks for the PR. However, it seems that the actual code changes for this PR might not have been pushed. I can only see formatting changes commit, I don’t see the code update that displays the payment request title instead of the ID in the wallet transaction tooltips. Could you please re-check and push the relevant changes? Thanks

BTCPayServer/HostedServices/TransactionLabelMarkerHostedService.cs
```diff
@@ -179,7 +181,17 @@ walletObjectData.Value.Type is WalletObjectData.Types.Tx &&
                     {
                         Attachment.Invoice(invoiceEvent.Invoice.Id)
                     };
-                        labels.AddRange(PaymentRequestRepository.GetPaymentIdsFromInternalTags(invoiceEvent.Invoice).Select(Attachment.PaymentRequest));
+                        
+                        var paymentRequestIds = PaymentRequestRepository.GetPaymentIdsFromInternalTags(invoiceEvent.Invoice);
+                        foreach (var prId in paymentRequestIds)
+                        {
+                            var pr = await _paymentRequestRepository.FindPaymentRequest(prId, null);
```
NicolasDorier: If you have 100 lines, this will make 100 DB roundtrip, which may timeout the page.
You need to find a way to fetch that all in one request.

dstrukt (approved):
tACK - If this passes dev review, works for me

My only nit might be to reduce the copy `Received through a payment request: ` to something more simple or succinct like `Payment request: `, but arguably it's a little less specific, so could go either way!

NicolasDorier:
Rebased, as a change in master removed Title from the blob.

PR #6985, opened 2025-11-11
Fix: When running more than one test in a test run, only the first would load plugins correctly

NicolasDorier:
Reported by @napoly.

## Actual behavior

Two tests were created. When we would run the tests through the runner,
ASP.NET wouldn't find the registered view of the plugin during the second test run.

## Expected behavior

The second test should work fine, ASP.NET should properly find the
views of the plugin when there are more than one test in the same test
run.

## Cause

If we detected that a plugin assembly was already in the AppDomain, then
we were not loading the ApplicationParts of such assembly.
This wasn't the case for the first test run, but would be after.

The reason for initially doing this was that long time ago, we would
test plugins by referencing them from BTCPaySevrer project. But since
this is not how we are doing things anymore, I think it is safe to
remove this "Feature".
This feature was broken anyway since we started loading plugins in their
own assembly context... which isn't possible when directly referencing
plugins project from BTCPayServer. (#6851)


NicolasDorier:
Got confirmation that it solved napoly's issue.

PR #6997, opened 2025-11-14
Feature: Plugin can extend whether an account can login or not

NicolasDorier:
This allow plugins to control if a user can log in or not by implementing `LoginExtension`.

The `LoginExtension.Check` fill out text or html message to send back to the caller.

I plan to use that in monetization. If the user is disabled because the plan doesn't allow it, we should send proper error back, along with a link to proceed to payment.

This PR also allow localization of the error messages during login.


NicolasDorier:
@coderabbitai review

PR #7001, opened 2025-11-14
Validate scopes of service injections in tests

NicolasDorier:
Bug such as #7000 and a5ff655 were introduced because we injected scoped services into singletons. This PR makes sure tests would catch that.

To avoid breaking too much things, I will just enforce it in tests.

This PR also remove `LocalBTCPayClient`. The Factory was a singleton injecting Controllers. It was difficult to maintain, and we were not using it. The purpose of this class was to allow plugins to use greenfield API, but it has never been really used.

I suspect there is also a way to do this without the maintenance burden.


NicolasDorier:
@coderabbitai  review

PR #7011, opened 2025-11-24
Fix clipboard copy for fiat amounts and other numeric values

### LINKED ISSUE #6933: Bug: Unable to copy fiat amount in Invoice Checkout
Clicking on the fiat amount copy `NaN` in the keyboard (Total Fiat and Exchange Rate)

Image

psam21:
Convert asNumber() results to strings using optional chaining and toString() to fix clipboard copy functionality. Previously, numeric values were being copied as 'NaN' because data-clipboard attribute expected strings.

Fixes #6933

### DIFF
--- a/BTCPayServer/wwwroot/checkout/checkout.js
+++ b/BTCPayServer/wwwroot/checkout/checkout.js
@@ -57,7 +57,27 @@ function updateLanguage(lang) {
 }
 
 function asNumber(val) {
-    return val && parseFloat(val.toString().replace(/\s/g, '')); // e.g. sats are formatted with spaces: 1 000 000
+    if (!val) return null;
+    let str = val.toString();
+    
+    // Remove leading non-numeric characters (e.g. currency symbols like $, €, £)
+    str = str.replace(/^[^\d-]+/, '');
+    
+    // Detect decimal separator: the last occurrence of . or , is the decimal separator
+    // Everything before it is thousands separators
+    const lastComma = str.lastIndexOf(',');
+    const lastDot = str.lastIndexOf('.');
+    
+    if (lastComma > lastDot) {
+        // European format: 91.121,60 - comma is decimal, dots/spaces are thousands
+        str = str.replace(/[\s.]/g, '').replace(',', '.');
+    } else {
+        // US/UK format: 91,121.60 - dot is decimal, commas/spaces are thousands
+        str = str.replace(/[\s,]/g, '');
+    }
+    
+    const num = parseFloat(str);
+    return isNaN(num) ? null : num;
 }
 
 Vue.use(VueI18next);


rockstardev (approved):
The issue was caused by parsing in javascript code... if you send $10.00, it wasn't parse it as 10 because of the leading currency symbol.

So the initial approach with `?.toString()` would not work. @dstrukt requested that from `$99.00` value `99.00` (and not `$99.00` string) - https://github.com/btcpayserver/btcpayserver/issues/6933#issuecomment-3362096248

I've changed the code and am merging now to resolve this issue.

rockstardev:
By the way, are you on https://chat.btcpayserver.org @psam21? If you are DM me, want to touch on future PRs.

psam21:
Thanks @rockstardev ; Have DMed you on chat.

NicolasDorier:
This solution feel like duck tape to me. Maybe better than what was there before, but I expect that to break with some other random currency or format... IMHO `asNumber` should have been completely removed, and the numbers should have been passed by the server backend with AND without formatting.

rockstardev:
@NicolasDorier I disagree with your feeling & think this is more than enough robust for function. It definitely fixes the problem, finally. Adding all the extra code on server side would significantly increase complexity, but feel free to implement it however you prefer it.

NicolasDorier:
@rockstardev not covered:

* Format with space at delimiters.
* Format which have . and not comma, or comma and no `,`

NicolasDorier:
@rockstardev , to prove my point.

Try create an invoice of `123456789 JPY`. When you copy your clipboard is `123.456`.
I prefer NaN, as at least the user doesn't have erroneous information.

I think that Sats may have also an issue, as the separator is space. Maybe not, did you tried?

Frankly, rather than trying to parse, maybe the best is to literally copy the formatted value and let user clean it up. This is better than having this type of bugs.

NicolasDorier:
@psam21 I think you should stop trying to vibe code  that. There are hundreds of currency format in the world, parsing formatted data should not be done at all. And if you really insist, then you better write a test that check them all.

I think the answer is to remove `AsNumber` altogether, and just copy the formatted amount to the user. Simple, less code, less bug, good enough, and even in some cases, maybe what the user actually wants.

psam21:
@NicolasDorier sure; let me think through this.

rockstardev:
@psam21, considering what @NicolasDorier said, I don't think we should invest more into code to handle special cases, but rather simplify.

You can pretty much open new PR doing a similar thing you did in https://github.com/btcpayserver/btcpayserver/pull/7011/commits/6226438fbf4c914bb6f074964c2dc0638e81fb9d - but instead of ?.toString, just in those same places drop `asNumber` and also remove that function from javascript. We swich to copying whatever formatted text is displayed instead of trying to parse numbers.

psam21:
@rockstardev  sure, I'm on it.

NicolasDorier:
Reverted on https://github.com/btcpayserver/btcpayserver/commit/afcad151f2b6a72913a12f4e28007bfdee109386

The commit wasn't working for every currency (notably, JPY). It is better to copy NaN, where the user will notice that something is wrong, than copying a wrong amount that can wreck accounting if the user doesn't notice the amount isn't correct.

rockstardev:
@NicolasDorier we'll do follow-up PR where we drop `asNumber`.

rockstardev:
@psam21 couldn't wait any more for you to open PR, so I opened it myself: https://github.com/btcpayserver/btcpayserver/pull/7036

PR #7015, opened 2025-12-01
[Feature] Payment Request Improvements

### LINKED ISSUE #6989: Improve Payment Requests usability - improving UX, categorization, search, reporting
The current Payment Requests page is quite different from other payment sections, like Invoices or Pull Payments. It works fine for a small number of requests, but once you have multiple pages, it becomes difficult to search or review revenue effectively.

After speaking with users, it seems that a minimal but impactful set of improvements would make a big difference. The goals should be to:

1. **Categorize Payment Requests** - this could leverage the existing Label functionality from the Bitcoin wallet.
2. **Improve search and management** - allow quick searches by title or category, and enable easy recategorization when needed.
3. **Generate reports** - be able to generate reports by date, leveraging added categories. For example: show total revenue between 01/01/2025 and 12/31/2025 for category X, broken down by payment method (Bitcoin onchain, Lightning, stablecoins, etc.).

Additionally, small UX tweaks like having and displaying *Created Date* column would make the page far more usable, especially for users managing thousands of payment requests.

Image

thgO-O:
## Changes

- Added Created Date column to the Payment Requests list
- Added Labels column with inline label manager
- Added Title and Amount search filter
- Added Date filters (Last 24h, 3d, 7d, Custom range modal)
- Added Label filter 
- Implemented Clear All
- Added Report with PaymentRequestsReportProvider
- Tests

Resolves #6989 

https://github.com/user-attachments/assets/8dbaefb5-0bc0-4e39-8543-76a0f072b616

https://github.com/user-attachments/assets/8ede51e9-2f3d-4f25-8d6a-e21ef58fd24c

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,8 @@
 
 * Subscriptions: Allow merchants to accept recurring payments from customers. (#6922) @NicolasDorier
 * Greenfield: Subscriptions API (#7022) @NicolasDorier
+* Report: Add payment requests reports (#7015) @thgO-O @rockstardev
+* Add better filtering capabilities to in the payment requests list (#7015) @thgO-O @rockstardev
 * Ambassadors can monetize access to their server. (#6986) @NicolasDorier
 * Ability to install [language packs](https://github.com/btcpayserver/btcpayserver-translator/tree/main/translations) for the backend UI. (#6943) @Abhijay007
 * Email rules


dstrukt:
Nice! 

Some great looking changes - excited to test this one!

NicolasDorier:
You need to fix `Per-item label fetching causes N+1 query problem.` that coderabbit talked about, the page will probably timeout  for people having too many payment requests.

Also you should NOT concatenate strings for making HTML, this is full of script injection vulnerabilities.

The search feature is probably a table scan, and if not, a scan on all payment requests of the store. I suspect this will take a long time to run on anybody person using PaymentRequests. You can't really fix it, but I expect it will be unusable for a subset of users.

BTCPayServer/wwwroot/main/site.js
```diff
@@ -15,131 +15,212 @@ const switchTimeFormat = event => {
...
+                        a.dataset.bsHtml = 'false';
+                        a.dataset.bsToggle = 'tooltip';
+                        a.dataset.bsCustomClass = 'transaction-label-tooltip';
+
+                        const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+                        svg.setAttribute('role', 'img');
+                        svg.classList.add('icon', 'icon-info');
+
+                        const use = document.createElementNS('http://www.w3.org/2000/svg', 'use');
+                        use.setAttributeNS('http://www.w3.org/1999/xlink', 'href', '/img/icon-sprite.svg#info');
```
NicolasDorier: If the server run with `ROOT_PATH`, the icon-sprite path will be wrong.

Instead of this unreadable JS mess, I advice you to follow the example `UIWalletsController.WalletTransactions`, which returns a razor view, and the `loadMoreTransactions` function using it in the view.

BTCPayServer/wwwroot/main/site.js
```diff
@@ -15,131 +15,212 @@ const switchTimeFormat = event => {
...
+                select.items.forEach(item => {
+                    selectElementI.add(new Option(item, item, true, true));
+                });
+            }
+            if (!updateUrl) return;
+            select.lock();
+            try {
+                const response = await fetch(updateUrl, {
+                    method: 'POST',
+                    credentials: 'include',
```
NicolasDorier: I think you should remove this. This was there before, but this shouldn't be needed as we are at same orign.

BTCPayServer/Services/WalletRepository.cs
```diff
@@ -305,6 +305,69 @@ private Dictionary<string, WalletTransactionInfo> GetWalletTransactionsInfoCore(
                 .Select(FormatToLabel).ToArray();
         }
 
+        public async Task<(string Label, string Color)[]> GetWalletLabelsByLinkedType(WalletId walletId, string linkedType)
+        {
+            await using var ctx = _ContextFactory.CreateContext();
```
NicolasDorier, line 310: If that work for you it's OK, but this is a typical case where using raw SQL is actually more readable and easy to write than EF.

BTCPayServer/Services/PaymentRequests/PaymentRequestRepository.cs
```diff
@@ -138,17 +135,58 @@ public async Task<PaymentRequestData[]> GetExpirablePaymentRequests(Cancellation
...
+                queryable = context.PaymentRequests.FromSqlRaw("""
+                                                                   SELECT *
+                                                                   FROM "PaymentRequests"
+                                                                   WHERE
+                                                                       "StoreDataId" = {0}
+                                                                       AND (
+                                                                           "ReferenceId" = {1}
+                                                                           OR "Id" = {1}
+                                                                           OR jsonb_extract_path_text("Blob2", 'title') = {1}
+                                                                           OR CAST("Amount" AS TEXT) = {1}
```
NicolasDorier: This one might make people frustrated. IMHO, you should convert the user input from C# and check for a match. This will avoid issues like the user not finding the `10 USD` item, because he didn't wrote `10.00 USD` instead.

rockstardev (changes requested):
I like the improvements made. There is still some UX polish needed before this is release-ready.

Most obvious items:

* You need to be able to add, remove, and manage labels while creating or editing a request.
* The new Label column has some quirky behavior. For example, when you add a label and then click it, it becomes selected and the UI changes. See video:
  
https://github.com/user-attachments/assets/d4a92a62-ee56-4096-93e0-d697f76b77ca

* The Clear All button now causes alignment issues. See video:

https://github.com/user-attachments/assets/916cdb99-cb91-451b-8d6d-10443db7175c

Honestly, it is rough to see the current state of Payment Requests. Storing Title in a JSON blob when it is a required field is bad engineering. It causes silly issues like search being case sensitive: if the title is "Testing" and you search for "testing", it does not match.

It is understandable how we got here, since Payment Requests were built over time by multiple developers, but we need to raise the quality bar in this area of the codebase. I think we can still improve this for 2.3:

* Add label management to create/edit Payment Requests
* Fix label UI quirks in the grid
* Fix Clear All layout issues
* Add a migration to extract Title from JSON into its own column and improve search
* Improve the "View" Payment Request UI and the "Copy URL" flow

rockstardev:
OK, here is a demo of Payment Requests with all the improvements:

https://github.com/user-attachments/assets/8ede51e9-2f3d-4f25-8d6a-e21ef58fd24c

I did not address the following:

* Fix label UI quirks in the grid
* Fix Clear All layout issues

These appear to be component-level problems. Since @thgO-O migrated these elements from Invoices, it would be better to track them in a separate, dedicated issue. I will sync with him in our meeting on the best way to approach this issue.

@NicolasDorier, for the Title column migration ([10c110a](https://github.com/btcpayserver/btcpayserver/pull/7015/commits/10c110aed7e4db04fe5966a008bf29ba925843ba)), I implemented it to illustrate approach. Feel free to revert the commit if you do not like it. At this point you can take over. Anything you do not like, rip it out. Let us just get this across the finish line in whatever way you prefer.

rockstardev:
@coderabbitai, regarding your comment - you're incorrect:

> BTCPayServer/Controllers/GreenField/GreenfieldPaymentRequestsController.cs (1)
> > `166-177`: **Making Title required in the Greenfield API is a breaking change—confirm it’s intended**
> > `CreateOrUpdatePaymentRequest` now treats an empty `request.Title` as a validation error and persists Title only via `PaymentRequestData.Title` (not in the blob). That’s consistent with the new schema, but it changes the API contract: previously, clients could create payment requests without a title; after this change they’ll start receiving 400s.
> > If backward compatibility is important, consider either:
> > 
> > * Defaulting missing titles server-side (e.g., a generic label or based on ReferenceId) while keeping the field optional, or
> > * Clearly documenting this as a breaking Greenfield change in release notes/migration docs so integrators can adjust.
> > 
> > Also applies to: 221-255

We had Title validation that it is required, see: https://github.com/btcpayserver/btcpayserver/blob/a3df88dfac0edf26d45968db16ad17efefac2dfd/BTCPayServer/Controllers/GreenField/GreenfieldPaymentRequestsController.cs#L175-L176

dstrukt:
> OK, here is a demo of Payment Requests with all the improvements:

Love the additional changes, the usability of this feature / view is looking much better!

BTCPayServer.Data/Data/PaymentRequestData.Migration.cs
```diff
@@ -18,7 +18,7 @@ public partial class PaymentRequestData : MigrationInterceptor.IHasMigration
         public bool TryMigrate()
         {
 #pragma warning disable CS0618 // Type or member is obsolete
-            if (Blob is (null or { Length: 0 }) && Blob2 is not null && Currency is not null)
+            if (Blob is (null or { Length: 0 }) && Blob2 is not null && Currency is not null && Title is not null)
```
NicolasDorier: Remove `Title is not null`. This code must not run again... it's only ran once for 1.x migration.

BTCPayServer.Data/Migrations/20251216120000_pr_title.cs
```diff
@@ -0,0 +1,35 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "Title",
+                table: "PaymentRequests",
+                type: "text",
+                nullable: true);
+
+            // Create case-insensitive index for Title searches
+            migrationBuilder.Sql(@"CREATE INDEX ""IX_PaymentRequests_Title"" ON ""PaymentRequests"" (LOWER(""Title""));");
```
NicolasDorier: this index is not used by any query.

PR #7036, opened 2025-12-18
Fixing clipboard NaN problem on checkout

rockstardev:
Fixes: #6933

As discussed in the previous pull request (https://github.com/btcpayserver/btcpayserver/pull/7011#issuecomment-3644951159), the best way to address this issue for now is by copying the strings for the fiat amount and exchange rate. Here’s a video demonstration of the code in action:

https://github.com/user-attachments/assets/7e5d969f-3b27-43ff-8614-86ece54f52e8

We cannot eliminate the `asNumber` function since it is utilized by other methods in `checkout.js`. Therefore, we’re making minimal changes in this pull request to resolve the issue with NaN.


dstrukt (approved):
tACK - LGTM

Tested in lasted Chrome & FF

PR #7050, opened 2025-12-25
[Feature] Store-scoped labels + linkedType-scoped LabelManager suggestions

thgO-O:
## Changes 

- Added optional linkedType support to LabelManager/LabelsJson so suggestions can be scoped per object type (instead of all wallet labels)
- Fixed Payment Requests reporting grouping when labels are empty by rendering them as “Unlabeled”, preventing repeated group blocks in the table
- Introduced StoreLabels / StoreLabelLinks (store-scoped) with migration from wallet label links for Payment Requests
- Extended LabelManager to support store scope
- Updated Payment Requests flows to read/write store-scoped labels

### DIFF
--- a/BTCPayServer.Data/Migrations/20260114053517_StoreScopedLabels.cs
+++ b/BTCPayServer.Data/Migrations/20260114053517_StoreScopedLabels.cs
@@ -25,14 +25,14 @@ namespace BTCPayServer.Migrations
                 },
                 constraints: table =>
                 {
-                    table.PrimaryKey("PK_StoreLabels", x => new { x.store_id, x.id });
+                    table.PrimaryKey("PK_store_labels", x => new { x.store_id, x.id });
                 });
 
-            migrationBuilder.CreateIndex(
-                name: "IX_StoreLabels_StoreId_Type_Text",
-                table: "store_labels",
-                columns: new[] { "store_id", "type", "text" },
-                unique: true);
+            migrationBuilder.Sql(
+                """
+                CREATE UNIQUE INDEX "IX_store_labels_store_id_type_text_lower"
+                ON store_labels (store_id, type, lower(text));
+                """);
 
 
             migrationBuilder.CreateTable(
@@ -45,9 +45,9 @@ namespace BTCPayServer.Migrations
                 },
                 constraints: table =>
                 {
-                    table.PrimaryKey("PK_StoreLabelLinks", x => new { x.store_id, x.store_label_id, x.object_id });
+                    table.PrimaryKey("PK_store_label_links", x => new { x.store_id, x.store_label_id, x.object_id });
                     table.ForeignKey(
-                        name: "FK_StoreLabelLinks_StoreLabels_StoreId_StoreLabelId",
+                        name: "FK_store_label_links_store_labels_store_id_store_label_id",
                         columns: x => new { x.store_id, x.store_label_id },
                         principalTable: "store_labels",
                         principalColumns: new[] { "store_id", "id" },
@@ -55,11 +55,9 @@ namespace BTCPayServer.Migrations
                 });
 
             migrationBuilder.CreateIndex(
-                name: "IX_StoreLabelLinks_StoreId_ObjectId",
+                name: "IX_store_label_links_store_id_object_id",
                 table: "store_label_links",
                 columns: new[] { "store_id", "object_id" });
-
-            migrationBuilder.Sql(@"CREATE EXTENSION IF NOT EXISTS pgcrypto;");
 
             // Copy Payment Request label objects (label metadata) into StoreLabels
             migrationBuilder.Sql(@"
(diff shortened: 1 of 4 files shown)


rockstardev:
I reviewed the code and tested the PR locally. One behavior that feels a bit unexpected is that if I create a label and then delete it from a request (there are 0 links of specific label to requests) - the label itself gets deleted.

This video shows it in action:
[https://github.com/user-attachments/assets/80f41ccd-e371-4afe-ab95-deb0b65edb87](https://github.com/user-attachments/assets/80f41ccd-e371-4afe-ab95-deb0b65edb87)

For now, I am OK with this behavior so we can merge and move forward.

@NicolasDorier, please share your review as well if you are OK with merging the PR.

NicolasDorier:
@rockstardev the reason why this behave like this is because of `GetStoreLabelsByLinkedType` which check all the links of all PRs, and give you one of each.

I don't really like this, as this is O(N), this call will take longer as the number of PR increases.
Instead, @thgO-O could just use the "StoreLabels" table. But the problem if he does that, is that then there is no way to remove those labels from the list... which mean we need another way to remove those, which would make this PR bigger.

I guess this is something to tackle for another day.

NicolasDorier:
@coderabbitai please review.

thgO-O:
hey @NicolasDorier, I spoke with @rockstardev  about moving the type to StoreLabel in order to isolate labels better. I'm working on this right now and will push the changes shortly

NicolasDorier:
@thgO-O whatever is fine for me. There are drawbacks to both methods.

BTCPayServer.Data/Migrations/20260114053517_StoreScopedLabels.cs
```diff
@@ -0,0 +1,164 @@
...
+                        principalColumns: new[] { "store_id", "id" },
+                        onDelete: ReferentialAction.Cascade);
+                });
+
+            migrationBuilder.CreateIndex(
+                name: "IX_StoreLabelLinks_StoreId_ObjectId",
+                table: "store_label_links",
+                columns: new[] { "store_id", "object_id" });
+
+            migrationBuilder.Sql(@"CREATE EXTENSION IF NOT EXISTS pgcrypto;");
```
NicolasDorier: Remove this, this add a requirement to hosts and will complicate the migration of postgres major versions.

BTCPayServer/Services/Labels/StoreLabelRepository.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        var oldLabelLower = oldLabel.ToLowerInvariant();
+        var newLabelLower = newLabel.ToLowerInvariant();
+        var oldId = await conn.QuerySingleOrDefaultAsync<string?>(
+            """
+            SELECT id
+            FROM store_labels
+            WHERE store_id = @storeId
+              AND type     = @type
+              AND lower(text) = @oldLabelLower
```
NicolasDorier, line 210: You add an index above. But this index will not be used is you do `lower(text) = @oldLabelLower`.

If you want the index to work for case insensitive, you need to modify the index to use the `lower(text)` expression instead of the `text` column.

BTCPayServer/Services/Labels/StoreLabelRepository.cs
```diff
@@ -0,0 +1,362 @@
...
+        if (oldId is null)
+            return false;
+
+        var newId = await conn.QuerySingleOrDefaultAsync<string?>(
+            """
+            SELECT id
+            FROM store_labels
+            WHERE store_id = @storeId
+              AND type     = @type
+              AND lower(text) = @newLabelLower
```
NicolasDorier, line 225: Same as above remark as above.

BTCPayServer/Services/Labels/StoreLabelRepository.cs
```diff
@@ -0,0 +1,362 @@
...
+                SET text = @newLabel
+                WHERE store_id = @storeId
+                  AND id       = @oldId
+                """, new { storeId, oldId, newLabel }, transaction: dbTx);
+
+            await tx.CommitAsync();
+            return updatedLabels > 0;
+        }
+
+        await conn.ExecuteAsync(
```
NicolasDorier, line 245: From here, I don't really understand, can you add a comment what you are doing.

BTCPayServer/Services/Labels/StoreLabelRepository.cs
```diff
@@ -0,0 +1,362 @@
...
+            return result;
+
+        var lowerTexts = texts.Select(t => t.ToLowerInvariant()).ToArray();
+        var existing = (await conn.QueryAsync<(string Id, string Text)>(
+            """
+            SELECT id, text
+            FROM store_labels
+            WHERE store_id = @storeId
+              AND type     = @type
+              AND lower(text) = ANY(@lowerTexts)
```
NicolasDorier: Same problem as I said above.
But on top of this I don't think `ANY` use indices properly, especially on older postgres versions.

What instead, you should `unnest(@lowerTexts)` and join against it.

PR #7066, opened 2026-01-01
PoS: Can disable zero amount invoices

### LINKED ISSUE #7035: [Bug]: Pos allows 0-amount invoices to be created - please give option to prevent this

BTCPay Server v2.2.1+790616d2a 

Docker

### What happened?

From time to time I notice invoices in my store that have 0 amount and show as settled. This is pretty much spam or an attempt to go over my GAP Limit (which I prevented by allowing onchain payment option for the store only starting at a certain amount)

The same was reported on Telegram 2 days ago: https://t.me/btcpayserver/93374

The proposed solution by @rockstardev is probably to specific. As the original poster ran into the issue with the keypad app. I'd like to point out that this also happens with the product list with no products and "custom payments" enabled.

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox on Win 11.

### Additional information

none

NicolasDorier:
Fix #7035 

We discussed about whether such settings should be in the store settings instead.
But we couldn't really come with example of such problem in other context than the PoS.

We can always add this in the store settings if we want later on, in such case, we would replace the checkbox with a select which can be set to `Inherit the store's settings`.

Added a settings in PoS settings:
image

Error message in PoS:
image

WIth Keypad:
image

### DIFF
--- a/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
+++ b/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
@@ -330,6 +330,9 @@ namespace BTCPayServer.Plugins.PointOfSale.Controllers
                           selectedChoices.Any(c => c.PriceType == AppItemPriceType.Topup);
 
             var receiptData = PosReceiptData.Create(isTopup, selectedChoices, jposData, order, summary, settings.Currency, _displayFormatter);
+            if (!isTopup && summary.PriceTaxIncludedWithTips <= 0m && settings.DisableZeroAmountInvoice is true)
+                return Error(StringLocalizer["Zero amount invoices are disabled"].Value);
+
             try
             {
                 var invoice = await _invoiceController.CreateInvoiceCoreRaw(new CreateInvoiceRequest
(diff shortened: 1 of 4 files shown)


kayth21:
"Disable zero amount invoices" option is not available if "Keypad" is selected as "Point of Sale Style" in v2.3.2. One have to switch "Point of Sale Style" to some other option, activate "Disable zero amount invoices" and get back to the original "Point of Sale Style". This works. No more zero-amount invoices can be created, but no error message is shown (fine for me, just sayin').

NicolasDorier:
hey @kayth21 where is `no error message is shown` ? As my two other screenshot show above, the error should show.

> One have to switch "Point of Sale Style" to some other option, activate "Disable zero amount invoices" and get back to the original "Point of Sale Style".

yuck, fixing that now. (https://github.com/btcpayserver/btcpayserver/pull/7071)

kayth21:
@NicolasDorier, in the Keypad Point of Sale. If I click "0", the "Charge" Button gets active, if I click it the "Charge" Button get inactive, but no error message is shown (although "disable zero amount invoices" is active) but mainly, no invoice is created, which is good. 
{D546D737-BC84-4FD0-B189-992E52C22119}

PR #7068, opened 2026-01-02
Feature: Cold wallet transaction support via Greenfield API

Aeroverra:
### ❄️ Feature: Cold Wallet Transaction Support (PSBT) via Greenfield API

This PR introduces the ability to perform cold wallet transactions (PSBT workflows) through the Greenfield API. Previously, the API primarily supported hot wallet transactions where the server held the private keys.

This update allows users to:

* Generate unsigned PSBTs.
* Broadcast externally signed transactions.

### Motivation

The primary goal is to enable more compliance options and higher security in automated environments. This allows setups utilizing multisig, external signing devices (hardware wallets), and air-gapped machines to use the Greenfield API for transaction construction and broadcasting without exposing private keys to the server.

**Key Benefits:**

* **Regulatory Compliance:** Allows for setups that meet stricter regulatory requirements.
* **Enhanced Security:** Keeps private keys offline while using the API.
* **Secure Automation:** Enables fully automated workflows where signing is cryptographically separated from transaction creation, utilizing an isolated HSM or Policy Engine.

### Context

I believe this to be related to the following discussion, which contains a more in-depth explanation:
**[Discussion #6262: Feature Request details](https://github.com/btcpayserver/btcpayserver/discussions/6262)**

### DIFF
--- a/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
+++ b/BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
@@ -567,8 +567,7 @@ namespace BTCPayServer.Controllers.Greenfield
             {
                 return Ok(new CreateOnChainTransactionResponse
                 {
-                    PSBT = psbt.PSBT.ToBase64(),
-                    Transaction = null
+                    PSBT = psbt.PSBT.ToBase64()
                 });
             }
 
--- a/BTCPayServer.Client/Models/CreateOnChainTransactionResponse.cs
+++ b/BTCPayServer.Client/Models/CreateOnChainTransactionResponse.cs
@@ -4,6 +4,5 @@ namespace BTCPayServer.Client.Models
     public class CreateOnChainTransactionResponse
     {
         public string PSBT { get; set; } = string.Empty;
-        public string? Transaction { get; set; }
     }
 }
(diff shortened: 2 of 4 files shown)


Aeroverra:
The docstring check I don't quite understand. Is this asking me to add summary comments? I do for my own projects usually but I was following the convention that I saw and I didn't see any in the files I touched. So if I need to add them let me know and I will.

NicolasDorier:
@Aeroverra ignore the `Docstring Coverage`. Thanks for this PR I will review it.

BTCPayServer/Controllers/GreenField/GreenfieldStoreOnChainWalletsController.cs
```diff
@@ -556,6 +563,15 @@ public async Task<IActionResult> CreateOnChainTransaction(string storeId, string
 
             derivationScheme.RebaseKeyPaths(psbt.PSBT);
 
+            if (!request.SignWithSeed)
+            {
+                return Ok(new CreateOnChainTransactionResponse
+                {
+                    PSBT = psbt.PSBT.ToBase64(),
+                    Transaction = null
```
NicolasDorier: `CreateOnChainTransactionResponse.Transaction` seems never used?
I am not sure it makes sense to expose it.

NicolasDorier (changes requested):

NicolasDorier:
Great job. Just minor issue on unused property I think...

Aeroverra:
@NicolasDorier thanks for reviewing. I think it was left behind from my research phase. My bad.

I have pushed an update but I don't think these test failures are due to me but if thats not the case please let me know.

PR #7073, opened 2026-01-03
Include UI extension point for pos

TChukwuleta:
To aid improvement to Voucher plugin.. 

We are requiring UI extension to be added to the PoS keypad view...

PR for reference:
https://github.com/rockstardev/BTCPayServerPlugins.RockstarDev/pull/108

Cc. @rockstardev


rockstardev (approved):
We need this exension point for Vouchers, tACK.

PR #6897, opened 2025-08-20
Refactor : Added missing translations strings for various files 

Abhijay007:
**Description:**

This PR adds missing translation strings by including `text-translate="true"` attributes for previously untranslated static text across the BTCPayServer user interface. This ensures that all hardcoded English text can now be properly localized.

Dynamic content and payment-related files (e.g., `checkout`, etc.) were intentionally ignored for now, following the comments in the last PR (ref: https://github.com/btcpayserver/btcpayserver/pull/6845 ).

**Areas covered in this PR:**

- Wallet Management
- Store Management
- Server Administration
- User Account
- Shared Components
- Invoice and Payment
- Other miscellaneous files

PR #7113, opened 2026-01-17
UI: clarify LND seed recovery expectations

### LINKED ISSUE #5612: [Bug]: UI clarify information about restoring LND backup.

BTCPay v1.12.3

docker

### What happened?

The text around the link to LND revovery procedure is unclear. It should mention that the recovery won't work with LND in BTCPay as the `lncli create` command won't work as it only works on first startup and is done automatically on BTCPay deployment.

image

Maybe link to docs and there explain it better and also mention blixt as a way to close channels and recover funds as SCB won't allow restore the channels but only to close them anyway. 

(not tested myself due the lack of LND enabled on BTCPay) should be on the LND backup page in the BTCPay UI

### Relevant log output

_No response_

### What browser do you use?

_No response_

### Additional information

_No response_

esixce:
Fixes #5612.

Clarifies the LND Seed Backup page copy to set correct expectations for recovery steps under some deployments (e.g. BTCPay Server Docker), where LND may be initialized/unlocked automatically and recovery may require an interactive shell (TTY) and temporarily disabling automatic init/unlock.

Changes:
- Add a short clarification paragraph on the LND Seed Backup page.
- Add the new string to Translations.Default.cs for localization support.

How to verify:
- Server → Services → LND Seed Backup: confirm the new paragraph appears above the link to LND's recovery documentation.

Note: PR #6897 also touches this view for translation-related refactors; happy to rebase/adjust if it lands first.

### DIFF
--- a/BTCPayServer/Services/Translations.Default.cs
+++ b/BTCPayServer/Services/Translations.Default.cs
@@ -1639,6 +1639,7 @@ namespace BTCPayServer.Services
   "The label has been successfully renamed.": "",
   "The Lightning node could not be registered.": "",
   "The LND seed backup is useful to recover on-chain funds of your LND wallet in case of a corruption of your server.": "",
+  "In some deployments (for example BTCPay Server Docker), LND may be initialized/unlocked automatically and recovery may require running the steps with an interactive shell (TTY) and temporarily disabling any automatic init/unlock behavior.": "",
   "The name of the field in the invoice's metadata.": "",
   "The name should be maximum 50 characters.": "",
   "The node is offline": "",
--- a/BTCPayServer/Views/UIServer/LndSeedBackup.cshtml
+++ b/BTCPayServer/Views/UIServer/LndSeedBackup.cshtml
@@ -1,3 +1,5 @@
+@using BTCPayServer.Configuration
+@inject BTCPayServerOptions BtcPayServerOptions
 @model LndSeedBackupViewModel
 @{
     ViewData.SetLayoutModel(new LayoutModel(nameof(ServerNavPages.Services), StringLocalizer["LND Seed Backup"])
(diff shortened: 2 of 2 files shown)


Kukks (changes requested):
* We can detect if we are in the BTCPay docker environment through `BTCPayServerOptions.DockerDeployment`, so that we can show this message conditionally (without the brackets comment)
* We should add docs in docs.btcpayserver.org and link to it cc @pavlenex  wdyt?

pavlenex:
@Kukks I'm not following what needs to be added to docs.btcpayserver.org, can you elaborate?

Kukks:
> @Kukks I'm not following what needs to be added to docs.btcpayserver.org, can you elaborate?

Currently we have https://docs.btcpayserver.org/FAQ/LightningNetwork/#where-can-i-find-recovery-seed-backup-for-my-lightning-network-wallet-in-btcpay-server but no docs on how to actually use that recovery within btcpay docker to restore and recover

ndeet:
Afair the issue was that our scripts automatically open the wallet on startup so you can't really do a restore with the file in some place. There were thoughts about an env variable so that it does not open the wallet on startup, so the backup can get mounted and imported manually. Not sure if that ever was done.

esixce:
Understood. In progress.

esixce:
Hello @Kukks — I can implement the conditional Docker detection you asked for.

The only complete documentation that exists is [@warioishere's German guide](https://yourdevice.ch/der-backup-restore-leitfaden-fuer-lighning-fullnodes/#LND_Wallet_von_BTCPay_Docker_wiederherstellen)

**"jetzt geht's um die Wurst"**

In researching what the recovery instructions should say, I realized the only step that must be terminal is lncli create (TTY requirement). Everything else — disabling auto-init, deleting wallet data, restarting services — could be UI-driven.

Would it be worth considering handling prep/cleanup via UI, leaving only seed entry to SSH?

On a related note `backup-restore.md` promises *"we will provide such a script with a future update"*. It was never built. Users need:
   - SCB download button on seed backup page
   - Prompt when channels change ("Download updated backup")
   - Opt-in Auto-backup script

Would a separate pr on this matter be useful?

ABC-8888:
text-translate="true" should not wrap elements that contain child HTML tags like ``` <a > ```. The translation system replaces the entire innerHTML with a plain text string, which causes the link to render as raw HTML instead of an anchor element.
Screenshot 2026-04-01 at 14 14 54
Also, external links in this codebase consistently use target="_blank" alongside rel="noreferrer noopener".

Issue #6955, opened 2025-10-14
Add min / max lightning invoice expiry requirements


jackstar12:
Some lightning implementations (e.g. boltz) have restrictions around the invoice expiry they can produce. The payment method interface should include a `GetExpiryLimits` or similar function to avoid misconfiguration by the user.

Shamsudeen12:
@pavlenex Can I take this up?

pavlenex:
Sure, I'll assign it to you when there's a PR opened.

Shamsudeen12:
@pavlenex I've opened a PR

Shamsudeen12:
@rockstardev Regarding your [comment](https://github.com/btcpayserver/btcpayserver/pull/7114#issuecomment-3786666553 ) on the initial PR.

Since we're closing that PR, can we discuss and understand the requirements clearly on here.

If the custom node itself the user has set has a different minimum expiry time. Do we just prevent invoices from being able to get created and show the user why, since the store setting set on BTCPay is what should be followed?

rockstardev:
After reviewing that PR, the comment from @Kukks at https://github.com/btcpayserver/btcpayserver/pull/7114#pullrequestreview-3679726343, and this issue - and thinking about the request from the perspective of the overall BTCPay Server architecture - my conclusion is that we should close this issue as well.

I do not see value in adding support for payment methods to override invoice expiry requirements. The existing limitation is good: if a merchant has set expiry to 15 minutes, plugins like Boltz can read that value, subscribe to invoice expiry, and cancel or reject any offers they created once the BTCPay Server invoice has expired. In race-condition scenarios, we already have the paidLate event, and a plugin can still handle the transition from expired to paidLate appropriately.

Thank you @Shamsudeen12 for taking the time to look into this. I would recommend that before taking on future PRs, you first coordinate with other contributors in https://chat.btcpayserver.org or https://t.me/btcpayserver so we can discuss, align on an approach, and point you to an issue that is a good fit.

jackstar12:
The general issue, which I intended to describe, is that a merchant shouldn't be able to misconfigure invoice expiries only to find out later that invoice generation stopped working. We've been seeing logs of such requests failing and thought it would be nice to improve the UX here.

Shamsudeen12:
Alright. Thanks @rockstardev 
I'll close the PR

rockstardev:
[...]

PR #7114, opened 2026-01-18
Impl: Get Invoice Expiry Limits

Shamsudeen12:
# Add Invoice Expiry Limits Support for Lightning Implementations - #6955 

## Summary

Some lightning implementations (e.g., Boltz submarine swaps) have restrictions around the invoice expiry they can produce. This PR adds a `GetExpiryLimits` function to the payment method interface to allow lightning implementations to specify their minimum and maximum expiry constraints, preventing misconfiguration by users.

## Problem

When a user configures a store with a 15-minute invoice expiry but connects to a lightning implementation like Boltz (which requires minimum 12-hour expiry), invoice creation would fail or behave unexpectedly. There was no way for the system to:
1. Query the lightning implementation for its expiry constraints
2. Validate or adjust the configured expiry to fit within allowed limits
3. Inform the user when adjustments were made

### New `ExpiryLimits` Record

Added a new record type to represent expiry constraints:

```csharp
public record ExpiryLimits(TimeSpan? MinExpiry, TimeSpan? MaxExpiry)
{
    public TimeSpan Clamp(TimeSpan expiry);
    public bool IsWithinLimits(TimeSpan expiry);
}
```

### Extended Interfaces

1. **`IPaymentMethodHandler`** - Added `GetExpiryLimits()` method with default null implementation
2. **`IExtendedLightningClient`** - Added `GetExpiryLimits()` method for lightning clients to report their constraints

### Automatic Expiry Adjustment

When creating a lightning invoice:
1. Query the lightning implementation for expiry limits
2. If limits exist and the configured expiry is outside the range, clamp it
3. Update both the BOLT11 invoice expiry AND the BTCPay invoice expiration
4. Log a warning message for visibility

Screenshot 2026-01-17 at 16 19 49

Screenshot 2026-01-17 at 16 20 04

### For Lightning Implementation Developers

To add expiry limits to a custom lightning client, implement `IExtendedLightningClient`:

```csharp
public class BoltzLightningClient : IExtendedLightningClient
{
    public Task<ExpiryLimits?> GetExpiryLimits(CancellationToken cancellationToken = default)
    {
        // Boltz requires minimum 12 hours expiry
        return Task.FromResult<ExpiryLimits?>(new ExpiryLimits(
            MinExpiry: TimeSpan.FromHours(12),
            MaxExpiry: null
        ));
    }
    
    // ... other implementations
}
```

### User Experience

When an invoice is created with an expiry outside the lightning implementation's limits:

1. The expiry is automatically clamped to fit within limits
2. A warning appears in the invoice logs:
   > "Invoice expiry adjusted from 15 to 60 minutes due to lightning implementation limits (min: 60, max: 1440). BTCPay invoice expiration also extended."
3. Both the BOLT11 and BTCPay invoice expirations are updated

### Unit Tests Added

- `ExpiryLimits_Clamp_RespectsMinimum`
- `ExpiryLimits_Clamp_RespectsMaximum`
- `ExpiryLimits_Clamp_RespectsMinAndMax`
- `ExpiryLimits_Clamp_PassesThroughWhenWithinLimits`
- `ExpiryLimits_IsWithinLimits_ReturnsFalseWhenBelowMin`
- `ExpiryLimits_IsWithinLimits_ReturnsFalseWhenAboveMax`
- `ExpiryLimits_NullLimits_AllowsAnyExpiry`
- `ExpiryLimits_EdgeCases`

### Manual Testing

1. Configure any lightning node
2. Create an invoice with default 15-minute expiry
3. (With test limits enabled) Verify invoice logs show expiry adjustment
4. Verify BTCPay invoice expiration is extended to match

## Breaking Changes

None. The new interface methods have default implementations that return `null` (no limits), maintaining backward compatibility with existing lightning implementations.

## Related Issues

Fixes: #6955 

## Checklist

- [x] Added `ExpiryLimits` record with `Clamp()` and `IsWithinLimits()` methods
- [x] Added `GetExpiryLimits()` to `IPaymentMethodHandler` interface
- [x] Added `GetExpiryLimits()` to `IExtendedLightningClient` interface  
- [x] Implemented expiry clamping in `LightningLikePaymentHandler.ConfigurePrompt()`
- [x] Implemented expiry clamping in `UILNURLController`
- [x] Extended BTCPay invoice expiration when BOLT11 expiry is clamped
- [x] Added unit tests for `ExpiryLimits`
- [x] No breaking changes to existing implementations

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -782,13 +782,19 @@ namespace BTCPayServer
                         var expiry = i.ExpirationTime.ToUniversalTime() - DateTimeOffset.UtcNow;
 
                         // Check if the lightning implementation has expiry limits and apply them
+                        // Use a short timeout linked to the request cancellation token
                         var lightningHandler = _handlers.GetLightningHandler(network);
                         try
                         {
-                            var expiryLimits = await lightningHandler.GetExpiryLimits(lnConfig);
+                            using var limitsCts = CancellationTokenSource.CreateLinkedTokenSource(HttpContext.RequestAborted);
+                            limitsCts.CancelAfter(LightningLikePaymentHandler.LightningTimeout);
+                            var expiryLimits = await lightningHandler.GetExpiryLimits(lnConfig, limitsCts.Token);
                             if (expiryLimits is not null)
                             {
-                                expiry = expiryLimits.Clamp(expiry);
+                                // Apply limits without extending beyond the invoice window
+                                // Take the minimum of the original expiry and the clamped value
+                                var clampedExpiry = expiryLimits.Clamp(expiry);
+                                expiry = expiry < clampedExpiry ? expiry : clampedExpiry;
                             }
                         }
                         catch
--- a/BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs
+++ b/BTCPayServer/Payments/LNURLPay/LNURLPayPaymentHandler.cs
@@ -115,6 +115,11 @@ namespace BTCPayServer.Payments.Lightning
         /// Returns the expiry limits from the underlying lightning node.
         /// LNURL's limits are determined by the lightning node it uses.
         /// </summary>
+        /// <remarks>
+        /// This method requires the LightningPaymentMethodConfig, not the LNURLPaymentMethodConfig.
+        /// If you have store access, resolve the lightning config first using:
+        /// store.GetPaymentMethodConfigs()[PaymentTypes.LN.GetPaymentMethodId(cryptoCode)]
+        /// </remarks>
         public async Task<ExpiryLimits?> GetExpiryLimits(object config, CancellationToken cancellationToken = default)
         {
             // LNURL uses the underlying lightning node, so we delegate to the lightning handler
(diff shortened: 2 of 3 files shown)


Kukks (changes requested):
The interface changes will make every plugin implementing them crash.


This PR does not help " avoid misconfiguration by the user" burt rather mutates the invoice's   expiry time after creation. If multiple payment methods have their own expiry limits, they end up competing on who mutates last. 

I'm unsure if it is persisted in the db after they are edited in `BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs`. I assume so since it's during the initial stage anyway.


I'm wondering if a better approach would be handled directly inside the ILightningClient implementation where if btcpay invoice expiry > LN bolt invoice expiry, we automate generate a brand new invoice for the rest of the time remainder. This however, only handles if the expiry max is shorter than the invoice expiry.

I'm assuming the bigger issue is the opposite scenario @jackstar12 ?

In the other direction, we can call CancelInvoice on ILightningClient when a btcpay invoice is no longer active. But it does require ln clients to implemenet that option which has been spotty so far

Shamsudeen12:
@Kukks

For the interface changes, I'm not sure I understand your concern about breaking changes, the method GetExpiryLimits returns a null value which would make the implementation default to the stores value, so that shouldn't break any existing process.

I understand the concern regarding mutation race condition, maybe we could query all limits first and then clamp the expiry time once, that way each handler won't handle mutation independently. I'm saying this because:

If the user has a custom lightning implementation that has a lower bound limit of 30 minutes and they have it configured on the store as 15 minutes. Shouldn't we automatically extend the time for the invoice to match their custom node requirements as I assume that should take precedence in the hierarchy – since using that custom node is intentional

rockstardev:
Honestly, the original issue is severely underdefined.

It doesn't matter that Boltz invoices expire in 12 hours. If the store owner has defined that for them, invoices expire in 15 minutes that should be respected. It could be they have set even less (5 minutes) because they have exchange rate restrictions. Doing these crazy extensions to our codebase to create lightning invoices that expire in 12 hours regardless of settings is absolutely orthogonal to the original intent of BTCPay Server invoices.

And the fact that work started on this issue without even fully understanding the problem or defining it is silly. We're now wasting time reviewing completely suboptimal code, that's would further complicate our codebase and introduce unnecessary complexity... all while not supporting and focusing on much more pressing issues & PRs.

This PR should be closed and sent back to the drawing board.

PR #7133, opened 2026-01-23
feat(wallets): add transaction search and date filters.

### LINKED ISSUE #7091: Add support for search, specifically for TX id inside the wallet view
As mentioned in [issue #3954](https://github.com/btcpayserver/btcpayserver/issues/3954), the wallet transactions view should be standardized line invoice or payment requests views and provide a search functionality, allowing users to immediately find a transaction based on criteria such as the transaction ID.

- Implement a search bar  in the wallet transactions view so users can filter or rapidly locate transactions using their transaction ID.

### LINKED ISSUE #7092: Standardize Wallet View: Add Date Filter for Transactions
**Context:**
As mentioned in [Issue #3954](https://github.com/btcpayserver/btcpayserver/issues/3954), there is a need to standardize the Wallet view so that it offers feature parity with other views in BTCPay Server.

Image

**Request:**
Enhance the Wallet transaction view to allow users to filter transactions by date in the same way as the Payment Requests and Invoices views. This would help users more easily locate relevant transactions and improve consistency across different UI components.

Image

**Suggested Acceptance Criteria:**
- Add date filtering controls to the wallet transaction view, following the existing UI/UX patterns used for payment requests or invoices.
- Ensure date filtering works as expected in combination with other filters ().
- Reuse UI components or logic from other views to maintain consistency.

Additional feedback or implementation notes can be found in #3954.

Sup3rlativ3:
## Summary
Adds invoice‑style search and date filtering to the wallet transactions view, so users can quickly locate transactions by TXID or time range. The UI mirrors the invoice/payment request patterns and filters apply consistently to infinite scroll and exports.

## Problem
Per #7091 and #7092 (and #3954), wallet transactions lacked:
- A search bar for transaction IDs (and related metadata)
- Date filtering controls consistent with invoices/payment requests
- Combined filter support when paging or exporting

This made it harder to locate specific transactions and created UX inconsistency across views.

### Search
- Added search input to wallet transactions
- Supports TXID and metadata (labels/comments/attachments)
- Added filter syntax for labels and direction:
  - `nolabel:true`, `label:<text>`, `direction:in|out`
- No‑label filtering uses `nolabel:true` (no special casing of label:none)

### Date filter parity
- Added “Last 24 hours / 3 days / 7 days + Custom Range” dropdown
- Custom range modal matches invoice/payment request UX
- Uses same `SearchString` filter keys: `startdate`, `enddate`

### Consistency
- Filters apply to infinite scroll and exports
- UI and behavior align with existing invoice/payment request patterns

### Testing
- Integration coverage added within existing wallet flow to assert:
  - TXID search
  - label/direction filters
  - start/end date filters

image

### Related Issues
Fixes: #7091, #7092  
See also: #3954

### Checklist
- [x] Search bar for wallet transactions (TXID + metadata)
- [x] Date filter controls matching invoices/payment requests
- [x] Filters combine correctly and persist through pagination/export
- [x] Integration assertions for search + date filters
- [x] New UI string marked for translation

### Screenshots
**No filtering**
image

**Filter for date**
image

**Filter for label "Label1"**
image

**Filter for no labels**
image

**Filter for label "label1" in the last 24 hours**
image

**Filter for free text "label"**
image

** Filter for transaction ID**
image


dstrukt (approved):
I can't speak to code quality, but functionally this works, and covers all of our bases. Always good to have another pair of eyes on this for edge cases, etc.. though.

The only functional part that I'm not a huge fan of is the generated/displayed filter string, and this is my own subjective view, but I do see the value of it.

@pavlenex What are your thoughts on the Filters string generated? Should it be so present/visible? 

I have minor adjustments and suggestions for the UI, but I'm not sure if we want to try to cover these in the PR, or in a future follow up (ignore the TX rows and everything out of scope of this PR)

It might also arguably make dropping that filter string between the Search/input and the filter buttons.

Image

pavlenex:
> The only functional part that I'm not a huge fan of is the generated/displayed filter string, and this is my own subjective view, but I do see the value of it.

I agree this isn't pretty and not following design patterns in the Invoice view and remove that text string, what user needs is already available from the filter dropdown. 

I am also curious if we can make all these filters in line

Screenshot 2026-02-25 at 11 43 40


Screenshot 2026-02-25 at 11 26 39

My bigger concerns with this PR is that by looking at description it should add IN/OUT filter, however by looking at screenshots and then testing locally,  it doesn't seem to have a filter for that? Is there a reason why this was left out @Sup3rlativ3?


https://github.com/user-attachments/assets/3abfe12c-c57c-4101-b22b-88e555aa5e87

Sup3rlativ3:
Apologies, I have been working 60+ hour weeks recently and just haven’t had the time to check GH or personal projects.

> I agree this isn't pretty and not following design patterns in the Invoice view and remove that text string, what user needs is already available from the filter dropdown.

That is fair feedback on the presentation.

Just to clarify though: the current UI only exposes the date range and label filters directly. The text was added because there are also additional advanced filters available via search syntax, such as `direction:in`, `direction:out`, `nolabel:true`, and custom `startdate` / `enddate` values, and those are not otherwise discoverable from the current dropdowns.

So I agree the helper text is not the prettiest solution, but it is currently the only place that tells the user those advanced filters exist. If the concern is mainly UI consistency, I’m happy to rework that and move the help into something less noisy like a tooltip / popover, or alternatively add proper inline controls so the text is no longer needed.

> I am also curious if we can make all these filters in line

Yes, I think that would be the cleaner approach.

Right now, the page is a bit mixed: date is inline, label is in its own dropdown, and the more advanced filters are only available through search syntax. I will rework that so the filters are presented inline in a way that is more consistent with the existing Invoice / Payment Request views.

> My bigger concerns with this PR is that by looking at description it should add IN/OUT filter, however by looking at screenshots and then testing locally, it doesn't seem to have a filter for that? Is there a reason why this was left out @Sup3rlativ3?

The IN / OUT filter is implemented, but at the moment it is only available through the advanced search syntax, for example `direction:in` or `direction:out`.

I just tested that on this branch and it is working for me, so I think the issue here is really discoverability rather than the filter itself not existing. That is also why I added the helper text in the first place.

I’ll make another commit to address the UI issues raised here. I’ll rework the filtering so it uses inline buttons/controls in the same style as the date range and the Invoice screen, which should make the available filters clearer and remove the need for the helper text.

@pavlenex @dstrukt

dstrukt:
Excited to see the changes, and excellent work so far! @Sup3rlativ3

NicolasDorier:
@Sup3rlativ3 can you fix the conflicts?

@pavlenex you can take another look?

dstrukt:
@NicolasDorier I don't believe any of the suggested changes have been made yet.

cc @pavlenex

NicolasDorier:
Seems to have an issue caught by the tests @Sup3rlativ3

Sup3rlativ3:
@NicolasDorier That looks like an issue with ubuntu not the code?

> 4.831 After this operation, 172 MB of additional disk space will be used.
> 4.831 Err:1 http://security.ubuntu.com/ubuntu noble-security/main amd64 libudev1 amd64 255.4-1ubuntu8.14
> 4.831   404  Not Found [IP: 91.189.91.81 80]
> 4.845 Err:2 http://security.ubuntu.com/ubuntu noble-security/main amd64 libsystemd-shared amd64 255.4-1ubuntu8.14
> 4.845   404  Not Found [IP: 91.189.91.81 80]
> 4.846 Err:3 http://security.ubuntu.com/ubuntu noble-security/main amd64 libsystemd0 amd64 255.4-1ubuntu8.14
> 4.846   404  Not Found [IP: 91.189.91.81 80]
> 4.846 Err:4 http://security.ubuntu.com/ubuntu noble-security/main amd64 systemd-dev all 255.4-1ubuntu8.14
> 4.846   404  Not Found [IP: 91.189.91.81 80]
> 4.846 Err:5 http://security.ubuntu.com/ubuntu noble-security/main amd64 systemd amd64 255.4-1ubuntu8.14
> 4.846   404  Not Found [IP: 91.189.91.81 80]
> 4.846 Err:6 http://security.ubuntu.com/ubuntu noble-security/main amd64 systemd-sysv amd64 255.4-1ubuntu8.14
> 4.846   404  Not Found [IP: 91.189.91.81 80]
> 4.846 Err:7 http://security.ubuntu.com/ubuntu noble-security/main amd64 udev amd64 255.4-1ubuntu8.14
> 4.847   404  Not Found [IP: 91.189.91.81 80]
> 4.847 Err:8 http://security.ubuntu.com/ubuntu noble-security/main amd64 libpam-systemd amd64 255.4-1ubuntu8.14
> 4.847   404  Not Found [IP: 91.189.91.81 80]failed to solve: process "/bin/sh -c apt-get update && apt-get install -y --no-install-recommends chromium-driver     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
> Exited with code exit status 17

Sup3rlativ3:
The test failure (other than the ubuntu mirror issues) turned out not to be caused by this wallet filtering change itself, but by a stale Playwright selector. A Playwright test was added (#7210) for label filtering after this PR was created, and it was still trying to open the old `#Filter button.dropdown-toggle` control. That markup had already been removed when the wallet transaction filters were refactored to use separate dropdown toggles for direction, label, and date as per the request from @pavlenex and @dstrukt above, so the test was timing out waiting for an element that no longer exists. I've updated the test to use the current wallet transactions UI, specifically `#LabelOptionsToggle`, `#LabelDropdownMenu`, and `#LabelSearch`. The test still verifies the same behavior it was meant to cover, that the label dropdown opens, the label search narrows the list correctly, the expected label counts are shown, and selecting a label applies the filter to the transaction list.

I have found some other issues with how things are rendered when resizing and also how the buttons look on mobile (the export and reporting are right aligned while the filtering ones are left aligned). I will look into those today and see what I can do about those but wanted to give everyone here an update first.

Sup3rlativ3:
The resize issues should now also be resolved.

If someone wanted to review that would be appreciated. Perhaps @dstrukt or @pavlenex ?

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -1983,7 +2150,7 @@ public async Task<IActionResult> EditWalletLabel(
             return null;
         }
 
-        private string? GetUserId() => User.GetIdOrNull();
+        private string? GetUserId() => _userManager.GetUserId(User)!;
```
NicolasDorier: Why did you changed this? I've done this refactoring recently so we don't need UserManager just to get the Id of a user.

NicolasDorier (changes requested):

BTCPayServer/Controllers/UIWalletsController.cs
Sup3rlativ3: Apologies, this was an oversight during the merge conflict. I'll revert it back to your refactor.

Sup3rlativ3:
For those following along, the failures were a transient error with the exhange rate provider

> Failed CanGetRateCryptoCurrenciesByDefault [647 ms]
>   Error Message:
>    Impossible to get the rate ERR_RATE_UNAVAILABLE(bitfinex, UST_BTC) * (70999.5, 70999.6)
>   Stack Trace:
>      at BTCPayServer.Tests.ThirdPartyTests.CanGetRateCryptoCurrenciesByDefault() in /source/BTCPayServer.Tests/ThirdPartyTests.cs:line 445
> --- End of stack trace from previous location ---

@coderabbit review

BTCPayServer/Controllers/UIWalletsController.cs
```diff
@@ -49,6 +50,7 @@ namespace BTCPayServer.Controllers
 {
     [Route("wallets")]
     [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
+    [AutoValidateAntiforgeryToken]
```
NicolasDorier: This isn't needed. This is enforced globally.

NicolasDorier (changes requested):
@Sup3rlativ3 your PR is adding a lots of feature but barely testing anything.

Also, I noticed a bug. When I put a time filter, the timezoneOffset passed to the search is 0, while I am living in Japan.

Also, the button UI seems off.

image

On master

image

BTCPayServer/Controllers/UIWalletsController.cs
Sup3rlativ3: Fixed

Sup3rlativ3:
@coderabbitai review

ABC-8888:
1. When custom date is used to make a search, the end date is not included in search. For example, if there is a transaction on 2026-03-31, but if I choose dates for example, start date = 2026-03-24, end date = 2026-03-31, no transaction is returned. Likewise, if I would like to see transaction only on 2026-09-03, I would enter the same date for start date and end date, but since end date is not included I would not get any transaction in return.

https://github.com/user-attachments/assets/66ed347f-0e0d-49ba-a7c4-aac8633f0991

Also, I think it would be good to show to the user the date span that is being used for the search.

2. Incorrect result transaction set when date fillter 3 Last days is used. I assume that month is not being taken into considaration only day.

https://github.com/user-attachments/assets/09a58384-de26-4502-9a21-7929e416356e

Sup3rlativ3:
> 1. When custom date is used to make a search, the end date is not included in search. For example, if there is a transaction on 2026-03-31, but if I choose dates for example, start date = 2026-03-24, end date = 2026-03-31, no transaction is returned. Likewise, if I would like to see transaction only on 2026-09-03, I would enter the same date for start date and end date, but since end date is not included I would not get any transaction in return.

You didn't set the time. So you were searching for midnight but the transaction didn't happen until 13:46. If you set the time in the same query to 13:50 it will work as expected. This follows the same convention as the invoices filtering.

https://github.com/user-attachments/assets/851dc0ae-507a-4011-be6e-2bdd8b00cff4

> Also, I think it would be good to show to the user the date span that is being used for the search.

It was determined above that having extra text should not be included. Happy to change if that's what the team wants.

> Incorrect result transaction set when date fillter 3 Last days is used. I assume that month is not being taken into considaration only day.

I'm assuming this is the same as the first point. I see no reason why this wouldn't be working. 

It does bring up the point though that technically the 3 day filter is actually a 72 hour filter. Would it be more user friendly to change the preset options to include 3 calendar days instead of the 72 hours. This is also how the invoices page works. So if the decision were made to make the change to 3 calendar days then 1) it should be in a new PR and 2) the invoices page would also need to be changed.

@Sanja22B

NicolasDorier:
Since I want to releease 2.3.7 end of the week and this PR is 1000+ line of code of review, I will kick the can for 2.4.0. (or 2.3.8 if we ever need a minor)

Sup3rlativ3:
Uncle changed the tag to 2.4.0 the other day. Just let me know if there are any issues and I will try to resolve as soon as I can.

ABC-8888:
@Sup3rlativ3 regarding the first point, thanks for the answer it does make sense. I was just thinking it would be better than to set default time for end date to 23:59:59 (not sure why the seconds are included in the search when they are not shown in the result), because setting the time is extra step and is counterintuitive. As far as I know, it is a standard to put 23:59 for end date, but if you want to keep it consistent with other views, that is fine. I will add @pavlenex 

And for the second point, you can reproduce the issue if you enter 3 in input field and choose Last 3 days in date filter, or if you enter 7 in input field and choose Last 7 days. I had to do a little of debugging, but I think the issue arises because search text and search term are combined and later function WithoutSearchText() removes all 3s from the combined text.
For example,
combinedSearchTerm = "3,direction:in,label:belgrade,startdate:-3d"
after WithoutSearchText() removes both 3, and returns "direction:in,label:belgrade,startdate:-d"
since -d does not exist as a case in GetFilterDate it does not set startDate and all transaction are returned.
I am not saying what is the solution because I just quickly debugged it, and I assumed incorrectly in the previous post. as this is not something that would happen often I do not think it is a big problem.

Sup3rlativ3:
> I was just thinking it would be better than to set default time for end date to 23:59:59 (not sure why the seconds are included in the search when they are not shown in the result)

I was thinking about a few QoL fixes for this part of the app and this was one of them. While it isn't shown in the interface, the seconds are used in the filtering (you can see this if you open the source for the page). The other QoL fix I was thinking about adding was persisting the dates in the filter button so that a user wouldn't need to enter the start and end date again if they wanted to adjust the query. That would need to go into another PR though as this one is large and old enough already.

> combinedSearchTerm = "3,direction:in,label:belgrade,startdate:-3d"
after WithoutSearchText() removes both 3, and returns "direction:in,label:belgrade,startdate:-d"
since -d does not exist as a case in GetFilterDate it does not set startDate and all transaction are returned.

I'll try to investigate this over the weekend.

Sup3rlativ3:
I’ve confirmed the bug. The root cause predates this PR and goes back to commit c063c70b07 from May 2024.

The issue is in the shared ```SearchString.WithoutSearchText()``` helper, which removes free-text search by raw substring replacement. Because that helper is shared, this also affects the Invoices view, not just wallet transactions.

My suggestion is to raise a separate issue and address it in a dedicated follow-up PR, since the fix should be handled as a shared search/filtering change rather than as part of this wallet-focused PR, especially given the feedback that this PR is already large and costly to review. If the team would prefer me to address it in this PR instead, I’m happy to do that as well.

@pavlenex @NicolasDorier @dstrukt

Edit: Created the bug to track either way
https://github.com/btcpayserver/btcpayserver/issues/7306

BTCPayServer.Tests/FastTests.cs
```diff
@@ -1501,6 +1512,51 @@ public void CanParseFilter()
             Assert.Null(modified.GetFilterArray("enddate"));
         }
 
+        [Fact]
+        public void BuildWalletTransactionsFilterSeparatesTextAndStructuredTerms()
```
NicolasDorier, line 1518: rebuild that test without reflection. If you need access to something just put visibility to internal.

NicolasDorier (changes requested):

NicolasDorier:
@Sup3rlativ3 sorry for the delay, there was another big PR on wallet refactoring that was pending and this got merged.
Can you fix the conflicts?

NicolasDorier:
@coderabbitai review

PR #7139, opened 2026-01-26
Fix: Some HTTP codes should not restart cause a webhook to be retried

### LINKED ISSUE #3146: Webhook does not get fired anymore (if it failed before multiple times)

NicolasDorier:
Fix #3146

Some after consulting with ChatGPT, the best practice for webhooks is the following.
The webhook should be re-attempted in the following cases:
* A server error is returned (5xx)
* 429 (Rate limiting)
* 408 (Request timeout)
* Network / connection errors

For the rest, the webhook will not be retried:
* 4xx (except 508, 429)
* 2xx and 3xx (success)


ndeet:
This sounds like a good strategy to avoid 404 errors clogging up the webhook queue :clap:

PR #7140, opened 2026-01-26
Feature: Add DeliveryTime property to webhook deliveries

### LINKED ISSUE #3146: Webhook does not get fired anymore (if it failed before multiple times)

NicolasDorier:
Fix #3146 

As @ndeet explained, the UI in Invoices was showing the time of creation of the webhook delivery. Not when the delivery was actually sent to the server.

Normally, those two timestamps are near identical. However, if the downstream service receiving the webhook fails to process a delivery, the webhooks to this service become queued.

When the downstream service come back, the deliveries get retried and successfully sent.

Now, it is possible to hover on the time, to see when the event has been sent.

image

If the difference between the creation and the delivery is more than one minute, a badge appear.

image

### DIFF
--- a/BTCPayServer.Client/Models/WebhookDeliveryData.cs
+++ b/BTCPayServer.Client/Models/WebhookDeliveryData.cs
@@ -10,6 +10,8 @@ namespace BTCPayServer.Client.Models
         public string Id { get; set; }
         [JsonConverter(typeof(NBitcoin.JsonConverters.DateTimeToUnixTimeConverter))]
         public DateTimeOffset Timestamp { get; set; }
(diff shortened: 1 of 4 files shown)


ndeet:
This is great! Very helpful with debugging if there are any issues caused by network or server outages. Thank you very much :clap: :green_heart:

ndeet:
Maybe also good to add it here?
https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Webhooks_GetWebhookDelivery
https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Webhooks_GetWebhookDeliveries

PR #7169, opened 2026-02-07
feat: add Desiboard rate provider for BTC/USD and BTC/INR rates

Abhijay007:
Currently, we have a few rate providers available like (freecurrencyrate), but they are not very accurate for INR rates. Some merchants mentioned that it would be very helpful if we could add INR rate sources such as DesiBoard INR rates or GetBit INR rates, as this would give them more flexibility when charging customers.

image

credits @thevikas for the rate provider


NicolasDorier:
Hey shouldn't it be the recommended exchange for INR? Check `RegisterExchangeRecommendations` and probably better to add it in the list?

Abhijay007:
> Hey shouldn't it be the recommended exchange for INR? Check `RegisterExchangeRecommendations` and probably better to add it in the list?

Sure will add it

PR #7170, opened 2026-02-08
feat: add getBit rate provider for BTC/INR rates

Abhijay007:
Added getBit INR rates : 

image


Kukks (changes requested):
Please add it to the rates test.  

Should this also be the default for INR or?

Abhijay007:
> Please add it to the rates test.
> 
> Should this also be the default for INR or?

will do that, the default will be coinDCX, in talks with the getBIT team to know more about the API doc

PR #7181, opened 2026-02-13
Beldex integration in btcpay-server

codeman-crypto:


NicolasDorier:
you don't need this PR. Those are related to backward compatible stuff when XMR wasn't a plugin.

PR #7186, opened 2026-02-14
Improve accessibility and UX for GlobalNav topbar icons

### LINKED ISSUE #6902: [WIP] Updated Global Topbar & Search
# W.I.P

# Goal

The goal of updating the sidebar/topbar would be help delineate the Account/Server functionality from the Store specific functionality further.

We have recognized that having everything on the left sidebar has blurred this line, and we're trying to offer more mass Store functionality, as well as specific Server level control that was otherwise lacking.

# Initial Pass (Not Recommended)

I explored a few options loosely (and very early - during the time we made the shift to the left sidebar in 1.4), by separating the Stores in their own secondary sidebar. However, I don't think we gain anything special for the space we lose - imo, this isn't the direction to go.

Image

# Secondary Pass (Recommended)

I would suggest we keep left sidebar as is, and incorporate a new fixed topbar on the right side. This would include the Search, Notifications, Server Overview, Server Settings, Account, and possibly Plugins(? - TBD, we have a few decisions to discuss here). The icons would have a hover tooltip to help identify what it is, and hidden upon click. Additionally, body content would be hidden upon scroll like so:

Image

This is very early concept, but hypothtically the Global Search would allow the user to navigate to specific pages, identify Settings sections making it easier to discover information. We could explore the idea of Store level filtering, or perhaps multi-store level filtering, etc.. for the Server Admins - the functionality could be dependent on the user's permission.

Image

# WIP

I'm still working through how this will degrade on mobile, but I'll update with mocks once I've finished.

-----

The goal of posting this is to provide something concrete to discuss, and push concepts the team has been interested in exploring. This will not and is not to be implemented until we have agreed as a team that this is the direction to move.

I will continue to refine the idea, but thought I'd share the rough state it's in now to garner feedback and discussion.

AayushKrGupta:
## Summary

Improve accessibility and UX consistency for GlobalNav topbar icons.

## Changes

- Added aria-label attributes to icon-only buttons for better screen reader support
- Added title attributes to provide hover tooltips
- Reused existing localization strings (no new translations added)
- Improved hover styling for Notifications dropdown button to match GlobalNav buttons

## Scope

- No layout or structural changes
- No backend logic changes
- Pure accessibility + UI polish

Fixes #6902

### DIFF
--- a/BTCPayServer/Blazor/NotificationsDropDown.razor
+++ b/BTCPayServer/Blazor/NotificationsDropDown.razor
@@ -16,7 +16,7 @@
 @inject EventAggregator _EventAggregator
 
 <div id="Notifications" class="dropdown">
-	<button id="NotificationsHandle" class="mainMenuButton dropdown-toggle" title="@StringLocalizer["Notifications"]" type="button" data-bs-toggle="dropdown" data-bs-display="static" data-bs-auto-close="outside" aria-expanded="false">
+	<button id="NotificationsHandle" class="mainMenuButton dropdown-toggle" aria-label="@StringLocalizer["Notifications"]" title="@StringLocalizer["Notifications"]" type="button" data-bs-toggle="dropdown" data-bs-display="static" data-bs-auto-close="outside" aria-expanded="false">
 		<Icon Symbol="nav-notifications" />
 		@if (HasUnseenNotifications)
 		{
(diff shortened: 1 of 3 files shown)


NicolasDorier:
can you rebase on master? pavlenex made a commit on master that I needed to remove.

AayushKrGupta:
> can you rebase on master? pavlenex made a commit on master that I needed to remove.

yeah okk i'll do it

AayushKrGupta:
> can you rebase on master? pavlenex made a commit on master that I needed to remove.

Rebased on latest master 👍

AayushKrGupta:
> can you rebase on master? pavlenex made a commit on master that I needed to remove.

Rebased on latest master and force-pushed the updated branch.
Please let me know if anything else needs adjustment

PR #7192, opened 2026-02-17
Update wallet folder access instructions for Wasabi

nopara73:
Not sure what's "Tools" and "Wallet Manager" is. I suspect they don't exist.

image


nopara73:
first🚀

PR #7197, opened 2026-02-20
Refactor: Move translations classes in its own plugin

NicolasDorier:


NicolasDorier:
@coderabbitai can you check if my idea to delete UpdateLanguagePack and reuse DownloadLanguagePack instead wouldn't break the update?

PR #7201, opened 2026-02-22
feat: add cheat-mode endpoint for default translations

Abhijay007:
ref pr : https://github.com/btcpayserver/btcpayserver-translator/pull/23

### DIFF
--- a/BTCPayServer/Plugins/Translations/Controllers/CheatTranslationController.cs
+++ b/BTCPayServer/Plugins/Translations/Controllers/CheatTranslationController.cs
@@ -10,8 +10,8 @@ namespace BTCPayServer.Plugins.Translations.Controllers;
 
 [AllowAnonymous]
 [Area(TranslationsPlugin.Area)]
-public class UICheatTranslationController(
-    IEnumerable<IDefaultTranslationProvider> defaultTranslationProviders) : Controller
+public class CheatTranslationController(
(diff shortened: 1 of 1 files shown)


BTCPayServer/Plugins/Translations/Controllers/UICheatTranslationController.cs
```diff
@@ -0,0 +1,28 @@
...
+using BTCPayServer.Filters;
+using BTCPayServer.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace BTCPayServer.Plugins.Translations.Controllers;
+
+[AllowAnonymous]
+[Area(TranslationsPlugin.Area)]
+public class UICheatTranslationController(
```
NicolasDorier: Don't call it `UI...Controller` this isn't a UI.

BTCPayServer/Plugins/Translations/Controllers/UICheatTranslationController.cs
```diff
@@ -0,0 +1,28 @@
...
+    IEnumerable<IDefaultTranslationProvider> defaultTranslationProviders) : Controller
```
NicolasDorier: Use ControllerBase, not Controller. Controller is for UI related controllers.

NicolasDorier (changes requested):

PR #7202, opened 2026-02-23
Add: Implement UpdateCrowdfundApp endpoint and client method

webiumsk:
- Added a new API endpoint to update properties of a Crowdfund app in GreenfieldAppsController.
- Introduced request validation for the update operation, ensuring required fields are checked.
- Updated Swagger documentation to include the new PUT operation for Crowdfund apps.
- Added corresponding client method in BTCPayServerClient for making requests to the new endpoint.

### DIFF
--- a/BTCPayServer.Tests/GreenfieldAPITests.cs
+++ b/BTCPayServer.Tests/GreenfieldAPITests.cs
@@ -632,7 +632,7 @@ namespace BTCPayServer.Tests
             Assert.Equal("test description", retrievedCfApp.Description);
             Assert.False(retrievedCfApp.Archived);
 
-            // Test that we can update the crowdfund app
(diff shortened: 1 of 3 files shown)


NicolasDorier:
need test

rockstardev:
pretty cool PR @webiumsk 

+1 on test

webiumsk:
Test added. @NicolasDorier please, check it now

Abhijay007:
Hi @webiumsk can you please look into merge conficlts

NicolasDorier:
This `HttpPut` is actually doing a patch. (Fields that are not present are ignored)
Why is this the case? We never did patching. The users who want to patch should GET, update then PUT instead, all of our APIs behave this way. (except maybe one or two exceptions)

PR #7203, opened 2026-02-24
feat: Add search-first label filter for scalable label management

### LINKED ISSUE #7109: [Feature]: Improve label system to work when we have 10+ labels
Now that we are integrating the label system into Payment Requests as well (https://github.com/btcpayserver/btcpayserver/pull/7050), we should improve label UX so it scales to 10, 100, or 1,000 labels.

The current wallet UI that dumps all labels into the filter dropdown will not scale:

https://github.com/user-attachments/assets/a9de89de-b877-472e-a776-921a46da87c9

From the video, it is clear the labels filter dropdown should never exceed page height. A good approach would be to reuse the dropdown we already use in transaction rows for applying labels:

Image

Proposal:

1. In the filter dropdown, show the top ~20 labels by default, sorted alphanumerically.
2. If more labels exist, show a hint that the user can type to search (typeahead) to find and apply other labels:

Image

3. Apply the same approach to row label pickers, ideally defaulting to the top 20 most-used labels.

With this in place, label filtering and assignment should scale cleanly to any number of labels.

aizuanjeme:
Implements a search-first filter approach for label dropdowns that scales to 1000+ labels. Instead of preloading all labels in a visible dropdown, labels are hidden by default and revealed dynamically as the user types.

Changes:
- Add initLabelFilterDropdowns() in site.js to handle search/filter logic
- Add CSS styles for label-filter-dropdown, search input, hint, and no-results states in site.css
- Update WalletTransactions.cshtml with search-first label filter UI
- Update GetPaymentRequests.cshtml with search-first label filter UI
- Labels sorted alphabetically for consistent ordering
- Dropdown stays open during search (data-bs-auto-close=outside)
- Auto-focus search input on dropdown open, reset on close

Fixes #7109

### DIFF
--- a/BTCPayServer/Views/UIPaymentRequest/GetPaymentRequests.cshtml
+++ b/BTCPayServer/Views/UIPaymentRequest/GetPaymentRequests.cshtml
@@ -239,11 +239,11 @@
 
             <ul class="dropdown-menu label-filter-dropdown" aria-labelledby="LabelOptionsToggle">
                 <li class="label-filter-search">
-                    <input type="text" class="form-control form-control-sm label-filter-search-input" placeholder="@StringLocalizer["Search labels..."]" autocomplete="off" />
+                    <input type="text" class="form-control form-control-sm label-filter-search-input" placeholder="@StringLocalizer["Search labels..."]" aria-label="@StringLocalizer["Search labels..."]" autocomplete="off" />
                 </li>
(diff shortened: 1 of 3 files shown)


rockstardev:
https://github.com/btcpayserver/btcpayserver/issues/7109#issuecomment-3954403659

Issue #6895, opened 2025-08-20
Multisig number of signatures is inconsistencies counted ocassionaly


rockstardev:
We occasionally run into issues counting the number of collected signatures in multisig. Here is a recent example showing the problem:

<img width="1349" height="168" alt="Image" src="https://github.com/user-attachments/assets/ab0862cf-cdc6-4fbf-885d-64c5cf8716b5" />

This occurred for the transaction: https://mempool.space/tx/38ccc8e80852dc9d3122b2e0ed16d743736ebbbc2a606cf04796fc07b372abd1

The relevant code is here: https://github.com/btcpayserver/btcpayserver/blob/5ce83d55f7f56b5afe069081f87640f513041644/BTCPayServer/HostedServices/PendingTransactionService.cs#L200

monicamuyama:
I have been reviewing the logic where we estimate the number of collected signatures and also been doing some research on the correct implementation
The approach currently used,  where we divide the signatures by the inputs works for single-sig but can be inaccurate for multisig transactions especially when multiple inputs require more than one signature each 
 I am planning to improve this by replacing the estimation with a count derived directly from the PSBT data. 
Specifically, I’ll:
- Iterate through each PSBTInput
-  Use PartialSigs.Count to determine how many signatures have been collected for that input
-  Detect the required number of signatures from the redeem or witness script when it’s a multisig
- Aggregate these counts to update SignaturesCollected and SignaturesNeeded accurately
Does this approach sound good before I proceed

psam21:
Hi @rockstardev  @monicamuyama 

My understanding:
So current code does signatures divided by inputs. This integer division can produce wrong results with many inputs. 

From the the transaction below that has 50 inputs and would show fully signed transaction displaying 0 signatures with state as Signed.
https://mempool.space/tx/edf619bc2c11a091e35c09fe083bef6cbdb4fd906f1c90b9b998d8ab9cc99897

Scenarios to handle:
Mixed inputs: 2 multisig inputs (2 of 4) + 1 single sig in same transaction
One person with multiple keys: Same person signs with 2 of the 4 keys

Proposal:
We compute all of this accurately from PSBT in the backend. Track unique signers, min signatures per input, and broadcast readiness as separate values. The UI currently shows a single number, but if we have the full set of values available, the frontend can decide what to display or expand on it later.

rockstardev:
[...]

PR #7218, opened 2026-03-06
feat: Add multisig wallet setup

### LINKED ISSUE #6651: Create new Roles and Permissions for accessing Bitcoin wallet and participating in multisig
With the release of v2.1, we will introduce multisig support, necessitating that we introduce access control to Store Wallets. This will allow creation of roles and inviting of users that only have store access in order to participate in signing multisig transactions.

At a minimum, we should extract the `Manage Wallets` permission, which can include the following categories:

- `Modify Bitcoin Onchain`
- `Modify Bitcoin Lightning`
- `Modify Other Wallets` (for altcoins)

Within each wallet, we could have permissions such as `View Wallet` and `Manage Settings`. For Bitcoin Onchain multisig, there would be one more permission group `Manage Transactions`, and within it:

- Create Transactions
- Sign Transactions
- Broadcast Transactions
- Cancel Transactions

Access to the transaction PSBT and its download will be restricted to those who can manage transactions.

Image

Right now, our Store Roles are very payment-centric, and we could add a few more focused on multisig:
- Wallet Manager (root `Manage Wallets` permission) 
- Multisigner (`View Wallet`, `Manage Transactions`) 
- Multisigner Guest (`View Wallet`, `Sign Transactions`)

thgO-O:
## Summary

This PR adds a guided multisig wallet setup flow for BTC stores.

Store owners can start a multisig setup, choose the signing policy, invite store users to submit their signer keys, review generated addresses, and create the wallet once all required signer keys are collected.

## Changes

- Add Multisig plugin with setup and invite flows.
- Add signer key submission by manual entry or BTCPay Server Vault.
- Add dashboard and wallet list entries for multisig setups in progress.
- Add email notifications for signer requests, submitted keys, wallet creation, and pending multisig signatures.
- Track pending multisig signature progress safely.
- Prevent duplicate/no-op signature submissions from emitting misleading notifications.
- Add tests for multisig setup, signer invites, pending signatures, and concurrency handling.

Resolves #6651 #6895 

https://github.com/user-attachments/assets/f3fa6a0c-a9e0-403b-8939-41313fdb80ae


BTCPayServer/Plugins/Multisig/Controllers/UIMultisigInviteController.cs
```diff
@@ -0,0 +1,181 @@
...
+        if (!IsSupportedCryptoCode(cryptoCode))
+            return NotFound();
+        var result = await LoadInviteViewModel(storeId, cryptoCode, token);
+        if (result.Status is InviteLoadStatus.WrongUser)
+            return RedirectToAction("Login", "UIAccount", new { area = "", returnUrl = Request.Path + Request.QueryString });
+        if (result.Status is not InviteLoadStatus.Ok)
+            return NotFound();
+        var current = result.ViewModel!;
+
+        var pending = await storeRepository.GetSettingAsync<PendingMultisigSetupData>(storeId, MultisigService.GetPendingMultisigSettingName(cryptoCode));
```
rockstardev: There is a race condition here. The request loads the full pending setup, checks whether this participant has already submitted, mutates the in-memory object, and then writes the whole setting back. Because StoreRepository.UpdateSetting replaces the serialized value without any concurrency token/version check, two concurrent submissions can both pass the precondition and the later write can overwrite the earlier one.

BTCPayServer/Plugins/Wallets/Services/OnChainWalletSetupService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    public async Task<OnChainWalletSetupResult> SaveWallet(StoreData store, BTCPayNetwork network, DerivationSchemeSettings derivationSchemeSettings, WalletSetupRequest? setupRequest)
+    {
+        var wallet = walletProvider.GetWallet(network);
+        if (wallet is null)
+            return new OnChainWalletSetupResult(false, "Wallet is not available.");
+
+        try
+        {
+            await wallet.TrackAsync(derivationSchemeSettings.AccountDerivation);
```
rockstardev: This whole block is non-atomic. TrackAsync runs before the store update is persisted. If UpdateStore fails after tracking succeeds, NBXplorer starts tracking a wallet that BTCPay does not persist, and there is no rollback path here. 

Also, catch block reports every failure as an NBXplorer tracking problem, which can hide other errors (DB command failed).

BTCPayServer/Plugins/Multisig/Controllers/UIMultisigInviteController.cs
```diff
@@ -0,0 +1,181 @@
...
+        participant.SubmittedAt = DateTimeOffset.UtcNow;
+        await storeRepository.UpdateSetting(storeId, MultisigService.GetPendingMultisigSettingName(cryptoCode), pending);
+        await multisigNotificationService.NotifyRequesterOfSubmission(HttpContext, storeId, cryptoCode, pending, participant);
+
+        current.Submitted = true;
+        TempData[WellKnownTempData.SuccessMessage] = "Signer key submitted successfully.";
+        return View("MultisigInvite", current);
+    }
+
+    private async Task<InviteLoadResult> LoadInviteViewModel(string storeId, string cryptoCode, string token)
```
rockstardev: The invite flow validates the token payload and authenticated user id, but it never re-checks that the user still belongs to the store. A user who was invited and later removed from the store can still submit a signer key as long as their participant entry remains in the pending request.

BTCPayServer/Plugins/Multisig/Controllers/UIMultisigSetupController.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        PendingMultisigSetupData? finalizedMultisigRequest = null;
+        if (!string.IsNullOrEmpty(vm.MultisigRequestId))
+        {
+            finalizedMultisigRequest = await multisigService.GetPendingMultisigSetup(vm.StoreId, vm.CryptoCode, vm.MultisigRequestId)
+                                      ?? await multisigService.GetLatestPendingMultisigSetup(vm.StoreId, vm.CryptoCode);
+        }
+
+        if (finalizedMultisigRequest is not null)
+            await multisigNotificationService.SendWalletCreatedEmails(HttpContext, vm.StoreId, vm.CryptoCode, finalizedMultisigRequest);
```
rockstardev: At this point SaveWallet has already succeeded. If SendWalletCreatedEmails throws, the wallet remains created but the pending multisig request is never cleared and the user gets a failed request. Notification delivery should be best-effort here, or the cleanup should happen before the email step.

BTCPayServer/Plugins/Multisig/Models/PendingMultisigSetupData.cs
```diff
@@ -0,0 +1,31 @@
...
+    public string CryptoCode { get; set; }
+    public string RequestedByUserId { get; set; }
+    public string RequestedByEmail { get; set; }
+    public string RequestedByName { get; set; }
+    public string ScriptType { get; set; }
+    public int RequiredSigners { get; set; }
+    public int TotalSigners { get; set; }
+    public DateTimeOffset CreatedAt { get; set; }
+    public DateTimeOffset ExpiresAt { get; set; }
+    public bool Finalized { get; set; }
```
rockstardev: It seems that `Finalized` is checked in a few places, but I couldn’t find anywhere it is ever set to `true`. As a result, the flag is not actually enforcing a completed/finalized lifecycle right now. Is that intentional, or does this need to be cleaned up?

BTCPayServer/Plugins/Multisig/Controllers/UIMultisigInviteController.cs
rockstardev: Also, tokens are not single-use - after a participant has submitted an AccountKey, later POSTs with the same invite are used to show the existing submission. If that is the intended replay behavior, it may be worth documenting here so code reviewers in future do not assume the token itself is meant to be single-use.

rockstardev (changes requested):

r2ckstardev:
Reviewed against current HEAD (commit aa9c129). The signature-counting refactor in `PendingTransactionService` is clean — per-input PSBT-based counting via `GetSignatureProgress` / `BuildEffectivePsbt` / `ApplyProgress` is the right approach.

Two issues from the prior review round still need attention, plus one role-scope concern.

### 1. Lost-update race in `PendingTransactionService.CollectSignature`

`BTCPayServer/HostedServices/PendingTransactionService.cs` lines 129-187.

Two concurrent signers can both read the same `PendingTransaction` row, both append a `CollectedSignature` to their local blob, and the second `SaveChangesAsync` silently overwrites the first. There is no `RowVersion` / `[Timestamp]` / concurrency token on `PendingTransaction`, and the blob is stored as a single JSON column — so EF Core has nothing to detect lost updates with.

The dedupe check at line 150 (`s.ReceivedPSBT == newPsbtBase64`) does not save it: in the race window both writers dedupe against the same stale snapshot, so neither sees the other's signature.

**Suggested fix:** add a `[Timestamp] public byte[] RowVersion { get; set; }` to `PendingTransaction`, and wrap `CollectSignature` in a retry-on-`DbUpdateConcurrencyException` loop that re-reads, re-merges and retries.

### 2. `wallet.TrackAsync` ordering in `OnChainWalletSetupService.SetupWallet`

`BTCPayServer/Services/Wallets/OnChainWalletSetupService.cs` lines 56-71.

`wallet.TrackAsync(...)` is awaited at line 58, but `storeRepository.UpdateStore(store)` is not called until line 71. If `UpdateStore` throws (transient DB error, validation, etc.), NBXplorer is left tracking a derivation scheme that was never persisted to the store.

**Suggested fix:** either persist the store first and only `TrackAsync` after a successful `UpdateStore`, or add a compensating untrack in the failure path.

### 3. `Multisigner Guest` permission scope

`BTCPayServer.Data/Migrations/20260203123000_AddWalletRoles.cs` lines 36-43.

Both `Multisigner` and `Multisigner Guest` are seeded with `btcpay.store.canmodifybitcoinonchain`. That permission lets the holder reconfigure the on-chain wallet derivation scheme — not just sign transactions. A guest signer should not be able to mutate the wallet config.

**Suggested fix:** narrow `Multisigner Guest` to `canviewwallet` + `cansigntransactions` only. Consider whether `Multisigner` (non-guest) really needs full `canmodifybitcoinonchain` either, or whether a narrower `canmanagewallettransactions` is sufficient for the signing-coordinator role.

---

Happy to code the fixes for any of the above if you'd like — ping me and I'll open a follow-up PR with the `RowVersion` migration, the ordering fix, and a tightened role seed.

r2ckstardev:
On the permission naming - recommendation:

**Best fit: `canusebitcoinonchain`** (+ matching `canusebitcoinlightning`, `canuseotherwallets`)

Why it's the right word:
- `canuse*` already exists in the codebase: `canuselightningnode`, `canuseinternallightningnode`. Not a new verb, stays inside the standard.
- Semantically accurate: it's a **use/participation gate** for a wallet surface, not a modify verb. You need it to show up in the bitcoin-onchain UI at all, then specific verbs (`canmanagewalletsettings`, `cansigntransactions`, etc.) gate the actual actions.
- Multisigner Guest holding `canusebitcoinonchain + canviewwallet + cansigntransactions` reads correctly on first glance. No AI reviewer (or human) would flag it as "guest can modify the wallet" because the verb no longer implies mutation.

NicolasDorier:
@coderabbitai  review

pavlenex:
Sick updates @thgO-O can't wait to try it out!

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -360,6 +360,7 @@ public static IServiceCollection AddBTCPayServer(this IServiceCollection service
             services.TryAddSingleton<StoreLabelRepository>();
             services.TryAddSingleton<PaymentRequestRepository>();
             services.TryAddSingleton<BTCPayWalletProvider>();
+            services.TryAddSingleton<OnChainWalletSetupService>();
```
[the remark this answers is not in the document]
thgO-O: I checked this and it looks like a false positive. AddDataProtection() registers IDataProtectionProvider as singleton, not scoped, and this service only creates/caches an IDataProtector for a fixed purpose. The other dependencies are singleton too, so keeping OnChainWalletSetupService as singleton is fine.

NicolasDorier:
@coderabbitai can you do another review?

BTCPayServer/Plugins/Wallets/Services/OnChainWalletSetupService.cs
```diff
@@ -0,0 +1,77 @@
+#nullable enable
```
NicolasDorier: This file should be in the plugin dir.

BTCPayServer/Plugins/Multisig/Controllers/UIMultisigInviteController.cs
```diff
@@ -0,0 +1,208 @@
...
+
+            var participant = pending.Participants.Find(p => string.Equals(p.UserId, current.UserId, StringComparison.Ordinal));
+            if (participant is null)
+                return NotFound();
+
+            if (!string.IsNullOrWhiteSpace(participant.AccountKey))
+            {
+                // Invite links remain reusable for the intended signer until the request expires,
+                // but once a key is submitted the page is read-only and never overwrites it.
+                current.AccountKey = participant.AccountKey;
```
NicolasDorier: Can you rename this one to `DisplayAccountKey` and add another property `AccountKey` of type `BitcoinExtKey`, so you can follow what the type `AccountKeySettings` is doing?

That way you can remove `TryNormalizeAccountKeyForNetwork` and `NormalizeAccountKey`.

This will:
* If one day our normalization rules changes, we don't have to do a database migration
* Forgetting to normalize the current AccountKey (I think you forgot just bellow on `current.AccountKey`)

BTCPayServer/Plugins/Multisig/Services/MultisigService.cs
```diff
@@ -0,0 +1,471 @@
...
+            var key = NormalizeAccountKey(participant.AccountKey, network);
+            if (string.IsNullOrEmpty(key))
+                continue;
+
+            participantsByKey.TryAdd(key, participant);
+        }
+
+        foreach (var accountSettings in strategy.AccountKeySettings)
+        {
+            var key = NormalizeAccountKey(accountSettings.AccountKey?.ToString(), network);
```
NicolasDorier: What are you normalizing here? This AccountKey is a BitcoinExtKey, so it's already normalized?

NicolasDorier (changes requested):

BTCPayServer/Plugins/Multisig/Services/MultisigService.cs
```diff
@@ -0,0 +1,471 @@
...
+            normalizedAccountKey = new BitcoinExtPubKey(accountKey.Trim(), network.NBitcoinNetwork).ToString();
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    public static string NormalizePath(string path)
```
NicolasDorier: This one could return a `KeyPath?` instead of string. Everywhere you are parsing just after.

BTCPayServer/HostedServices/PendingTransactionService.cs
```diff
@@ -123,121 +110,134 @@ public async Task<PendingTransaction> CreatePendingTransaction(string storeId, s
...
             Data = pendingTransaction,
+            SignerUserId = null,
             Type = PendingTransactionEvent.Created
         });
 
         return pendingTransaction;
     }
 
-    public async Task<PendingTransaction?> CollectSignature(PendingTransactionFullId id, PSBT psbt, CancellationToken cancellationToken)
+    public async Task<PendingTransaction?> CollectSignature(PendingTransactionFullId id, PSBT psbt, CancellationToken cancellationToken, string? signerUserId = null)
```
NicolasDorier, line 138: You don't have to fix this, I can cleanup later.
But note that in this situation, the use of `goto retry`  is way more readable.
Also, you should just rethrow the exception if too many fail rather than creating a new one.

BTCPayServer/Plugins/Multisig/Services/MultisigNotificationService.cs
```diff
@@ -0,0 +1,291 @@
...
+
+namespace BTCPayServer.Plugins.Multisig.Services;
+
+public class MultisigNotificationService(
+    EmailSenderFactory emailSenderFactory,
+    StoreRepository storeRepository,
+    PermissionService permissionService,
+    MultisigService multisigService,
+    ILogger<MultisigNotificationService> logger)
+{
```
NicolasDorier, line 23: We already have a the email rule system, reuse it.

NicolasDorier (changes requested):

BTCPayServer/Plugins/Multisig/Services/MultisigNotificationService.cs
```diff
@@ -0,0 +1,113 @@
...
+            participant.Email,
+            participant.Name,
+            setupLink));
+        return Task.CompletedTask;
+    }
+
+    public Task PublishWalletCreatedEvent(HttpContext httpContext, string storeId, string cryptoCode, PendingMultisigSetupData pending)
+    {
+        var walletId = new WalletId(storeId, cryptoCode).ToString();
+        var walletLink = $"{httpContext.Request.PathBase}/wallets/{walletId}";
```
NicolasDorier: Use `LinkGenerator` for generating links, not string concat.

BTCPayServer/Plugins/Multisig/Views/Shared/Components/MultisigDashboardSetupGuide/Default.cshtml
```diff
@@ -0,0 +1,32 @@
+@model IReadOnlyList<BTCPayServer.Plugins.Multisig.Models.MultisigInProgressViewModel>
+
+@foreach (var multisig in Model)
+{
+    var setupState = multisig.ReadyToCreateWallet
```
NicolasDorier, line 5: You should localize this with the `StringLocalizer`

BTCPayServer/Plugins/Multisig/Views/Shared/Components/MultisigDashboardSetupGuide/Default.cshtml
```diff
@@ -0,0 +1,32 @@
...
+            ? "Create wallet"
+            : "View status";
+    var actionUrl = multisig.CanSubmitSignerKey ? multisig.InviteUrl : multisig.SetupUrl;
+    <a href="@actionUrl"
+       id="SetupGuide-Multisig-@multisig.RequestId"
+       class="list-group-item list-group-item-action d-flex align-items-center order-1">
+        <vc:icon symbol="wallet-new" />
+        <div class="content">
+            <h5 class="mb-0" text-translate="true">Multisig wallet setup</h5>
+            <div class="text-secondary small">@($"{multisig.RequiredSigners}-of-{multisig.TotalSigners} {multisig.CryptoCode} · {multisig.SubmittedSigners}/{multisig.TotalSigners} submitted · {setupState} · {participationState}")</div>
```
NicolasDorier: Should be using `StringLocalizer`.

NicolasDorier (changes requested):

NicolasDorier:
@coderabbitai review

NicolasDorier:
@coderabbitai, it should be fine now, can you check

PR #7226, opened 2026-03-10
show payment method on receipt

### LINKED ISSUE #7174: Show payment method on receipt

BTCPay Server 2.3.4

On the invoice receipt page we don't show an indicator which payment method was used. You can kind of guess it with ln addresses or Bitcoin addressed but not for others. Especially when you use e.g. the new Stripe plugin you have no idea what it was. Also for the cash payment plugin it would be good to have the payment method noted somewhere. cc @dstrukt 

Image

### What happened?

Paid invoice
Looked at receipt 😆 

.

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

TChukwuleta:
Resolve #7174 

image

@dstrukt  what do you think?

### DIFF
--- a/BTCPayServer/Views/UIInvoice/InvoiceReceipt.cshtml
+++ b/BTCPayServer/Views/UIInvoice/InvoiceReceipt.cshtml
@@ -139,7 +139,7 @@
 												<td>
 													@if (!string.IsNullOrEmpty(payment.Destination))
 													{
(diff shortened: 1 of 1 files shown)


dstrukt:
Looks great!

Thanks for picking up!

My only other thought, should we flip the payment, BTC amount (larger and in black) on top, fiat amount (smaller and in grey), on bottom?

TChukwuleta:
did the flip @dstrukt 

image

PR #7240, opened 2026-03-14
Allow multiple payouts to the same destination address

rockstardev:
Previously, `SingleOrDefault` was used to find a payout matching a detected transaction's destination address. This would throw if multiple payouts existed for the same address. Changed to query all `AwaitingPayment` payouts for that address, then pick the first one whose amount matches the transaction output sum. Payouts that have already moved past `AwaitingPayment` are naturally excluded from the query, so each incoming payment matches the next unprocessed payout in FIFO order.

## Summary

- Remove the `duplicate-destination` check that rejected payouts when an active payout to the same address already existed
- Update on-chain transaction detection to match the correct payout by amount when multiple payouts share a destination

## Tests

Three new integration tests:
- Two payouts to the same address can be created without `duplicate-destination` error
- Two payouts to the same address with different amounts -- each payment matches only the correct one
- Three payouts to the same address with the same amount -- payments are matched one at a time sequentially

### DIFF
--- a/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
+++ b/BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
@@ -381,6 +381,26 @@ public class BitcoinLikePayoutHandler : IPayoutHandler, IHasNetwork
 
 
             await using var ctx = _dbContextFactory.CreateContext();
+            var txId = newTransaction.NewTransactionEvent.TransactionData.TransactionHash;
+
+            // Check if this transaction is already claimed by an InProgress payout for this destination.
+            // This prevents a single transaction from matching multiple payouts when the event fires
+            // more than once (e.g. mempool detection then block confirmation).
+            var inProgressPayouts = await ctx.Payouts
+                .Where(p => p.State == PayoutState.InProgress)
+                .Where(p => p.PayoutMethodId == PaymentMethodId.ToString())
+#pragma warning disable CA1307 // Specify StringComparison
+                .Where(p => destination.Equals(p.DedupId))
+#pragma warning restore CA1307 // Specify StringComparison
+                .ToListAsync();
+
+            foreach (var existing in inProgressPayouts)
+            {
+                var existingProof = ParseProof(existing) as PayoutTransactionOnChainBlob;
+                if (existingProof?.Candidates.Contains(txId) == true)
+                    return;
+            }
+
             var payouts = await ctx.Payouts
                 .Include(o => o.StoreData)
                 .Include(o => o.PullPaymentData)
@@ -389,7 +409,18 @@ public class BitcoinLikePayoutHandler : IPayoutHandler, IHasNetwork
 #pragma warning disable CA1307 // Specify StringComparison
                 .Where(p => destination.Equals(p.DedupId))
 #pragma warning restore CA1307 // Specify StringComparison
+                .OrderBy(p => p.Date)
                 .ToListAsync();
+
+            // Also check AwaitingPayment payouts for already-claimed txIds.
+            // This covers external payments where the payout stays in AwaitingPayment
+            // but already had its proof updated by a previous event (e.g. mempool detection).
+            foreach (var p in payouts)
+            {
+                var existingProof = ParseProof(p) as PayoutTransactionOnChainBlob;
+                if (existingProof?.Candidates.Contains(txId) == true)
+                    return;
+            }
 
             var payout = payouts.FirstOrDefault(p =>
                 p.Amount is not null &&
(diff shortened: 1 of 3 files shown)


Kukks:
I remember we used to have a dB unique constraint index at one point. I guess not if the tests pass 🤔


What happens if you have 5 payoutsnto same address at 0.2 btc each, and then someone sends 1 btc in one go to the address? Would you batch process them all? I assume this would potentially happen often if it's the payee stacking up bills over a period and then settling at once. 

If you have a single tx with 5 outputs of 0. 2 to the same address, would that also settle all those payouts too?

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -381,24 +381,23 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
 #pragma warning disable CA1307 // Specify StringComparison
                 .Where(p => destination.Equals(p.DedupId))
 #pragma warning restore CA1307 // Specify StringComparison
-                .FirstOrDefaultAsync();
+                .ToListAsync();
+
+            var payout = payouts.FirstOrDefault(p =>
+                p.Amount is not null &&
+                destinationSum ==
+                BTCPayServer.Extensions.RoundUp(p.Amount.Value, Network.Divisibility));
```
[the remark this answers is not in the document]
rockstardev: Addressed in commit e0d3654.

rockstardev:
@Kukks, no - I didn't add any grouping logic. Everything in that regard stayed the same as it was for single payouts - needed to match address and amount one by one.

I understand that even though this is existing behavior (single payouts also required exact amount matching), if we make it multi-payout, this will happen more often.

So it's an implementation call - should we encourage people who need multiple payouts to the same address to just use the wallet functionality directly? Or maybe take some queuing approach?

Kukks:
> @Kukks, no - I didn't add any grouping logic. Everything in that regard stayed the same as it was for single payouts - needed to match address and amount one by one.
> 
> I understand that even though this is existing behavior (single payouts also required exact amount matching), if we make it multi-payout, this will happen more often.
> 
> So it's an implementation call - should we encourage people who need multiple payouts to the same address to just use the wallet functionality directly? Or maybe take some queuing approach?

One thing to note is that the internal wallet may load those payouts as individual outputs but the default behavior of NBitcoin will be to merge them into one output. I think multiple libs do this too

NicolasDorier:
What is the rational for changing this? (what business case?) Please add this to the description.

At the very least, having two `Payout+Amount` at the same time should be forbidden, as it could result in a situation where the payment is not attributed to the right pull payment.

I am super worried about this code. Pull payments had lot's of bug which could have resulted into merchant losing money... I have lot's of trouble reviewing the code that is added, for a use case that seems to aim at accommodating somebody who wants to address reuse.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -381,24 +381,57 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
+            foreach (var existing in claimedPayouts)
+            {
+                var existingProof = ParseProof(existing) as PayoutTransactionOnChainBlob;
+                if (existingProof?.Candidates?.Contains(txId) == true)
+                    return;
+                if (existingProof?.TransactionId == txId)
+                    return;
+            }
+
+            var payouts = await ctx.Payouts
```
NicolasDorier, line 407: what is the difference betwen `claimedPayouts` and `payouts`? do we need 2 request to db here?

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -381,24 +381,57 @@ private async Task UpdatePayoutsAwaitingForPayment(NewOnChainTransactionEvent ne
...
                 .Where(p => destination.Equals(p.DedupId))
 #pragma warning restore CA1307 // Specify StringComparison
-                .FirstOrDefaultAsync();
+                .OrderBy(p => p.Date)
+                .ToListAsync();
+
+            // Also check AwaitingPayment payouts for already-claimed txIds.
+            // This covers external payments where the payout stays in AwaitingPayment
+            // but already had its proof updated by a previous event (e.g. mempool detection).
+            foreach (var p in payouts)
```
NicolasDorier, line 421: What is the difference with

```
  foreach (var existing in claimedPayouts)
            {
                var existingProof = ParseProof(existing) as PayoutTransactionOnChainBlob;
                if (existingProof?.Candidates?.Contains(txId) == true)
                    return;
                if (existingProof?.TransactionId == txId)
                    return;
            }
```

PR #7247, opened 2026-03-17
fix: store dashboard wallet balance chart and numbers widget

### LINKED ISSUE #7246: [Bug]: store dashboard: wallet balance switch broken

BTCPay Server v2.3.6 (at least also in 2.3.5)

docker

### What happened?

On the store dashboard wallet balance we have a toggle between 1W 1M and 1Y. When you switch to 1W or 1Y it always shows 21. Jan, 21. Jan on the timeline. Also when you click back to 1W it does not update the screen to show March dates again but stays broken. Only reload fixes screen display.

This also happens on stores with more transactions than what I have in the screenshots below.

Important: the graphic updates just fine, only the x-axis dates are broken

Bonus: also the right widget saying "Invoices paid in the last 7 days" should update accordingly to the selected timeframe on the wallet balance?

Image

1. go to store dashboard
2. in wallet balance graphic, switch to 1M or 1Y
3. see x-axis updates only to 21. Jan, 21. Jan instead of showing correct dates

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

Abhijay007:
closes #7246 

https://github.com/user-attachments/assets/fdda35d0-e20e-4aed-b575-535868b2547b

### DIFF
--- a/BTCPayServer/Components/StoreNumbers/StoreNumbers.cs
+++ b/BTCPayServer/Components/StoreNumbers/StoreNumbers.cs
@@ -27,15 +27,14 @@ public class StoreNumbers : ViewComponent
         _invoiceRepository = invoiceRepository;
     }
 
-    public async Task<IViewComponentResult> InvokeAsync(StoreData store, string cryptoCode, bool initialRendering = false, int timeframeDays = 7)
+    public async Task<IViewComponentResult> InvokeAsync(StoreData store, string cryptoCode, bool initialRendering)
     {
         var vm = new StoreNumbersViewModel
         {
             StoreId = store.Id,
             CryptoCode = cryptoCode,
             InitialRendering = initialRendering,
-            WalletId = new WalletId(store.Id, cryptoCode),
-            TimeframeDays = timeframeDays
+            WalletId = new WalletId(store.Id, cryptoCode)
         };
 
         if (vm.InitialRendering)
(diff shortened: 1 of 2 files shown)


TChukwuleta:
Hey @Abhijay007  thank you for looking at this issue..


While looking at the PR, I think the issue is that server-side labels are ISO date strings  but the API returns Unix timestamps in seconds

so changes to the view model, or introducing numbersUrl  is abit more extra


Under the Update function in the Default.cshtml file try changing:

`
 data = await response.json();
        render(data);
`

to 

`
const json = await response.json();
data = { ...json, labels: json.labels.map(l => new Date(l * 1000).toISOString()) };
render(data);
`

I think that is all you need and that should resolve the issue

Abhijay007:
> Hey @Abhijay007 thank you for looking at this issue..
> 
> While looking at the PR, I think the issue is that server-side labels are ISO date strings but the API returns Unix timestamps in seconds
> 
> so changes to the view model, or introducing numbersUrl is abit more extra
> 
> Under the Update function in the Default.cshtml file try changing:
> 
> `data = await response.json(); render(data);`
> 
> to
> 
> `const json = await response.json(); data = { ...json, labels: json.labels.map(l => new Date(l * 1000).toISOString()) }; render(data);`
> 
> I think that is all you need and that should resolve the issue

Thanks for the suggestion @TChukwuleta,  I've applied your fix in the update function by transforming the API's Unix timestamp labels to ISO strings to match the server-side format, keeping the chart renderer untouched.

I also kept the `numbersUrl` period sync feature since it improves UX, switching between 1W/1M/1Y now also refreshes the "Paid invoices in the last X days" card to reflect the selected period. Let me know if you'd prefer this in a separate PR.

TChukwuleta:
Sounds great.. yea

Well done

we can have that in a separate PR

Lets just leave the iso string resolution for this @Abhijay007

Abhijay007:
> Sounds great.. yea
> 
> Well done
> 
> we can have that in a separate PR
> 
> Lets just leave the iso string resolution for this @Abhijay007

Sure will refactor

PR #7250, opened 2026-03-19
Add LUD-21 (LNURL-pay Verify) support

### LINKED ISSUE #7248: Add LUD-21 (LNURL-pay Verify) support
## Summary

Add [LUD-21](https://github.com/lnurl/luds/blob/luds/21.md) (LNURL-pay verify endpoint) support to BTCPay Server. This enables external services to verify that a Lightning invoice created via LNURL-pay (LUD-06) has been settled, without needing direct access to the store's Lightning node.

## Problem

BTCPay Server already supports LNURL-pay (LUD-06) for creating invoices via Lightning Address and LNURL endpoints. However, it does not expose the LUD-21 `verify` endpoint. This means:

1. External services (wallet aggregators, point-of-sale apps, tipping services) cannot independently verify payment status after directing a user to pay via LNURL.
2. The only way to check payment status is through the BTCPay API with authentication, which requires sharing API keys.
3. Services that use BTCPay as the LNURL backend for Lightning Addresses cannot offer payment verification to their users without polling the authenticated invoice API.

### 1. Add Verify Endpoint to UILNURLController

New endpoint: `GET /lnurlp/{username}/verify/{payment_hash}`

```
Response (settled):
{ "status": "OK", "settled": true, "preimage": "<hex>", "pr": "<bolt11>" }

Response (unsettled):
{ "status": "OK", "settled": false, "preimage": null, "pr": "<bolt11>" }

Response (not found):
{ "status": "ERROR", "reason": "Not found" }
```

This endpoint is unauthenticated per the LUD-21 spec - it only reveals whether a specific payment hash was settled, not any other store data.

### 2. Include `verify` URL in LNURL-pay Callback Response

In the existing callback handler (`UILNURLController.cs`), add the `verify` field to `LNURLPayRequestCallbackResponse`:

```csharp
return Ok(new LNURLPayRequest.LNURLPayRequestCallbackResponse
{
    Disposable = true,
    Routes = Array.Empty<string>(),
    Pr = lightningPaymentMethod.Destination,
    SuccessAction = successAction,
    Verify = _linkGenerator.GetUriByAction(
        "LnurlPayVerify", "UILNURL",
        new { username, paymentHash = promptDetails.PaymentHash?.ToString() },
        Request.Scheme, Request.Host, Request.PathBase)
});
```

### 3. Add Toggle to LNURLPaymentMethodConfig

Add `LUD21Enabled` boolean following the existing `LUD12Enabled` pattern:

```csharp
[JsonProperty("lud21Enabled")]
public bool LUD21Enabled { get; set; } = true;
```

### 4. Implementation Details

The verify endpoint needs to:
1. Look up the Lightning Address username to find the store
2. Query the invoice repository for an invoice with the matching payment hash
3. Check the Lightning node for settlement status via the existing `ILightningClient`
4. Return the appropriate LUD-21 response

Key files to modify:
- `BTCPayServer/Controllers/UILNURLController.cs` - Add verify endpoint and include verify URL in callback
- `BTCPayServer/Payments/LNURLPay/LNURLPaymentMethodConfig.cs` - Add `LUD21Enabled` toggle
- `BTCPayServer/Payments/LNURLPay/LNURLPayPaymentMethodDetails.cs` - Ensure `PaymentHash` is persisted (already is)

The `LNURLPayRequestCallbackResponse` class in the LNURL NuGet package may not have a `Verify` property yet. If not, it could be contributed upstream or the response class extended locally.

### 5. Security Considerations

- The verify endpoint is public (no auth) per LUD-21 spec. This is intentional - it only exposes boolean settlement status and preimage for a specific payment hash.
- Payment hashes are cryptographically random (32 bytes) so the endpoint is not enumerable.
- Rate limiting should be applied to prevent abuse.
- The preimage is only revealed after settlement (standard Lightning behavior).

## Scope

- Core BTCPay Server change (not a plugin)
- Extends existing `UILNURLController`
- Minimal new code (~50-80 lines)
- No database schema changes
- No breaking changes to existing LNURL-pay flow
- Backwards compatible (`verify` URL is optional in the response)

With LUD-21 in BTCPay Server, great wallets like [agi.cash](https://agi.cash) would be able to verify Lightning payments against any BTCPay instance without needing a single API key - just pure, unauthenticated LNURL magic.

r1ckstardev:
## Summary

Implements [LUD-21](https://github.com/lnurl/luds/blob/luds/21.md) (LNURL-pay verify endpoint) for Lightning Address payments. This enables external services to verify that a Lightning invoice created via LNURL-pay has been settled, without needing direct access to the store's Lightning node or any API keys.

## Changes

- **New verify endpoint**: `GET /lnurlp/{username}/verify/{paymentHash}` - unauthenticated endpoint per LUD-21 spec that returns settlement status and preimage
- **Callback response**: LNURL-pay callback responses now include a `verify` URL when the payment was initiated via Lightning Address and LUD-21 is enabled
- **Search indexing**: Payment hashes are indexed in `InvoiceSearchData` for efficient lookup by the verify endpoint
- **LUD21Enabled toggle**: New boolean config option on `LNURLPaymentMethodConfig` (defaults to `true`), following the existing `LUD12Enabled` pattern
- **Settings UI**: Toggle added to Lightning Settings page under LNURL section
- **Swagger**: API docs updated with new `lud21Enabled` property

## Verify endpoint response format

```json
// Settled
{ "status": "OK", "settled": true, "preimage": "<hex>", "pr": "<bolt11>" }

// Not yet settled
{ "status": "OK", "settled": false, "preimage": null, "pr": "<bolt11>" }

// Not found
{ "status": "ERROR", "reason": "Not found" }
```

## Security

- Endpoint is intentionally unauthenticated per LUD-21 spec
- Only exposes boolean settlement status and preimage for a specific payment hash
- Payment hashes are cryptographically random (32 bytes) so the endpoint is not enumerable
- Scoped to a specific Lightning Address username (resolved to store)
- Can be disabled per-store via the LUD21Enabled toggle

## Motivation

With LUD-21 in BTCPay Server, wallets like [agi.cash](https://agi.cash) can verify Lightning payments against any BTCPay instance without needing a single API key - just pure, unauthenticated LNURL verification.

Closes #7248

### DIFF
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -462,15 +462,14 @@ namespace BTCPayServer
         }
 
         /// <summary>
-        /// LUD-21: Verify payment status for a Lightning Address payment.
+        /// LUD-21: Verify payment status for an LNURL payment.
         /// Returns settlement status and preimage for a given payment hash.
         /// </summary>
-        [HttpGet("~/lnurlp/{username}/verify/{paymentHash}")]
+        [HttpGet("~/lnurlp/verify/{paymentHash}")]
         [EnableCors(CorsPolicies.All)]
-        [IgnoreAntiforgeryToken]
-        public async Task<IActionResult> LnurlPayVerify(string username, string paymentHash)
+        public async Task<IActionResult> LnurlPayVerify(string paymentHash)
         {
-            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(paymentHash))
+            if (string.IsNullOrEmpty(paymentHash))
                 return NotFound();
 
             // A Lightning payment hash is exactly 32 bytes / 64 hex chars.
@@ -480,23 +479,22 @@ namespace BTCPayServer
 
             paymentHash = paymentHash.ToLowerInvariant();
 
-            var lightningAddressSettings = await _lightningAddressService.ResolveByAddress(username);
-            if (lightningAddressSettings is null)
-                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
-
-            var store = await _storeRepository.FindStore(lightningAddressSettings.StoreDataId);
-            if (store is null)
-                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
-
             var cryptoCode = "BTC";
-            var pmi = GetLNUrlPaymentMethodId(cryptoCode, store, out var lnUrlMethod);
-            if (pmi is null || !lnUrlMethod.LUD21Enabled)
-                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not available" });
+            var pmi = PaymentTypes.LNURL.GetPaymentMethodId(cryptoCode);
 
             // Find invoice by payment hash via AddressInvoices index
             var invoice = await _invoiceRepository.GetInvoiceFromAddress(pmi, paymentHash);
-            if (invoice is null || invoice.StoreId != store.Id)
+            if (invoice is null)
                 return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
+
+            var store = await _storeRepository.FindStore(invoice.StoreId);
+            if (store is null)
+                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
+
+            var lnUrlPmi = GetLNUrlPaymentMethodId(cryptoCode, store, out var lnUrlMethod);
+            if (lnUrlPmi is null || !lnUrlMethod.LUD21Enabled)
+                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
+
             var prompt = invoice.GetPaymentPrompt(pmi);
             if (prompt is null)
                 return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
@@ -504,8 +502,7 @@ namespace BTCPayServer
             var handler = (LNURLPayPaymentHandler)_handlers[pmi];
             var details = handler.ParsePaymentPromptDetails(prompt.Details);
 
-            // Verify the payment hash matches (case-insensitive for hex)
-            if (!string.Equals(details.PaymentHash?.ToString(), paymentHash, StringComparison.OrdinalIgnoreCase))
+            if (!string.Equals(details.PaymentHash?.ToString(), paymentHash, StringComparison.Ordinal))
                 return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
 
             var settled = invoice.Status == InvoiceStatus.Settled ||
@@ -885,19 +882,14 @@ namespace BTCPayServer
                     SuccessAction = successAction
                 });
 
-                // LUD-21: Add verify URL if enabled and payment was via Lightning Address
+                // LUD-21: Add verify URL if enabled
                 if (lnurlSupportedPaymentMethod.LUD21Enabled &&
-                    promptDetails.ConsumedLightningAddress is not null &&
                     promptDetails.PaymentHash is not null)
                 {
-                    var username = promptDetails.ConsumedLightningAddress.Split('@').FirstOrDefault();
-                    if (username is not null)
-                    {
-                        callbackResponse["verify"] = _linkGenerator.GetUriByAction(
-                            nameof(LnurlPayVerify), "UILNURL",
-                            new { username, paymentHash = promptDetails.PaymentHash.ToString().ToLowerInvariant() },
-                            Request.Scheme, Request.Host, Request.PathBase);
-                    }
+                    callbackResponse["verify"] = _linkGenerator.GetUriByAction(
+                        nameof(LnurlPayVerify), "UILNURL",
+                        new { paymentHash = promptDetails.PaymentHash.ToString().ToLowerInvariant() },
+                        Request.Scheme, Request.Host, Request.PathBase);
                 }
 
                 return Ok(callbackResponse);
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -506,8 +506,8 @@ namespace BTCPayServer
             var handler = (LNURLPayPaymentHandler)_handlers[pmi];
             var details = handler.ParsePaymentPromptDetails(prompt.Details);
 
-            // Verify the payment hash matches
-            if (details.PaymentHash?.ToString() != paymentHash)
+            // Verify the payment hash matches (case-insensitive for hex)
+            if (!string.Equals(details.PaymentHash?.ToString(), paymentHash, StringComparison.OrdinalIgnoreCase))
                 return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
 
             var settled = invoice.Status == InvoiceStatus.Settled ||
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -39,7 +39,6 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Localization;
 using NBitcoin;
-using NicolasDorier.RateLimits;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using LightningAddressData = BTCPayServer.Data.LightningAddressData;
@@ -469,7 +468,6 @@ namespace BTCPayServer
         [HttpGet("~/lnurlp/{username}/verify/{paymentHash}")]
         [EnableCors(CorsPolicies.All)]
         [IgnoreAntiforgeryToken]
-        [RateLimitsFilter(ZoneLimits.Verify, Scope = RateLimitsScope.RemoteAddress)]
         public async Task<IActionResult> LnurlPayVerify(string username, string paymentHash)
         {
             if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(paymentHash))
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -868,15 +868,13 @@ namespace BTCPayServer
 
                 if (updatePaymentMethod)
                 {
-                    await _invoiceRepository.UpdatePrompt(invoiceId, lightningPaymentMethod);
+                    // Index payment hash for LUD-21 verify lookup via AddressInvoices,
+                    // flushed atomically with the prompt update.
+                    var trackedDestinations = promptDetails.PaymentHash is not null
+                        ? new[] { promptDetails.PaymentHash.ToString().ToLowerInvariant() }
+                        : null;
+                    await _invoiceRepository.UpdatePrompt(invoiceId, lightningPaymentMethod, trackedDestinations);
                     _eventAggregator.Publish(new InvoiceNewPaymentDetailsEvent(invoiceId, promptDetails, pmi));
-
-                    // Index payment hash for LUD-21 verify lookup via AddressInvoices
-                    if (promptDetails.PaymentHash is not null)
-                    {
-                        await _invoiceRepository.AddAddressInvoice(invoiceId, pmi,
-                            promptDetails.PaymentHash.ToString().ToLowerInvariant());
-                    }
                 }
 
                 var callbackResponse = JObject.FromObject(new LNURLPayRequest.LNURLPayRequestCallbackResponse
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -482,11 +482,11 @@ namespace BTCPayServer
 
             var lightningAddressSettings = await _lightningAddressService.ResolveByAddress(username);
             if (lightningAddressSettings is null)
-                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Unknown username" });
+                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
 
             var store = await _storeRepository.FindStore(lightningAddressSettings.StoreDataId);
             if (store is null)
-                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Unknown username" });
+                return NotFound(new LNUrlStatusResponse { Status = "ERROR", Reason = "Not found" });
 
             var cryptoCode = "BTC";
             var pmi = GetLNUrlPaymentMethodId(cryptoCode, store, out var lnUrlMethod);
--- a/BTCPayServer/Controllers/UILNURLController.cs
+++ b/BTCPayServer/Controllers/UILNURLController.cs
@@ -39,6 +39,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Localization;
 using NBitcoin;
+using NicolasDorier.RateLimits;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using LightningAddressData = BTCPayServer.Data.LightningAddressData;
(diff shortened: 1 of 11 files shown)


NicolasDorier:
1. You need to add tests.
2. Using `AdditionalSearchTerms` for finding an invoice can be abused. This property is literally meant for users who want to search an invoice in a list (UI), it's not meant for building your logic on top of.

Instead of `AdditionalSearchTerms`, you can add use `context.TrackedDestinations.Add(paymentHash)` in the `LightningLikePaymentHandler` when the Lud 21 is enabled. Then just use `InvoiceRepository.GetInvoiceFromAddress` to fetch the invoice per paymentHash. (don't forget to double check that the returned invoice's storeId match the lnaddress's store id)

r1ckstardev:
@NicolasDorier @coderabbitai - all feedback addressed, ready for another look.

**Nicolas**
- Tests added: `CanUseLUD21VerifyEndpoint` in `BTCPayServer.Tests/LightningTests.cs` covers the unknown-payment-hash case, the unknown-username case, the happy-path verify on a fresh LNURL invoice (status / settled / preimage / pr), and the cross-store isolation case where store1's payment hash cannot be resolved through store2's username.
- `AdditionalSearchTerms` removed. The verify endpoint now uses `InvoiceRepository.GetInvoiceFromAddress(pmi, paymentHash)` against the `AddressInvoices` index, and `LightningLikePaymentHandler` registers the payment hash via `context.TrackedDestinations.Add(...)` when LUD-21 is enabled, exactly as you suggested.
- The post-lookup `invoice.StoreId != store.Id` check is in place at `UILNURLController.cs:498`, and the new test exercises that specific branch (it asserts both the 404 status and the response body's `reason == "Not found"`, which can only come from the post-lookup branch, not from the earlier feature-gate).

**CodeRabbit**
- Case-insensitive payment hash comparison: addressed in dd3626f via `StringComparison.OrdinalIgnoreCase` plus `ToLowerInvariant()` normalization at the entry of `LnurlPayVerify`.
- `paymentHash` validation before lookup: 64-hex sanity check added at the top of `LnurlPayVerify` (515a693), garbage input is rejected with `Not found` before any DB hit.
- Standardize not-found reasons: `Unknown username` collapsed to `Not found` (1311e33). The LUD-21-disabled branch intentionally keeps `Not available` because that's a distinct operational condition (feature off vs resource missing), and the new test relies on the distinction to prove it actually reached the store-isolation branch.
- `AddAddressInvoice` idempotency: handled in dd3626f via a check-then-insert-or-update pattern on `(Address, PaymentMethodId)`, so duplicate inserts no longer crash on the unique key.

Test plan locally: build clean, `BTCPayServer.Tests` LUD-21 verify test exercises all four branches (unknown hash, unknown username, happy path, cross-store isolation).

r1ckstardev:
@coderabbitai - addressed in 60f4007.

**1. TOCTOU race in `AddAddressInvoice`** - fixed. The insert path now catches `DbUpdateException` and swallows the unique-key violation, since both concurrent writers are inserting identical `(Address, PaymentMethodId)` rows. The update branch keeps its own `SaveChangesAsync` so any real failure there still propagates.

**2. Fragile null-preimage assertion** - fixed. Replaced with `JTokenType.Null` comparison so the assertion is robust regardless of whether the global serializer omits nulls or emits them as JSON `null`.

**3. `LUD21Enabled` gating for `AddAddressInvoice` / `TrackedDestinations.Add`** - intentionally deferred. Both call sites are in code paths that already require LNURL setup; the cost of an extra `AddressInvoices` row when LUD-21 is off is a single small DB write per invoice and not on a hot path. Adding the gate inside `LightningLikePaymentHandler` would mean threading the store's `LNURLPaymentMethodConfig` lookup into the payment handler, which is more invasive than the saving justifies. Happy to revisit if a maintainer wants it in this PR.

Also, separately: pushed 8818d7b which fixes the actual integration-test failure on this PR. `CanUseLUD21VerifyEndpoint` was calling `/lnurlp/{username}` to fetch the LN address, but the LUD-16 resolver lives at `/.well-known/lnurlp/{username}` (`UILNURLController.cs:404`). The bare path has no route, so the test 404'd at line 995 before any of the verify logic could run. Wrong route since the test was first added in d91d6b8, not a regression.

r1ckstardev:
dcfdf88 refactors the LUD-21 payment-hash indexing to address the architectural concern raised earlier: the LNURL callback was doing `UpdatePrompt` + a separate `AddAddressInvoice` as two repository round-trips, which is the compensation pattern for not having a flush-aware overload on the simple prompt path (`NewPaymentPrompt(PaymentMethodContext)` does flush `TrackedDestinations`, but constructing a full `PaymentMethodContext` inside a controller just to reach that flush is heavier and more fragile than the alternative).

The fix adds an optional `IEnumerable<string> trackedDestinations = null` parameter to `UpdatePrompt`. When supplied, the `AddressInvoices` rows are inserted inside the same `DbContext` / `SaveChangesAsync` as the prompt-blob update, matching the idempotency pattern already used by `AddAddressInvoice` (check-then-insert with a `DbUpdateException` swallow for concurrent writers inserting the same row). The two existing callers (`LightningListener`, `NBXplorerListener`) are bit-identical because the new param defaults to `null`. Only the LNURL callback passes the payment hash, and it only does so inside the existing `if (updatePaymentMethod)` gate, so the flush fires only on real state transitions.

@coderabbitai please take another look - does this cleanly close the #2/#3 items from your previous review, or do you see a sharper angle (catch-breadth on `DbUpdateException`, race semantics, anything else)? Happy to tighten if you spot something.

r1ckstardev:
`b237809` adds three assertions inside the existing `CanLNURLPayVerify` integration test to illustrate and guard the `UpdatePrompt(trackedDestinations)` refactor without new infrastructure:

1. **Repeat LNURL callback with the SAME amount.** Exercises the idempotent flush branch - the `AddressInvoices` row for the payment hash already exists, so the flush loop takes the `existing is not null` path. Second callback must return 200 and the verify endpoint must still resolve the hash afterward (no duplicate row, no throw).
2. **Repeat LNURL callback with a DIFFERENT amount.** A new payment hash is minted, the new verify URL must resolve - proves the new hash is indexed in `AddressInvoices` and the prompt blob transition persists through the split-SaveChanges fix from `11a26fb`.
3. **Malformed `paymentHash` (non-hex / wrong length) on verify.** Must return 404 with `Reason: Not found` at the 64-hex guard before any DB lookup. Exercises the validation added in `515a693`.

Full PR commit series on this branch now:
- `60f4007` TOCTOU fix in `AddAddressInvoice` + preimage assertion hardening
- `dcfdf88` refactor: flush tracked payment hash via `UpdatePrompt` overload
- `11a26fb` fix: split SaveChanges, exclude `DbUpdateConcurrencyException` from inner catch
- `b237809` test: repeat-callback idempotency + hash validation coverage

Awaiting your final approval @coderabbitai.

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -461,6 +461,66 @@ public async Task<IActionResult> ResolveLightningAddress(string username)
             return Ok(lnurlRequest);
         }
 
+        /// <summary>
+        /// LUD-21: Verify payment status for a Lightning Address payment.
+        /// Returns settlement status and preimage for a given payment hash.
+        /// </summary>
+        [HttpGet("~/lnurlp/{username}/verify/{paymentHash}")]
```
NicolasDorier: Can you simplify the url by not including the `username`?
I believe you don't need it, you have the `paymentHash`, so you can get the invoice and from this you have the `storeId`. (in the current code you only need the username to get the `storeid`)

It would also simplify the test so you don't need to check what happen if username/paymenthash doesn't match.

On top of this, this will make it possible to support it for all lnurl, not only lnaddress

NicolasDorier (changes requested):

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -808,17 +868,39 @@ public async Task<IActionResult> GetLNURLForInvoice(string invoiceId, string cry
...
                     Pr = lightningPaymentMethod.Destination,
                     SuccessAction = successAction
                 });
+
+                // LUD-21: Add verify URL if enabled and payment was via Lightning Address
+                if (lnurlSupportedPaymentMethod.LUD21Enabled &&
+                    promptDetails.ConsumedLightningAddress is not null &&
+                    promptDetails.PaymentHash is not null)
+                {
+                    var username = promptDetails.ConsumedLightningAddress.Split('@').FirstOrDefault();
```
NicolasDorier: This code to fetch the username becomes uneeded if you remove the username from the route.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -346,7 +379,40 @@ public async Task UpdatePrompt(string invoiceId, PaymentPrompt prompt)
...
+                    // AddressInvoices must not be able to roll this back.
                     await context.SaveChangesAsync();
+
+                    if (trackedDestinations is not null)
+                    {
+                        var pmi = prompt.PaymentMethodId.ToString();
+                        foreach (var tracked in trackedDestinations)
+                        {
+                            var trackedLocal = tracked;
+                            var existing = await context.AddressInvoices
```
NicolasDorier: Can you do an upsert instead?
The code will be smaller requiring only 1 line of SQL rather than ~20 lines of Entity Framework.
Also, it will require only a single rountrip rather than N*2 roundtrip.

It would also fix an edge case where if only some of the `trackedDestinations` are dups, the others would not get inserted properly. (When an exception such as DbUpdateException is thrown in EF, the dbcontext ends in a weird state causing subsequent `.SaveChanges()` to also fail)

```
INSERT INTO ... ()
ON CONFLICT DO NOTHING
```

NicolasDorier (changes requested):

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -461,6 +461,66 @@ public async Task<IActionResult> ResolveLightningAddress(string username)
             return Ok(lnurlRequest);
         }
 
+        /// <summary>
+        /// LUD-21: Verify payment status for a Lightning Address payment.
+        /// Returns settlement status and preimage for a given payment hash.
+        /// </summary>
+        [HttpGet("~/lnurlp/{username}/verify/{paymentHash}")]
+        [EnableCors(CorsPolicies.All)]
+        [IgnoreAntiforgeryToken]
```
NicolasDorier: `IgnoreAntiforgeryToken` should be needed on HttpGet routes

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -142,6 +138,12 @@ public async Task ConfigurePrompt(PaymentMethodContext context)
                 NodeInfo = (await nodeInfo).FirstOrDefault()?.ToString()
             };
             paymentPrompt.Details = JObject.FromObject(details, Serializer);
+
+            // Track payment hash for LUD-21 verify lookup
```
NicolasDorier: If it is not enabled, and not created by LNURL, do we need to track the paymenthash?

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
```diff
@@ -142,6 +138,12 @@ public async Task ConfigurePrompt(PaymentMethodContext context)
...
+            if (details.PaymentHash is not null)
+            {
+                context.TrackedDestinations.Add(details.PaymentHash.ToString().ToLowerInvariant());
```
NicolasDorier: Actually, shouldn't that be in `LNURLPayPaymentHandler` instead?

BTCPayServer/Controllers/UILNURLController.cs
r1ckstardev: Done in 3016e38. Removed `username` from the route entirely — now `~/lnurlp/verify/{paymentHash}`. The endpoint looks up the invoice by payment hash, gets the store from the invoice, then checks LUD-21 is enabled on that store. Tests simplified accordingly (cross-store-via-username and unknown-username checks dropped since they no longer apply).

BTCPayServer/Controllers/UILNURLController.cs
r1ckstardev: Removed in 3016e38. The verify URL is now generated without needing the username at all — just the payment hash.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
r1ckstardev: Done in 3016e38. Replaced all EF check-then-act patterns with `INSERT ... ON CONFLICT DO NOTHING` — applies to `AddAddressInvoice`, `UpdatePrompt` tracked destinations, and `NewPaymentPrompt` tracked destinations. Single SQL roundtrip each, no partial-dup DbContext corruption risk.

BTCPayServer/Controllers/UILNURLController.cs
r1ckstardev: Removed in 3016e38.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
r1ckstardev: Right — removed the tracking from `LightningLikePaymentHandler` in 3016e38. Regular LN invoices don't need it; the LNURL path already handles tracking via `UpdatePrompt(trackedDestinations)` in the callback.

BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs
r1ckstardev: Removed from `LightningLikePaymentHandler` entirely in 3016e38. The LNURL callback in `UILNURLController` already indexes the payment hash via `UpdatePrompt(trackedDestinations)` after creating the bolt11, so `LNURLPayPaymentHandler` doesn't need to duplicate it.

r1ckstardev:
@coderabbitai review

rockstardev:
@coderabbitai, is there anything else to address in this PR or it's good to merge in your opinion? Referencing previous things you and Nicolas raised.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
```diff
@@ -346,7 +353,21 @@ public async Task UpdatePrompt(string invoiceId, PaymentPrompt prompt)
...
                     await context.SaveChangesAsync();
+
+                    if (trackedDestinations is not null)
+                    {
+                        var pmi = prompt.PaymentMethodId.ToString();
+                        foreach (var tracked in trackedDestinations)
+                        {
+                            await context.Database.ExecuteSqlInterpolatedAsync(
+                                $"""INSERT INTO "AddressInvoices" ("Address", "PaymentMethodId", "InvoiceDataId") VALUES ({tracked}, {pmi}, {invoiceId}) ON CONFLICT ("Address", "PaymentMethodId") DO NOTHING""");
+                        }
```
NicolasDorier, line 373: This code is exactly the same as  https://github.com/btcpayserver/btcpayserver/pull/7250/changes#diff-c545b4806714f17ef9b042626a9db7614f96ba9abac22993a00f7ebc7d6398f9R395

https://github.com/btcpayserver/btcpayserver/pull/7250/changes#diff-c545b4806714f17ef9b042626a9db7614f96ba9abac22993a00f7ebc7d6398f9R83

Can you refactor it to avoid duplication?

BTCPayServer/Controllers/UILNURLController.cs
```diff
@@ -808,17 +865,34 @@ public async Task<IActionResult> GetLNURLForInvoice(string invoiceId, string cry
 
                 if (updatePaymentMethod)
                 {
-                    await _invoiceRepository.UpdatePrompt(invoiceId, lightningPaymentMethod);
+                    // Index payment hash for LUD-21 verify lookup via AddressInvoices,
+                    // flushed atomically with the prompt update.
+                    var trackedDestinations = promptDetails.PaymentHash is not null
```
NicolasDorier: The trackedDestinations should be null if LUD21 is not enabled.

NicolasDorier (changes requested):

BTCPayServer/Controllers/UILNURLController.cs
r1ckstardev: Fixed in 23efe7f - added `lnurlSupportedPaymentMethod.LUD21Enabled` guard so `trackedDestinations` is null when LUD-21 is disabled.

BTCPayServer/Services/Invoices/InvoiceRepository.cs
r1ckstardev: Refactored in 23efe7f - extracted a `private static UpsertAddressInvoice` helper. All three call sites (`AddAddressInvoice`, `UpdatePrompt`, `NewPaymentPrompt`) now go through it.

PR #7258, opened 2026-03-22
Subscriptions: Allow upgrade/downgrade at period end

TChukwuleta:
Hey @NicolasDorier 

Closes [#7147](https://github.com/btcpayserver/btcpayserver/discussions/7147)

Adds support for scheduling subscription plan changes to take effect at the end of the current billing period, rather than immediately. This mirrors the behavior described in Stripe's customer portal and addresses the two sub-features requested in the discussion:
Subscribers can choose to change plans immediately or at period end
Admins can configure the timing behavior per plan transition

I wrote a query locally to update period_end and grace_period_end in subs_subscribers in order to test the plan upgrade/downgrade with respect to end period

### DIFF
--- a/BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
+++ b/BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
@@ -259,7 +259,6 @@ public class SubscriptionHostedService(
                             m.StartNextPlan(now);
                             subCtx.AddEvent(new SubscriptionEvent.PlanStarted(m, planBefore)
                             {
-                                PreviousPlan = planBefore,
                                 AutoRenew = planBefore.Id == m.PlanId
                             });
                         }
@@ -272,13 +271,11 @@ public class SubscriptionHostedService(
 
                 if (newPhase is PhaseTypes.Expired or PhaseTypes.Grace && m is { NewPlan: not null, NewPlanId: not null } && m.NewPlanId != m.PlanId)
                 {
-                    var prevPlanId = m.PlanId;
                     var prevPlan = m.Plan;
                     (m.PlanId, m.Plan) = (m.NewPlanId, m.NewPlan);
                     (m.NewPlanId, m.NewPlan) = (null, null);
                     subCtx.AddEvent(new SubscriptionEvent.PlanStarted(m, prevPlan)
                     {
-                        PreviousPlan = prevPlan,
                         AutoRenew = false
                     });
                 }
@@ -504,10 +501,7 @@ public class SubscriptionHostedService(
             }
         }
         if (checkout.PlanStarted)
-            subCtx.AddEvent(new SubscriptionEvent.PlanStarted(sub, prevPlan)
-            {
-                PreviousPlan = prevPlan
-            });
+            subCtx.AddEvent(new SubscriptionEvent.PlanStarted(sub, prevPlan));
     }
 
     record MoveTimeRequest(MemberSelector MemberSelector, TimeSpan Period);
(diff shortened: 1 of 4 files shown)


BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
```diff
@@ -270,14 +271,21 @@ private async Task UpdateSubscriptionStates(SubscriptionContext subCtx, MemberSe
                     }
                 }
 
+                if (newPhase is PhaseTypes.Expired or PhaseTypes.Grace && m is { NewPlan: not null, NewPlanId: not null } && m.NewPlanId != m.PlanId)
+                {
+                    var prevPlan = m.Plan;
+                    (m.PlanId, m.Plan) = (m.NewPlanId, m.NewPlan);
+                    (m.NewPlanId, m.NewPlan) = (null, null);
+                    subCtx.AddEvent(new SubscriptionEvent.PlanStarted(m, prevPlan)
```
NicolasDorier, line 277: 1. `new SubscriptionEvent.PlanStarted(m, prevPlan)` you are passing `prevPlan` instead of `m.Plan` I believe.
2. Are you sure you need to add `AddEvent` here? It wasn't there before, was it a bug?

BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
```diff
@@ -651,10 +659,18 @@ public class UpdateCreditParameters
...
+                .FirstOrDefault(pc => pc.PlanId == portal.Subscriber.PlanId && pc.PlanChangeId == planId);
+            if (planChangeRecord == null)
                 return null;
+
+            if (planChangeRecord.Timing == PlanChangeData.ChangeTiming.AtPeriodEnd)
+            {
+                portal.Subscriber.NewPlanId = planId;
+                portal.Subscriber.NewPlan = plan;
+                await ctx.SaveChangesAsync();
+                return ScheduledResult;
```
NicolasDorier: Don't return a magic string here. Caller expect a valid `checkoutId` or null.

When you want to do this kind of change, you should change the return type. Then create a `discriminated union` with records in .NET. (check the type `AddOrUpdateStoreUserResult` for example of what it looks like)

The caller can then use pattern matching to decide what to do with the result.

NicolasDorier (changes requested):
I've seen two small change needed.

On top of this:
* You need to add a test.
* Update swagger since you updated SubscriberModel

I am also concerned about the `PlanStarted` event you added. Either we had a bug before that the tests didn't catch, or we are now triggering the event twice. Can you check?

TChukwuleta:
Hey Emperor I have made the changes and also included test for it

Ou7law007:
That's super nice.

In the portal, when the user tries to upgrade/downgrade, there's a pop-up that shows up to confirm the change. Currently, regardless of if it's 'immediate' or 'at period end', the pop looks exactly the same. Could you make it clear in the pop-up exactly what's going to happen once the user clicks confirm? Maybe add some extra text that explains what's going to happen exactly.

image

image

Immediate changes specifically are more 'dangerous', I would say, so maybe they deserve an extra highlighting in red or something idk.

Also maybe make it clear from the beginning so that the user doesn't have to click each plan to know if that change is immediate or at period end:

image

Cheap suggestion:
image

But I would say, the pop-ups definitely need changing.

Ou7law007:
Also, is it possible to allow both?

Unless I'm missing something, I don't see a reason why a plan change should be restricted to either immediate or at period end and not both, if the admin wants to allow this.

If this is done, then once both are allowed, the portal also needs updating to allow the user to choose which kind of change they want to make. Maybe 2 buttons for plan changes that allow both or just 1 choice/select field called "Change kind" or something with a button "Upgrade" or "Downgrade" to make that change.

PR #7263, opened 2026-03-23
lightning: support amountless bolt11 for top-up invoices

bigg-bb:
## Summary

This PR enables the BTC-LN payment method to work on **top-up (open-amount) invoices** by attempting to create an amountless BOLT11 invoice instead of immediately throwing `PaymentMethodUnavailableException`.

Depends on: btcpayserver/BTCPayServer.Lightning#175 (already merged, published as `BTCPayServer.Lightning.All` v1.6.14)

---

### `BTCPayServer/BTCPayServer.csproj`
- Bump `BTCPayServer.Lightning.All` from `1.6.13` → `1.6.14`

### `BTCPayServer/Payments/Lightning/LightningLikePaymentHandler.cs`

**Before:**
```csharp
if (context.InvoiceEntity.Type == InvoiceType.TopUp)
{
    throw new PaymentMethodUnavailableException("Lightning Network payment method is not available for top-up invoices");
}
```

**After:**
```csharp
                    ....
                    var isTopUpInvoice = invoice.Type == InvoiceType.TopUp;
                    ...
                    ...
                    // For top-up invoices, pass LightMoney.Zero to request an amountless bolt11.
                    // The backend decides whether it supports this
                    var invoiceAmount = isTopUpInvoice ? LightMoney.Zero : new LightMoney(due, LightMoneyUnit.BTC);
                    var request = new CreateInvoiceParams(invoiceAmount, description, expiry);
                    request.PrivateRouteHints = storeBlob.LightningPrivateRouteHints;
                    lightningInvoice = await client.CreateInvoice(request, cts.Token);

                    // Since No Invoice Amount so Skip the Tweak Fee check
                    if (!isTopUpInvoice)
                    {
                        var diff = request.Amount - lightningInvoice.Amount;
                        if (diff != LightMoney.Zero)
                        {
                            // Some providers doesn't round up to msat. So we tweak the fees so the due match the BOLT11's amount.
                            paymentPrompt.AddTweakFee(-diff.ToUnit(LightMoneyUnit.BTC));
                        }
                    }
```

The amount-diff fee-tweak is also skipped for top-up invoices since there is no fixed due amount to reconcile.

---

## How It Works

Instead of hard-blocking all lightning backends for top-up invoices, we now attempt the call with `LightMoney.Zero`. The backend itself decides whether it can create an amountless invoice:

- **LND** — natively supports amountless bolt11. When `value_msat` is omitted (or zero), LND skips adding the amount tag to the bolt11 (see [`addinvoice.go:340`](https://github.com/lightningnetwork/lnd/blob/v0.20.0-beta/lnrpc/invoicesrpc/addinvoice.go#L340)). The BTC-LN tab appears on the checkout page.
- **CLN, Eclair, Phoenixd, and others** — if they reject a zero/amountless invoice, the exception is caught and re-thrown as `PaymentMethodUnavailableException`, causing BTC-LN to be silently excluded from the invoice — exactly the same behaviour as today. Nothing changes for these backends.

This approach is forward-compatible: if another backend adds amountless invoice support in the future, it will work automatically without any changes here.

---

## Settlement — No Breaking Changes

When a payer settles an amountless invoice, LND reports:
- `value_msat = "0"` (the encoded invoice amount — zero for amountless)
- `amt_paid_msat = "5000"` (the actual amount the payer sent)

In `LightningListener.cs`, `ConvertLndInvoice` maps these to:
- `notification.Amount = LightMoney(0)` — **not null**, so the `notification.Amount != null` guard at line 552 passes correctly
- `notification.AmountReceived = LightMoney(5000)` — the actual paid amount

In `AddPayment` at line 634:
```csharp
Amount = (notification.AmountReceived ?? notification.Amount).ToDecimal(LightMoneyUnit.BTC)
```
`AmountReceived` is used (not the zero amount), so the correct paid amount is recorded. For top-up invoices there is no fixed due amount — any payment settles it.

**No changes are needed in `LightningListener.cs`.**

---

## What Does Not Change

- On-chain (BTC-CHAIN) and LNURL (BTC-LNURL) behaviour for top-up invoices is completely unchanged
- All non-top-up (standard) invoice flows are completely unchanged
- No changes to `ILightningClient`, `CreateInvoiceParams`, or any other backend client
- The existing `PaymentMethodUnavailableException` fallback mechanism is preserved

We'd appreciate a review of this change. Happy to address any feedback. Thank you!

### DIFF
--- a/BTCPayServer.Tests/PlaywrightTests.cs
+++ b/BTCPayServer.Tests/PlaywrightTests.cs
@@ -356,10 +356,20 @@ namespace BTCPayServer.Tests
             // Topup Invoice test
             var i = await s.CreateInvoice(storeId, null, cryptoCode);
             await s.GoToInvoiceCheckout(i);
+
+
+            async Task<BOLT11PaymentRequest> AssertBolt11(LightMoney expectedValue = null)
+            {
+                var bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
+                Assert.NotNull(bolt11);
+                var b = BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network);
+                if (expectedValue is not null)
+                    Assert.Equal(expectedValue, b.MinimumAmount);
+                return b;
+            }
+
             // Top-up invoices now display a zero-amount BOLT11 on checkout, like standard invoices.
-            var bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
-            Assert.NotNull(bolt11);
-            BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network);
+            await AssertBolt11(LightMoney.Zero);
             var invoiceId = s.Page.Url.Split('/').Last();
             LNURLPayRequest fetchedRequest;
             using (var resp = await s.Server.PayTester.HttpClient.GetAsync("BTC/lnurl/pay/i/" + invoiceId))
@@ -400,8 +410,7 @@ namespace BTCPayServer.Tests
             i = await s.CreateInvoice(storeId, 0.0000001m, cryptoCode);
             await s.GoToInvoiceCheckout(i);
             // BOLT11 is also displayed for standard invoice (not LNURL, even if it is available)
-            bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
-            BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network);
+            await AssertBolt11(LightMoney.Coins(0.0000001m));
             invoiceId = s.Page.Url.Split('/').Last();
             using (var resp = await s.Server.PayTester.HttpClient.GetAsync("BTC/lnurl/pay/i/" + invoiceId))
             {
@@ -446,9 +455,7 @@ namespace BTCPayServer.Tests
 
             i = await s.CreateInvoice(storeId, null, cryptoCode);
             await s.GoToInvoiceCheckout(i);
-            bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
-            Assert.NotNull(bolt11);
-            BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network);
+            await AssertBolt11();
             paymentMethods = await greenfield.GetInvoicePaymentMethods(storeId, i);
             lnurlMethod = Assert.Single(paymentMethods, p => p.PaymentMethodId == "BTC-LNURL");
             var lnurl = lnurlMethod.PaymentLink.Replace("lightning:", "", StringComparison.OrdinalIgnoreCase);
@@ -464,9 +471,7 @@ namespace BTCPayServer.Tests
             Assert.Contains($"{cryptoCode} Lightning settings successfully updated", await (await s.FindAlertMessage()).TextContentAsync());
             var invForPP = await s.CreateInvoice(null, cryptoCode);
             await s.GoToInvoiceCheckout(invForPP);
-            bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
-            Assert.NotNull(bolt11);
-            BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network);
+            await AssertBolt11();
             paymentMethods = await greenfield.GetInvoicePaymentMethods(newStoreId, invForPP);
             lnurlMethod = Assert.Single(paymentMethods, p => p.PaymentMethodId == "BTC-LNURL");
             lnurl = lnurlMethod.PaymentLink.Replace("lightning:", "", StringComparison.OrdinalIgnoreCase);


bigg-bb:
@NicolasDorier Please check if everything correct

NicolasDorier:
@bigg-bb, I will review when tests pass

bigg-bb:
@NicolasDorier 

I made 2 test case changes:

**1. CanUseLNURL**

**Before:**
```
Assert.Single(paymentMethods, p => p.AdditionalData["providedComment"]!.Value<string>() == "lol2");
```
Previously, there was only one payment method (`BTC-LNURL`) with `providedComment` in `AdditionalData`.

Since this PR introduces topup BOLT11 invoices, the invoice now has 2 payment methods (`BTC-LN` & `BTC-LNURL`), causing the test to fail — `BTC-LN` does not have `providedComment` in its `AdditionalData`, which is expected behavior.

**After:**
```
var lnurlMethod = Assert.Single(paymentMethods, p => p.PaymentMethodId == "BTC-LNURL");
Assert.Equal("lol2", lnurlMethod.AdditionalData["providedComment"]!.Value<string>());
```
We now check `providedComment` only on the LNURL-specific payment method object, since this is a test for LNURL behavior.

No breaking changes. No tests skipped for LNURL.

---

**2. CanConfigureCheckout**

**Before:**
```
Assert.Contains("?lightning=lnurl", payUrl);
Assert.DoesNotContain("amount=", payUrl);
Assert.StartsWith("bcrt", copyAddressOnchain);
Assert.StartsWith("lnurl", copyAddressLightning);
Assert.StartsWith($"bitcoin:{copyAddressOnchain!.ToUpperInvariant()}?lightning=LNURL", qrValue);
Assert.Contains($"bitcoin:{copyAddressOnchain}?lightning=lnurl", clipboard);
```
With a topup invoice, only LNURL was expected. However, since a BOLT11 invoice is now created successfully, the test was breaking.

**After:**
```
Assert.True(payUrl.Contains("?lightning=lnurl") || payUrl.Contains("?lightning=lnbcrt"),
    $"payUrl should contain either ?lightning=lnurl or ?lightning=lnbcrt, got: {payUrl}");
Assert.True(copyAddressLightning!.StartsWith("lnurl") || copyAddressLightning.StartsWith("lnbcrt"),
    $"copyAddressLightning should start with lnurl or lnbcrt, got: {copyAddressLightning}");
var lightningParam = copyAddressLightning.StartsWith("lnurl") ? "lnurl" : "lnbcrt";
var lightningParamUpper = lightningParam.ToUpperInvariant();
Assert.StartsWith($"bitcoin:{copyAddressOnchain!.ToUpperInvariant()}?lightning={lightningParamUpper}", qrValue);
Assert.Contains($"bitcoin:{copyAddressOnchain}?lightning={lightningParam}", clipboard);
```
We now assert that either BOLT11 or LNURL is present. If BOLT11 is unavailable, the test falls back to verifying LNURL.

No breaking changes. No tests skipped for LNURL.

---

Please let me know if there is a better way to approach this.

BTCPayServer.Tests/CheckoutUITests.cs
```diff
@@ -320,12 +320,16 @@ await TestUtils.EventuallyAsync(async () =>
             copyAddressOnchain = await s.Page.Locator("#Address_BTC-CHAIN .truncate-center").GetAttributeAsync("data-text");
             copyAddressLightning = await s.Page.Locator("#Lightning_BTC-CHAIN .truncate-center").GetAttributeAsync("data-text");
             Assert.StartsWith($"bitcoin:{copyAddressOnchain}", payUrl);
-            Assert.Contains("?lightning=lnurl", payUrl);
+            Assert.True(payUrl.Contains("?lightning=lnurl") || payUrl.Contains("?lightning=lnbcrt"),
```
NicolasDorier: Please don't test both, test the one you expect to be

BTCPayServer.Tests/CheckoutUITests.cs
```diff
@@ -320,12 +320,16 @@ await TestUtils.EventuallyAsync(async () =>
...
+                $"payUrl should contain either ?lightning=lnurl or ?lightning=lnbcrt, got: {payUrl}");
             Assert.DoesNotContain("amount=", payUrl);
             Assert.StartsWith("bcrt", copyAddressOnchain);
-            Assert.StartsWith("lnurl", copyAddressLightning);
-            Assert.StartsWith($"bitcoin:{copyAddressOnchain!.ToUpperInvariant()}?lightning=LNURL", qrValue);
-            Assert.Contains($"bitcoin:{copyAddressOnchain}?lightning=lnurl", clipboard);
+            Assert.True(copyAddressLightning!.StartsWith("lnurl") || copyAddressLightning.StartsWith("lnbcrt"),
```
NicolasDorier: Don't test one, test the one you expect to be

BTCPayServer.Tests/CheckoutUITests.cs
```diff
@@ -320,12 +320,16 @@ await TestUtils.EventuallyAsync(async () =>
...
+                $"copyAddressLightning should start with lnurl or lnbcrt, got: {copyAddressLightning}");
+            var lightningParam = copyAddressLightning.StartsWith("lnurl") ? "lnurl" : "lnbcrt";
```
NicolasDorier: No shady logic, just test the one your expect to be

bigg-bb:
@NicolasDorier

thank you for review, your requested changes are commited.

but its back to back 2 times CI pipelines are failing for other internal reasons...

i don't want more CI build to be triggered. please take care of rest.

Let me know if i did incorrect here

- biggbb

NicolasDorier:
Thank you @bigg-bb , I am right now pinging @Kukks because he had some objections into having 0 amount invoices in BTCPay long time ago, which I am not sure are still relevant.

He said,

> The main thing is not all LN providers are able to support it
like the Boltz/Blink/Strike ones cant do  0 amount invoices 
i guess in btcpay you would fall back gracefully because of how lnurl  works

Your PR fallback gracefully so it should be OK, I will need to double check with Blink plugin though.

NicolasDorier (changes requested):
@bigg-bb I've tested manually the PR, if I create a new invoice with top up 0, LND in backend, the QR Code is still a LNURL. The BOLT11 is still created, so I guess something with the `GetCheckout()` route might need to change so LNURL isn't the default shown.

The test that you had to change is relevant when a "unified QR code" including both Bitcoin and lightning show up.

I confirmed that Blink nicely fallback to LNURL.

bigg-bb:
@NicolasDorier 
Thank you for considering this much needed feature

NicolasDorier:
Can you fix https://github.com/btcpayserver/btcpayserver/pull/7263#pullrequestreview-4010744537 ?
I think we will make a new minor in one or two weeks, so this PR is a nice addition.

BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -758,10 +758,6 @@ public async Task<IActionResult> Checkout(string? invoiceId, string? id = null,
                 displayedPaymentMethods.Remove(lnurlId);
             }
 
-            // BOLT11 doesn't really support payment without amount
-            if (invoice.IsUnsetTopUp())
-                displayedPaymentMethods.Remove(lnId);
```
bigg-bb, line 763: @NicolasDorier this one removed. BOLT11 support amount less

Sorry i missed your change request message earlier

NicolasDorier: Now need to fix the tests :)

BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -353,24 +353,6 @@ await ConnectChannels.ConnectAll(s.Server.ExplorerNode,
...
-            Assert.NotEqual(1m, fetchedRequest.MaxSendable.ToDecimal(LightMoneyUnit.Satoshi));
-            var lnurlResponse = await fetchedRequest.SendRequest(new LightMoney(0.000001m, LightMoneyUnit.BTC),
-                network, new HttpClient(), comment: "lol");
-
-            Assert.Equal(new LightMoney(0.000001m, LightMoneyUnit.BTC),
-                lnurlResponse.GetPaymentRequest(network).MinimumAmount);
-
-            var lnurlResponse2 = await fetchedRequest.SendRequest(new LightMoney(0.000002m, LightMoneyUnit.BTC),
-                network, new HttpClient(), comment: "lol2");
-            Assert.Equal(new LightMoney(0.000002m, LightMoneyUnit.BTC), lnurlResponse2.GetPaymentRequest(network).MinimumAmount);
```
bigg-bb, line 373: @NicolasDorier 
Not sure if i did correct, but there is no way TopUp invoice can use LNURL with All methods selected

BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -353,24 +353,6 @@ await ConnectChannels.ConnectAll(s.Server.ExplorerNode,
             await s.Page.CheckAsync("#LUD12Enabled");
             await s.ClickPagePrimary();
 
-            // Topup Invoice test
```
NicolasDorier, line 356: @bigg-bb do not delete this, instead you should check that the invoice checkout is using BOLT11.

bigg-bb: @NicolasDorier but this test suppose to be for "CanUseLNURL"?. i mean just asking...

NicolasDorier: Good point. While this would mean this part should be moved to another test, it's not worth troubling yourself with this, just modify this test to assert that a 0 amount BOLT11 is returned. It's just a test, good enough.

bigg-bb: @NicolasDorier 
I see below this topup invoice snippet, there is also a standard invoice snippet, which just assume there is LNURL and request a bolt11 with callback url GetAsync("BTC/lnurl/pay/i/" + invoiceId)

i guess this make sense, there is no QR but we can still use LNURL this way.

going for same approach in topup invoice as well

NicolasDorier: Yeah everywhere that was expecting LNUrl in tests, and the behavior changed, the test need to change.

Not only those tests are useful for checking against regressions, but it allow to understand better what user facing behavior this PR is changing.

BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -353,6 +353,32 @@ await ConnectChannels.ConnectAll(s.Server.ExplorerNode,
             await s.Page.CheckAsync("#LUD12Enabled");
             await s.ClickPagePrimary();
 
+            // Topup Invoice test
+            var i = await s.CreateInvoice(storeId, null, cryptoCode);
+            await s.GoToInvoiceCheckout(i);
+            // Top-up invoices now display a zero-amount BOLT11 on checkout, like standard invoices.
+            var bolt11 = await s.Page.Locator("#Lightning_BTC-LN .truncate-center").GetAttributeAsync("data-text");
+            Assert.NotNull(bolt11);
+            Assert.Null(BOLT11PaymentRequest.Parse(bolt11!, s.Server.ExplorerNode.Network).MinimumAmount);
```
NicolasDorier: I don't think this one can be null, isn't it 0 instead?

bigg-bb: yea, but below in standard invoice test, we are not doing any bolt11 test.
this one is all about fetch and pay lnurl 

so i am removing these check.

BTCPayServer.Tests/PlaywrightTests.cs
bigg-bb, line 479: @NicolasDorier These lines will also cause problems, as now we have Bolt11 instead of lnurl

any recommended approach for this one?

bigg-bb:
Finally

PR #7265, opened 2026-03-24
feat: add ability to add comment to the transaction on the send view

### LINKED ISSUE #6687: [FR] Add the ability to add a comment to the transaction on the Send view
Originally reported by @astupidmoose.

This is a feature request to simply add the ability to add a comment on the TX at the moment you're constructing it from the Send view, instead of having to wait until the TX is broadcast, then being able to comment on the list view.

Would be a nice QOL update, and is fairly simple from a design/UX perspective, but If there are technical considerations to add, please feel free to do so.

Abhijay007:
closes #6687

https://github.com/user-attachments/assets/b473c7f2-e834-4e91-9357-c707dcfe5e48

### DIFF
--- a/BTCPayServer/Controllers/UIWalletsController.PSBT.cs
+++ b/BTCPayServer/Controllers/UIWalletsController.PSBT.cs
@@ -576,6 +576,12 @@ namespace BTCPayServer.Controllers
                             TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Transaction broadcasted successfully ({0})", transaction.GetHash()].Value;
                         }
 
+                        if (!string.IsNullOrEmpty(vm.SigningContext?.Comment))
+                        {
+                            var txObjId = new WalletObjectId(walletId, WalletObjectData.Types.Tx, transaction.GetHash().ToString());
+                            await WalletRepository.SetWalletObjectComment(txObjId, vm.SigningContext.Comment);
+                        }
+
                         if (vm.SigningContext.PendingTransactionId is not null)
                         {
                             await _pendingTransactionService.Broadcasted(GetPendingTxId(walletId, vm.SigningContext.PendingTransactionId));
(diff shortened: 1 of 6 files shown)


Abhijay007:
cc @dstrukt , @NicolasDorier

pavlenex:
@Abhijay007 Has this been done in a way so it mimics what we have in wallet view? We should ensure no fragmentation for same features.
For example, will user see a comment from a transaction list, or can they edit it there.

Screenshot 2026-03-25 at 17 44 59

Abhijay007:
> @Abhijay007 Has this been done in a way so it mimics what we have in wallet view? We should ensure no fragmentation for same features. For example, will user see a comment from a transaction list, or can they edit it there.

Yes, fully integrated, no fragmentation. It uses the exact same `SetWalletObjectComment` call and storage as the list view, so the comment shows up pre-filled in the list popup, the icon highlights correctly, and the user can still edit/clear it from there as before.

NicolasDorier:
Concept ACK. We definitely need comments in the send. However, UX is bit weird, having to scroll down.
Let's wait @dstrukt feedback

dstrukt:
Sorry it took me a moment, gave it some thought, and I think this implementation strikes the best balance in the current UI - it's short and unobtrusive, basically 2 lines, is placed in the correct logical order, and works with multiple destinations, etc... Great job @Abhijay007!

Given it's so unobtrusive, it doesn't make sense to tuck it behind an accordion, or behind an icon that would require an additional click (had to see for myself though) - but if we were to get feedback that it was obtrusive or unnecessary, tucking it away could be a reasonable path forward, i.e:

https://github.com/user-attachments/assets/bc840a0e-8493-423a-92e4-82e179b7bf82

My only suggestion might be the naming of the feature from a `Comment` to a `Note`, but given we use `Comment` already in the list view, and likely in the code(?) that shouldn't be changed in this PR - just a personal preference for consideration. cc @pavlenex

Functional tACK!

PR #7266, opened 2026-03-24
Add txid search to wallet transactions view

### LINKED ISSUE #7091: Add support for search, specifically for TX id inside the wallet view

j-kon:
## Summary

Fixes #7091 by adding transaction ID search to the wallet transactions page.

## What changed

- added a GET search input to the wallet transactions page
- filtered wallet transactions by transaction ID
- preserved `searchTerm` through label filters and lazy loading
- showed a filtered empty state when no transactions match
- added regression coverage for partial and no-match txid search

## Notes

- the change stays scoped to wallet transaction TXID search only
- the implementation follows BTCPay’s existing list-page pattern used by invoice and payment-request views
- reporting/export behavior was left unchanged intentionally:
  - nearby reporting links do not preserve list search state
  - wallet export is not currently search-aware

## Testing

- `dotnet build` passed
- `git diff --check` passed
- attempted `dotnet test BTCPayServer.Tests/BTCPayServer.Tests.csproj --filter "FullyQualifiedName~CanRescanWallet"`
- the test failed during `ServerTester` initialization because the local regtest test infrastructure was not running
- the failure occurs before the new #7091 TXID search assertions execute

@pavlenex check it this out


pavlenex:
Hi, thanks for your contribution but afaik this issue is already being tackled as a part of https://github.com/btcpayserver/btcpayserver/pull/7133

j-kon:
Thanks for the heads-up. I checked #7133 and it does appear to cover the wallet transaction TXID search work already.

Since this PR overlaps with that effort, I’m happy to close this one to avoid duplicate review/work. If a smaller TXID-only piece from this branch would still be useful separately, I’m also happy to help extract it.

Issue #7145, opened 2026-01-27
[Bug]: Deleting a subscription plan with customer is possible if you delete the whole offering


Ou7law007:
### What is your BTCPay version?

© BTCPay Server v2.3.3+ee0ce92e1

### How did you deploy BTCPay Server?

Locally with docker

### What happened?

It seems like the system's intent is to not allow deletion of plans if they have active subscriptions.
This is not stated anywhere but 2 observations:

1. Only plans with no active subscribers are allowed to be deleted.
2. Plans with no active subscribers cannot be deleted

<img width="877" height="556" alt="Image" src="https://github.com/user-attachments/assets/313e50b7-1460-4865-b85c-fcf1d46e47fd" />

<img width="877" height="556" alt="Image" src="https://github.com/user-attachments/assets/d65a4810-1426-4150-83bd-a21eef3c2822" />

<img width="877" height="556" alt="Image" src="https://github.com/user-attachments/assets/0c053075-76a0-4537-9c77-dfdfbaf3a1b1" />

However, deleting the whole offering is still allowed and it deletes all plans and all subscriptions and all subscribers!

<img width="877" height="556" alt="Image" src="https://github.com/user-attachments/assets/d49eee39-78e0-4a84-a6da-5113d4484cab" />

<img width="877" height="105" alt="Image" src="https://github.com/user-attachments/assets/a4b89587-9145-4bce-854b-1a506a3a9096" />

Now the whole thing is gone!

### How did you encounter this bug?

Description above

### Relevant log output

```shell

```

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

NicolasDorier:
Yes, I kicked the can for this one. Basically, when deleting a plan, we need to ask the question to the user where to migrates the subs in it. (a sub can't exist without a plan)

So we need to ask:
* Migrate to a new plan, and reimburse what hasn't been used
* Migrate to a new plan for free. Start from day 0.
* Migrate to a new plan when the current period is over
* Migrate to a new plan. Start from day X. (so if the user is currently in day X of the retired plan, he will end up to day X of the new plan)

Feature wise, maybe all of those options aren't really needed. But I couldn't decide before shipping. That's something we need to do though.

If you have suggestion of what options would make sense let me know.
I need input on what a migration of subs into a new plan should look like.

Ou7law007:
[...]

PR #7270, opened 2026-03-25
Subscriptions: SubscriberDisabled webhook now includes the reason why subscriber has been disabled

NicolasDorier:
Implement one of the suggestions of @ou7law007. https://github.com/btcpayserver/btcpayserver/issues/7145#issuecomment-3817443684

> Using only webhooks (e.g. SubscriberDisabled), you cannot reliably distinguish between when the subscription is suspended in the middle of the subscription period (e.g. using the UI website, or using cancel now with refund) or the subscription has actually fully ended without the user resubscribing. This is important if you're trying to work around the issue (2) by cutting of the subscriber's continuous benefit when the subscription actually ends vs when it was ended mid subscription.

A `reason` enum with value `Suspension` or `Expired` has been added to the webhook, along with `suspensionReason`.


Ou7law007:
Looks great, Emperor :D

PR #7271, opened 2026-03-25
Add transaction comment field to wallet send view

### LINKED ISSUE #6687: [FR] Add the ability to add a comment to the transaction on the Send view

j-kon:
## Summary

Fixes #6687 by adding a transaction comment field to the wallet send view.

## What changed

- added a comment field to the wallet send page
- preserved the comment through the existing signing flow
- preserved the comment when creating a pending transaction
- saved the comment onto the transaction after successful broadcast
- added Playwright coverage for the send-comment flow

## Notes

- reused the existing wallet transaction comment pattern
- kept the change scoped to the wallet send flow
- no migration is needed because the pending transaction blob is stored as JSON

## Testing

- `dotnet build`
- `dotnet test BTCPayServer.Tests/BTCPayServer.Tests.csproj --filter "FullyQualifiedName~WalletTests.CanManageWallet"`
- `dotnet test BTCPayServer.Tests/BTCPayServer.Tests.csproj --filter "FullyQualifiedName~MultisigTests"`
- `git diff --check`


pavlenex:
Thanks for your contrbution but this is already being worked on https://github.com/btcpayserver/btcpayserver/pull/7265 , so I'll close it to avoid review duplication.

Issue #7023, opened 2025-12-09
Subscriptions missing features


NicolasDorier:
## API

Asked by @ndeet on https://github.com/btcpayserver/btcpayserver/pull/7022#issuecomment-3631105737

- [x] https://github.com/btcpayserver/btcpayserver/pull/7296
- [x] https://github.com/btcpayserver/btcpayserver/pull/7297
- [ ] PUT /api/v1/plan-checkout/{checkoutId} (not sure if this is really possible or needed)
- [ ] GET subscribers

## Features

- [ ] Subscriber Portal: Customer should be able to give a notification email
- [ ] Subscriber Portal: Customer should be able to give a notification nostr address
- [ ] Subscriber Portal: Customer should be able to configure payment NwC for paying future renewal
- [ ] Subscriber Portal: Customer should be able to put a pull payment link for paying future renewal
- [ ] Backend: Allow migration of subscribers to new plan
  - [ ]  Migration without changing current entitlements
  - [ ]  Migration with immediate update of entitlements
- [ ] Plan Retirement
- [x] Removed expired portal checkout and portal session from the DB
- [ ] Allows credit refunds via Pull Payments
- [ ] Reports
- [x] Documentation
- [ ] Show logs of customer in the backend.
- [ ] [Monetization: Assign users created by a subscriber to an existing subscription](https://github.com/btcpayserver/btcpayserver/discussions/7101)

## Customers

Need more thought about integrating customers features.

Ou7law007:
Has this been fully implemented now? Is it stable now? Are there any massive reworks to be expected?

The webhooks in the documentation still don't document subscription related events.

I can see:

1. https://blog.btcpayserver.org/btcpay-server-2-3-0/
2. https://github.com/btcpayserver/btcpayserver/pull/7022
3. https://docs.btcpayserver.org/Subscriptions/#8-subscription-api-access also says "Public documentation:
https://docs.btcpayserver.org/API/Greenfield/v1/" but the link does not actually document any subscription related webhook events like "SubscriberPhaseChanged" or "SubscriberActivated" which are in fact being correctly sent but just not documented anywhere that I can see!

I can only see these webhooks:

<img width="263" height="1115" alt="Image" src="https://github.com/user-attachments/assets/374c3915-392c-47a8-a8af-31f7d7903615" />

NicolasDorier:
@Ou7law007 the subscription events are now documented

Ou7law007:
[...]

PR #7284, opened 2026-03-27
subscription:  allows credit refunds via Pull Payments

TChukwuleta:
According to issue https://github.com/btcpayserver/btcpayserver/issues/7023

This PR aims to allows credit refunds via Pull Payments

image

Cc. @NicolasDorier


BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
```diff
@@ -358,6 +387,83 @@ private async Task UpdateSubscriptionStates(SubscriptionContext subCtx, MemberSe
...
+        if (portal is null)
+            return null;
+
+        var sub = portal.Subscriber;
+        var credit = sub.GetCredit();
+        if (amount <= 0 || amount > credit)
+            return null;
+
+        var store = sub.Plan.Offering.App.StoreData;
+        var pullPaymentId = await pullPaymentService.CreatePullPayment(store, new CreatePullPaymentRequest
```
NicolasDorier: Why not creating the pull payment only if the CreatedDebit succeed?

BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
```diff
@@ -358,6 +387,83 @@ private async Task UpdateSubscriptionStates(SubscriptionContext subCtx, MemberSe
...
+        {
+            Name = $"Credit refund for {sub.Customer.GetPrimaryIdentity()}",
+            Amount = amount,
+            Currency = sub.Plan.Currency,
+            AutoApproveClaims = false
+        });
+        var debitResult = await subCtx.TryCreditDebitSubscriber(sub, $"Credit refund (Pull Payment: {pullPaymentId})", 0m, amount);
+        if (debitResult is null)
+        {
+            await pullPaymentService.Cancel(new PullPaymentHostedService.CancelRequest(pullPaymentId));
```
NicolasDorier: if you create only when that succeed, you don't need to cancel anything.

BTCPayServer/Plugins/Subscriptions/SubscriptionsPlugin.cs
```diff
@@ -69,6 +69,20 @@ WHERE id IN (SELECT id FROM expired_plan_checkout)
...
+                SELECT pull_payment_id FROM subs_credit_refunds
+                WHERE deducted = false
+                ORDER BY pull_payment_id LIMIT 1000
+            ),
+            deleted AS (
+                DELETE FROM subs_credit_refunds
+                WHERE pull_payment_id IN (SELECT pull_payment_id FROM completed_refunds)
+                RETURNING *
+            )
+            SELECT COUNT(*) FROM deleted;
```
NicolasDorier: You can remove this. I don't expect there will be a huge amount of credit refund, and if there is, we can fix it later in a way that doesn't trigger a full table scan on the table every 10 min.

BTCPayServer/Plugins/Subscriptions/SubscriptionHostedService.cs
```diff
@@ -358,6 +387,83 @@ private async Task UpdateSubscriptionStates(SubscriptionContext subCtx, MemberSe
...
+            .Include(r => r.Subscriber).ThenInclude(s => s.Customer)
+            .Where(r => r.Deducted)
+            .Join(ctx.Payouts,
+                r => r.PullPaymentId,
+                p => p.PullPaymentDataId,
+                (r, p) => new { Refund = r, Payout = p })
+            .Where(x => x.Payout.State == PayoutState.Cancelled)
+            .Select(x => x.Refund).ToListAsync();
+
+        foreach (var refund in cancelledRefunds)
```
NicolasDorier: The fact a payout is cancelled doesn't mean that the pull payment is.
A customer could create a payout that will be cancelled, then a second one that will succeed. If I understand this code, this mean that the customer will have get refunded credit.

Don't attempt to reverse automatically. The merchant can manually add credits and archive the pull payment if he wants to.

NicolasDorier (changes requested):
This shouldn't be exposed on the subscription portal. Rather, it should be an action the merchant can manually do on the Offering page.

TChukwuleta:
Hey Emperor @NicolasDorier 

I have removed credit refunds from the subscription portal to the offering page, so that merchants can issue refunds the same way they credit and charge... 

I have also stopped auto reversal allowing the merchants manually top up credits for failed payouts. 


I included email rule for this use case

image

TChukwuleta:
Rebased issue

credit refund email now improved with a tag

@NicolasDorier 

image

NicolasDorier:
Seems good. The email seems like a spam but I don't know what to do better :D

NicolasDorier:
image

The subscriber id shouldn't be shown. It isn't useful to the user.
Please instead put the link to the pull payment (like we do with `Inv: invoice_id``

See 

image

NicolasDorier:
image

Adjust the order.
1. Credit
2. Charge
3. Refund

NicolasDorier:
image

is it broked by this PR? there should be a small space on the right of the Test badge.

BTCPayServer/Plugins/Subscriptions/Controllers/UIOfferingController.cs
```diff
@@ -158,6 +159,22 @@ await SubsService.UpdateCredit(
                 });
             TempData.SetStatusSuccess(message);
         }
+        else if (command is "refund" && amount is > 0)
+        {
+            var pullPaymentId = await SubsService.CreateCreditRefund(sub.Id, amount.Value, Request.GetRequestBaseUrl());
```
NicolasDorier: You need to check permission for `Policies.CanCreatePullPayments`. As is, any user with the permission `btcpay.store.canviewofferings` is able to drain the wallet completely.

What we do for invoice's refund is that we requires `Policies.CanCreateNonApprovedPullPayments`.

And then, if `Policies.CanCreatePullPayments` is missing, we turn the auto approval off on the pull request.

Issue #6713, opened 2025-05-12
Phase 2: Multiple taxes per product, tax included or tax excluded keypad input, tax and tips configurability 


pavlenex:
Overview:
This is a tracker issue, upon successfully closing https://github.com/btcpayserver/btcpayserver/pull/6724 we'll be tracking the remaining to-do's here, to ensure this feature is fully prepared for 2.2.0, while https://github.com/btcpayserver/btcpayserver/issues/6712 is shipped with 2.1.2 and will be used at an event, which was our priority, great work Nicolas for shipping this one fast!

Scope of Work:

Extend the initial POS global tax functionality https://github.com/btcpayserver/btcpayserver/issues/6712  by allowing  or per category, enabling greater flexibility and accuracy in tax calculation.

- [ ]https://github.com/btcpayserver/btcpayserver/pull/6739
- [ ] Multiple taxes can be added per product
- [x] Clearly reflect these custom tax rates during checkout, receipt generation, and reporting.
- [ ] Tax included or tax excluded price in keypad (allowing merchants to configure if the amount they apply in keypad includes tax or not)
- [ ] Apply tax to tips (configurability)
- [ ] Optional: Category-Specific Tax: Allow merchants to set tax rates per category, applying automatically to all items within that category (unless item-specific tax is defined). This one may be a bit tricky with how we currently do categories (as tags) and would  require input from @NicolasDorier

PR #7290, opened 2026-04-01
Allows merchants configure tax inclusion or exclusion

TChukwuleta:
Part of pending tax issues as stated here https://github.com/btcpayserver/btcpayserver/issues/6713

This PR aims to resolve
"Tax included or tax excluded price in keypad (allowing merchants to configure if the amount they apply in keypad includes tax or not)"

image

image

image

Cc. @NicolasDorier


TChukwuleta:
Might need to rerun the test.. when I clicked on the circle ci nothing seems to be failing

NicolasDorier:
@TChukwuleta can you double check that this settings works on cart view? I think it should be ok, but probably need to double check.

It is also missing test. When it is checked, you need to check that the keypad is behaving properly.

Also, I believe we have a `taxIncluded` invoice metadata that it is used by the tax feature. If it is tax included, I think this should be populated.

TChukwuleta:
Hey @NicolasDorier 

 Yea.. I did test for the Cart view.. Here are screenshots

Tax Inclusive:

image


Tax exclusive

image


I have included tests for this.

Regarding this:

"Also, I believe we have a taxIncluded invoice metadata that it is used by the tax feature. If it is tax included, I think this should be populated."

I think what exists in the code already caters to this

`
                        TaxIncluded = summary.Tax == 0m ? null : summary.Tax,
`

for tax-exclusive invoices it records the tax added on top, and for tax-inclusive invoices it records the tax extracted from the price.

pavlenex (changes requested):
Good direction, 

Feedback:
Screenshot 2026-05-14 at 12 57 32

1. It seems there's a rounding error in how tax is calculated? 10% of 10 is not 0.91
2. Replace the current helper text:
- Current: $10.00 (incl. $0.91 tax @ 10%)
- Proposed: $10.00 (incl. $0.91 tax 10%) (without @)

I stopped testing after seeing rounding error, it seems like a serious blocker that needs to be addressed same logic as from the tax excluded should be used, there's no rounding error there.

TChukwuleta:
Removed the @ from the helper text 

Pav and I have aligned on rounding question, the calculation is correct as-is (pre-tax + (10% × pre-tax) = $10.00)

pavlenex (approved):
tACK. My previous comment about calculation logic isn't correct. 

I've done quick testing on this one, it seems to be correctly shown in POS various modes, product or keypad, it exposes tax in receipt and in reports. 

Image
Image
Image

Good work @TChukwuleta!

PR #7293, opened 2026-04-03
Allow qr-code login expiry after a day

### LINKED ISSUE #6801: [Bug]: qr-code login expires after a few hours

BTCPay 2.1.5

docker

### What happened?

Logged in user(employee) on pos terminal with qr code. After some hours the cookie expired or something else invalidated. Imho this did not happen before and the user stayed logged in much longer. This is a problem because last invoices/payments are then not accessible anymore and you need to create normal logins which costs a lot of time.

go to pos
click on qr code icon
assuming you have an employee role user, select that with login qr
scan qr with pos terminal
add page to home 
open it again after some hours

### Relevant log output

```shell

```

### What browser do you use?

chrome, standard on pos devices

### Additional information

_No response_

TChukwuleta:
Resolves #6801 

Extending login code to be persistent up to a day

### DIFF
--- a/BTCPayServer/Controllers/UIAccountController.cs
+++ b/BTCPayServer/Controllers/UIAccountController.cs
@@ -141,10 +141,13 @@ namespace BTCPayServer.Controllers
                 }
 
                 _logger.LogInformation("User {Email} logged in with a login code", user!.Email);
+                var now = DateTimeOffset.UtcNow;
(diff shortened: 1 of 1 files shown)


TChukwuleta:
Hey @NicolasDorier 

could you help review this

NicolasDorier:
Merging this, but I think we should rework login code to make it configurable.

PR #7298, opened 2026-04-06
Include configuration for tax on tips

TChukwuleta:
Part implementation for phase 2 of tax implementation (https://github.com/btcpayserver/btcpayserver/issues/6713)

Cc. @NicolasDorier 

This PR introduces configuration to apply tax to tips

image

POS KeyPad

image

image

POS Cart

image

image


NicolasDorier:
I think adding coderabbit suggestion `Consider mirroring this scenario in CanUsePOSKeypad.` make sense. (it generated a prompt you can give to your agent for this)

@pavlenex can you check that this is what you want?

TChukwuleta:
I have added the suggestion in the CanUsePOSKeypad test mirroring keypad PoS for tax in tip

dstrukt (commented):
Love to see the work on this, great job @TChukwuleta!

### 1 

Just a thought for the tip / tax calculation display.

Do we want to simplify it to match what Square, etc.. do in the states? The tax rate (say 9.25%) is the same for both the Tip & the Goods/Services, it's just called a "Sales Tax" - can't speak to other areas of the world. Are there instances where you want a different tax rate for just Tips?

cc @pavlenex Curious to hear your thoughts here.

Regarding the display to the end user via the Receipt or Cart I.e.

A) Tip is calculated and displayed as a % of the Subtotal + Taxes
```
Goods/Services = 20.00
Taxes (10%) = 2.00
Subtotal = 22.00

Tip (20%) = 4.40 (Could have a info hover showing the Subtotal + Tip tax is how this is calculated)
-----
Total = 26.40
```

B) What we do now, a more thorough, but arguably more complex breakdown many are not used to see (again, at least here in the states)
```
Goods/Services = 20.00
Taxes (10%) = 2.00
Subtotal = 22.00

Tip (20%) = 4.00
Tip Tax (10%) = 0.40
-----
Total = 26.40
```

Both are valid paths, and our current approach gives more flexibility to the merchant, but B is a bit more complicated on the end-user display, so just thought I'd flag it for discussion!

Perhaps there is some accounting context I'm not aware of or taking into consideration as well. 

-----

### 2 

I have a few UI nits and suggestions that might help the above as well - in the process of typing up now, will update this review this evening.

TChukwuleta:
Hey @dstrukt 

dropping a nudge on the suggested UI improvements..

pavlenex:
@dstrukt When I spoke with someone who requested this feature, I think they mentioned that in their country, which I believe was in the EU, tips had a different tax rate because they are treated as a gift, even though they are still taxed. For example, if sales tax is 10%, tips may only be taxed at 2%.
My concern is that bundling everything together could create transparency issues around what the customer is actually being charged for. In my opinion, for now it's better to expose the breakdown clearly, so both the customer and merchant can see exactly how charges are calculated. If "tax on tips" is disabled, merchants won't even have to deal with this complexity.


@TChukwuleta I reviewed this PR and only have a few minor consistency nits. Mainly around having a consistent way of displaying tips.


We should display tips tax % in receipt same way we do it for tax, and tip %.
Screenshot 2026-05-14 at 15 11 38
Screenshot 2026-05-14 at 16 35 41

TChukwuleta:
Hey @pavlenex 


I have updated it to follow the format

image

image

TChukwuleta:
Added Tip on tax to invoice reporting

image



Also removed it from product information since all information are present in receipt data

image


Cc. @pavlenex

pavlenex:
I've give it another rundown, it looks good to me. Long-term we really need to do something about information architecture in the invoice list, but it's beyond the scope of what this PR is trying to address.

Screenshot 2026-05-18 at 18 29 27

@NicolasDorier good to go as far as user-testing goes.

pavlenex (approved):

PR #7301, opened 2026-04-07
Include get subscribers API

TChukwuleta:
Breaking down subscription portal improvements into smaller PR (https://github.com/btcpayserver/btcpayserver/pull/7284)

This PR adds GET subscribers (API)

image

image

Cc. @NicolasDorier


NicolasDorier:
I need to think more about this API.
This is because as is, it isn't very useful, and it is hugely inefficient.

Each subscriber returned include the offering information, which is data duplicated as many time as there is subscribers.
On top of this, as soon as there will be too many subscribers, this will start timing.

I am wondering if a `Get Subscribers` can be actually useful for anything.

PR #7304, opened 2026-04-09
Fix reserved addresses after wallet replacement

### LINKED ISSUE #7295: [Bug]: Reserved address of replaced wallet still showing

2.3.7

docker

### What happened?

Reported by @petzsch , ping @thgO-O 

> not sure if this is expected behaviour. I've replaced my onchain wallet about 3 5 days ago, but under reserved addresses there are still some addresses listed from the old wallet:
perhaps there is a cleanup step missing?

Image

* Create a wallet
* Reserve an address
* Create a new wallet
* Still see it in the reserved address window

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

thgO-O:
Fixes #7295.

Clears persisted reserved receive address markers when an on-chain wallet is replaced or deleted, so old addresses no longer appear in Reserved Addresses.

Adds a Playwright regression test for the wallet replacement flow.

### DIFF
--- a/BTCPayServer.Tests/WalletTests.cs
+++ b/BTCPayServer.Tests/WalletTests.cs
@@ -515,6 +515,89 @@ public class WalletTests(ITestOutputHelper helper) : UnitTestBase(helper)
 
     [Fact]
     [Trait("Playwright", "Playwright-2")]
+    public async Task CanHideReservedAddressesFromReplacedWallet()
+    {
+        await using var s = CreatePlaywrightTester();
+        await s.StartAsync();
+        await s.RegisterNewUser(true);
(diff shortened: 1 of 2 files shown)


NicolasDorier:
This mean that if you switch from one derivation scheme to another one and back, you will lose the tags. Also current wallets would need to be fixed.

@thgO-O what about instead you query NBXplorer to check if those addresses are in the current derivation scheme?
Just use raw SQL on NBXplorer directly (using `unnest(@addresses)`) so you have only a single request going to postgres.

That way no need of migration, no tags are lost, and no risk to have a bug where we forget to call `UnmarkReservedReceiveAddresses`

PR #7307, opened 2026-04-10
Remove support for importing keys to RPC

NicolasDorier:
This feature is barely used, not understood by most users, while requiring quite heavy maintenance on my part in NBXplorer.

I will keep the support on NBXplorer to make it harder to break people who were using it, but I will stop investing handling corner cases.

image


Kukks:
Please note that this will very likely kill/brick liquid user funds.

NicolasDorier:
@Kukks yeah, that's my only worry, though I would expect people using samrock/aqua instead.

Maybe on NBXplorer I should crash when generating a new address for impacted wallet rather than silently not importing the keys. This way users are at least alerted.

That said, if I do it on BTCPay Server and not on NBXplorer for now, it would prevent at least people to rely on it now.

NicolasDorier:
Ok so I've checked and asked about various plugin like samrock: It's not using RPC at all. I will disable the option in BTCPay, and for NBX, maybe leave it but just undocument it and not test it anymore.

Issue #7314, opened 2026-04-13
[Bug]: Double encoding


Sanja22B:
### What is your BTCPay version?

BTCPay Server 2.3.7

### How did you deploy BTCPay Server?

Local Docker 

### What happened?

**Title**  
`TranslateTagHelper` re-translates already localized dynamic content, causing double-encoded output in some views

**Description**  
`TranslateTagHelper` should only be applied to static text nodes in Razor views. In a few places it is currently attached to content that is already localized or dynamically generated, which causes a second localization/encoding pass.

This can lead to user-visible rendering issues. For example, in `NotificationSettings.cshtml`, localized text containing non-ASCII characters can end up displayed literally as HTML entities such as `&#x17E;` instead of `Ž/ž`.

**Current behavior**  
When `text-translate="true"` is used on elements whose child content is already localized, `TranslateTagHelper` reads the rendered child content and tries to localize it again:

- `BTCPayServer/TagHelpers/TranslateTagHelper.cs`
- `output.GetChildContentAsync().GetContent()`
- `_localizer[originalContent.Trim()]`
- `output.Content.SetContent(newContent)`

If the child content is already localized and already encoded by Razor, this second pass can result in double-encoded output being rendered to the browser.

**Example**  
In `BTCPayServer/Views/UIManage/NotificationSettings.cshtml`:

```cshtml
<label class="form-check-label cursor-pointer"
       asp-for="DisabledNotifications[index].Selected"
       text-translate="true">@item.Text</label>
```

`@item.Text` is already localized in `UIManageController.Notifications.cs` via notification handler metadata. Applying `text-translate="true"` again causes an unnecessary second translation/encoding pass.

A similar issue exists in:

```cshtml
<span class="user-status badge bg-@status.Item2" text-translate="true">@status.Item1</span>
```

in `BTCPayServer/Views/UIServer/ListUsers.cshtml`, where `@status.Item1` is already localized in the view.

**Expected behavior**  
`text-translate="true"` should only be used for static literal text in Razor views, not for:

- already localized values
- `@item.Text`
- `@status.Item1`
- `@ViewData["Title"]`
- other dynamic child content

These cases should render directly without another `TranslateTagHelper` pass.
**Affected locations**  
User-visible bug:
- `BTCPayServer/Views/UIManage/NotificationSettings.cshtml`
- `BTCPayServer/Views/UIServer/ListUsers.cshtml`

Additional incorrect usage, but not necessarily user-visible:
[...]

PR #7315, opened 2026-04-13
Fix double translation of already localized view content

Sanja22B:
This PR should close the https://github.com/btcpayserver/btcpayserver/issues/7314 issue.


teamssUTXO:
Nice PR @Sanja22B!

Maybe the `text-translate` label can be removed here :

```
<option value="taproot" text-translate="true">@ui.StringLocalizer["Taproot"]</option>
```

File : BTCPayServer\Blazor\VaultBridge\Elements\XPubSelect.razor:16

PR #7318, opened 2026-04-14
Allow server admin to specify if an invited user subscribes for monetization or not

TChukwuleta:
Allow admin to exempt specific users from subscription requirements when creating users. 

Added a  "Bypass monetization for this user" toggle to the server admin create user page. 

This toggle only appears as long as 

Default toggle value is false, which expect all users to subscribe to monetization as long as monetization is enabled for this server. 
When enabled, the newly created user will not be enrolled into the subscription offering regardless of whether monetization is active on the server.

This doesn't apply to store adding users, as every user added under a particular store is expected to subscribe as long as the server has monetization toggled on

This is part of https://github.com/btcpayserver/btcpayserver/discussions/7101 aiming to tackle the second problem to an extent (Have a global monetization toggle of the same effect)

Create a user with  monetization

image

Create a user without monetization

image

and in the users listing

image

### DIFF
--- a/BTCPayServer.Tests/MonetizationTests.cs
+++ b/BTCPayServer.Tests/MonetizationTests.cs
@@ -223,7 +223,6 @@ public class MonetizationTests(ITestOutputHelper helper) : UnitTestBase(helper)
         await offeringPMO.GoToSubscribers();
         await offeringPMO.AssertHasNotSubscriber("normal-guest@gmail.com");
 
-
         await GoToMonetization(s);
         await ClickSetupOffering(s);
 
@@ -237,7 +236,11 @@ public class MonetizationTests(ITestOutputHelper helper) : UnitTestBase(helper)
         await ClickSetupOffering(s);
         await s.Page.ClickAsync("#RequireSubscriptionForInvitedUsers");
         await s.FindAlertMessage(partialText: "Server-invited users now require a subscription.");
-        await CreateUserAsAdmin(s, "enrolled-invited@gmail.com");
+        var evInvited = await s.Server.WaitForEvent<SubscriptionEvent.NewSubscriber>(async () =>
+        {
+            await CreateUserAsAdmin(s, "enrolled-invited@gmail.com");
+        });
+        Assert.Equal("enrolled-invited@gmail.com", evInvited.Subscriber.Customer.Email.Get());
         await AssertSubscribed(s, "enrolled-invited@gmail.com", true);
     }
 
--- a/BTCPayServer.Tests/MonetizationTests.cs
+++ b/BTCPayServer.Tests/MonetizationTests.cs
@@ -217,32 +217,34 @@ public class MonetizationTests(ITestOutputHelper helper) : UnitTestBase(helper)
         await s.GoToServer(ServerNavPages.Users);
         var users = new PMO.UsersPMO(s);
         await users.DeleteUser("normal-guest@gmail.com");
-
-        await GoToMonetization(s);
-        await GoToOffering(s);
-        await offeringPMO.GoToSubscribers();
-        await offeringPMO.AssertHasNotSubscriber("normal-guest@gmail.com");
-
-        await GoToMonetization(s);
-        await ClickSetupOffering(s);
-
-        await s.Page.ClickAsync("#RequireSubscriptionForInvitedUsers");
-        await s.FindAlertMessage(partialText: "Server-invited users are now exempt from subscription requirements.");
-        await CreateUserAsAdmin(s, "exempt-invited@gmail.com");
-        await AssertSubscribed(s, "exempt-invited@gmail.com", false);
-        await CanLog(s, "exempt-invited@gmail.com");
-
-        await GoToMonetization(s);
-        await ClickSetupOffering(s);
-        await s.Page.ClickAsync("#RequireSubscriptionForInvitedUsers");
-        await s.FindAlertMessage(partialText: "Server-invited users now require a subscription.");
-        var evInvited = await s.Server.WaitForEvent<SubscriptionEvent.NewSubscriber>(async () =>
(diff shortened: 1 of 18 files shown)


TChukwuleta:
Cc. @NicolasDorier 

 kindly review

NicolasDorier:
@TChukwuleta rather that having this in the monetization settings, wouldn't it make sense to add this as part of the invitation?

This would solve two issues:
1. UX wise, the admin doesn't necessarily wants for the invite to skipped monetization
2. There is a security vulns. Because normal users actually can invite third party user, allowing them to bypass monetization.

Also, I think this would be less code.

TChukwuleta:
hey @NicolasDorier  I have move this to the invitation flow.. 

The server admin can decide if a particular invited user should monetize or not (default to monetize)

The option isnt available for store owner, as long as monetization is enabled all invited store user are expected to monetize

NicolasDorier:
@TChukwuleta can you fix conflicts and update the PR description then?

TChukwuleta:
Hey @NicolasDorier  I have updated the PR description and also resolved the conflict

BTCPayServer/Controllers/UIStoresController.Users.cs
```diff
@@ -74,7 +74,9 @@ public async Task<IActionResult> StoreUsers(string storeId, StoreUsersViewModel
                     (await _userManager.CreateAsync(user)) is { Succeeded: true })
                 {
                     var invitationEmail = await _emailSenderFactory.IsComplete();
+                    var monetizationEnabled = _monetizationSettings.Settings.IsSetup();
                     var evt = (UserEvent.Invited)await UserEvent.Registered.Create(user!, currentUser, _callbackGenerator, invitationEmail);
+                    evt.SkipMonetization = !monetizationEnabled;
```
NicolasDorier: Why do you need this? If Monetization is enabled, then no monetization skipped => this is the default already.
If Monetization is not enabled, we don't care about this property.

BTCPayServer/Controllers/UIServerController.Users.cs
```diff
@@ -234,6 +237,7 @@ public async Task<IActionResult> CreateUser()
         public async Task<IActionResult> CreateUser(RegisterFromAdminViewModel model)
         {
             await PrepareCreateUserViewData();
+            ViewData["MonetizationEnabled"] = _monetizationSettings.Settings.IsSetup();
```
NicolasDorier: Why using a `ViewData`, why not just adding a property to `RegisterFromAdminViewModel`?

NicolasDorier (changes requested):

BTCPayServer.Tests/MonetizationTests.cs
```diff
@@ -217,13 +217,35 @@ public async Task CanMonetizeServer()
...
+    public async Task CanSkipMonetizationOnInvite()
+    {
+        await using var s = CreatePlaywrightTester(newDb: true);
+        await s.StartAsync();
+        await s.RegisterNewUser(true);
+        await s.CreateNewStore();
         await GoToMonetization(s);
-        await GoToOffering(s);
-        await offeringPMO.GoToSubscribers();
-        await offeringPMO.AssertHasNotSubscriber("normal-guest@gmail.com");
```
NicolasDorier, line 224: Why did you delete this test?

NicolasDorier (changes requested):
@TChukwuleta I am just thinking of a more fundamental problem with this PR related to UX.

### Problem 1: If the admin disable monetization and re-monetize

If the administrator wants to change the migration's offering OR disable temporarily monetization, the user will get monetized again.

### Problem 2: Ad-hoc demonetization

Consider:
1. Ambassador opens free registration
2. His friend Nic create a new account
3. Ambassador decides to monetize the server
4. His friend Nic now need to pay
5. But Nic is his friend! He doesn't want to make him pay.

In this situation, the Ambassador doesn't have a way to demonetize his friend Nic after the fact.

### Solution

First, whether a user should bypass monetization should be persisted in a new column.

Second, the "Edit user" page, a checkbox should be added about whether or not monetization bypass is activated for such user.
If Monetization isn't activated, then this is just an update in the database.
However, if Monetization is activated, you need to properly handle this in the MonetizationHostedService, you need to unlock him (with UpdateUserLockout).
Also you need to adapt the queries so you are not migrating users with this flag on.

### Implementation note

I believe `NonAdminUserQuery` should be renamed to `AllUsersQuery`, and the SQL query in `MigrateUsers` should filter  out users in `users_to_migrate` who are either admin or `BypassMonetization`. I believe you can then remove the check to `userService.IsAdminUser` as it won't be needed anymore since the query itself filter the admin out.

TChukwuleta:
Hey @NicolasDorier 


Thanks for the feedback, I have taken out time to checkout several use case to this and pushed changes. Here's what was implemented:

Added BypassMonetization column to ApplicationUser — persists the bypass flag so it survives monetization being disabled and re-enabled.

Renamed NonAdminUserQuery to AllUsersQuery and added AND u."BypassMonetization" = false to the SQL filter so that bypassed users are excluded from bulk migration. 

Added BypassMonetization checkbox to the Edit User page so the admin can exempt a user 

When bypass is toggled on Edit User, a BypassMonetizationChanged event is published and handled in MonetizationHostedService:

Bypass added => unlock user if they have a subscriber record
Bypass removed, no subscriber => enroll them via MigrateUsers
Bypass removed, has subscriber => re-evaluate lockout based on subscription state


MonetizationLoginExtension returns early when BypassMonetization = true.

Removed SkipMonetization from UserEvent.Invited, BypassMonetization on the user is now the source of truth.



I kept the IsAdminUser check... the SQL only filters admins during bulk migration, not when a single user registers.

PR #7320, opened 2026-04-15
feat: allow max stores per user

### LINKED ISSUE #7311: Max stores per user
### Related issue in https://github.com/btcpayserver/btcpayserver/issues/7302

<sup>Originally posted by **robwoodgate** July  2, 2025</sup>
As a server admin, I would like to be able to set the maximum number of stores a user can have, ideally as a global default setting, with the ability to override this on a user by user basis.

Eg, under `Server Settings` (Policies), there could be an `Users` option something like _"Non-admins can have this many stores"_ with an integer input. Then under `Users`, there could be an override option to set quota on a per user basis.</div>

Abhijay007:
closes #7311 

https://github.com/user-attachments/assets/5dc0f69e-7f25-4889-86b3-7dce8b1f1227

### DIFF
--- a/BTCPayServer.Client/Models/UpdateApplicationUserRequest.cs
+++ b/BTCPayServer.Client/Models/UpdateApplicationUserRequest.cs
@@ -27,3 +27,4 @@ public class UpdateApplicationUserRequest
     /// </summary>
     public string NewPassword { get; set; }
 }
+
--- a/BTCPayServer/Controllers/UIUserStoresController.cs
+++ b/BTCPayServer/Controllers/UIUserStoresController.cs
@@ -7,9 +7,11 @@ using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client;
 using BTCPayServer.Data;
 using BTCPayServer.Models.StoreViewModels;
+using BTCPayServer.Services;
 using BTCPayServer.Services.Rates;
 using BTCPayServer.Services.Stores;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.Extensions.Localization;
@@ -23,18 +25,24 @@ namespace BTCPayServer.Controllers
         private readonly IStringLocalizer StringLocalizer;
         private readonly DefaultRulesCollection _defaultRules;
         private readonly RateFetcher _rateFactory;
+        private readonly PoliciesSettings _policiesSettings;
+        private readonly UserManager<ApplicationUser> _userManager;
         public string CreatedStoreId { get; set; }
 
         public UIUserStoresController(
 			DefaultRulesCollection defaultRules,
             StoreRepository storeRepository,
             IStringLocalizer stringLocalizer,
-            RateFetcher rateFactory)
+            RateFetcher rateFactory,
+            PoliciesSettings policiesSettings,
+            UserManager<ApplicationUser> userManager)
         {
             _repo = storeRepository;
             StringLocalizer = stringLocalizer;
             _defaultRules = defaultRules;
             _rateFactory = rateFactory;
+            _policiesSettings = policiesSettings;
+            _userManager = userManager;
         }
 
         [HttpGet]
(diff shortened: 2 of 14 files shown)


Abhijay007:
@coderabbitai review this PR

Abhijay007:
cc : @dstrukt

BTCPayServer/Controllers/GreenField/GreenfieldUsersController.cs
```diff
@@ -415,6 +415,56 @@ public async Task<IActionResult> CreateUser(CreateApplicationUserRequest request
             return CreatedAtAction(string.Empty, model);
         }
 
+        [Authorize(Policy = Policies.CanModifyServerSettings, AuthenticationSchemes = AuthenticationSchemes.Greenfield)]
+        [HttpPut("~/api/v1/users/{idOrEmail}")]
+        public async Task<IActionResult> UpdateUser(string idOrEmail, AdminUpdateApplicationUserRequest request)
```
NicolasDorier: Can we keep that out of the PR?
This would deserve proper attention on its own.

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -428,10 +438,48 @@ private async Task<bool> EnsureRemainingOwner(DbSet<UserStore> userStore, string
...
+                    StoreDataId = storeData.Id,
+                    ApplicationUserId = ownerId,
+                    StoreRoleId = StoreRoleId.Owner.Id,
+                };
+                ctx.Add(storeData);
+                ctx.Add(userStore);
+                await ctx.SaveChangesAsync();
+                await tx.CommitAsync();
+                _eventAggregator.Publish(new StoreUserEvent.Added(storeData.Id, ownerId, StoreRoleId.Owner.Id));
+                _eventAggregator.Publish(new StoreEvent.Created(storeData));
```
NicolasDorier: Bad idea. Since the transaction is serializable, this has high change on being triggered twice.
The events should be publish outside the transaction.

BTCPayServer/Services/Stores/StoreRepository.cs
```diff
@@ -428,10 +438,48 @@ private async Task<bool> EnsureRemainingOwner(DbSet<UserStore> userStore, string
                 store.ApplicationUserId != userId);
         }
 
+        public async Task<bool> TryCreateStore(string ownerId, StoreData storeData, int? quota)
```
NicolasDorier: This one need to be rewritten.

* You shouldn't use `strategy.Execute` because EF db context get corrupted if one try fails.
* You should publish event only when this has committed successfully
* You need to try/catch failure on the SaveChange (postgres error 40001) and retry the whole thing again if it failed, included the creation of the context (you can use `goto` for that, this is easier to read). This is due to the isolation mode Serializable that you are using. (this can, and will fail, and need a retry)

NicolasDorier (changes requested):

NicolasDorier:
@Abhijay007 can you do the three points that coderabbit suggest? Previously you added a route to change the user data, but we already have one it seems `UpdateApplicationUserRequest`, can you just add the property there?

BTCPayServer/Controllers/UIUserStoresController.cs
```diff
@@ -112,6 +160,13 @@ public async Task<IActionResult> CreateStore(CreateStoreViewModel vm)
             });
         }
 
+        private async Task<int?> GetEffectiveStoreLimitAsync()
```
NicolasDorier, line 155: Instead of guarding everything with `User.IsInRole(Roles.ServerAdmin)`, maybe you should just add the check inside `GetEffectiveStoreLimitAsync` and return null if the user is admin.

BTCPayServer.Client/Models/UpdateApplicationUserRequest.cs
```diff
@@ -26,4 +26,10 @@ public class UpdateApplicationUserRequest
     /// new password of the user
     /// </summary>
     public string NewPassword { get; set; }
+
+    /// <summary>
+    /// per-user override for the max number of stores this user can create. Null means the server default applies. Only settable by admins.
+    /// </summary>
+    public int? StoreQuota { get; set; }
```
NicolasDorier: You haven't hooked this to the greenfield API. Also need a test.
Your test doesn't check that the limit set to specific user is actually applied.

NicolasDorier (changes requested):

NicolasDorier (changes requested):
Need test, and small cleanup

NicolasDorier:
1. The tests needed to be in a separate DB because we pollute the server settings.
2. I rewrote the quota calculation in SQL, the code with SERIALIZABLE was tricky.

Issue #7322, opened 2026-04-16
[Bug]: BOLT11 amount-less top-up invoices


notraiday:
### What is your BTCPay version?

I'm running BTCPay Server v2.3.7 + clightning

### How did you deploy BTCPay Server?

docker

### What happened?

If you create top-up invoice(none amount) in btcpayserver with BTC-LN payment method enabled it gives amount-less BOLT11 address which is fine. But if this invoice is paid using BTC Lightning btcpayserver doesn't mark one as settled. 
Problem is not with clightning, it sees transfer and mark invoice as paid. But not btcpayserver
There's no errors or any appearance of invoice payment in btcpayserver logs. 

### How did you encounter this bug?

1. Create top-up(none amount) btcpayserver invoice with BTC-LN enabled
2. Pay it
3. Be upset

### Relevant log output

```shell

```

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

PR #7323, opened 2026-04-16
fix: use AmountReceived instead of Amount in lightning invoices

notraiday:
Issue: https://github.com/btcpayserver/btcpayserver/issues/7322

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningListener.cs
+++ b/BTCPayServer/Payments/Lightning/LightningListener.cs
@@ -549,7 +549,7 @@ namespace BTCPayServer.Payments.Lightning
                         notification.GetPaymentHash(_network.NBitcoinNetwork) == GetPaymentHash(listenedInvoice)))
                     {
                         if (notification.Status == LightningInvoiceStatus.Paid &&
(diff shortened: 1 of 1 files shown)


BTCPayServer/Payments/Lightning/LightningListener.cs
```diff
@@ -622,6 +622,15 @@ public async Task<bool> AddPayment(LightningInvoice notification, string invoice
             if (notification?.PaidAt is null || invoiceEntity is null)
                 return false;
 
+            var paidAmount = notification.AmountReceived ?? notification.Amount;
```
Kukks, line 625: Should we also just fall back to the invoice(bolt11) min amount?

Kukks: ah yes, we did those unintuitive 0 amount invoices  so this would break anyway

notraiday: Yeah, for these invoices it's all zero anyway. I think the node-reported received amount is the only reliable source here. And if the BOLT11 amount is set, it can't be overridden at payment time anyway.

PR #7329, opened 2026-04-22
Add granular wallet permissions and move wallet to plugin 

thgO-O:
## Summary

This PR moves the on-chain wallet UI into the Wallets plugin area and introduces granular wallet permissions for wallet access, settings, and transaction workflows.

Instead of relying mostly on broad store-settings access, wallet routes now use wallet-specific policies for viewing, configuring, creating PSBTs, signing, broadcasting, canceling pending transactions, and managing wallet metadata.

## Changes

- Moves wallet controllers/views/routes into the `Wallets` plugin area.
- Adds granular wallet permissions:
  - `CanViewWallet`
  - `CanManageWalletSettings`
  - `CanManageWalletTransactions`
  - `CanCreateWalletTransactions`
  - `CanSignWalletTransactions`
  - `CanBroadcastWalletTransactions`
  - `CanCancelWalletTransactions`
- Adds default wallet-focused roles:
  - Wallet Manager
  - Multisigner
  - Multisigner Guest
- Adds command-level authorization for send, bump fee, PSBT, pending transaction, broadcast, cancel, labels/comments, settings, seed, import/generate, replace/delete, rescan, and history actions.
- Keeps seed and hot-wallet key material behind the existing store-settings permission.
- Prevents cross-store wallet setup retargeting by using route `storeId`/`cryptoCode` as the source of truth.
- Updates wallet UI controls and redirects so role-scoped users only see actions they can complete.
- Adds test coverage for wallet role behavior and pending transaction broadcast matching.


BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.PSBT.cs
```diff
@@ -576,15 +584,18 @@ public async Task<IActionResult> WalletPSBTReady(
                             TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Transaction broadcasted successfully ({0})", transaction.GetHash()].Value;
                         }
 
-                        if (!string.IsNullOrEmpty(vm.SigningContext?.Comment))
+                        if (!string.IsNullOrEmpty(vm.SigningContext?.Comment) &&
+                            await AuthorizeWalletAsync(walletId, Policies.CanManageWalletTransactions))
                         {
                             var txObjId = new WalletObjectId(walletId, WalletObjectData.Types.Tx, transaction.GetHash().ToString());
                             await WalletRepository.SetWalletObjectComment(txObjId, vm.SigningContext.Comment);
                         }
```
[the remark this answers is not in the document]
thgO-O: I’m leaving this as-is. The comment field is already hidden from users without CanManageWalletTransactions, and the send controller only copies vm.Comment into SigningContext when that permission is present.

So broadcast-only users do not have a normal UI path to enter a comment here. This check is just defense in depth for crafted POSTs.

BTCPayServer.Data/Migrations/20260203123000_AddWalletRoles.cs
```diff
@@ -0,0 +1,36 @@
...
+                SELECT 'Wallet Manager', 'Wallet Manager', ARRAY['btcpay.store.canmanagewallets']::TEXT[]
+                WHERE NOT EXISTS (SELECT 1 FROM "StoreRoles" WHERE "Id" = 'Wallet Manager');
+
+                INSERT INTO "StoreRoles" ("Id", "Role", "Permissions")
+                SELECT 'Multisigner', 'Multisigner', ARRAY['btcpay.store.canmanagewallettransactions']::TEXT[]
+                WHERE NOT EXISTS (SELECT 1 FROM "StoreRoles" WHERE "Id" = 'Multisigner');
+
+                INSERT INTO "StoreRoles" ("Id", "Role", "Permissions")
+                SELECT 'Multisigner Guest', 'Multisigner Guest', ARRAY['btcpay.store.cansigntransactions']::TEXT[]
+                WHERE NOT EXISTS (SELECT 1 FROM "StoreRoles" WHERE "Id" = 'Multisigner Guest');
```
NicolasDorier: Can you instead do like `20240229092905_AddManagerAndEmployeeToStoreRoles.cs`?

Btw, this code also isn't necessarily wrong, but idempotent insert are better served with `ON CONFLICT DO NOTHING`.

BTCPayServer/Data/Payouts/BitcoinLike/BitcoinLikePayoutHandler.cs
```diff
@@ -315,11 +315,14 @@ public async Task<IActionResult> InitiatePayment(string[] payoutIds)
             }
         }
         if (bip21.Any())
-            return new RedirectToActionResult("WalletSend", "UIWallets", new { walletId = new WalletId(storeId, Network.CryptoCode).ToString(), bip21 });
-        return new RedirectToActionResult("Payouts", "UIWallets", new
+            return new RedirectToActionResult("WalletSend", "UIWallets", new { area = "Wallets", walletId = new WalletId(storeId, Network.CryptoCode).ToString(), bip21 });
```
NicolasDorier: Can you follow the pattern we use in all plugins? Put the string in a const, like `WalletsPlugin.Area`

BTCPayServer/Plugins/Wallets/Controllers/UIStoreOnChainWalletsController.cs
```diff
@@ -42,14 +103,24 @@ public async Task<ActionResult> SetupWallet(WalletSetupViewModel vm)
 
         var perm = await CanUseHotWallet();
         vm.SetPermission(perm);
+        vm.CanGenerateNewWallet = await CanAccessWalletSeedMaterial(vm.StoreId) &&
+                                  (vm.CanUseHotWallet || vm.CanCreateNewColdWallet);
 
-        return View(vm);
+        return View("SetupWallet", vm);
```
NicolasDorier: `nameof(SetupWallet)` which you are at it.

BTCPayServer/Plugins/Wallets/Controllers/UIStoreOnChainWalletsController.cs
```diff
@@ -437,13 +541,23 @@ public async Task<IActionResult> WalletSettings(string storeId, string cryptoCod
...
-
     [HttpPost("{storeId}/onchain/{cryptoCode}/settings/wallet")]
-    [Authorize(Policy = Policies.CanModifyStoreSettings, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
-    public async Task<IActionResult> UpdateWalletSettings(WalletSettingsViewModel vm)
+    public async Task<IActionResult> UpdateWalletSettings(
+        WalletSettingsViewModel vm,
+        [FromRoute] string storeId = null,
+        [FromRoute] string cryptoCode = null)
     {
+        if (RouteValueMismatch(vm.StoreId, storeId, StringComparison.Ordinal) ||
```
NicolasDorier: is `RouteValueMismatch` neceesarry given we set the vm.StoreId to storeId just below.

NicolasDorier (changes requested):

BTCPayServer/Components/MainNav/Default.cshtml
```diff
@@ -235,7 +239,8 @@
                     </div>
                 </div>
             }
-            <div class="accordion-item">
+            <div class="accordion-item"
+                 permission="@($"{Policies.CanViewStoreSettings},{Policies.CanModifyServerSettings}")">
```
NicolasDorier: This shouldn't be guarded, this would prevent plugins to show anything if CanViewStoreSettings isn't set... but that's up to the plugin itself to decide whether the permission is required or not.

NicolasDorier:
one test fail

BTCPayServer.Client/Permissions.cs
```diff
@@ -21,6 +21,14 @@ public class Policies
         public const string CanViewInvoices = "btcpay.store.canviewinvoices";
         public const string CanCreateInvoice = "btcpay.store.cancreateinvoice";
         public const string CanModifyInvoices = "btcpay.store.canmodifyinvoices";
+        public const string CanManageWallets = "btcpay.store.canmanagewallets";
```
NicolasDorier: Move those permissions to the Wallets plugin directory.

BTCPayServer/Components/MainNav/Default.cshtml
```diff
@@ -94,26 +94,29 @@
...
                         <div text-translate="true" class="accordion-button">Wallets</div>
                     </header>
-                    <div id="Nav-Wallets" class="accordion-collapse" aria-labelledby="Nav-Wallets-Header" permission="@Policies.CanModifyStoreSettings">
+                    <div id="Nav-Wallets" class="accordion-collapse" aria-labelledby="Nav-Wallets-Header" permission="@Policies.CanViewWallet">
                         <div class="accordion-body">
                             <ul class="navbar-nav">
                                 @foreach (var scheme in Model.DerivationSchemes.OrderBy(scheme => scheme.Collapsed))
                                 {
                                     var isSetUp = !string.IsNullOrWhiteSpace(scheme.Value);
+                                    const string walletViewPermission = Policies.CanViewWallet;
```
NicolasDorier: Why do you need variables here? Just use the constant directly.

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -130,16 +133,27 @@ public UIStoresController(
...
     {
+        var userId = GetUserId();
+        if (userId is null)
+            return Forbid();
+        var store = await _storeRepo.FindStore(storeId, userId);
+        if (store is null)
+            return NotFound();
+        // Keep selected store in context/cookie even for limited roles that will be redirected away from dashboard to wallets.
+        HttpContext.SetStoreData(store);
+        HttpContext.SetPreferredStoreId(storeId);
```
NicolasDorier: So the problem of doing this is that if the user has no CanViewInvoices and no CanModifyStoreSettings, then he is stuck is seeing the "Forbid" page forever and can't navigate out of that... because it's the preferred store!

BTCPayServer/Controllers/UIStoresController.cs
```diff
@@ -130,16 +133,27 @@ public UIStoresController(
...
         {
-            HttpContext.SetPreferredStoreId(storeId);
             return RedirectToAction("Dashboard", new { storeId });
         }
         if ((await _authorizationService.AuthorizeAsync(User, Policies.CanViewInvoices)).Succeeded)
         {
-            HttpContext.SetPreferredStoreId(storeId);
             return RedirectToAction("ListInvoices", "UIInvoice", new { storeId });
         }
+        if (store.HasPolicy(userId, Policies.CanViewWallet, _permissionService))
```
NicolasDorier: use the `authorizationService`

BTCPayServer/Plugins/Wallets/Views/UIWallets/WalletSend.cshtml
```diff
@@ -188,7 +192,7 @@
         }
     </div>
 
-    <div class="form-group">
+    <div class="form-group" permission="@walletManageTxPermission" and-mode="true">
```
NicolasDorier: why using `and-mode` ? there is a single permission.

BTCPayServer/Plugins/Wallets/Views/UIWallets/WalletTransactions.cshtml
```diff
@@ -13,6 +17,12 @@
     var sortedLabels = Model.Labels.OrderBy(l => l.Text, StringComparer.OrdinalIgnoreCase).ToList();
     var popularLabels = Model.PopularLabels.OrderBy(l => l.Text, StringComparer.OrdinalIgnoreCase).ToList();
     var initialLabels = popularLabels.Any() ? popularLabels : sortedLabels.Take(maxVisibleLabels).ToList();
+    var walletViewPermission = Policies.CanViewWallet;
+    var walletCreateTxPermission = Policies.CanCreateWalletTransactions;
+    var walletSettingsPermission = Policies.CanManageWalletSettings;
+    var walletBroadcastTxPermission = Policies.CanBroadcastWalletTransactions;
+    var walletCancelTxPermission = Policies.CanCancelWalletTransactions;
+    var walletAnyTxPermission = $"{Policies.CanCreateWalletTransactions},{Policies.CanSignWalletTransactions},{Policies.CanBroadcastWalletTransactions},{Policies.CanCancelWalletTransactions}";
```
NicolasDorier: Is it needed? Those are protecting `ViewPendingTransaction`

BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.cs
```diff
@@ -2044,6 +2134,69 @@ public async Task<IActionResult> EditWalletLabel(
             return null;
         }
 
+        private async Task<bool> AuthorizeWalletAsync(WalletId walletId, params string[] policies)
```
NicolasDorier: I don't understand all this hacks.

NicolasDorier (changes requested):
I think `AuthorizeWalletAsync` and `HasAnyWalletTransactionPermission` are code smell indicating that no permission properly fit the routes that you are attempting to protect.

Rather than hacking around, I believe you should replace all those hacks with a single permission check.

BTCPayServer/Plugins/Wallets/Views/UIWallets/WalletPSBTDecoded.cshtml
```diff
@@ -132,7 +125,7 @@ else if (isReady)
         <div class="d-flex flex-column flex-sm-row flex-wrap justify-content-center align-items-sm-center">
             @if (!string.IsNullOrEmpty(Model.SigningContext?.PayJoinBIP21))
             {
-                <span permission="@walletPayjoinPermission" and-mode="true">
+                <span permission="@($"{WalletPolicies.CanBroadcastWalletTransactions},{WalletPolicies.CanSignWalletTransactions}")" and-mode="true">
```
NicolasDorier, line 128: is it normal that here you need two permissions, but in the controller's route, you only check CanBroadcast?

BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.PSBT.cs
```diff
@@ -654,24 +672,10 @@ public async Task<IActionResult> WalletPSBTCombine([ModelBinder(typeof(WalletIdM
...
-                "payjoin" => AuthorizeWalletAsync(walletId, Policies.CanBroadcastWalletTransactions, Policies.CanSignWalletTransactions),
-                _ => AuthorizeWalletAsync(walletId)
-            };
-        }
-
         private async Task<bool> CanAutoSignWithHotWallet(WalletId walletId, DerivationSchemeSettings derivationSchemeSettings)
         {
             return derivationSchemeSettings?.IsHotWallet is true &&
-                   await AuthorizeWalletAsync(walletId, Policies.CanModifyStoreSettings) &&
+                   (await _authorizationService.AuthorizeAsync(User, walletId.StoreId, Policies.CanModifyStoreSettings)).Succeeded &&
```
NicolasDorier: Why you don't use the `CanSignWalletTransactions` on this one?

NicolasDorier (changes requested):

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -569,6 +570,7 @@ CREATE INDEX IF NOT EXISTS idx_invoices_expired_cleanup
                         Policies.CanModifyWebhooks,
                         Policies.CanModifyPaymentRequests,
                         Policies.CanManagePayouts,
+                        WalletPolicies.CanManageWallets,
```
NicolasDorier: Remove it from here, you can add this in the PolicyDefinition of `CanManageWallets` with the property `IncludedByPermissions`.

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -29,6 +29,7 @@
 using BTCPayServer.Payments.PayJoin;
 using BTCPayServer.PayoutProcessors;
 using BTCPayServer.Plugins;
+using BTCPayServer.Plugins.Wallets;
```
NicolasDorier: Can then remove this after https://github.com/btcpayserver/btcpayserver/pull/7329/changes#r3185577981

BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.PSBT.cs
```diff
@@ -576,15 +607,18 @@ public async Task<IActionResult> WalletPSBTReady(
...
                             var txObjId = new WalletObjectId(walletId, WalletObjectData.Types.Tx, transaction.GetHash().ToString());
                             await WalletRepository.SetWalletObjectComment(txObjId, vm.SigningContext.Comment);
                         }
 
                         if (vm.SigningContext.PendingTransactionId is not null)
                         {
-                            await _pendingTransactionService.Broadcasted(GetPendingTxId(walletId, vm.SigningContext.PendingTransactionId));
+                            await _pendingTransactionService.Broadcasted(
+                                GetPendingTxId(walletId, vm.SigningContext.PendingTransactionId),
+                                transaction.GetHash().ToString());
```
NicolasDorier: Are you sure about this? I don't remember everything, but I think the `PendingTransactionId` is different from the tx id on purpose.

This is because bitcoin transaction ids are malleable. If you sign from a legacy wallet, the transaction's hash you are passing here is probably not the same as the one you saved in the database.

NicolasDorier (changes requested):
You also need to remove `EnsureWalletStoreContextAsync`. This set the store data, which bypass permissions checks that should have checked by ASP.NET prior to calling the route.

BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.PSBT.cs
thgO-O: Yep.. you were right. I added the check to prevent a tampered PendingTransactionId from marking an unrelated pending tx as broadcast.

I changed it to avoid txid comparison: now it compares the final transaction intent with the stored pending PSBT, ignoring scriptSig/witness for malleability.

PR #7333, opened 2026-04-24
Add cache to circle ci

NicolasDorier:


NicolasDorier:
I failed to make it faster. I tried:
* Docker layer cache
* CircleCI cache for nuget dependencies

The upside was very limited for the added complexity, so I decided to give up on this.

PR #7345, opened 2026-04-30
feat: add separate CanSendStoreEmail permission for store email API

### LINKED ISSUE #6159: Add separate permission for sending store email through API
We have SendStoreEmail API, currently it needs canmodifystoresettings permission which is too broad for that task, if api key is leaked or integration compromised. We should have a separate permission only for sending store email.

Abhijay007:
closes : https://github.com/btcpayserver/btcpayserver/issues/6159


Abhijay007:
cc : @pavlenex, @NicolasDorier

Issue #7341, opened 2026-04-30
Download of Translation Dictionaries doesn't work


mp21-org:
> Currently Downloads of i.e. the spanish and german language pack leads to a 404 error - is this related?
> see: https://t.me/btcpayserver/97503
> 
> <img width="686" height="66" alt="Image" src="https://github.com/user-attachments/assets/41dabc23-3797-48a9-ae28-7b298d05b336" />
> 
> cc @rockstardev  

 _Originally posted by @petzsch in [#7331](https://github.com/btcpayserver/btcpayserver/issues/7331#issuecomment-4341526233)_

It seems that this issue is unrelated to #7331

mp21-org:
@NicolasDorier fixed this: https://github.com/btcpayserver/btcpayserver-translator/commit/2522bc9d9f8ac61646f1b56efed026f3a58fc558

PR #7347, opened 2026-05-02
Server-side translation manager: install, update, and remove language packs from the BTCPay UI

### LINKED ISSUE #7331: Replace /server/dictionaries with /server/translations: install/uninstall language packs from upstream translator repo
This issue tracks Phase 1 of the [BTCPay Server translations revamp roadmap](https://docs.google.com/document/d/1mO0MuWrZ-vwSpN1qkXj2aREj62vG2DEam4yYDN10gu4/edit).

## Goal

Replace the current `/server/dictionaries` raw CRUD surface with a `/server/translations` page that:

- Lists **installed** languages with maintainer attribution (handle + GitHub link), native name, last-updated date.
- Lists languages **available to install** from `btcpayserver-translator` upstream.
- One-click install and update from upstream.
- No popup or multi-step modal.

## Data source

Single `manifest.json` at the root of `btcpayserver-translator`, regenerated by CI on every merge to `main`:

```json
{
  "languages": [
    { "code": "fr",
      "bcp47": "fr-FR",
      "name": "French",
      "native": "Francais",
      "file": "translations/french.json",
      "sha": "...",
      "maintainer": "teamssUTXO|https://github.com/teamssUTXO",
      "updated": "2026-04-12T10:30:00Z" }
  ],
  "_redirect": null
}
```

The server fetches `manifest.json` once per hour (cached). Individual language files are fetched only on user-initiated install. The `bcp47` field is what the future per-user dropdown will match against `Accept-Language`.

## Backwards compatibility

For BTCPay versions hardcoded to fetch `translations/<lang>.json` directly (pre-manifest), the static files at the old URLs remain in place. New versions read from `manifest.json` and follow its entries.

The manifest top-level `_redirect` field is reserved for a future move to a standalone translations repo - clients follow it without coordinated server-side changes. The default manifest URL inside BTCPay Server source is updated on each release so the redirect is a fallback, not the primary path.

## Dependencies

This page depends on these `btcpayserver-translator` PRs landing first:

- [`#47`](https://github.com/btcpayserver/btcpayserver-translator/pull/47) (1amKhush): core language-pack validation. Catches malformed language packs at install time.
- [`#49`](https://github.com/btcpayserver/btcpayserver-translator/pull/49) (teamssUTXO): unit tests + CI workflow. Foundation for the manifest.json regeneration. Per its description, merges after #47.

Both bundled into Phase 1 of the roadmap so the install path has working validation + the manifest comes online with CI behind it.

## Target

Phase 1 target completion: **2026-05-04** (2 weeks from 2026-04-23).

teamssUTXO:
This PR turns the existing translation editor into a full
language-pack manager that store admins can use without leaving
the BTCPay UI.

## What changes for users

- A new **Translations** page (Server Settings > Translations)
  shows every translation BTCPay knows about - the built-in ones
  ("English", "fallback") plus every language available from the
  community translator at
  https://github.com/btcpayserver/btcpayserver-translator.
- For each available language, admins can:
  - **Install** it with a single click. The pack downloads from
    the translator repo, gets a SHA-256 integrity check, and is
    stored locally.
  - **Set as default** so checkout, emails, and the dashboard
    use it.
  - **Update** when the maintainer publishes a newer version. An
    "Update available" badge appears next to packs that are
    behind.
  - **Uninstall** (with the typed-confirmation modal restored
    in this PR after the prior refactor dropped it).
- Maintainer attribution is shown on the page (handle + link),
  so admins can see who's responsible for the translation they
  install.
- If the translator repo is unreachable, the page degrades
  gracefully: installed translations still work and are listed,
  the "available to install" list just shows a note that the
  manifest could not be fetched.

## Why

Translations were previously edited through a generic dictionary
UI that did not know anything about language-pack lifecycle.
Operators had to clone the translator repo, hand-edit JSON, and
restart - or wait for a new BTCPay release to ship updated
translations. This PR closes that loop in the running server.

## What stays the same

- Existing custom translations created via the dictionary editor
  keep working - they are tagged Source = "Custom" and behave as
  before.
- Built-in translations cannot be uninstalled (server-side guard
  added).
- Old `/server/dictionaries/...` URLs continue to work via a
  catch-all redirect, so any existing bookmarks or external docs
  remain valid.
- API permissions are unchanged. (An earlier draft consolidated
  CanSendStoreEmail into CanModifyStoreSettings; that was reverted
  to keep this PR scoped to translations.)

## Notes for reviewers

- Manifest fetches from the translator repo are cached for 1 hour
  and back off for 60 seconds after a failure, so the Translations
  page is cheap to refresh.
- Downloaded language packs are SHA-256 verified against the
  manifest before being persisted - rejected on mismatch (test
  in BTCPayServer.Tests/LanguageServiceTests.cs).
- Closes #7331.

## UI
new_ui_translations-btcpay


rockstardev (changes requested):
Solid design + strong test foundation; concrete asks below before merge.

## MAJOR

1. **[security] SHA verification missing.** `LanguagePackUpdateService.FetchLanguagePackFromRepository()` (BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs lines around the new `RawBaseUrl + filePath` fetch) returns the manifest's `Sha` field as the version but never hashes the downloaded translation content to verify it. The manifest's `Sha` is SHA-256 of the source file content (verified live: matches `sha256sum` on dutch.json + french.json), so the trust anchor is in place - it just isn't checked. Fix: compute SHA256 over the downloaded bytes, compare case-insensitive against `entry["Sha"]`, reject on mismatch + add a unit test covering the rejection path. Defense-in-depth against MITM / CDN tampering between btcpay-translator main and the user's server.

2. **[scope] `CanSendStoreEmail` permission removal is unrelated to the translations refactor.** BTCPayServer.Client/Permissions.cs:18 deletes the constant; BTCPayServer/Hosting/BTCPayServerServices.cs lines around 549-573 remove the policy definition and its inclusion in the `CanModifyStoreSettings` umbrella; BTCPayServer/Controllers/GreenField/GreenfieldStoreEmailController.cs:30 elevates the endpoint authorization. This is a breaking API change for keys holding the narrow `btcpay.store.cansendstoreemails` permission; it scope-expands email-send to require full store-settings access. Ask: split into a separate PR, or amend this PR's title/description to disclose the permission removal as a breaking change. The translations refactor and the permission consolidation should not co-merge silently.

3. **[compat] Backward-compat shim covers only one of 8+ old routes.** UITranslationController.cs only adds `[HttpGet("server/dictionaries")] -> RedirectToActionPermanent(ListTranslations)`. The old route table also exposed GET/POST `server/dictionaries/create`, GET/POST `server/dictionaries/{name}`, POST `server/dictionaries/{name}/select`, POST `server/dictionaries/{name}/delete`, POST `server/dictionaries/download`. Under the new controller all of these 404. The PR description claims "preserve existing integrations and prevent breaking changes for legacy clients" - if the intent is that only the list page is preserved, please amend the description; if the intent is full preservation, expand the shim (probably easiest as a small `[Route("server/dictionaries/{**catchall}")]` action that 308-redirects to the equivalent translations URL).

4. **[ux/safety] Delete confirmation modal removed.** Old `ListDictionaries.cshtml` line ~75 used the `_Confirm` partial with `data-confirm-input="Delete"` requiring the user to type "Delete" to enable the destructive action. New `ListTranslations.cshtml:115` is a direct `<button type="submit">` with no confirmation - one accidental click on the wrong row uninstalls a custom translation pack (which is irreversible without re-import + can be hours of user-edited entries). Restore the typed-confirmation modal on the Uninstall action.

## MINOR

5. **Redundant routes.** UITranslationController.cs has both `[HttpPost("server/translations/{translation}/delete")] DeleteTranslation` and `[HttpPost("server/translations/{translation}/uninstall")] UninstallLanguagePack`. Both call `localizer.DeleteTranslation(translation)` with identical TempData; the view only references Uninstall. Pick one and remove the other, or differentiate semantics (e.g. uninstall = remove downloaded pack only, delete = full row).

6. **Catch-all exception in `CheckForLanguagePackUpdate`.** LanguagePackUpdateService.cs around the modified catch block now catches `Exception`; old version caught `HttpRequestException` + `TaskCanceledException`. Loses information about *why* a check fails (network vs malformed manifest vs missing entry); narrower catch is preferable.

7. **No back-off on manifest-fetch failure.** `FetchManifest()` in LanguagePackUpdateService.cs retries on every translations-page load. Under sustained upstream outage this hammers the source. Suggest: cache the failure with a short TTL (e.g. 60s) so transient outages back off.

8. **Test coverage gaps.** BTCPayServer.Tests/LanguageServiceTests.cs new tests cover happy path + degraded mode + cache invalidation; please also add (a) malformed manifest JSON returns degraded, (b) missing `Languages` key returns degraded, (c) `FetchLanguagePackFromRepository` rejects unknown language with `ArgumentException`, (d) **SHA mismatch rejection** (depends on fix to MAJOR #1), (e) the backward-compat redirect from `GET /server/dictionaries` to `ListTranslations`.

## Praise

- `PoliciesSettings.cs:32` - `[JsonProperty("LangDictionary", DefaultValueHandling = DefaultValueHandling.Populate)]` keeps existing serialized settings loadable while the C# property is renamed to `LangTranslation`. Clean data-shape preservation; CR-caught, fix is correct.
- `LanguagePackUpdateService.FetchLanguagePackFromRepository` returning manifest-`Sha` as version (vs old behavior of recomputing SHA256 of full file every time): right design, dramatic efficiency win on update-checks (no full-download needed for the comparison). Verified live the manifest's `Sha` is SHA-256 of source file content.
- `FetchManifest()` double-check locking - cache check before AND after acquiring `SemaphoreSlim` is correctly implemented.
- `GetManifestLanguages()` returning `(empty, true)` on exception + controller rendering a graceful degraded UI - fail-discoverable rather than fail-silent.
- `+198` lines in BTCPayServer.Tests/LanguageServiceTests.cs using a clean `StubHttpMessageHandler` pattern. 4/4 passing locally on .NET 10 RC.2; build clean (0 errors).

rockstardev:
@coderabbitai review

rockstardev:
@coderabbitai review

rockstardev:
@coderabbitai resume

rockstardev:
@coderabbitai review

NicolasDorier:
@rockstardev review :trollface:

NicolasDorier:
* Do not try/catch in `GetManifestLanguages`, let the error propagates and callers handle it.
* Remove `GetAvailableLanguages` (unused, but by test)
* Use a cache for `GetManifestLanguages` (1H?) this isn't a data changing a lot.
* Make the PR description and title human readable, especially by non tech. What is it trying to achieve? This is a functional change, not just some internal tech stuff.

NicolasDorier:
Links to `Translator/translations` are now broken due to fix https://github.com/btcpayserver/btcpayserver/issues/7341

teamssUTXO:
Sorry, I mistakenly rebased instead of merging, which rewrote the commit history. The changes are the same, only the history was affected.

teamssUTXO:
cc : @NicolasDorier

BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs
```diff
@@ -1,75 +1,161 @@
...
-            };
+                if (token is JObject entry &&
+                    string.Equals(entry["Name"]?.ToString(), language, StringComparison.OrdinalIgnoreCase))
+                    return entry;
+            }
+            return null;
         }
 
-        public async Task<(string translationsJson, string version)> FetchLanguagePackFromRepository(string language)
+        private static LanguageManifestEntry ToManifestEntry(JObject entry)
```
NicolasDorier: Instead of doing all this code by hand, can you just use normal NewtonsoftJSON deserialization? (along with some helper methods in the model if you need)

BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs
```diff
@@ -1,75 +1,161 @@
...
+        {
+            if (_entriesCache is { } cached && DateTime.UtcNow - cached.FetchedAt < _cacheExpiration)
+                return cached.Entries;
+
+            var languages = await FetchManifest();
+            var entries = languages.OfType<JObject>()
+                .Select(ToManifestEntry)
+                .Where(e => !string.IsNullOrEmpty(e.Name))
+                .ToArray();
+            _entriesCache = (entries, DateTime.UtcNow);
```
NicolasDorier: Just use `IMemoryCache` of dotnet, don't implement that by youself.

BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs
```diff
@@ -1,75 +1,161 @@
...
+        private (LanguageManifestEntry[] Entries, DateTime FetchedAt)? _entriesCache;
+        private DateTime _manifestNextFetchAllowedAt = DateTime.MinValue;
+        private readonly SemaphoreSlim _manifestLock = new(1, 1);
+
+        private const string ManifestUrl = "https://raw.githubusercontent.com/btcpayserver/btcpayserver-translator/main/manifest.json";
+        private const string RawBaseUrl = "https://raw.githubusercontent.com/btcpayserver/btcpayserver-translator/main/";
+
+        private async Task<JArray> FetchManifest()
+        {
+            if (_manifestCache is { } cached && DateTime.UtcNow - cached.FetchedAt < _cacheExpiration)
```
NicolasDorier: Why there is a different cache that `_entriesCache` ? Isn't one enough?
Need to use IMemoryCache.

BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs
```diff
@@ -1,75 +1,161 @@
...
+                return cached.Languages;
+            if (DateTime.UtcNow < _manifestNextFetchAllowedAt)
+                throw new InvalidOperationException("Manifest fetch is in back-off after a recent failure.");
+
+            await _manifestLock.WaitAsync();
```
NicolasDorier: Why do you need locks?

BTCPayServer/Plugins/Translations/LanguagePackUpdateService.cs
```diff
@@ -1,75 +1,161 @@
...
+                    httpClient.Timeout = TimeSpan.FromSeconds(30);
+                    var json = await httpClient.GetStringAsync(ManifestUrl);
+                    var languages = JObject.Parse(json)["Languages"] as JArray
+                        ?? throw new InvalidOperationException("Manifest is missing the 'Languages' array.");
+                    _manifestCache = (languages, DateTime.UtcNow);
+                    return languages;
+                }
+                catch
+                {
+                    _manifestNextFetchAllowedAt = DateTime.UtcNow + _manifestFailureBackoff;
```
NicolasDorier: Do not backoff, if it fails and we have no cache, just throw.

BTCPayServer/Plugins/Translations/Controllers/UITranslationController.cs
```diff
@@ -132,67 +169,98 @@ public async Task<IActionResult> EditDictionary(string dictionary, EditDictionar
...
+    public IActionResult RedirectToTranslation()
+    {
+        return RedirectToActionPermanent(nameof(ListTranslations));
+    }
+
+    [Route("server/dictionaries/{**catchall}")]
+    public IActionResult RedirectDictionariesSubpath(string catchall)
+    {
+        var newPath = "/server/translations" + (string.IsNullOrEmpty(catchall) ? "" : $"/{catchall}");
+        return RedirectPermanentPreserveMethod(newPath + Request.QueryString);
```
NicolasDorier, line 264: Why did you kept those? This isn't an API, should be concerned with changing route's path.

BTCPayServer/Plugins/Translations/Controllers/UITranslationController.cs
```diff
@@ -25,63 +26,99 @@ public class UITranslationController(
...
 
-        foreach (var dictionary in dictionaries)
+        foreach (var translation in translations)
         {
-            var isSelected = policiesSettings.LangDictionary == dictionary.DictionaryName ||
-                             (policiesSettings.LangDictionary is null && dictionary.Source == "Default");
-            var isDownloadedPack = downloadableLanguages.Contains(dictionary.DictionaryName);
+            var isSelected = policiesSettings.LangTranslation == translation.TranslationName ||
+                             (policiesSettings.LangTranslation is null && translation.Source == "Default");
+            var isDownloadedPack = translation.Source == "Custom";
```
NicolasDorier: Why is `translation.Source` set to "Custom" for downloaded pack?

Is it possible to have a source for downloaded pack like "LanguagePack" instead of "Custom" so we can differentiate them?

NicolasDorier (changes requested):

NicolasDorier:
I've pushed some review. Is there a reason behind the rename of the routes?

The problem is that in the backend we use the "dictionary" terminology, but the frontend is now using "translations", this is kind of confusing.

rockstardev:
@NicolasDorier "Translations" is much more meaningful to users than "Dictionaries".

We originally used "dictionaries" because the system was just a way to translate backend strings. Since then, it has evolved to cover both backend and frontend strings, spanning almost the entire BTCPay Server. As part of this refactoring, it makes sense to complete the transition to Translations.

image

We'll also need to update this description accordingly.

BTCPayServer/Plugins/Translations/Controllers/UITranslationController.cs
rockstardev: I think @teamssUTXO just wanted to preserve it in case we had some undocumented flow. If you're fine with it, we can remove it.

Issue #7348, opened 2026-05-03
Add missing localization markers in Razor UI views


ABC-8888:
## Summary

Several Razor UI views still contain user-facing strings that are not covered by the localization system.

This issue tracks adding missing localization markers in `.cshtml` files using the existing project patterns:
- `text-translate="true"` for static text
- `StringLocalizer[...]` for plain text and formatted strings
- `ViewLocalizer[...]` for strings that contain HTML or links

## Goal

Ensure that user-visible strings in the affected Razor views are picked up by the translation extraction flow and can be translated properly.

## Scope

The following views should be updated:
### UIManage
- `BTCPayServer/Views/UIManage/TwoFactorAuthentication.cshtml`

### UILNURL
- `BTCPayServer/Views/UILNURL/EditLightningAddress.cshtml`

### Shared
- `BTCPayServer/Views/Shared/Lightning/LightningLikeMethodCheckout.cshtml`
- `BTCPayServer/Views/Shared/Lightning/ViewLightningLikePaymentData.cshtml`
- `BTCPayServer/Views/Shared/PostRedirect.cshtml`
- `BTCPayServer/Views/Shared/CameraScanner.cshtml`
- `BTCPayServer/Views/Shared/ConfirmModal.cshtml`
- `BTCPayServer/Views/Shared/LayoutPartials/SyncModal.cshtml`

### UIAccount
- `BTCPayServer/Views/UIAccount/ForgotPassword.cshtml`

### UIWallets
- `BTCPayServer/Views/UIWallets/WalletReceive.cshtml`
- `BTCPayServer/Views/UIWallets/WalletLabels.cshtml`

### UIInvoice
- `BTCPayServer/Views/UIInvoice/Invoice.cshtml`
- `BTCPayServer/Views/UIInvoice/InvoiceReceipt.cshtml`
- `BTCPayServer/Views/UIInvoice/InvoiceReceiptPrint.cshtml`

### UIPaymentRequest
- `BTCPayServer/Views/UIPaymentRequest/GetPaymentRequests.cshtml`
- `BTCPayServer/Views/UIPaymentRequest/PaymentRequestLabels.cshtml`
- `BTCPayServer/Views/UIPaymentRequest/ViewPaymentRequest.cshtml`

### UIServer
- `BTCPayServer/Views/UIServer/ListPlugins.cshtml`

## Notes

When making these updates, please follow the existing conventions carefully:

- Use `text-translate="true"` only for static literal text
[...]

PR #7350, opened 2026-05-03
Add missing localization markers in Razor views

ABC-8888:
This PR should resolve issue https://github.com/btcpayserver/btcpayserver/issues/7348


BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -40,6 +40,16 @@
...
+
+    string LocalizePendingCommand(string command) =>
+        command switch
+        {
+            "install" => StringLocalizer["install"].Value,
+            "update" => StringLocalizer["update"].Value,
+            "delete" => StringLocalizer["delete"].Value,
+            "enable" => StringLocalizer["enable"].Value,
+            _ => command
+        };
```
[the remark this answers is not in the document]
ABC-8888: Thanks. I checked the current code and I’m leaving this unchanged in this PR.
There may be a separate command-token consistency issue here, but it is outside the scope of this PR.

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -516,7 +540,7 @@ else
                             else
                             {
                                 <form asp-action="InstallPlugin" asp-route-plugin="@plugin.Identifier" asp-route-version="@plugin.Version">
-                                    <button title="Schedule install for when the dependencies have been met to ensure a smooth update" data-bs-toggle="tooltip" type="submit" class="btn btn-primary" text-translate="true">Schedule install</button>
+                                    <button title="@StringLocalizer["Schedule install for when the dependencies have been met to ensure a smooth update"]" data-bs-toggle="tooltip" type="submit" class="btn btn-primary" text-translate="true">Schedule install</button>
```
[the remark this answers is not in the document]
ABC-8888: Thanks. This looks like a valid copy issue, but it is outside the scope of this PR.

This PR only adds missing localisation coverage. Changing this existing localised string would also change the translation key, so I’d prefer to handle that separately.

BTCPayServer/Views/UILNURL/EditLightningAddress.cshtml
```diff
@@ -36,9 +36,15 @@
 @if (Context.Request.PathBase.ToString() != string.Empty)
 {
     <div class="alert alert-warning" role="alert">
```
NicolasDorier, line 38: This should be a single translated string, not three different one

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -398,7 +415,7 @@ else
                             @if (!string.IsNullOrEmpty(plugin.Author))
                             {
                                 <span class="text-muted text-nowrap">
-                                    by
```
NicolasDorier, line 401: `by {}` should be a single translation

NicolasDorier (changes requested):

BTCPayServer/Views/UIServer/ListPlugins.cshtml
ABC-8888: Resolved by joining with a AuthorLink and Author.

NicolasDorier:
There are too many slops here and there, it takes too much time to review, please break it down in smaller PRs and double review if the changes make sense,.

PR #7359, opened 2026-05-12
Include expired invoice in Invoice report

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver/discussions/7342

Allow CSV export of expired invoices in reporting

Cc. @NicolasDorier


NicolasDorier:
I am not sure we should do this. Will discuss in the issue.

PR #7369, opened 2026-05-18
Localization of POS keypad labels and invoice statuses

ABC-8888:
https://github.com/user-attachments/assets/66220a97-a6ee-4eb6-ab0b-761d4711e3e8

### DIFF
--- a/BTCPayServer/Extensions/InvoiceStateExtensions.cs
+++ b/BTCPayServer/Extensions/InvoiceStateExtensions.cs
@@ -1,0 +1,41 @@
+using BTCPayServer.Client.Models;
+using BTCPayServer.Services.Invoices;
+using Microsoft.Extensions.Localization;
+
+namespace BTCPayServer
+{
+    public static class InvoiceStateExtensions
+    {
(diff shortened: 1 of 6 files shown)


NicolasDorier:
We have not hooked up translations to public pages such as Point of Sale, because the language should be set at the store level, not server level, (as we do with Checkout).

Unfortunately, we don't have the underlying system in place to connect the server translations with the store's translation. I need to think more about it.

I close this PR, because this isn't hooking up the server's language to the store's language, but doing so isn't something we can do until I've properly thought about it.

pavlenex:
@NicolasDorier How do you explain then that `Charge` call to action is translated in POS at the moment?
Screenshot 2026-05-18 at 18 43 20

NicolasDorier:
@pavlenex I would explain it as a bug, it's something I didn't catch during review.

I am not opposed to use the server's translation for those public pages, but this is confusing as Checkout isn't using it for historical reason. If we start doing this, we will need to revert later when we put it back up to the store's settings.

PR #7372, opened 2026-05-20
Server report for plugin stats

TChukwuleta:
Include  telemetry for installed plugins

Periodically reports installed plugins to the plugin builder via POST /api/v1/telemetry/plugins, leveraging PluginUpdateFetcher periodic task.

Adds BTCPayServer/{version} User-Agent header to all plugin builder requests.

Skipped on Debug builds and Regtest/Development environments.

Links to https://github.com/btcpayserver/btcpayserver-plugin-builder/pull/220

Cc. @NicolasDorier


TChukwuleta:
Failed CheckNoDeadLink [20 s] ..
  
  Might need to rerun pipeline

NicolasDorier:
I really hate it, I don't like tracking. I understand plugin builders want this data, but this is against the interests of BTCPay Server users.

At the very least, it should be Opt-in and users need to be prompted.

PR #7373, opened 2026-05-20
Add default discount setting for Point of Sale apps

Wired4ncer:
## Summary

Adds an optional default discount percentage for Point of Sale apps. When discounts are enabled, the store can set a value that pre-fills the discount field — the customer or cashier can still change it.

Surfaced in the POS settings UI (collapsed under the "User can input discount" toggle) and in the Greenfield API.

## Details

- New `DefaultDiscount` (`int?`, 0–100) on `PointOfSaleSettings`, wired through the settings form, the public POS view, and the Greenfield request/response models. It is a whole percentage because that is what both public views can express: the cart input is `step="1"` and the keypad enters the discount at divisibility 0.
- The public POS only applies the default when `ShowDiscount` is enabled, so disabling discounts never leaves a hidden discount applied.
- `common.js` pre-fills `discountPercent` from the setting; the watcher is `immediate` so the order total reflects the default on initial load, and `clear()` restores the default rather than blanking it, so it survives between customers on a terminal that stays open.

## Test plan

- [x] Greenfield test in `CanCreateReadUpdateAndDeletePointOfSaleApp` asserts `DefaultDiscount` roundtrips on create and GET, and that an out-of-range value is rejected.
- [x] Playwright test `CanUsePOSDefaultDiscount` asserts the cart pre-fills the discount field, discounts the total on load, and keeps the default after the cart is emptied.


Wired4ncer:
The `fast_tests` failure on this branch is unrelated to the change. That job runs the `Fast` and `ThirdParty` traits together, and the only failing test is `ThirdPartyTests.CanGetRateCryptoCurrenciesByDefault`, which hits live exchange APIs:

```
Failed CanGetRateCryptoCurrenciesByDefault
  Impossible to get the rate ERR_RATE_UNAVAILABLE(bitfinex, DSH_BTC) * (64016.2, 64016.3)
  at BTCPayServer.Tests.ThirdPartyTests.CanGetRateCryptoCurrenciesByDefault():line 445
```

Bitfinex did not return a DSH_BTC rate during the run. A re-run of that job should clear it. `integration_tests`, `playwright_tests` and `playwright_2_tests` are green.

For reviewers, the last commit changes two things from the original proposal:

- `DefaultDiscount` is now an `int?` (0–100) rather than a decimal. Both public views can only express whole percentages — the cart input is `step="1"` and the keypad enters the discount at divisibility 0 — so a fractional default pre-filled the cart with a value that fails the input's step validation, which blocks the form submit and leaves the customer unable to pay.
- `clear()` now restores the default instead of blanking the discount, so it survives between customers on a terminal that stays open rather than only applying to the first sale.

Both are covered by tests: `CanUsePOSDefaultDiscount` (Playwright) and the `DefaultDiscount` assertions in `CanCreateReadUpdateAndDeletePointOfSaleApp` (Greenfield).

PR #7375, opened 2026-05-22
add allow-user-association plan feature

TChukwuleta:
Related to https://github.com/btcpayserver/btcpayserver/discussions/7101

Every user on a monetized BTCPay Server instance must be a paying subscriber. This means a store owner adding an employee to their store forces that employee to also subscribe even though they're not the one running the business.

This PR adds an allow-user-association plan feature. When enabled on a plan, users invited by a subscriber get attached to their subscription instead of needing their own. 
They inherit the subscriber's access. When the feature is removed, associated users are detached, enrolled as their own expired subscribers, and redirected to billing on next login.
The per-user model remains the default, this is opt-in per plan.

Let me know what you think, and if this is a good first step, I can go ahead and add tests

Cc. @NicolasDorier

### DIFF
--- a/BTCPayServer.Tests/MonetizationTests.cs
+++ b/BTCPayServer.Tests/MonetizationTests.cs
@@ -480,7 +480,10 @@ public class MonetizationTests(ITestOutputHelper helper) : UnitTestBase(helper)
         // Delete sponsor — associated employee should get detached
         await s.GoToServer(ServerNavPages.Users);
         var users = new PMO.UsersPMO(s);
-        await users.DeleteUser("sponsor@gmail.com");
+        await s.Server.WaitForEvent<MonetizationHostedService.MonetizationLockoutUpdated>(async () =>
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Ping @golyalpha is it what you need?

golyalpha:
Yup, this is pretty much what I asked for

TChukwuleta:
Hey @NicolasDorier  I have improved this PR and also included tests..

This is now open for review

NicolasDorier:
I need to think more about this, I've dumped some thoughts on https://github.com/btcpayserver/btcpayserver/discussions/7419

I think BTCPay Server have a deeper architectural limitation that prevent this solution to be easy to use and understand. We probably need to revisit the concept of ownership and maybe introduce the concept of Organisation.

But this isn't something this PR should focus on, this is a deeper architectural re-design.

PR #7377, opened 2026-05-22
Fix Yadio invalid rate handling

### LINKED ISSUE #7376: Zero or negative primary rates bypass fallback rate sources
# Zero or Negative Primary Rates Bypass Fallback Rate Sources

## Summary

BTCPay currently treats a primary rate source as successful when it returns a non-null `BidAsk`, even if the rate is unusable such as `0` or negative. Because fallback only runs when the primary result has `BidAsk == null`, an invalid primary result can block fallback resolution and disable payment methods.

## Example

Invoice logs:

```text
Error retrieving a matching payment method or rate.
05/22/2026 02:12:24:Info Creation of invoice starting
05/22/2026 02:12:24:Info Rate for BTC_USD: yadio(BTC_USD) = 0 = 0
05/22/2026 02:12:24:Error The calculated rate should not be 0.
05/22/2026 02:12:24:Error BTC-LN: Unable to get rate(s) BTC_USD, this payment method disabled for this invoice.
05/22/2026 02:12:24:Error BTC-LNURL: Unable to get rate(s) BTC_USD, this payment method disabled for this invoice.
```

Yadio response:

```json
{
  "BTC": {
    "USD": 0
  },
  "base": "BTC"
}
```

## Current Behavior

- The provider returns `BTC_USD = 0` as a `PairRate`.
- `RateFetcher` sees a non-null `BidAsk` and considers the primary result successful.
- The fallback rule set is not used.
- Later invoice code rejects the zero rate and disables payment methods.

## Expected Behavior

If a primary rate resolves to a non-positive value, it should be treated as unavailable or invalid by the shared rating flow so that:

- fallback rate sources can run, and
- non-fallback consumers do not treat unusable rates as successful results.

## Root Cause

`FallbackGetRuleValue(...)` currently only checks whether `primaryResult.BidAsk != null`.

The rating layer also allows evaluated expressions to produce `BidAsk(0)` or negative values without normalizing them into an invalid result.

## Proposed Fix

- Add a shared `BidAsk` usability check such as `Bid > 0 && Ask > 0`.
- Use that shared check in fallback selection instead of only `BidAsk != null`.
- Normalize non-positive evaluated `BidAsk` values to an unavailable result in the rating layer.

## Why This Should Be Shared

Some providers already filter zero rates locally, but many do not. Relying on every current and future provider to remember to reject `0` is error-prone. Shared validation in the rating flow keeps fallback and non-fallback behavior consistent.

BuffaloDyl:
Fixes #7376

## Summary

Skip non-positive rates returned by Yadio.

## Problem

Yadio can return invalid rates such as `0`. BTCPay was accepting those values from the provider.

## Fix

Only add Yadio rates when the parsed value is greater than zero.

### DIFF
--- a/BTCPayServer.Rating/Providers/YadioRateProvider.cs
+++ b/BTCPayServer.Rating/Providers/YadioRateProvider.cs
@@ -27,7 +27,7 @@ namespace BTCPayServer.Services.Rates
             {
                 string name = ((JProperty)item).Name;
                 var value = results[name].Value<decimal?>();
-                if (value.HasValue)
+                if (value is > 0m)
                     list.Add(new PairRate(new CurrencyPair("BTC", name), new BidAsk(value.Value)));
             }
 


NicolasDorier:
@BuffaloDyl this should be fixed at the provider level. If there is a provider returning invalid rate, it should return null.
The bug is in the specific provider, fix it there.

BuffaloDyl:
Updated per feedback. I removed the shared fallback/rating-layer changes and narrowed the fix to  so non-positive values are skipped at the provider level. I also replaced the broad fallback tests with a focused provider test covering , negative, , and positive values.\n\nThe branch has been force-pushed and the relevant FastTests pass locally.

BuffaloDyl:
Updated per feedback. I removed the shared fallback/rating-layer changes and narrowed the fix to `YadioRateProvider` so non-positive values are skipped at the provider level. I also replaced the broad fallback tests with a focused provider test covering `0`, negative, `null`, and positive values.

The branch has been force-pushed and the relevant FastTests pass locally.

NicolasDorier:
Remove the tests, this isn't testing anything and adding noise.

NicolasDorier:
Then fix the PR desc and title

BuffaloDyl:
Updated per feedback. I removed the test additions and narrowed the PR to the provider fix only. The remaining change is in `YadioRateProvider`, which now skips non-positive values from the Yadio response.

PR #7380, opened 2026-05-27
feat:add includePaymentMethods to POST invoice creation endpoint 

### LINKED ISSUE #7366: Feature Request: includePaymentMethods should be possible from POST /api/v1/stores/{storeId}/invoices
Currently when someone whats to show the BTC address (or other payment destinations) for a newly created invoice, they need to querry the invoice with includePaymentMethods set to true or call /api/v1/stores/{storeId}/invoices/{invoiceId}/payment-methods

In order to decrease network delays it would be nice to be able to do this in one API call (when creating the invoice) and getting returned the payment methods.

Thanks for considering it.

Idea came from a discussion with a  [TG user](https://t.me/btcpayserver/97888).

Godwin-88:
## Summary

Adds `includePaymentMethods` support to the POST invoice creation endpoint.

When `true`, payment methods are returned inline with the invoice creation response, avoiding an additional round-trip to:
`GET .../payment-methods`

Closes #7366

### DIFF
--- a/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
@@ -8,5 +8,6 @@ namespace BTCPayServer.Client.Models
         [JsonConverter(typeof(NumericStringJsonConverter))]
         public decimal? Amount { get; set; }
         public string[] AdditionalSearchTerms { get; set; }
+        public bool IncludePaymentMethods { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


NicolasDorier:
1. This should be a query string parameter of the route
2. This need to be documented in the api docs.

comwanga:
A few things I noticed going through the diff:

**`IncludePaymentMethods` belongs in the query string, not the request body.** The existing GET endpoint already uses `?includePaymentMethods=true` as a query param. Putting a response-shaping flag inside `CreateInvoiceRequest` mixes concerns — the body is for defining the invoice, not for controlling what the response looks like. Something like this would be more consistent:

```csharp
[HttpPost("~/api/v1/stores/{storeId}/invoices")]
public async Task<IActionResult> CreateInvoice(
    string storeId,
    CreateInvoiceRequest request,
    bool includePaymentMethods = false)
```

**Removing the default on `PaymentMethods` is a breaking change.** Clients that previously got `[]` from `CreateInvoice` will now get `null` unless they opt in. That's a silent regression for anyone iterating the array without a null check. I'd keep `= Array.Empty<InvoicePaymentMethodDataModel>()` and only populate it when the flag is true.

**Worth adding a doc comment** for the new param so it shows up properly in the generated API docs.

PR #7381, opened 2026-05-28
feat: embed plugin directory in plugin management

thgO-O:
This PR embeds the Plugin Builder public directory into the BTCPay Server plugin management page, replacing the old available plugins list while keeping installed, disabled, pending, and fallback flows managed by BTCPay Server.

It also surfaces Plugin Builder reviews and ratings directly in BTCPay Server, so users can see those trust signals while browsing and installing plugins.

Depends on [Plugin Builder PR](https://github.com/btcpayserver/btcpayserver-plugin-builder/pull/225)

https://github.com/user-attachments/assets/e0db8411-97b9-4e25-be69-4f300b9a58a0

## Changes

- Adds an embedded Plugin Builder directory iframe for browsing public plugins.
- Surfaces Plugin Builder reviews and ratings directly in BTCPay Server's plugin management UI.
- Lets users choose which plugin version to install from the embedded plugin details page.
- Adds a selected plugin panel that shows BTCPay Server state and actions for the selected directory plugin.
- Keeps installed, disabled, update, uninstall, cancel, and restart-required flows in BTCPay Server.
- Preserves plugin update notifications and anchors to the relevant plugin card.


r1ckstardev:
Request changes. Cross-PR work with the plugin-builder embed (PR #225 - already approved) shows clean protocol alignment, solid parent-side origin/source verification, and a well-scoped iframe sandbox. The review surfaced 3 install-flow gaps + 5 lower-severity items + a worth-flagging hidden bug-fix.

MEDIUM:
1. **`pb:install-requested.version` bypasses server-side compatibility/version validation.** The iframe-supplied version flows directly into `DownloadPluginAndDependencies` (`UIServerController.Plugins.cs:131`) without re-querying plugin-builder for the compatible-version set or verifying the requested version is in the advertised one for the current `btcpayVersion`/pre-release policy. A tampered embed (or a compromised plugin-builder origin) could request any existing version - including one not compatible with the current BTCPay. Fix: server-side re-query for compatible versions + verify the requested version is in that set + verify the manifest identifier matches the requested identifier.
2. **Explicit-version download skips manifest-identity check.** The null-version path in `DownloadRemotePlugin` filters candidates by `v.Identifier == pluginIdentifier`; the explicit-version path takes the API response as-is. A plugin-builder selector ambiguity or compromise could deliver a manifest with a mismatched identifier, producing file/queued-command divergence. Fix: after deserialization, fail closed unless `manifest.Identifier.Equals(pluginIdentifier, StringComparison.OrdinalIgnoreCase)`.

LOW/MEDIUM:
3. **The "fallback behavior for legacy/non-embedded available plugin management" claimed in the PR description is not present in `PluginDirectory.cshtml`.** When the iframe is unreachable/misconfigured, `showDirectoryError()` (`plugins-embed.js:53-66`) hides the embed shell and shows a static warning - there's no server-rendered available-plugin list that takes over. Either restore the non-embedded fallback or update the PR description to acknowledge that available-plugin browsing now requires embed availability.

LOW:
4. **Hidden plugin identifiers JSON serialization** in `PluginDirectory.cshtml:113` uses `@Html.Raw(JsonSerializer.Serialize(...))` inside `<script type="application/json">`. A malicious plugin manifest with `Identifier = "</script><script>alert(1)</script>"` would escape the script block. The threat model requires an already-installed malicious plugin (which has code execution anyway), so low marginal risk. Cheap fix: pass `new JsonSerializerOptions { Encoder = JavaScriptEncoder.Default }` so `<` and `>` get encoded.
5. **8-second timeout to declare directory failure** (`plugins-embed.js:47`) with no retry UI - cold-start or transient network blip leaves a permanent failure state until page reload. Worth a retry button or "loading slowly" intermediate state.
6. **No retry on panel fetch failure** (`plugins-embed.js:244-253`).
7. **`darkThemeLink.getAttribute("rel") === "stylesheet"` color-mode detection** (`plugins-embed.js:167-172`) is brittle - depends on the current way BTCPay toggles theme link state. Worth a code comment documenting the contract so future theme-handling changes preserve it.
8. **No `frame-src` / `default-src` CSP directive** is set in `Filters/ContentSecurityPolicyAttribute.cs`, so browsers fall back to allowing any iframe origin. The iframe URL comes from admin-configured `PoliciesSettings.PluginSource` - same trust boundary as PR #225's `data-origin`. By design but worth a one-line code comment noting the trust assumption.

Positive callout: `PluginUpdateFetcher` line 88 swap `pair.Name -> pair.Identifier` corrects a previous disabled-plugins lookup that was matching against the wrong field; the consistent `StringComparer.OrdinalIgnoreCase` plumbing through the dictionary handling is clean.

Tested locally against pr-7381 branch head 94bd632f. Cross-checked the postMessage protocol against PR #225's iframe-side (`public-embed.js`) - 5 message types align exactly; the parent-side origin/source verification (`event.origin !== origin` + `getFrameByWindow(event.source)`) closes the inbound boundary cleanly. 7 new unit tests on `FastTests.cs` cover `BuildDirectoryIframeUrl` + projection state combinations.

BTCPayServer/Views/UIServer/ListPlugins.cshtml
```diff
@@ -1,243 +1,165 @@
...
-    ViewData.SetLayoutModel(new LayoutModel(nameof(ServerNavPages.Plugins), StringLocalizer["Plugins"]));
-    var installed = Model.Installed;
-    var availableAndNotInstalled = new List<PluginService.AvailablePlugin>();
-    var availableAndNotInstalledx = Model.Available
-        .Where(plugin => !installed.ContainsKey(plugin.Identifier))
-        .GroupBy(plugin => plugin.Identifier)
-        .ToList();
-
-    foreach (var availableAndNotInstalledItem in availableAndNotInstalledx)
+    ViewData.SetLayoutModel(new LayoutModel(nameof(ServerNavPages.Plugins), StringLocalizer["Installed Plugins"])
```
rollforsats: This is really nitpick, but when I click on the Installed Plugins page I'm not fully at the top and have to scroll to see the "Installed Plugins" header.

https://github.com/user-attachments/assets/ea8f9c07-a6e3-4db2-b916-7a5ac41ca308

BTCPayServer/Views/UIServer/PluginDirectory.cshtml
```diff
@@ -0,0 +1,143 @@
+@using System.Collections.Generic
+@using System.Linq
+@using System.Text.Encodings.Web
+@using System.Text.Json
+@model ManagePluginsShellViewModel
+@{
+    Layout = "_Layout";
+    ViewData.SetLayoutModel(new LayoutModel(nameof(ServerNavPages.Plugins), StringLocalizer["Plugin Directory"])
```
rollforsats: Another nitpick:  I have to click on certain areas to dismiss the iframe. Unless it's the intended behavior, I'd expect it to dismiss if I clicked on an empty area on the Plugin Directory page.

https://github.com/user-attachments/assets/73ff39a3-5f0c-4d0a-b434-6e2f28898fee

rollforsats (commented):
@thgO-O This looks good to me! I've tested the embedded plugins page for the following:
* Installing a plugin in the iframe
* Canceling a pending install in the iframe
* Clicking on the different links (view source, write a review, contributor links) and verified I’m navigated to the appropriate pages
* Reviews are showing
* Installing/enabling/uninstalling plugins on the Installed Plugins page, including disabling plugins
* Searching on the Plugin Directory filters the plugins

Did not test:
* Multiple versions dropdown (Not sure how to add older versions on my local plugin-builder instance)
* Multiple reviews and review filter

^ If you want me to test these let me know

r1ckstardev:
Approve. All review items addressed cleanly across 3 commits:

- `4308bbb fix: validate plugin install requests` closes the two install-flow MEDIUMs. `SelectCompatiblePluginVersion` now re-queries plugin-builder for the compatible set and rejects requested versions outside it; `DownloadRemotePlugin` adds identifier + version match checks against the returned manifest before any file fetch. The 4 new unit tests in `FastTests.cs` (`PluginVersionSelection_RejectsRequestedVersionOutsideCompatibleSet` / `_DoesNotUseVersionsFromOtherIdentifiers` + `DownloadRemotePlugin_RejectsManifestIdentifierMismatch` with a stubbed HttpMessageHandler asserting `downloadRequested == false` after the identity-throw) directly cover the contract. Bonus `Uri.EscapeDataString(identifier)` on `GetPluginVersionsForDownload` is good defense-in-depth.

- `d930956 fix: add retry states for plugin embed failures` closes the LOW UX + hardening items: Retry button + `startDirectory()` reset flow for the 8s-timeout case; `renderPanelError(identifier, slug)` with retry button for panel-fetch failures; `JsonSerializerOptions { Encoder = JavaScriptEncoder.Default }` for the hidden-identifiers JSON; inline comments documenting the theme-link / `data-origin` trust assumptions.

- `bbef9b fix: polish plugin management navigation` swaps the custom offcanvas pointerdown handlers for standard Bootstrap `backdrop: true` behavior and drops a stale `asp-fragment` anchor. No drift on the security surface.

PR description also updated to drop the "Adds fallback behavior for legacy/non-embedded available plugin management" bullet - matches the actual code.

Tested locally against `bbef9b16`. Cross-PR alignment with PR #225 still holds.

BTCPayServer/Views/UIServer/PluginDirectory.cshtml
```diff
@@ -0,0 +1,143 @@
...
+    <div id="plugins-embed-shell"
+         class="pb-5"
+         data-origin="@Model.DirectoryOrigin"
+         data-iframe-url="@Model.DirectoryIframeUrl"
+         data-panel-url="@Model.PanelUrl"
+         data-selected-identifier="@Model.SelectedPluginIdentifier"
+         data-selected-slug="@Model.SelectedPluginSlug"
+         data-panel-error-message="@StringLocalizer["BTCPay Server could not load the plugin panel."]"
+         data-panel-retry-label="@StringLocalizer["Retry"]">
+        <script type="application/json" id="plugins-embed-hidden-plugin-identifiers">@Html.Raw(hiddenPluginIdentifiersJson)</script>
```
NicolasDorier: `@Html.Raw` give me heart attacks don't do this. If you want to serialize a json object, use Safe.Json

BTCPayServer/HostedServices/PluginUpdateFetcher.cs
```diff
@@ -37,9 +37,13 @@ protected override void FillViewModel(PluginUpdateNotification notification, Not
                 vm.Identifier = notification.Identifier;
                 vm.Type = notification.NotificationType;
                 vm.Body = StringLocalizer["New {0} plugin version {1} released!", notification.Name, notification.Version];
-                vm.ActionLink = linkGenerator.GetPathByAction(nameof(UIServerController.ListPlugins),
+                var actionLink = linkGenerator.GetPathByAction(
+                    nameof(UIServerController.ListPlugins),
                     "UIServer",
-                    new {plugin = notification.PluginIdentifier}, options.RootPath);
+                    pathBase: options.RootPath);
+                vm.ActionLink = actionLink is null
```
NicolasDorier: Why woudl actionLink be null?

BTCPayServer/HostedServices/PluginUpdateFetcher.cs
```diff
@@ -37,9 +37,13 @@ protected override void FillViewModel(PluginUpdateNotification notification, Not
...
+                    ? null
+                    : $"{actionLink}#{Uri.EscapeDataString(notification.PluginIdentifier)}";
```
NicolasDorier: I think GetPathByAction has a `FragmentString` parameter so you don't have to do that by hand.

NicolasDorier:
As I scroll, the list of plugins, it crashes

image

NicolasDorier:
Please can you refactor the `StubHttpMessageHandler` ? It's like the third time this class is copy/pasted by AI in tests.

NicolasDorier:
Ok so small adjustment/cleanup while we are at it:

Can you move the Plugin out of ServerController, and into its own UI controller in a new plugin "PluginManager"?
Instead of injecting in routes via `[FromService]`, you can inject directly in the ctor of the controller.

NicolasDorier:
The `ProjectionSource`  is something that gather all the mess of before in an intermediate viewmodel... that is used to generate the real viewmodel your UI depends on.

At first glance, you can probably make the code easier to understand by just building the ViewModel your UI depends on directly inside the controller. No intermediate type, no services. Potentially some ctor in the ViewModel if that make life easier.

For example, most of the methods in the Projection Services have like 10 parameters, then creating the Projection, then creating the view model. It's easier to just create the view model directly rather than going through this dance.
It will be more maintainable as well, because right now if one day we need to add something to the projection because of one specific view model, you'll have to update all the parameter list of all the methods.

BTCPayServer/Controllers/UIPluginManagerController.cs
```diff
@@ -0,0 +1,843 @@
...
+    IStringLocalizer stringLocalizer)
+    : Controller
+{
+    public IStringLocalizer StringLocalizer { get; } = stringLocalizer;
+
+    [HttpGet]
+    public async Task<IActionResult> ListPlugins()
+    {
+        var remotePlugins = await LoadRemotePlugins(setErrorStatusMessage: true);
+        var model = CreateInstalledPluginsViewModel(
```
NicolasDorier: Why do you need to pass 6 parameters to `CreateInstalledPluginsViewModel` ? `pluginService` is in the scope, so it can be accessed as a field inside `CreateInstalledPluginsViewModel`.

This would replace this 7 args method into a single argument for remotePlugins.

And this last argument can even be optional, as most of the time you only want to call `LoadRemotePlugins`.

BTCPayServer/Plugins/PluginManager/PluginUpdateFetcher.cs
```diff
@@ -21,26 +22,19 @@ internal class PluginUpdateNotification : BaseNotification
...
-                }
-            }
+            public override (string identifier, string name)[] Meta => [(Type, stringLocalizer["Plugin update"])];
 
             protected override void FillViewModel(PluginUpdateNotification notification, NotificationViewModel vm)
             {
                 vm.Identifier = notification.Identifier;
                 vm.Type = notification.NotificationType;
-                vm.Body = StringLocalizer["New {0} plugin version {1} released!", notification.Name, notification.Version];
+                vm.Body = stringLocalizer["New {0} plugin version {1} released!", notification.Name, notification.Version];
```
NicolasDorier: You should keep `StringLocalizer` (uppercase). This is because tha'ts why our `UpdateDefaultTranslations` relies on.

NicolasDorier (changes requested):

NicolasDorier:
Removed sandbox stuff, but need that.
https://github.com/btcpayserver/btcpayserver-plugin-builder/pull/237

Issue #7385, opened 2026-06-01
[Bug]: Custom Range month dropdown has white text on white background in Dark Mode


senutpal:
### What is your BTCPay version?

Master branch / Latest

### How did you deploy BTCPay Server?

I did not deploy it myself, I encountered this issue while testing on the Demo version on the BTCPay Server website.

### What happened?

When using the dark theme the month selection drop-down inside the Custom Range date picker has a white background with white text making the months unreadable. In light mode it behaves normally.

<img width="2879" height="1799" alt="Image" src="https://github.com/user-attachments/assets/574b09ec-57bd-4633-96fd-bf6ac0c72159" />

### How did you encounter this bug?

1. Switch BTCPay Server to use the dark theme.
2. Go to the Invoices section (`/invoices`).
3. Click the date filter dropdown (e.g., "All Time").
4. Select **Custom Range**.
5. In the modal that appears, click the input field for **Start Date** or **End Date**.
6. When the `flatpickr` calendar appears, click on the current Month name (e.g., "May") to open the month selector drop-down.
7. Observe that the drop-down list text is white against a white background.

### Relevant log output

N/A

### What browser do you use?

Brave (Latest Version)

### Additional information

I have verified the issue locally and it seems to exclusively affect the dark theme because the `<option>` elements inherit the dark mode white text but fall back to the browser's default white background. 

If the maintainers approve, I would love to open a PR and work on fixing this issue.

### Are you sure this is a bug report?

- [x] I confirm this is a bug report

PR #7388, opened 2026-06-04
fix: flatpickr dropdown focus inside modals

### LINKED ISSUE #7384: [Bug]: Invoice Date Filter - Date Selector Widget closes month list when clicked

BTCPay Server v2.3.9+4c7580104

Docker

### What happened?

Attempting to filter invoices by date, I click on "Custom" next to "All Status" and select "Custom Range", then I click on the Month (May), which automatically closes the list on me. See the attached video for the issue:

https://github.com/user-attachments/assets/d6d23a3e-ecdc-4caa-bd1c-6e67324ba382

see above

### Relevant log output

```shell
none
```

### What browser do you use?

Firefox 151.0.2 on Windows 11

### Additional information

none

senutpal:
## Description
This PR fixes an issue where the Flatpickr month dropdown immediately closes when clicked inside a Bootstrap modal. 

Bootstrap modals enforce focus traps (`enforceFocus`) to prevent keyboard navigation outside the modal. By default, Flatpickr renders its calendar at the end of the `document.body`. When a user clicks the month `<select>` dropdown on the calendar, Bootstrap detects focus leaving the modal and immediately steals it back, which aborts and closes the native dropdown.

This fix forces Flatpickr to render inline within the modal's DOM tree by setting `static: true` for instances initialized via the `data-fdtp` attribute, completely avoiding the focus collision.

Fixes #7384

## Before Vs After 

https://github.com/user-attachments/assets/ed1decf6-7d6c-44bb-8369-abbf60db9e9f

### DIFF
--- a/BTCPayServer/wwwroot/main/site.js
+++ b/BTCPayServer/wwwroot/main/site.js
@@ -353,7 +353,7 @@ document.addEventListener("DOMContentLoaded", () => {
 
         // support for initializing with special options per instance
         if (fdtp) {
-            var parsed = JSON.parse(fdtp);
+            var parsed = Object.assign({}, JSON.parse(fdtp), { static: true });
             flatpickrInstances.push(element.flatpickr(parsed));
         } else {
             var min = element.attr("min");


pavlenex:
Color on picker seems to be inverted, perhaps we can fix that as well?
image

senutpal:
yeah we could. i opened #7385 for that earlier. want me to fix it in this PR or do a separate one ?

PR #7391, opened 2026-06-05
fix: localization

ABC-8888:
Pages covered so far are sections of:
1. Translations
2. Users
3. Store - general settings


rockstardev:
Moving over to: https://github.com/btcpayserver/btcpayserver/pull/7398

PR #7392, opened 2026-06-06
fix(translation UI) : deletion translations languagesPack

### LINKED ISSUE #7390: [Bug]: Deinstall button missing and language not editable

2.3.9

Docker locally

### What happened?

After deinstalling and installing a language, the language is not any more clickable and button Deinstall is not shown.

1. Click on Translations in the server menu
2. Install a new language

**Expected behaviour:**
The name of the language is clickable and the button Deinstall is shown.

**Actual behaviour:**
The name of the language is not clickable and the button Deinstall is not shown.

Image

### Relevant log output

```shell

```

### What browser do you use?

Chrome 148

### Additional information

_No response_

teamssUTXO:
Closes #7390 

Users can now uninstall languagePacks from translator repo but still can't edit them.


teamssUTXO:
**Question for the TOCTOU in translation uninstall :** `lang_dictionaries. fallback' is currently 'ON DELETE SET NULL' Do we want to keep policy enforcement in app code only, or should we also enforce it at DB level with a migration (e. g. 'ON DELETE RESTRICT/NO ACTION ) to make this invariant hard-guaranteed outside application paths too?

BTCPayServer/Plugins/Translations/Views/ListTranslations.cshtml
```diff
@@ -109,7 +109,7 @@
 									<button id="Select-@v.TranslationName" type="submit" class="btn btn-link p-0" text-translate="true">Select</button>
 								</form>
 							}
-							@if (!v.IsSelected && v.Editable)
+							@if (!v.IsSelected && (v.Editable || v.IsDownloadedLanguagePack))
```
NicolasDorier: Why do you need 2 boolean? LanguagePack should have Editable to false.

NicolasDorier: Can you cleanup this by having a `v.IsDeletable` and use that instead. You can then probably remove `IsDownloadedLanguagePack`.

BTCPayServer/Plugins/Translations/LocalizerService.cs
```diff
@@ -160,11 +170,48 @@ public async Task<Translation> CreateTranslation(string langName, string? fallba
...
         }
 
-        public async Task DeleteTranslation(string translationName)
+        public async Task<DeleteTranslationResult> DeleteTranslation(string translationName)
         {
             await using var ctx = contextFactory.CreateContext();
             var db = ctx.Database.GetDbConnection();
-            await db.ExecuteAsync("DELETE FROM lang_dictionaries WHERE dict_id=@dict_id AND source='Custom'", new { dict_id = translationName });
+            await db.OpenAsync();
+            await using var transaction = await db.BeginTransactionAsync();
```
NicolasDorier: You don't need transactions for that.

BTCPayServer/Plugins/Translations/LocalizerService.cs
```diff
@@ -160,11 +170,48 @@ public async Task<Translation> CreateTranslation(string langName, string? fallba
...
+
+            var existing = await db.QueryFirstOrDefaultAsync<(string dict_id, string source)>(
+                "SELECT dict_id, COALESCE(source, '') source FROM lang_dictionaries WHERE dict_id=@dict_id FOR UPDATE",
```
NicolasDorier: You don't need FOR UPDATE

BTCPayServer/Plugins/Translations/LocalizerService.cs
```diff
@@ -160,11 +170,48 @@ public async Task<Translation> CreateTranslation(string langName, string? fallba
...
+            }
+
+            if (existing.source != "LanguagePack" && existing.source != "Custom")
+            {
+                await transaction.RollbackAsync();
+                return new DeleteTranslationResult(DeleteTranslationResultType.NotUserInstalled, []);
+            }
+
+            var fallbackUsers = (await db.QueryAsync<string>(
+                "SELECT dict_id FROM lang_dictionaries WHERE fallback=@fallback ORDER BY dict_id FOR UPDATE",
```
NicolasDorier: Just fetch all the dictionaries with the fallbacks from the database. Use `GetTranslations`.

NicolasDorier (changes requested):

BTCPayServer/Plugins/Translations/LocalizerService.cs
teamssUTXO: @NicolasDorier For the delete method in the LocalizerService file : The fallback foreign key is configured to be initialized to NULL upon deletion. So, the database won't prevent us from deleting a dictionary used by others as a fallback key, it will set their fallback key to NULL. The `UsedAsFallback` check below is the only protection. We should use "FOR UPDATE" on this line for the entire duration of the transaction so that a simultaneous insert/update of a dependency (which locks this line with a foreign key) cannot occur between the check and the deletion.

You can look at my comment about the TOCTOU above.

NicolasDorier: Wouldn't then a better idea to just let the DB throw an error if it is referencing a deleted dictionary rather than setting to the fallback column to null?
Because basically that's what we want: Not allowing delete of dictionaries which are used as fallback.

So we wouldn't initialize to NULL upon deletion.

teamssUTXO: Yes this would be better. So we have to make a migration ?

NicolasDorier: Yes, that would make more sense, and probably simplify the code a bit.

teamssUTXO:
@NicolasDorier can you re-run failed test? I don't know why `CanUseSSHService` test failed

teamssUTXO:
Ready for review

NicolasDorier:
Way better!

PR #7395, opened 2026-06-08
Fix #5868: Allow admins to delete other users' stores

### LINKED ISSUE #5868: [Feature]: As admin allow stores of other users to be deleted

kallal79:
### Description
Currently, when users are removed from the server, their stores remain orphaned because server administrators are only granted view-level permissions (`Policies.CanViewStoreSettings`) over stores by default. 
 
### Changes
* Upgraded `ServerAdminRolePermissions` in `BuiltInPermissionHandler.cs` from `CanViewStoreSettings` to `CanModifyStoreSettings`.

### DIFF
--- a/BTCPayServer.Tests/PlaywrightTests.cs
+++ b/BTCPayServer.Tests/PlaywrightTests.cs
@@ -2193,31 +2193,27 @@ namespace BTCPayServer.Tests
             string GetStorePath(string subPath) => $"/stores/{storeId}/{subPath}";
 
             // Admin access
-            await s.AssertPageAccess(false, GetStorePath(""));
+            await s.AssertPageAccess(true, GetStorePath(""));
             await s.AssertPageAccess(true, GetStorePath("reports"));
             await s.AssertPageAccess(true, GetStorePath("invoices"));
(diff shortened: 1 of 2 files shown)


NicolasDorier:
The fact admins can't modify stores is on purpose.

I have attempted to do this https://github.com/btcpayserver/btcpayserver/pull/7324 to give them more power, but this has been shut down. I agree that technically an administrator has unlimited power on the server. However, it would currently require somebody technically competent to leverage.

I close this PR because just changing the admin permission isn't acceptable.
However, I recognize the deeper issue: If the user is disabled, the stores public pages shouldn't be accessible. Or it should be possible to cleanup orphaned stores. If you have a different approach, open another PR.

PR #7396, opened 2026-06-09
Add copy and QR actions for store ID

BuffaloDyl:
## Summary

Adds inline QR and copy actions to the `Store Id` field on the Store Settings page.

## Motivation

Some setup flows require moving a store ID from the BTCPay UI to another device or app. This change makes that easier without changing store behavior or API surface.

## Changes

- wraps the `Store Id` input in an input group
- adds a QR button to display the raw store ID in the existing QR modal
- adds a copy button to copy the store ID to the clipboard
- removes the need for a separate text link by keeping both actions inline with the field

## Notes

- UI-only change
- QR payload is the raw store ID only

## Testing

- `dotnet build BTCPayServer/BTCPayServer.csproj`
- manually verified the Store Settings page in a local BTCPay instance
- verified the QR modal opens from the new inline button
- verified the copy action still renders correctly

### DIFF
--- a/BTCPayServer/Views/UIStores/GeneralSettings.cshtml
+++ b/BTCPayServer/Views/UIStores/GeneralSettings.cshtml
@@ -1,10 +1,12 @@
 @using BTCPayServer.Abstractions.Contracts
 @using BTCPayServer.Client
+@inject BTCPayServer.Security.ContentSecurityPolicies Csp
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@coderabbitai can you double check if there isn't some obivous risk at calling  `Csp.UnsafeEval();` on this store setting page?

PR #7402, opened 2026-06-12
fix: lightning invoice silently dropped when node doesn't return amount on reconnect

atharrva01:
so i was digging into how the lightning listener handles reconnects and found something that's been bugging me

basically when the server reconnects to a lightning node, it runs a catch-up poll (PollAllListenedInvoices) to check if any payments came in while we were offline. the problem is, if the node says the invoice is Paid but doesn't return an amount (which can happen with top-up/amount-less invoices or just transient node state right after a restart), AddPayment returns false. but the old code was still removing the invoice from _ListenedInvoices like everything was fine. so the payment just... disappears. no error, no UI feedback, merchant never gets notified.

the fix is pretty small, i changed PollPayment to return a tuple (status, recorded) and now PollAllListenedInvoices only evicts the invoice if the payment was actually written to the db. if it wasn't, the invoice stays in the list and gets retried on the next poll cycle.

there's a similar issue in the notification path too (the WaitInvoice loop) where a paid notification with null amount gets silently skipped, didn't touch that here but worth a follow up

tested the build locally, no new errors

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningListener.cs
+++ b/BTCPayServer/Payments/Lightning/LightningListener.cs
@@ -484,21 +484,23 @@ namespace BTCPayServer.Payments.Lightning
             ConnectionString = connectionString;
         }
 
+        internal enum RecordedState { RecordedNow, AlreadyRecorded, RetryLater }
+
         internal bool AddListenedInvoice(ListenedInvoice invoice)
         {
             return _ListenedInvoices.TryAdd(invoice.PaymentMethodDetails.InvoiceId, invoice);
         }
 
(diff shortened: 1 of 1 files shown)


atharrva01:
hi @NicolasDorier this pr fixes a silent payment loss in lightning reconnect poll when node returns no amount.

NicolasDorier:
Thanks for digging. I will include that in 2.4.0

NicolasDorier:
@coderabbitai can you review?

NicolasDorier:
@coderabbitai

> BOLT11/hash cross-check removed

This is on purpose, it is unclear why it was useful.

> AlreadyRecorded for null invoice

This is fine. We just want to stop tracking it.

> Unconditional AddPayment on WaitInvoice	

This is correct, `AddPayment` is looking at the state, so no need for the callers to do it.

Can you review once more?

PR #7405, opened 2026-06-15
don't silently drop paid LN invoices when amount is missing on expiry

atharrva01:
Found this while going through the lightning poll logic after #7402. Basically if a node reports an invoice as Paid but doesn't return an amount yet (happens sometimes after reconnects), and the BTCPay invoice expires before the amount shows up, RemoveExpiredInvoices was just dropping it silently - no retry, no log, nothing. Payment essentially vanishes from our tracking even though the customer paid.

This PR gives it one last poll before giving up. If the node still can't give us an amount, we now log a warning so at least it's visible instead of just disappearing.

plit the old RetryLater state into RetryLaterUnpaid and RetryLaterMissingAmount so we can tell these two cases apart (unpaid expiry is fine to drop quietly, missing-amount is the risky one).

Added a regression test (LightningListenerTests.cs) covering both states and the final-poll-before-evict behavior.

### DIFF
--- a/BTCPayServer.Tests/LightningListenerTests.cs
+++ b/BTCPayServer.Tests/LightningListenerTests.cs
@@ -1,0 +1,142 @@
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using BTCPayServer.Lightning;
+using BTCPayServer.Payments;
+using BTCPayServer.Payments.Lightning;
+using BTCPayServer.Services;
+using BTCPayServer.Services.Invoices;
+using Dapper;
+using Microsoft.EntityFrameworkCore;
+using NBitcoin;
+using Xunit;
+using static BTCPayServer.Payments.Lightning.LightningInstanceListener;
+
+namespace BTCPayServer.Tests
+{
+    [Trait("Integration", "Integration")]
+    public class LightningListenerTests(ITestOutputHelper helper) : UnitTestBase(helper)
+    {
+        [Fact]
+        public async Task DoesNotSilentlyDropInvoiceWaitingForAmount()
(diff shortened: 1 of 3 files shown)


atharrva01:
hi @NicolasDorier this pr fxes a silent payment-loss bug where expired LN invoices with a missing amount got dropped without retry now polls once more and warns if still unrecorded, would love your review whenever you get a chance , thanks

atharrva01:
also @NicolasDorier , if i had doubts or required your guidance on maybe a issue or approach on a pr, where am i suppose to contact you?

NicolasDorier:
@atharrva01 chat.btcpayserver.org, you can DM

NicolasDorier:
Remove the test and the new `virtual`. I believe that will break some plugins, and I don't want so close to release.
The test is just too complex to maintain for such a corner case.

Also, it seems to trying to paper over a bug coming from a lightning client implementation itself.
Which lightning implementation exactly is causing this bug?

Having a AmountReceived to null, when the invoice has been paid is a bug from the underlying, specific provider that we should fix.

atharrva01:
hey @NicolasDorier,  i'll drop the test and the virtual, no problem.

on the implementation question, i honestly don't have a confirmed specific node that triggers this. i traced it from the existing AmountReceived ?? Amount null-check that was already in the code (from #7323), and assumed it was a real scenario worth guarding. but if that null case is already fixed at the provider level or it's not actually happening in practice, then yeah this PR is just papering over something that doesn't need patching here.

do you know which provider #7323 was originally fixing? if it's the same one, i can check if it still returns null on paid invoices and whether the fix belongs in the client library instead.

PR #7407, opened 2026-06-17
fix: bulk archive pull payments scoped to current store

atharrva01:
I found a missing store check in the bulk archive endpoint added in #7400

when you POST to mass-action with a list of pull payment IDs, there was no check that those IDs actually belong to your store. so a user with archive permission on store A could archive pull payments from store B just by including foreign IDs in the request body.

the single archive endpoint is fine because pullPaymentId is a route param and gets validated by the auth middleware. bulk action takes IDs from the body so that check never runs.

fix is pretty small, pass the current storeId into the cancel request and filter in HandleCancel before touching anything in the db.

### DIFF
--- a/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
+++ b/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
@@ -282,7 +282,7 @@ namespace BTCPayServer.Controllers
                 switch (command)
                 {
                     case "archive":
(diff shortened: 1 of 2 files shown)


atharrva01:
hi @NicolasDorier this pr fixes the IDOR introduced in #7400 where MassAction had no store ownership check on the provided pull payment IDs

NicolasDorier:
Good catch

PR #7409, opened 2026-06-18
fix: lightning invoices stuck when node doesn't return amount during live session

atharrva01:
so basically found this while looking at the reconnect fix from #7402

the issue is when a LN node sends a paid notification but with a null amount (phoenixd does this, some CLN setups too), `AddPaymentCore` returns `RetryLater` and the invoice stays in `_ListenedInvoices` waiting for the next poll. thats expected.

but the problem is `PollAllListenedInvoices` only runs when a new session starts (basically on reconnect). if the websocket session stays alive the whole time, those `RetryLater` invoices just sit there. they never get re-polled.

then eventually the invoice expires, `RemoveExpiredInvoices` cleans it up, and the payment never gets recorded in BTCPay even tho the sats are already sitting in the merchants wallet.

the fix is pretty simple. `CheckConnection` already runs every minute from the timer, so added a check there - if the session is alive and the last full poll was more than 5 minutes ago, trigger a poll.

also moved the `LastFullPoll` update to before dispatching the poll instead of after it finishes. otherwise if the node is slow theres a chance multiple polls can get kicked off before the timestamp gets updated.

before:

* `RetryLater` invoices only got rescued if a reconnect happened

after:

* they get rescued within 5 min even if the websocket session never drops

### DIFF
--- a/BTCPayServer/Payments/Lightning/LightningListener.cs
+++ b/BTCPayServer/Payments/Lightning/LightningListener.cs
@@ -284,17 +284,8 @@ namespace BTCPayServer.Payments.Lightning
                 if (!instance.Empty)
                 {
                     instance.EnsureListening(_Cts.Token);
-                    // Re-poll periodically during live sessions to catch invoices stuck in RetryLater
-                    // (e.g. nodes that return null amount in the initial WS notification).
-                    // PollAllListenedInvoices is normally only called on session start (reconnect path),
-                    // leaving RetryLater invoices unresolved until the next disconnect.
-                    if (instance.IsListening &&
-                        instance.LastFullPoll is { } last &&
-                        DateTimeOffset.UtcNow - last > TimeSpan.FromMinutes(5.0))
-                    {
-                        instance.LastFullPoll = DateTimeOffset.UtcNow;
-                        _ = instance.PollAllListenedInvoices(_Cts.Token);
-                    }
+                    if (instance.IsListening && !instance.PaidButNoAmountEmpty)
+                        _ = instance.PollPaidButNoAmountInvoices(_Cts.Token);
                 }
             }
         }
@@ -511,6 +502,7 @@ namespace BTCPayServer.Payments.Lightning
             if (lightningInvoice is null)
             {
                 _ListenedInvoices.TryRemove(listenedInvoice.PaymentMethodDetails.InvoiceId, out _);
+                _PaidButNoAmount.TryRemove(listenedInvoice.PaymentMethodDetails.InvoiceId, out _);
                 return;
             }
             if (await AddPayment(lightningInvoice, listenedInvoice.InvoiceId, listenedInvoice.PaymentMethod.PaymentMethodId))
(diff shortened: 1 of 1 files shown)


atharrva01:
@nicolasdorier found this while digging into #7402, silent payment loss when node omits amount in first notification and session never drops

NicolasDorier:
@atharrva01 the problem is that poll will create way bigger problem when there are a lot of concurrent invoices. And instead of a bug affecting only CLN and Phoenixd, this will affect everybody.

It is a bug from CLN or Phoenixd providers to not have the "PaidAmount" in notifications. Does it really happen?

What node and version of node did you use to have this issue? Can you repro manually consistently?

atharrva01:
hi @NicolasDorier

On whether it happens, it does. #7322 is exactly that case: CLN with an amountless/top-up invoice where the payment succeeds on the node side, but BTCPay never records it. That's also why #7323 added the warning about the node reporting a payment without an amount.

On the performance side, you're right. I overcorrected a bit. `PollAllListenedInvoices` ends up touching every invoice in `_ListenedInvoices`, including ones that are simply unpaid and waiting normally.

The real issue is that `RetryLater` is used for two different cases:

* invoice still unpaid → nothing to do
* node says paid but returns no amount → this is the stuck case

I think a better approach is to track only the second case in a separate `_PaidButNoAmount` collection and have `CheckConnection` poll just those invoices. That keeps the scope small and avoids sweeping everything.

NicolasDorier:
@atharrva01 rather that papering over a bug from CLN, we should probably fix it at the source.
An amountless invoice that is paid should have the AmountReceived that isn't null.

I've been unable to reproduce this. Which version of CLN? Tested with 25.05

atharrva01:
reworked this based on your feedback

on the CLN version, don't have a specific one, the report in #7322 didn't mention it and i can't pin it down. if 25.05 is clean that's fine, but the warning log and RetryLater path from #7323/#7402 are already in master as a defensive layer for when it does happen.

on the performance concern, you were right, the full sweep was too broad. reworked it so instead of polling everything in _ListenedInvoices, there's now a separate _PaidButNoAmount dict that only gets populated when a notification comes in with Status=Paid but null amount (the specific line 633 path). normal unpaid invoices waiting for payment never touch it.

CheckConnection now only calls PollPaidButNoAmountInvoices when that dict is non-empty so for nodes that always return amounts (LND, CLN 25.05 etc) the overhead is literally zero, just an IsEmpty check per minute.

when PollPayment runs on a stuck invoice and GetInvoice comes back with the amount this time, AddPayment resolves it, clears it from both _ListenedInvoices and _PaidButNoAmount, and it's done.

NicolasDorier:
I can't accept this PR until you can write a test reproducing the bug on a specific lightning implementation.
This add complexity on a code that is already sufficiently complex, it has more potential to break stuff than fix things.

If WaitInvoice returns an invoice in the Paid state, the ReceivedAmount shouldn't be null. If it is, then there is a bug at the provider level, or the node level, and it must be fixed there. This PR is trying to hide a bug from the provider, which is worse, as it will make things even harder to debug when somebody will wonder why the payment amount doesn't match the received amount.

I bumped core lightning and improved the tests on BTCPayServer.Lightning in an attempt to reproduce the bug, and wasn't able to.

atharrva01:
@NicolasDorier went back and dug into this properly.

the original bug in #7322 was in BTCPay, not CLN. before #7323 the WS path checked notification.Amount != null, but for amountless invoices that's always null since it's the invoice amount, not what was received. so every paid amountless invoice silently failed that check and got dropped. #7323 already fixed it by switching to AmountReceived ?? Amount, and CLN does return amount_received_msat on paid invoices so that works correctly now.

the null-amount RetryLater path this PR defends against doesn't actually get hit on any current CLN version. should've traced this further before opening. closing.

PR #7413, opened 2026-06-21
 Greenfield: support payoutMethods array on invoice refund 

### LINKED ISSUE #6185: **API: Invoice refund inconsistency**
**API: Invoice refund inconsistency**
With the [pull payments endpoint](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Invoices_Refund) you can set paymentMethods as an array and if not set it will apply all available payment methods, with the [invoice refund endpoint](https://docs.btcpayserver.org/API/Greenfield/v1/#operation/Invoices_Refund) you only can pass a string and it is required too.

Suggestion: Make it a nullable array too to be consistent and more flexible

_Originally posted by @ndeet in https://github.com/btcpayserver/btcpayserver/issues/5899#issuecomment-2071870864_

atharrva01:
Closes #6185.

Added optional `payoutMethods` array to refund endpoint, same as pull payment already has. `payoutMethodId` still works, just marked obsolete now.

Pull payment took `payoutMethods` as optional array already, refund only took single required `payoutMethodId` - that's the mismatch issue was about. Two options in the thread: Nicolas wanted to just drop `payoutMethodId`, but that's breaking and ndeet called it out. Went non-breaking since we're on 2.4.0.

Resolve order: `payoutMethods` first, falls back to `payoutMethodId`, falls back to invoice default if neither passed. Old clients see no difference.

- `RefundInvoiceRequest.cs` - new `PayoutMethods` field, `payoutMethodId` marked obsolete
- `GreenfieldInvoiceController.cs` - resolving logic, pull payment now gets all methods not just one
- swagger json - new field documented, old one deprecated
- refund test updated

Extended `CanRefundInvoice` instead of writing a new test. Still covers all old `payoutMethodId` cases plus added array path.

image

Note: test store only has BTC onchain, so didn't test refund with multiple different payout methods (needs lightning). Array just passes straight into `CreatePullPaymentRequest.PayoutMethods` which already handles multiple, and that's covered by pull payment tests.

### DIFF
--- a/BTCPayServer.Client/Models/RefundInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/RefundInvoiceRequest.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using BTCPayServer.JsonConverters;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
@@ -17,7 +18,15 @@ namespace BTCPayServer.Client.Models
     public class RefundInvoiceRequest
     {
         public string? Name { get; set; } = null;
+
+        [Obsolete("Use PayoutMethods instead")]
         public string? PayoutMethodId { get; set; }
+
+        /// <summary>
+        /// The payout methods the refund can be claimed in. If null, falls back to <see cref="PayoutMethodId"/>,
+        /// and if that is also null, to the default payout method of the original invoice.
+        /// </summary>
+        public string[]? PayoutMethods { get; set; }
         public string? Description { get; set; } = null;
         
         [JsonConverter(typeof(StringEnumConverter))]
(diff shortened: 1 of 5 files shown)


atharrva01:
@NicolasDorier this closes #6185, non-breaking `payoutMethods` array on refunds, `payoutMethodId` kept as deprecated fallback.

BTCPayServer.Client/Models/RefundInvoiceRequest.cs
```diff
@@ -17,7 +18,15 @@ public enum RefundVariant
     public class RefundInvoiceRequest
     {
         public string? Name { get; set; } = null;
+
+        [Obsolete("Use PayoutMethods instead")]
```
thgO-O, line 22: IMHO it is better to keep this deprecated only in Swagger for now. Adding [Obsolete] forces CS0618 suppressions while the legacy field still needs to be supported. If we decide to keep, then UnitTest1.cs:2939 should also be updated to use PayoutMethods to avoid warning churn.

NicolasDorier: I think it's fine to keep Obsolete here, also you can remove `payoutMethodId` documentation from the swagger. Let's not document what is deprecated.

But yeah, the tests shouldn't have a warning.

PR #7423, opened 2026-07-02
Localize Blazor strings that the text-translate TagHelper can't reach

Wired4ncer:
## Problem

`text-translate` / `html-translate` are implemented by `TranslateTagHelper`, an MVC `TagHelper`. TagHelpers only run while rendering Razor (`.cshtml`) views — **Blazor components do not execute TagHelpers**. As a result these attributes are inert inside `.razor` files and the literal English text is rendered regardless of the active language.

This is visible in the notifications dropdown (`NotificationsDropDown.razor`): the heading **"Notifications"**, **"Mark all as seen"**, **"No new notifications"** and **"View all"** always render in English, even when a translation for e.g. "Notifications" exists in the dictionary. The button tooltip on the same component (`@StringLocalizer["Notifications"]`) *is* translated — which confirms the C# localizer path works in Blazor while `text-translate` does not.

## Fix

Replace the inert `text-translate="true"` usages in Blazor components with `@StringLocalizer[...]` (or `@ui.StringLocalizer[...]` where a UI helper is already used):

- **`NotificationsDropDown.razor`**: `Notifications`, `Mark all as seen`, `No new notifications`, `View all`
- **`PosLoginCode.razor`**: `Signed in user`, `None, just open the URL` (adds the `IStringLocalizer` injection)
- **`VaultBridge/Elements/XPubSelect.razor`**: `Account`

These were the only `.razor` files still using `text-translate` / `html-translate`.

The newly-translatable strings are registered in `Translations.Default.cs`.


Wired4ncer:
Thanks — on the `"View all"` vs `"View All"` nitpick: this is intentional. `NotificationsDropDown.razor` already rendered the literal text `View all` (lowercase) before this change; the PR only routes it through `StringLocalizer`. Since lookups are exact, case-sensitive matches on the rendered English string, the key has to be `"View all"` to resolve — reusing `"View All"` would silently fail to translate. The two casings live in distinct contexts already in the codebase (`"View All"` in the dashboard Recent Transactions/Invoices/Store Numbers components vs. the notifications dropdown here), so normalising the casing would be a separate, user-visible text change outside the scope of this i18n fix. Leaving the string as-is to keep the change faithful to the existing UI.

PR #7424, opened 2026-07-03
Improve date range filtering and shared search components

NicolasDorier:
# Features

## Allow time zone configuration in date range selectors

A new text input lets users select the time zone used by date range selectors.

image

As before, the default selection is based on the user's browser settings.

image

## Add more relevant default date range filters

Previously, the available filters were limited to `Last 24H`, `Last 3 days`, and `Last 7 days`.

These options did not fit common business reporting periods, which are usually based on days, months, quarters, or years.

This PR adds new default filters that better match those use cases.

image

## Unify the label and date range filter design

This PR unifies the design of the date range selector and label selector across the following pages:

  * Notifications
  * Payment requests
  * Wallet transactions
  * Invoices

## Use the browser's preferred hour format in the date time picker

The date time picker now uses the browser's preference to decide whether to display time in AM/PM or 24-hour format.

image

# Bug fixes

## Fix date range selector results at DST boundaries

When a custom date filter was applied, the backend assumed that both dates used the browser’s current timezone offset. This was not always correct.

For example, the selected date could fall within daylight saving time, while the current browser session does not. This could cause the selected date range to be off by one hour.

This PR now passes the browser timezone instead of the timezone offset, allowing the backend to calculate the correct time for the selected dates.

## Fix the `Mark as seen` button in the notifications list

The `Mark as seen` button in the notifications list was not working.

image

## Incorrect date range filtering across DST boundaries

Fixes date-based searches by using the browser's timezone instead of a fixed timezone offset when parsing search filters.

Using only the browser offset can produce incorrect results around DST transitions. For example, a custom date filter may refer to a date where the DST offset is different from the current browser session offset, causing the backend to select dates off by one hour.

# Architecture

## New razor view components

This PR removes duplicated search and filter code from the following pages:

The following Razor view components have been created and should be used by plugins and future development:

* `<vc:clear-all-filters>`
* `<vc:date-range-selector>`
* `<vc:label-selector>`
* `<vc:search-string-input>`

image 

## Improved SearchString

Added `timezone:` and `date-range:` to `SearchString` in order to make sure the class properly calculate the date range of the user.


NicolasDorier:
@coderabbitai  review

PR #7426, opened 2026-07-04
Greenfield: support includePaymentMethods on get invoice

atharrva01:
The list endpoint already lets you pass includePaymentMethods=true to get payment methods inline, but the single invoice GET didn't. So if you had one invoice and wanted its addresses, you had to make a second call to /payment-methods.

This just wires the same flag into GET invoice so you can get everything in one call. Off by default, so nothing changes for existing callers.

Also updated the client SDK and swagger docs to match.


atharrva01:
hi @NicolasDorier , would love your review on this

PR #7429, opened 2026-07-06
Return NotFound in payouts view when the pull payment no longer exists

atharrva01:
The payouts view loaded the pull payment name like this:

```csharp
vm.PullPaymentName = (await ctx.PullPayments.FindAsync(pullPaymentId)).GetBlob().Name;
```

`FindAsync` returns `null` when the pull payment id no longer exists — for example a deleted/archived pull payment, or a stale bookmarked `/stores/{storeId}/pull-payments/{pullPaymentId}/payouts` URL. Calling `.GetBlob()` on that `null` throws a `NullReferenceException`, so the page returns a 500.

This looks up the pull payment up front and returns `NotFound()` if it's missing, matching the pattern used elsewhere in the controllers.

### DIFF
--- a/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
+++ b/BTCPayServer/Controllers/UIStorePullPaymentsController.PullPayments.cs
@@ -547,8 +547,11 @@ namespace BTCPayServer.Controllers
                 ctx.Payouts.Where(p => p.StoreDataId == storeId && (p.PullPaymentDataId == null || !p.PullPaymentData.Archived));
             if (pullPaymentId != null)
             {
+                var pp = await ctx.PullPayments.FindAsync(pullPaymentId);
+                if (pp is null)
+                    return NotFound();
                 payoutRequest = payoutRequest.Where(p => p.PullPaymentDataId == vm.PullPaymentId);
-                vm.PullPaymentName = (await ctx.PullPayments.FindAsync(pullPaymentId)).GetBlob().Name;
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Attempt to reproduce the crash or leak on https://mainnet.demo.btcpayserver.org/  (but even name leaking is not a problem, as if you have the ppId, you can go to the pull payment page without permission anyway)

Your tests bypass permissions checks because it is calling the controllers directly without passing through all the permission handling checks.

If you can't reproduce in prod, please do not open any PR trying to fix stuff that isn't broken. This waste my time in reviewing and cross checking, and add risks of breaking else something that actually matter.

atharrva01:
Yeah, you're right on both counts. The name isn't actually sensitive since anyone with the id can open the pull payment page anyway, so I dropped that whole angle (and the store-scoping). And the old test was bogus, calling the controller directly skipped the auth checks, so it didn't prove anything.

I did manage to repro the crash on its own though, no permissions bypass involved: if you open the payouts page for a pull payment that doesn't exist anymore in your own store, a deleted/archived one, or just a stale bookmarked URL, FindAsync returns null and .GetBlob() throws a 500:


System.NullReferenceException : Object reference not set to an instance of an object.
  at BTCPayServer.Data.PullPaymentsExtensions.GetBlob(PullPaymentData data)
  at UIStorePullPaymentsController.Payouts(...) UIStorePullPaymentsController.PullPayments.cs:line 551


So I've trimmed the PR down to just the null check and updated the test to cover that case.

PR #7430, opened 2026-07-07
Support LUD-290 (pinLimit) for LNURL-Withdraw in checkout

Kukks:
## What

Adds [LUD-290](https://github.com/lnurl/luds/pull/290) support to the NFC checkout. When a customer pays an invoice by tapping an LNURL-Withdraw that requires a PIN for the amount, BTCPay now asks for the 4-digit PIN and forwards it to complete the payment.

This is the wallet side of the spec, for Bolt Cards: issuers like [boltcard-tools-terminal](https://github.com/boltcard/boltcard-tools-terminal) and [boltcard-lndhub](https://github.com/boltcard/boltcard-lndhub) can require a PIN above a set amount, so a lost or stolen card can't be drained.

## Screenshot

_PIN entry: the 4-digit field shows the amount, with Confirm / Cancel._

## Also fixes

NFC withdraw has been returning 400 since #7199 applied CSRF protection to all UI controllers — it caught `NFCController` too, even though the checkout posts without a token. Restored with `[IgnoreAntiforgeryToken]`, like the other LNURL endpoints.

## Testing

- Fast unit test for the PIN transport guard.
- Lightning integration test driving the full flow against a fake withdraw service: prompt → wrong PIN → retry → success, plus below-threshold (no PIN) and three-strikes-blocked.

### DIFF
--- a/BTCPayServer/Plugins/NFC/NFCController.cs
+++ b/BTCPayServer/Plugins/NFC/NFCController.cs
@@ -46,12 +46,14 @@ namespace BTCPayServer.Plugins.NFC
             public string InvoiceId { get; set; }
             public long? Amount { get; set; }
 
-            // LUD-290 PIN-protected withdraw: set on the second (completion) call.
+            // LUD-290: set on the PIN completion call.
             public string Pin { get; set; }
             public string Token { get; set; }
         }
 
+        // Public checkout endpoint POSTed via fetch without a token, like the LNURL endpoints.
         [AllowAnonymous]
+        [IgnoreAntiforgeryToken]
         public async Task<IActionResult> SubmitLNURLWithdrawForInvoice([FromBody] SubmitRequest request)
         {
             var invoice = await _invoiceRepository.GetInvoice(request.InvoiceId);
@@ -60,8 +62,7 @@ namespace BTCPayServer.Plugins.NFC
                 return NotFound();
             }
 
-            // LUD-290: the customer has entered a PIN for a withdraw we already resolved and stashed.
-            // Complete it using the stored session (single fetch of the withdrawRequest, reused k1).
+            // LUD-290: complete a PIN-protected withdraw from the stashed session.
             if (!string.IsNullOrEmpty(request.Token))
             {
                 return await CompletePinProtectedWithdraw(request, invoice);
@@ -196,9 +197,8 @@ namespace BTCPayServer.Plugins.NFC
                 return BadRequest("Could not fetch BOLT11 invoice to pay to.");
             }
 
-            // LUD-290: if the withdraw service asks for a PIN at this amount, don't contact the callback
-            // yet. Stash the already-fetched request (so we reuse the same single-use k1) and ask the
-            // browser for the PIN; the callback is only hit once the customer submits it.
+            // LUD-290: PIN required - stash the fetched request (reusing its single-use k1) and ask
+            // the browser for a PIN instead of calling the callback now.
             if (info.PinLimit is not null && withdrawAmount is not null &&
                 withdrawAmount.MilliSatoshi >= info.PinLimit.MilliSatoshi)
             {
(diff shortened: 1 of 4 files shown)


NicolasDorier:
Please clean up this horrible PR description it gives me a headache. Don't repeat what the code does to such details, I can read it and people who read the release notes and come here aren't probably devs so they don't care about an english version of C#.

At least add a screenshot of the pin, and tell us if there is some wallet somewhere that implement that or how we can use it in the real life?

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -194,11 +231,85 @@ public async Task<IActionResult> SubmitLNURLWithdrawForInvoice([FromBody] Submit
...
+            catch (Exception ex)
+            {
+                return BadRequest(ex.Message);
+            }
+        }
+
+        // The PIN is a plaintext query param, so only forward it over https, onion, or loopback.
+        internal static bool IsPinTransportSecure(Uri callback)
+        {
+            return callback.Scheme == Uri.UriSchemeHttps || callback.IsOnion() || callback.IsLoopback;
```
NicolasDorier: instead of `.IsLoopback`, use `Extensions.IsLocalNetwork`

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -194,11 +231,85 @@ public async Task<IActionResult> SubmitLNURLWithdrawForInvoice([FromBody] Submit
...
+                {
+                    _memoryCache.Remove(PinSessionCacheKey(request.Token));
+                }
+
+                return BadRequest(reason);
+            }
+            catch (Exception ex)
+            {
+                return BadRequest(ex.Message);
+            }
```
NicolasDorier, line 267: Put the try/catch around what is actually expected to crash for some unknown reason, that is  `session.Info.SendRequest`

BTCPayServer/Plugins/NFC/Views/CheckoutEnd.cshtml
```diff
@@ -51,13 +64,25 @@ Vue.component("lnurl-withdraw-checkout", {
...
+            return `Enter the 4-digit PIN to withdraw ${this.pinAmountSats} sats`
         }
     },
     data () {
         return {
 				url: @Safe.Json(Url.ActionAbsolute(Context.Request, "SubmitLNURLWithdrawForInvoice", "NFC").ToString()),
             amount: 0,
-            submitting: false
+            submitting: false,
+            pinRequired: false,
```
NicolasDorier: Can you make an object `pin`, with all those properties inside?

BTCPayServer/Plugins/NFC/NFCController.cs
```diff
@@ -24,26 +26,35 @@ public class NFCController : Controller
...
             public long? Amount { get; set; }
+
+            // LUD-290: set on the PIN completion call.
+            public string Pin { get; set; }
+            public string Token { get; set; }
         }
 
+        // Public checkout endpoint POSTed via fetch without a token, like the LNURL endpoints.
         [AllowAnonymous]
+        [IgnoreAntiforgeryToken]
```
NicolasDorier, line 57: Does it means NFC wasn't working anymore?? Can you double check?

Kukks:
@NicolasDorier all addressed:

- `IsPinTransportSecure` now uses `Extensions.IsLocalNetwork` instead of `.IsLoopback` (heads-up: that also allows a plaintext-http PIN over a private LAN, not just loopback — it matches the boundary the LNURL lib itself uses in `Parse`/`EncodeBech32`; say the word if you'd rather keep it loopback-only)
- try/catch narrowed to just `SendRequest`
- checkout PIN state grouped into a single `pin` object
- trimmed the PR description, added a screenshot + real-world usage (Bolt Card issuers)

On the CSRF question: yes — NFC withdraw has been broken since #7199 applied global CSRF to UI controllers. `NFCController` inherits `Controller` so the filter caught it, and the checkout posts without a token → empty 400. `[IgnoreAntiforgeryToken]` restores it, and the new integration test confirms it (empty 400 before, green after).

NicolasDorier:
No screenshot

image

NicolasDorier:
I have ported https://github.com/btcpayserver/btcpayserver/commit/ebaa263a656e81ee164699f5c3d4104e1899e8cf on master as I don't know if you'll have time to cleanup before 2.4.1

PR #7431, opened 2026-07-07
Allow merchants to manually resend an invoice's triggered email

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver/discussions/6549

"My smtp relay went offline. So I have paid invoices whom did not receive their email tickets. It would be nice to be able to manually trigger an email for an invoice. The button could live in the invoice detail page where you could have a dropdown menu with your custom emails you can select to re-send."

This PR adds "Resend email" action to the invoice detail page, letting merchants manually re-fire one of their store's Email Rules against a specific invoice.

image

image

Cc. @NicolasDorier

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.WebSockets;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Abstractions.Constants;
@@ -11,6 +12,7 @@ using BTCPayServer.Abstractions.Models;
 using BTCPayServer.Client;
 using BTCPayServer.Client.Models;
 using BTCPayServer.Data;
+using BTCPayServer.Events;
 using BTCPayServer.Filters;
 using BTCPayServer.Models;
 using BTCPayServer.Models.AppViewModels;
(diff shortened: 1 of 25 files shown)


BTCPayServer/Controllers/UIInvoiceController.UI.cs
```diff
@@ -189,9 +194,61 @@ public async Task<IActionResult> Invoice(string invoiceId)
...
+
+            var model = new JObject { ["Store"] = new JObject { ["Id"] = invoice.StoreId, ["Name"] = store.StoreName } };
+            InvoiceTriggerProvider.AddInvoiceToModel(model, invoice, _linkGenerator);
+            _EventAggregator.Publish(new TriggerEvent(invoice.StoreId, trigger, model, new ResendTriggerOwner(invoice)));
+
+            TempData[WellKnownTempData.SuccessMessage] = StringLocalizer["Email resent."].Value;
+            return RedirectToAction(nameof(Invoice), new { invoiceId });
+        }
+
+        private class ResendTriggerOwner(InvoiceEntity invoice) : ITriggerOwner
```
NicolasDorier: You don't need this class, you can just pass the `InvoiceTriggerProvider`. It is a trigger owner, and the DI has it registered. (Actually I think you'd need to put WebhookTriggerOwner as internal and use this to wrap the InvoiceTriggerProvider)

NicolasDorier:
I think this PR works, however, I believe it is too narrowly scoped.

There should be a page dedicated to see recently sent email, which give the ability to see whether sending failed, and also resend them with mass actions.

While https://github.com/btcpayserver/btcpayserver/discussions/6549 reported a specific complain from @pavlenex , a similar issue could happen for any email sent from btcpayserver.

TChukwuleta:
@NicolasDorier I have implemented a new page dedicated to see recently sent email, whether successful or not and also resend them with mass actions.

I am also prunning down to the last 10 days

This is for store emails

image


This is for server emails

image

PR #7433, opened 2026-07-07
 Scope payout mark/mark-paid endpoints to the store in the route

atharrva01:
I found that `POST /api/v1/stores/{storeId}/payouts/{payoutId}/mark` (and `/mark-paid`) checks the `CanManagePayouts` policy against the `{storeId}` in the URL, but then looks the payout up by id alone. It never confirms the payout belongs to that store. `ApprovePayout`, `CancelPayout`, and `GetStorePayout` all scope by store; `mark` is the odd one out.

So a key that can manage payouts on its own store can change the state and proof of *any* store's payout, as long as it knows the payout id. That's cross-tenant write access to payment state. Worse, `InProgress → AwaitingPayment` is allowed, which clears the proof and re-arms the payout for the processor to pay again.

To be fair, you need the target payout id (160-bit random), so it's not sprayable. But payout ids get handed to recipients, so a malicious recipient with a throwaway store qualifies. It's cross-tenant write access to payment state, not an info leak.

The fix mirrors what `ApprovePayout` already does: I resolve the payout scoped to the route store first, and return not-found if it isn't there.

**How I verified it**

I wrote an integration test (`Repro_MarkPayout_CrossStore_IDOR` in `GreenfieldAPITests.cs`) that runs against a real BTCPay instance over HTTP: real Greenfield client, real auth pipeline, no calling controllers directly. It sets up two separate accounts: a victim store with an approved payout, and an attacker who's a normal non-admin with an API key scoped only to its own store.

The same test proves the key is genuinely unprivileged, then shows it can still tamper cross-store:

- The attacker gets a 403 reading the victim's payout (`GetStorePayout(victimStoreId, ...)`). The real `CanManagePayouts` policy denies it.
- The attacker then calls `POST /stores/{attackerStore}/payouts/{victimPayoutId}/mark` with the same key, same pipeline.

**The proof, same test, same real HTTP + auth pipeline, only the fix toggled:**

- **Fix off:** the mark call returns 200 and flips the victim's payout to `InProgress` with an attacker-supplied proof. The test fails at the "expect 404" assertion precisely because no error was thrown:
  ```
  Repro_MarkPayout_CrossStore_IDOR [FAIL]
  Expected: typeof(BTCPayServer.Client.GreenfieldAPIException)   // i.e. got 200, not an error
  at GreenfieldAPITests.Repro_MarkPayout_CrossStore_IDOR() line 1229
  ```
- **Fix on:** the cross-store mark/mark-paid return 404, the victim's payout is untouched, and the store's own owner can still mark its own payout. Test passes; `CanProcessPayoutsExternally` still passes too.

So the identical real-HTTP request goes from 200 (payout tampered) to 404 (rejected) with only the guard toggled. Denied on read (403) but allowed on write (200) is the bug in one line.

**Why this is a genuine reproduction, not a test artifact**

Three facts, all inside that one test, all through the real pipeline:

1. It's real HTTP. `MarkPayout → SendHttpRequest → _httpClient.SendAsync(...)` hits the running Kestrel server. No controller is instantiated directly.
2. The attacker key is genuinely unprivileged, proven by the pipeline itself. The 403 on `GetStorePayout` comes from the real `[Authorize(Policy = CanManagePayouts)]` handler. If the test were bypassing permission checks, that 403 wouldn't happen.
3. Same key, same pipeline: denied on read (403), allowed on write (200). That gap is exactly the bug.

**Reproduce it yourself**

```
dotnet test BTCPayServer.Tests --filter FullyQualifiedName~Repro_MarkPayout_CrossStore_IDOR
```

Fails on master, passes on this branch. Or do it manually with the two-account curl steps below:

1. Register account A, create a store, make an API key with *Manage payouts* for it.
2. Grab a payout id from store B (claim a payout from B's pull payment, since recipients see their own id).
3. `POST /api/v1/stores/{A}/payouts/{payout_id_from_B}/mark-paid` with A's key. On master this returns 200 and flips B's payout to Completed; with this PR, 404.


atharrva01:
hi @NicolasDorier what do you think about this bug? would love your thoughts whenever u get chance

PR #7440, opened 2026-07-09
POS: format keypad amounts using the store checkout language

### LINKED ISSUE #7365: [Bug]: Store checkout language selection has no affect on POS keypad comma separator nor button label

v2.3.9+4c7580104

Umbrel

### What happened?

I changed the store checkout language, and it had no affect on POS keypad comma separator (e.g. changing 1,234 to 1.234) nor the "Charge" button label.

Trying to use the software in a non-English setting.

### Relevant log output

```shell

```

### What browser do you use?

_No response_

### Additional information

_No response_

TowyTowy:
The Point of Sale keypad derived its number-formatting locale solely from the currency code (`getLocale` in `pos/common.js`: USD→en-US, EUR→de-DE, JPY→ja-JP, else `navigator.language`), so the store checkout language had no effect on how amounts are grouped and separated (e.g. `1,234` vs `1.234`), as reported in #7365.

This passes the store checkout `DefaultLang` to the POS view model and prefers it as the `Intl.NumberFormat` locale when set, falling back to the existing currency-based heuristic when the store has no checkout language configured — so default behavior and existing tests are unchanged.

Scope note: this fixes the number/separator formatting, which is the part that can follow the per-store checkout language. The keypad static text ("Charge", tab labels) is localized via the server-side `text-translate` system (a single server-wide language, not per-store); making it per-store would need client-side i18n and is left as a separate follow-up.

@NicolasDorier flagged in the issue that this should be better supported.

Fixes #7365

### DIFF
--- a/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
+++ b/BTCPayServer/Plugins/PointOfSale/Controllers/UIPointOfSaleController.cs
@@ -143,6 +143,7 @@ namespace BTCPayServer.Plugins.PointOfSale.Controllers
                 AppId = appId,
                 StoreId = store.Id,
                 HtmlLang = settings.HtmlLang,
+                DefaultLang = storeBlob.DefaultLang,
                 HtmlMetaTags= settings.HtmlMetaTags,
                 Description = settings.Description,
                 NotAvailable = StringLocalizer["Not available in Keypad POS"].Value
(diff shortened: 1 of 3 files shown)


NicolasDorier:
This might be a wider problem we need to address.
Because currently, the currency format everywhere in BTCPay Server is based on the currency.

I've started doing some work to properly display dates based on user's browser preference. (Should it even be a browser preference? or a store one? Or a user one?)
It probably need to follow this example.

I need to think more about a better solution. I think we have been in the wrong direction on formatting in general in BTCPay, and probably need a step back rather than continuing in this direction.

TowyTowy:
That's fair, and thanks for saying so plainly rather than leaving it in limbo.

Agreed the formatting story is bigger than this PR — keypad amounts are just one symptom of currency-driven formatting, and patching them in isolation would add another special case to the thing you're trying to unify. I'll close this so it isn't sitting in your queue implying a direction you've decided against.

If you do land on an approach for locale/date/currency formatting and there's a piece worth picking up, happy to have a go at it.

Issue #7027, opened 2025-12-12
Subscription API testing


ndeet:
Setup: I did not click anything in the UI but created store, setup wallet and generated an unrestricted api key.

**Endpoint: `POST /api/v1/stores/{storeId}/offering`**

1. In the post data the naming of the offering `appName` seems to be a copy&paste thing, should be renamed to `offeringName`
```
{
  "appName": "Example App",
  "successRedirectUrl": "https://example.com/success",
  ....
}
```

2. In the response there is also `appName` and `appId` that imo should be renamed? 

3: What is also confusing, why there is `id` and `appId`? `appId` seems to be redundant/useless here?

```
"id": "offering_6x2bhwEcp3ufsjKVsN",
        "storeId": "G1Ex5rpU81MtZJCqWVhe1pQKkALRaB5zdhRwVSwwzKCm",
        "appName": "Example App",
        "appId": "36gxPQK4pej273Mw9pvhfFq6hGWL",
```

**`POST /api/v1/stores/{storeId}/offerings/{offeringId}/plans`**

1. In the request params there is `features` (formerly entitlements) **missing**. Afaik they are hardcoded so would be good to also list the available ones. `can-login`, `one-store`, `unlimited-store` (nitpick: typo, missing s, should  be `unlimited-stores`)

2. (nitpick) In the list of request params the `name` should be at the top:

<img width="679" height="399" alt="Image" src="https://github.com/user-attachments/assets/779a227d-44a8-4a60-87d2-06b9da887e82" />

3. In the UI plan list `API ID` should likely be `Plan ID`:

<img width="977" height="419" alt="Image" src="https://github.com/user-attachments/assets/adea08a5-00e9-49d2-91a6-a1a29ed3fdc7" />

**`POST /api/v1/plan-checkout` **

1. (nitpick) importing the swagger json I get this error on the storeId param (while others work fine):
`"value": "reference #/components/schemas/StoreId not found in the OpenAPI spec"`

2. with the example payload I get the error:
` {
        "path": "NewSubscriberEmail",
        "message": "If customerSelector is specified, newSubscriberEmail cannot be specified"
    }`

In the description says it can be "Null" but it actually means the param needs to be removed from the payload because if the value is "Null" then the same error as above happens.

3. When not passing `newSubscriberEmail` but keeping `customerSelector` as e.g. "user@example.com" I get this:
`{"code":"subscriber-not-found","message":"The subscriber was not found"}` 

[...]

PR #7441, opened 2026-07-09
Greenfield: document offering and plan creation as returning 200

TowyTowy:
While going through the subscription Greenfield API in #7027, @ndeet noticed that creating an offering and creating a plan return `200 OK`, even though the API reference documents both as `201 Created`.

Both endpoints build their response by delegating to the corresponding `GET` helper (`GetOffering` / `GetOfferingPlan`), which returns `Ok(...)` — so the created resource came back with a 200. This wraps the success response to return 201 with the exact same body, matching the Swagger definition. It also lines up with the user-creation endpoint, the only other creation endpoint documented as 201, which already returns it.

Error and not-found paths are untouched (the wrapper only rewrites the successful `OkObjectResult`). No client or response-shape change, so existing consumers and the `CanUseSubscriptionAPI` integration test are unaffected — the C# client accepts any 2xx.

This is just the status-code point from #7027; the other items in that issue are separate, so I've left it as `Refs #7027` rather than closing it.

Refs #7027

### DIFF
--- a/BTCPayServer/wwwroot/swagger/v1/swagger.template.subscriptions.json
+++ b/BTCPayServer/wwwroot/swagger/v1/swagger.template.subscriptions.json
@@ -159,7 +159,7 @@
                     }
                 },
                 "responses": {
-                    "201": {
+                    "200": {
                         "description": "Offering created successfully.",
                         "content": {
                             "application/json": {
@@ -211,7 +211,7 @@
                     }
                 },
                 "responses": {
-                    "201": {
+                    "200": {
                         "description": "Plan created successfully.",
                         "content": {
                             "application/json": {


NicolasDorier:
@TowyTowy I believe every other routes creating entities in our API returns 200?
If that's the case, we should fix the documentation to align with what is done everywhere rather than the code.

TowyTowy:
You're right — the other entity-creation routes return `Ok(...)` (200), e.g. `CreateStore`. I've reverted the code change and instead updated the Subscriptions swagger so `CreateOffering` and `CreateOfferingPlan` both document a `200` response, matching the rest of the Greenfield API. It's now a docs-only change.

PR #7442, opened 2026-07-09
Implementation for external payment

TChukwuleta:
Aim to resolve https://github.com/btcpayserver/btcpayserver/discussions/6635

Adds a generic external payment method (MISC-EXTERNAL) for recording payments settled outside BTCPay without a plugin. No checkout UI,

An integrator creates an API key scoped to a new btcpay.store.canregisterexternalpayment permission, then calls the endpoint with necessary data, this creates a real PaymentData row through the normal PaymentService.AddPayment

@NicolasDorier trying to clean up some things and then add tests to it, but before I go ahead does this match what you had in mind, particularly the API shape and the decision to keep it to one generic method for now rather than named ones? Also curious whether the store-toggle and per-store named methods feel like they belong in this same PR or should be separate follow-ups

image

image

image


NicolasDorier:
Cooooool! This is really needed.

NicolasDorier:
@TChukwuleta can you run `EnsureSwaggerPermissionsDocumented` just so the test pass?

TChukwuleta:
Hey @NicolasDorier, I tested it
the test ran

image

TChukwuleta:
@NicolasDorier this is open for review

NicolasDorier:
Is `ExternalPaymentAllowedLabels` really needed? doesn't seems used. I think any label should just be authorized.
The reference is great, but very useful would also be a link to the external payment on the external system.

PR #7445, opened 2026-07-11
Improve topbar visual separation

pavlenex:
This PR giive the topbar a distinct background and bottom border, and increase the contrast of the search field and navigation icons. 

This makes the user account (icons on the right) and search section easier to identify without changing the existing layout or mobile behavior and it's very minimal code change.

@NicolasDorier pointed out a concern that in that case secondary search, like on the invoice or wallet can lose on visibility, so I am attaching screenshots to showcase that they don't imo.

## Before this PR

Screenshot 2026-07-11 at 16 12 57

## After this PR
Screenshot 2026-07-11 at 16 17 49
Screenshot 2026-07-11 at 16 19 42
Screenshot 2026-07-11 at 16 19 50
Screenshot 2026-07-11 at 16 20 03
Screenshot 2026-07-11 at 16 19 21

### DIFF
--- a/BTCPayServer/wwwroot/main/layout.css
+++ b/BTCPayServer/wwwroot/main/layout.css
@@ -161,7 +161,7 @@
     border: 0;
     cursor: pointer;
     outline: none;
-    color: var(--btcpay-body-text-muted);
+    color: var(--btcpay-nav-link);
     border-radius: var(--btcpay-border-radius-m);
 }
 
@@ -875,7 +875,8 @@
         min-height: var(--global-nav-height);
         padding: var(--btcpay-space-s) var(--content-padding-horizontal);
         justify-content: space-between;
-        background: var(--btcpay-body-bg);
+        background: var(--btcpay-header-bg);
+        border-bottom: var(--menu-border);
     }
 
     #StoreSelector {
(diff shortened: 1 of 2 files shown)


dstrukt:
FWIW, I'm still a fan of the same background / no bottom border, but instead a White background in the Search for the light theme, and a Black background for the dark theme.

Keeps it very minimal and still addresses the wash out concern that prompted the PR.

pavlenex:
@dstrukt can you do a PR of how you want it and we can compare?

dstrukt:
Sure, I'll get something up this evening or tomorrow morning!

dstrukt:
https://github.com/btcpayserver/btcpayserver/pull/7450 for consideration

cc @NicolasDorier

NicolasDorier:
I will merge this one, I think the top bar being set to `--btcpay-header-bg` make more sense, it reflect UI from VS Code.
And it also draw a bit more the attention on the global bar, which I think isn't a bad idea.

dstrukt:
Digging up this old PR, as I've been using the updated header, and it just feels messier, more cluttered, and more intrusive:

Screenshot 2026-07-29 at 11 58 41 AM

I still think this direction creates a cleaner distinction and looks better in almost every instance:

image

https://github.com/btcpayserver/btcpayserver/pull/7450

I reserve opinions on a lot of the subjective things, but I think it's worth revisiting imo.

cc @pavlenex @NicolasDorier

dstrukt:
Reposting for thoughts @NicolasDorier @pavlenex

PR #7447, opened 2026-07-12
Support running headed in container

deverickapollo:
Change enables headed browser for  playwright while in container to support devcontainer based development


NicolasDorier:
@deverickapollo can you just remove the `Server.PayTester.InContainer`. Somebody who wants headed in container can now just set the env variable.

PR #7448, opened 2026-07-12
Containerize store settings views into sections & clean up views

### LINKED ISSUE #6901: Improved Settings view containerization
# Goal

We have a lot of settings across BTCPay Server, and we've addressed the paging issue by pushing it to the left side bar, but the specific goal of this change is to improve the legibility and chunking of the long lists on each page.

Outside of better visual hierarchy and legibility, the additional long-term goal is to allow each of these sections to be easily discovered via a Global Search and thus better labelled, so that each corresponding section and it's contents Are more intuitive to the respective user.

The mocks below are fairly W.I.P, but they encapsulate the idea and the visual style intended. 

Each section will have a proper header, and its contents will be grouped based on its intended purpose. This will require a bit of work, and each page will likely need to be combed through and cleaned up.

Image

Image

There's quite a few open questions involved here, but hopefully this opens up the discussion so we can attempt to tackle this in sections.

dstrukt:
Containerizes the store settings views: each page's content is grouped into labelled sections, each in its own card, using one shared reusable style for consistent hierarchy and legibility. Also cleans up the page headers and applies the same styling to the related subpages.

## Before:

https://github.com/user-attachments/assets/fcfc2a47-a6a2-46ac-85d7-bbdc003e115a

Screenshot 2026-08-04 at 1 25 59 PM
Screenshot 2026-08-04 at 1 26 08 PM
Screenshot 2026-08-04 at 1 26 18 PM

## After:

https://github.com/user-attachments/assets/05edf93b-b232-4823-a7ca-374d1fa22d65

Screenshot 2026-08-04 at 1 10 04 PM
Screenshot 2026-08-04 at 1 10 13 PM
Screenshot 2026-08-04 at 1 10 21 PM

## Changes

- **Shared styling:** reusable `.settings-section` / `.settings-section-heading` classes for headed section cards.
- **Pages:** General, Rates, Checkout Experience, Email Settings, Email Rules, Access Tokens, Payout Processors, Roles, Store Users, Webhooks, and Forms.
- **Headers:** page descriptions moved into the header as a subtitle (full width, with action buttons wrapping below on mobile) instead of floating above the content.
- **Subpages:** Create Token, Update Role, Pairing Permission, and the On-Chain/Lightning payout processor configs use the same section styling.

Purely a layout/markup change — no settings, fields, or behavior were changed.

Closes: #6901

### DIFF
--- a/BTCPayServer/Plugins/Bitpay/Views/CreateToken.cshtml
+++ b/BTCPayServer/Plugins/Bitpay/Views/CreateToken.cshtml
@@ -17,6 +17,7 @@
     <partial name="_StatusMessage" />
     <div class="row">
         <div class="col-xxl-constrain col-xl-8">
+            <div class="settings-section">
             <div class="form-group">
                 <label asp-for="Label" class="form-label"></label>
                 @if (ViewBag.HidePublicKey)
@@ -28,7 +29,7 @@
             </div>
             @if (!ViewBag.HidePublicKey)
             {
-                <div class="form-group">
+                <div class="form-group@(((bool)ViewBag.ShowStores) ? "" : " mb-0")">
                     <label asp-for="PublicKey" class="form-label"></label>
                     <input asp-for="PublicKey" class="form-control" />
                     <span asp-validation-for="PublicKey" class="text-danger"></span>
@@ -37,7 +38,7 @@
             }
             @if (ViewBag.ShowStores)
             {
-                <div class="form-group">
+                <div class="form-group mb-0">
                     <label asp-for="Stores" class="form-label"></label>
                     <select asp-for="StoreId" asp-items="Model.Stores" class="form-select"></select>
                     <span asp-validation-for="StoreId" class="text-danger"></span>
@@ -47,6 +48,7 @@
             {
                 <input type="hidden" asp-for="StoreId" />
             }
+            </div>
         </div>
     </div>
 </form>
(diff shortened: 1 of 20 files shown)


NicolasDorier:
I think we shouldn't show this.

`Store Settings > General > Store Details` (also I think we already have a link for this page?)

Because  100% of the other links doesn't have the hierarchy in the name. Instead, they are under the "Group", `Store`.

If you added Branding it should rather be "Configure the store's branding"

dstrukt:
Make sense, will revert, I *think* there are a few cases that are not covered, but will document and do a focused follow up PR..

NicolasDorier:
Why you closed? I was commenting on the search entry text.
I am saying that those entry here shouldn't show `Store Settings > General > Branding` for example in those text search, but something more like `Configure store branding`. Where clicking on this one, would go to the anchor you've added in the group.

image

NicolasDorier:
Reopening as I think the sections make sense, my comments was only about the global search.

dstrukt:
@NicolasDorier I think this one is ready for review. If all looks good, I'll start doing the same for the Server Settings and the Account views.

I reverted the Global Search anchor code a while back as well, and broke it into quite a commits so we can easily trace any issues that might have arisen from such a large sweeping change.

dstrukt:
Pinging you again  @NicolasDorier as this one is also ready to go, and would like to merge asap if it looks good - will create rebase hell eventually.

PR #7450, opened 2026-07-13
Refine global search colors

dstrukt:
Small visual tweaks to the global search to make it read as one consistent surface across light and dark themes. It's more consistent with what we currently do, and still maintains a strong presence, while being minimal and unobtrusive.

## Screenshots

Screenshot 2026-07-13 at 10 22 18 AM
Screenshot 2026-07-13 at 10 22 27 AM
Screenshot 2026-07-13 at 10 42 18 AM
Screenshot 2026-07-13 at 10 42 26 AM

## Changes
- **Input bar background:** white on the light theme, our darkest shade on the dark theme.
- **Results panel / hover:** swapped the two so the panel background matches the input (white / darkest) and the hovered item is the offwhite/lighter shade instead of the other way around.
- **Hover corners:** fixed the result item to use a valid radius token so the hover highlight has rounded (4px) corners instead of square ones.

### DIFF
--- a/BTCPayServer/wwwroot/plugins/GlobalSearch/global-search.css
+++ b/BTCPayServer/wwwroot/plugins/GlobalSearch/global-search.css
@@ -19,7 +19,8 @@
     padding: 0 var(--btcpay-space-m);
     border: 1px solid var(--btcpay-body-border-light);
     border-radius: var(--btcpay-border-radius-l);
-    background: var(--btcpay-body-bg);
+    /* Darkest surface shade on dark theme (#0d1117), white on light theme */
+    background: var(--btcpay-bg-tile);
     transition: border-color var(--btcpay-transition-duration-fast), box-shadow var(--btcpay-transition-duration-fast);
 }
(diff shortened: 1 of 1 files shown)


NicolasDorier:
I've merged https://github.com/btcpayserver/btcpayserver/pull/7445, but I think this one is still adding some tweak that isn't really conflicting with 7445?

dstrukt:
Yeah it probably wouldn't cause any merge conflicts - as I solely focused on the search input background and dropdown colors alone, which would be white or black respectively, while given what was done in #7445, it no longer makes sense to do here, as the while topbar + search would effectively become white or black.

NicolasDorier:
> it no longer makes sense to do here

So you will push another commit to this PR for removing what doesn't make sense?

dstrukt:
Oh, sorry if that wasn't clear, it should just be closed, other than perhaps the background on the global search drop-down, it wouldn't be an improvement.

Will close.

PR #7463, opened 2026-07-17
Greenfield: support webhookUrl on invoice creation

atharrva01:
Closes #7052.

Webhooks can only be configured per store, so when several websites share one store, every site receives every invoice's events and has to filter out or forward what isn't theirs. The legacy BitPay API had a per-invoice notificationURL for exactly this, but it never made it to Greenfield, and the legacy one sends old unsigned IPN payloads anyway.

This adds an optional `webhookUrl` (plus `webhookSecret`) to invoice creation. When set, all webhook events of that invoice are also delivered to that URL, signed with `BTCPay-Sig` like regular webhooks. Store webhooks behave exactly as before, and nothing changes if you don't pass the field. The URL is also visible on the invoice details page.

Tested on a local instance with two receivers sharing one store: one as the store webhook, the other passed as `webhookUrl` on a single invoice. That invoice's created/settled events arrived at both endpoints with a valid signature from the per-invoice secret, while other invoices only hit the store webhook. Invalid URLs are rejected at creation. There's a new integration test covering delivery, signature and validation.

One note: since these webhooks aren't stored on the store, their deliveries don't appear in the webhook delivery list and can't be redelivered from the UI. Automatic retry on failed delivery still applies.

pr-evidence-invoice-webhookurl

### DIFF
--- a/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
+++ b/BTCPayServer.Client/Models/CreateInvoiceRequest.cs
@@ -8,5 +8,7 @@ namespace BTCPayServer.Client.Models
         [JsonConverter(typeof(NumericStringJsonConverter))]
         public decimal? Amount { get; set; }
         public string[] AdditionalSearchTerms { get; set; }
+        public string WebhookUrl { get; set; }
+        public string WebhookSecret { get; set; }
     }
 }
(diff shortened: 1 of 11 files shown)


atharrva01:
hi @NicolasDorier this adds an optional webhookUrl (+ secret) on invoice creation so sites sharing one store get signed webhook events for their own invoices only, store webhooks untouched.

PR #7465, opened 2026-07-17
[WIP] Redesign the invoice detail view

dstrukt:
## Summary

Redesigns the invoice detail view (**Invoices → Invoice**) into a card based layout, reusing existing BTCPay components and theme tokens. The redesign also answers the question of why a merchant opens this page in the first place: usually because an invoice is in an unexpected state and they need enough information to decide what to do.

## Screenshots

TODO

## What the page shows

**Header and hero.** Breadcrumb, title, status and exception badges, and the actions on one row. Below that, what was paid and what is still due, with a progress bar and the network fee totalled per currency.

**Contextual banners.** Partial, overpaid, paid late and invalid states each get a short explanation with the actions that make sense for that state.

**Order summary.** For point of sale and cart invoices, an itemised table with subtotal, tax and total.

**Payments.** One row per payment with method, amount, fee, confirmations and transaction, expandable for the details. Replaced payments are struck through.

**Activity.** Lifecycle events in plain language. Debug entries stay collapsed behind a toggle that says how many are hidden. Anything logged as a warning or an error is always shown, never collapsed.

**Invoice info and buyer.** All the well known metadata fields, including tax, tax on tip and whether the order is physical. Order Id links to the external system when an order URL is present. The comment sits here as a row with an inline editor.

**Receipt data and additional information.** Anything in the metadata beyond the well known fields, rendered as a readable table, with the raw JSON kept as a fallback.

**Webhooks.** Every delivery with its event, HTTP status code, time, a link to the exact payload that was sent, and a redeliver action on failures. The section is always shown, so an empty state can say whether nothing was sent because nothing is configured, or because this invoice simply did not trigger anything.

## Partial payments on an expired invoice

Previously the page told the merchant to send the payer an updated link no matter what state the invoice was in. Once an invoice expires that link is a dead end, so the advice was wrong exactly when it mattered.

The banner now distinguishes three cases:

* Still live: send the payer the link for the remainder.
* Expired but still monitored: payments are still detected until the monitoring date, which is now shown, so the link still works.
* Expired and no longer monitored: the link will not credit anything, and the message says so.

There is also a **Charge remainder** action. It opens the invoice form prefilled with the outstanding amount, the same currency and the same Order Id, with the original invoice recorded in the metadata. The customer pays the difference, no refund is needed, and both invoices reconcile against one order. **Issue refund** is offered alongside it for merchants who would rather return the partial payment.

## Notes for reviewers

* `HttpCode` was already stored on webhook deliveries but not exposed to the view. It is now on the delivery view model, so the status column shows the real code rather than just success or failure.
* The invoice creation page previously ignored everything except the store id, so it could not be prefilled. It now accepts an amount, currency, order id and source invoice id, which is what Charge remainder uses.
* The comment uses the same markup and classes as the pattern merged in #7464, so the existing browser test continues to drive it.

## Not included

* Automatic webhook retries are saved as separate deliveries and are not counted, so the status column shows the HTTP code without a retry position.

### DIFF
--- a/BTCPayServer/Controllers/UIInvoiceController.UI.cs
+++ b/BTCPayServer/Controllers/UIInvoiceController.UI.cs
@@ -162,6 +162,8 @@ namespace BTCPayServer.Controllers
                                     .ToList()
             };
 
+            model.StoreHasWebhooks = (await _StoreRepository.GetWebhooks(invoice.StoreId)).Length > 0;
+
             var details = InvoicePopulatePayments(invoice);
             model.CryptoPayments = details.CryptoPayments;
             model.Payments = details.Payments;
@@ -1160,7 +1162,8 @@ namespace BTCPayServer.Controllers
         [HttpGet("/stores/{storeId}/invoices/create")]
         [HttpGet("invoices/create")]
         [Authorize(Policy = Policies.CanCreateInvoice, AuthenticationSchemes = AuthenticationSchemes.Cookie)]
-        public async Task<IActionResult> CreateInvoice(InvoicesModel? model = null)
+        public async Task<IActionResult> CreateInvoice(InvoicesModel? model = null, decimal? amount = null,
+            string? currency = null, string? orderId = null, string? sourceInvoiceId = null)
         {
             if (string.IsNullOrEmpty(model?.StoreId))
             {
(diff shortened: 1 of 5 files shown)


NicolasDorier:
Step in the right direction.
I feel that comments are taking lot's of estate for what it is.

Also, there should be a way to visualize the Invoice Metadata. Those can include relevant information from the user's own system.

dstrukt:
Yeah I noticed you tucked the comment away further, and this direction was based off what was merged in the original comment PR, but I can opt for a more minimal direction.

Mind elaborating a bit more of what you'd like to see? I have an idea, but all feedback is helpful in reducing scope, and getting the most ideal solution, considering how large the scope of change is.

Glad you like where it's headed though.

@NicolasDorier

NicolasDorier:
The question is why do merchant comes in this view?

In general it is because there is an invoice with some unexpected state (for example "Invalid" because it took too much time to confirm or "Paid Partial", or "Paid Late")

In this case we need to provide them as much information as we can to take a proper decision.

This information in general include: Invoice Metadata.
This is a JSON object defined by the user. (For example internal an OrderId)

The Metadata has some "well known field" (see https://docs.btcpayserver.org/Development/InvoiceMetadata/).
Those are popular field, that we can display not just as the JSONB blob, but more nicely. (OrderId + OrderUrl can be rendered into a clickable link to user's external system for example)

Also, there ware web hooks delivery. Sometimes, a webhook need to be sent on the customer's separate e-commerce, but didn't reach it. That is why they can re-deliver manually a webhook on this page, and see why it failed to deliver in the first place.

I like how you hide the "Debug logs". Most of those logs aren't really relevant. Though sometimes, there are warning and errors logs in there, that we shouldn't bury.

PR #7466, opened 2026-07-18
Update default global search bar suggestions

NicolasDorier:
The previous suggestions were not actually working in the first place.

Before

image

After

image


pavlenex:
Seems like UX downgrade? Now users don't know what they can search?

NicolasDorier:
I don't think so because anything the user type show up 20 different possible choices.
There are probably around 100 different pages the users can search for, so in any case discovering those would invoice typing something, anything.

Maybe by default I can show 20 random ones instead if there is no recent queries?

dstrukt:
Tend to agree with @pavlenex complete removal would be a downgrade, but agree with you @NicolasDorier that it's not as useful as it could be. A middle ground for the time being until more time is spent improving the more complex queries might be to hide some of the `Suggested` just for the `Recents` after it's been used a few times, one replaces the other kind of idea.

I also like the random by default, or if we had a basic list of what a lot of people struggle to find or *might* find useful.

Will give it more thought, but just some initial ideas, but also don't love just removing entirely, FWIW.

pavlenex (approved):
cACK https://github.com/btcpayserver/btcpayserver/commit/2643ce1ef4cf2a21ffc52d692d7f05ff04d52c0d

PR #7472, opened 2026-07-20
Limit Kraken ticker response to supported pairs

### LINKED ISSUE #7467: [Bug]: Background rate fetcher downloads Kraken's full ticker (~200 KB) every 30s for 24h after any request — default USD/EUR stores poll ~626 MB/day permanently

BTCPay Server v2.4.0+50ae4bfc5 — confirmed current via btcpay-update.sh on
2026-07-17 ("Image is up to date"). Code references below are from the
v2.4.0 source tarball.

Docker (btcpayserver-docker), Ubuntu 24.04 VM. Single store, BTC only,
Bitcoin Knots via custom fragment, Tor. My instance is now fine (details
below) — I'm filing because the default configuration carries this cost
permanently, and diagnosing it is currently very hard.

### What happened?

While preparing my server for a metered cellular failover link, I measured
~626 MB/day of traffic from the btcpayserver container to api.kraken.com —
a persistent keep-alive connection re-established ~70 s after every process
start, transferring ~434 KB/min indefinitely. My store was configured
yadio-only (primary source, rate script, default pairs; fallback disabled),
`docker inspect` showed no rate env vars, and `pg_dumpall | grep -i kraken`
returned **zero rows** — nothing configurable referenced kraken, yet it
polled continuously across restarts.

Reading the source explains everything. Four behaviors combine:

**1. Full-ticker payload for single-pair needs.**
`KrakenExchangeRateProvider` fetches the bare `/0/public/Ticker` endpoint —
every pair on the exchange, ~200+ KB per response — to answer lookups that
need one pair. (`BTCPayServer.Rating/Providers/KrakenExchangeRateProvider.cs`)

**2. 30-second cadence.**
`AddScheduledTask<RatesHostedService>(TimeSpan.FromSeconds(30))` plus
`BackgroundFetcherRateProvider.RefreshRate = TimeSpan.FromSeconds(30)`.
2 fetches/min × ~217 KB = ~434 KB/min — matching my packet capture exactly.
(`BTCPayServer/HostedServices/RatesHostedService.cs`,
`BTCPayServer.Rating/Providers/BackgroundFetcherRateProvider.cs`)

**3. A 24-hour polling tail after any request.**
`RatesHostedService.IsStillUsed` keeps refreshing any provider whose
`LastRequested` is within `TimeSpan.FromDays(1.0)`. One request touching a
kraken rule buys 24 hours × 626 MB/day of polling.

**4. The tail survives restarts, defeating diagnosis.**
`SaveRateCache()`/`TryLoadRateCache()` persist `LastRequested` to
`rates-cache.json` in the datadir volume. Every fresh process reloads it
and, if inside the 24 h window, resumes polling within ~70 s — while the
string "kraken" appears nowhere in the database or environment, and no log
line is emitted at any level about fetcher start, stop, or volume.

**Why default stores never escape the tail:** an empty `PreferredExchange`
resolves via `StoreBlob.GetPreferredExchange` →
`DefaultRulesCollection.GetRecommendedExchange`, which returns kraken for
USD/EUR/CAD/GBP/CHF (`BTCPayServer/Hosting/BTCPayServerServices.cs` ~L680)
— and for **any currency without a recommendation** via the hardcoded
fallback (`BTCPayServer/DefaultRules.cs` ~L71:
`RecommendedExchanges.TryGetValue(currency, out var ex) ? ex : "kraken"`).
Every invoice, checkout, and wallet-nav render re-touches the rule and
resets `LastRequested`, so the window never closes. **A default-configured
store pricing in USD downloads ~19 GB/month of Kraken ticker data,
permanently, invisibly.**

Confirmation the mechanism is complete: after I converted everything to
yadio, polling self-extinguished exactly 24 h after my last kraken-touching
request, and `rates-cache.json` now contains only a yadio entry. The code
predicted my box's behavior to the kilobyte and the hour.

Suggested directions (any subset helps):
- Fetch only requested pairs where the exchange API allows it, or scale
  RefreshRate to payload size (Kraken's ticker at 30 s is ~626 MB/day;
  yadio's at the same cadence is ~4 MB/day).
- Reconsider the 24 h IsStillUsed tail (e.g., 1–2 h covers dashboard use).
- Log fetcher lifecycle (started polling X / stopped polling X) so this is
  diagnosable without tcpdump.
- Document that unlisted currencies default to kraken (DefaultRules.cs:71).

1. Measured per-container traffic while planning a metered failover link;
   found ~626 MB/day from the btcpayserver container to a Cloudflare IP.
2. tcpdump SNI + conntrack identified api.kraken.com and the container.
3. Eliminated every configurable reference: store primary source, rate
   script, default pairs, fallback (disabled), env (docker inspect),
   database (pg_dumpall grep = zero rows). Polling persisted across
   `docker restart`, re-establishing within ~70 s each time.
4. Read the v2.4.0 source; found the recommendation table, the 30 s
   scheduled task, the 24 h IsStillUsed window, and the rates-cache.json
   persistence.
5. Verified the model's prediction: polling self-extinguished 24 h after
   the last kraken-touching request; rates-cache.json now yadio-only.

Repro on any default USD store: create an invoice (or just load the
dashboard), then watch the container hold a connection to 104.17.185–189
.205:443 (api.kraken.com) and transfer ~434 KB/min. `docker restart` the
container: the connection returns within ~90 s with no request made.

### Relevant log output

```shell
None exists — that is part of the report. The fetcher logs nothing at
default levels when polling, and nothing when its egress is blocked (I
firewalled 104.17.184.0/21:443 as containment; BTCPay was silent and
pricing was unaffected, consistent with the traffic serving no configured
function on my instance).

Wire evidence in place of logs:
- conntrack after restart: tcp ESTABLISHED src=<btcpayserver container>
  dst=104.17.189.205 dport=443, re-created ~70 s into process life.
- Interface counters bracketing the connection: ~434 KB/min received.
- rates-cache.json (datadir) before conversion: kraken state with
  LastRequested; after 24 h idle: yadio only.
```

### What browser do you use?

N/A — server-side network behavior (verified via tcpdump/conntrack, not a
browser).

### Additional information

_No response_

ZenulAbidin:
Fixes #7467 

The Kraken rate provider previously downloaded ticker data for every market Kraken offers, even though BTCPay only needs a small subset. This PR limits each ticker request to relevant currency pairs, reducing response size and bandwidth usage without changing the rate-fetch frequency.

The retained markets are:

- BTC paired with fiat currencies present in BTCPay’s currency catalogue and offered by Kraken.
- "DASH/USD", "DOGE/BTC", and "LTC/BTC", required by BTCPay’s built-in payment-network rate rules.
- "XMR/BTC", required by the external Monero plugin.

A provider test was also created to test Kraken pair fetching.


BTCPayServer.Tests/KrakenExchangeRateProviderTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Threading.Tasks;
+using BTCPayServer.Rating;
+using BTCPayServer.Services.Rates;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace BTCPayServer.Tests
+{
+    [Trait("Fast", "Fast")]
+    public class KrakenExchangeRateProviderTests
```
NicolasDorier: Remove this test, there is no point into making a mock test of that.

BTCPayServer.Rating/Providers/KrakenExchangeRateProvider.cs
```diff
@@ -30,7 +31,10 @@ public HttpClient HttpClient
         }
 
         HttpClient _LocalClient;
-        static readonly HttpClient _Client = new HttpClient();
+        static readonly HttpClient _Client = new HttpClient(new HttpClientHandler
```
NicolasDorier: Just inject the HttpClient from the constructor, the DI will inject it.
Then just configure the httpclientfactory to use compressed data.
I don't know why we have a signleton here.

NicolasDorier:
Instead of having complicated code to detect what pair make sense, I think we should just hard code the relevant pairs directly.

What are the current pairs? we should cherry-pick them directly.

ZenulAbidin:
> What are the current pairs? we should cherry-pick them directly.

Currently, this set resolves to BTC/AUD, BTC/CAD, BTC/CHF, BTC/EUR, BTC/GBP, BTC/JPY, BTC/USD, DASH/USD, DOGE/BTC, LTC/BTC, and XMR/BTC.

BTCPayServer/Hosting/BTCPayServerServices.cs
```diff
@@ -728,7 +728,12 @@ internal static void RegisterRateSources(IServiceCollection services)
 
             // Handmade providers
             services.AddRateProvider<HitBTCRateProvider>();
-            services.AddRateProvider<KrakenExchangeRateProvider>();
+            services.AddHttpClient<KrakenExchangeRateProvider>()
+                .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler
+                {
+                    AutomaticDecompression = System.Net.DecompressionMethods.All
+                });
+            services.AddSingleton<IRateProvider>(provider => provider.GetRequiredService<KrakenExchangeRateProvider>());
```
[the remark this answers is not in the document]
NicolasDorier: @ZenulAbidin so coderabbit is correct here. It's probaly a bug we also have elsewhere in the code.
The fix is rather to inject the `IHttpClientFactory` and create a new client when you need it instead.

NicolasDorier:
I am quite worried with this PR, as it is easy to miss pair people are or may be using in the future.
Probably should add any well known crypto currency to the list, not only the currently used.

ZenulAbidin:
As in fiat currencies or crypto?

What do you have in mind currently?

-------- Original Message --------

Kukks:
I would say concept nack here. Extra pairs are harmless and actually add in fallback options instead of specifically hardwired providers for specific rates.

slyronious:
Measured the batched form: ?pair= with the 11 pairs in this PR returns 3,098 bytes (vs 307 B for one pair, ~200 KB unfiltered). So the marginal cost is ~279 bytes per pair — at the 30 s cadence, 11 pairs is about 8.9 MB/day against the 626 MB/day I measured previously. 

Maybe that addresses the concern about missing pairs: the list can be liberal without losing much. 30 pairs ≈ 24 MB/day, 50 ≈ 40 MB/day (93.2% reduction). Erring toward over-inclusion costs a few hundred bytes per entry and keeps the vast majority of the saving, so the list doesn't have to be exactly right to be worth shipping.

ZenulAbidin:
Since 2.4.2 has already been taken, perhaps this PR could be adjusted to target 2.4.3?

NicolasDorier:
Ok so I updated the list.
Just on the list, it is 96% saving without even talking about the compression.
There is a lot of shitcoins we can indeed just remove

image

PR #7474, opened 2026-07-21
Adds table (wallet transactions) column controls (hide/show/re-order)

### LINKED ISSUE #7473: Add the ability to hide/show table columns
Tracking this simple task based on @NicolasDorier's request while implementing #7470

> I think that before adding one more column to this table, we should think about adding a way to show/hide columns from the table and save that in the localStorage of the user. This table is out of control right now.

Functionally this show enable the ability to hide/show table columns, with an additional goal of ordering the columns, and the "default" column/data items should be what was implemented in #7470 (essentially sans the Rate column)

dstrukt:
Closes: #7473. 

Adds a Columns control to the wallet transactions table so users can show, hide, and reorder columns, with the choice saved in localStorage.

- A gear dropdown lists the table's content columns, each with a checkbox to show/hide and a drag handle to reorder, plus a Reset to default action.
- Preferences persist in localStorage and reapply on page load and after infinite scroll loads more rows.
- Column agnostic: the control is built from the header cells tagged with data-col, so it adapts automatically as columns are added or removed (for example when the Type column from #7470 lands).
- At least one column always stays visible.
- The shared date column component gains an optional column key so it can participate; other tables that use it are unaffected.

## Screenshots

Screenshot 2026-07-22 at 9 35 43 PM

Screenshot 2026-07-20 at 11 45 07 PM


NicolasDorier:
Love it, add screenshots to the PR, it's because people reading the changelog needs something to visualize. (Also when we do posts)

I wonder if the settings gear is adapted for that. Some pages have already a settings button.
If that work well, we will eventually bring this pattern to other views. (focus just on wallet transactions for this PR)

dstrukt:
Will do, and most likely will utilize a different icon, have a few minor adjustments left to make.

NicolasDorier:
let me know when ready. UX wise, this is exactly what we need.

dstrukt:
Assuming all is good on the code side, we're ready to go @NicolasDorier.

Updated the icon, resolved the code review comments. 

I will rebase https://github.com/btcpayserver/btcpayserver/pull/7470 on this one, and some of the changes on that PR will help improve this view + mobile.

PR #7428, opened 2026-07-04
feat : add rtl language support with custom css

teamssUTXO:
Closes #7408 

This PR add RTL language support by changing the CSS.

### Image : 

image

PR #7481, opened 2026-07-23
Fix missing Bootstrap styles in RTL layouts

pavlenex:
## Summary

- load the committed Bootstrap RTL bundle from its actual static asset path
- fix both the shared layout and printable invoice receipt
- add a regression test that verifies every RTL stylesheet referenced by Razor views exists under `wwwroot`

## Cause

The RTL support merged in #7428 referenced `/main/rtl/bootstrap.rtl.css`, while the asset was committed as `/main/bootstrap/bootstrap.rtl.css`. The resulting 404 left the generated RTL layout overrides without their Bootstrap base and broke the Arabic UI layout.

## Tests

- RTL static-asset regression test: passed
- Full `BTCPayServer.Tests` run: project compiled; 233 passed and 59 failed before exit 137. The failures were environmental, primarily missing Playwright Chromium, plus an altcoin synchronization timeout.

### DIFF
--- a/BTCPayServer.Tests/StaticAssetTests.cs
+++ b/BTCPayServer.Tests/StaticAssetTests.cs
@@ -1,0 +1,34 @@
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Xunit;
+
+namespace BTCPayServer.Tests;
+
+public class StaticAssetTests
+{
+    [Fact]
+    [Trait("Unit", "Unit")]
+    public void RtlStylesheetReferencesExist()
+    {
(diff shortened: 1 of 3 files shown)


teamssUTXO:
Hey @pavlenex, good catch, it's because I moved the rtl.css file in the boostrap directory as Nicolas suggested but this broke links between css and view files. Do you want me to fix it ?

pavlenex:
Go for it @teamssUTXO!

teamssUTXO:
I think I can't push here in your PR so I open another

teamssUTXO:
What your test was for ?

PR #7482, opened 2026-07-24
fix bootstrap RTL CSS links in layouts

teamssUTXO:
## Fix Summary

Fix both the shared layout and printable invoice receipt

## Cause

The RTL support merged in https://github.com/btcpayserver/btcpayserver/pull/7428 referenced /main/rtl/bootstrap.rtl.css, while the asset was committed as /main/bootstrap/bootstrap.rtl.css. The resulting 404 left the generated RTL layout overrides without their Bootstrap base and broke the Arabic UI layout.

cc : @pavlenex

### DIFF
--- a/BTCPayServer/Views/Shared/LayoutHead.cshtml
+++ b/BTCPayServer/Views/Shared/LayoutHead.cshtml
@@ -26,7 +26,7 @@
 }
 @if (isRtl)
 {
-    <link href="~/main/rtl/bootstrap.rtl.css" asp-append-version="true"  rel="stylesheet" />
+    <link href="~/main/bootstrap/bootstrap.rtl.css" asp-append-version="true"  rel="stylesheet" />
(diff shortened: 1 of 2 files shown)


pavlenex:
@teamssUTXO please add a video or screen shots, we can't really trust ai blindly without trying locally, it's what made the bug in the first place. With surge of PR's I also cannot prioritize PR's which people don't test locally first.

teamssUTXO:
Sure here is the video

https://github.com/user-attachments/assets/91c44016-aff4-433a-aa4f-95ccf09525c8

To be on the same page, I understand what you say and agree, but this bug isn't an AI mistake, Nicolas just told me to change a file's path and did this simplest thing without testing again...

NicolasDorier:
Yeah, I confirm this wasn't an AI bug, it was working before I asked him to change this.
Surprised coderabbit didn't catch the mistake though.

PR #7484, opened 2026-07-28
Fix pending multisig signature retries

rockstardev:
Pre 2.4.0, each distinct signed PSBT was retained, combined, and finalization was retried.
Commit e48a8438 introduced:Math.Min(validExpectedPartialSigCount, multisigParams.SignatureCount)

For 2-of-4, adding signer three produces Math.Min(3, 2) == 2.
HasMeaningfulDelta therefore sees 2 → 2 and returns false.
The service exits before storing signer three and before retrying TryFinalize.
The record stays Pending; the UI only shows Broadcast for State == Signed.

In some cases NBitcoin finalization errors are thrown, added logging to identify exact issues with why edge cases are happening.

### DIFF
--- a/BTCPayServer.Tests/MultisigTests.cs
+++ b/BTCPayServer.Tests/MultisigTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


thgO-O (approved):
LGTM! 

Non-blocking:

Non-multisig transactions report 0/0 progress, so normal finalization failures are logged as warnings (0 >= 0). Maybe we should also require SignaturesNeeded > 0. 

The optional additionalPsbt parameter on BuildEffectivePsbt no longer has any callers. Consider removing it.

PR #7488, opened 2026-08-01
PSBT signing and pending transaction status made more reliable

rockstardev:
Addressing two transaction edge cases:

- SegWit PSBTs now include the transaction data expected by more signing devices and wallets.
- Fully signed multisig transactions no longer remain stuck as **Pending** because their stored status is stale.

On safety front - BTCPay only changes a transaction to **Signed** when it has enough signatures **and** the PSBT can be finalized successfully. Incomplete or invalid transactions remain **Pending**.

Existing PSBT data is preserved for compatibility.

* Before HWI 3.2: HWI read `non_witness_utxo`, found the referenced output, extracted its value and script, then constructed Jade’s legacy signing request. So `non_witness_utxo` alone was sufficient.
* HWI 3.2 + Jade firmware ≥ 0.1.47: HWI sends the **raw PSBT** to Jade’s new native sign_psbt implementation. HWI no longer performs that extraction first.
* We now want BTCPay-generated PSBTs to include `witness_utxo` when `non_witness_utxo` is present for a SegWit input.

We follow what Bitcoin Core is doing:
https://github.com/bitcoin/bitcoin/blob/master/src/psbt.cpp#L741-L749

image

## Tested

- Native, nested, and Taproot SegWit PSBT inputs
- Legacy inputs remain unchanged
- Stale multisig status repairs itself and persists
- Invalid signatures do not mark a transaction as signed

No UI changes.

### DIFF
--- a/BTCPayServer/Extensions/PSBTExtensions.cs
+++ b/BTCPayServer/Extensions/PSBTExtensions.cs
@@ -26,9 +26,9 @@ namespace BTCPayServer
 
                 var previousOutput = nonWitnessUtxo.Outputs[(int)input.PrevOut.N];
                 var redeemScript = input.RedeemScript;
-                if (IsSegwit(previousOutput.ScriptPubKey) ||
+                if (previousOutput.ScriptPubKey.IsScriptType(ScriptType.Witness) ||
                     redeemScript is not null &&
-                    IsSegwit(redeemScript) &&
+                    redeemScript.IsScriptType(ScriptType.Witness) &&
                     redeemScript.Hash.ScriptPubKey == previousOutput.ScriptPubKey)
                 {
                     input.WitnessUtxo = previousOutput;
@@ -36,13 +36,6 @@ namespace BTCPayServer
             }
 
             return psbt;
-        }
-
-        private static bool IsSegwit(Script scriptPubKey)
-        {
-            return PayToWitPubKeyHashTemplate.Instance.CheckScriptPubKey(scriptPubKey) ||
-                   PayToWitScriptHashTemplate.Instance.CheckScriptPubKey(scriptPubKey) ||
-                   PayToTaprootTemplate.Instance.CheckScriptPubKey(scriptPubKey);
         }
     }
 }
--- a/BTCPayServer/Extensions/PSBTExtensions.cs
+++ b/BTCPayServer/Extensions/PSBTExtensions.cs
@@ -12,5 +12,37 @@ namespace BTCPayServer
             var after = psbt.ToBase64();
             return before != after;
         }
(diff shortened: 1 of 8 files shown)


BTCPayServer/Extensions/PSBTExtensions.cs
```diff
@@ -12,5 +12,37 @@ public static bool PSBTChanged(this PSBT psbt, Action act)
...
+                    redeemScript.Hash.ScriptPubKey == previousOutput.ScriptPubKey)
+                {
+                    input.WitnessUtxo = previousOutput;
+                }
+            }
+
+            return psbt;
+        }
+
+        private static bool IsSegwit(Script scriptPubKey)
```
thgO-O: Non-blocking: could we use scriptPubKey.IsScriptType(ScriptType.Witness) here instead? NBitcoin already provides generic witness-program detection, so we wouldn’t need to maintain a list of P2WPKH/P2WSH/P2TR as new witness versions are introduced.

BTCPayServer/Plugins/Wallets/Controllers/UIWalletsController.PSBT.cs
```diff
@@ -69,6 +69,7 @@ public async Task<CreatePSBTResponse> CreatePSBT(string storeId, BTCPayNetwork n
             if (psbt == null)
                 throw new NotSupportedException(StringLocalizer["You need to update your version of NBXplorer"]);
 
+            psbt.PSBT.AddWitnessUtxoToSegwitInputs();
```
thgO-O, line 72: This covers regular wallet sends, but CPFP and RBF call explorer.CreatePSBTAsync directly in UIWalletsController.cs and then pass those PSBTs to createpending, sign, or analyze-psbt. Fee bumps can therefore still miss WitnessUtxo. Could we apply this helper to those two responses as well, or centralize the normalization for all BTCPay-generated PSBTs?

rockstardev: [ede7cc7](https://github.com/btcpayserver/btcpayserver/pull/7488/commits/ede7cc706797ae2328517a6ca4a177a538137a2d)

BTCPayServer/Extensions/PSBTExtensions.cs
rockstardev: [ede7cc7](https://github.com/btcpayserver/btcpayserver/pull/7488/commits/ede7cc706797ae2328517a6ca4a177a538137a2d)

thgO-O (approved):
LGTM

PR #7493, opened 2026-08-05
Add store-wide tax rates as server plugin: PayButton, Crowdfund

TChukwuleta:
Adds a Tax plugin (Plugins/Tax/) with StoreTaxRate, TaxCalculator, TaxRateResolver, and a settings page to define named store tax rates.

BTCPay Server apps can opt into a rate via TaxRateId (Done for Corwdfund and Pay button)

image

Would love a first look plus feedback 

Cc, @NicolasDorier

### DIFF
--- a/BTCPayServer/Data/StoreBlob.cs
+++ b/BTCPayServer/Data/StoreBlob.cs
@@ -29,6 +29,7 @@ namespace BTCPayServer.Data
             RecommendedFeeBlockTarget = 1;
             PaymentMethodCriteria = new List<PaymentMethodCriteria>();
             ReceiptOptions = InvoiceDataBase.ReceiptOptions.CreateDefault();
+            TaxRates = new List<StoreTaxRate>();
         }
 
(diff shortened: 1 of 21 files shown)


NicolasDorier:
I think for this one, you will have to see with @pavlenex, as he is the one who often pushed for it.

I'm always shy to start this, because there are many different way of calculating taxes, and choosing the wrong path will be very difficult to backtrack. I think the best thing to do is to copy what more mature solutions are doing (that mean following the path of ERPs such as Microsoft Dynamics)

I believe that this is something we should do though and it scares me. Ideally, somebody with prior ERP experience could guide us and provide good feedback.

dstrukt:
I'm with @NicolasDorier w.r.t introducing this as a whole, despite knowing we'll probably be forced to down the road.

However, what I'm most curious is, why introduce as a plugin instead of a Store / Server Setting? It can be disabled by default, then enabled, with all of the same UI you're presenting here, and then the Opt in from that view or other respective views.

Maybe there's something I'm missing, but wanted to hear your thinking.

Otherwise, visually and functionally, it looks solid!

PR #7494, opened 2026-08-05
NFC: block SSRF via LNURL-Withdraw target validation

r1ckstardev:
Server-Side Request Forgery hardening in the NFC plugin's LNURL-Withdraw endpoint.

## Context

`SubmitLNURLWithdrawForInvoice` (BTCPayServer/Plugins/NFC/NFCController.cs:46-48) is `[AllowAnonymous][IgnoreAntiforgeryToken]` and takes a caller-supplied `request.Lnurl`. After parsing, the server issues a fetch to whatever URL the caller supplied.

The response body is not returned to the caller, so this is a blind primitive. But an unauthenticated remote can still use error timing / boolean status as a reachability oracle for:

- Cloud metadata endpoints (169.254.169.254)
- Localhost admin panels on the BTCPay host
- Adjacent-container services on a shared docker network

Severity is low (no exfil, no auth bypass, no fund movement) but the class is real and the fix is small, so this is opportunistic hardening rather than an incident response.

## Change

Two guards before `LNURL.FetchInformation` runs:

- Require `uri.Scheme` in {http, https}, rejecting file / ftp / gopher / data.
- Reject hosts flagged by `Extensions.IsLocalNetwork(...)` (link-local, RFC1918, `.internal` / `.local` / `.lan`, single-label). Same helper the existing `ILightningClient.IsSafe` path already relies on for LN connection strings.

No new configuration surface. If an operator legitimately needs to point at a private LNURL-Withdraw server, we can revisit the opt-out once that request actually surfaces.

## Diff

9 added lines in one file, no dependencies touched.

## Notes

- .onion targets remain reachable: `IsLocalNetwork` only matches `.internal` / `.local` / `.lan` / single-label / RFC1918 / link-local, so `xxx.onion` still passes and continues to route through the existing onion `HttpClient` factory branch.
- Attribution: original finding via an AI-assisted audit pass (Kimi K3); this PR is the manually-verified minimal-surface fix.

### DIFF
--- a/BTCPayServer.Tests/NFCTests.cs
+++ b/BTCPayServer.Tests/NFCTests.cs
@@ -1,0 +1,94 @@
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using BTCPayServer.Client;
+using BTCPayServer.Client.Models;
+using BTCPayServer.Lightning;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace BTCPayServer.Tests;
+
+[Collection(nameof(NonParallelizableCollectionDefinition))]
+public class NFCTests(ITestOutputHelper testOutputHelper) : UnitTestBase(testOutputHelper)
+{
+    // Documents the SSRF surface on the [AllowAnonymous] NFC LNURL-Withdraw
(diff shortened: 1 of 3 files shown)


rockstardev:
Closing in favor of #7495

PR #7495, opened 2026-08-05
Harden NFC LNURL withdraw requests against SSRF

rockstardev:
Superseeds #7494 

`SubmitLNURLWithdrawForInvoice` (BTCPayServer/Plugins/NFC/NFCController.cs:46-48) is `[AllowAnonymous][IgnoreAntiforgeryToken]` and takes a caller-supplied `request.Lnurl`. After parsing, the server issues a fetch to whatever URL the caller supplied.

The response body is not returned to the caller, so this is a blind primitive. But an unauthenticated remote can still use error timing / boolean status as a reachability oracle for:

- Cloud metadata endpoints (169.254.169.254)
- Localhost admin panels on the BTCPay host
- Adjacent-container services on a shared docker network

Severity is low (no exfil, no auth bypass, no fund movement) but the class is real and the fix is small, so this is opportunistic hardening rather than an incident response.

### DIFF
--- a/BTCPayServer.Tests/NFCTests.cs
+++ b/BTCPayServer.Tests/NFCTests.cs
@@ -1,0 +1,91 @@
+using System;
+using System.Net;
+using System.Net.Http;
+using BTCPayServer.Plugins.NFC;
+using Xunit;
+
+namespace BTCPayServer.Tests;
+
+[Trait("Fast", "Fast")]
+public class NFCTests
+{
+    [Theory]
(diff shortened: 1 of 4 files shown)


NicolasDorier:
Concept NACK. This isn't exploitable, add a lot of code that need review and have potential bugs, and doesn't cover unsafe internal DNS addresses. (which depends on where btcpay is installed)

I am fine for a comment to avoid AI to pester us on this.
But I'm fine on leaving things as is, as attackers waste token and time to exploit something that isn't exploitable.

PR #7498, opened 2026-08-07
README: add temporary security notice for v2.4.2

orangesurf:
Adds copy of [official security advisory](https://blog.btcpayserver.org/security-advisory-btcpay-server-2-4-2/) to Readme 

image


NicolasDorier:
Not adding it here. Because not everyone is impacted. Those not running the btcpay-docker are not impacted.

PR #7518, opened 2026-08-26
Fix blank plugin details panel for unlisted plugins

### LINKED ISSUE #7508: [Bug]: Selecting an unlisted plugin from directory search opens a blank details panel

v2.4.3+8f2f998ee

Start9 / StartOS

### What happened?

An unlisted plugin can be found by searching for its exact name in the Plugin Directory, but selecting the search result opens an empty details panel on the right. The plugin cannot be installed from that panel because no metadata or install button is rendered.

Expected behavior: selecting an unlisted plugin returned by an exact search should display its details and allow installation.

Actual behavior: the offcanvas opens, but its content is blank.

The issue was reproduced with the unlisted Bare Bitcoin plugin (`barebitcoin`, `BTCPayServer.Plugins.BareBitcoin`, stable version `2.0.1.0`). Its public Plugin Builder details page is available at:

https://plugin-builder.btcpayserver.org/public/plugins/barebitcoin

1. Open **Server Settings → Plugins → Plugin Directory**.
2. Search for the exact name of an unlisted plugin, for example `Bare Bitcoin`.
3. Confirm that the plugin appears with the **Unlisted** badge.
4. Select the plugin result.
5. Observe that the right-hand details panel opens with no content.

This also occurs when navigating directly to:

`/server/plugins/directory?selectedSlug=barebitcoin`

### Relevant log output

No relevant server log entry was observed. The panel request itself reproduces the failure:

```text
GET /server/plugins/panel?slug=barebitcoin
HTTP 200 OK
Content-Type: text/html; charset=utf-8
Body: "\n"
```

### What browser do you use?

Google Chrome 151.0.0.0 on macOS

### Additional information

The embedded Plugin Builder directory loads normally, returns the unlisted plugin for the exact search, and sends `pb:plugin-selected`. The blank panel appears to be caused by the server-side panel lookup:

- `SelectedPluginPanel` calls `pluginService.GetRemotePlugins(null)`.
- The no-search remote catalog does not include unlisted plugins.
- `CreateSelectedPluginPanelViewModel` cannot match `CatalogSlug == "barebitcoin"` and returns an empty view model.
- `_SelectedPluginPanel` consequently renders an empty response.

The selected-plugin lookup likely needs a slug/name-specific remote lookup that permits an unlisted exact match, consistent with the directory search behavior.

ZenulAbidin:
Fixes #7508

Unlisted plugins can be found by searching the Plugin Directory, but selecting the search result opened an empty details panel with no metadata or install button. This PR falls back to a slug-specific catalog lookup when the selected plugin is not part of the default listing, mirroring how the directory search itself surfaces unlisted plugins.

- Selecting an unlisted plugin now shows its details and allows installation.
- Direct links like `/server/plugins/directory?selectedSlug=barebitcoin` render correctly.
- Listed plugins are unaffected and do not trigger the extra lookup.

Unit tests were added covering the fallback for unlisted slugs and the no-op for listed ones.

### DIFF
--- a/BTCPayServer.Tests/PluginManagerTests.cs
+++ b/BTCPayServer.Tests/PluginManagerTests.cs
@@ -289,6 +289,138 @@ namespace BTCPayServer.Tests
         }
 
         [Fact]
+        public async Task SelectedPluginPanel_FallsBackToSearchForUnlistedPlugin()
+        {
+            var requestedQueries = new List<string>();
+            using var httpClient = new HttpClient(new TestHttpMessageHandler(request =>
+            {
+                requestedQueries.Add(request.RequestUri!.Query);
+                if (request.RequestUri.Query.Contains("searchPluginName=unlistedplugin"))
+                {
+                    return TestHttpMessageHandler.JsonResponse("""
+                                                                [{
+                                                                    "projectSlug": "unlistedplugin",
+                                                                    "buildId": 2,
+                                                                    "manifestInfo": {
+                                                                        "identifier": "UnlistedPlugin",
+                                                                        "name": "Unlisted Plugin",
+                                                                        "version": "1.0.0"
+                                                                    },
+                                                                    "buildInfo": {}
+                                                                }]
+                                                                """);
+                }
+
+                return TestHttpMessageHandler.JsonResponse("""
+                                                            [{
+                                                                "projectSlug": "listedplugin",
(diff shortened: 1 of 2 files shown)


thgO-O:
Thanks for working on this. I closed the PR to avoid duplicate work, as I’m assigned to #7508 and implementing it.

PR #7523, opened 2026-08-27
Fix/monetization cannot override admin lockout

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver2/issues/160

Two separate concerns: monetization could clear an administrator lockout, and the anonymous signup form could attach a subscription to someone else's existing account.

To reproduce: Reproduced manually: registered a user, disabled it via admin, confirmed login fails with "Your user account is currently disabled." Went through /monetization/new-user with that same email, paid, got "Payment Successful." Logged in again, account was active again. No admin action involved.

I removed MonetizationHostedService from touching LockoutEnd.

also plan checkout controller only checked whether the submitted email already had a subscription to the specific offering being purchased. It never checked whether the email belonged to an existing BTCPay login account. There is now a usermanager account check

Cc. @NicolasDorier

### DIFF
--- a/BTCPayServer/Plugins/Subscriptions/Controllers/UIPlanCheckoutController.cs
+++ b/BTCPayServer/Plugins/Subscriptions/Controllers/UIPlanCheckoutController.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 using System.Threading;
 using System.Threading.Tasks;
 using BTCPayServer.Client.Models;
@@ -7,6 +7,7 @@ using BTCPayServer.Models;
 using BTCPayServer.Services;
 using BTCPayServer.Views.UIStoreMembership;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Localization;
--- a/BTCPayServer/Events/UserEvent.cs
+++ b/BTCPayServer/Events/UserEvent.cs
@@ -61,6 +61,11 @@ public class UserEvent(ApplicationUser user)
         public bool Bypass { get; } = bypass;
         public RequestBaseUrl RequestBaseUrl { get; } = requestBaseUrl;
     }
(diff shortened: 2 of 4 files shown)


BTCPayServer/Plugins/Subscriptions/Controllers/UIPlanCheckoutController.cs
```diff
@@ -112,6 +114,12 @@ public async Task<IActionResult> PlanCheckout(string checkoutId, PlanCheckoutVie
                     ModelState.AddModelError(nameof(vm.Email), "This email already has a subscription to this offering");
                     return await PlanCheckout(checkoutId);
                 }
+                var existingAccount = await userManager.FindByEmailAsync(checkout.NewSubscriberEmail!);
```
NicolasDorier: Can't do this, as this is only relevant to monetization, not to every offerings.

BTCPayServer/Plugins/Subscriptions/Controllers/GreenfieldOfferingController.cs
```diff
@@ -419,6 +421,10 @@ public async Task<IActionResult> ProceedPlanCheckout(string checkoutId, [FromQue
             if (checkout is { NewSubscriber: true, NewSubscriberEmail: null })
                 ModelState.AddModelError(nameof(email), "You need to pass `email` as query string");
 
+            if (checkout is { NewSubscriber: true, NewSubscriberEmail: not null } &&
+                await userManager.FindByEmailAsync(checkout.NewSubscriberEmail) is not null)
+                ModelState.AddModelError(nameof(email), "An account with this email already exists. Please log in to continue.");
```
NicolasDorier: Can't do this, only monetization works like this

NicolasDorier:
I can't repro, when I go to monetization/new-user, I can't just reactivate the user.
EDIT: It's only for users already in the subscription.

image

TChukwuleta:
@NicolasDorier 

Updated the PR to suspend subscription (if exists) if an admin disables a user

NicolasDorier:
@coderabbitai review

PR #7526, opened 2026-08-28
Rename CryptoMarket rate provider to Notbank

NicolasDorier:
CryptoMarket exchange rates stopped working because the old CryptoMarket API host no longer resolves. CryptoMarket has moved to Notbank, so this updates the rate source to use Notbank for the same market rates.

Existing stores and rate rules that referenced `cryptomarket` are migrated to `notbank`, so merchants keep their configured rate source after upgrading.

Public announcement: https://www.theclinic.cl/2025/12/09/notbank-nueva-etapa-cryptomarket/


NicolasDorier:
@coderabbitai review

PR #7533, opened 2026-08-31
Validate pay button server IPN url

Kukks:
Rejects Server IPN values targeting local network addresses on the public pay button invoice endpoint, and extends IsLocalNetwork to cover link-local ranges.

### DIFF
--- a/BTCPayServer.Tests/FastTests.cs
+++ b/BTCPayServer.Tests/FastTests.cs
@@ -63,6 +63,22 @@ namespace BTCPayServer.Tests
             Assert.Equal("https://example.com/root/api/v1/stores/..%2Fusers%2Fme/webhooks/delivery%2F..%2Fx?email=a%2Bb%40example.com", request.RequestUri!.AbsoluteUri);
         }
 
+        [Fact]
+        public void IsLocalNetwork()
+        {
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("127.0.0.1"));
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("10.0.0.1"));
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("192.168.1.1"));
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("172.16.0.1"));
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("169.254.0.1"));
+            Assert.True(BTCPayServer.Extensions.IsLocalNetwork("::1"));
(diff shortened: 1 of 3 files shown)


r2ckstardev:
Thanks @Kukks. Catching local addresses in the form is useful, but I would add a check where BTCPay actually opens the connection before treating this as complete protection. Checking the address someone typed does not guarantee where the notification will go.

For `99921e3`, these are the changes I would suggest:

1. **Keep the form check, and enforce the rule in the IPN client.** This path uses `BitpayIPNSender` and the `bitpay-ipn` client registered in [BitpayPlugin](https://github.com/btcpayserver/btcpayserver/blob/99921e32bbc8286e5d756648d9a209b90a75934c/BTCPayServer/Plugins/Bitpay/BitpayPlugin.cs#L28-L32). That is where the delivery guard belongs. The earlier bot comment points to `WebhookSender`, which is a different path. Putting the guard in the shared IPN client also covers callers outside this form.

2. **Check the resolved IP and connect to that exact IP.** A `SocketsHttpHandler.ConnectCallback` is a reasonable approach for direct connections. Resolve the hostname, filter the results through a dedicated IPN address policy, and connect without another DNS lookup. Cover IPv4 and IPv6, including mapped addresses, loopback, private, link-local, unique-local, unspecified and shared address space. Keep the original hostname for HTTP and TLS certificate validation. I would keep this policy separate from `IsLocalNetwork`, since that helper also affects HTTPS and certificate-checking decisions elsewhere.

3. **Make redirects and proxies explicit.** Either disable automatic redirects or ensure every destination stays within the same policy. Also account for system/environment proxies: with a proxy, the callback can check the connection to the proxy rather than the final destination. Checking the proxy's IP alone is insufficient. Use a deliberately direct client or enforce the destination policy through the supported proxy path; document the compatibility choice.

4. **Preserve working callbacks.** If the first allowed DNS result is unreachable, try the other already-validated addresses within the cancellation/timeout budget. Do not fall back to an unchecked hostname connection. Decide how existing private receivers are handled; any exception should be controlled by the server administrator, not the invoice caller.

5. **Test delivery as well as address classification.** Add committed tests using the registered IPN client for DNS resolution, redirects, IPv4/IPv6 boundaries, multiple-address fallback, retries and proxy behavior. Include an allowed callback that succeeds, and an invoice-to-callback integration test. Controlled DNS and receivers will make these repeatable without depending on third-party services.

The evidence so far is source review and local component checks. A complete invoice-to-callback run on an unmodified installation is still outstanding, so I would keep the validation claims limited until that passes.

PR #7537, opened 2026-09-01
Validate support URL scheme to prevent stored script injection

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver2/issues/207

`StoreSupportUrl` was stored as a free-text string with no scheme restriction. This PR enforces these validation.
Not so high as this was only possible with can modify store settings.. but still a worthy validation

Cc. @NicolasDorier


TChukwuleta:
Hey Emperor @NicolasDorier .. Can you help rerun this.. 

This  test is failing from a timeout on a monetization test

PR #7539, opened 2026-09-01
Require CanModifyInvoices to activate an invoice payment method

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver2/issues/201

Cc. @NicolasDorier


NicolasDorier:
I don't think this is a bug. It might actually even be OK to be "[Anonymous]". The activation of a payment method happen when you go on the invoice page and change the payment method while this one is on.

image

TChukwuleta:
Ohh yea.. Confirmed.. closing this

PR #7549, opened 2026-09-03
Comment to avoid AI reports on intentional behavior

TChukwuleta:
Resolves https://github.com/btcpayserver/btcpayserver2/issues/149

Resolves https://github.com/btcpayserver/btcpayserver2/issues/148

Cc. @NicolasDorier


TChukwuleta:
CI failing on wrong test

NicolasDorier:
@TChukwuleta this is on purpose, so users can see it is expired. There is a db script running automatically to clean them up every month.

Can you add a comment there so AI doesn't bother us again on this?

TChukwuleta:
yea @NicolasDorier , added the comments on both end

PR #7554, opened 2026-09-03
Make btcpay-host integration opt-in

NicolasDorier:
Host integration is now disabled by default, even when a `btcpay-host` executable is available. Deployments must explicitly enable it with `btcpayhostenabled=true` or `BTCPAY_BTCPAYHOSTENABLED=true`.

When disabled, BTCPay Server does not run `btcpay-host env`. Maintenance and SSH features backed by host commands remain hidden from Server Settings. This keeps host access unavailable unless the deployment has deliberately configured the integration.

The existing `btcpayhostexecutable` setting still selects the executable when host integration is enabled.

Related documentation: [Deployment Host Integration](https://docs.btcpayserver.org/Development/HostIntegration/).

### DIFF
--- a/BTCPayServer.Tests/BTCPayServerTester.cs
+++ b/BTCPayServer.Tests/BTCPayServerTester.cs
@@ -163,7 +163,10 @@ namespace BTCPayServer.Tests
             config.AppendLine($"debuglog=debug.log");
             config.AppendLine($"nocsp={NoCSP.ToString().ToLowerInvariant()}");
             if (btcpayHostExecutable is not null)
+            {
                 config.AppendLine($"btcpayhostexecutable={btcpayHostExecutable}");
+                config.AppendLine("btcpayhostenabled=true");
+            }
 
             if (!String.IsNullOrEmpty(Postgres))
                 config.AppendLine($"postgres=" + Postgres);
--- a/BTCPayServer/Plugins/Maintenance/CheckHostCommandsHostedService.cs
+++ b/BTCPayServer/Plugins/Maintenance/CheckHostCommandsHostedService.cs
@@ -22,6 +22,12 @@ namespace BTCPayServer.Plugins.Maintenance
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
+            if (!processRunner.BTCPayHostEnabled)
+            {
+                Logs.PayServer.LogInformation("Host integration is disabled");
+                return Task.CompletedTask;
+            }
+
             _testingConnection = TestConnection();
             return Task.CompletedTask;
         }
--- a/.agents/skills/btcpayserver-configuration/SKILL.md
+++ b/.agents/skills/btcpayserver-configuration/SKILL.md
@@ -1,0 +1,25 @@
+---
+name: btcpayserver-configuration
+description: Use when adding or reviewing BTCPay Server startup configuration options. Covers command-line registration, defaults, environment variables, tests, and documentation.
+---
+
+# BTCPay Server Configuration Options
+
+When adding a public startup configuration option, treat every supported configuration source as part of the feature.
+
+## Implementation Checklist
+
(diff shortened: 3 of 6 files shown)


BTCPayServer.Tests/BTCPayServerTester.cs
```diff
@@ -164,6 +164,8 @@ public async Task StartAsync()
             config.AppendLine($"nocsp={NoCSP.ToString().ToLowerInvariant()}");
             if (btcpayHostExecutable is not null)
                 config.AppendLine($"btcpayhostexecutable={btcpayHostExecutable}");
+            if (btcpayHostEnabled is not null)
```
NicolasDorier: No need of a separate settings. If `btcpayHostExecutable` is not null, then enable it.

BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -904,6 +905,23 @@ public async Task CanUseSSHService()
             Assert.DoesNotContain("test2", text);
         }
 
+        [Fact]
+        public async Task HostIntegrationIsDisabledByDefault()
```
NicolasDorier: You can remove this test, other tests will fail anyway if not properly implemented.

BTCPayServer.Tests/ServerTester.cs
```diff
@@ -309,8 +309,9 @@ public MailPitClient GetMailPitClient()
             return mailPitClient;
         }
 
-        public async Task InstallHostCommands()
+        public async Task InstallHostCommands(bool? enabled = true)
```
NicolasDorier: you can remove `enabled` as there is no need of ` PayTester.btcpayHostEnabled`

BTCPayServer/Plugins/Maintenance/CheckHostCommandsHostedService.cs
```diff
@@ -22,6 +22,9 @@ public class CheckHostCommandsHostedService(ProcessRunner processRunner, Logs lo
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
+            if (!processRunner.BTCPayHostEnabled)
```
NicolasDorier, line 25: Add a small log message saying the host integration is disabled.

NicolasDorier (commented):
Can you also document btcpayhost configuration in DefaultConfiguration.cs ?

BTCPayServer.Tests/BTCPayServerTester.cs
NicolasDorier: Done. Setting the fake btcpay-host executable now automatically writes btcpayhostenabled=true.

BTCPayServer.Tests/PlaywrightTests.cs
NicolasDorier: Removed the standalone disabled-default test.

BTCPayServer/Plugins/Maintenance/CheckHostCommandsHostedService.cs
NicolasDorier: Added a startup log message when host integration is disabled.

BTCPayServer.Tests/ServerTester.cs
NicolasDorier: Removed the enabled argument and the separate test setting.

NicolasDorier:
Addressed the review: documented `btcpayhostenabled` and `btcpayhostexecutable` in the default configuration template, simplified the test setup, removed the redundant test, and added the disabled startup log.

NicolasDorier:
Added both `--btcpayhostenabled` and `--btcpayhostexecutable` to `CreateCommandLineApplicationCore`. They now appear in `--help`; the generated config template still documents their defaults.

PR #7556, opened 2026-09-04
Require current password for account email changes

NicolasDorier:
Account settings now requires the current password before changing the account email, matching the protection already present in the Greenfield API. This prevents someone with temporary access to an authenticated browser session from redirecting password recovery to another email address.

This is classified as low impact because exploitation requires an existing foothold such as a stolen session cookie, XSS, or access to a shared machine. Thanks to `llen` for responsibly reporting the issue.

The 2.4.4 changelog includes the fix and attribution. An end-to-end browser test covers missing, incorrect, and correct current passwords.

image

### DIFF
--- a/BTCPayServer.Tests/PlaywrightTests.cs
+++ b/BTCPayServer.Tests/PlaywrightTests.cs
@@ -990,6 +990,40 @@ namespace BTCPayServer.Tests
         }
 
         [Fact]
+        public async Task ChangingAccountEmailRequiresCurrentPassword()
+        {
+            await using var s = CreatePlaywrightTester();
+            await s.StartAsync();
+            var oldEmail = await s.RegisterNewUser();
+            var newEmail = $"{RandomUtils.GetUInt256().ToString()[..20]}@example.com";
+            await s.SkipWizard();
+            await s.GoToUrl("/account");
+
+            await s.Page.FillAsync("#Email", newEmail);
+            await s.ClickPagePrimary();
+            await Expect(s.Page.Locator("[data-valmsg-for=CurrentPassword]")).ToContainTextAsync("The current password is not correct.");
+
+            await s.GoToUrl("/account");
+            await Expect(s.Page.Locator("#Email")).ToHaveValueAsync(oldEmail);
+            await s.Page.FillAsync("#Email", newEmail);
+            await s.Page.FillAsync("#CurrentPassword", "incorrect");
+            await s.ClickPagePrimary();
+            await Expect(s.Page.Locator("[data-valmsg-for=CurrentPassword]")).ToContainTextAsync("The current password is not correct.");
+
+            await s.GoToUrl("/account");
+            await Expect(s.Page.Locator("#Email")).ToHaveValueAsync(oldEmail);
+            await s.Page.FillAsync("#Email", newEmail);
+            await s.Page.FillAsync("#CurrentPassword", "123456");
+            await s.ClickPagePrimary();
+            await s.FindAlertMessage(partialText: "Your profile has been updated");
+            await Expect(s.Page.Locator("#Email")).ToHaveValueAsync(newEmail);
+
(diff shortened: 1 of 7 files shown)


BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -989,6 +989,42 @@ public async Task NewUserLogin()
...
+            await s.SkipWizard();
+            await s.GoToUrl("/account");
+
+            await s.Page.FillAsync("#Email", newEmail);
+            await s.ClickPagePrimary();
+            await Expect(s.Page.Locator("[data-valmsg-for=CurrentPassword]")).ToContainTextAsync("The current password is not correct.");
+
+            await s.GoToUrl("/account");
+            await s.Page.WaitForLoadStateAsync();
+            Assert.Equal(oldEmail, await s.Page.Locator("#Email").InputValueAsync());
```
NicolasDorier: Never use this parttern.
```csharp
   await s.Page.WaitForLoadStateAsync();
            Assert.Equal(oldEmail, await s.Page.Locator("#Email").InputValueAsync());
```

Instead, use `Assert` API and add `using static Microsoft.Playwright.Assertions;` in namespace to assert a value.

This should be part of the btcpayserver skills for writing test. Update the skills if that's not the case.

BTCPayServer.Tests/PlaywrightTests.cs
```diff
@@ -989,6 +989,42 @@ public async Task NewUserLogin()
...
+            await s.Page.FillAsync("#Email", newEmail);
+            await s.Page.FillAsync("#CurrentPassword", "123456");
+            await s.ClickPagePrimary();
+            await s.FindAlertMessage(partialText: "Your profile has been updated");
+            Assert.Equal(newEmail, await s.Page.Locator("#Email").InputValueAsync());
```
NicolasDorier: use Assert API

BTCPayServer/Controllers/UIManageController.cs
```diff
@@ -154,7 +154,19 @@ public async Task<IActionResult> Index(IndexViewModel model, [FromForm] bool Rem
...
-            if (model.Email != email)
+            var blob = user.GetBlob() ?? new();
+            var setNewEmail = model.Email != email && ModelState.IsValid;
+            if (setNewEmail && (string.IsNullOrEmpty(model.CurrentPassword) ||
+                                !await _userManager.CheckPasswordAsync(user, model.CurrentPassword)))
+            {
+                ModelState.AddModelError(nameof(model.CurrentPassword), StringLocalizer["The current password is not correct."].Value);
+                model.EmailConfirmed = user.EmailConfirmed;
+                model.RequiresEmailConfirmation = user.RequiresEmailConfirmation;
+                model.ImageUrl = string.IsNullOrEmpty(blob.ImageUrl) ? null : await _uriResolver.Resolve(Request.GetAbsoluteRootUri(), UnresolvedUri.Create(blob.ImageUrl));
```
NicolasDorier: Can you refactor the code so what is in the `Index` is reused?

NicolasDorier: Refactored the GET action and validation-error path to share GetIndexViewModel.

BTCPayServer.Tests/PlaywrightTests.cs
NicolasDorier: Updated all value checks to use Playwright Expect/ToHaveValue without explicit load-state waits. I also documented this exact pattern in the existing playwright-test-patterns skill and linked that skill from AGENTS.md.

BTCPayServer.Tests/PlaywrightTests.cs
NicolasDorier: Updated this assertion to use Expect(...).ToHaveValueAsync(...).

PR #7561, opened 2026-09-05
Harden API key storage

NicolasDorier:
API key secrets are currently stored in the database, so a leaked database or backup exposes usable credentials. This change stores hashes and derived IDs instead, and clears plaintext secrets shortly after creation. Existing Greenfield keys continue to work after migration.

We are also now tracking the creation date of the API Key.

## Breaking changes

These routes now accept `apiKeyId` instead of the API key secret:

- `DELETE /api/v1/api-keys/{apiKeyId}`
- `DELETE /api/v1/users/{idOrEmail}/api-keys/{apiKeyId}`

Legacy BitPay Basic-auth API keys are deleted and no longer supported.

API key responses now include `id`, and the account screen displays the ID and the `akid_` prefix. (for example, `akid_4743fdafe0c5807a`)

### DIFF
--- a/BTCPayServer.Client/BTCPayServerClient.APIKeys.cs
+++ b/BTCPayServer.Client/BTCPayServerClient.APIKeys.cs
@@ -30,15 +30,15 @@ public partial class BTCPayServerClient
         await SendHttpRequest($"api/v1/api-keys/current", null, HttpMethod.Delete, token);
     }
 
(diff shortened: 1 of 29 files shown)


NicolasDorier:
@coderabbitai review

NicolasDorier:
@coderabbitai review

